PR #15, opened 2022-10-06
Add parsing logic for neural query

### LINKED ISSUE #14: Create custom neural query
Create a new query type: "neural". Internally, it will use ML-Commons to take a query string and create a vector from it. From there, it should build a k-NN query.

Interface will look like:
```
GET <index_name>/_search
{
  "size": int,
  "query": {
    "neural": {
      "<vector_field>": {
        "query_text": "string",
        "model_id": "string",
        "k": int
      }
    }
  }
}
```

**vector_field** — Field to execute k-NN query against
**query_text** — (string) Query text to be used to produce queries against.
**model_id** — (string) ID of model to do vector to encoding.
**k** — (int) Number of results to return from the k-NN search

For more details, refer to https://github.com/opensearch-project/neural-search/issues/11#issue-1398115115

jmazanec15:
### Description
Adds new query builder to plugin: NeuralQueryBuilder. Adds logic to parse the user provided parameters for the NeuralQueryBuilder as well as a few other smaller methods. Query text to dense vector logic will be added in a future PR

The query has the following structure:
```
{
  "neural": {
    "<VECTOR_FIELD>": {
    "query_text": "string",
    "model_id": "string",
    "k": int,
    "boost": float, (optional - comes with default query builder)
    "_name": "string", (optional - comes with default query builder)
  }
}
```

Adds unit tests as well as test utility class

### Issues Resolved
#14  (partially resolved)

### Check List
    - [X] All tests pass

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -119,6 +119,7 @@ allprojects {
 }
 
 repositories {
+    mavenLocal()
     maven { url "https://aws.oss.sonatype.org/content/repositories/snapshots" }
     mavenCentral()
     maven { url "https://plugins.gradle.org/m2/" }
--- a/src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
@@ -124,12 +124,13 @@ public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder>
         if (parser.currentToken() != XContentParser.Token.END_OBJECT) {
             throw new ParsingException(
                 parser.getTokenLocation(),
-                String.format(
-                    "[%s] query doesn't support multiple fields, found [%s] and [%s]",
-                    NAME,
-                    neuralQueryBuilder.fieldName(),
-                    parser.currentName()
-                )
+                "["
+                    + NAME
+                    + "] query doesn't support multiple fields, found ["
+                    + neuralQueryBuilder.fieldName()
+                    + "] and ["
+                    + parser.currentName()
+                    + "]"
             );
         }
         requireValue(neuralQueryBuilder.queryText(), "Query text must be provided for neural query");
--- a/src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
@@ -52,7 +52,7 @@ public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder>
     @VisibleForTesting
     static final ParseField K_FIELD = new ParseField("k");
 
-    private static int DEFAULT_K = 10;
+    private static final int DEFAULT_K = 10;
 
     private String fieldName;
     private String queryText;
--- a/src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
@@ -122,7 +122,15 @@ public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder>
         parseQueryParams(parser, neuralQueryBuilder);
         parser.nextToken();
         if (parser.currentToken() != XContentParser.Token.END_OBJECT) {
-            throw new ParsingException(parser.getTokenLocation(), "Token must be END_OBJECT");
+            throw new ParsingException(
+                parser.getTokenLocation(),
+                String.format(
+                    "[%s] query doesn't support multiple fields, found [%s] and [%s]",
+                    NAME,
+                    neuralQueryBuilder.fieldName(),
+                    parser.currentName()
+                )
+            );
         }
         requireValue(neuralQueryBuilder.queryText(), "Query text must be provided for neural query");
         requireValue(neuralQueryBuilder.fieldName(), "Field name must be provided for neural query");
--- a/src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
@@ -1,0 +1,227 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.neuralsearch.plugin.query;
+
+import java.io.IOException;
+
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.extern.log4j.Log4j2;
+
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+import org.apache.lucene.search.Query;
+import org.opensearch.common.ParseField;
+import org.opensearch.common.ParsingException;
+import org.opensearch.common.io.stream.StreamInput;
+import org.opensearch.common.io.stream.StreamOutput;
+import org.opensearch.common.xcontent.XContentBuilder;
+import org.opensearch.common.xcontent.XContentParser;
+import org.opensearch.index.mapper.NumberFieldMapper;
+import org.opensearch.index.query.AbstractQueryBuilder;
+import org.opensearch.index.query.QueryShardContext;
+
+/**
+ * NeuralQueryBuilder is responsible for producing "neural" query types. A "neural" query type is a wrapper around a
+ * k-NN vector query. It uses a ML language model to produce a dense vector from a query string that is then used as
+ * the query vector for the k-NN search.
+ */
+
+@Log4j2
+@Getter
+@NoArgsConstructor
+public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder> {
+
+    public static final String NAME = "neural";
+
+    static final ParseField QUERY_TEXT_FIELD = new ParseField("query_text");
+
+    static final ParseField MODEL_ID_FIELD = new ParseField("model_id");
+
+    static final ParseField K_FIELD = new ParseField("k");
+
+    private String fieldName;
+    private String queryText;
+    private String modelId;
+    private int k;
+
+    /**
+     * Set the fieldName this query will be executed against
+     *
+     * @param fieldName name of k-NN vector field that query will be executed against
+     * @return this
+     */
+    public NeuralQueryBuilder fieldName(String fieldName) {
+        this.fieldName = fieldName;
+        return this;
+    }
+
+    /**
+     * Set the queryText that will be translated into the dense query vector used for k-NN search.
+     *
+     * @param queryText Text of a query that should be translated to a dense vector
+     * @return this
+     */
+    public NeuralQueryBuilder queryText(String queryText) {
+        this.queryText = queryText;
+        return this;
+    }
+
+    /**
+     * Set the modelId that should produce the dense query vector
+     *
+     * @param modelId ID of model to produce query vector
+     * @return this
+     */
+    public NeuralQueryBuilder modelId(String modelId) {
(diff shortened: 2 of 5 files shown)


src/test/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilderTests.java
```diff
@@ -0,0 +1,326 @@
...
+public class NeuralQueryBuilderTests extends OpenSearchTestCase {
+
+    private static final String FIELD_NAME = "testField";
+    private static final String QUERY_TEXT = "Hello world!";
+    private static final String MODEL_ID = "mfgfgdsfgfdgsde";
+    private static final int K = 10;
+    private static final float BOOST = 1.8f;
+    private static final String QUERY_NAME = "queryName";
+
+    public void testFromXContent_valid_withDefaults() throws IOException {
```
navneet1v, line 36: see if we want to use @SneakyThrows

src/test/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilderTests.java
```diff
@@ -0,0 +1,326 @@
...
+        XContentParser contentParser = createParser(xContentBuilder);
+        NeuralQueryBuilder neuralQueryBuilder = NeuralQueryBuilder.fromXContent(contentParser);
+
+        assertEquals(FIELD_NAME, neuralQueryBuilder.getFieldName());
+        assertEquals(QUERY_TEXT, neuralQueryBuilder.getQueryText());
+        assertEquals(MODEL_ID, neuralQueryBuilder.getModelId());
+        assertEquals(K, neuralQueryBuilder.getK());
+    }
+
+    public void testFromXContent_valid_withOptionals() throws IOException {
```
navneet1v, line 64: see if we want to rename the test function to something like testFromXConent_whenXXXX_thenYYYYY for all the functions. This can help in just getting to know what is being tested with what just be reading the name of the fuction.

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
```diff
@@ -0,0 +1,227 @@
...
+        this.modelId = in.readString();
+        this.k = in.readInt();
+    }
+
+    @Override
+    protected void doWriteTo(StreamOutput out) throws IOException {
+        out.writeString(this.fieldName);
+        out.writeString(this.queryText);
+        out.writeString(this.modelId);
+        out.writeInt(this.k);
```
navneet1v, line 114: see if want to use writeVInt() instead of Int. VInt uses less bytes than int.

src/test/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilderTests.java
```diff
@@ -0,0 +1,326 @@
...
+        String modelId1 = "model-1";
+        String modelId2 = "model-2";
+        float boost1 = 1.8f;
+        float boost2 = 3.8f;
+        String queryName1 = "query-1";
+        String queryName2 = "query-2";
+        int k1 = 1;
+        int k2 = 2;
+
+        NeuralQueryBuilder neuralQueryBuilder1 = new NeuralQueryBuilder().fieldName(fieldName1)
```
navneet1v, line 230: rather than using 1,2,3,4,5 can we rename them in fashion with which just by variable name we can tell what is missing/what default values we have used

jmazanec15:
> ```
> {
>   "neural": {
>     "<VECTOR_FIELD>": {
>     "query_text": "string",
>     "model_id": "string",
>     "k": int,
>     "boost": float, (optional - comes with default query builder)
>     "_name": "string", (optional - comes with default query builder)
>   }
> }
> ```
> 
> From this query example, user could run neural search on multiple `VECTOR_FIELD`s , Is that true?

No, user would be able to run one neural search on one field per query.

ylwu-amzn:
> > ```
> > {
> >   "neural": {
> >     "<VECTOR_FIELD>": {
> >     "query_text": "string",
> >     "model_id": "string",
> >     "k": int,
> >     "boost": float, (optional - comes with default query builder)
> >     "_name": "string", (optional - comes with default query builder)
> >   }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > From this query example, user could run neural search on multiple `VECTOR_FIELD`s , Is that true?
> 
> No, user would be able to run one neural search on one field per query.

If we can only support one filed per neural search, how about change to this format to avoid confusion
```
{
  "neural": {
    "vector_field": "string",
    "query_text": "string",
    "model_id": "string",
    "k": int,
    "boost": float, (optional - comes with default query builder)
    "_name": "string", (optional - comes with default query builder)
  }
}
```

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
```diff
@@ -0,0 +1,227 @@
...
+    /**
+     * Set the fieldName this query will be executed against
+     *
+     * @param fieldName name of k-NN vector field that query will be executed against
+     * @return this
+     */
+    public NeuralQueryBuilder fieldName(String fieldName) {
+        this.fieldName = fieldName;
+        return this;
+    }
```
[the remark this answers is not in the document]
jmazanec15: I thought about this, but I didnt see how it would work, given we are extending a builder class. Instead, I followed how AbstractQueryBuilder handled it for the boost field: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/AbstractQueryBuilder.java#L172

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
```diff
@@ -0,0 +1,227 @@
...
+@NoArgsConstructor
+public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder> {
+
+    public static final String NAME = "neural";
+
+    static final ParseField QUERY_TEXT_FIELD = new ParseField("query_text");
+
+    static final ParseField MODEL_ID_FIELD = new ParseField("model_id");
+
+    static final ParseField K_FIELD = new ParseField("k");
```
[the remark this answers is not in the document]
jmazanec15: Yes, it was for testing purposes. I will add annotation.

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
```diff
@@ -0,0 +1,227 @@
...
+        equalsBuilder.append(queryText, obj.queryText);
+        equalsBuilder.append(modelId, obj.modelId);
+        equalsBuilder.append(k, obj.k);
+        return equalsBuilder.isEquals();
+    }
+
+    @Override
+    protected int doHashCode() {
+        return new HashCodeBuilder().append(fieldName).append(queryText).append(modelId).append(k).toHashCode();
+    }
```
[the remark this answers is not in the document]
jmazanec15: Reason for this is that we are not implementing "hashCode" and "equals" method, but instead "doHashCode" and "doEquals".

src/test/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilderTests.java
jmazanec15: Will update

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
jmazanec15: Will update

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
```diff
@@ -0,0 +1,227 @@
...
+        parseQueryParams(parser, neuralQueryBuilder);
+        parser.nextToken();
+        if (parser.currentToken() != XContentParser.Token.END_OBJECT) {
+            throw new ParsingException(parser.getTokenLocation(), "Token must be END_OBJECT");
+        }
+
+        requireValue(neuralQueryBuilder.getQueryText(), "Query text must be provided for neural query");
+        requireValue(neuralQueryBuilder.getFieldName(), "Field name must be provided for neural query");
+        requireValue(neuralQueryBuilder.getModelId(), "Model ID must be provided for neural query");
+        requireValue(neuralQueryBuilder.getK(), "K must be provided for neural query");
```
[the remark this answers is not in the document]
jmazanec15: I was thinking about this. Yes, I think we can make it 10 by default.

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
```diff
@@ -0,0 +1,227 @@
...
+                throw new ParsingException(
+                    parser.getTokenLocation(),
+                    "[" + NAME + "] unknown token [" + token + "] after [" + currentFieldName + "]"
+                );
+            }
+        }
+    }
+
+    @Override
+    protected Query doToQuery(QueryShardContext queryShardContext) throws IOException {
```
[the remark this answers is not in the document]
jmazanec15: "toQuery" will build a Query object using the queryShardContext and the queryBuilder. "doToQuery" is used internally to produce the query in [AbstractQueryBuilder](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/AbstractQueryBuilder.java#L116).

Its null right now because we will be implementing in future PR. Ill add a comment.

src/test/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilderTests.java
jmazanec15: Sure, will update

src/test/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilderTests.java
jmazanec15: Will update

navneet1v (approved):

ylwu-amzn:
> @ylwu-amzn Yes, I see what you mean. The main reason that I opted to make the field name the key was to follow OpenSearch convention.
> 
> For instance, for the text matching query types, the [field is the key](https://opensearch.org/docs/latest/opensearch/ux/). Similarly, this is how it is done for [k-NN](https://opensearch.org/docs/latest/search-plugins/knn/approximate-knn/).

Thanks, that makes sense. Make sure the error message is readable to tell user we only support one query field. like this one
```
POST /_search
{
  "query": {
    "match_phrase_prefix": {
      "message": {
        "query": "quick brown f"
      },
      "k1": {
        "query": "test"
      }
    }
  }
}
```
```
{
  "error" : {
    "root_cause" : [
      {
        "type" : "parsing_exception",
        "reason" : "[match_phrase_prefix] query doesn't support multiple fields, found [message] and [k1]",
        "line" : 7,
        "col" : 13
      }
    ],
    "type" : "parsing_exception",
    "reason" : "[match_phrase_prefix] query doesn't support multiple fields, found [message] and [k1]",
    "line" : 7,
    "col" : 13
  },
  "status" : 400
}

```

jmazanec15:
@ylwu-amzn Makes sense. Just updated.

dhrubo-os (approved):

build.gradle
```diff
@@ -119,6 +119,7 @@ allprojects {
 }
 
 repositories {
+    mavenLocal()
```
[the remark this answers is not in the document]
navneet1v: I know it is required for local development, but ideally we should not put this, as it can lead to errors with different people can different versions. 

These should be added as a part of Development.md file that we need to enable mavenLocal(). But good for now.

Issue #1, opened 2022-09-24
Setup of Neural Search Plugin


navneet1v:
### Description
The task involves the setting up of the Neural Search plugin for development.

### Tasks:

- [x] Setup the code for the repo
- [x] Update the Description of the repo
- [ ] Setup the Developer guide
- [x] Add the merge policy like: Don't merge the code without approval, 2 person approval.
- [x] Setup the labels.
- [x] Setup the 2.x branches
- [x] Add update version task 
- [x] Show Jacoco reports when Pull request is created
- [x] Run CI checks periodically
- [x] Add the badges to Readme.

martin-gaievski:
I'm pretty sure that's done, closing

PR #18, opened 2022-10-12
Add text embedding processor to neural search

zane-neo:
Signed-off-by: Zan Niu <zaniu@amazon.com>

### Description

### Issues Resolved

### Check List
    - [x] All tests pass

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,9 @@
  * Learn more about Gradle by exploring our samples at https://docs.gradle.org/7.5.1/samples
  * This project uses @Incubating APIs which are subject to change.
  */
+
 import org.opensearch.gradle.test.RestIntegTestTask
+
 import java.util.concurrent.Callable
 
 apply plugin: 'java'
@@ -137,6 +139,9 @@ dependencies {
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-knn', version: "${opensearch_build}"
     compileOnly fileTree(dir: knnJarDirectory, include: '*.jar')
     api group: 'org.opensearch', name:'opensearch-ml-client', version: "${opensearch_build}"
+    implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.10'
+
+    testImplementation group: 'commons-io', name: 'commons-io', version: '2.11.0'
 }
 
 // From maven, we can get the k-NN plugin as a zip. In order to add the jar to the classpath, we need to unzip the
--- a/build.gradle
+++ b/build.gradle
@@ -140,8 +140,6 @@ dependencies {
     compileOnly fileTree(dir: knnJarDirectory, include: '*.jar')
     api group: 'org.opensearch', name:'opensearch-ml-client', version: "${opensearch_build}"
     implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.10'
-
-    testImplementation group: 'commons-io', name: 'commons-io', version: '2.11.0'
 }
 
 // From maven, we can get the k-NN plugin as a zip. In order to add the jar to the classpath, we need to unzip the
--- a/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
@@ -5,15 +5,10 @@
 
 package org.opensearch.neuralsearch.ml;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.stream.Collectors;
-
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.log4j.Log4j2;
-
+import org.opensearch.action.ActionFuture;
 import org.opensearch.action.ActionListener;
 import org.opensearch.ml.client.MachineLearningNodeClient;
 import org.opensearch.ml.common.FunctionName;
@@ -21,10 +16,17 @@ import org.opensearch.ml.common.dataset.MLInputDataset;
 import org.opensearch.ml.common.dataset.TextDocsInputDataSet;
 import org.opensearch.ml.common.input.MLInput;
 import org.opensearch.ml.common.model.MLModelTaskType;
+import org.opensearch.ml.common.output.MLOutput;
 import org.opensearch.ml.common.output.model.ModelResultFilter;
 import org.opensearch.ml.common.output.model.ModelTensor;
 import org.opensearch.ml.common.output.model.ModelTensorOutput;
 import org.opensearch.ml.common.output.model.ModelTensors;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.stream.Collectors;
 
 /**
  * This class will act as an abstraction on the MLCommons client for accessing the ML Capabilities
@@ -99,23 +101,54 @@ public class MLCommonsClientAccessor {
         @NonNull final List<String> inputText,
         @NonNull final ActionListener<List<List<Float>>> listener
     ) {
-        final ModelResultFilter modelResultFilter = new ModelResultFilter(false, true, targetResponseFilters, null);
-        final MLInputDataset inputDataset = new TextDocsInputDataSet(inputText, modelResultFilter);
-        final MLInput mlInput = new MLInput(FunctionName.TEXT_EMBEDDING, null, inputDataset, MLModelTaskType.TEXT_EMBEDDING);
-        final List<List<Float>> vector = new ArrayList<>();
-
+        MLInput mlInput = createMLInput(targetResponseFilters, inputText);
         mlClient.predict(modelId, mlInput, ActionListener.wrap(mlOutput -> {
-            final ModelTensorOutput modelTensorOutput = (ModelTensorOutput) mlOutput;
-            final List<ModelTensors> tensorOutputList = modelTensorOutput.getMlModelOutputs();
-            for (final ModelTensors tensors : tensorOutputList) {
-                final List<ModelTensor> tensorsList = tensors.getMlModelTensors();
-                for (final ModelTensor tensor : tensorsList) {
-                    vector.add(Arrays.stream(tensor.getData()).map(value -> (Float) value).collect(Collectors.toList()));
-                }
-            }
+            final List<List<Float>> vector = buildVectorFromResponse(mlOutput);
             log.debug("Inference Response for input sentence {} is : {} ", inputText, vector);
             listener.onResponse(vector);
         }, listener::onFailure));
+    }
+
+    /**
+     * Abstraction to call predict function of api of MLClient with provided targetResponseFilters. It uses the
+     * custom model provided as modelId and run the {@link MLModelTaskType#TEXT_EMBEDDING}. The return will be sent
+     * using the actionListener which will have a {@link List} of {@link List} of {@link Float} in the order of
+     * inputText. We are not making this function generic enough to take any function or TaskType as currently we need
+     * to run only TextEmbedding tasks only. Please note this method is a blocking method, use this only when the processing
+     * needs block waiting for response, otherwise please use {@link #inferenceSentences(String, List, ActionListener)}
+     * instead.
+     * @param modelId {@link String}
+     * @param inputText {@link List} of {@link String} on which inference needs to happen.
+     * @return {@link List} of {@link List} of {@link String} represents the text embedding vector result.
+     * @throws ExecutionException If the underlying task failed, this exception will be thrown in the future.get().
+     * @throws InterruptedException If the thread is interrupted, this will be thrown.
+     */
+    public List<List<Float>> inferenceSentences(@NonNull final String modelId, @NonNull final List<String> inputText)
+        throws ExecutionException, InterruptedException {
+        final MLInput mlInput = createMLInput(TARGET_RESPONSE_FILTERS, inputText);
+        final ActionFuture<MLOutput> outputActionFuture = mlClient.predict(modelId, mlInput);
+        final List<List<Float>> vector = buildVectorFromResponse(outputActionFuture.get());
+        log.debug("Inference Response for input sentence {} is : {} ", inputText, vector);
+        return vector;
+    }
+
+    private MLInput createMLInput(final List<String> targetResponseFilters, List<String> inputText) {
+        final ModelResultFilter modelResultFilter = new ModelResultFilter(false, true, targetResponseFilters, null);
+        final MLInputDataset inputDataset = new TextDocsInputDataSet(inputText, modelResultFilter);
+        return new MLInput(FunctionName.TEXT_EMBEDDING, null, inputDataset, MLModelTaskType.TEXT_EMBEDDING);
+    }
+
+    private List<List<Float>> buildVectorFromResponse(MLOutput mlOutput) {
+        final List<List<Float>> vector = new ArrayList<>();
+        final ModelTensorOutput modelTensorOutput = (ModelTensorOutput) mlOutput;
+        final List<ModelTensors> tensorOutputList = modelTensorOutput.getMlModelOutputs();
+        for (final ModelTensors tensors : tensorOutputList) {
+            final List<ModelTensor> tensorsList = tensors.getMlModelTensors();
+            for (final ModelTensor tensor : tensorsList) {
+                vector.add(Arrays.stream(tensor.getData()).map(value -> (Float) value).collect(Collectors.toList()));
+            }
+        }
+        return vector;
     }
 
 }
--- a/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
@@ -5,9 +5,16 @@
 
 package org.opensearch.neuralsearch.ml;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.stream.Collectors;
+
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.log4j.Log4j2;
+
 import org.opensearch.action.ActionFuture;
 import org.opensearch.action.ActionListener;
 import org.opensearch.ml.client.MachineLearningNodeClient;
@@ -21,12 +28,6 @@ import org.opensearch.ml.common.output.model.ModelResultFilter;
 import org.opensearch.ml.common.output.model.ModelTensor;
 import org.opensearch.ml.common.output.model.ModelTensorOutput;
 import org.opensearch.ml.common.output.model.ModelTensors;
-
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.concurrent.ExecutionException;
-import java.util.stream.Collectors;
 
 /**
  * This class will act as an abstraction on the MLCommons client for accessing the ML Capabilities
--- a/src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
+++ b/src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
@@ -42,6 +42,8 @@ import org.opensearch.watcher.ResourceWatcherService;
  */
 public class NeuralSearch extends Plugin implements ActionPlugin, SearchPlugin, IngestPlugin {
 
+    private MLCommonsClientAccessor clientAccessor;
+
     @Override
     public Collection<Object> createComponents(
         final Client client,
@@ -56,8 +58,6 @@ public class NeuralSearch extends Plugin implements ActionPlugin, SearchPlugin, 
         final IndexNameExpressionResolver indexNameExpressionResolver,
         final Supplier<RepositoriesService> repositoriesServiceSupplier
     ) {
-        final MachineLearningNodeClient machineLearningNodeClient = new MachineLearningNodeClient(client);
-        final MLCommonsClientAccessor clientAccessor = new MLCommonsClientAccessor(machineLearningNodeClient);
         NeuralQueryBuilder.initialize(clientAccessor);
         return List.of(clientAccessor);
     }
@@ -80,8 +80,7 @@ public class NeuralSearch extends Plugin implements ActionPlugin, SearchPlugin, 
 
     @Override
     public Map<String, Processor.Factory> getProcessors(Processor.Parameters parameters) {
-        final MachineLearningNodeClient machineLearningNodeClient = new MachineLearningNodeClient(parameters.client);
-        final MLCommonsClientAccessor clientAccessor = new MLCommonsClientAccessor(machineLearningNodeClient);
+        clientAccessor = new MLCommonsClientAccessor(new MachineLearningNodeClient(parameters.client));
         return Collections.singletonMap(TextEmbeddingProcessor.TYPE, new TextEmbeddingProcessorFactory(clientAccessor));
     }
 }
--- a/src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
+++ b/src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
@@ -8,6 +8,7 @@ package org.opensearch.neuralsearch.plugin;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.function.Supplier;
 
 import org.opensearch.action.ActionRequest;
@@ -19,12 +20,16 @@ import org.opensearch.common.io.stream.NamedWriteableRegistry;
 import org.opensearch.common.xcontent.NamedXContentRegistry;
 import org.opensearch.env.Environment;
 import org.opensearch.env.NodeEnvironment;
+import org.opensearch.ingest.Processor;
 import org.opensearch.ml.client.MachineLearningNodeClient;
 import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
 import org.opensearch.neuralsearch.plugin.query.NeuralQueryBuilder;
+import org.opensearch.neuralsearch.processor.TextEmbeddingProcessor;
+import org.opensearch.neuralsearch.processor.factory.TextEmbeddingProcessorFactory;
 import org.opensearch.neuralsearch.transport.MLPredictAction;
 import org.opensearch.neuralsearch.transport.MLPredictTransportAction;
 import org.opensearch.plugins.ActionPlugin;
+import org.opensearch.plugins.IngestPlugin;
 import org.opensearch.plugins.Plugin;
 import org.opensearch.plugins.SearchPlugin;
 import org.opensearch.repositories.RepositoriesService;
@@ -35,7 +40,7 @@ import org.opensearch.watcher.ResourceWatcherService;
 /**
  * Neural Search plugin class
  */
-public class NeuralSearch extends Plugin implements ActionPlugin, SearchPlugin {
+public class NeuralSearch extends Plugin implements ActionPlugin, SearchPlugin, IngestPlugin {
 
     @Override
     public Collection<Object> createComponents(
@@ -71,5 +76,12 @@ public class NeuralSearch extends Plugin implements ActionPlugin, SearchPlugin {
         return Collections.singletonList(
             new QuerySpec<>(NeuralQueryBuilder.NAME, NeuralQueryBuilder::new, NeuralQueryBuilder::fromXContent)
         );
+    }
+
+    @Override
+    public Map<String, Processor.Factory> getProcessors(Processor.Parameters parameters) {
+        final MachineLearningNodeClient machineLearningNodeClient = new MachineLearningNodeClient(parameters.client);
+        final MLCommonsClientAccessor clientAccessor = new MLCommonsClientAccessor(machineLearningNodeClient);
+        return Collections.singletonMap(TextEmbeddingProcessor.TYPE, new TextEmbeddingProcessorFactory(clientAccessor));
     }
 }
--- a/src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
+++ b/src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
@@ -81,6 +81,6 @@ public class NeuralSearch extends Plugin implements ActionPlugin, SearchPlugin, 
     @Override
     public Map<String, Processor.Factory> getProcessors(Processor.Parameters parameters) {
         clientAccessor = new MLCommonsClientAccessor(new MachineLearningNodeClient(parameters.client));
-        return Collections.singletonMap(TextEmbeddingProcessor.TYPE, new TextEmbeddingProcessorFactory(clientAccessor));
+        return Collections.singletonMap(TextEmbeddingProcessor.TYPE, new TextEmbeddingProcessorFactory(clientAccessor, parameters.env));
     }
 }
--- a/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
@@ -73,8 +73,8 @@ public class TextEmbeddingProcessor extends AbstractProcessor {
 
     @Override
     public IngestDocument execute(IngestDocument ingestDocument) {
-        validateEmbeddingFieldsValue(ingestDocument, fieldMap);
-        Map<String, Object> knnMap = buildMapWithKnnKeyAndOriginalValue(ingestDocument, fieldMap);
+        validateEmbeddingFieldsValue(ingestDocument);
+        Map<String, Object> knnMap = buildMapWithKnnKeyAndOriginalValue(ingestDocument);
         try {
             List<List<Float>> vectors = mlCommonsClientAccessor.inferenceSentences(this.modelId, createInferenceList(knnMap));
             appendVectorFieldsToDocument(ingestDocument, knnMap, vectors);
@@ -122,7 +122,7 @@ public class TextEmbeddingProcessor extends AbstractProcessor {
     }
 
     @VisibleForTesting
-    Map<String, Object> buildMapWithKnnKeyAndOriginalValue(IngestDocument ingestDocument, Map<String, Object> fieldMap) {
+    Map<String, Object> buildMapWithKnnKeyAndOriginalValue(IngestDocument ingestDocument) {
         Map<String, Object> sourceAndMetadataMap = ingestDocument.getSourceAndMetadata();
         Map<String, Object> mapWithKnnKeys = new LinkedHashMap<>();
         for (Map.Entry<String, Object> fieldMapEntry : fieldMap.entrySet()) {
@@ -228,9 +228,9 @@ public class TextEmbeddingProcessor extends AbstractProcessor {
         return numbers;
     }
 
-    private static void validateEmbeddingFieldsValue(IngestDocument ingestDocument, Map<String, Object> embeddingFields) {
+    private void validateEmbeddingFieldsValue(IngestDocument ingestDocument) {
         Map<String, Object> sourceAndMetadataMap = ingestDocument.getSourceAndMetadata();
-        for (Map.Entry<String, Object> embeddingFieldsEntry : embeddingFields.entrySet()) {
+        for (Map.Entry<String, Object> embeddingFieldsEntry : fieldMap.entrySet()) {
             Object sourceValue = sourceAndMetadataMap.get(embeddingFieldsEntry.getKey());
             if (sourceValue != null) {
                 String sourceKey = embeddingFieldsEntry.getKey();
--- a/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
@@ -11,11 +11,14 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ExecutionException;
+import java.util.function.Supplier;
 import java.util.stream.IntStream;
 
 import lombok.extern.log4j.Log4j2;
 
 import org.apache.commons.lang3.StringUtils;
+import org.opensearch.env.Environment;
+import org.opensearch.index.mapper.MapperService;
 import org.opensearch.ingest.AbstractProcessor;
 import org.opensearch.ingest.IngestDocument;
 import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
@@ -43,12 +46,15 @@ public class TextEmbeddingProcessor extends AbstractProcessor {
 
     private final MLCommonsClientAccessor mlCommonsClientAccessor;
 
+    private final Environment environment;
+
     public TextEmbeddingProcessor(
         String tag,
         String description,
         String modelId,
         Map<String, Object> fieldMap,
-        MLCommonsClientAccessor clientAccessor
+        MLCommonsClientAccessor clientAccessor,
+        Environment environment
     ) {
         super(tag, description);
         if (StringUtils.isBlank(modelId)) throw new IllegalArgumentException("model_id is null or empty, can not process it");
@@ -57,6 +63,7 @@ public class TextEmbeddingProcessor extends AbstractProcessor {
         this.modelId = modelId;
         this.fieldMap = fieldMap;
         this.mlCommonsClientAccessor = clientAccessor;
+        this.environment = environment;
     }
 
     private void validateEmbeddingConfiguration(Map<String, Object> fieldMap) {
@@ -236,7 +243,7 @@ public class TextEmbeddingProcessor extends AbstractProcessor {
                 String sourceKey = embeddingFieldsEntry.getKey();
                 Class<?> sourceValueClass = sourceValue.getClass();
                 if (List.class.isAssignableFrom(sourceValueClass) || Map.class.isAssignableFrom(sourceValueClass)) {
-                    validateNestedTypeValue(sourceKey, sourceValue);
+                    validateNestedTypeValue(sourceKey, sourceValue, () -> 1);
                 } else if (!String.class.isAssignableFrom(sourceValueClass)) {
                     throw new IllegalArgumentException("field [" + sourceKey + "] is neither string nor nested type, can not process it");
                 } else if (StringUtils.isBlank(sourceValue.toString())) {
@@ -247,11 +254,17 @@ public class TextEmbeddingProcessor extends AbstractProcessor {
     }
 
     @SuppressWarnings({ "rawtypes", "unchecked" })
-    private static void validateNestedTypeValue(String sourceKey, Object sourceValue) {
-        if ((List.class.isAssignableFrom(sourceValue.getClass()))) {
+    private void validateNestedTypeValue(String sourceKey, Object sourceValue, Supplier<Integer> maxDepthSupplier) {
+        int maxDepth = maxDepthSupplier.get();
+        if (maxDepth > MapperService.INDEX_MAPPING_DEPTH_LIMIT_SETTING.get(environment.settings())) {
+            throw new IllegalArgumentException("map type field [" + sourceKey + "] reached max depth limit, can not process it");
+        } else if ((List.class.isAssignableFrom(sourceValue.getClass()))) {
             validateListTypeValue(sourceKey, sourceValue);
         } else if (Map.class.isAssignableFrom(sourceValue.getClass())) {
-            ((Map) sourceValue).values().stream().filter(Objects::nonNull).forEach(x -> validateNestedTypeValue(sourceKey, x));
+            ((Map) sourceValue).values()
+                .stream()
+                .filter(Objects::nonNull)
+                .forEach(x -> validateNestedTypeValue(sourceKey, x, () -> maxDepth + 1));
         } else if (!String.class.isAssignableFrom(sourceValue.getClass())) {
             throw new IllegalArgumentException("map type field [" + sourceKey + "] has non-string type, can not process it");
         } else if (StringUtils.isBlank(sourceValue.toString())) {
--- a/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
@@ -5,20 +5,23 @@
 
 package org.opensearch.neuralsearch.processor;
 
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.ImmutableMap;
-import lombok.extern.log4j.Log4j2;
-import org.apache.commons.lang3.StringUtils;
-import org.opensearch.ingest.AbstractProcessor;
-import org.opensearch.ingest.IngestDocument;
-import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
-
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ExecutionException;
+import java.util.stream.IntStream;
+
+import lombok.extern.log4j.Log4j2;
+
+import org.apache.commons.lang3.StringUtils;
+import org.opensearch.ingest.AbstractProcessor;
+import org.opensearch.ingest.IngestDocument;
+import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableMap;
 
 /**
  * This processor is used for user input data text embedding processing, model_id can be used to indicate which model user use,
@@ -49,28 +52,32 @@ public class TextEmbeddingProcessor extends AbstractProcessor {
     ) {
         super(tag, description);
         if (StringUtils.isBlank(modelId)) throw new IllegalArgumentException("model_id is null or empty, can not process it");
-        if (fieldMap == null || fieldMap.size() == 0 || isEmbeddingConfigValid(fieldMap)) throw new IllegalArgumentException(
-            "Unable to create the TextEmbedding processor as field_map is null or empty."
-        );
+        validateEmbeddingConfiguration(fieldMap);
 
         this.modelId = modelId;
         this.fieldMap = fieldMap;
         this.mlCommonsClientAccessor = clientAccessor;
     }
 
-    private boolean isEmbeddingConfigValid(Map<String, Object> fieldMap) {
-        return fieldMap.entrySet()
-            .stream()
-            .anyMatch(x -> StringUtils.isBlank(x.getKey()) || Objects.isNull(x.getValue()) || StringUtils.isBlank(x.getValue().toString()));
+    private void validateEmbeddingConfiguration(Map<String, Object> fieldMap) {
+        if (fieldMap == null
+            || fieldMap.size() == 0
+            || fieldMap.entrySet()
+                .stream()
+                .anyMatch(
+                    x -> StringUtils.isBlank(x.getKey()) || Objects.isNull(x.getValue()) || StringUtils.isBlank(x.getValue().toString())
+                )) {
+            throw new IllegalArgumentException("Unable to create the TextEmbedding processor as field_map has invalid key or value");
+        }
     }
 
     @Override
     public IngestDocument execute(IngestDocument ingestDocument) {
-        validateEmbeddingFieldsType(ingestDocument, fieldMap);
-        Map<String, Object> knnMap = buildKnnMap(ingestDocument, fieldMap);
+        validateEmbeddingFieldsValue(ingestDocument, fieldMap);
+        Map<String, Object> knnMap = buildMapWithKnnKeyAndOriginalValue(ingestDocument, fieldMap);
         try {
             List<List<Float>> vectors = mlCommonsClientAccessor.inferenceSentences(this.modelId, createInferenceList(knnMap));
-            processResponse(ingestDocument, knnMap, vectors);
+            appendVectorFieldsToDocument(ingestDocument, knnMap, vectors);
         } catch (ExecutionException | InterruptedException e) {
             log.error("Text embedding processor failed with exception: ", e);
             throw new RuntimeException("Text embedding processor failed with exception", e);
@@ -79,22 +86,22 @@ public class TextEmbeddingProcessor extends AbstractProcessor {
         return ingestDocument;
     }
 
-    void processResponse(IngestDocument ingestDocument, Map<String, Object> knnMap, List<List<Float>> vectors) {
+    void appendVectorFieldsToDocument(IngestDocument ingestDocument, Map<String, Object> knnMap, List<List<Float>> vectors) {
         Objects.requireNonNull(vectors, "embedding failed, inference returns null result!");
         log.debug("Text embedding result fetched, starting build vector output!");
-        Map<String, Object> vectorOutput = buildVectorOutput(knnMap, vectors, ingestDocument.getSourceAndMetadata());
-        vectorOutput.forEach(ingestDocument::appendFieldValue);
+        Map<String, Object> textEmbeddingResult = buildTextEmbeddingResult(knnMap, vectors, ingestDocument.getSourceAndMetadata());
+        textEmbeddingResult.forEach(ingestDocument::appendFieldValue);
     }
 
     @SuppressWarnings({ "unchecked" })
-    private List<String> createInferenceList(Map<String, Object> knnMap) {
+    private List<String> createInferenceList(Map<String, Object> knnKeyMap) {
         List<String> texts = new ArrayList<>();
-        knnMap.entrySet().stream().filter(knnMapEntry -> knnMapEntry.getValue() != null).forEach(knnMapEntry -> {
+        knnKeyMap.entrySet().stream().filter(knnMapEntry -> knnMapEntry.getValue() != null).forEach(knnMapEntry -> {
             Object sourceValue = knnMapEntry.getValue();
             if (sourceValue instanceof List) {
-                ((List<String>) sourceValue).stream().filter(StringUtils::isNotBlank).forEach(texts::add);
+                texts.addAll(((List<String>) sourceValue));
             } else if (sourceValue instanceof Map) {
-                processMapTypeInput(sourceValue, texts);
+                createInferenceListForMapTypeInput(sourceValue, texts);
             } else {
                 texts.add(sourceValue.toString());
             }
@@ -103,9 +110,9 @@ public class TextEmbeddingProcessor extends AbstractProcessor {
     }
 
     @SuppressWarnings("unchecked")
-    private void processMapTypeInput(Object sourceValue, List<String> texts) {
+    private void createInferenceListForMapTypeInput(Object sourceValue, List<String> texts) {
         if (sourceValue instanceof Map) {
-            ((Map<String, Object>) sourceValue).forEach((k, v) -> processMapTypeInput(v, texts));
+            ((Map<String, Object>) sourceValue).forEach((k, v) -> createInferenceListForMapTypeInput(v, texts));
         } else if (sourceValue instanceof List) {
             ((List<String>) sourceValue).stream().filter(StringUtils::isNotBlank).forEach(texts::add);
         } else {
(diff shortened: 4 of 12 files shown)


src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,278 @@
...
+        return res -> {
+            Objects.requireNonNull(res, "embedding failed!");
+            Map<String, Object> vectorOutput = buildVectorOutput(knnMap, res, ingestDocument.getSourceAndMetadata());
+            vectorOutput.forEach(ingestDocument::appendFieldValue);
+        };
+    }
+
+    @VisibleForTesting
+    Consumer<Exception> exceptionConsumer() {
+        return exception -> log.error(exception.getMessage(), exception);
```
navneet1v, line 84: we should enhance the log message

navneet1v:
> Signed-off-by: Zan Niu [zaniu@amazon.com](mailto:zaniu@amazon.com)
> 
> ### Description
> [Describe what this change achieves]
> 
> ### Issues Resolved
> [List any issues this PR will resolve]
> 
> ### Check List
> * [x]  New functionality includes testing.
>   
>   * [x]  All tests pass
> * [ ]  New functionality has been documented.
>   
>   * [ ]  New functionality has javadoc added
> * [x]  Commits are signed as per the DCO using --signoff
> 
> By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license. For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Update the description and issue link.

build.gradle
```diff
@@ -136,6 +136,17 @@ dependencies {
...
+
+    implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.10'
+    testImplementation group: 'commons-io', name: 'commons-io', version: '2.11.0'
+    testImplementation group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.1.2'
+    testImplementation group: 'com.fasterxml.jackson.core', name: 'jackson-core', version: '2.13.3'
+    testImplementation group: 'com.fasterxml.jackson.core', name: 'jackson-annotations', version: '2.1.2'
+    /**Power mock**/
+    testImplementation "org.powermock:powermock-core:1.7.3"
+    testImplementation "org.powermock:powermock-module-junit4:1.7.3"
+    testImplementation "org.powermock:powermock-api-mockito2:1.7.3"
```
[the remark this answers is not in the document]
zane-neo: Done, I was using Mockito only, so removing these dependencies is fine.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,278 @@
...
+
+    public TextEmbeddingProcessor(String tag, String description, String modelId, Map<String, Object> fieldMap, Client client) {
+        super(tag, description);
+        this.modelId = Objects.requireNonNull(modelId, "model_id is null, can not process it");
+        if (fieldMap == null || fieldMap.size() == 0 || checkEmbeddingConfigNotValid(fieldMap)) {
+            throw new IllegalArgumentException("filed_map is null, can not process it");
+        } else {
+            this.fieldMap = fieldMap;
+        }
+        this.mlCommonsClientAccessor = new MLCommonsClientAccessor(new MachineLearningNodeClient(client));
```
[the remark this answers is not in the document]
zane-neo: New Class is not a best practice indeed, but `Processors` are created by `Factory` instead of injection, also the `NeuralSearch` plugin needs an implementation of `IngestionPlugin`, check below code in `NeuralSearch`:
```
@Override
    public Map<String, Processor.Factory> getProcessors(Processor.Parameters parameters) {
        return Collections.singletonMap(TextEmbeddingProcessor.TYPE, new TextEmbeddingProcessor.Factory(parameters.client));
    }
```
We need to return a factory map, and the instance creation happens out of our code by invoking the factory.create automatically, and a `@inject` field won't be initialized correctly.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,278 @@
...
+        validateEmbeddingFieldsType(ingestDocument, fieldMap);
+        Map<String, Object> knnMap = buildKnnMap(ingestDocument, fieldMap);
+
+        ActionListener<List<List<Float>>> internalListener = ActionListener.wrap(
+            responseConsumer(ingestDocument, knnMap),
+            exceptionConsumer()
+        );
+
+        mlCommonsClientAccessor.inferenceSentences(this.modelId, buildMLInput(knnMap), internalListener);
+        return ingestDocument;
```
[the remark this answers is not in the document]
zane-neo: Yes, please check the `responseConsumer`, it's a `CheckedConsumer` and being passed to the ActionListener, the onResponse invocation will invoke the function created in `responseConsumer`.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,278 @@
...
+            Objects.requireNonNull(res, "embedding failed!");
+            Map<String, Object> vectorOutput = buildVectorOutput(knnMap, res, ingestDocument.getSourceAndMetadata());
+            vectorOutput.forEach(ingestDocument::appendFieldValue);
+        };
+    }
+
+    @VisibleForTesting
+    Consumer<Exception> exceptionConsumer() {
+        return exception -> log.error(exception.getMessage(), exception);
+    }
```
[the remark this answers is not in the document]
zane-neo: User has the flexibility to determine what to do when error arise, please refer to this: https://www.elastic.co/guide/en/elasticsearch/reference/master/ingest.html#handling-pipeline-failures. And a processor itself doesn't have the capability to prevent document ingestion.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,278 @@
...
+
+    @SuppressWarnings({ "unchecked" })
+    private List<String> buildMLInput(Map<String, Object> knnMap) {
```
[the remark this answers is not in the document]
zane-neo: Done

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,278 @@
...
+        );
+
+        mlCommonsClientAccessor.inferenceSentences(this.modelId, buildMLInput(knnMap), internalListener);
+        return ingestDocument;
+    }
+
+    @VisibleForTesting
+    CheckedConsumer<List<List<Float>>, Exception> responseConsumer(IngestDocument ingestDocument, Map<String, Object> knnMap) {
+        return res -> {
+            Objects.requireNonNull(res, "embedding failed!");
```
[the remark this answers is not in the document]
zane-neo: This is to check inference return value, null without exception is a rare case. Added a little more info to make the error more clear.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,278 @@
...
+            throw new IllegalArgumentException("nested type field [" + key + "] has non-string type, can not process it");
+        }
+    }
+
+    @Override
+    public String getType() {
+        return TYPE;
+    }
+
+    public static final class Factory implements Processor.Factory {
```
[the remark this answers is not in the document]
zane-neo: It seems this is a standard style of processor factory, this is an example: https://github.com/opensearch-project/OpenSearch/blob/d15795a7aca488c1fadb04b3c8d9f1a3b02e4056/modules/ingest-common/src/main/java/org/opensearch/ingest/common/SetProcessor.java#L115

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,278 @@
...
+            String description,
+            Map<String, Object> config
+        ) throws Exception {
+            String modelId = readStringProperty(TYPE, processorTag, config, MODEL_ID_FIELD);
+            Map<String, Object> filedMap = readOptionalMap(TYPE, processorTag, config, FIELD_MAP_FIELD);
+            return new TextEmbeddingProcessor(processorTag, description, modelId, filedMap, client);
+        }
+    }
+
+    static class IndexWrapper {
```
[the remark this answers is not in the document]
zane-neo: Done

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: Done

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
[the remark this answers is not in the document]
model-collapse: The use can set the tags of ignore_failure to proceed. But at least we need to announce the failure if they choose not to ignore_failure.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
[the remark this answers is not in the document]
model-collapse: This is the opensearch style, because the factory code is little and the logic of processor creation is extremely related to the ingestion logic.

navneet1v:
> > Signed-off-by: Zan Niu [zaniu@amazon.com](mailto:zaniu@amazon.com)
> > ### Description
> > [Describe what this change achieves]
> > ### Issues Resolved
> > [List any issues this PR will resolve]
> > ### Check List
> > 
> > * [x]   New functionality includes testing.
> >   
> >   * [x]   All tests pass
> > * [ ]   New functionality has been documented.
> >   
> >   * [ ]   New functionality has javadoc added
> > * [x]   Commits are signed as per the DCO using --signoff
> > 
> > By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license. For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).
> 
> Update the description and issue link.

@zane-neo please fix this

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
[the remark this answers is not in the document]
navneet1v: Even if we want to announce the failures, we should put a proper message instead of saying exception.getMessage().

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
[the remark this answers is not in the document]
navneet1v: even with the 

```
@Override
    public Map<String, Processor.Factory> getProcessors(Processor.Parameters parameters) {
        return Collections.singletonMap(TextEmbeddingProcessor.TYPE, new TextEmbeddingProcessor.Factory(parameters.client));
    }
```

You can use the same instance that we have created using createComponents. I would say rather than passing client in the `new TextEmbeddingProcessor.Factory(parameters.client)` pass the MLCommonsAccessor.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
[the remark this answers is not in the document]
navneet1v: Just a quick comment its not an OpenSearch style. Given the size of Processor class which you have provided as example it might be fine to put in the same file, but the current TextEmbeddingProcessor class is big enough which already makes it difficult to read. So, as best practice let's move Factory class out of this file. 

You can put the class under `src/main/java/org/opensearch/neuralsearch/processor/factory`.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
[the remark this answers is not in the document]
navneet1v: the question was around, does this function execution will be stopped until we get the inference response?

build.gradle
```diff
@@ -136,6 +136,12 @@ dependencies {
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-knn', version: "${opensearch_build}"
     api fileTree(dir: knnJarDirectory, include: '*.jar')
     api group: 'org.opensearch', name:'opensearch-ml-client', version: "${opensearch_build}"
+
+    implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.10'
+    testImplementation group: 'commons-io', name: 'commons-io', version: '2.11.0'
+    testImplementation group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.1.2'
+    testImplementation group: 'com.fasterxml.jackson.core', name: 'jackson-core', version: '2.13.3'
+    testImplementation group: 'com.fasterxml.jackson.core', name: 'jackson-annotations', version: '2.1.2'
```
navneet1v, line 144: why do we exactly need these dependencies? As per my understanding all the Jackson related features are abstracted under XContentBuilder in openSearch. You might want to check the usage of these.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,286 @@
...
+import org.opensearch.ingest.IngestDocument;
+import org.opensearch.ingest.Processor;
+import org.opensearch.ml.client.MachineLearningNodeClient;
+import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableMap;
+
+@Log4j2
+public class TextEmbeddingProcessor extends AbstractProcessor {
```
navneet1v, line 30: add Java Doc on all the public functions and public classes in main/java.

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
```diff
@@ -71,4 +75,9 @@ public List<QuerySpec<?>> getQueries() {
             new QuerySpec<>(NeuralQueryBuilder.NAME, NeuralQueryBuilder::new, NeuralQueryBuilder::fromXContent)
         );
     }
+
+    @Override
+    public Map<String, Processor.Factory> getProcessors(Processor.Parameters parameters) {
+        return Collections.singletonMap(TextEmbeddingProcessor.TYPE, new TextEmbeddingProcessor.Factory(parameters.client));
```
navneet1v, line 81: rather than passing client here, pass the MLCommonsClientAccessor which we are creating via createComponents. You can achieve this by creating a class variable for MLCommonsClientAccessor and then passing the same variable while creating the TextEmbeddingProcessor

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,286 @@
...
+
+    private final String modelId;
+
+    private final Map<String, Object> fieldMap;
+
+    private final MLCommonsClientAccessor mlCommonsClientAccessor;
+
+    public TextEmbeddingProcessor(String tag, String description, String modelId, Map<String, Object> fieldMap, Client client) {
+        super(tag, description);
+        this.modelId = Objects.requireNonNull(modelId, "model_id is null, can not process it");
```
navneet1v, line 46: we should validate modelId for null and empty string.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,286 @@
...
+        if (fieldMap == null || fieldMap.size() == 0 || checkEmbeddingConfigNotValid(fieldMap)) {
+            throw new IllegalArgumentException("filed_map is null, can not process it");
```
navneet1v, line 48: Should we say something like this in error message:

`Unable to create the TextEmbedding processor as field_map is null or empty.`

I would recommend putting more thoughts on each and every error message that is getting generated as those messages are directly read by users.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,286 @@
...
+        } else {
+            this.fieldMap = fieldMap;
+        }
+        this.mlCommonsClientAccessor = new MLCommonsClientAccessor(new MachineLearningNodeClient(client));
+    }
+
+    private boolean checkEmbeddingConfigNotValid(Map<String, Object> fieldMap) {
+        return fieldMap.entrySet().stream().anyMatch(x -> StringUtils.isBlank(x.getKey()) || Objects.isNull(x.getValue()));
```
navneet1v, line 56: Missing checks for empty Strings, maps in the x.getValue, please add those checks

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,286 @@
...
+                numbers.add(ImmutableMap.of(LIST_TYPE_NESTED_MAP_KEY, modelTensorList.get(indexWrapper.index++)));
+            }
+        }
+        return numbers;
+    }
+
+    private static void validateEmbeddingFieldsType(IngestDocument ingestDocument, Map<String, Object> embeddingFields) {
+        Map<String, Object> sourceAndMetadataMap = ingestDocument.getSourceAndMetadata();
+        for (Map.Entry<String, Object> entry : embeddingFields.entrySet()) {
+            Object obj = sourceAndMetadataMap.get(entry.getKey());
```
navneet1v, line 220: use better names for local variables, fix this throughout the PR. Having good variable names improves readability.

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
```diff
@@ -0,0 +1,161 @@
...
+
+    private static final String indexName = "text_embedding_index";
+
+    private static final String pipelineName = "pipeline-hybrid";
+
+    private final ClassLoader classLoader = this.getClass().getClassLoader();
+
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    public void test_text_embedding_processor() throws IOException {
```
navneet1v, line 34: [Nit-Pick] : Add @SneakyThrows on the method signature which will clean up the function signature by removing the "throws IOException".

Do it for all the functions in the test.

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
```diff
@@ -0,0 +1,161 @@
...
+        while (!isComplete) {
+            taskQueryResult = getTaskQueryResponse(taskId);
+            isComplete = checkComplete(taskQueryResult);
+        }
+        String modelId = taskQueryResult.get("model_id").asText();
+        assertNotNull(modelId);
+        return modelId;
+    }
+
+    private void loadModel(String modelId) throws IOException {
```
navneet1v, line 68: This is a great common function which will be used by Query too, move this function to an abstract class named: BaseNeuralSearchIT.class and then extend the new class to use these functions.

The IT for Query can then use these common functions.

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
```diff
@@ -0,0 +1,161 @@
...
+
+    public void test_text_embedding_processor() throws IOException {
+        String modelId = uploadModel();
+        loadModel(modelId);
+        createPipelineProcessor(modelId);
+        createIndex();
+        ingestDocument();
+    }
+
+    private String uploadModel() throws IOException {
```
navneet1v, line 42: This is a great common function which will be used by Query too, move this function to an abstract class named: BaseNeuralSearchIT.class and then extend the new class to use these functions.

The IT for Query can then use these common functions.

src/test/java/org/opensearch/neuralsearch/utils/TestHelper.java
```diff
@@ -0,0 +1,293 @@
...
+                    Collections.emptyMap(),
+                    Collections.singleton(DATA_ROLE),
+                    Version.CURRENT
+                )
+            );
+        }
+        return state(new ClusterName("test"), indexName, mapping, clusterManagerNode, clusterManagerNode, allNodes);
+    }
+
+    public static ClusterState setupTestClusterState() {
```
navneet1v, line 255: move this to a base class, as suggested for upload and loadModel functions.

src/test/java/org/opensearch/neuralsearch/utils/TestHelper.java
```diff
@@ -0,0 +1,293 @@
...
+    public static RestRequest getStatsRestRequest() {
+        RestRequest request = new FakeRestRequest.Builder(getXContentRegistry()).build();
+        return request;
+    }
+
+    public static RestRequest getStatsRestRequest(String nodeId, String stat) {
+        RestRequest request = new FakeRestRequest.Builder(getXContentRegistry()).withParams(ImmutableMap.of("nodeId", nodeId, "stat", stat))
+            .build();
+        return request;
+    }
```
navneet1v, line 175: not able to see the usage of these 2 functions

src/test/java/org/opensearch/neuralsearch/utils/TestHelper.java
```diff
@@ -0,0 +1,293 @@
...
+        String method,
+        String endpoint,
+        Map<String, String> params,
+        HttpEntity entity,
+        List<Header> headers
+    ) throws IOException {
+        return makeRequest(client, method, endpoint, params, entity, headers, false);
+    }
+
+    public static Response makeRequest(
```
navneet1v, line 116: there are already created functions in OpenSearchRestTestCase class, try to look into them. If you are still not able to find it, move these functions to base class as suggested for upload and loadModel functions, so that all IT can take benefit of.

src/test/java/org/opensearch/neuralsearch/utils/TestHelper.java
```diff
@@ -0,0 +1,293 @@
...
+        return new StringEntity(toJsonString(object), APPLICATION_JSON);
+    }
+
+    public static HttpEntity toHttpEntity(String jsonString) throws IOException {
+        return new StringEntity(jsonString, APPLICATION_JSON);
+    }
+
+    public static RestStatus restStatus(Response response) {
+        return RestStatus.fromCode(response.getStatusLine().getStatusCode());
+    }
```
navneet1v, line 153: not able to see usage of this function.

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
```diff
@@ -0,0 +1,161 @@
...
+            client(),
+            "POST",
+            indexName + "/_doc",
+            null,
+            TestHelper.toHttpEntity(
+                FileUtils.readFileToString(new File(classLoader.getResource("processor/IngestDocument.json").getFile()), "utf-8")
+            ),
+            ImmutableList.of(new BasicHeader(HttpHeaders.USER_AGENT, "Kibana"))
+        );
+        JsonNode node = objectMapper.readTree(EntityUtils.toString(response.getEntity()));
```
navneet1v, line 140: check other Integration tests on how to read the responses.

build.gradle
zane-neo: Checked on XContentBuilder but failed to find the function I would like to use, what I want is to parse string to map or parse a specific field from a json string, please let me know how I can achieve these with this builder. Thanks.

src/test/java/org/opensearch/neuralsearch/utils/TestHelper.java
zane-neo: Didn't found the function similar to this makeRequest, moved upload and loadModel and this function to BaseNeuralSearchIT.class

build.gradle
jmazanec15: For xContentBuilder to map I usually use some combination of: 

1. https://github.com/opensearch-project/neural-search/blob/main/src/test/java/org/opensearch/neuralsearch/plugin/TestUtils.java#L22
2. https://github.com/opensearch-project/k-NN/blob/main/src/testFixtures/java/org/opensearch/knn/KNNRestTestCase.java#L154-L164

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,269 @@
...
+        if (fieldMap == null || fieldMap.size() == 0 || checkEmbeddingConfigNotValid(fieldMap)) throw new IllegalArgumentException(
+            "Unable to create the TextEmbedding processor as field_map is null or empty."
+        );
+
+        this.modelId = modelId;
+        this.fieldMap = fieldMap;
+        this.mlCommonsClientAccessor = clientAccessor;
+    }
+
+    private boolean checkEmbeddingConfigNotValid(Map<String, Object> fieldMap) {
```
jmazanec15, line 58: nit:  isEmbeddingConfigValid (negate output) -> related post https://softwareengineering.stackexchange.com/questions/196830/boolean-method-naming-affirmative-vs-negative

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,269 @@
...
+            log.error("Text embedding processor failed with exception: " + e.getMessage(), e);
+            throw new RuntimeException("Text embedding processor failed with exception", e);
+        }
+        log.info("Text embedding completed, returning ingestDocument!");
+        return ingestDocument;
+    }
+
+    void processResponse(IngestDocument ingestDocument, Map<String, Object> knnMap, List<List<Float>> vectors) {
+        Objects.requireNonNull(vectors, "embedding failed, inference returns null result!");
+        log.info("Text embedding result fetched, starting build vector output!");
```
jmazanec15, line 81: qq: If this method is executed on each doc ingest, I dont think we should log.info -> itd be better to switch to debug. What happens if 1M documents are ingested?

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,269 @@
...
+            }
+        });
+        return texts;
+    }
+
+    @SuppressWarnings("unchecked")
+    private void processMapTypeInput(Object sourceValue, List<String> texts) {
+        if (sourceValue instanceof Map) {
+            ((Map<String, Object>) sourceValue).forEach((k, v) -> processMapTypeInput(v, texts));
+        } else if (sourceValue instanceof List) {
```
jmazanec15, line 106: When would this be a list? https://github.com/opensearch-project/neural-search/pull/18/files#diff-703b4a7aad2b342a1385adb42d1b7c2b4fa3bd778a56b35f313a4a52e77df436R94 shows its already a map?

src/test/java/org/opensearch/neuralsearch/common/BaseNeuralSearchIT.java
```diff
@@ -0,0 +1,139 @@
...
+            toHttpEntity(""),
+            ImmutableList.of(new BasicHeader(HttpHeaders.USER_AGENT, "Kibana"))
+        );
+        JsonNode uploadResJson = objectMapper.readTree(EntityUtils.toString(uploadResponse.getEntity()));
+        String taskId = uploadResJson.get("task_id").asText();
+        assertNotNull(taskId);
+
+        JsonNode taskQueryResult = getTaskQueryResponse(taskId);
+        boolean isComplete = checkComplete(taskQueryResult);
+        while (!isComplete) {
```
jmazanec15, line 74: Will this hang forever if load fails?

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private static final String pipelineName = "pipeline-hybrid";
+
+    private final ClassLoader classLoader = this.getClass().getClassLoader();
+
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    private static final Locale locale = Locale.getDefault();
+
+    public void test_text_embedding_processor() throws Exception {
```
jmazanec15, line 40: Camelcase for each component in test name (suggestion from [Google style guide](https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names)) -> testTextEmbeddingProcessor

If there are multiple tests, then we can separate components with _. For example: testTextEmbeddingProcessor_whenInputInvalid_thenThrowException

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
```diff
@@ -0,0 +1,108 @@
...
+        createTextEmbeddingIndex();
+        ingestDocument();
+    }
+
+    private String uploadTextEmbeddingModel() throws Exception {
+        String requestBody = Files.readString(Path.of(classLoader.getResource("processor/UploadModelRequestBody.json").toURI()));
+        return uploadModel(requestBody);
+    }
+
+    private void createPipelineProcessor(String modelId) throws Exception {
```
jmazanec15, line 53: Should this be moved to BaseClass so other ITs can use it?

src/test/java/org/opensearch/neuralsearch/utils/TestHelper.java
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.neuralsearch.utils;
+
+public class TestHelper {
+
```
jmazanec15, line 9: Whats this for?

src/test/resources/processor/UploadModelRequestBody.json
```diff
@@ -0,0 +1,13 @@
...
+  "version": 1,
+  "model_format": "TORCH_SCRIPT",
+  "model_task_type": "text_embedding",
+  "model_config": {
+    "model_type": "bert",
+    "embedding_dimension": 384,
+    "framework_type": "sentence_transformers",
+    "all_config": "{\"architectures\":[\"BertModel\"],\"max_position_embeddings\":512,\"model_type\":\"bert\",\"num_attention_heads\":12,\"num_hidden_layers\":6}"
+  },
+  "url": "https://api.quip-amazon.com/2/blob/MdZ9AAsfqat/y-6nBQpg6Ma_UEE3pYt2NQ?name=all-MiniLM-L6-v2.zip&oauth_token=TUhMOU1BV1gwWUE%3D%7C1695445530%7CNN8X0Y0SQ0NfJvMxNZCnumpJaurxCDaT%2FdK70Al%2Bgh0%3D&s=YkW8AVqTosiF"
```
jmazanec15, line 12: Is this link available outside of amazon? If no, we cannot embed it here.

jmazanec15, line 12: Is this link available outside of amazon? If no, just check in the model in order to run with ITs.

ylwu-amzn:
How about adding more details in description? Suggest adding some examples there so people know what feature this PR is building.

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
```diff
@@ -0,0 +1,84 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.ImmutableList;
+
+public class TextEmbeddingProcessorIT extends BaseNeuralSearchIT {
+
+    private static final String indexName = "text_embedding_index";
+
+    private static final ObjectMapper objectMapper = new ObjectMapper();
```
navneet1v, line 31: make static final variable UPPER_CASE, that is a general convention for all static final variables

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private static final Locale locale = Locale.getDefault();
```
navneet1v, line 33: This variable is not used.

src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
```diff
@@ -95,6 +103,25 @@ public void testInferenceSentences_whenExceptionFromMLClient_thenFailure() {
         Mockito.verifyNoMoreInteractions(resultListener);
     }
 
+    public void test_blockingInferenceSentences() throws ExecutionException, InterruptedException {
```
navneet1v, line 106: use @SneakyThrows.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -72,23 +75,54 @@ public void inferenceSentences(
...
+     * @param modelId {@link String}
+     * @param inputText {@link List<String>} on which inference needs to happen.
+     * @return {@link List} of {@link List<String>} represents the text embedding vector result.
+     * @throws ExecutionException If the underlying task failed, this exception will be thrown in the future.get().
+     * @throws InterruptedException If the thread is interrupted, this will be thrown.
+     */
+    public List<List<Float>> inferenceSentences(@NonNull final String modelId, @NonNull final List<String> inputText)
+        throws ExecutionException, InterruptedException {
+        final MLInput mlInput = createMLInput(TARGET_RESPONSE_FILTERS, inputText);
+        ActionFuture<MLOutput> outputActionFuture = mlClient.predict(modelId, mlInput);
```
navneet1v, line 103: [nit-pick] make this variable final, like all other local variables.

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
```diff
@@ -71,4 +76,11 @@ public List<QuerySpec<?>> getQueries() {
             new QuerySpec<>(NeuralQueryBuilder.NAME, NeuralQueryBuilder::new, NeuralQueryBuilder::fromXContent)
         );
     }
+
+    @Override
+    public Map<String, Processor.Factory> getProcessors(Processor.Parameters parameters) {
+        final MachineLearningNodeClient machineLearningNodeClient = new MachineLearningNodeClient(parameters.client);
+        final MLCommonsClientAccessor clientAccessor = new MLCommonsClientAccessor(machineLearningNodeClient);
```
navneet1v, line 83: This is what was not suggested in the older comment. The idea was to use the instance of  MLCommonsClientAccessor which was created in createComponents Function

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,269 @@
...
+
+    public TextEmbeddingProcessor(
+        String tag,
+        String description,
+        String modelId,
+        Map<String, Object> fieldMap,
+        MLCommonsClientAccessor clientAccessor
+    ) {
+        super(tag, description);
+        if (StringUtils.isBlank(modelId)) throw new IllegalArgumentException("model_id is null, can not process it");
```
navneet1v, line 48: it should be modelId is null or Empty, not only null.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,269 @@
...
+    @SuppressWarnings({ "unchecked" })
+    private void processConfiguredMapType(
+        String parentKey,
+        Object knnKey,
+        Map<String, Object> sourceAndMetadataMap,
+        Map<String, Object> treeRes
+    ) {
+        if (knnKey == null || sourceAndMetadataMap == null) return;
+        if (knnKey instanceof Map) {
+            Map<String, Object> next = new LinkedHashMap<>();
```
navneet1v, line 141: why we are using LinkedHashMap instead of HashMap only?

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,269 @@
...
+        List<Map<String, List<Float>>> numbers = new ArrayList<>();
+        for (String strSourceValue : sourceValue) {
+            if (StringUtils.isNotBlank(strSourceValue)) {
+                numbers.add(ImmutableMap.of(LIST_TYPE_NESTED_MAP_KEY, modelTensorList.get(indexWrapper.index++)));
+            }
+        }
+        return numbers;
+    }
+
+    private static void validateEmbeddingFieldsType(IngestDocument ingestDocument, Map<String, Object> embeddingFields) {
```
navneet1v, line 221: any reason for making this and other functions static? if no please make them no static.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,269 @@
...
+        vectorOutput.forEach(ingestDocument::appendFieldValue);
+    }
+
+    @SuppressWarnings({ "unchecked" })
+    private List<String> createInferenceList(Map<String, Object> knnMap) {
+        List<String> texts = new LinkedList<>();
+        knnMap.entrySet().stream().filter(knnMapEntry -> knnMapEntry.getValue() != null).forEach(knnMapEntry -> {
+            Object sourceValue = knnMapEntry.getValue();
+            if (sourceValue instanceof List) {
+                ((List<String>) sourceValue).stream().filter(StringUtils::isNotBlank).forEach(texts::add);
```
navneet1v, line 92: can we add a comment here, which tells what when we will be building the output vector list for this list, we need to add empty vector list to define the string was empty hence no text embeddings.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,269 @@
...
+    private List<Map<String, List<Float>>> processListOut(
+        List<String> sourceValue,
+        List<List<Float>> modelTensorList,
+        IndexWrapper indexWrapper
+    ) {
+        List<Map<String, List<Float>>> numbers = new ArrayList<>();
+        for (String strSourceValue : sourceValue) {
+            if (StringUtils.isNotBlank(strSourceValue)) {
+                numbers.add(ImmutableMap.of(LIST_TYPE_NESTED_MAP_KEY, modelTensorList.get(indexWrapper.index++)));
+            }
```
navneet1v, line 216: if we remove blank strings here, then the order vector lists will be wrong. Example:
```
Input
["This is"," ","Hello world"]

Output vectors in document:
[{"vector":[1.0,2.0]},{"vector":[4.0,3.0]}]
```
This seems to be wrong. I also think we add an empty vector list here that will cause the exception. Can we go with vector list of 0.0 values ?

@jmazanec15 do we see any issue with 0.0 value? Will is cause any impact on the K-NN? 

@ylwu-amzn if we pass a blank string or a string with spaces to ML Algorithms what will happen? will it fail? I think ans is no. want to confirm.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -72,23 +75,54 @@ public void inferenceSentences(
...
+     * Abstraction to call predict function of api of MLClient with provided targetResponseFilters. It uses the
+     * custom model provided as modelId and run the {@link MLModelTaskType#TEXT_EMBEDDING}. The return will be sent
+     * using the actionListener which will have a {@link List} of {@link List} of {@link Float} in the order of
+     * inputText. We are not making this function generic enough to take any function or TaskType as currently we need
+     * to run only TextEmbedding tasks only. Please note this method is a blocking method, use this only when the processing
+     * needs block waiting for response, otherwise please use {@link #inferenceSentences(String, List, ActionListener)}
+     * instead.
+     * @param modelId {@link String}
+     * @param inputText {@link List<String>} on which inference needs to happen.
+     * @return {@link List} of {@link List<String>} represents the text embedding vector result.
```
navneet1v, line 96:
```suggestion
     * @param inputText {@link List} of {@link String} on which inference needs to happen.
     * @return {@link List} of {@link List} of {@link String} represents the text embedding vector result.
```

Do this, otherwise it is a java doc error.

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
navneet1v: Screen Shot 2022-10-14 at 7 41 45 PM

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
zane-neo: Done

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
zane-neo: Done

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,269 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.neuralsearch.processor;
+
+import java.util.*;
```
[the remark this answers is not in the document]
zane-neo: Done

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,269 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.opensearch.ingest.AbstractProcessor;
+import org.opensearch.ingest.IngestDocument;
+import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableMap;
+
+/**
+ * Before querying with neural search, all the documents should be ingested in the form of embedded vectors. Leaving the embedding process
```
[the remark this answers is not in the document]
zane-neo: Done

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: Done

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: Changed log to debug level.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: This would be a mixed type, e.g. 
```
{
    "key": ["v1", "v2"]
}
```

src/test/java/org/opensearch/neuralsearch/common/BaseNeuralSearchIT.java
```diff
@@ -0,0 +1,164 @@
...
+            taskQueryResult = getTaskQueryResponse(taskId);
+            isComplete = checkComplete(taskQueryResult);
```
navneet1v, line 82: This is a very brute force way of checking if model is uploaded or not. This is very resource intensive plus there is no time out which makes can make it run forever and just increase the build time.

Suggestion:
1. We should move this checking of model upload to another thread.
2. Add a timeout(3 times than the actual time, or let say 1 min) after which we will fail the test saying that model is not uploaded, and provide proper response why the tests failed and add things like increase the time out and other things.
3. To avoid the resource intensive work, we should also provide some sleep time in the thread which is checking the model upload if we apply \#1 suggestion.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: For map type user input, we're performing a pre-order traversal on this map to extract all the texts to build the list. And when we build the vector response with the vector list, we need to fill the list value into the map with a correct order. A LinkedHashMap can maintain the order of the keys when iteration on it, but HashMap not.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: Static functions means this method do not depends on any stateful data, so other methods can use this without concerns of the state changing. Please let me know your thoughts on making it no static. Thanks.

src/test/java/org/opensearch/neuralsearch/common/BaseNeuralSearchIT.java
zane-neo: No because failure loading task will return a result with error not empty. That said, added sleep to avoid CPU consuming and max time to limit the method running.

src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
zane-neo: Done

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
zane-neo: Done

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
zane-neo: removed

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
```diff
@@ -0,0 +1,84 @@
...
+                    .startObject("index")
+                    .field("knn", true)
+                    .field("knn.algo_param.ef_search", 100)
+                    .field("refresh_interval", "30s")
+                    .field("default_pipeline", pipelineName)
+                    .endObject()
+                    .field("number_of_shards", 1)
+                    .field("number_of_replicas", 0)
+                    .endObject()
+                    .endObject()
```
navneet1v, line 61: I tried running this test it failed at the line 61. I see that we are adding 1 more extra endObject which is not required.

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
```diff
@@ -0,0 +1,84 @@
...
+            );
+            createIndex(
+                indexName,
+                Settings.builder().loadFromSource(settings, XContentType.JSON).build(),
+                Files.readString(Path.of(classLoader.getResource("processor/IndexMappings.json").toURI()))
```
navneet1v, line 66: as we are reading a JSON object from file, while reading the object you might want to remove "{" and "}" from start and end. I tried removing then it worked for. 

Example reference: https://github.com/opensearch-project/k-NN/blob/48d2303ad6964d386709ab5ae5fdbb0965420cb8/src/testFixtures/java/org/opensearch/knn/KNNRestTestCase.java#L638-L639

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
navneet1v: Command to rerun the test:
```
./gradlew ':integTest' --tests "org.opensearch.neuralsearch.processor.TextEmbeddingProcessorIT.test_text_embedding_processor" -Dtests.seed=173A58A0D4C0E3A0 -Dtests.security.manager=false -Dtests.locale=es-SV -Dtests.timezone=Asia/Seoul
```

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
navneet1v: So, this function is not dependent on any stateful correct that is correct, but it is always used in the context of a Stateful and that too only in this class.

So, seems like with your assumption if a function is not dependent on a state it should always be static. That seems to be an overkill. Static function lives in JVM heap and generally created to keep data in heap and make them run faster. We don't have any of the case here.

If this function was present in another class then we might want to make this function static as we don't want to create objects at runtime as object creation is expensive. But that is not the case here.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,269 @@
...
+
+    @Override
+    public IngestDocument execute(IngestDocument ingestDocument) {
+        validateEmbeddingFieldsType(ingestDocument, fieldMap);
+        Map<String, Object> knnMap = buildKnnMap(ingestDocument, fieldMap);
+        try {
+            List<List<Float>> vectors = mlCommonsClientAccessor.inferenceSentences(this.modelId, createInferenceList(knnMap));
+            processResponse(ingestDocument, knnMap, vectors);
+        } catch (ExecutionException | InterruptedException e) {
+            log.error("Text embedding processor failed with exception: " + e.getMessage(), e);
```
navneet1v, line 72:
```suggestion
            log.error("Text embedding processor failed with exception: ", e);
```
When exception trace is printed the exception message will be printed along that. So we can avoid doing e.getMessage()

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: > Static function lives in JVM heap and generally created to keep data in heap and make them run faster. We don't have any of the case here.
If this function was present in another class then we might want to make this function static as we don't want to create objects at runtime as object creation is expensive. But that is not the case here.

The reducing of time complexity is the benefit we can get from a static method instead of the unnecessary of object creation, please refer to: https://stackoverflow.com/a/135038.  This running faster itself is a benefit for us.

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
```diff
@@ -71,4 +76,11 @@ public List<QuerySpec<?>> getQueries() {
             new QuerySpec<>(NeuralQueryBuilder.NAME, NeuralQueryBuilder::new, NeuralQueryBuilder::fromXContent)
         );
     }
+
+    @Override
+    public Map<String, Processor.Factory> getProcessors(Processor.Parameters parameters) {
+        final MachineLearningNodeClient machineLearningNodeClient = new MachineLearningNodeClient(parameters.client);
+        mlCommonsClientAccessor = new MLCommonsClientAccessor(machineLearningNodeClient);
```
zane-neo, line 83: We need to create the MLCommonsClientAccessor in the getProcessors method since this method executes before `createComponents`, check this line: https://github.com/opensearch-project/OpenSearch/blob/e44158d4d10d4f8905895ffa50bf9398b8550667/server/src/main/java/org/opensearch/node/Node.java#L515, and this line: https://github.com/opensearch-project/OpenSearch/blob/e44158d4d10d4f8905895ffa50bf9398b8550667/server/src/main/java/org/opensearch/node/Node.java#L711.

src/test/resources/processor/UploadModelRequestBody.json
zane-neo: @ylwu-amzn , please chime in and offer help on this.

ylwu-amzn: This is not a public link. Suggest put the model inside your test resource folder, then use local file url.

src/main/java/org/opensearch/neuralsearch/processor/factory/TextEmbeddingProcessorFactory.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Override
+    public TextEmbeddingProcessor create(
+        Map<String, Processor.Factory> registry,
+        String processorTag,
+        String description,
+        Map<String, Object> config
+    ) throws Exception {
+        String modelId = readStringProperty(TYPE, processorTag, config, MODEL_ID_FIELD);
+        Map<String, Object> filedMap = readOptionalMap(TYPE, processorTag, config, FIELD_MAP_FIELD);
```
jmazanec15, line 34: Why optional? If this doesnt exist, what is the purpose of the processor?

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,272 @@
...
+        if (fieldMap == null || fieldMap.size() == 0 || isEmbeddingConfigValid(fieldMap)) throw new IllegalArgumentException(
+            "Unable to create the TextEmbedding processor as field_map is null or empty."
+        );
+
+        this.modelId = modelId;
+        this.fieldMap = fieldMap;
+        this.mlCommonsClientAccessor = clientAccessor;
+    }
+
+    private boolean isEmbeddingConfigValid(Map<String, Object> fieldMap) {
```
jmazanec15, line 61: It should return true if it is valid. False otherwise.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,272 @@
...
+            processResponse(ingestDocument, knnMap, vectors);
+        } catch (ExecutionException | InterruptedException e) {
+            log.error("Text embedding processor failed with exception: ", e);
+            throw new RuntimeException("Text embedding processor failed with exception", e);
+        }
+        log.debug("Text embedding completed, returning ingestDocument!");
+        return ingestDocument;
+    }
+
+    void processResponse(IngestDocument ingestDocument, Map<String, Object> knnMap, List<List<Float>> vectors) {
```
jmazanec15, line 82: Suggestion: Can we call this "appendVectorFieldsToDocument"? IMO it is hard to understand what is meant by process.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: Done

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
zane-neo: This method encapsulation is really not friendly to use, I prefer to create our own method.

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
zane-neo: Fixed.

src/test/java/org/opensearch/neuralsearch/common/BaseNeuralSearchIT.java
zane-neo: Add max retry time and sleep time in the thread.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: The final decision is: null or empty string value in a list will cause exception.

src/main/java/org/opensearch/neuralsearch/processor/factory/TextEmbeddingProcessorFactory.java
zane-neo: Changed to non optional method, this will throw a configuration exception when this map is missing.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,272 @@
...
+                processMapTypeInput(sourceValue, texts);
+            } else {
+                texts.add(sourceValue.toString());
+            }
+        });
+        return texts;
+    }
+
+    @SuppressWarnings("unchecked")
+    private void processMapTypeInput(Object sourceValue, List<String> texts) {
```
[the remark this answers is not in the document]
zane-neo: Sure, changed to `createInferenceListForMapTypeInput`

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,272 @@
...
+        } else if (sourceValue instanceof List) {
+            ((List<String>) sourceValue).stream().filter(StringUtils::isNotBlank).forEach(texts::add);
+        } else {
+            if (sourceValue == null || StringUtils.isBlank(sourceValue.toString())) return;
+            texts.add(sourceValue.toString());
+        }
+    }
+
+    @VisibleForTesting
+    Map<String, Object> buildKnnMap(IngestDocument ingestDocument, Map<String, Object> fieldMap) {
```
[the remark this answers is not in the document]
zane-neo: changed to `buildMapWithKnnKeyAndOriginalValue`, this name is self-explaination.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,272 @@
...
+            treeRes.put(parentKey, next);
+        } else {
+            String key = String.valueOf(knnKey);
+            treeRes.put(key, sourceAndMetadataMap.get(parentKey));
+        }
+    }
+
+    @SuppressWarnings({ "unchecked" })
+    @VisibleForTesting
+    Map<String, Object> buildVectorOutput(
```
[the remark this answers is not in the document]
zane-neo: changed to `buildTextEmbeddingResult`, this is more descriptive.

src/test/java/org/opensearch/neuralsearch/utils/TestHelper.java
zane-neo: Removed.

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
zane-neo: Changed tests method names to google style.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -72,23 +75,54 @@ public void inferenceSentences(
...
+     * @throws ExecutionException If the underlying task failed, this exception will be thrown in the future.get().
+     * @throws InterruptedException If the thread is interrupted, this will be thrown.
+     */
+    public List<List<Float>> blockingInferenceSentences(@NonNull final String modelId, @NonNull final List<String> inputText)
```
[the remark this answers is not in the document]
zane-neo: Change the naming.

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
zane-neo: Changed to XContent approach.

src/test/java/org/opensearch/neuralsearch/utils/TestHelper.java
zane-neo: Done

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: Changed to debug level.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: Sure

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,288 @@
...
+        if (StringUtils.isBlank(modelId)) throw new IllegalArgumentException("model_id is null or empty, can not process it");
+        validateEmbeddingConfiguration(fieldMap);
+
+        this.modelId = modelId;
+        this.fieldMap = fieldMap;
+        this.mlCommonsClientAccessor = clientAccessor;
+    }
+
+    private void validateEmbeddingConfiguration(Map<String, Object> fieldMap) {
+        if (fieldMap == null || fieldMap.size() == 0 || fieldMap.entrySet()
```
jmazanec15, line 61: I am wondering if a user can create an arbitrarily large fieldMap? If so, should this be limited, from a security perspective?

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,288 @@
...
+
+    private static void validateEmbeddingFieldsValue(IngestDocument ingestDocument, Map<String, Object> embeddingFields) {
+        Map<String, Object> sourceAndMetadataMap = ingestDocument.getSourceAndMetadata();
+        for (Map.Entry<String, Object> embeddingFieldsEntry : embeddingFields.entrySet()) {
+            Object sourceValue = sourceAndMetadataMap.get(embeddingFieldsEntry.getKey());
+            if (sourceValue != null) {
+                String sourceKey = embeddingFieldsEntry.getKey();
+                Class<?> sourceValueClass = sourceValue.getClass();
+                if (List.class.isAssignableFrom(sourceValueClass) || Map.class.isAssignableFrom(sourceValueClass)) {
+                    validateNestedTypeValue(sourceKey, sourceValue);
```
jmazanec15, line 232: Related to above comment: are there any limits on the depth of nested parameter that could be passed?

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,288 @@
...
+        } else if (sourceValue instanceof List) {
+            ((List<String>) sourceValue).stream().filter(StringUtils::isNotBlank).forEach(texts::add);
+        } else {
+            if (sourceValue == null || StringUtils.isBlank(sourceValue.toString())) return;
+            texts.add(sourceValue.toString());
+        }
+    }
+
+    @VisibleForTesting
+    Map<String, Object> buildMapWithKnnKeyAndOriginalValue(IngestDocument ingestDocument, Map<String, Object> fieldMap) {
```
jmazanec15, line 121: private method: do we need to pass fieldMap if its a member?

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: For now, there's no limits. Nested type will not be explicitly mentioned in the doc, so users use only raw string or list type. If we receive feedback on supporting nested type, we can then tell user how.

jmazanec15 (approved):

model-collapse (approved):
LGTM

PR #20, opened 2022-10-13
Integrate model inference to build query

### LINKED ISSUE #14: Create custom neural query

jmazanec15:
### Description
Integrates ml-commons model inference capabilities to transform NeuralQueryBuilder into a KNNQueryBuilder. Minor changes to parsing logic and build.gradle to fix bugs. Minor enhancement to MLCommonsCLientAccessor to add single sentence inference.

Added unit tests to test functionality. 

Working on integration tests. Confirmed it works on a local cluster by setting up k-NN index with 1K docs and running the following query:
```
$ curl -XPOST "localhost:9200/test-index/_search?_source_excludes=cool_field&pretty" -H 'Content-Type: application/json' -d'
{
  "query": {
    "neural": {
      "cool_field": {
        "query_text": "Hello world!",
        "model_id": "rlAg04MB3cG1ZCLOBuDF",
        "k": 1000
      }
    }
  },
  "size": 5
}
'
{
  "took" : 14,
  "timed_out" : false,
  "_shards" : {
    "total" : 3,
    "successful" : 3,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 1000,
      "relation" : "eq"
    },
    "max_score" : 0.002477972,
    "hits" : [
      {
        "_index" : "test-index",
        "_id" : "613",
        "_score" : 0.002477972,
        "_source" : { }
      },
      {
        "_index" : "test-index",
        "_id" : "864",
        "_score" : 0.0024763448,
        "_source" : { }
      },
      {
        "_index" : "test-index",
        "_id" : "190",
        "_score" : 0.002475292,
        "_source" : { }
      },
      {
        "_index" : "test-index",
        "_id" : "818",
        "_score" : 0.002458808,
        "_source" : { }
      },
      {
        "_index" : "test-index",
        "_id" : "340",
        "_score" : 0.0024474028,
        "_source" : { }
      }
    ]
  }
}
```

### Issues Resolved
#14 

### Check List
    - [X] All tests pass


src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
```diff
@@ -54,10 +66,20 @@ public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder>
 
     private static final int DEFAULT_K = 10;
 
+    private static MLCommonsClientAccessor ML_CLIENT;
+
+    public static void initialize(MLCommonsClientAccessor mlClient) {
+        NeuralQueryBuilder.ML_CLIENT = mlClient;
+    }
```
navneet1v, line 73: we have MLCommonsClientAccessor instance created via createComponents, you might want to reuse it.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -35,6 +35,33 @@ public class MLCommonsClientAccessor {
...
 
+    /**
+     * Wrapper around {@link #inferenceSentences} that expected a single input text and produces a single floating
+     * point vector as a response.
+     *
+     * @param modelId {@link String}
+     * @param inputText {@link List} of {@link String} on which inference needs to happen
+     * @param listener {@link ActionListener} which will be called when prediction is completed or errored out
+     */
+    public void inferenceSentence(
```
navneet1v, line 46: you don't need this, when we already have a function which does for a list of sentences, please try to use that.

src/test/java/org/opensearch/neuralsearch/common/VectorUtilTests.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package org.opensearch.neuralsearch.common;
+
+import java.util.Collections;
+import java.util.List;
+
+import org.opensearch.test.OpenSearchTestCase;
+
+public class VectorUtilTests extends OpenSearchTestCase {
```
navneet1v, line 13: add Java docs on the class and also on the function.

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
```diff
@@ -172,10 +193,32 @@ private static void parseQueryParams(XContentParser parser, NeuralQueryBuilder n
         }
     }
 
+    @Override
+    protected QueryBuilder doRewrite(QueryRewriteContext queryRewriteContext) {
+        if (vectorSupplier() != null && vectorSupplier.get() != null) {
+            return new KNNQueryBuilder(fieldName(), vectorSupplier.get(), k());
+        }
```
navneet1v, line 200: why do we need these checks, if the query builder is running the vectorSupplier will be null isn't it?

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
```diff
@@ -172,10 +193,32 @@ private static void parseQueryParams(XContentParser parser, NeuralQueryBuilder n
...
+
+        if (vectorSupplier() == null) {
+            SetOnce<float[]> vectorSetOnce = new SetOnce<>();
+            queryRewriteContext.registerAsyncAction(
+                ((client, actionListener) -> ML_CLIENT.inferenceSentence(modelId(), queryText(), ActionListener.wrap(floatList -> {
+                    vectorSetOnce.set(vectorAsListToArray(floatList));
+                    actionListener.onResponse(null);
```
navneet1v, line 207: why setting null on the OnReponse?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
jmazanec15: Im going to be writing the code to convert functionality to single sentence/single vector. Why not incorporate it here so it can be used by other features of the code as well?

src/main/java/org/opensearch/neuralsearch/common/VectorUtil.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package org.opensearch.neuralsearch.common;
+
+import java.util.List;
+
+/**
+ * Utility class for working with vectors
+ */
+public class VectorUtil {
```
navneet1v, line 13: nit pick, you can add @NoArgsConstructor(accessLevel=PRIVATE) on this class to make sure that we are not creating instance of this class anywhere

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
jmazanec15: VectorSupplier is null initially and then gets set during rewrite. The actual vector will not get set until the async call finishes completely.

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
jmazanec15: This was how it was done for Geo: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/AbstractGeometryQueryBuilder.java#L519. 

The [rewrite context](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/QueryRewriteContext.java#L113) specifies listener response type as wildcard. That being said, we dont control the listener that is passed in, so all we can do is return null.

src/main/java/org/opensearch/neuralsearch/common/VectorUtil.java
jmazanec15: Good suggestion - will update

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
navneet1v: All you need to do is do a get on the vector list, for that if we are creating a new function that is over abstraction. We already have 2 versions of this function in the class, and add third which is very specific seems to be an overkill for now.

jmazanec15: I prefer to keep this here. I dont see a downside to having it. The name of the method "inferenceSentence" distinguishes it from "inferenceSentences" so there won't be any confusion when to use what. Also, I think single versus collection isnt so specific that it couldn't be useful outside my current use case. Many other Java interfaces/classes have methods for both.

> All you need to do is do a get on the vector list,

True, but you also have to ensure only 1 vector is returned - so each client would also have to add this check/error handling. Using this function, we can centralize that check. Additionally, it is clunky to pass in a List\<List\<Float\>\> listener when you expect that there will only be 1 List\<Float\>

Given that the purpose of this class is to build an easy to use abstraction over MLClient, I think it fits to add this method here.

jmazanec15: I can add parity method with inferenceSentences where filters are passed in as well.

navneet1v: Can we abstract the single sentence function in the queryBuilder class only? I just want to avoid confusion around the usage and 1 more function where we will have use case of not using the TARGET_RESPONSE_FILTERS, for 1 single sentence.

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
navneet1v: I think the reason of making onResponse(null) is because we have already set the response to vector supplier. 

Can we add some more details around how this rewrite query function will work with first if condition like vectorSupplier != null and all.

Because what I am getting from this whole function is the re-writeQuery function can be called atleast 2 times. where one time vectorSupplier will be null and next time it will not be null. is that understanding correct?

build.gradle
```diff
@@ -59,6 +59,7 @@ opensearchplugin {
     classname "${projectPath}.${pathToPlugin}.${pluginClassName}"
     licenseFile rootProject.file('LICENSE')
     noticeFile rootProject.file('NOTICE')
+    extendedPlugins = ['opensearch-knn']
```
navneet1v, line 62: should we add MLCommons also over here? or this is just for the plugins who we depend during compile time?

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
```diff
@@ -172,10 +193,32 @@ private static void parseQueryParams(XContentParser parser, NeuralQueryBuilder n
...
+                }, actionListener::onFailure)))
+            );
+            vectorSupplier(vectorSetOnce::get);
+        }
+
+        // Rewrites will continuously happen until the supplier is set and the vector is generated. Rewrites will stop
+        // once this object is returned. Hence, if we get here, we need to return a new object
+        return new NeuralQueryBuilder(fieldName(), queryText(), modelId(), k(), vectorSupplier());
```
navneet1v, line 215: does this ensure that main thread which called the rewriteQuery comes back and check if the vector response has come or not?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
jmazanec15: I dont think there is any usage confusion, given the method name and signature are descriptive of the functionality.

In terms of maintainability, I dont think it makes sense for other components that will want to use the inferenceSentence method to depend on the queryBuilder class. I think that the ml package should be responsible for providing easy/intuitive interaction with ml-commons for the components in the plugin and should therefore be able to house this functionality. 

That being said, I think it makes sense to either integrate it into this class or build a new class that uses this class to provide higher level abstractions. In a similar case, OpenSearch has a "HighLevel" [rest client](https://github.com/opensearch-project/OpenSearch/blob/main/client/rest-high-level/src/main/java/org/opensearch/client/RestHighLevelClient.java) that is built on top of the lower level RestClient to provide more functionality. We could do this, or we could treat the MLCommonsClientAccessor as the higher level client.

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
jmazanec15: > I think the reason of making onResponse(null) is because we have already set the response to vector supplier.

We are not setting the response, but instead setting the value of the supplier.

> Can we add some more details around how this rewrite query function will work with first if condition like vectorSupplier != null and all.
> Because what I am getting from this whole function is the re-writeQuery function can be called atleast 2 times. where one time vectorSupplier will be null and next time it will not be null. is that understanding correct?

Yes, will add additional context in the comment. From my understanding, the query will be rewritten until the rewrite method returns the object itself. So, on first rewrite, the supplier will be null. On second rewrite, the supplier will be set, but the vector may be null, depending if the async call finished in time or not. Let me double check this though and add a comment.

build.gradle
jmazanec15: For ml-commons, we are okay because we just take the client as a dependency.

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
jmazanec15: Yes thats my understanding. Let me provide better comment on this.

jmazanec15: Here is what I was basing this off of: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/QueryBuilder.java#L90-L98

jmazanec15: Here is the logic: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/Rewriteable.java#L83

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
```diff
@@ -54,10 +66,20 @@ public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder>
 
     private static final int DEFAULT_K = 10;
 
+    private static MLCommonsClientAccessor ML_CLIENT;
+
+    public static void initialize(MLCommonsClientAccessor mlClient) {
+        NeuralQueryBuilder.ML_CLIENT = mlClient;
+    }
+
```
navneet1v, line 74: Question: Can we init the MLCommonsClientAccessor via the constructor of NeuralQueryBuilder class? is it not possible to do so?

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
navneet1v: > > I think the reason of making onResponse(null) is because we have already set the response to vector supplier.
> 
> We are not setting the response, but instead setting the value of the supplier.
> 

I mean here the value only.

navneet1v: > > Can we add some more details around how this rewrite query function will work with first if condition like vectorSupplier != null and all.
> > Because what I am getting from this whole function is the re-writeQuery function can be called atleast 2 times. where one time vectorSupplier will be null and next time it will not be null. is that understanding correct?
> 
> Yes, will add additional context in the comment. From my understanding, the query will be rewritten until the rewrite method returns the object itself. So, on first rewrite, the supplier will be null. On second rewrite, the supplier will be set, but the vector may be null, depending if the async call finished in time or not. Let me double check this though and add a comment.

My understanding is if we keep on returning the NeuralQueryBuilder, then rewrite will keep on happening(which we are doing). 

To stop query re-write we added `vectorSupplier() != null && vectorSupplier.get() != null) ` condition so that we can return another QueryBuilder(KNNQueryBuilder) which only happens when we have response from MLCommonsClientAccessor. 

Please confirm this understanding is correct or not.

navneet1v (commented):
Overall code looks good to me. Minor comments.

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
zane-neo, line 207: My understanding here: doRewrite method will be invoked by [rewriteAndFetch](https://github.com/opensearch-project/OpenSearch/blob/e44158d4d10d4f8905895ffa50bf9398b8550667/server/src/main/java/org/opensearch/index/query/Rewriteable.java#L109) which is a recursive method by wrapping itself invocation in the listener. And once the predict invocation is done, the vectorSupplier will be populated with valid data, and a KNNQueryBuilder is returned, next recursive call will hit the check on vectorSupplier and return KNNQueryBuilder again, then this method stop. 
When the action gets invoked at [executeAsyncActions](https://github.com/opensearch-project/OpenSearch/blob/e44158d4d10d4f8905895ffa50bf9398b8550667/server/src/main/java/org/opensearch/index/query/QueryRewriteContext.java#L152), the `actionListener.onResponse(null)` here means to invoke the [wildcardListener's on Response](https://github.com/opensearch-project/OpenSearch/blob/e44158d4d10d4f8905895ffa50bf9398b8550667/server/src/main/java/org/opensearch/index/query/QueryRewriteContext.java#L135) to count down.

zane-neo: > On second rewrite, the supplier will be set, but the vector may be null, depending if the async call finished in time or not. Let me double check this though and add a comment.
I think when second rewrite happens, the supplier vector shouldn't be null, because, the second call will only happen when the async call complete because the for loop is returned [here](https://github.com/opensearch-project/OpenSearch/blob/e44158d4d10d4f8905895ffa50bf9398b8550667/server/src/main/java/org/opensearch/index/query/Rewriteable.java#L135). Once the recursive call is complete and a KNNQueryBuilder is returned, the [passed-in listener](https://github.com/opensearch-project/OpenSearch/blob/e44158d4d10d4f8905895ffa50bf9398b8550667/server/src/main/java/org/opensearch/action/search/TransportSearchAction.java#L388) will be executed and continue the query flow.

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
jmazanec15: Good question. I thought a little bit about this, but I am not sure about how to handle stream constructors. Any ideas on this?

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
jmazanec15: Right, so we need to call listener on null so countdown happens

jmazanec15: > My understanding is if we keep on returning the NeuralQueryBuilder, then rewrite will keep on happening(which we are doing).
>
> To stop query re-write we added vectorSupplier() != null && vectorSupplier.get() != null) condition so that we can return another QueryBuilder(KNNQueryBuilder) which only happens when we have response from MLCommonsClientAccessor.
>
> Please confirm this understanding is correct or not.

Yes this is correct.

> I think when second rewrite happens, the supplier vector shouldn't be null, because, the second call will only happen when the async call complete because the for loop is returned here. Once the recursive call is complete and a KNNQueryBuilder is returned, the passed-in listener will be executed and continue the query flow.

Right, thats correct, but I return a copy at the end instead of "this" to prevent a case where rewrite stops early.

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
```diff
@@ -207,12 +212,10 @@ protected QueryBuilder doRewrite(QueryRewriteContext queryRewriteContext) {
...
                 }, actionListener::onFailure)))
             );
-            vectorSupplier(vectorSetOnce::get);
+            return new NeuralQueryBuilder(fieldName(), queryText(), modelId(), k(), vectorSetOnce::get);
         }
 
-        // Rewrites will continuously happen until the supplier is set and the vector is generated. Rewrites will stop
-        // once this object is returned. Hence, if we get here, we need to return a new object
-        return new NeuralQueryBuilder(fieldName(), queryText(), modelId(), k(), vectorSupplier());
+        return this;
```
ylwu-amzn, line 218: I think the third commit fixed the rescore exception of too many rewrite rounds. The key part is no need to return a new NeuralQueryBuilder. That's why the [PoC code](https://github.com/ylwu-amzn/k-NN/commit/d1164974695fe85186d4f0b5f40c5eebfa16e2a7#diff-19fa1b5328b894071230d33a4d215f96ec7cefd853500baf7b770b0fddf53ca6R172) worked for rescore before

ylwu-amzn (approved):
LGTM

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
ylwu-amzn: Thanks for fixing the rescore issue. The code LGTM. 
Please also test other possible use cases.

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
navneet1v: @jmazanec15 please add these code links as documentation on top of this code.

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
```diff
@@ -172,10 +193,35 @@ private static void parseQueryParams(XContentParser parser, NeuralQueryBuilder n
...
+
+        if (vectorSupplier() == null) {
+            SetOnce<float[]> vectorSetOnce = new SetOnce<>();
+            queryRewriteContext.registerAsyncAction(
+                ((client, actionListener) -> ML_CLIENT.inferenceSentence(modelId(), queryText(), ActionListener.wrap(floatList -> {
+                    vectorSetOnce.set(vectorAsListToArray(floatList));
+                    actionListener.onResponse(null);
+                }, actionListener::onFailure)))
+            );
+            return new NeuralQueryBuilder(fieldName(), queryText(), modelId(), k(), vectorSetOnce::get);
```
navneet1v, line 215: [Query] : is there any reason why we are not returning "this" here and returning a new query object? and also I am not able to understand in which case line number 218 will hit?
vectorSupplier can be null or not null.

jmazanec15: We are not returning this here because we are changing the supplier of the query builder. If we return this, this check https://github.com/opensearch-project/OpenSearch/blob/e44158d4d10d4f8905895ffa50bf9398b8550667/server/src/main/java/org/opensearch/index/query/Rewriteable.java#L109 will see the same reference is indicate another round of rewrites does not need to be performed. So instead, we copy it. This is how Geo does it: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/AbstractGeometryQueryBuilder.java#L509.

> vectorSupplier can be null or not null.

Good point. Ill simplify

src/main/java/org/opensearch/neuralsearch/plugin/query/NeuralQueryBuilder.java
jmazanec15: Sure, will leave as is

navneet1v (approved):

PR #25, opened 2022-10-19
Upgrade plugin to 2.4 and refactor zip dependencies

jmazanec15:
### Description
Upgrades plugin to 2.4. Fixes a couple compile time dependencies with ml-commons. 

Refactors dependencies on zips for ml-commons and k-NN. Currently, both ml-commons and k-NN post zips to maven. We already depend on k-NN for the jar. We can refactor dependencies to rely only on these artifacts and not have to download separately.

### Check List


jmazanec15:
Seems that CI is checking out wrong commit: "HEAD is now at 272d803 Integrate model inference to build query (#20)" from https://github.com/opensearch-project/neural-search/actions/runs/3284862131/jobs/5411313863

build.gradle
```diff
@@ -137,22 +131,21 @@ def knnJarDirectory = "$buildDir/dependencies/opensearch-knn"
 dependencies {
     api "org.opensearch:opensearch:${opensearch_version}"
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-knn', version: "${opensearch_build}"
+    zipArchive group: 'org.opensearch', name:'opensearch-ml-plugin', version: "${opensearch_build}"
```
navneet1v, line 134: Why we need ML-Plugin as dependency? Seems like we just need this for integ tests? If yes what was issue with older behavior?

jmazanec15: They dependency is for integ tests.  We specify a custom dependency configuration "zipArchive" to define the scope of the dependency. Declaring a dependency on it this way, we can avoid hardcoding urls and manually downloading from urls.

martin-gaievski (commented):
How we are using zip file with models from resources folder (src/test/resources/model/all-MiniLM-L6-v2.zip)?

jmazanec15:
> How we are using zip file with models from resources folder (src/test/resources/model/all-MiniLM-L6-v2.zip)?

This is a minimal model used for integration tests. Ideally, it would be smaller and/or located somewhere we could download it from, but for now we need to check it in.

vamshin (approved):
LGTM! Thanks

build.gradle
navneet1v: The problem with this approach is opensearch-ml-plugin will be downloaded in those cases where we don't need to run the integration tests. But with the URLs approach we make sure that we download this only when it is required not in every case. I can see this as a downside of this approach. Also, how zip archive is making sure that we are downloading the right platform version of the dependency?

navneet1v (approved):
Approving with minor comments.

build.gradle
jmazanec15: > The problem with this approach is opensearch-ml-plugin will be downloaded in those cases where we don't need to run the integration tests. 

From [here](https://docs.gradle.org/current/userguide/dependency_management.html) I saw this: 
"At runtime, Gradle will locate the declared dependencies if needed for operating a specific task. The dependencies might need to be downloaded from a remote repository, retrieved from a local directory or requires another project to be built in a multi-project setting. This process is called dependency resolution." So, gradle should be able to identify when a dependency is needed and identify that only the k-NN zip is required for compileJava and only download the ml-commons zip for integ test cluster. However, if this isnt the case, we could create a new configuration that extends zipArchive so that only the k-NN zip is downloaded for compileJava. Ill leave this FAR.

> Also, how zip archive is making sure that we are downloading the right platform version of the dependency?

Unsure at the moment how this is done, but will investigate. OpenSearch core dependency is platform specific so Im guessing a similar mechanism is followed.

PR #27, opened 2022-10-24
Change text embedding processor to async mode for better isolation

zane-neo:
### Description
Changed text embedding processor to async mode when handling user input. 
Previously since inferencing runs in async mode, and overriding only the `execute(IngestDocument)` method TextEmbeddingProcessor needs a blocking approach to make sure the document enrichment is complete and then the indexing happens. This has a drawback that the blocking happens in the `write` thread pool, this thread pool has only `availableProcessors + 1` threads, and this could have impact on non text embedding indexing.
Changing this to async mode by overriding `execute(IngestDocument, BiConsumer)` method, thus threads in `write` thread pool will not be blocked and this has better isolation for non text embedding indexing.

### Issues Resolved
An enhancement without issue resolve.

### Check List
    - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
@@ -10,13 +10,14 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.concurrent.ExecutionException;
+import java.util.function.BiConsumer;
 import java.util.function.Supplier;
 import java.util.stream.IntStream;
 
 import lombok.extern.log4j.Log4j2;
 
 import org.apache.commons.lang3.StringUtils;
+import org.opensearch.action.ActionListener;
 import org.opensearch.env.Environment;
 import org.opensearch.index.mapper.MapperService;
 import org.opensearch.ingest.AbstractProcessor;
@@ -80,17 +81,29 @@ public class TextEmbeddingProcessor extends AbstractProcessor {
 
     @Override
     public IngestDocument execute(IngestDocument ingestDocument) {
-        validateEmbeddingFieldsValue(ingestDocument);
-        Map<String, Object> knnMap = buildMapWithKnnKeyAndOriginalValue(ingestDocument);
-        try {
-            List<List<Float>> vectors = mlCommonsClientAccessor.inferenceSentences(this.modelId, createInferenceList(knnMap));
-            appendVectorFieldsToDocument(ingestDocument, knnMap, vectors);
-        } catch (ExecutionException | InterruptedException e) {
-            log.error("Text embedding processor failed with exception: ", e);
-            throw new RuntimeException("Text embedding processor failed with exception", e);
-        }
-        log.debug("Text embedding completed, returning ingestDocument!");
         return ingestDocument;
+    }
+
+    /**
+     * When received a bulk indexing request, the pipeline will be executed in the <a href="https://github.com/opensearch-project/OpenSearch/blob/8fda187bb459757164cc80c91ca305d274ca2b53/server/src/main/java/org/opensearch/action/bulk/TransportBulkAction.java#L226">doInternalExecute</a> method
+     * Before the pipeline execution, the pipeline will be marked as resolved (means executed), and then this overriding method will be invoked when executing the text embedding processor.
+     * After the inference completes, the handler will invoke the doInternalExecute method again to run actual write operation.
+     * @param ingestDocument
+     * @param handler
+     */
+    @Override
+    public void execute(IngestDocument ingestDocument, BiConsumer<IngestDocument, Exception> handler) {
+        try {
+            validateEmbeddingFieldsValue(ingestDocument);
+            Map<String, Object> knnMap = buildMapWithKnnKeyAndOriginalValue(ingestDocument);
+            mlCommonsClientAccessor.inferenceSentences(this.modelId, createInferenceList(knnMap), ActionListener.wrap(x -> {
+                appendVectorFieldsToDocument(ingestDocument, knnMap, x);
+                handler.accept(ingestDocument, null);
+            }, e -> { handler.accept(null, e); }));
+        } catch (Exception e) {
+            handler.accept(null, e);
+        }
+
     }
 
     void appendVectorFieldsToDocument(IngestDocument ingestDocument, Map<String, Object> knnMap, List<List<Float>> vectors) {
--- a/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
@@ -85,19 +85,23 @@ public class TextEmbeddingProcessor extends AbstractProcessor {
     }
 
     /**
-     * When received a bulk indexing request, the pipeline will be executed in the <a href="https://github.com/opensearch-project/OpenSearch/blob/8fda187bb459757164cc80c91ca305d274ca2b53/server/src/main/java/org/opensearch/action/bulk/TransportBulkAction.java#L226">doInternalExecute</a> method
-     * Before the pipeline execution, the pipeline will be marked as resolved (means executed), and then this overriding method will be invoked when executing the text embedding processor.
-     * After the inference completes, the handler will invoke the doInternalExecute method again to run actual write operation.
-     * @param ingestDocument
-     * @param handler
+     * This method will be invoked by PipelineService to make async inference and then delegate the handler to
+     * process the inference response or failure.
+     * @param ingestDocument {@link IngestDocument} which is the document passed to processor.
+     * @param handler {@link BiConsumer} which is the handler which can be used after the inference task is done.
      */
     @Override
     public void execute(IngestDocument ingestDocument, BiConsumer<IngestDocument, Exception> handler) {
+        // When received a bulk indexing request, the pipeline will be executed in this method, (see
+        // https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/bulk/TransportBulkAction.java#L226).
+        // Before the pipeline execution, the pipeline will be marked as resolved (means executed),
+        // and then this overriding method will be invoked when executing the text embedding processor.
+        // After the inference completes, the handler will invoke the doInternalExecute method again to run actual write operation.
         try {
             validateEmbeddingFieldsValue(ingestDocument);
             Map<String, Object> knnMap = buildMapWithKnnKeyAndOriginalValue(ingestDocument);
-            mlCommonsClientAccessor.inferenceSentences(this.modelId, createInferenceList(knnMap), ActionListener.wrap(x -> {
-                appendVectorFieldsToDocument(ingestDocument, knnMap, x);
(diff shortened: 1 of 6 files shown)


jmazanec15:
@zane-neo Please add brief description to PR and/or issue this relates to.

navneet1v:
@zane-neo can you please add, some more details like when client will be sending the bulk request what would be the flow? is there anything changing?

Its good that we are moving towards async just want to understand the flow.

navneet1v:
@zane-neo please resolve the conflicts and also can we check why the checks are failing? as per the last commit all the checks were passing. You might want to do some rebasing.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -80,17 +81,21 @@ private void validateEmbeddingConfiguration(Map<String, Object> fieldMap) {
 
     @Override
     public IngestDocument execute(IngestDocument ingestDocument) {
-        validateEmbeddingFieldsValue(ingestDocument);
-        Map<String, Object> knnMap = buildMapWithKnnKeyAndOriginalValue(ingestDocument);
+        return ingestDocument;
+    }
+
+    public void execute(IngestDocument ingestDocument, BiConsumer<IngestDocument, Exception> handler) {
```
navneet1v, line 87: add proper java doc.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -80,17 +81,21 @@ private void validateEmbeddingConfiguration(Map<String, Object> fieldMap) {
...
-            throw new RuntimeException("Text embedding processor failed with exception", e);
+            validateEmbeddingFieldsValue(ingestDocument);
+            Map<String, Object> knnMap = buildMapWithKnnKeyAndOriginalValue(ingestDocument);
+            mlCommonsClientAccessor.inferenceSentences(this.modelId, createInferenceList(knnMap), ActionListener.wrap(x -> {
+                appendVectorFieldsToDocument(ingestDocument, knnMap, x);
+                handler.accept(ingestDocument, null);
+            }, e -> { handler.accept(null, e); }));
+        } catch (Exception e) {
+            handler.accept(null, e);
         }
```
navneet1v, line 97: which function is throwing the exception which we are catching here?

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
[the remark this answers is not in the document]
zane-neo: Yes, this is an override function, added the override annotation also added proper java doc on this method.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -80,17 +81,21 @@ private void validateEmbeddingConfiguration(Map<String, Object> fieldMap) {
 
     @Override
     public IngestDocument execute(IngestDocument ingestDocument) {
-        validateEmbeddingFieldsValue(ingestDocument);
-        Map<String, Object> knnMap = buildMapWithKnnKeyAndOriginalValue(ingestDocument);
+        return ingestDocument;
+    }
```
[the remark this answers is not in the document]
zane-neo: We can't skip this, there's no where implement this and TextEmbeddingProcessor is not abstract either.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: Done as address in previous comments.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -80,17 +81,29 @@ private void validateEmbeddingConfiguration(Map<String, Object> fieldMap) {
...
+
+    /**
+     * When received a bulk indexing request, the pipeline will be executed in the <a href="https://github.com/opensearch-project/OpenSearch/blob/8fda187bb459757164cc80c91ca305d274ca2b53/server/src/main/java/org/opensearch/action/bulk/TransportBulkAction.java#L226">doInternalExecute</a> method
```
jmazanec15, line 88: minor: Why this commit 8fda187bb459757164cc80c91ca305d274ca2b53 in the link?

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -80,17 +81,29 @@ private void validateEmbeddingConfiguration(Map<String, Object> fieldMap) {
...
+    public void execute(IngestDocument ingestDocument, BiConsumer<IngestDocument, Exception> handler) {
         try {
-            List<List<Float>> vectors = mlCommonsClientAccessor.inferenceSentences(this.modelId, createInferenceList(knnMap));
-            appendVectorFieldsToDocument(ingestDocument, knnMap, vectors);
-        } catch (ExecutionException | InterruptedException e) {
-            log.error("Text embedding processor failed with exception: ", e);
-            throw new RuntimeException("Text embedding processor failed with exception", e);
+            validateEmbeddingFieldsValue(ingestDocument);
+            Map<String, Object> knnMap = buildMapWithKnnKeyAndOriginalValue(ingestDocument);
+            mlCommonsClientAccessor.inferenceSentences(this.modelId, createInferenceList(knnMap), ActionListener.wrap(x -> {
```
jmazanec15, line 99: nit: can we keep name "vectors" instead of "x"

jmazanec15 (commented):
Few minor comments, but overall it looks good.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -80,17 +81,29 @@ private void validateEmbeddingConfiguration(Map<String, Object> fieldMap) {
...
+    }
+
+    /**
+     * When received a bulk indexing request, the pipeline will be executed in the <a href="https://github.com/opensearch-project/OpenSearch/blob/8fda187bb459757164cc80c91ca305d274ca2b53/server/src/main/java/org/opensearch/action/bulk/TransportBulkAction.java#L226">doInternalExecute</a> method
+     * Before the pipeline execution, the pipeline will be marked as resolved (means executed), and then this overriding method will be invoked when executing the text embedding processor.
+     * After the inference completes, the handler will invoke the doInternalExecute method again to run actual write operation.
+     * @param ingestDocument
+     * @param handler
+     */
+    @Override
```
jmazanec15, line 94: In the future, could we refactor this method to batch calls to the model so we can improve throughput?

navneet1v: @jmazanec15 what kind of batching you are suggesting here? Is it like batching the write or batching the inference calls?

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -80,17 +81,29 @@ private void validateEmbeddingConfiguration(Map<String, Object> fieldMap) {
...
-        Map<String, Object> knnMap = buildMapWithKnnKeyAndOriginalValue(ingestDocument);
+        return ingestDocument;
+    }
+
+    /**
+     * When received a bulk indexing request, the pipeline will be executed in the <a href="https://github.com/opensearch-project/OpenSearch/blob/8fda187bb459757164cc80c91ca305d274ca2b53/server/src/main/java/org/opensearch/action/bulk/TransportBulkAction.java#L226">doInternalExecute</a> method
+     * Before the pipeline execution, the pipeline will be marked as resolved (means executed), and then this overriding method will be invoked when executing the text embedding processor.
+     * After the inference completes, the handler will invoke the doInternalExecute method again to run actual write operation.
+     * @param ingestDocument
+     * @param handler
```
navneet1v, line 92: Improve the documentation to explain what this function is doing not that what the whole pipeline is doing, and try to remove the commit ids from the java doc and point to right function using @link of java doc.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: Checked again, not only the method `validateEmbeddingFieldsValue` is throwing exception but also the `mlClient.predict`, and the listener.onFailure is not invoked when exception arise, this exception could populate to `execute(IngestDocument, BiConsumer)`, if not catching it, this will further populate to upper methods which could impact the pipeline execution, this is what we don't want. Also the parent method also use try catch to catch all the exceptions please refer to: https://github.com/opensearch-project/OpenSearch/blob/8c9ca4e858e6333265080972cf57809dbc086208/server/src/main/java/org/opensearch/ingest/Processor.java#L64.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: Sure, change to vectors

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: Sure, add the pipeline execution in the method body.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: Generally it's doable, but it's risky to make this a bulk approach now since the result mapping is complicated and error prone.

jmazanec15: @zane-neo That makes sense. It might make sense to create a proof of concept test in future to see what the benefit would be.

@navneet1v the idea would be to batch the inference calls. For the async action, instead of calling inference directly, submit the update to some kind of queue that will then create 1 large request for multiple docs and call the handlers once it completes.

jmazanec15 (approved):
Failure https://github.com/opensearch-project/neural-search/actions/runs/3328176376/jobs/5504257477 is unrelated to this PR. It appears to be related to ml-commons permission issue. Approving.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
navneet1v: > @zane-neo That makes sense. It might make sense to create a proof of concept test in future to see what the benefit would be.
> 
> @navneet1v the idea would be to batch the inference calls. For the async action, instead of calling inference directly, submit the update to some kind of queue that will then create 1 large request for multiple docs and call the handlers once it completes.

I am not like 100% convinced on the idea of batching and the reason behind that is batching the data and making 1 call, will help if the number of threads are less at the ML node level and multiple requests are competing for the threads(api call threads). 
Secondly, for creating the batch we need to put the data in a sync queue, that will further slow down the processing.
Third, batches create problem where 1 single input sentence can delay all the documents processing. 

But we can discuss further on this, but before even doing the POC, we should first do some deep-dive on the ML-Commons code to know if batching can really help or not.

navneet1v (approved):

PR #48, opened 2022-11-01
Switch processor IT to use Lucene

jmazanec15:
### Description
Currently, Processor IT silently fails because it relies on k-NN plugin's nmslib jni library which is not shipped with the zip at the moment. It fails silently because on ingest, the index is not refreshed.

Updated processor test to use Lucene engine to avoid dependency issue. Refactored ingest document to refresh after ingestion and confirm doc gets indexed.

### Check List


vamshin (approved):
LGTM! Thanks.

It will be good to have these tests run with other engines as well. We can take in other Pr

navneet1v (approved):

PR #72, opened 2022-11-18
Change the behavior when embedding fields not present

### LINKED ISSUE #71: Not enough error information in error response when ingestion fails due to fields not present.
A customer got an error while trying out the neural plugin. Steps followed

Forum link: https://forum.opensearch.org/t/feedback-neural-search-plugin-experimental-release/11501/4

```
I got an error with the example in the documentation. Below is the code I tried. I tested the model and it is loading and working fine. However, ingesting the document fails giving the below error!

Environment: This was done on windows in development mode (one node as cluster_manager, data, ingest and ml)

Error:
{
“error” : {
“root_cause” : [
{
“type” : “illegal_argument_exception”,
“reason” : “empty docs”
}
],
“type” : “illegal_argument_exception”,
“reason” : “empty docs”
},
“status” : 400
}

My Code:
POST /_plugins/_ml/models/_upload
{
“name”: “all-MiniLM-L6-v2”,
“version”: “1.0.0”,
“description”: “test model”,
“model_format”: “TORCH_SCRIPT”,
“model_config”: {
“model_type”: “bert”,
“embedding_dimension”: 384,
“framework_type”: “sentence_transformers”
},
“url”: “[https://github.com/ylwu-amzn/ml-commons/blob/2.x_custom_m_helper/ml-algorithms/src/test/resources/org/opensearch/ml/engine/algorithms/text_embedding/all-MiniLM-L6-v2_torchscript_sentence-transformer.zip?raw=true 1](https://github.com/ylwu-amzn/ml-commons/blob/2.x_custom_m_helper/ml-algorithms/src/test/resources/org/opensearch/ml/engine/algorithms/text_embedding/all-MiniLM-L6-v2_torchscript_sentence-transformer.zip?raw=true)”
}

POST /_plugins/_ml/models/kI6NhoQB3oLQzIJTkldg/_load

POST /_plugins/_ml/models/kI6NhoQB3oLQzIJTkldg/_predict
{
“text_docs”:[ “today is sunny”]
}

PUT _ingest/pipeline/nlp-pipeline
{
“description”: “An example neural search pipeline”,
“processors” : [
{
“text_embedding”: {
“model_id”: “kI6NhoQB3oLQzIJTkldg”,
“field_map”: {
“text”: “text_knn”
}
}
}
]
}

PUT /my-nlp-index-1
{
“settings”: {
“index.knn”: true,

    "default_pipeline": "nlp-pipeline"
},
"mappings": {
    "properties": {
        "passage_embedding": {
            "type": "knn_vector",
            "dimension": 384,
            "method": {
                "name": "hnsw",
                "space_type": "l2",
                "engine": "nmslib",
                "parameters": {
                  "ef_construction": 128,
                  "m": 24
                }
            }
        },
        "passage_text": { 
            "type": "text"            
        }
    }
}
}

POST my-nlp-index-1/_doc
{
“passage_text”: “Hello world”
}
```

zane-neo:
Signed-off-by: Zan Niu <zaniu@amazon.com>

### Description
This PR is to optimize the error prompt message when inference list is empty, which is caused by user configuration error or expected fields empty. Previously the error message is too general for user to understand what's the root cause.

### Issues Resolved
#71 & #73 

### Check List
    - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
@@ -102,12 +102,13 @@ public class TextEmbeddingProcessor extends AbstractProcessor {
             Map<String, Object> knnMap = buildMapWithKnnKeyAndOriginalValue(ingestDocument);
             List<String> inferenceList = createInferenceList(knnMap);
             if (inferenceList.size() == 0) {
-                throw new IllegalArgumentException("Unable to process embedding since no text found from corresponding source fields");
-            }
-            mlCommonsClientAccessor.inferenceSentences(this.modelId, inferenceList, ActionListener.wrap(vectors -> {
-                appendVectorFieldsToDocument(ingestDocument, knnMap, vectors);
                 handler.accept(ingestDocument, null);
-            }, e -> { handler.accept(null, e); }));
+            } else {
+                mlCommonsClientAccessor.inferenceSentences(this.modelId, inferenceList, ActionListener.wrap(vectors -> {
+                    appendVectorFieldsToDocument(ingestDocument, knnMap, vectors);
+                    handler.accept(ingestDocument, null);
+                }, e -> { handler.accept(null, e); }));
+            }
         } catch (Exception e) {
             handler.accept(null, e);
         }
--- a/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
@@ -100,7 +100,11 @@ public class TextEmbeddingProcessor extends AbstractProcessor {
         try {
             validateEmbeddingFieldsValue(ingestDocument);
             Map<String, Object> knnMap = buildMapWithKnnKeyAndOriginalValue(ingestDocument);
-            mlCommonsClientAccessor.inferenceSentences(this.modelId, createInferenceList(knnMap), ActionListener.wrap(vectors -> {
+            List<String> inferenceList = createInferenceList(knnMap);
+            if (inferenceList.size() == 0) {
+                throw new IllegalArgumentException("Unable to process embedding since no text found from corresponding source fields");
+            }
+            mlCommonsClientAccessor.inferenceSentences(this.modelId, inferenceList, ActionListener.wrap(vectors -> {
                 appendVectorFieldsToDocument(ingestDocument, knnMap, vectors);
                 handler.accept(ingestDocument, null);
             }, e -> { handler.accept(null, e); }));
--- a/src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorTests.java
@@ -127,10 +127,8 @@ public class TextEmbeddingProcessorTests extends OpenSearchTestCase {
         verify(handler).accept(isNull(), any(RuntimeException.class));
     }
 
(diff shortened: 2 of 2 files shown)


jmazanec15:
@zane-neo add #71 to the description under issues resolved?

jmazanec15 (approved):

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -100,7 +100,11 @@ public void execute(IngestDocument ingestDocument, BiConsumer<IngestDocument, Ex
         try {
             validateEmbeddingFieldsValue(ingestDocument);
             Map<String, Object> knnMap = buildMapWithKnnKeyAndOriginalValue(ingestDocument);
-            mlCommonsClientAccessor.inferenceSentences(this.modelId, createInferenceList(knnMap), ActionListener.wrap(vectors -> {
+            List<String> inferenceList = createInferenceList(knnMap);
```
navneet1v, line 103: Can we validate and see what will happen if knnMap is empty or null?

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -100,7 +100,11 @@ public void execute(IngestDocument ingestDocument, BiConsumer<IngestDocument, Ex
...
+            if (inferenceList.size() == 0) {
+                throw new IllegalArgumentException("Unable to process embedding since no text found from corresponding source fields");
+            }
```
navneet1v, line 106: The inferenceList can be empty if the field which needs to inference is not present in the document. Please handle that use case in this PR.

Case when inference fields not present:
1. If for the document the field is genuinely not present.
2. In the update document use case the field to inference is not present as it doesn't need to be updated.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: The knnMap won't be null since it's manually created, but it could be empty if user doesn't configure anything in the pipeline field_map. This case the inference list will be empty.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zane-neo: @jmazanec15 @navneet1v Yeah, This is something I want to discuss with you guys. The error information in fact is not a real issue. Since the cases Navneet mentioned above, we should succeed the ingestion without any exception raised. This way user will never see the exception so no need to handle the proper prompt. I've changed the code in this PR, please take a look.

navneet1v: @zane-neo 
I checked the code and it handles the case when user is updating the document, or even in the case when he is indexing the document for first time and he misses the fields. While this looks good. But the current code doesn't handle the case when user wants the ingestion request to fail because of lack of fields. I think we should think of that case too.

zane-neo: @navneet1v These two are conflict cases, if we prompt the user lacking of fields which means we don't allow user's insert/update(particular without embedding fields). If we allow user to ingest without embedding fields, we shouldn't raise error to user. I prefer allowing user's ingestion without embedding fields, since this is a valid case for user(Some documents do not have embedding fields).

navneet1v: @zane-neo The idea which I was thinking is let customer configure what they want to do in case the embedding fields are not present. One way to do this is by adding a parameter in the processor. But as of now I think we can go ahead with this change.

PR #91, opened 2022-12-27
Add retry mechanism for neural search inference.

zane-neo:
### Description
This feature is an enhancement to neural-search and the purpose is to improve the availability of the opensearch ml cluster. The core idea is add retry for inference operation, once a request is been sending or have sent to a ml instance which is not available now, rather than simply fail this request we use retry to improve the success rate of requests.

### Issues Resolved
https://github.com/opensearch-project/neural-search/issues/92

### Check List
    - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
@@ -5,9 +5,15 @@
 
 package org.opensearch.neuralsearch.ml;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.log4j.Log4j2;
+
 import org.opensearch.action.ActionListener;
 import org.opensearch.ml.client.MachineLearningNodeClient;
 import org.opensearch.ml.common.FunctionName;
@@ -20,11 +26,6 @@ import org.opensearch.ml.common.output.model.ModelTensor;
 import org.opensearch.ml.common.output.model.ModelTensorOutput;
 import org.opensearch.ml.common.output.model.ModelTensors;
 import org.opensearch.neuralsearch.util.RetryUtil;
-
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.stream.Collectors;
 
 /**
  * This class will act as an abstraction on the MLCommons client for accessing the ML Capabilities
@@ -123,7 +124,6 @@ public class MLCommonsClientAccessor {
             }
         }));
     }
-
 
     private MLInput createMLInput(final List<String> targetResponseFilters, List<String> inputText) {
         final ModelResultFilter modelResultFilter = new ModelResultFilter(false, true, targetResponseFilters, null);
--- a/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
@@ -108,11 +108,11 @@ public class MLCommonsClientAccessor {
     }
 
     private void inferenceSentencesWithRetry(
-        @NonNull final List<String> targetResponseFilters,
-        @NonNull final String modelId,
-        @NonNull final List<String> inputText,
+        final List<String> targetResponseFilters,
+        final String modelId,
+        final List<String> inputText,
         final int retryTime,
-        @NonNull final ActionListener<List<List<Float>>> listener
+        final ActionListener<List<List<Float>>> listener
     ) {
         MLInput mlInput = createMLInput(targetResponseFilters, inputText);
         mlClient.predict(modelId, mlInput, ActionListener.wrap(mlOutput -> {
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -98,13 +103,32 @@ public void inferenceSentences(
...
+    ) {
+        inferenceSentencesWithRetry(targetResponseFilters, modelId, inputText, 0, listener);
+    }
+
+    private void inferenceSentencesWithRetry(
+        @NonNull final List<String> targetResponseFilters,
+        @NonNull final String modelId,
+        @NonNull final List<String> inputText,
+        final int retryTime,
+        @NonNull final ActionListener<List<List<Float>>> listener
```
navneet1v, line 115: remove @NonNull from private functions. Given that it is private function we can be sure that the inputs shouldn't be null.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -98,13 +103,32 @@ public void inferenceSentences(
...
         MLInput mlInput = createMLInput(targetResponseFilters, inputText);
         mlClient.predict(modelId, mlInput, ActionListener.wrap(mlOutput -> {
             final List<List<Float>> vector = buildVectorFromResponse(mlOutput);
             log.debug("Inference Response for input sentence {} is : {} ", inputText, vector);
             listener.onResponse(vector);
-        }, listener::onFailure));
+        }, e -> {
+            final int nodeNotConnectedExceptionIndex = ExceptionUtils.indexOfThrowable(e, NodeNotConnectedException.class);
+            final int nodeDisconnectExceptionIndex = ExceptionUtils.indexOfThrowable(e, NodeDisconnectedException.class);
+            if ((nodeDisconnectExceptionIndex != -1 || nodeNotConnectedExceptionIndex != -1) && retryTime < MAX_RETRY) {
```
navneet1v, line 125: We should encapsulate this in a different method which can give a boolean response if we should do a retry or not. This will allow us in future to add more retryable exceptions and will also allow us to evolve the conditions for doing the retry.

src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
```diff
@@ -114,6 +119,28 @@ public void testInferenceSentences_whenExceptionFromMLClient_thenFailure() {
         Mockito.verifyNoMoreInteractions(resultListener);
     }
 
+    public void testInferenceSentences_whenNodeNotConnectedException_thenRetry() {
```
navneet1v, line 122: Add these test cases also:
1. Test case which tests when all the retries are finished.
2. Test cases which test when these retryable exceptions are not thrown we should get a failure.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
zane-neo: Removed.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
zane-neo: Refactored to a new method.

src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
zane-neo: Two cases covered:
1. check the method `testInferenceSentences_whenNodeNotConnectedException_thenRetry_3Times` and the `Mockito.verify(client, times(4))` stands for the client method has been invoked for 4 times in total.
2. Add a new method `testInferenceSentences_whenNotConnectionException_thenNoRetry` and the verify has time of 1, meaning method only run once.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -98,13 +103,36 @@ public void inferenceSentences(
...
+            } else {
+                listener.onFailure(e);
+            }
+        }));
+    }
+
+    private boolean shouldRetry(Exception e, int retryTime) {
+        final int nodeNotConnectedExceptionIndex = ExceptionUtils.indexOfThrowable(e, NodeNotConnectedException.class);
+        final int nodeDisconnectExceptionIndex = ExceptionUtils.indexOfThrowable(e, NodeDisconnectedException.class);
+        return (nodeDisconnectExceptionIndex != -1 || nodeNotConnectedExceptionIndex != -1) && retryTime < MAX_RETRY;
```
navneet1v, line 135: I would suggest 1 more simplification here:

```suggestion
    private boolean shouldRetry(Exception e, int retryTime) {
        
        return (RetryUtil.containsException(INFERENCE_API_RETRYABLE_EXCEPTION_LIST, e)) && retryTime < MAX_RETRY;
}

class RetryUtil {
    
    public static boolean containsException(@NonNull final List<Class<?>> exceptionClassesList, final Exception ex) {
     // call below function to check if any exception is present and then return true.
    }
    
    private static boolean containsException(final Exception ex, final ClassClass<?> exceptionClazz) {
         return ExceptionUtils.indexOfThrowable(ex, exceptionClazz) != -1;
    }
}

```

zane-neo: In fact, we have analyzed all the exceptions if they're retryable and it doesn't seem to have more exceptions could be retried, otherwise the retry may cause cluster level instability, I wouldn't suggest go too far and over design for this for now.

jmazanec15: @zane-neo This logic would retry if a model is not loaded, correct?

jmazanec15 (commented):
Retries would be useful for ephemeral issues, correct? So which exceptions would we expect to be resolved on failure like this?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
navneet1v: > In fact, we have analyzed all the exceptions if they're retryable and it doesn't seem to have more exceptions could be retried, otherwise the retry may cause cluster level instability, I wouldn't suggest go too far and over design for this for now.

So you are saying there will be no case in future where we need add exception? Thats very strange. I would recommend lets build retry mechanism extensible. 

Also the suggested code is just an example.

zane-neo: > @zane-neo This logic would retry if a model is not loaded, correct?

No, this logic only retries the network issue: NodeNotConnectedException and NodeDisconnectedException. When a model is not loaded the requests won't be dispatched to that node at the beginning.

zane-neo: > > In fact, we have analyzed all the exceptions if they're retryable and it doesn't seem to have more exceptions could be retried, otherwise the retry may cause cluster level instability, I wouldn't suggest go too far and over design for this for now.
> 
> So you are saying there will be no case in future where we need add exception? Thats very strange. I would recommend lets build retry mechanism extensible.
> 
> Also the suggested code is just an example.

Are you talking about business level retries? These should be retried at underlying layers to avoid network calls. But I can do the refactor to the potential business retries at client side.

navneet1v: > Are you talking about business level retries?

No.

I am not saying to refractor any code at client side. I am saying just refractor the code at the Neural Search plugin side as suggested in my older comment.

zane-neo:
> Retries would be useful for ephemeral issues, correct? So which exceptions would we expect to be resolved on failure like this?

Two exceptions we're going to handle are: NodeNotConnectedException and NodeDisconnectedException

jmazanec15 (approved):

navneet1v (approved):

Issue #194, opened 2023-06-03
[FEATURE] New Doc Collector for Normalization and Score Combination Query


martin-gaievski:
## Description

For Normalization and Score Combination feature, we need to collect results of different sub-queries and pass those results for actual post processing (normalization and score combination) before the Fetch phase.

## Solution

Solution we are proposing is to create new implementation of a [Collector](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/Collector.java) for results.

Because new collector cannot be plugged to OpenSearch directly (there no suitable extension points) we also need a new implementation of [QueryPhaseSearcher](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/query/QueryPhaseSearcher.java). Plugin can register new QueryPhaseSearcher (with certain limitations), and that searcher will be executing new collector. 

## Tasks

- [X] Create RFC for new Document Collector: https://github.com/opensearch-project/neural-search/issues/193
- [x] Implementation of a Query clause
- [x] Testing

## Reference Links
1. https://github.com/opensearch-project/neural-search/issues/126
2. https://github.com/opensearch-project/neural-search/issues/174
3. https://github.com/opensearch-project/neural-search/issues/193

PR #198, opened 2023-06-09
Add Doc collector, required for Query Phase Searcher 

martin-gaievski:
### Description
Adding doc collector for Normalization feature. This will be used by Query Phase Searcher. Future QueryPhaseSearcher will call IndexSearcher.search(query, docCollector) to accumulate query results and then collector.topDocs() to get final DTO with scores for each sub-query, similar to what is done in [this test](https://github.com/opensearch-project/neural-search/pull/198/files#diff-bd9df5c4786267ae2499fbc95d99106f1f8f9f76ae7d37e40e55374e00d56acd).

Basic refactoring in test classes, pulled methods that are common between few test classes into base test class

More tests will be added in later PRs. 

### Issues Resolved
https://github.com/opensearch-project/neural-search/issues/194

### Check List
    - [X] All tests pass


src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
```diff
@@ -39,7 +39,7 @@ public final class HybridQueryScorer extends Scorer {
 
     private final Map<Query, Integer> queryToIndex;
 
-    HybridQueryScorer(Weight weight, List<Scorer> subScorers) throws IOException {
+    public HybridQueryScorer(Weight weight, List<Scorer> subScorers) throws IOException {
```
martin-gaievski, line 42: Increase visibility of constructor for tests, previous package default was not required

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
```diff
@@ -0,0 +1,158 @@
...
+@Log4j2
+public class HybridTopScoreDocCollector implements Collector {
+    int docBase;
+    float minCompetitiveScore;
+    final HitsThresholdChecker hitsThresholdChecker;
+    ScoreDoc pqTop;
+    protected TotalHits.Relation totalHitsRelation = TotalHits.Relation.EQUAL_TO;
+    protected int[] totalHits;
+    public static final TopDocs EMPTY_TOPDOCS = new TopDocs(new TotalHits(0, TotalHits.Relation.EQUAL_TO), new ScoreDoc[0]);
+    int numOfHits;
```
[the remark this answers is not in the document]
martin-gaievski: ack, will change

src/main/java/org/opensearch/neuralsearch/search/HybridQueryTopDocs.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.lucene.search.TotalHits;
+
+/**
+ * Class stores collection of TodDocs for each sub query from hybrid query
+ */
+@ToString(includeFieldNames = true)
+public class HybridQueryTopDocs extends TopDocs {
+
+    @Getter
+    TopDocs[] hybridQueryTopdDocs;
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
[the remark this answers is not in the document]
martin-gaievski: ack, will change

src/main/java/org/opensearch/neuralsearch/search/HybridQueryTopDocs.java
```diff
@@ -0,0 +1,45 @@
...
+    public HybridQueryTopDocs(TotalHits totalHits, ScoreDoc[] scoreDocs) {
+        super(totalHits, scoreDocs);
+    }
+
+    public HybridQueryTopDocs(TotalHits totalHits, TopDocs[] docs) {
+        super(totalHits, copyScoreDocs(docs[0].scoreDocs));
+        this.hybridQueryTopdDocs = docs;
+    }
+
+    private static ScoreDoc[] copyScoreDocs(ScoreDoc[] original) {
```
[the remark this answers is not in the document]
martin-gaievski: static is actually required as we're calling this before the super constructor, it's the utility method to get copy of first array element and pass that to super constructor.

navneet1v (approved):

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package org.opensearch.neuralsearch.search;
+
+import java.util.Locale;
+
+import org.apache.lucene.search.ScoreMode;
+
+/**
+ *  Defines algorithm to allow searches to terminate early
```
jmazanec15, line 13: Add more context in description

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public class HitsThresholdChecker {
+    private int hitCount;
+    private final int totalHitsThreshold;
+
+    private HitsThresholdChecker(int totalHitsThreshold) {
+        if (totalHitsThreshold < 0) {
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "totalHitsThreshold must be >= 0, got %d", totalHitsThreshold));
+        }
+        assert totalHitsThreshold != Integer.MAX_VALUE;
```
jmazanec15, line 23: Should this be an assertion error or should we throw an IllegalArgumentException

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
```diff
@@ -0,0 +1,46 @@
...
+        this.totalHitsThreshold = totalHitsThreshold;
+    }
+
+    int getHitsThreshold() {
```
jmazanec15, line 27: lombok?

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
```diff
@@ -0,0 +1,46 @@
...
+    void incrementHitCount() {
+        ++hitCount;
+    }
+
+    boolean isThresholdReached() {
+        return hitCount > getHitsThreshold();
+    }
+
+    ScoreMode scoreMode() {
+        return ScoreMode.TOP_SCORES;
```
jmazanec15, line 40: Whats the point of this?

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
martin-gaievski: will change to exception, assertion is not a requirement here

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
jmazanec15: But this class does not implement Collector interface

martin-gaievski: right, this method called from doc collector https://github.com/martin-gaievski/neural-search/blob/feature/normalization-query-phase-searcher/src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java#L93, I found that in other doc collectors it's done in such way, e.g. [TopScoreDocCollector](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java#L314)

jmazanec15: Shouldnt we extend or use https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/HitsThresholdChecker.java?

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
```diff
@@ -0,0 +1,48 @@
...
+        }
+        this.totalHitsThreshold = totalHitsThreshold;
+    }
+
+    void incrementHitCount() {
+        ++hitCount;
+    }
+
+    boolean isThresholdReached() {
+        return hitCount > getTotalHitsThreshold();
```
jmazanec15, line 38: Wouldnt it be reached when it is ==?

src/main/java/org/opensearch/neuralsearch/search/HybridQueryTopDocs.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.apache.lucene.search.ScoreDoc;
+import org.apache.lucene.search.TopDocs;
+import org.apache.lucene.search.TotalHits;
+
+/**
+ * Class stores collection of TodDocs for each sub query from hybrid query
+ */
+@ToString(includeFieldNames = true)
+public class HybridQueryTopDocs extends TopDocs {
```
jmazanec15, line 21: Is HybridQueryTopDocs the right name? This class is basically a TopDocs implementation that stores multiple TopDocs - so maybe CompondTopDocs would be better?

jmazanec15: I dont think the name of this class needs to map to the query api. It should be descriptive of its functionality. I think CompoundTopDocs makes sense.

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
```diff
@@ -0,0 +1,157 @@
...
+            float localMinScore = Math.nextUp(pqTop.score);
+            if (localMinScore > minCompetitiveScore) {
+                scorer.setMinCompetitiveScore(localMinScore);
+                totalHitsRelation = TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO;
+                minCompetitiveScore = localMinScore;
+            }
+        }
+    }
+
+    public TopDocs[] topDocs() {
```
jmazanec15, line 109: add comment

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
martin-gaievski: oh nice catch, I think not inclusive it more logical here, let me change

src/main/java/org/opensearch/neuralsearch/search/HybridQueryTopDocs.java
martin-gaievski: I actually put it as CompoundTopDocs in design initially, but then changed to hybrid to make it consistent. If at least one more person came to same idea I'll change it to CompoundTopDocs, will push commit soon

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
martin-gaievski: for other question, can if use https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/HitsThresholdChecker.java, it's not a public class and factory method is not public, so no use and no extension possible

src/main/java/org/opensearch/neuralsearch/search/HybridQueryTopDocs.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Getter
+    private TopDocs[] hybridQueryTopdDocs;
+
+    public HybridQueryTopDocs(TotalHits totalHits, ScoreDoc[] scoreDocs) {
+        super(totalHits, scoreDocs);
+    }
+
+    public HybridQueryTopDocs(TotalHits totalHits, TopDocs[] docs) {
+        super(totalHits, cloneScoreDocs(docs[0].scoreDocs));
```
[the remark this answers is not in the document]
martin-gaievski: on second though let me change this logic, I think we can return TopDocs of sub-query with max number of hits, that will match length of result set that we return and read later in injector processor

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
```diff
@@ -0,0 +1,48 @@
...
+    private int hitCount;
+    @Getter
+    private final int totalHitsThreshold;
+
+    private HitsThresholdChecker(int totalHitsThreshold) {
+        if (totalHitsThreshold < 0) {
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "totalHitsThreshold must be >= 0, got %d", totalHitsThreshold));
+        }
+        if (totalHitsThreshold == Integer.MAX_VALUE) {
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "totalHitsThreshold must be less than max integer value"));
```
heemin32, line 28: What is wrong with `totalHitsThreshold` having `Integer.MAX_VALUE` but okay with `Integer.MAX_VALUE - 1`?

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
```diff
@@ -0,0 +1,48 @@
...
+        }
+        this.totalHitsThreshold = totalHitsThreshold;
+    }
+
+    void incrementHitCount() {
```
heemin32, line 33: Can we put either `protected` or `public` unless there is a specific reason for it to be default? I think having explicit accessor has better readability.

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
heemin32: Shouldn't we extend https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/HitsThresholdChecker.java#L23?

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
```diff
@@ -0,0 +1,48 @@
...
+
+    boolean isThresholdReached() {
+        return hitCount >= getTotalHitsThreshold();
+    }
+
+    ScoreMode scoreMode() {
+        return ScoreMode.TOP_SCORES;
+    }
+
+    public static HitsThresholdChecker create(final int totalHitsThreshold) {
```
heemin32, line 45: Why do we want to have this method? Can we just call a constructor?

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
```diff
@@ -0,0 +1,48 @@
...
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "totalHitsThreshold must be >= 0, got %d", totalHitsThreshold));
+        }
+        if (totalHitsThreshold == Integer.MAX_VALUE) {
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "totalHitsThreshold must be less than max integer value"));
+        }
+        this.totalHitsThreshold = totalHitsThreshold;
+    }
+
+    void incrementHitCount() {
+        ++hitCount;
```
heemin32, line 34: Shouldn't it be thread safe?

src/main/java/org/opensearch/neuralsearch/search/CompoundTopDocs.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Getter
+    private TopDocs[] compoundTopDocs;
+
+    public CompoundTopDocs(TotalHits totalHits, ScoreDoc[] scoreDocs) {
+        super(totalHits, scoreDocs);
+    }
+
+    public CompoundTopDocs(TotalHits totalHits, TopDocs[] docs) {
+        // we pass clone of score docs from the sub-query that has most hits
```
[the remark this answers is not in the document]
martin-gaievski: problem they are facing is that TopDocs is used by core classes and we can't extend or change those classes. So we're keeping TopDocs up to the contract and also add out specific information that we'll use in later stages from pipeline post processor. Constructor ```CompoundTopDocs(TotalHits totalHits, ScoreDoc[] scoreDocs)``` is existing core super constructor, we can't change it. ```CompoundTopDocs(TotalHits totalHits, TopDocs[] docs)``` is what we added to have sub-queries results. 

As per core contract we can have only one set of results, and only one query. In case of hybrid search some sub-queries at shard level may have no hits, so if we pick those sub-query results and use them it may influence later stages (in case they do have some optimizations to drop such shard results). Picking one with most results (any in there is a tie) looks more relevant for hybrid queries, as after normalization and combination we'll get max (total_hits) results. 

For example:

shard level results:
```
sub-query_1 {doc2: 1.0},
sub-query_2 {doc1: 0.4, doc2: 2.0},
sub_query_3 {}
```
we build CompoundTopDocs
```
{doc1: 0.4, doc2: 2.0}, total_hits: 2
```

after scores combination:
```
{doc2: 3.0, doc1: 0.4} 
```

resulting list looks much more like sub_query results with max hits.

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
```diff
@@ -0,0 +1,161 @@
...
+            );
+        }
+
+        if (start < 0) {
+            throw new IllegalArgumentException(
+                String.format(Locale.ROOT, "Expected value of starting position is between 0 and %d, got %d", size, start)
+            );
+        }
+
+        if (start >= size || howMany == 0) {
```
heemin32, line 138:
```suggestion
        if (start >= howMany || howMany == 0) {
```

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
```diff
@@ -0,0 +1,161 @@
...
+            return EMPTY_TOPDOCS;
+        }
+
+        howMany = Math.min(size - start, howMany);
```
heemin32, line 142: I would not change value in parameter variable.
```suggestion
        int size = Math.min(howMany - start, howMany);
```

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
```diff
@@ -0,0 +1,161 @@
...
+        ScoreDoc[] results = new ScoreDoc[howMany];
```
heemin32, line 143:
```suggestion
        ScoreDoc[] results = new ScoreDoc[size];
```

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
```diff
@@ -0,0 +1,161 @@
...
+        // pq's pop() returns the 'least' element in the queue, therefore need
+        // to discard the first ones, until we reach the requested range.
+        for (int i = pq.size() - start - howMany; i > 0; i--) {
```
heemin32, line 146:
```suggestion
        for (int i = pq.size() - start - size; i > 0; i--) {
```

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
```diff
@@ -0,0 +1,161 @@
...
+            pq.pop();
+        }
+
+        // Get the requested results from pq.
+        populateResults(results, howMany, pq);
```
heemin32, line 151:
```suggestion
        populateResults(results, size, pq);
```

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
```diff
@@ -0,0 +1,161 @@
...
+
+        return new TopDocs(new TotalHits(totalHits, totalHitsRelation), results);
+    }
+
+    protected void populateResults(ScoreDoc[] results, int howMany, PriorityQueue<ScoreDoc> pq) {
```
heemin32, line 156: No need to check if `results` size is big enough for howMany value and `pq` also has enough items in it?

src/main/java/org/opensearch/neuralsearch/search/CompoundTopDocs.java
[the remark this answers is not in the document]
heemin32: What happen in this case? Will it still return two results?
```
sub-query_1 {doc2: 1.0},
sub-query_2 {doc1: 0.4},
sub_query_3 {}
```

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
martin-gaievski: we don't want to spill over and reset to negative, this what will happen on next increment if current value is ```Integer.MAX_VALUE```. ```Integer.MAX_VALUE - 1``` doesn't have such issue. 
For clarification, we're counting number of hits for one sub-query in one shard. Integer type should handle max number of hits (used by Lucene in similar class https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/HitsThresholdChecker.java)

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
martin-gaievski: we cannot extend, Lucene class isn't public.
no specific reason for factory method, will change to plain constructor

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
martin-gaievski: it's not necessary, for each query it will be new instance of checker and executions in shard are single-threaded

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
martin-gaievski: it's checked in previous logic, but let me create one more explicit check here

src/main/java/org/opensearch/neuralsearch/search/CompoundTopDocs.java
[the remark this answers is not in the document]
martin-gaievski: In this case CompoundTopDocs will have:
```
ScoreDocs: {doc2: 1.0}, totalHits = 1

TopDocs: 
{doc2: 1.0},
{doc1: 0.4},
{}
```
then after normalization, if  query has ```size=1```  then we will return:
```
{doc2: 1.0}
```
because it has the highest score, if ```size >= 2``` then return
```
{doc2: 1.0},
{doc1: 0.4}
```

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
```diff
@@ -131,31 +129,34 @@ private TopDocs topDocsPerQuery(int start, int howMany, PriorityQueue<ScoreDoc>
...
 
         // Get the requested results from pq.
-        populateResults(results, howMany, pq);
+        populateResults(results, size, pq);
 
         return new TopDocs(new TotalHits(totalHits, totalHitsRelation), results);
     }
 
     protected void populateResults(ScoreDoc[] results, int howMany, PriorityQueue<ScoreDoc> pq) {
         for (int i = howMany - 1; i >= 0; i--) {
```
heemin32, line 155:
```suggestion
        for (int i = howMany - 1; i >= 0 && i < results.length && pq.isEmpty() == false; i--) {
```

martin-gaievski: I think it's not exactly same as we do have now, we may not even start the loop. E.g. if heap has elements and ```howMany = 10``` and ```results.length = 4```, then with existing logic we'll do 4 iterations, but with suggested change there will be none.
I'll add the heap size check though, it makes sense

PR #214, opened 2023-07-12
Add feature flag for QueryPhaseSearcher

martin-gaievski:
### Description
Added feature flag ```neural_search_hybrid_search_enabled``` for query phase searcher. This is required as a safeguard in case concurrent search feature is enabled and another implementation of query phase searcher is registered. 
This new flag can be set as system property or in opensearch.yml (with use of ```transport.features```), by default flag is ```false``` (feature is disabled). Enable flag for integ tests.

### Issues Resolved
https://github.com/opensearch-project/neural-search/issues/194

### Check List
    - [X] All tests pass


src/main/java/org/opensearch/neuralsearch/util/PluginFeatureFlags.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public class PluginFeatureFlags {
+
+    /**
+     * Used to test feature flags whose values are expected to be booleans.
+     * This method returns true if the value is "true" (case-insensitive),
+     * and false otherwise.
+     * Checks alternative flag names as they may be different for plugins
+     */
+    public static boolean isEnabled(String featureFlagName) {
```
martin-gaievski, line 23: we need to try "transport" feature prefix, as this is the way core adding features that are set in plugins.

heemin32, line 23:
```suggestion
    public static boolean isEnabled(final String featureFlagName) {
```

heemin32: Then why both? Should we only check with "transport" feature prefix?

martin-gaievski: I wanted to encapsulate that logic so client can do only one call. Otherwise on a client we need to do two calls:
```featureFlags.isEnabled(flag) || customFeatureFlags.isEnabled(flag)```. Depending on scenario one feature flag can be set in one form or another (with or without prefix)

Comment (martin-gaievski) [reply src/main/java/org/opensearch/neuralsearch/util/PluginFeatureFlags.java:23]: ack

Comment (heemin32) [reply src/main/java/org/opensearch/neuralsearch/util/PluginFeatureFlags.java]: I still think it is client's responsibility to pass correct key. In this case `transport.features.neural_search_hybrid_search_enabled`. 

There is no such case that `neural_search_hybrid_search_enabled` can be enabled by setting `neural_search_hybrid_search_enabled` as true but only by setting `transport.features.neural_search_hybrid_search_enabled` as true I believe?

Comment (martin-gaievski) [reply src/main/java/org/opensearch/neuralsearch/util/PluginFeatureFlags.java]: I see your point, it can be that flag is set in two places and as we read it with OR and one that set in plugin is always TRUE then we cannot set it to FALSE ever. I'll rework to only use one name or another, in fact it should be always with the prefix except the definition.

File: src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java [hunk 2]
```diff
@@ -50,9 +52,14 @@ public class NeuralSearch extends Plugin implements ActionPlugin, SearchPlugin,
      * Gates the functionality of hybrid search
      * Currently query phase searcher added with hybrid search will conflict with concurrent search in core.
      * Once that problem is resolved this feature flag can be removed.
+     * Key is the name string, value is key + transport feature specific prefix, prefix is added by core when we register feature
+     * We need to write and read by the value, key is only for definition
```
heemin32, line 56: I think this previous comment having a code link to core was good and would be nice to have it here. 
https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/plugins/PluginsService.java#L277

heemin32 (approved):

naveentatikonda (approved):

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
```diff
@@ -52,15 +50,9 @@ public class NeuralSearch extends Plugin implements ActionPlugin, SearchPlugin,
...
-     * Key is the name string, value is key + transport feature specific prefix,
-     * prefix is added by core when we register feature (https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/plugins/PluginsService.java#L277)
-     * We need to write and read by the value, key is only for definition
      */
     @VisibleForTesting
-    public static final Pair<String, String> NEURAL_SEARCH_HYBRID_SEARCH_DISABLED = Pair.of(
-        "neural_search_hybrid_search_disabled",
-        String.join(".", TransportSettings.FEATURE_PREFIX, "neural_search_hybrid_search_disabled")
-    );
+    public static final String NEURAL_SEARCH_HYBRID_SEARCH_ENABLED = "neural_search_hybrid_search_enabled";
```
navneet1v: The variable is already public we can remove @VisbleForTesting

src/test/java/org/opensearch/neuralsearch/plugin/NeuralSearchTests.java
```diff
@@ -42,15 +42,15 @@ public void testQueryPhaseSearcher() {
         assertNotNull(queryPhaseSearcher);
         assertTrue(queryPhaseSearcher.isEmpty());
 
-        System.setProperty(NEURAL_SEARCH_HYBRID_SEARCH_DISABLED.getValue(), "true");
+        System.setProperty(NEURAL_SEARCH_HYBRID_SEARCH_ENABLED, "true");
```
navneet1v, line 45: should we set this property in base class?

and why test are extending OpenSearchTestCase, there is should be a base Neural Search test class

navneet1v:
happy to see all checks succeeding. This is happening after a long time

src/test/java/org/opensearch/neuralsearch/plugin/NeuralSearchTests.java
martin-gaievski: I was thinking that for some unit tests we may want to disable it, that would be complex if it's set in base class. 

I'll make setting in a method and will call it in a setup phase for this test class.

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
martin-gaievski: I made it public only because it's required for tests otherwise private will be enough, my understanding is that @VisbleForTesting is required in this case to mark that it's not by mistake.

navneet1v: oh.. sorry my bad. I read it wrong then

src/test/java/org/opensearch/neuralsearch/plugin/NeuralSearchTests.java
navneet1v: sure

navneet1v:
@martin-gaievski tests are failing.

navneet1v (approved):

Issue #228, opened 2023-07-19
[FEATURE] Provide way of defining methods for score normalization and combination in scope of Hybrid search


martin-gaievski:
## Description

For Normalization and Score Combination feature, we need actual processing unit that will process scores collected on Query phase of Hybrid search. We need approach to define different techniques for score normalization and combination. 

## Solution

Solution we are proposing is to create new implementation of a [Search phase result processor](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/pipeline/SearchPhaseResultsProcessor.java). This Processor will be setup as part of search pipeline to be called between Query and Fetch phases. More details on such processors can be found in[ corresponding core PR](https://github.com/opensearch-project/OpenSearch/pull/7283)

Processor will support predefined set of techniques for normalization and combination. Exact techniques are defined using search pipeline API and then it must be referenced from ```_search``` call. We start from ```min-max``` for normalization and ```arithmetic mean``` for combination. 

Processor definition may look something like this:
```
{
    "description": "Post processor for hybrid search",
    "phase_results_processors": [
        {
            "normalization-processor": {
                "normalization": {
                    "technique": "MIN_MAX"
                },
                "combination": {
                    "technique": "ARITHMETIC_MEAN",
                    "parameters": {
                        "weights": [
                            0.4, 0.7
                        ]
                    }
                }
            }
        }
    ]
}
```

## Tasks

- [x] Implementation of a Search phase result processor
- [ ] Testing

## Reference Links
1. https://github.com/opensearch-project/neural-search/issues/126
2. https://github.com/opensearch-project/neural-search/issues/174
3. https://github.com/opensearch-project/neural-search/issues/193
4. https://github.com/opensearch-project/OpenSearch/pull/7283

austintlee:
I didn't realize this feature aspires to implement a generic hybrid search.  I was under the impression that it simply combines a BM25 search and a KNN search which is why I thought you'd always have two weights that add up to 1.0.

Don't the weights need to sum to 1?  It looks like in the current implementation, you assign a weight of 1.0 to sub-queries that are not matched to the weights specified in the query.  In other words, if you have 2 weights in the input and 4 sub-queries, the 3rd and 4th sub-queries seem to get a weight of 1.0?

navneet1v:
> Don't the weights need to sum to 1?

Yes the weights need to sum up to 1. We didn't add this check at start. This needs to be added. 

[...]

Issue #126, opened 2023-02-28
[RFC] High Level Approach and Design For Normalization and Score Combination


navneet1v:
## Introduction

This issue talks about the various high level directions which are being proposed for Score combination and Normalization techniques to improve Semantic Search/Neural Search Queries in OpenSearch(META ISSUE: https://github.com/opensearch-project/neural-search/issues/123). The proposals tries to use already created extensions in OpenSearch as much as possible. Also, we try to make sure that directions we are choosing are long term; provides different level of customizations for users to tweak the Semantics Search as per their needs. The document also proposes phased design and implementation plan which will help us to improve and add new feature with every phase.

For information how normalization improves the overall Quality of Results please refer this OpenSearch Blog for Science Benchmarks: https://opensearch.org/blog/semantic-science-benchmarks/

## Current Architecture

For simplicity, let's us consider a 3 node OpenSearch cluster, where we have 2 data nodes and 1 coordinator node. The data node stores the data and coordinator node helps in running the request. This is how OpenSearch works on a very high level.

1. Request lands on a Coordinator node and this node will rewrite the queries and send out the request to all the required shards as part of Query Phase.
2. The shards then use that query and pass it to Lucene(Index Search) along with all the relevant Collectors. Lucene then again rewrite the query and start performing the leaf queries on the segments sequentially. Its uses the different collectors like TopDocsCollector to collect the documents with Top Scores and return this result back to Coordinator node.
3. Once the query results arrive on the Coordinator node, we move to next phase which is Fetch phase. In the Fetch phase we have different sub-fetch phases that runs. Important one is Source Phase and Score phase. The Score phase will again send out the request to Data nodes to get the sources for the document ids and build the response to send back to the clients.

![NormalizationInNeuralSearch-Current HLD drawio (1)](https://user-images.githubusercontent.com/6162415/221723507-7c1c3587-d776-4ec1-a2e2-91bf476a1e5d.png)

## Requirements

The customer here is referenced as the OpenSearch customers, who wants to use OpenSearch and wants to run Semantics Search in their application.

1. As a customer, I want to combine different results coming from Transformers query(K-NN Query) and BM-25 based queries to obtain better search results. Github link: https://github.com/opensearch-project/k-NN/issues/717#issue-1529124743
2. As a customer I want to experiment with different ways to combine and rank the results in final search output. Github Link: https://github.com/opensearch-project/OpenSearch/issues/4557

[...]

PR #235, opened 2023-07-27
Adding weights param for combination technique

martin-gaievski:
### Description
Adding parameter "weights" for combination technique in normalization processor. Weights are use to multiply scores for each sub-query from the Hybrid search query. Weights are mapped to sub-query based on position (index), if there are less weights provided we use "1.0" for the rest of sub-query scores. If there are more weights provided we take first N and ignore the rest. 

Parameter is set as part of the search pipeline, example of such request:

```
{
    "phase_results_processors": [
        {
            "normalization-processor": {
                "normalization": {
                    "technique": "min_max"
                },
                "combination": {
                    "parameters": {
                        "weights": [
                            0.4, 0.7
                        ]
                    }
                }
            }
        }
    ]
}
```

### Issues Resolved

### Check List
- [X] All tests pass

skip

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
@@ -5,14 +5,13 @@
 
 package org.opensearch.neuralsearch.processor.combination;
 
-import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-
-import org.opensearch.OpenSearchParseException;
+import java.util.stream.Collectors;
 
 /**
  * Abstracts combination of scores based on arithmetic mean method
@@ -21,17 +20,23 @@ public class ArithmeticMeanScoreCombinationTechnique implements ScoreCombination
 
     public static final String TECHNIQUE_NAME = "arithmetic_mean";
     public static final String PARAM_NAME_WEIGHTS = "weights";
+    private static final Set<String> SUPPORTED_PARAMS = Set.of(PARAM_NAME_WEIGHTS);
     private static final Float ZERO_SCORE = 0.0f;
-    private final List<Double> weights;
+    private final List<Float> weights;
 
     public ArithmeticMeanScoreCombinationTechnique(final Map<String, Object> params) {
         validateParams(params);
+        weights = getWeights(params);
+    }
+
+    private List<Float> getWeights(final Map<String, Object> params) {
         if (Objects.isNull(params) || params.isEmpty()) {
-            weights = List.of();
-            return;
+            return List.of();
         }
         // get weights, we don't need to check for instance as it's done during validation
-        weights = (List<Double>) params.getOrDefault(PARAM_NAME_WEIGHTS, new ArrayList<>());
+        return ((List<Double>) params.getOrDefault(PARAM_NAME_WEIGHTS, List.of())).stream()
+            .map(Double::floatValue)
+            .collect(Collectors.toUnmodifiableList());
     }
 
     /**
@@ -44,13 +49,11 @@ public class ArithmeticMeanScoreCombinationTechnique implements ScoreCombination
     public float combine(final float[] scores) {
         float combinedScore = 0.0f;
         int count = 0;
-        for (int i = 0; i < scores.length; i++) {
-            float score = scores[i];
+        for (int indexOfSubQuery = 0; indexOfSubQuery < scores.length; indexOfSubQuery++) {
+            float score = scores[indexOfSubQuery];
             if (score >= 0.0) {
-                // apply weight for this sub-query if it's set for particular sub-query
-                if (i < weights.size()) {
-                    score = (float) (score * weights.get(i));
-                }
+                float weight = getWeightForSubQuery(indexOfSubQuery);
+                score = score * weight;
                 combinedScore += score;
                 count++;
             }
@@ -66,20 +69,36 @@ public class ArithmeticMeanScoreCombinationTechnique implements ScoreCombination
             return;
         }
         // check if only supported params are passed
-        Set<String> supportedParams = Set.of(PARAM_NAME_WEIGHTS);
         Optional<String> optionalNotSupportedParam = params.keySet()
             .stream()
-            .filter(paramName -> !supportedParams.contains(paramName))
+            .filter(paramName -> !SUPPORTED_PARAMS.contains(paramName))
             .findFirst();
         if (optionalNotSupportedParam.isPresent()) {
-            throw new OpenSearchParseException("provided parameter for combination technique is not supported");
+            throw new IllegalArgumentException(
+                String.format(
+                    Locale.ROOT,
+                    "provided parameter for combination technique is not supported. supported parameters are [%s]",
+                    SUPPORTED_PARAMS.stream().collect(Collectors.joining(","))
+                )
+            );
         }
 
         // check param types
         if (params.keySet().stream().anyMatch(PARAM_NAME_WEIGHTS::equalsIgnoreCase)) {
             if (!(params.get(PARAM_NAME_WEIGHTS) instanceof List)) {
-                throw new OpenSearchParseException("parameter {} must be a collection of numbers", PARAM_NAME_WEIGHTS);
+                throw new IllegalArgumentException(
+                    String.format(Locale.ROOT, "parameter [%s] must be a collection of numbers", PARAM_NAME_WEIGHTS)
+                );
             }
         }
+    }
+
+    /**
+     * Get weight for sub-query based on its index in the hybrid search query. Use user provided weight or 1.0 otherwise
+     * @param indexOfSubQuery 0-based index of sub-query in the Hybrid Search query
+     * @return weight for sub-query, use one that is set in processor/pipeline definition or 1.0 as default
+     */
+    private float getWeightForSubQuery(int indexOfSubQuery) {
+        return indexOfSubQuery < weights.size() ? weights.get(indexOfSubQuery).floatValue() : 1.0f;
     }
 }
--- a/src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
@@ -5,16 +5,34 @@
 
 package org.opensearch.neuralsearch.processor.combination;
 
-import lombok.NoArgsConstructor;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+
+import org.opensearch.OpenSearchParseException;
 
 /**
  * Abstracts combination of scores based on arithmetic mean method
  */
-@NoArgsConstructor
 public class ArithmeticMeanScoreCombinationTechnique implements ScoreCombinationTechnique {
 
     public static final String TECHNIQUE_NAME = "arithmetic_mean";
+    public static final String PARAM_NAME_WEIGHTS = "weights";
     private static final Float ZERO_SCORE = 0.0f;
+    private final List<Double> weights;
+
+    public ArithmeticMeanScoreCombinationTechnique(final Map<String, Object> params) {
+        validateParams(params);
+        if (Objects.isNull(params) || params.isEmpty()) {
+            weights = List.of();
+            return;
+        }
+        // get weights, we don't need to check for instance as it's done during validation
+        weights = (List<Double>) params.getOrDefault(PARAM_NAME_WEIGHTS, new ArrayList<>());
+    }
 
     /**
      * Arithmetic mean method for combining scores.
@@ -26,8 +44,13 @@ public class ArithmeticMeanScoreCombinationTechnique implements ScoreCombination
     public float combine(final float[] scores) {
         float combinedScore = 0.0f;
         int count = 0;
-        for (float score : scores) {
+        for (int i = 0; i < scores.length; i++) {
+            float score = scores[i];
             if (score >= 0.0) {
+                // apply weight for this sub-query if it's set for particular sub-query
+                if (i < weights.size()) {
+                    score = (float) (score * weights.get(i));
+                }
                 combinedScore += score;
                 count++;
             }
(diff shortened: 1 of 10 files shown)


src/main/java/org/opensearch/neuralsearch/processor/factory/NormalizationProcessorFactory.java
```diff
@@ -53,9 +55,12 @@ public SearchPhaseResultsProcessor create(
         Map<String, Object> combinationClause = readOptionalMap(NormalizationProcessor.TYPE, tag, config, COMBINATION_CLAUSE);
 
         ScoreCombinationTechnique scoreCombinationTechnique = ScoreCombinationFactory.DEFAULT_METHOD;
+        Map<String, Object> combinationParams;
         if (Objects.nonNull(combinationClause)) {
-            String combinationTechnique = (String) combinationClause.getOrDefault(TECHNIQUE, "");
-            scoreCombinationTechnique = scoreCombinationFactory.createCombination(combinationTechnique);
+            String combinationTechnique = readOptionalStringProperty(NormalizationProcessor.TYPE, tag, combinationClause, TECHNIQUE);
```
heemin32, line 60: combinationTechnique is not optional but we use `readOptionalStringProperty`?

src/main/java/org/opensearch/neuralsearch/processor/factory/NormalizationProcessorFactory.java
```diff
@@ -53,9 +55,12 @@ public SearchPhaseResultsProcessor create(
...
+            // check for optional combination params
+            combinationParams = readOptionalMap(NormalizationProcessor.TYPE, tag, combinationClause, PARAMETERS);
```
heemin32, line 62:
```suggestion
            Map<String, Object> combinationParams = readOptionalMap(NormalizationProcessor.TYPE, tag, combinationClause, PARAMETERS);
```

src/main/java/org/opensearch/neuralsearch/processor/factory/NormalizationProcessorFactory.java
```diff
@@ -53,9 +55,12 @@ public SearchPhaseResultsProcessor create(
         Map<String, Object> combinationClause = readOptionalMap(NormalizationProcessor.TYPE, tag, config, COMBINATION_CLAUSE);
 
         ScoreCombinationTechnique scoreCombinationTechnique = ScoreCombinationFactory.DEFAULT_METHOD;
+        Map<String, Object> combinationParams;
```
heemin32, line 58: Move inside `if` block
```suggestion

```

src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
```diff
@@ -5,16 +5,34 @@
...
+    private final List<Double> weights;
+
+    public ArithmeticMeanScoreCombinationTechnique(final Map<String, Object> params) {
+        validateParams(params);
+        if (Objects.isNull(params) || params.isEmpty()) {
+            weights = List.of();
+            return;
+        }
+        // get weights, we don't need to check for instance as it's done during validation
+        weights = (List<Double>) params.getOrDefault(PARAM_NAME_WEIGHTS, new ArrayList<>());
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
```diff
@@ -5,16 +5,34 @@
...
 public class ArithmeticMeanScoreCombinationTechnique implements ScoreCombinationTechnique {
 
     public static final String TECHNIQUE_NAME = "arithmetic_mean";
+    public static final String PARAM_NAME_WEIGHTS = "weights";
     private static final Float ZERO_SCORE = 0.0f;
+    private final List<Double> weights;
+
+    public ArithmeticMeanScoreCombinationTechnique(final Map<String, Object> params) {
+        validateParams(params);
+        if (Objects.isNull(params) || params.isEmpty()) {
```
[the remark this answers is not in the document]
martin-gaievski: I think after refactoring you suggested this will be changed

src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
```diff
@@ -5,16 +5,34 @@
...
 /**
  * Abstracts combination of scores based on arithmetic mean method
  */
-@NoArgsConstructor
 public class ArithmeticMeanScoreCombinationTechnique implements ScoreCombinationTechnique {
 
     public static final String TECHNIQUE_NAME = "arithmetic_mean";
+    public static final String PARAM_NAME_WEIGHTS = "weights";
     private static final Float ZERO_SCORE = 0.0f;
+    private final List<Double> weights;
```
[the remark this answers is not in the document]
martin-gaievski: our parser does doubles, those are not vectors but just factors for multiplication, so double or float doesn't really matter. Otherwise we need to do one more round of conversion, not sure if that's necessary

src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
```diff
@@ -26,8 +44,13 @@ public class ArithmeticMeanScoreCombinationTechnique implements ScoreCombination
     public float combine(final float[] scores) {
         float combinedScore = 0.0f;
         int count = 0;
-        for (float score : scores) {
+        for (int i = 0; i < scores.length; i++) {
+            float score = scores[i];
             if (score >= 0.0) {
+                // apply weight for this sub-query if it's set for particular sub-query
+                if (i < weights.size()) {
```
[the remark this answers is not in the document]
martin-gaievski: like having weight for all cases, just it will be 1.0 if it's not explicitly set by user? I like this form, let me rework

src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
```diff
@@ -37,4 +60,26 @@ public float combine(final float[] scores) {
...
+            return;
+        }
+        // check if only supported params are passed
+        Set<String> supportedParams = Set.of(PARAM_NAME_WEIGHTS);
+        Optional<String> optionalNotSupportedParam = params.keySet()
+            .stream()
+            .filter(paramName -> !supportedParams.contains(paramName))
+            .findFirst();
+        if (optionalNotSupportedParam.isPresent()) {
+            throw new OpenSearchParseException("provided parameter for combination technique is not supported");
```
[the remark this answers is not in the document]
martin-gaievski: ack, I used parse as this is suppose to go up to the factory, and I didn't want to put that knowledge of params into factory. We can catch illigalargument and throw parseexception in the factory

src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
```diff
@@ -37,4 +60,26 @@ public float combine(final float[] scores) {
...
+        }
+
+        // check param types
+        if (params.keySet().stream().anyMatch(PARAM_NAME_WEIGHTS::equalsIgnoreCase)) {
+            if (!(params.get(PARAM_NAME_WEIGHTS) instanceof List)) {
+                throw new OpenSearchParseException("parameter {} must be a collection of numbers", PARAM_NAME_WEIGHTS);
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombinationFactory.java
```diff
@@ -28,7 +29,17 @@ public class ScoreCombinationFactory {
      * @return instance of ScoreCombinationTechnique for technique name
      */
     public ScoreCombinationTechnique createCombination(final String technique) {
+        return createCombination(technique, Map.of());
+    }
+
+    /**
+     * Get score combination method by technique name
+     * @param technique name of technique
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/factory/NormalizationProcessorFactory.java
martin-gaievski: it's optional, user may not set it and we need a default in this case, same for normalization

src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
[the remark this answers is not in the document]
heemin32: Please let me know if I am wrong. I think we only use ParseException when customer provided document cannot be parsed. In this case, it is API parameter which is invalid not document to be indexed.

src/main/java/org/opensearch/neuralsearch/processor/factory/NormalizationProcessorFactory.java
heemin32: I think `combinationClause` is optional but not `combinationTechnique`
`scoreCombinationFactory.createCombination` throws an exception when `combinationTechnique` is not set.

src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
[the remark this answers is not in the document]
heemin32: Isn't this the place where we get weight? Here can we just cast it to `(List<Float>)`?
```
weights = (List<Double>) params.getOrDefault(PARAM_NAME_WEIGHTS, new ArrayList<>());
```

src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
```diff
@@ -26,8 +44,13 @@ public class ArithmeticMeanScoreCombinationTechnique implements ScoreCombination
...
-        for (float score : scores) {
+        for (int i = 0; i < scores.length; i++) {
+            float score = scores[i];
             if (score >= 0.0) {
+                // apply weight for this sub-query if it's set for particular sub-query
+                if (i < weights.size()) {
+                    score = (float) (score * weights.get(i));
+                }
                 combinedScore += score;
                 count++;
```
HenryL27, line 55: Maybe `count += weight`? I think weighted arithmetic mean wants to divide by total weight of the scores, not count. Otherwise you throw off your scaling, e.g. I have weights [300, 650, 50]. Doesn't matter all that much but I guess it does raise the question of what happens if the provided weights are way out of scale with the default weight (1.0)

navneet1v: @HenryL27 it needs to divide by the count. "mean" always mean divide by the number of entities you have. 

weights: here is a representation to say how much importance we want to give for a score of the document coming from a specific query. which in turn means how important the results of the query belong in the final result set.

Please correct me if there is another definition of arithmetic mean

navneet1v: @martin-gaievski https://en.wikipedia.org/wiki/Weighted_arithmetic_mean

src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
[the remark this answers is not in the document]
martin-gaievski: cast wouldn't work. we can only convert it

src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
martin-gaievski: nice, let me correct implementation, it should be divided by sum of weights as @HenryL27 mentioned

src/main/java/org/opensearch/neuralsearch/processor/factory/NormalizationProcessorFactory.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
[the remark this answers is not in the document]
martin-gaievski: You're correct, I checked other code in neural-search, we can throw IllegalArgument exceptin directly without wrapping it into ParsingException. I'll change that

src/main/java/org/opensearch/neuralsearch/processor/factory/NormalizationProcessorFactory.java
martin-gaievski: I see, that's true we assign default technique in case the whole "clause" is missing. Let me add same for case when technique name is missing

src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
```diff
@@ -5,16 +5,39 @@
...
+        validateParams(params);
+        weights = getWeights(params);
+    }
+
+    private List<Float> getWeights(final Map<String, Object> params) {
+        if (Objects.isNull(params) || params.isEmpty()) {
+            return List.of();
+        }
+        // get weights, we don't need to check for instance as it's done during validation
+        return ((List<Double>) params.getOrDefault(PARAM_NAME_WEIGHTS, List.of())).stream()
```
heemin32, line 37: This does not work?
```suggestion
        return ((List<Float>) params.getOrDefault(PARAM_NAME_WEIGHTS, List.of())).stream()
```

src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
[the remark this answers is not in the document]
martin-gaievski: I added explicit conversion double -> float

src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
martin-gaievski: nope, this gives classcastexception

heemin32: Curious how (List<Float>) is throwing classcastexception but not (List<Double>)

martin-gaievski: Problem is on list type cast, here is what is gives if I put List<Float>:
```
java.lang.ClassCastException: class java.lang.Double cannot be cast to class java.lang.Float (java.lang.Double and java.lang.Float are in module java.base of loader 'bootstrap')
```

src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
martin-gaievski: Thank you for pointing this out, we implemented "normalized weighted linear combination", I will change implementation accordingly

heemin32 (approved):

junqiu-lei (approved):

PR #236, opened 2023-07-28
Adding L2 norm technique

martin-gaievski:
### Description
Adding L2 normalization technique for normalization processor. Essentially L2 norm is "score" divided by square root of sum of score squares (more details [here](https://machinelearningmastery.com/vector-norms-machine-learning/)). Example of pipeline with processor config:

```
{
    "phase_results_processors": [
        {
            "normalization-processor": {
                "normalization": {
                    "technique": "l2"
                }
            }
        }
    ]
}
```

### Issues Resolved

### Check List
    - [X] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/combination/HarmonicMeanScoreCombinationTechnique.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/combination/HarmonicMeanScoreCombinationTechnique.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.neuralsearch.processor.combination;
+
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Abstracts combination of scores based on arithmetic mean method
+ */
+public class HarmonicMeanScoreCombinationTechnique implements ScoreCombinationTechnique {
+
+    public static final String TECHNIQUE_NAME = "arithmetic_mean";
+    public static final String PARAM_NAME_WEIGHTS = "weights";
+    private static final Set<String> SUPPORTED_PARAMS = Set.of(PARAM_NAME_WEIGHTS);
+    private static final Float ZERO_SCORE = 0.0f;
+    private final List<Float> weights;
+
+    public HarmonicMeanScoreCombinationTechnique(final Map<String, Object> params) {
+        validateParams(params);
+        weights = getWeights(params);
+    }
+
+    private List<Float> getWeights(final Map<String, Object> params) {
+        if (Objects.isNull(params) || params.isEmpty()) {
+            return List.of();
+        }
+        // get weights, we don't need to check for instance as it's done during validation
+        return ((List<Double>) params.getOrDefault(PARAM_NAME_WEIGHTS, List.of())).stream()
+            .map(Double::floatValue)
+            .collect(Collectors.toUnmodifiableList());
+    }
+
+    /**
+     * Arithmetic mean method for combining scores.
+     * score = (weight1*score1 + weight2*score2 +...+ weightN*scoreN)/(weight1 + weight2 + ... + weightN)
+     *
+     * Zero (0.0) scores are excluded from number of scores N
+     */
+    @Override
+    public float combine(final float[] scores) {
+        float combinedScore = 0.0f;
+        float weights = 0;
+        for (int indexOfSubQuery = 0; indexOfSubQuery < scores.length; indexOfSubQuery++) {
+            float score = scores[indexOfSubQuery];
+            if (score >= 0.0) {
+                float weight = getWeightForSubQuery(indexOfSubQuery);
+                score = score * weight;
+                combinedScore += score;
+                weights += weight;
+            }
+        }
+        if (weights == 0.0f) {
+            return ZERO_SCORE;
+        }
+        return combinedScore / weights;
+    }
+
+    private void validateParams(final Map<String, Object> params) {
+        if (Objects.isNull(params) || params.isEmpty()) {
+            return;
+        }
+        // check if only supported params are passed
+        Optional<String> optionalNotSupportedParam = params.keySet()
+            .stream()
+            .filter(paramName -> !SUPPORTED_PARAMS.contains(paramName))
+            .findFirst();
+        if (optionalNotSupportedParam.isPresent()) {
+            throw new IllegalArgumentException(
+                String.format(
+                    Locale.ROOT,
+                    "provided parameter for combination technique is not supported. supported parameters are [%s]",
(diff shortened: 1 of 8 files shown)


src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * L2 normalization method.
+     * n_score_i = score_i/sqrt(score1^2 + score2^2 + ... + scoren^2)
+     * Main algorithm steps:
+     * - calculate sum of squares of all scores
+     * - iterate over each result and update score as per formula above where "score" is raw score returned by Hybrid query
+     */
+    @Override
+    public void normalize(final List<CompoundTopDocs> queryTopDocs) {
+        int numOfSubqueries = queryTopDocs.stream()
```
heemin32, line 33: Can we move this inside `getL2Norm`?

src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
```diff
@@ -0,0 +1,80 @@
...
+            .filter(Objects::nonNull)
+            .filter(topDocs -> topDocs.getCompoundTopDocs().size() > 0)
+            .findAny()
+            .get()
+            .getCompoundTopDocs()
+            .size();
+        // get l2 norms for each sub-query
+        float[] normsPerSubquery = getL2Norm(queryTopDocs, numOfSubqueries);
```
heemin32, line 41: Can we switch to List<Float>?

src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
```diff
@@ -0,0 +1,80 @@
...
+     * Main algorithm steps:
+     * - calculate sum of squares of all scores
+     * - iterate over each result and update score as per formula above where "score" is raw score returned by Hybrid query
+     */
+    @Override
+    public void normalize(final List<CompoundTopDocs> queryTopDocs) {
+        int numOfSubqueries = queryTopDocs.stream()
+            .filter(Objects::nonNull)
+            .filter(topDocs -> topDocs.getCompoundTopDocs().size() > 0)
+            .findAny()
```
heemin32, line 36: Shouldn't we find the max value? Or `getCompoundTopsDocs()` is same for all `queryTopDocs`?

src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private float[] getL2Norm(final List<CompoundTopDocs> queryTopDocs, final int numOfSubqueries) {
+        float[] l2Norms = new float[numOfSubqueries];
+        for (CompoundTopDocs compoundQueryTopDocs : queryTopDocs) {
+            if (Objects.isNull(compoundQueryTopDocs)) {
+                continue;
+            }
+            List<TopDocs> topDocsPerSubQuery = compoundQueryTopDocs.getCompoundTopDocs();
+            IntStream.range(0, topDocsPerSubQuery.size()).forEach(index -> {
```
heemin32, line 65: Better to use `for` statement as I don't see much benefit of using stream this way..

src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
martin-gaievski: yes, it's either empty is shard does not have any results for all of sub-queries, or it has results for all the sub-queries. In edge case of let's say shard has results only for one sub-query, then for all the rest there will be total_hits = 0.

src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
martin-gaievski: it's more a style, my understanding we should use stream everywhere other than if array is required

src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
heemin32: Cool. Would be nice to have a comment.

src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
heemin32: We use stream when it is less error prone thanks to definitive way of writing a code. I don't think this is true for this case because we are accessing topDocsPerSubQuery by its index anyway and not difference with `for` statement.

martin-gaievski: that makes sense, let me change to the for loop

src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
martin-gaievski: ack

heemin32 (approved):

navneet1v (approved):

src/main/java/org/opensearch/neuralsearch/processor/combination/HarmonicMeanScoreCombinationTechnique.java
```diff
@@ -0,0 +1,104 @@
...
+                score = score * weight;
+                combinedScore += score;
+                weights += weight;
+            }
+        }
+        if (weights == 0.0f) {
+            return ZERO_SCORE;
+        }
+        return combinedScore / weights;
+    }
```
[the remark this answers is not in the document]
martin-gaievski: Thanks @HenryL27 , that's what I'm working now and it will be a new PR for harmonic mean and weighted geometric combination. Probably checked in this class incidentally in a half-ready form.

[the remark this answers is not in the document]
HenryL27: ah, no worries. Thanks!

PR #239, opened 2023-08-01
Add geometric mean normalization for scores

martin-gaievski:
### Description
Adding geometric mean technique, that is a generalization of the mean that is based on product and N-th root of N values (more details [here](https://en.wikipedia.org/wiki/Weighted_geometric_mean)). Weights are supported similarly what it's done in arithmetic mean. Example of pipeline with processor config:

```
{
    "phase_results_processors": [
        {
            "normalization-processor": {}, 
            "combination": {
                 "technique": "geometric_mean",
                  "parameters": {
                        "weights": [
                            0.4, 0.7
                        ]
                    }
             }
        }
    ]
}
```

In addition to main changes there are some refactoring in integ tests. I have to put it to this PR because with few new tests added for geometric mean auto redeploy feature started acting more aggressively and tests became flaky. 
 
 - move from global model id to reading it at the beginning of each test case. this is required because model can be redeployed by the [auto redeploy feature of ml-commons](https://github.com/opensearch-project/ml-commons/pull/852)
 - split on large class into three smaller ones based on area of responsibility 
 - cleanup outdated tests

### Issues Resolved

### Check List
    - [X] All tests pass

### DIFF
--- a/src/test/java/org/opensearch/neuralsearch/processor/ScoreNormalizationIT.java
+++ b/src/test/java/org/opensearch/neuralsearch/processor/ScoreNormalizationIT.java
@@ -1,0 +1,295 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.neuralsearch.processor;
+
+import static org.opensearch.neuralsearch.TestUtils.assertHybridSearchResults;
+import static org.opensearch.neuralsearch.TestUtils.createRandomVector;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Map;
+
+import lombok.SneakyThrows;
+
+import org.junit.After;
+import org.junit.Before;
+import org.opensearch.index.query.QueryBuilders;
+import org.opensearch.knn.index.SpaceType;
+import org.opensearch.neuralsearch.common.BaseNeuralSearchIT;
+import org.opensearch.neuralsearch.query.HybridQueryBuilder;
+import org.opensearch.neuralsearch.query.NeuralQueryBuilder;
+
+import com.google.common.primitives.Floats;
+
+public class ScoreNormalizationIT extends BaseNeuralSearchIT {
+    private static final String TEST_MULTI_DOC_INDEX_ONE_SHARD_NAME = "test-neural-multi-doc-one-shard-index";
+    private static final String TEST_QUERY_TEXT3 = "hello";
+    private static final String TEST_DOC_TEXT1 = "Hello world";
+    private static final String TEST_DOC_TEXT2 = "Hi to this place";
+    private static final String TEST_DOC_TEXT3 = "We would like to welcome everyone";
+    private static final String TEST_DOC_TEXT4 = "Hello, I'm glad to you see you pal";
+    private static final String TEST_KNN_VECTOR_FIELD_NAME_1 = "test-knn-vector-1";
+    private static final String TEST_TEXT_FIELD_NAME_1 = "test-text-field-1";
+    private static final int TEST_DIMENSION = 768;
+    private static final SpaceType TEST_SPACE_TYPE = SpaceType.L2;
+    private static final String SEARCH_PIPELINE = "phase-results-pipeline";
+    private final float[] testVector1 = createRandomVector(TEST_DIMENSION);
+    private final float[] testVector2 = createRandomVector(TEST_DIMENSION);
+    private final float[] testVector3 = createRandomVector(TEST_DIMENSION);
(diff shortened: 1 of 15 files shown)


src/test/java/org/opensearch/neuralsearch/processor/combination/GeometricMeanScoreCombinationTechniqueTests.java
```diff
@@ -0,0 +1,62 @@
...
+    public void testLogic_whenNotAllScoresAndWeightsPresent_thenCorrectScores() {
+        List<Double> weights = List.of(0.9, 0.2, 0.7);
+        ScoreCombinationTechnique technique = new GeometricMeanScoreCombinationTechnique(
+            Map.of(PARAM_NAME_WEIGHTS, weights),
+            scoreCombinationUtil
+        );
+        testLogic_whenNotAllScoresAndWeightsPresent_thenCorrectScores(technique, weights);
+    }
+
+    private float geometricMean(List<Float> scores, List<Double> weights) {
```
heemin32, line 49: Hope this comment is not too late. I should have paid attention to test code before.

Here, we are coping calculation logic from the production code which might not be able to catch a bug in the logic because the test code and production code is same.
Instead, we should calculate the end result manually and use it to compare withe the value returned by method. Same for other tests.

src/test/java/org/opensearch/neuralsearch/processor/ScoreNormalizationIT.java
```diff
@@ -0,0 +1,295 @@
...
+import org.junit.Before;
+import org.opensearch.index.query.QueryBuilders;
+import org.opensearch.knn.index.SpaceType;
+import org.opensearch.neuralsearch.common.BaseNeuralSearchIT;
+import org.opensearch.neuralsearch.query.HybridQueryBuilder;
+import org.opensearch.neuralsearch.query.NeuralQueryBuilder;
+
+import com.google.common.primitives.Floats;
+
+public class ScoreNormalizationIT extends BaseNeuralSearchIT {
```
heemin32, line 28: No need to change but just want to share my thought. Integration test is expensive. Therefore, it would be better to write IT just for the base functionality and leave all the other cases to unit test.
In this case, I would write IT using either L2 for min-max only as except the calculation logic, all others things are same between the two. The calculation logic can be covered in unit test.

src/test/java/org/opensearch/neuralsearch/processor/combination/GeometricMeanScoreCombinationTechniqueTests.java
martin-gaievski: I think we have two cases here: for the dimple dataset size, something like 2 sub-queries, 4-5 scores each we can calculate that manually and use it as expected value in assertion. For the larger size data manual calculation may not be feasible, so we need to calculate that ground truth value of expected values. Something to what we're doing in [knn repo tests](https://github.com/opensearch-project/k-NN/blob/main/src/testFixtures/java/org/opensearch/knn/TestUtils.java#L196). Let me rework into those two approaches.

heemin32: You can use the value returned by the method to get the initial value and hard code the value here so that if the logic changes in the future, we can catch a mistake.

src/test/java/org/opensearch/neuralsearch/processor/combination/GeometricMeanScoreCombinationTechniqueTests.java
```diff
@@ -0,0 +1,98 @@
...
+            scoreCombinationUtil
+        );
+        testRandomValues_whenNotAllScoresAndWeightsPresent_thenCorrectScores(technique, weights);
+    }
+
+    /**
+     * Verify score correctness by using alternative formula for geometric mean as n-th root of product of weighted scores,
+     * more details in here https://en.wikipedia.org/wiki/Weighted_geometric_mean
+     */
+    private float geometricMean(List<Float> scores, List<Double> weights) {
```
heemin32, line 83: I still have a doubt on the effectiveness of this test code. l believe we don’t need test code based on random number. I would like to here other opinions though.

heemin32 (approved):

src/test/java/org/opensearch/neuralsearch/processor/combination/GeometricMeanScoreCombinationTechniqueTests.java
navneet1v: I don't have any concern as long long as test is able to fail if there are changes in formula.

naveentatikonda (approved):

Issue #123, opened 2023-02-23
[META] Score Combination and Normalization for Semantics Search.  Score Normalization for k-NN and BM25


vamshin:
### Is your feature request related to a problem?
BM25 works well in exact match use cases and k-NN score works well in understanding context and getting relevant documents. It is important to get benefits from both of these relevancy mechanisms and one could achieve by combining these scores. One caveat is scores are on different scales and hence some kind of normalization is required.  

### Older Issues and Discussions:
1. https://github.com/opensearch-project/k-NN/issues/717
2. https://github.com/opensearch-project/OpenSearch/issues/4557
3. Science Benchmarks: https://opensearch.org/blog/semantic-science-benchmarks/

### Tasks
High Level Tasks:
- [X] Create RFC for the Feature Request. https://github.com/opensearch-project/neural-search/issues/126
- [X] Build POC for the Feature.
- [X] Low Level Design of various components.
    - [X] API Design https://github.com/opensearch-project/neural-search/issues/244
    - [X] Search Phase Results Processor: https://github.com/opensearch-project/neural-search/issues/152
    - [X] Different Normalization and Score combination Methods: https://github.com/opensearch-project/neural-search/issues/228
    - [X] New Doc Collector For Query Clause: https://github.com/opensearch-project/neural-search/issues/228
    - [X] LLD for New Query Phase Searcher: https://github.com/opensearch-project/neural-search/issues/228
- [x] Validate Accuracy of the solution.
- [ ] Perf Testing and release tasks.
- [x] Documentation issue for Normalization and Score combination feature: https://github.com/opensearch-project/documentation-website/issues/4351

### Community Requests
1. https://forum.opensearch.org/t/normalisation-in-hybrid-search/12996

navneet1v:
RFC created with HLD: https://github.com/opensearch-project/neural-search/issues/126

martin-gaievski:
RFC with design for Query: https://github.com/opensearch-project/neural-search/issues/174
Created new Feature issue for Query: https://github.com/opensearch-project/neural-search/issues/175

DarshitChanpura:
@vamshin Should this issue be moved to 2.11?

vamshin:
@DarshitChanpura we are releasing this feature for 2.10

navneet1v:
@martin-gaievski can we add the perf results on this issue and then resolve this github issue.

martin-gaievski:
Summary for performance benchmark

[...]

PR #259, opened 2023-08-22
Changed format for hybrid query results to a single list of scores with delimiter

martin-gaievski:
### Description
Addressing findings from testing of Hybrid Search on larger datasets (https://github.com/beir-cellar/beir#beers-available-datasets, mainly trec-covid and nfcorpus)

What's included in this change:
- fix for incorrect scores in case of multiple nodes

Case of multiple nodes has problem of serialization/deserialization of the custom class when data nodes are sending their results to coordinator. With current implementation of [serialization](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/lucene/Lucene.java#L430) and [deserialization](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/lucene/Lucene.java#L312) in core OpenSearch only single array of ScoreDocs will be transferred between nodes.

Change the approach of storing results, now we use existing TopDocs class. Scores will be stored in a single list and each sub-query results will end with a special delimiter score object (see below for exact format). CompoundTopDocs used as DTO only to keep results of multiple sub-queries between normalization processor steps.

- fix for runtime errors in case of a single shard

Fixing side-effect of approach with a single list and delimiters. Due to implementation of [optimization logic in core](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/SearchService.java#L615-L617) in such case processors will be called after the actual Fetch phase. For us that means we need to update results of fetch phase, not only query results.

- change default value for min score from 0.001 to 0.0. This is after consulting with science team, same approach has been used in their experiments. 

Example of format used for score list with delimiters:
```
     *  doc_id | magic_number_1    //start/stop
     *  doc_id | magic_number_2   //delimiter for sub-query 1
     *  ...
     *  doc_id | magic_number_2   //delimiter for sub-query 2
     *  ...
     *  doc_id | magic_number_2   //delimiter for sub-query 3
     *  ...
     *  doc_id | magic_number_1  //start/stop
```

### Issues Resolved
https://github.com/opensearch-project/neural-search/issues/123

### Check List
- [X] All tests pass


navneet1v:
@martin-gaievski can you run the spotlessApply as the build is failing.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -67,22 +75,71 @@ public void execute(
...
+        // 3. update original scores to normalized and combined values
+        // 4. order scores based on normalized and combined values
+        SearchHits searchHits = fetchSearchResult.hits();
+
+        // create map of docId to index of search hits
+        Map<Integer, SearchHit> docIdToSearchHit = new HashMap<>();
+        for (SearchHit searchHit : searchHits) {
+            docIdToSearchHit.put(searchHit.docId(), searchHit);
+        }
+        QuerySearchResult querySearchResult = querySearchResults.get(0);
```
heemin32, line 128: Why do we use only the first item in querySearchResults?

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -67,22 +75,71 @@ public void execute(
...
+        TopDocs topDocs = querySearchResult.topDocs().topDocs;
+        // iterate over the normalized/combined scores, that solves (1), (2) and (3)
```
heemin32, line 130: I think (2) is handled when we convert `fetchSearchResult.hits()` to `docIdToSearchHit` but not in this iteration?

martin-gaievski:
> @martin-gaievski can you run the spotlessApply as the build is failing.

done, not spotless checks are passing, thank you @navneet1v

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -110,29 +115,84 @@ private void setTopDocsInQueryResult(
     ) {
         final List<TopDocs> topDocs = collector.topDocs();
         final float maxScore = getMaxScore(topDocs);
-        final TopDocs newTopDocs = new CompoundTopDocs(getTotalHits(searchContext, topDocs), topDocs);
+        boolean isSingleShard = searchContext.numberOfShards() == 1;
```
heemin32, line 118: No final?
```suggestion
        final boolean isSingleShard = searchContext.numberOfShards() == 1;
```

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -110,29 +115,84 @@ private void setTopDocsInQueryResult(
...
+        final TopDocs newTopDocs = getNewTopDocs(getTotalHits(searchContext, topDocs, isSingleShard), topDocs);
         final TopDocsAndMaxScore topDocsAndMaxScore = new TopDocsAndMaxScore(newTopDocs, maxScore);
         queryResult.topDocs(topDocsAndMaxScore, getSortValueFormats(searchContext.sort()));
     }
 
-    private TotalHits getTotalHits(final SearchContext searchContext, final List<TopDocs> topDocs) {
+    TopDocs getNewTopDocs(final TotalHits totalHits, final List<TopDocs> topDocs) {
```
heemin32, line 124:
```suggestion
    private TopDocs getNewTopDocs(final TotalHits totalHits, final List<TopDocs> topDocs) {
```

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -110,29 +115,84 @@ private void setTopDocsInQueryResult(
...
+                .map(topDoc -> topDoc.scoreDocs)
+                .filter(scoreDoc -> scoreDoc.length > 0)
+                .map(scoreDoc -> scoreDoc[0].doc)
+                .findFirst()
+                .orElse(-1);
+            if (delimiterDocId == -1) {
+                return new TopDocs(totalHits, scoreDocs);
+            }
+            // format scores using following template:
+            // doc_id | magic_number_1
```
heemin32, line 143: Do we really need start and end mark?
Can we just do like
```
...
doc_id | delimiter
...
doc_id | delimiter
...
```

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -110,29 +115,84 @@ private void setTopDocsInQueryResult(
...
             : TotalHits.Relation.EQUAL_TO;
-        if (topDocs == null || topDocs.size() == 0) {
+        if (topDocs == null || topDocs.isEmpty()) {
             return new TotalHits(0, relation);
         }
         long maxTotalHits = topDocs.get(0).totalHits.value;
+        int totalSize = 0;
         for (TopDocs topDoc : topDocs) {
             maxTotalHits = Math.max(maxTotalHits, topDoc.totalHits.value);
+            if (isSingleShard) {
```
heemin32, line 179: I guess this line is not needed?

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -110,29 +115,84 @@ private void setTopDocsInQueryResult(
...
+                totalSize += topDoc.totalHits.value + 1;
+            }
         }
+        // add 1 qty per each sub-query and + 2 for start and stop delimiters
+        totalSize += 2;
+        if (isSingleShard) {
+            // for single shard we need to update total size as this is how many docs are fetched in Fetch phase
```
heemin32, line 186: Could be due to a lack of context, but how do we handle total size for multi shard?
In my understanding, we update totalSize because fetch is already done for single shard.
But for multi shard, still how do we get correct totalSize including start/stop mark and delimiter count?

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
```diff
@@ -0,0 +1,107 @@
...
+public class CompoundTopDocs {
+
+    @Getter
+    @Setter
+    private TotalHits totalHits;
+    @Getter
+    private final List<TopDocs> compoundTopDocs;
+    @Getter
+    @Setter
+    private ScoreDoc[] scoreDocs;
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
```diff
@@ -0,0 +1,107 @@
...
+
+    public CompoundTopDocs(final TotalHits totalHits, final List<TopDocs> compoundTopDocs) {
+        this.totalHits = totalHits;
+        this.compoundTopDocs = compoundTopDocs;
+        scoreDocs = cloneLargestScoreDocs(compoundTopDocs);
+    }
+
+    private static ScoreDoc[] cloneLargestScoreDocs(List<TopDocs> docs) {
```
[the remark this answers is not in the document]
martin-gaievski: ack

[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
```diff
@@ -0,0 +1,107 @@
...
+     *  doc_id | magic_number_2
+     *  ...
+     *  doc_id | magic_number_1
+     *
+     * where doc_id is one of valid ids from result
+     *
+     * @param topDocs object with scores from multiple sub-queries
+     * @return compound TopDocs object that has results from all sub-queries
+     */
+    public static CompoundTopDocs create(final TopDocs topDocs) {
```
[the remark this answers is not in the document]
martin-gaievski: let me work on this, probably constructor should be better, but need to check if that doesn't break other things

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
```diff
@@ -0,0 +1,107 @@
...
+                maxLength = topDoc.scoreDocs.length;
+                maxScoreDocs = topDoc.scoreDocs;
+            }
+        }
+        // do deep copy
+        return Arrays.stream(maxScoreDocs).map(doc -> new ScoreDoc(doc.doc, doc.score, doc.shardIndex)).toArray(ScoreDoc[]::new);
+    }
+
+    /**
+     * Create new instance from TopDocs by parsing scores of sub-queries. Final format looks like:
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
```diff
@@ -0,0 +1,107 @@
...
+     *  doc_id | magic_number_1
+     *  doc_id | magic_number_2
+     *  ...
+     *  doc_id | magic_number_2
+     *  ...
+     *  doc_id | magic_number_2
+     *  ...
+     *  doc_id | magic_number_1
```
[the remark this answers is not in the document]
martin-gaievski: it will complicate the parsing logic. I think current format is ok as it's more like a protocol. Also it's internal and other users/engineers should follow that standard for enhancements or changes

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
```diff
@@ -0,0 +1,107 @@
...
+     *
+     * @param topDocs object with scores from multiple sub-queries
+     * @return compound TopDocs object that has results from all sub-queries
+     */
+    public static CompoundTopDocs create(final TopDocs topDocs) {
+        ScoreDoc[] scoreDocs = topDocs.scoreDocs;
+        if (Objects.isNull(scoreDocs) || scoreDocs.length < 2) {
+            return new CompoundTopDocs(topDocs.totalHits, new ArrayList<>());
+        }
+        // skipping first two elements, it's a start-stop element and delimiter for first series
```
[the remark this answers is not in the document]
martin-gaievski: that is already validated in the NormalizationProcessor, if it's not like this we're not starting the workflow

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
```diff
@@ -0,0 +1,107 @@
...
+        List<TopDocs> topDocsList = new ArrayList<>();
+        List<ScoreDoc> scoreDocList = new ArrayList<>();
+        for (int index = 2; index < scoreDocs.length; index++) {
+            // getting first element of score's series
+            ScoreDoc scoreDoc = scoreDocs[index];
+            if (isHybridQueryDelimiterElement(scoreDoc) || isHybridQueryStartStopElement(scoreDoc)) {
```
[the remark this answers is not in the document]
martin-gaievski: good idea, just package level access because of tests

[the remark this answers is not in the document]
martin-gaievski: checked, we actually cannot, we need them in process class as well to check if results are from hybrid query. I think they belong more to ```search``` package, maybe need a utility class for this

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -67,22 +75,71 @@ public void execute(
     private List<CompoundTopDocs> getQueryTopDocs(final List<QuerySearchResult> querySearchResults) {
         List<CompoundTopDocs> queryTopDocs = querySearchResults.stream()
             .filter(searchResult -> Objects.nonNull(searchResult.topDocs()))
-            .filter(searchResult -> searchResult.topDocs().topDocs instanceof CompoundTopDocs)
-            .map(searchResult -> (CompoundTopDocs) searchResult.topDocs().topDocs)
+            .map(querySearchResult -> querySearchResult.topDocs().topDocs)
+            .map(CompoundTopDocs::create)
             .collect(Collectors.toList());
+        if (queryTopDocs.size() != querySearchResults.size()) {
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -67,22 +75,71 @@ public void execute(
...
-            float maxScore = updatedTopDocs.totalHits.value > 0 ? updatedTopDocs.scoreDocs[0].score : 0.0f;
-            TopDocsAndMaxScore updatedTopDocsAndMaxScore = new TopDocsAndMaxScore(updatedTopDocs, maxScore);
+    private void updateOriginalQueryResults(
+        final List<QuerySearchResult> querySearchResults,
+        final FetchSearchResult fetchSearchResult,
+        final List<CompoundTopDocs> queryTopDocs,
+        final boolean isSingleShard
+    ) {
+        int queryTopDocsIndex = 0;
+        for (QuerySearchResult querySearchResult : querySearchResults) {
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -67,22 +75,71 @@ public void execute(
...
+
+            // create final version of top docs with all updated values
+            TopDocs topDocs = new TopDocs(updatedTopDocs.getTotalHits(), updatedTopDocs.getScoreDocs());
+
+            TopDocsAndMaxScore updatedTopDocsAndMaxScore = new TopDocsAndMaxScore(topDocs, maxScore);
             querySearchResult.topDocs(updatedTopDocsAndMaxScore, null);
         }
+        // a workaround for a single shard case, fetch has happened, and we need to update both fetch and
+        // query results
+        if (isSingleShard && querySearchResults.size() == 1) {
```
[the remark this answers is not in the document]
martin-gaievski: agree on splitting into two methods for fetch and query results.

as for the general idea - that logic is there in core from the beginning, it doesn't change frequently. We cannot do one general flow of processing because for everything other than shard == 1 there will be empty fetch results. Normal sequence is QUERY phase - call processors registered between QUERY and FETCH - FETCH phase. Only for one shard this is different. You can think of our approach as a workaround for that behavior in core.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -67,22 +75,71 @@ public void execute(
...
+        // fetch results have list of document content, that includes start/stop and
+        // delimiter elements. list is in original order from query searcher. We need to:
+        // 1. filter out start/stop and delimiter elements
+        // 2. filter out duplicates from different sub-queries
+        // 3. update original scores to normalized and combined values
+        // 4. order scores based on normalized and combined values
+        SearchHits searchHits = fetchSearchResult.hits();
+
+        // create map of docId to index of search hits
+        Map<Integer, SearchHit> docIdToSearchHit = new HashMap<>();
```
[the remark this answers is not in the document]
martin-gaievski: it won't work as we do have duplicates of docIds, streams api is failing in such case

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
martin-gaievski: for one shard we do have only one query result

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
martin-gaievski: yes, you're right. it's already handled over there, will update description

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -43,6 +45,9 @@
 @Log4j2
 public class HybridQueryPhaseSearcher extends QueryPhase.DefaultQueryPhaseSearcher {
 
+    public static final Float MAGIC_NUMBER_START_STOP = 9549511920.4881596047f;
```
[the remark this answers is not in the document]
martin-gaievski: just randomly generated number, will add comment

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
martin-gaievski: I wanted to indicate the beginning and end of the result series, it makes the logic for parsing cleaner / simpler. Another intention was extensibility, if we need to add something to the "header" part it can be between first start/stop and first delimiter. From resource consumption there is no much difference which number to send, start/stop or delimiter

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
martin-gaievski: it is required, we compute and set total hits number only for 1 shard case

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
martin-gaievski: for multi shards we use total hits from the sub-query with max results. In the processor query results will be updated, and fetch phase will get already well formed results where total hits will match number of ScoreDocs in the list. It's only different for 1 shard case

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -145,4 +205,22 @@ private float getMaxScore(final List<TopDocs> topDocs) {
     private DocValueFormat[] getSortValueFormats(final SortAndFormats sortAndFormats) {
         return sortAndFormats == null ? null : sortAndFormats.formats;
     }
+
+    /**
+     * Create ScoreDoc object that is a start/stop element in case of hybrid search query results
+     * @param docId id of one of docs from actual result object, or -1 if there are no matches
+     * @return
+     */
+    public static ScoreDoc createStartStopElementForHybridSearchResults(final int docId) {
```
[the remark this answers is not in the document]
martin-gaievski: right, that's what I'll be pushing with next commit

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
heemin32: Not about resource consumption but more about readability.
When you have a list of data and delimiter, it is natural to think delimiter exist only between item.
For example, with `,a,b,c` people naturally think there might be an empty item at the first.
When you need to add a header, you should have different marker instead. Otherwise, it will be more confusing if it is header or just another item. For example, `e, f, g, h`, is e is header or item?

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
heemin32: I admit that this is not a big issue. Just wanted to point out that totalSize is set to `searchContext` in line 187 only when it is singleShard anyway. The code will work just find without this condition check here.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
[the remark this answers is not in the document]
martin-gaievski: I see what you mean, let me check if we can rely on fetch results being present / not null / not empty. I've added few integ tests for a single shard case, it should give some conference

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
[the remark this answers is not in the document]
martin-gaievski: changed check to "if fetch results present"

heemin32 (approved):

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -67,22 +78,78 @@ public void execute(
     private List<CompoundTopDocs> getQueryTopDocs(final List<QuerySearchResult> querySearchResults) {
         List<CompoundTopDocs> queryTopDocs = querySearchResults.stream()
             .filter(searchResult -> Objects.nonNull(searchResult.topDocs()))
-            .filter(searchResult -> searchResult.topDocs().topDocs instanceof CompoundTopDocs)
-            .map(searchResult -> (CompoundTopDocs) searchResult.topDocs().topDocs)
+            .map(querySearchResult -> querySearchResult.topDocs().topDocs)
+            .map(CompoundTopDocs::new)
             .collect(Collectors.toList());
+        if (queryTopDocs.size() != querySearchResults.size()) {
+            log.warn("Some of querySearchResults are not produced by hybrid query");
```
navneet1v, line 85: if we see this warning what should the action? like in which cases this can happen? can we add some details in the warning about this.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -67,22 +78,78 @@ public void execute(
...
+        // 1. filter out start/stop and delimiter elements
+        // 2. filter out duplicates from different sub-queries
+        // 3. update original scores to normalized and combined values
+        // 4. order scores based on normalized and combined values
+        FetchSearchResult fetchSearchResult = fetchSearchResultOptional.get();
+        SearchHits searchHits = fetchSearchResult.hits();
+
+        // create map of docId to index of search hits, handles (2)
+        Map<Integer, SearchHit> docIdToSearchHit = Arrays.stream(searchHits.getHits())
+            .collect(Collectors.toMap(SearchHit::docId, Function.identity(), (a1, a2) -> a1));
```
navneet1v, line 136: [Question] How we are removing the duplicates is not clear. Can you please explain how duplicates are removed?

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessor.java
```diff
@@ -56,7 +57,8 @@ public <Result extends SearchPhaseResult> void process(
             return;
         }
         List<QuerySearchResult> querySearchResults = getQueryPhaseSearchResults(searchPhaseResult);
-        normalizationWorkflow.execute(querySearchResults, normalizationTechnique, combinationTechnique);
+        FetchSearchResult fetchSearchResult = searchPhaseResult.getAtomicArray().asList().get(0).fetchResult();
```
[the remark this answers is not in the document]
martin-gaievski: I haven't seen it, but let me add checks for that. We can consider FetchResults null in this case

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
martin-gaievski: I think it's old case when results from different nodes were not serialized properly. I haven't seen this warning since we switched to a single list approach. We can probably fail for such case, as otherwise scores will be silently corrupted. Let me add exception and push new commit

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
martin-gaievski: for delimiter and start/stop all doc_ids are same, so if we use them as map key they will be collapsed into one key-entry pair. That is enough for us, as we're using "valid" doc_id for those elements we'll update it by the score from the processor and that will be a valid entry for the final list of scores.
I'll add comment to code

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessor.java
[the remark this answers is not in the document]
martin-gaievski: Added few checks and changed to approach with Optional

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -67,22 +78,89 @@ public void execute(
...
+        if (queryTopDocs.size() != querySearchResults.size()) {
+            log.error(
+                String.format(
+                    Locale.ROOT,
+                    "sizes of querySearchResults [%d] and queryTopDocs [%d] must match. Most likely some of query results were not formatted correctly by the hybrid query",
+                    querySearchResults.size(),
+                    queryTopDocs.size()
+                )
+            );
+            throw new IllegalStateException("found inconsistent system state while processing score normalization and combination");
```
navneet1v, line 93: Lets make the error message more informative.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -67,22 +78,89 @@ public void execute(
...
+        if (querySearchResults.size() != queryTopDocs.size()) {
+            log.error(
+                String.format(
+                    Locale.ROOT,
+                    "sizes of querySearchResults [%d] and queryTopDocs [%d] must match",
+                    querySearchResults.size(),
+                    queryTopDocs.size()
+                )
+            );
+            throw new IllegalStateException("found inconsistent system state while processing score normalization and combination");
```
navneet1v, line 108: same as above.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
martin-gaievski: sure, added more details to exception message and remove logging statement as exception will be logged anyway

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
martin-gaievski: ack, pushed commit with the change

navneet1v (approved):

PR #265, opened 2023-08-28
Added strong check on number of weights equals number of sub-queries for Hybrid Search

martin-gaievski:
### Description
Enforcing checks and limits of weights for score normalization.
Following rules added:
- each weight must be between 0.0 and 1.0
- total sum of weights for score combination must be 1.0
- number of sub-queries and number of weights must match

Weight value and total sum of all weights will be checked during search processor creation. Match between number of sub-queries and weights will be checked in search time.

### Issues Resolved
https://github.com/opensearch-project/neural-search/issues/123

### Check List
    - [X] All tests pass


heemin32:
I wonder why `each weight must be between 0.0 and 1.0`. As long as we divide it by sum of weights during calculation, the sum will be 1 and each weight will be between 0.0 and 1.0. Should we have to enforce it from customer input?

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombinationUtil.java
```diff
@@ -77,4 +88,58 @@ public void validateParams(final Map<String, Object> actualParams, final Set<Str
...
+     * weights were not provided, as this is valid default value
+     * @param scores collection of scores from all sub-queries of a single hybrid search query
+     * @param weights score combination weights that are defined as part of search result processor
+     */
+    protected void validateIfWeightsMatchScores(final float[] scores, final List<Float> weights) {
+        if (weights.isEmpty()) {
+            return;
+        }
+        if (scores.length != weights.size()) {
+            log.error(
```
heemin32, line 103: No need this logging.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombinationUtil.java
```diff
@@ -77,4 +88,58 @@ public void validateParams(final Map<String, Object> actualParams, final Set<Str
...
+                String.format(
+                    Locale.ROOT,
+                    "number of weights [%d] must match number of sub-queries [%d] in hybrid query",
+                    weights.size(),
+                    scores.length
+                )
+            );
+            throw new IllegalArgumentException("number of weights must match number of sub-queries in hybrid query");
```
heemin32, line 111: Returning size of weights and queries should be fine here. So security concern.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombinationUtil.java
```diff
@@ -77,4 +88,58 @@ public void validateParams(final Map<String, Object> actualParams, final Set<Str
...
+    /**
+     * Check if provided weights are valid for combination. Following conditions are checked:
+     * - every weight is between 0.0 and 1.0
+     * - sum of all weights must be equal 1.0
+     * @param weightsList
+     */
+    private void validateWeights(final List<Float> weightsList) {
+        boolean isOutOfRange = weightsList.stream().anyMatch(weight -> !Range.between(0.0f, 1.0f).contains(weight));
+        if (isOutOfRange) {
+            log.error(
```
heemin32, line 124: Same here. Returning submitted weights in exception should be fine rather than logging duplicated message because we already knows the value is float. And float won't cause any security risk by getting returned in exception.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombinationUtil.java
```diff
@@ -77,4 +88,58 @@ public void validateParams(final Map<String, Object> actualParams, final Set<Str
...
+                    Locale.ROOT,
+                    "all weights must be in range [0.0 ... 1.0], submitted weights: %s",
+                    Arrays.toString(weightsList.toArray(new Float[0]))
+                )
+            );
+            throw new IllegalArgumentException("all weights must be in range [0.0 ... 1.0]");
+        }
+        float sumOfWeights = weightsList.stream().reduce(0.0f, Float::sum);
+        if (!DoubleMath.fuzzyEquals(1.0f, sumOfWeights, DELTA_FOR_SCORE_ASSERTION)) {
+            log.error(
```
heemin32, line 135: Same. No need an additional logging here.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombinationUtil.java
martin-gaievski: Yup, I do it as a rule of thumb, but I guess that's ok here. Let me push a commit

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombinationUtil.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombinationUtil.java
martin-gaievski: ack

navneet1v (approved):

heemin32 (approved):

Issue #260, opened 2023-08-23
Add non dense vector response support


zane-neo:
### Is your feature request related to a problem?
In near future, neural search needs to support SPLADE model and there'll be a new processor type for this model, before integrating this model, we need to have an approach to parse the model response correctly. SPLADE model doesn't return dense vectors, instead it's response is a map like below:
```
{
    ".": 0.43459656834602356,
    "2": 0.2500864267349243,
    "7": 0.7361266613006592,
    "_": 1.1532046794891357,
    "a": 0.20618286728858948,
    "j": 1.7557533979415894,
    "y": 0.15952491760253906,
    "in": 0.010410528630018234
}
```
Current code only parse the dense vector type response: https://github.com/opensearch-project/neural-search/blob/70675c8ef019c18b7a93e6c49c9fad51cd4db7bb/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java#L134. In ml-commons, all non dense vector type response are encapsulated in a map called dataAsMap in ModelTensorOutput. So to support these response, we need to add a different method to extract result from this field, then when implementing SPLADE, this new method can be used to get the model response and apply other logic to the response.

### What solution would you like?
Adding a new method in neural search MLCommonsClientAccessor to fetch response from ModelTensorOutput's dataAsMap field.

### What alternatives have you considered?
Passing a flag in MLInput to let ml-commons return string type response in ModelTensorOutput's result field and this method accept Class parameter and return generic type result. Using Gson/Jackson to deserialize the string to generic type result to make this method works with object instead of map which is easier to use. The drawback is this needs more changes in ml-commons which could introduce error in ml-commons.

### Do you have any additional context?
NA

navneet1v:
@zane-neo as this is a new feature, and will probably go through multiple iterations before it is going to be released. Hence, for all new features let's not merge the changes directly in main. Lets merge keep reviewing the changes and merge them in a feature branch. Let me create a feature branch for this new feature.

So the process will go like this:

1. Neural Search plugin maintainer will cut a feature branch from main branch. (Feature Branch: https://github.com/opensearch-project/neural-search/tree/feature/sparseVectorSupport)
2. Contributors working on Sparse Vectors will raise the PR against the feature branch.
3. Once the PR is reviewed it will go in the feature branch.
4.Once all the changes are done and performance testing is done, commits of feature branch will be merged in the main branch.

Please let me know if there is any further questions.

zane-neo:
[...]

PR #333, opened 2023-09-22
Supporting sparse semantic retrieval in neural search

### LINKED ISSUE #230: [RFC] Supporting sparse semantic retrieval based on neural models
# [RFC] Supporting sparse semantic retrieval based on neural models
## Background
Dense retrieval based on neural models has achieved great success in search relevance tasks. However, dense methods use k-NN to search the most relevant docs, which consumes large amount of memory and CPU resource. It is very expensive.
Recent years, there are a lot of research about sparse retrieval based on neural models, such as DeepCT\[1\], SparTerm\[2\], SPLADE\[3,4\]. **Since sparse retrieval fashion can be naturally implemented using inverted index, these methods are as efficient as BM25**. After fine-tuning, neural sparse retrieval can achieve high search relevance on par with dense methods. Neural sparse methods also show great generalization ability. **SPLADE defeats all dense methods in BEIR benchmark with the same setting**. Thus we propose to implement support for sparse retrieval based on neural models.

**Example Comparison 1:** Dense and Sparse retrieval models on MS-MARCO dataset.

| Type | Algorithm | Model | MRR@10 | Latency* |
| ---- | ----- | ------ | ------- | -------- |
| Dense | Encoding | BERT-tasb | 0.347 | 120ms |
| Dense | Late Interaction | ColBERT| 0.359 | 193ms |
| Sparse | BM25 | - | 0.184 | 15ms |
| Sparse | Expansion | DeepCT | 0.243 | - |
| Sparse | Expansion | SparTerm | 0.279 | - |
| Sparse | Encoding | SPLADE-max^ | 0.340 | 117ms |
| Sparse | Encoding | DistilSPLADE-max | 0.368 | - |
| Sparse | Encoding | SPLADE-doc | 0.322 | 19ms | 

**\*:** All the experiments are conducted on a single OpenSearch node with 8 * 2.4G CPU cores and 32GB RAM. \
**^:** SPLADE-max conducts a BERT model inference for query encoding, thus have similar latency to dense methods.

**Example Comparison 2:** Splade vs. Others on BEIR benchmarking dataset.

| Metrics | BM25 | ColBERT | TASB | SPLADE |
| ---------- | ---- | ------- | ---- | ------ |
| Avg. NDCG@10 | 0.440 | 0.455 | 0.435 | 0.500 |
| Winners in BEIR | 2 | 2 | 0 | 11 | 

All the above performance are extracted from the SpladeV2 [paper][4].

**Example Comparison 3:** Splade vs. openai embedding on BEIR benchmarking subset.
| Metrics | Ada | Babbage | Curie | Davinci | SPLADE |
| ---------- | ---- | ------- | ---- | ------ | ----- |
| Avg. NDCG@10 | 0.490 | 0.505 | 0.509 | 0.528 | 0.527 |
| Winners in BEIR | 0 | 0 | 1 | 5 | 5 |

Above table is extracted from the [paper][5] of openai. The experiments are conducted on a subset of BEIR benchmark.

## What are we going to do?
- Implement **api to search** in the sparse semantic space, where the query can be encoded via either a simple tokenizer or a deep neural network. 
- Implement **api to ingest** documents in the form of sparse vectors (i.e. tokens and weights) generated by neural models. 
- Provide **out-of-box model weights** for neural sparse retrieval. The model should have good search relevance on general domains without fine-tuning.

## Design
We are going to implement one IngestionProcessor for document sparse encoding and one QueryBuilder for sparse querying. Before ingestion or querying, we will have the sparse encoding model deploy via the ml-commons plugin and the ingestion processor will consequently invoke the prediction action for the encoding result. If query encoding is enabled, the query builder will also encode the queries via prediction actions while passing the query through a bert tokenizer when query encoding is disabled. Because the encoding result is in the form of sparse vector, it is very natural to adopt the term vector based Lucene index. Here is the architecture diagram.

Architecture

### Term Weight integration with Lucene
Suggested by the SPLADE paper, the relevance score is calculated following $r = \Sigma_{t \in T}w_t$ where $w_t$ is the weight of sparse term $t$ and $T$ is the intersection term set of query and document. Standard Lucene indices will only store TF(term frequency) and DF(document frequency), we will implement an analyzer that interprets term weights and stores into the payload attribute. Since the above formula is not a standard Lucene relevance scoring function, we will involve a PayloadScorer with a sum operator in the query. 

## The out-of-box Model
The schema of the sparse encoding model will be similar to SparTerm or SPLADE, where the input is a natural language sentence while the output is a sparse vector(in SPLADE, the sparse terms are BERT tokens). We will mainly focus on cross-domain optimization for a better relevance over different scenarios. The models are planned to be released in [huggingface.co](https://www.huggingface.co) and have Appache 2.0 licence.

### Ingestion Processor
The ingestion processor can be created with following API, where the field ```field_map``` can specify the fields need to be encoded and the new field names after encoding.
```
PUT /_ingest/pipeline/sparse-pipeline
{
    "description": "Calling sparse model to generate expanded tokens",
    "processors": [
        {
            "neural_sparse": {
                "model_id": "fousVokBjnSupmOha8aN",
                "field_map": {
                    "body": "body_sparse"
                }
            }
        }
    ]
}
```

### Sparse Search Query
Similar to vector search based neural search, sparse retrieval is also bind with a query type called ```neural_sparse```. One can search the field ```body_sparse``` (sparse encoded fields only) via the API below.
```
GET /test-index-3/_search
{
    "query": {
        "neural_sparse": {
            "body_sparse": {
                "query_text": "i be going",
                "model_id": "fousVokBjnSupmOha8aN",
                "tokenizer": "fousVokBjdfs4vb6gGdgYl"
            }
        }
    }
}
```
The fields ```model_id``` and ```tokenizer``` are optional, if ```model_id``` presents, the query executor will call the sparse model for query encoding, while if ```tokenizer``` presents, the executor will only encode the query via tokenization.

## Reference
[\[1\] Dai et al, Context-Aware Sentence/Passage Term Importance Estimation For First Stage Retrieval, arxiv.org. 2019.][1]

[\[2\] Bai et al, SparTerm: Learning Term-based Sparse Representation for Fast Text Retrieval, arxiv.org. 2020.][2]

[\[3\] Formal et al, SPLADE: Sparse lexical and expansion model for first stage ranking, SIGIR, 2021.][3]

[\[4\] Formal et al, SPLADE v2: Sparse lexical and expansion model for information retrieval, arxiv.org. 2021.][4]

[\[5\] Neelakantan et al, Text and Code Embeddings by Contrastive Pre-Training, arxiv.org. 2022.][5]

[1]:https://arxiv.org/abs/1910.10687
[2]:https://arxiv.org/abs/2010.00768
[3]:https://arxiv.org/pdf/2107.05720
[4]:https://arxiv.org/pdf/2109.10086
[5]:https://arxiv.org/pdf/2201.10005.pdf

zhichao-aws:
### Description
Support using sparse encoding model e.g. SPLADE in opensearch. We'll introduce a new sparse encoding doc ingest processor, and a new sparse encoding query clause.

related PR: https://github.com/opensearch-project/ml-commons/pull/1301

### Issues Resolved
#230 

### Check List
    - [x] All tests pass


zhichao-aws:
Some integ tests fail because new sparse models in ml-commons are invoked, while the PR in ml-commons haven't been merged.

zhichao-aws:
We have finalized our code, it is ready for review now

zhichao-aws:
The force push aimed to fix DCO

gradle.properties
```diff
@@ -8,4 +8,4 @@ org.gradle.jvmargs=--add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAME
   --add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED \
   --add-exports jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED \
   --add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED \
-  --add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
+  --add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
```
zane-neo, line 11: what's this for?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -144,4 +175,20 @@ private List<List<Float>> buildVectorFromResponse(MLOutput mlOutput) {
         return vector;
     }
 
+    private List<Map<String, ?>> buildMapResultFromResponse(MLOutput mlOutput) {
+        final ModelTensorOutput modelTensorOutput = (ModelTensorOutput) mlOutput;
+        final List<ModelTensors> tensorOutputList = modelTensorOutput.getMlModelOutputs();
+        if (CollectionUtils.isEmpty(tensorOutputList) || CollectionUtils.isEmpty(tensorOutputList.get(0).getMlModelTensors())) {
+            throw new IllegalStateException("Empty model result produced. Expected 1 tensor output and 1 model tensor, but got [0]");
```
zane-neo, line 182:
```suggestion
            throw new IllegalStateException("Empty model result produced. Expected at least [1] tensor output and 1 model tensor, but got [0]");
```

gradle.properties
zhichao-aws: It's a typo, didn't clean up the local settings totally. Will fix it

src/main/java/org/opensearch/neuralsearch/processor/NLPProcessor.java
```diff
@@ -0,0 +1,319 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableMap;
+
+@Log4j2
+public abstract class NLPProcessor extends AbstractProcessor {
+
+    public static final String MODEL_ID_FIELD = "model_id";
+    public static final String FIELD_MAP_FIELD = "field_map";
+
+    protected final String type;
```
zane-neo, line 35: can we use private for these class level variables?

src/main/java/org/opensearch/neuralsearch/processor/NLPProcessor.java
```diff
@@ -0,0 +1,319 @@
...
+
+    protected final String listTypeNestedMapKey;
```
zane-neo, line 37:
```suggestion
    // For nested type field which using embedding type(knn_vector, rank_features), this key will be used as the default key to the nested object.
    protected final String listTypeNestedMapKey;
```

navneet1v: what is the reason behind naming this processor as NLP processor?

zane-neo, line 37: Please add comment to this field, maybe like this: For nested type field which using embedding type(knn_vector, rank_features), this key will be used as the default key to the nested object.

navneet1v:
As discussed earlier please the changes should be merged to the feature branch. https://github.com/opensearch-project/neural-search/tree/feature/sparseVectorSupport

Also, could please add the final proposal on the github.

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
```diff
@@ -0,0 +1,306 @@
...
+     *  "SAMPLE_FIELD": {
+     *    "query_text": "string",
+     *    "model_id": "string"
+     *  }
+     *
+     */
+    public static SparseEncodingQueryBuilder fromXContent(XContentParser parser) throws IOException {
+        SparseEncodingQueryBuilder sparseEncodingQueryBuilder = new SparseEncodingQueryBuilder();
+        if (parser.currentToken() != XContentParser.Token.START_OBJECT) {
+            throw new ParsingException(parser.getTokenLocation(), "First token of " + NAME + "query must be START_OBJECT");
```
zane-neo, line 129: Do we need this parser.getTokenLocation() since we already explicitly identifying this is the First token?

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
```diff
@@ -0,0 +1,306 @@
...
+
+        BooleanQuery.Builder builder = new BooleanQuery.Builder();
+        for (Map.Entry<String, Float> entry : queryTokens.entrySet()) {
+            builder.add(FeatureField.newLinearQuery(fieldName, entry.getKey(), entry.getValue()), BooleanClause.Occur.SHOULD);
+        }
+        return builder.build();
+    }
+
+    private static void validateForRewrite(String queryText, String modelId) {
+        if (null == queryText || null == modelId) {
```
zane-neo, line 254: Please use StringUtils.isBlank to avoid unnecessary invocations, empty queryText doesn't have concrete meaning from business perspective.

src/main/java/org/opensearch/neuralsearch/util/TokenWeightUtil.java
```diff
@@ -0,0 +1,67 @@
...
+                throw new IllegalArgumentException("The data object associated with field [" + RESPONSE_KEY + "] should be a list.");
+            }
+            results.addAll((List<?>) map.get("response"));
+        }
+        return results.stream().map(TokenWeightUtil::buildTokenWeightMap).collect(Collectors.toList());
+    }
+
+    private static Map<String, Float> buildTokenWeightMap(Object uncastedMap) {
+        if (!Map.class.isAssignableFrom(uncastedMap.getClass())) {
+            throw new IllegalArgumentException("The expected inference result is a Map with String keys and  " + " Float values.");
```
zane-neo, line 56:
```suggestion
            throw new IllegalArgumentException("The expected inference result is a Map with String keys and Float values.");
```

src/main/java/org/opensearch/neuralsearch/util/TokenWeightUtil.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+        Map<String, Float> result = new HashMap<>();
+        for (Map.Entry<?, ?> entry : ((Map<?, ?>) uncastedMap).entrySet()) {
+            if (!String.class.isAssignableFrom(entry.getKey().getClass()) || !Number.class.isAssignableFrom(entry.getValue().getClass())) {
+                throw new IllegalArgumentException("The expected inference result is a Map with String keys and  " + " Float values.");
```
zane-neo, line 61:
```suggestion
                throw new IllegalArgumentException("The expected inference result is a Map with String keys and Float values.");
```

src/main/java/org/opensearch/neuralsearch/util/TokenWeightUtil.java
```diff
@@ -0,0 +1,67 @@
...
+            }
+            result.put((String) entry.getKey(), ((Number) entry.getValue()).floatValue());
```
zane-neo, line 63: Is it possible getting a cast exception when the value is not a number?

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
```diff
@@ -0,0 +1,306 @@
...
+                    + QUERY_TEXT_FIELD.getPreferredName()
+                    + " and "
+                    + MODEL_ID_FIELD.getPreferredName()
+                    + " can not be null."
+            );
+        }
+    }
+
+    private static void validateFieldType(MappedFieldType fieldType) {
+        if (!fieldType.typeName().equals("rank_features")) {
```
zane-neo, line 268: Is it possible the fieldType.typeName() return null?

src/main/java/org/opensearch/neuralsearch/processor/NLPProcessor.java
zhichao-aws: Sure

src/main/java/org/opensearch/neuralsearch/processor/NLPProcessor.java
zhichao-aws: Currently we use this abstract class for texts dense vector encoding and texts sparse vector encoding. Both of them are NLP use cases. Other names are also fine for me, do you have any suggestions?

zhichao-aws: Sure

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
zhichao-aws: This is the first token of sparse_encoding query object, but its location is uncertain. The query object may be wrapped by other query objects.

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
zhichao-aws: Sure

src/main/java/org/opensearch/neuralsearch/util/TokenWeightUtil.java
zhichao-aws: For remote inference the response is uncertain, it's possible to get a response with wrong data format. I suggest we throw Exception to help user find the problem.

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
zhichao-aws: This is a good point. The fieldType can be null, and I missed this check, will add one. However the fieldType.typeName() is an abstract method in the super class, and the inheritors need to implement it explicitly. I sampled about 10 inheritors and all of them hard-code the return value or return a static String field of the inheritor. So I don't think fieldType.typeName() can return null.

zhichao-aws:
> As discussed earlier please the changes should be merged to the feature branch. https://github.com/opensearch-project/neural-search/tree/feature/sparseVectorSupport
> 
> Also, could please add the final proposal on the github.

Sure, we'll add our final proposal. Do we need a feature branch? All our works are ensembled in this branch, and I think this is same with creating another branch and sync all commits to it

navneet1v:
@zhichao-aws as this is a new feature, hence I don't want to put this change in main before everything is completed. I mentioned the same in the old PR and github issue.

https://github.com/opensearch-project/neural-search/issues/260#issuecomment-1689339940

navneet1v:
@zhichao-aws add the entry in the change log about what this change is about.

navneet1v:
Also I see the build gradle checks getting failed. Lets make sure that everything is fixed.

zhichao-aws:
> Also I see the build gradle checks getting failed. Lets make sure that everything is fixed.

The integ test failed because it depends on the PR in ml-commons, as we use the new ml functions. We can re-run the build after https://github.com/opensearch-project/ml-commons/pull/1301 get merged.

release-notes/opensearch-neural-search.release-notes-2.10.0.0.md
```diff
@@ -4,6 +4,7 @@ Compatible with OpenSearch 2.10.0
 
 ### Features
 * Improved Hybrid Search relevancy by Score Normalization and Combination ([#241](https://github.com/opensearch-project/neural-search/pull/241/))
+* Support sparse semantic retrieval by introducing `sparse_encoding` ingest processor and query builder ([#333](https://github.com/opensearch-project/neural-search/pull/333))
 
 ### Enhancements
 * Changed format for hybrid query results to a single list of scores with delimiter ([#259](https://github.com/opensearch-project/neural-search/pull/259))
```
navneet1v, line 10: as this is the first change in the neural search after 2.10 launch can you remove all other entries and just put the one you are adding.

navneet1v:
@zhichao-aws can we add documentation on all the public classes and functions.

release-notes/opensearch-neural-search.release-notes-2.10.0.0.md
zhichao-aws: Sure, or maybe we can create a 2.11 release note and put this entry there

build.gradle
```diff
@@ -151,6 +151,8 @@ dependencies {
     runtimeOnly group: 'org.reflections', name: 'reflections', version: '0.9.12'
     runtimeOnly group: 'org.javassist', name: 'javassist', version: '3.29.2-GA'
     runtimeOnly group: 'org.opensearch', name: 'common-utils', version: "${opensearch_build}"
+    runtimeOnly group: 'com.google.code.gson', name: 'gson', version: '2.10.1'
+    runtimeOnly group: 'org.json', name: 'json', version: '20230227'
```
navneet1v, line 155: why we need both?
and what is the usage of these dependencies?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -100,10 +102,39 @@ public void inferenceSentences(
...
+        retryableInferenceSentencesWithVectorResult(targetResponseFilters, modelId, inputText, 0, listener);
     }
 
-    private void inferenceSentencesWithRetry(
+    public void inferenceSentencesWithMapResult(
+        @NonNull final String modelId,
+        @NonNull final List<String> inputText,
+        @NonNull final ActionListener<List<Map<String, ?>>> listener
+    ) {
+        retryableInferenceSentencesWithMapResult(modelId, inputText, 0, listener);
```
navneet1v, line 113: why we want to do no retires?

src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
```diff
@@ -161,6 +162,123 @@ public void testInferenceSentences_whenNotConnectionException_thenNoRetry() {
         Mockito.verify(resultListener).onFailure(illegalStateException);
     }
 
+    public void test_inferenceSentencesWithMapResult_whenValidInput_thenSuccess() {
```
jmazanec15, line 165: nit: `test_inferenceSentencesWithMapResult_whenValidInput_thenSuccess` -> `testInferenceSentencesWithMapResult_whenValidInput_thenSuccess`

src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
```diff
@@ -161,6 +162,123 @@ public void testInferenceSentences_whenNotConnectionException_thenNoRetry() {
...
+        // final List<Map<String, String>> map = List.of(Map.of("key", "value"));
```
jmazanec15, line 166: nit: remove dead code

build.gradle
zhichao-aws: When calling inferenceSentencesWithMapResult using remote inference, it will use gson to load the response from remote endpoint. And it can throw JSONException during running

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
zhichao-aws: I think we are getting consistent with the existing `inferenceSentences` method. @zane-neo Could you please help answer this?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -100,10 +102,39 @@ public void inferenceSentences(
...
+    private void retryableInferenceSentencesWithMapResult(
+        final String modelId,
+        final List<String> inputText,
+        final int retryTime,
+        final ActionListener<List<Map<String, ?>>> listener
+    ) {
+        MLInput mlInput = createMLInput(null, inputText);
+        mlClient.predict(modelId, mlInput, ActionListener.wrap(mlOutput -> {
+            final List<Map<String, ?>> result = buildMapResultFromResponse(mlOutput);
+            log.debug("Inference Response for input sentence {} is : {} ", inputText, result);
```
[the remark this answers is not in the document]
zhichao-aws: Sure, will remove it.

release-notes/opensearch-neural-search.release-notes-2.10.0.0.md
zhichao-aws: Get it, will move the commit in this file to change log

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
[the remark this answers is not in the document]
zhichao-aws: Just found existing code also log inference result, https://github.com/opensearch-project/neural-search/blob/8484be9b319a09dbdd9a1c952ca1b8515dedf571/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java#L116. I think we can just get consistent between the 2 use case

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -144,4 +175,20 @@ private List<List<Float>> buildVectorFromResponse(MLOutput mlOutput) {
         return vector;
     }
 
+    private List<Map<String, ?>> buildMapResultFromResponse(MLOutput mlOutput) {
+        final ModelTensorOutput modelTensorOutput = (ModelTensorOutput) mlOutput;
```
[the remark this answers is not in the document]
zhichao-aws: Same as the above one. Maybe we can use consistent code between the 2 use case. And If they need to be fixed, we can create another PR to fix them. https://github.com/opensearch-project/neural-search/blame/8484be9b319a09dbdd9a1c952ca1b8515dedf571/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java#L136

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
```diff
@@ -87,14 +91,18 @@ public Collection<Object> createComponents(
...
+            new QuerySpec<>(HybridQueryBuilder.NAME, HybridQueryBuilder::new, HybridQueryBuilder::fromXContent),
+            new QuerySpec<>(SparseEncodingQueryBuilder.NAME, SparseEncodingQueryBuilder::new, SparseEncodingQueryBuilder::fromXContent)
         );
     }
 
     @Override
     public Map<String, Processor.Factory> getProcessors(Processor.Parameters parameters) {
         clientAccessor = new MLCommonsClientAccessor(new MachineLearningNodeClient(parameters.client));
-        return Collections.singletonMap(TextEmbeddingProcessor.TYPE, new TextEmbeddingProcessorFactory(clientAccessor, parameters.env));
+        Map<String, Processor.Factory> allProcessors = new HashMap<>();
```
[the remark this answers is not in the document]
zhichao-aws: Sure, will change it

src/main/java/org/opensearch/neuralsearch/processor/NLPProcessor.java
```diff
@@ -0,0 +1,320 @@
...
+import org.opensearch.index.mapper.MapperService;
+import org.opensearch.ingest.AbstractProcessor;
+import org.opensearch.ingest.IngestDocument;
+import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableMap;
+
+@Log4j2
+public abstract class NLPProcessor extends AbstractProcessor {
```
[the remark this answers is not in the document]
zhichao-aws: Sure

src/main/java/org/opensearch/neuralsearch/processor/NLPProcessor.java
```diff
@@ -0,0 +1,320 @@
...
+
+    private final Environment environment;
+
+    public NLPProcessor(
+        String tag,
+        String description,
+        String type,
+        String listTypeNestedMapKey,
+        String modelId,
+        Map<String, Object> fieldMap,
```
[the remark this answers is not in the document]
zhichao-aws: I don't think so. This field is used for field name mapping. The original use case:
https://github.com/opensearch-project/neural-search/blob/8484be9b319a09dbdd9a1c952ca1b8515dedf571/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java#L46

src/main/java/org/opensearch/neuralsearch/processor/NLPProcessor.java
```diff
@@ -0,0 +1,320 @@
...
+                )) {
+            throw new IllegalArgumentException("Unable to create the processor as field_map has invalid key or value");
+        }
+    }
+
+    @SuppressWarnings({ "rawtypes" })
+    private static void validateListTypeValue(String sourceKey, Object sourceValue) {
+        for (Object value : (List) sourceValue) {
+            if (value == null) {
+                throw new IllegalArgumentException("list type field [" + sourceKey + "] has null, can not process it");
```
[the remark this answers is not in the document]
zhichao-aws: Sure, will change these wordings

src/main/java/org/opensearch/neuralsearch/processor/NLPProcessor.java
```diff
@@ -0,0 +1,320 @@
...
+                .stream()
+                .anyMatch(
+                    x -> StringUtils.isBlank(x.getKey()) || Objects.isNull(x.getValue()) || StringUtils.isBlank(x.getValue().toString())
+                )) {
+            throw new IllegalArgumentException("Unable to create the processor as field_map has invalid key or value");
+        }
+    }
+
+    @SuppressWarnings({ "rawtypes" })
+    private static void validateListTypeValue(String sourceKey, Object sourceValue) {
```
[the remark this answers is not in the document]
zhichao-aws: Will fix this

src/main/java/org/opensearch/neuralsearch/processor/NLPProcessor.java
```diff
@@ -0,0 +1,320 @@
...
+        for (Object value : (List) sourceValue) {
```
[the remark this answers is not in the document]
zhichao-aws: I don't think so, the caller `validateNestedTypeValue` has validated `sourceValue` is a list object

src/main/java/org/opensearch/neuralsearch/processor/NLPProcessor.java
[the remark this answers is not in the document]
zhichao-aws: Sure

src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
zhichao-aws: Sure

src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
zhichao-aws: Sure

zhichao-aws:
> @zhichao-aws can we add documentation on all the public classes and functions.

Sure

build.gradle
zhichao-aws: If we don't add these two dependencies, ClassNotFoundException will be throwed

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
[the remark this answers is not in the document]
jmazanec15: Yes, I think we should remove it in 2 places.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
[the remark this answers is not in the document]
jmazanec15: We can fix the other one in the future, but I want to understand how we know for certain this wont cause class cast on invalid input.

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
```diff
@@ -0,0 +1,266 @@
...
+                    + sparseEncodingQueryBuilder.fieldName()
+                    + "] and ["
+                    + parser.currentName()
+                    + "]"
+            );
+        }
+
+        requireValue(sparseEncodingQueryBuilder.fieldName(), "Field name must be provided for " + NAME + " query");
+        requireValue(
+            sparseEncodingQueryBuilder.queryText(),
```
jmazanec15, line 141: Isnt query text optional?

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
```diff
@@ -0,0 +1,266 @@
...
+        this.fieldName = in.readString();
+        this.queryText = in.readOptionalString();
+        this.modelId = in.readOptionalString();
+    }
+
+    @Override
+    protected void doWriteTo(StreamOutput out) throws IOException {
+        out.writeString(fieldName);
+        out.writeOptionalString(queryText);
+        out.writeOptionalString(modelId);
```
jmazanec15, line 92: Why are these 2 optional? In the fromXContent it looks like they require a value.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 3.0](https://github.com/opensearch-project/neural-search/compare/2.x...HEAD)
 ### Features
+Support sparse semantic retrieval by introducing `sparse_encoding` ingest processor and query builder ([#333](https://github.com/opensearch-project/neural-search/pull/333))
```
martin-gaievski, line 8: I think we typically have PR for one version, if that's for 2.11 it should be under 2.x

build.gradle
martin-gaievski: confirming, I've seen same behavior while working on multimodal

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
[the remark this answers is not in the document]
zhichao-aws: Sure

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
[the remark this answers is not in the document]
zhichao-aws: @zane-neo could you please help answer this?

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
zhichao-aws: Good point. In original design there were 2 query forms and in final proposal we only keep this one. But I didn't change here. Will fix this.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -57,275 +34,19 @@ public TextEmbeddingProcessor(
...
-    }
-
-    @SuppressWarnings({ "unchecked" })
-    private void buildMapWithKnnKeyAndOriginalValueForMapType(
-        String parentKey,
-        Object knnKey,
-        Map<String, Object> sourceAndMetadataMap,
-        Map<String, Object> treeRes
+    public void doExecute(
+        IngestDocument ingestDocument,
```
martin-gaievski, line 42: arguments should be `final` at least for public methods, that's the convention if I remember correctly

src/main/java/org/opensearch/neuralsearch/processor/factory/SparseEncodingProcessorFactory.java
```diff
@@ -0,0 +1,43 @@
...
+
+import lombok.extern.log4j.Log4j2;
+
+import org.opensearch.env.Environment;
+import org.opensearch.ingest.Processor;
+import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
+import org.opensearch.neuralsearch.processor.SparseEncodingProcessor;
+
+@Log4j2
+public class SparseEncodingProcessorFactory implements Processor.Factory {
```
martin-gaievski, line 22: Please add description of the class

src/main/java/org/opensearch/neuralsearch/util/TokenWeightUtil.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    private static Map<String, Float> buildTokenWeightMap(Object uncastedMap) {
+        if (!Map.class.isAssignableFrom(uncastedMap.getClass())) {
+            throw new IllegalArgumentException("The expected inference result is a Map with String keys and Float values.");
+        }
+        Map<String, Float> result = new HashMap<>();
+        for (Map.Entry<?, ?> entry : ((Map<?, ?>) uncastedMap).entrySet()) {
+            if (!String.class.isAssignableFrom(entry.getKey().getClass()) || !Number.class.isAssignableFrom(entry.getValue().getClass())) {
+                throw new IllegalArgumentException("The expected inference result is a Map with String keys and  " + " Float values.");
```
martin-gaievski, line 69: why we use "+" to concatenate strings, you can just use everything as one string

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+        validateForRewrite(queryText, modelId);
+        SetOnce<Map<String, Float>> queryTokensSetOnce = new SetOnce<>();
+        queryRewriteContext.registerAsyncAction(
+            ((client, actionListener) -> ML_CLIENT.inferenceSentencesWithMapResult(
+                modelId(),
+                List.of(queryText),
+                ActionListener.wrap(mapResultList -> {
+                    queryTokensSetOnce.set(TokenWeightUtil.fetchListOfTokenWeightMap(mapResultList).get(0));
```
martin-gaievski, line 194: is this safe to assume that list will have at least one element without checking is it's not empty?

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
```diff
@@ -0,0 +1,266 @@
...
+        for (Map.Entry<String, Float> entry : queryTokens.entrySet()) {
+            builder.add(FeatureField.newLinearQuery(fieldName, entry.getKey(), entry.getValue()), BooleanClause.Occur.SHOULD);
+        }
+        return builder.build();
+    }
+
+    private static void validateForRewrite(String queryText, String modelId) {
+        if (StringUtils.isBlank(queryText) || StringUtils.isBlank(modelId)) {
+            throw new IllegalArgumentException(
+                QUERY_TEXT_FIELD.getPreferredName() + " and " + MODEL_ID_FIELD.getPreferredName() + " cannot be null."
```
martin-gaievski, line 223: please use String.format for this error message

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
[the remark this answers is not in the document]
zane-neo: The two inference method we created in neural-search(inferenceWithVectorResult & inferenceWithMapResult) are for text-embedding and sparse-embedding only, and they both return ModelTensorOutput. For later one in the future, it could be expanded to other use cases, e.g. remote inference, but in this case, it's still returning ModelTensorOutput. So it's safe to cast this output to ModelTensorOutput.

CHANGELOG.md
zhichao-aws: Sure, will fix this.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
zhichao-aws: Good point. Furthermore, we can mark the two inheritor processor class final, since we can only have one behavior for one ingest processor

src/main/java/org/opensearch/neuralsearch/processor/factory/SparseEncodingProcessorFactory.java
zhichao-aws: Sure, also add one for TextEmbeddingProcessorFactory

src/main/java/org/opensearch/neuralsearch/util/TokenWeightUtil.java
zhichao-aws: I think it's the automatic tidy tool merge 2 lines. will remove the "+"

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
zhichao-aws: Good point. Will add the check at the util class

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
zane-neo: The 0 here doesn't mean no retry, it's only a initial value, this value will be increased till max retry time(3), this  can be optimized to decrease in the future to make it more intuitive though.

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
zhichao-aws: Sure, has rewritten all strings having more than 2 args to String.format

jmazanec15 (approved):

zane-neo (approved):

Issue #70, opened 2022-11-16
[Semantic Search] Support default models for index/fields


jmazanec15:
## Problem Statement

Currently, the neural-search plugin search functionality relies on the user to pass the "model_id" with each query. 

```
  "query": {
    "neural": {
      "<vector_field>": {
        "query_text": "hello world",
        "model_id": "csdsdcsdsadasdcsad",
        "k": int
      }
    }
```

This offers a suboptimal user experience. The model IDs are randomized strings that add confusion to a given query. Additionally, search behavior has to change when the model is updated (the ID needs to be updated). While it may be possible to come up with some kind of alias scheme for the model ID (see https://github.com/opensearch-project/ml-commons/issues/554), the best user experience would be for the user writing the query to not need to know any details about the model_id.

### Goal

We want to offer a user experience like this:

```
  "query": {
    "neural": {
      "<vector_field>": {
        "query_text": "hello world",
        "k": int
      }
    }
```

Similarly, for indexing, the same information could be used if no model id is specified, so the experience would look like:
```
PUT _ingest/pipeline/<pipeline_name>
{
  "description": "string",
  "processors" : [
    {
      "text_embedding": {
        "field_map": {
           "<input_field_name>": "<output_field_name>",
           ...
        }
      }
    },
    ...
  ]
}
```

### 1. Rely on index meta field

In this option, we would associate the model mapping in a field in the _meta field of the index.
```
PUT my-neural-index
{
  "mappings": {
    "_meta": { 
      "neural_field_map": {
        "field_name_1": "<model_id>",
        "field_name_2": "<model_id>",
        "field_name_3": "<model_id>",
      }
    }
  }
}
```

### 2. Make model map index settings

Similar to _meta field, we could make the map an index setting (would need to validate that settings can in fact be maps). Index settings would give us more control over validation of input model ids as well as some hooks to trigger actions when settings are updated.

```
PUT my-neural-index/_settings
{
    "index": { 
      "neural_field_map": {
        "field_name_1": "<model_id>",
        "field_name_2": "<model_id>",
        "field_name_3": "<model_id>",
      }
    }
}
```

### 3. Use system index 

Using a system index is another approach to associating this model information with a given index. However, maintaining a system index is heavier than relying on a _meta field. This would require several APIs to manage the functionality. If we are to create a system index for model management, it would be better to group this functionality with ml-commons, which already has a model system index (see next option).

### 4. Rely on model index association during model management

[...]

PR #337, opened 2023-09-25
[Feature] Support for Default Model Id

vibrantvarun:
### Description
This PR provides a functionality to client to skip passing model id in the neural search query request. 

### Issues Resolved
[70](https://github.com/opensearch-project/neural-search/issues/70)

### Check List
    - [x] All tests pass


navneet1v:
@vibrantvarun remove the skip changelog tag and put the info in change log.

navneet1v:
@vibrantvarun fill the checklist too

navneet1v:
@vibrantvarun add documentation on all public classes and public functions

src/main/java/org/opensearch/neuralsearch/query/visitor/NeuralSearchQueryVisitor.java
```diff
@@ -0,0 +1,48 @@
...
+import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.index.query.QueryBuilderVisitor;
+import org.opensearch.neuralsearch.query.NeuralQueryBuilder;
+
+public class NeuralSearchQueryVisitor implements QueryBuilderVisitor {
+
+    private String modelId;
+    private Map<String, Object> neuralFieldMap;
+
+    public NeuralSearchQueryVisitor(String modelId, Map<String, Object> neuralFieldMap) {
```
navneet1v, line 20: we can use @AllArgsConstructor here

src/main/java/org/opensearch/neuralsearch/query/visitor/NeuralSearchQueryVisitor.java
```diff
@@ -0,0 +1,48 @@
...
+        this.modelId = modelId;
+        this.neuralFieldMap = neuralFieldMap;
+    }
+
+    @Override
+    public void accept(QueryBuilder queryBuilder) {
```
navneet1v, line 26: as per my understanding the logic here should be :
1. If model id is provided in query then do nothing.
2. If model id is present in the field map, use that model id
3. If no field is not in the field map, then use the model id provided in the SearchRequestProcessor. 

The current code doesn't seem to work like that. Can you share some light on the logic.

src/main/java/org/opensearch/neuralsearch/util/NeuralSearchClusterUtil.java
```diff
@@ -0,0 +1,63 @@
...
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import lombok.extern.log4j.Log4j2;
+
+import org.opensearch.Version;
+import org.opensearch.cluster.service.ClusterService;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@Log4j2
+public class NeuralSearchClusterUtil {
```
navneet1v, line 19: do we really need this class cannot we directly get it from any OpenSearch class?

src/main/java/org/opensearch/neuralsearch/query/visitor/NeuralSearchQueryVisitor.java
navneet1v: cool, lets fix the logic.

src/main/java/org/opensearch/neuralsearch/util/NeuralSearchClusterUtil.java
vibrantvarun: I would suggest let's keep this class as it provides abstraction from underlying opensearch cluster, also if i have to remove this then I have initialize cluster service variable in the create components and then have take it all the way to neural query builder.

vibrantvarun: Because Version data comes from clusterService.state().nodes().getMinNodeVersion()

src/test/java/org/opensearch/neuralsearch/common/BaseNeuralSearchIT.java
```diff
@@ -68,15 +76,33 @@ public abstract class BaseNeuralSearchIT extends OpenSearchSecureRestTestCase {
...
+        clusterService = createClusterService(threadPool);
         if (isUpdateClusterSettings()) {
             updateClusterSettings();
         }
+        NeuralSearchClusterUtil.instance().initialize(clusterService);
+    }
+
+    protected ThreadPool setUpThreadPool() {
+        return new TestThreadPool(getClass().getName(), threadPoolSettings());
+    }
```
navneet1v, line 98: lets move this function after all public functions.

src/test/java/org/opensearch/neuralsearch/common/BaseNeuralSearchIT.java
```diff
@@ -68,15 +76,33 @@ public abstract class BaseNeuralSearchIT extends OpenSearchSecureRestTestCase {
...
+    protected ClusterService clusterService;
+
     protected void setPipelineConfigurationName(String pipelineConfigurationName) {
         this.PIPELINE_CONFIGURATION_NAME = pipelineConfigurationName;
     }
 
     @Before
     public void setupSettings() {
+        threadPool = setUpThreadPool();
+        clusterService = createClusterService(threadPool);
```
navneet1v, line 89: do we really need this? what are really testing with these, as these are integration tests

src/test/java/org/opensearch/neuralsearch/util/NeuralSearchClusterUtilTests.java
```diff
@@ -0,0 +1,51 @@
...
+import static org.mockito.Mockito.when;
+import static org.opensearch.neuralsearch.util.NeuralSearchClusterTestUtils.mockClusterService;
+
+import org.opensearch.Version;
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.test.OpenSearchTestCase;
+
+public class NeuralSearchClusterUtilTests extends OpenSearchTestCase {
+
+    public void testSingleNodeCluster() {
```
navneet1v, line 18: try to put your tests function names like

test<WhatYourTesting>_when<ABCCondition>_then<XYZResult>

src/test/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessorIT.java
```diff
@@ -0,0 +1,81 @@
...
+import org.opensearch.neuralsearch.common.BaseNeuralSearchIT;
+import org.opensearch.neuralsearch.query.NeuralQueryBuilder;
+
+import com.google.common.primitives.Floats;
+
+public class NeuralQueryProcessorIT extends BaseNeuralSearchIT {
+
+    public static final String index = "my-nlp-index";
+    public static final String search_pipeline = "search-pipeline";
+    public static final String ingest_pipeline = "nlp-pipeline";
```
navneet1v, line 29: why public?

src/test/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessorIT.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @After
+    @SneakyThrows
+    public void tearDown() {
+        super.tearDown();
+        deleteSearchPipeline(search_pipeline);
+        findDeployedModels().forEach(this::deleteModel);
+    }
+
+    public void testNeuralQueryProcessor() throws Exception {
```
navneet1v, line 50: use @SneakyThrows

src/test/java/org/opensearch/neuralsearch/common/BaseNeuralSearchIT.java
vibrantvarun: Yes we need this because when we get node version from cluster. this.clusterService.getState.getNode.getMinNodeVersion.

While executing integ test case of NeuralQueryProcessor, **this.clusterService** throws null pointer if we do not mock it.

navneet1v (commented):
Can we add Unit tests for the visitor class. and test all the scenarios

src/test/java/org/opensearch/neuralsearch/common/BaseNeuralSearchIT.java
navneet1v: But these are integration steps, why do I need to mock it?

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
```diff
@@ -0,0 +1,94 @@
...
+         */
+        @Override
+        public NeuralQueryProcessor create(
+            Map<String, Processor.Factory<SearchRequestProcessor>> processorFactories,
+            String tag,
+            String description,
+            boolean ignoreFailure,
+            Map<String, Object> config,
+            PipelineContext pipelineContext
+        ) throws Exception {
```
jmazanec15, line 83: Can we specify the exception as IllegalArgumentException instead of Exception?

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -112,7 +123,9 @@ protected void doXContent(XContentBuilder xContentBuilder, Params params) throws
         xContentBuilder.startObject(NAME);
         xContentBuilder.startObject(fieldName);
         xContentBuilder.field(QUERY_TEXT_FIELD.getPreferredName(), queryText);
-        xContentBuilder.field(MODEL_ID_FIELD.getPreferredName(), modelId);
+        if (!isClusterOnOrAfterMinRequiredVersion() || (isClusterOnOrAfterMinRequiredVersion() && modelId != null)) {
```
jmazanec15, line 126: Can this be simplified to:
```
if (modelId != null) {

}
```

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -258,4 +272,8 @@ protected int doHashCode() {
     public String getWriteableName() {
         return NAME;
     }
+
+    private static boolean isClusterOnOrAfterMinRequiredVersion() {
```
jmazanec15, line 276: One problem with this is that the function name does not indicate that it has anything to do with default model ID. I would recommend parametrizing this with the version.

src/main/java/org/opensearch/neuralsearch/query/visitor/NeuralSearchQueryVisitor.java
```diff
@@ -0,0 +1,60 @@
...
+
+import lombok.AllArgsConstructor;
+
+import org.apache.lucene.search.BooleanClause;
+import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.index.query.QueryBuilderVisitor;
+import org.opensearch.neuralsearch.query.NeuralQueryBuilder;
+
+/**
+ * Neural Search Query Visitor. It visits the each and every component of query buikder tree.
```
jmazanec15, line 18: nit: "It visits the each and every component of query buikder tree." -> "It visits each and every component of query builder tree."

src/test/java/org/opensearch/neuralsearch/common/BaseNeuralSearchIT.java
vibrantvarun: Oops I used the wrong word here, instead of mock I should I have used initialize. Opensearch has created a test suite for integ test called OpenSearchIntegTestCase. We use is BaseNeuralSearchIT in our test cases. Now in OpenSearchIntegTestCase they [initialize](https://github.com/opensearch-project/OpenSearch/blob/8f4f9951de817a58d18815c5dd35896dae2bced1/test/framework/src/main/java/org/opensearch/test/OpenSearchIntegTestCase.java#L654) clusterService. In our case we don't initalize so that is the reason I was getting null in NeuralSearchClusterUtil. Now I have initialized.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
vibrantvarun: Yes, I was planning to do the same. But @navneet1v earlier suggested to put this clusterOnorAfter checks so to maintain the similarity in the work I used this one. Any thoughts @navneet1v ?

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
```diff
@@ -0,0 +1,94 @@
...
+import org.opensearch.action.search.SearchRequest;
+import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.ingest.ConfigurationUtils;
+import org.opensearch.neuralsearch.query.visitor.NeuralSearchQueryVisitor;
+import org.opensearch.search.pipeline.AbstractProcessor;
+import org.opensearch.search.pipeline.Processor;
+import org.opensearch.search.pipeline.SearchRequestProcessor;
+
+/**
+ * Neural Search Query Request Processor
```
martin-gaievski, line 19: Can we have more descriptive text here, this one looks more like a class name

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+public class NeuralQueryProcessor extends AbstractProcessor implements SearchRequestProcessor {
+
+    /**
+     * Key to reference this processor type from a search pipeline.
+     */
+    public static final String TYPE = "neural_query";
```
martin-gaievski, line 26: is this the agreed name for this processor? It sounds like a processor suppose to execute the query, maybe something describing actual functionality like enriching query with missing data.

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
```diff
@@ -0,0 +1,94 @@
...
+     * @return The Search Request.
+     */
+    @Override
+    public SearchRequest processRequest(SearchRequest searchRequest) {
+        QueryBuilder queryBuilder = searchRequest.source().query();
+        queryBuilder.visit(new NeuralSearchQueryVisitor(modelId, neuralFieldDefaultIdMap));
+        return searchRequest;
+    }
+
+    public static class Factory implements Processor.Factory<SearchRequestProcessor> {
```
martin-gaievski, line 66: this should be a separate class, please check existing factory classes for text_embedding for example

src/main/java/org/opensearch/neuralsearch/query/visitor/NeuralSearchQueryVisitor.java
```diff
@@ -0,0 +1,60 @@
...
+import org.opensearch.neuralsearch.query.NeuralQueryBuilder;
+
+/**
+ * Neural Search Query Visitor. It visits each and every component of query buikder tree.
+ */
+@AllArgsConstructor
+public class NeuralSearchQueryVisitor implements QueryBuilderVisitor {
+
+    private String modelId;
+    private Map<String, Object> neuralFieldMap;
```
martin-gaievski, line 24: if those are only set in constructor please make them final

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -93,7 +96,11 @@ public NeuralQueryBuilder(StreamInput in) throws IOException {
         super(in);
         this.fieldName = in.readString();
         this.queryText = in.readString();
-        this.modelId = in.readString();
+        if (isClusterOnOrAfterMinReqVersionForDefaultModelIdSupport()) {
```
martin-gaievski, line 99: please add a comment on what the new logic is. You may move it to separate private method `getModelId` and put it in the method's description

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
```diff
@@ -0,0 +1,94 @@
...
+ * Neural Search Query Request Processor
+ */
+public class NeuralQueryProcessor extends AbstractProcessor implements SearchRequestProcessor {
+
+    /**
+     * Key to reference this processor type from a search pipeline.
+     */
+    public static final String TYPE = "neural_query";
+
+    final String modelId;
```
jmazanec15, line 28: private?

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
```diff
@@ -0,0 +1,94 @@
...
+
+    final Map<String, Object> neuralFieldDefaultIdMap;
```
jmazanec15, line 30: private? Also, could we add a comment for this member?

Also, why is this map <String, Object> and not <String, String>

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
```diff
@@ -0,0 +1,94 @@
...
+     * Returns the type of the processor.
+     *
+     * @return The processor type.
+     */
+    @Override
+    public String getType() {
+        return TYPE;
+    }
+
+    protected NeuralQueryProcessor(
```
jmazanec15, line 42: Can this just be package-private or private? Will this class be sub-classed?

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
```diff
@@ -0,0 +1,94 @@
...
+        @Override
+        public NeuralQueryProcessor create(
+            Map<String, Processor.Factory<SearchRequestProcessor>> processorFactories,
+            String tag,
+            String description,
+            boolean ignoreFailure,
+            Map<String, Object> config,
+            PipelineContext pipelineContext
+        ) throws IllegalArgumentException {
+            String modelId = (String) config.remove(DEFAULT_MODEL_ID);
```
jmazanec15, line 84: Is this cast safe?

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
```diff
@@ -0,0 +1,94 @@
...
+    public String getType() {
+        return TYPE;
+    }
+
+    protected NeuralQueryProcessor(
+        String tag,
+        String description,
+        boolean ignoreFailure,
+        String modelId,
+        Map<String, Object> neuralFieldDefaultIdMap
```
jmazanec15, line 47: Could we add @Nullable annotation to indicate these can be null?

src/main/java/org/opensearch/neuralsearch/util/NeuralSearchClusterUtil.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Return minimal OpenSearch version based on all nodes currently discoverable in the cluster
+     * @return minimal installed OpenSearch version, default to Version.CURRENT which is typically the latest version
+     */
+    public Version getClusterMinVersion() {
+        try {
+            return this.clusterService.state().getNodes().getMinNodeVersion();
+        } catch (Exception exception) {
+            log.error(
```
jmazanec15, line 54: This seems dangerous to suppress all exceptions. This looks like it was taken from https://github.com/opensearch-project/k-NN/blob/ca5e483e1e70abdc19196e1018b3a7fd06908bf6/src/main/java/org/opensearch/knn/index/KNNClusterUtil.java#L46-L56. @martin-gaievski What was reasoning behind doing this?

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
vibrantvarun: I personally think this current one looks good, we are anyway explaining everything in docs and other places. Type can  be a little lighter as well

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
vibrantvarun: Text embedding processor is a ingest processor. Neural Query Processor is a search request processor. I have followed the same template of search request processor how they are implemented in OpenSearch. We should follow the same template of OS

[FilterQueryRequestProcessor](https://github.com/opensearch-project/OpenSearch/blob/main/modules/search-pipeline-common/src/main/java/org/opensearch/search/pipeline/common/FilterQueryRequestProcessor.java)

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
vibrantvarun: Yes I will add the comment. Moving into method looks unecessory.

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
vibrantvarun: Because [ConfigurationUtils](https://github.com/opensearch-project/OpenSearch/blob/d656e3db592f29466d35452867caa241f5429485/server/src/main/java/org/opensearch/ingest/ConfigurationUtils.java#L359) The class from Opensearch which is used to retrieve map from processor is generic and it Map<String, Object> defined in it. ConfigurationUtils is the designed to get values from the processor in OS. So I am using the same class

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
vibrantvarun: Yeah I can put a null pointer check

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
martin-gaievski: I would go with something like `enriching_query_defaults`, that is still light to me. The docs will be there for sure, what I mean is that having just `neural_query` gives wrong impression that we're executing that query, it's not about less details but about wrong direction. Plus with more generic name we can re-use that same processor for defaults in other fields.

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
martin-gaievski: I don't see a reason why to follow that pattern, what makes more sense to me for decision is if that class has important and complex logic/responsibility, or not. We've done a separate class for [normalization processor factory](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/factory/NormalizationProcessorFactory.java), processor is a search processor. 
If you think factory for your processor is simple we can leave it as inner class

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
vibrantvarun: makes sense.

src/main/java/org/opensearch/neuralsearch/util/NeuralSearchClusterUtil.java
martin-gaievski: if I'm not mistaken, the idea was to allow cluster work with a `current` version instead of failing, maybe that core api was throwing some exception at that time. I checked core code, seems it's pretty straightforward now, I think we can get rid of that catch_all approach

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
vibrantvarun: Yes it is simple.

src/main/java/org/opensearch/neuralsearch/query/visitor/NeuralSearchQueryVisitor.java
vibrantvarun: I think we should not keep it final because either of the value can be null. and variables with final keyword cannot have a value null

src/main/java/org/opensearch/neuralsearch/util/NeuralSearchClusterUtil.java
vibrantvarun: Taken care off now

navneet1v:
@vibrantvarun can we fix the github action: codecov/patch and codecov/project 

seems like the coverage has dropped

src/test/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessorTests.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Map;
+
+import org.opensearch.action.search.SearchRequest;
+import org.opensearch.neuralsearch.query.NeuralQueryBuilder;
+import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.test.OpenSearchTestCase;
+
+public class NeuralQueryProcessorTests extends OpenSearchTestCase {
+
+    public void testFactory() throws Exception {
```
martin-gaievski, line 19: please use naming convention from other tests - whatYouTesting_conditions_expectation, I think <conditions> may be optional sometimes, when it's obvious, but other two are must

src/test/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessorTests.java
```diff
@@ -0,0 +1,62 @@
...
+        NeuralQueryProcessor.Factory factory = new NeuralQueryProcessor.Factory();
+        NeuralQueryProcessor processor = createTestProcessor(factory);
+        assertEquals("vasdcvkcjkbldbjkd", processor.getModelId());
+        assertEquals("bahbkcdkacb", processor.getNeuralFieldDefaultIdMap().get("fieldName").toString());
+
+        // Missing "query" parameter:
```
martin-gaievski, line 25: why do we need this?

src/test/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessorTests.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    public void testFactory_whenModelIdIsNotString_thenFail() {
+        NeuralQueryProcessor.Factory factory = new NeuralQueryProcessor.Factory();
+        Map<String, Object> configMap = new HashMap<>();
+        configMap.put("default_model_id", 55555L);
+        expectThrows(IllegalArgumentException.class, () -> factory.create(Collections.emptyMap(), null, null, false, configMap, null));
+    }
+
+    public void testProcessRequest() throws Exception {
```
martin-gaievski, line 39: nit: you can use @sneakythrow for exceptions in tests

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
```diff
@@ -0,0 +1,106 @@
...
+import org.opensearch.search.pipeline.Processor;
+import org.opensearch.search.pipeline.SearchRequestProcessor;
+
+/**
+ * Neural Search Query Request Processor, It modifies the search request with neural query clause
+ * and adds model Id if not present in the search query.
+ */
+@Setter
+@Getter
+public class NeuralQueryProcessor extends AbstractProcessor implements SearchRequestProcessor {
```
martin-gaievski, line 28: I would rename class to something closer to a processor type name, although I left the final call to you

martin-gaievski (commented):
few minor things, over looks good

src/test/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessorTests.java
vibrantvarun: To check exception

src/main/java/org/opensearch/neuralsearch/processor/NeuralQueryProcessor.java
navneet1v: @martin-gaievski Enriching Query Processor is too generic. Can we rename it to NeuralQueryEnrichingProcessor. In my understanding we should limit the scope of this processor to only neural queries. Nothing else. If we go too generic then this will become a default processor to put any kind of enriching which will bloat up the responsibilities of this processor.

src/main/java/org/opensearch/neuralsearch/processor/EnrichingQueryDefaultProcessor.java
```diff
@@ -0,0 +1,106 @@
...
+ * and adds model Id if not present in the search query.
+ */
+@Setter
+@Getter
+public class EnrichingQueryDefaultProcessor extends AbstractProcessor implements SearchRequestProcessor {
+
+    /**
+     * Key to reference this processor type from a search pipeline.
+     */
+    public static final String TYPE = "enriching_query_defaults";
```
navneet1v, line 33: neural_query_enricher name seems better here

src/main/java/org/opensearch/neuralsearch/processor/EnrichingQueryDefaultProcessor.java
```diff
@@ -0,0 +1,106 @@
...
+            Map<String, Processor.Factory<SearchRequestProcessor>> processorFactories,
+            String tag,
+            String description,
+            boolean ignoreFailure,
+            Map<String, Object> config,
+            PipelineContext pipelineContext
+        ) throws IllegalArgumentException {
+            String modelId;
+            try {
+                modelId = (String) config.remove(DEFAULT_MODEL_ID);
```
navneet1v, line 93: why config.remove?

navneet1v, line 93: There is a better way to get the string. Check this: https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/factory/TextEmbeddingProcessorFactory.java#L40C9-L40C89

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
```diff
@@ -80,6 +80,7 @@ public Collection<Object> createComponents(
         final IndexNameExpressionResolver indexNameExpressionResolver,
         final Supplier<RepositoriesService> repositoriesServiceSupplier
     ) {
+        NeuralSearchClusterUtil.instance().initialize(clusterService);
```
navneet1v, line 83: We should start getting rid of these kind of inits and move towards singleton pattern without this kind of inits. 

May be an AI for maintainers

src/test/java/org/opensearch/neuralsearch/processor/EnrichingQueryDefaultProcessorIT.java
```diff
@@ -0,0 +1,82 @@
...
+        prepareModel();
+    }
+
+    @After
+    @SneakyThrows
+    public void tearDown() {
+        super.tearDown();
+        deleteSearchPipeline(search_pipeline);
+        findDeployedModels().forEach(this::deleteModel);
+    }
```
navneet1v, line 47: we should delete the index too.

src/test/resources/processor/IndexMappings.json
```diff
@@ -67,6 +67,13 @@
             }
           }
         }
+      },
+      "passage_embedding": {
+        "type": "knn_vector",
+        "dimension": 768
+      },
+      "passage_text": {
+        "type": "text"
```
navneet1v, line 76: why we need these new fields? why we cannot use older fields?

vibrantvarun: Because they don't have text matching

src/main/java/org/opensearch/neuralsearch/processor/EnrichingQueryDefaultProcessor.java
vibrantvarun: @martin-gaievski I am finally going with neural_query_enricher

src/main/java/org/opensearch/neuralsearch/processor/EnrichingQueryDefaultProcessor.java
vibrantvarun: Took reference from OS core

martin-gaievski (approved):

src/test/resources/processor/IndexMappings.json
navneet1v: do we need text matching?

vibrantvarun: yeah i wrote a test case where I am passing querytext as hello world. If I don't add this field here then it will fail

navneet1v (approved):

src/test/resources/processor/IndexMappings.json
```diff
@@ -67,6 +67,13 @@
             }
           }
         }
+      },
+      "passage_embedding": {
+        "type": "knn_vector",
+        "dimension": 768
```
martin-gaievski, line 73: this need to have a `lucene` as engine, otherwise it uses default which is `nmslib` and that fails, as nmslib is not part of the knn zip fopr integ tests @vibrantvarun

heemin32: Where is this field used by the way?

PR #348, opened 2023-09-27
Add max token score for SparseEncodingQueryBuilder and do renaming

### LINKED ISSUE #230: [RFC] Supporting sparse semantic retrieval based on neural models

zhichao-aws:
### Description
For sparse semantic retrieval in neural search, we use lucene FeatureField for storage and use lucene FeatureQuery to search. The feature queries of input tokens are wrapped by lucene BooleanQuery, which use WAND algorithm to accelerate the execution. The WAND algorithm leverage the score upper bound of sub-queries to skip non-competitive tokens. However, origin lucene FeatureQuery use Float.MAX_VALUE as the score upper bound, and this invalidates WAND.

To mitigate this issue, we rewrite the FeatureQuery to BoundedLinearFeatureQuery. The caller can set the token score upperbound of this query. And according to our use case, we use LinearFunction as the score function.

We have conducted several end to end benchmark experiments with this optimization. Using a doc-only SPLADE like model, this optimization reduce the query latency from P90 40ms to P90 26ms (1 million docs), and reduce the query latency from P90 231ms to P90 80ms (8.8 million docs).

After lucene version 9.8, the FeatureQuery are rewritten, and lucene optimize the speed for BooleanQuery. Then this optimization is no longer needed. However, we're most likely not upgrade to lucene 9.8 in 2.11 release https://github.com/opensearch-project/OpenSearch/pull/8668. So we create this PR for 2.x only, and in main, we'll create another PR to deperacate the `max_token_score` parameter in sparse query clause.

### Issues Resolved
#230 

### Check List
    - [x] All tests pass


zhichao-aws:
The changes we made to origin FeatureQuery:
https://github.com/opensearch-project/neural-search/pull/348/files/9d4a591e3bc62715efd635380cb2cea38ecc54fd..df99d0b4ab43e3eb8648e8384f3b804c309ea73c#diff-ff541c86a7b171bd13284d781b39e31ce83861938b6acea54894c0e5d6702bba

src/main/java/org/opensearch/neuralsearch/query/BoundedLinearFeatureQuery.java
```diff
@@ -0,0 +1,230 @@
...
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+/*
+ * This class is built based on lucene FeatureQuery. We use LinearFuntion to
+ * build the query and add an upperbound to it.
+ */
+
+package org.opensearch.neuralsearch.query;
```
navneet1v, line 36: lets move this to another package similar to what core has done for lucene classes.
`org.apache.lucene.query`


https://github.com/opensearch-project/OpenSearch/tree/main/server/src/main/java/org/apache/lucene

src/main/java/org/opensearch/neuralsearch/query/BoundedLinearFeatureQuery.java
```diff
@@ -0,0 +1,230 @@
...
+ * as the score function.
+ */
+
+public final class BoundedLinearFeatureQuery extends Query {
+
+    private final String fieldName;
+    private final String featureName;
+    private final Float scoreUpperBound;
+
+    public BoundedLinearFeatureQuery(String fieldName, String featureName, Float scoreUpperBound) {
```
navneet1v, line 77: can we comment on the lines which we have added on top Lucene code so that it is easy to debug and fix later.

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
```diff
@@ -99,6 +104,7 @@ protected void doXContent(XContentBuilder xContentBuilder, Params params) throws
         xContentBuilder.startObject(fieldName);
         xContentBuilder.field(QUERY_TEXT_FIELD.getPreferredName(), queryText);
         xContentBuilder.field(MODEL_ID_FIELD.getPreferredName(), modelId);
+        if (null != maxTokenScore) xContentBuilder.field(MAX_TOKEN_SCORE_FIELD.getPreferredName(), maxTokenScore);
```
navneet1v, line 107:
```suggestion
        if (maxTokenScore != null) {
            xContentBuilder.field(MAX_TOKEN_SCORE_FIELD.getPreferredName(), maxTokenScore);
        }
```

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
```diff
@@ -99,6 +104,7 @@ protected void doXContent(XContentBuilder xContentBuilder, Params params) throws
...
         printBoostAndQueryName(xContentBuilder);
```
navneet1v, line 108: why do we need to print ?

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
```diff
@@ -146,6 +153,9 @@ public static SparseEncodingQueryBuilder fromXContent(XContentParser parser) thr
             sparseEncodingQueryBuilder.modelId(),
             String.format(Locale.ROOT, "%s field must be provided for [%s] query", MODEL_ID_FIELD.getPreferredName(), NAME)
         );
+        if (null != sparseEncodingQueryBuilder.maxTokenScore && sparseEncodingQueryBuilder.maxTokenScore <= 0) {
```
navneet1v, line 156: same as above

src/main/java/org/opensearch/neuralsearch/query/BoundedLinearFeatureQuery.java
```diff
@@ -0,0 +1,230 @@
...
+        }
+    }
+
+    @Override
+    public String toString(String field) {
+        return "BoundedLinearFeatureQuery(field=" + fieldName + ", feature=" + featureName + ", scoreUpperBound=" + scoreUpperBound + ")";
+    }
+
+    // the field and decodeFeatureValue are modified from FeatureField.decodeFeatureValue
+    static final int MAX_FREQ = Float.floatToIntBits(Float.MAX_VALUE) >>> 15;
```
navneet1v, line 220: what is the value of this MAX_FREQ?

navneet1v:
Minor comments on the code.

src/main/java/org/opensearch/neuralsearch/query/BoundedLinearFeatureQuery.java
zane-neo: This is copied from https://github.com/apache/lucene/blob/6d764c3397d00f93bd4273bd8d1c9e51d6e104e6/lucene/core/src/java/org/apache/lucene/document/FeatureField.java#L207, including the below method is from `FeatureField` but we changed it to make our new parameter scoreUpperBound work.

src/main/java/org/opensearch/neuralsearch/query/BoundedLinearFeatureQuery.java
zane-neo: We can treat this a brand new class, it combines the FeatureQuery class and some methods from FeatureField class in lucene, we'll add some comments on top of the core function that makes our new parameter work.

zhichao-aws:
Also include renaming changes in this PR. The renaming changes in main: https://github.com/opensearch-project/neural-search/pull/353

zhichao-aws:
Also include renaming changes in this PR. The renaming changes in main: https://github.com/opensearch-project/neural-search/pull/353

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
zhichao-aws: max_token_score is a parameter for this query clause, and it should be treated same as other parameters

zane-neo: This is similar to exist code: https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java#L120, I think this is for debugging purpose.

src/main/java/org/opensearch/neuralsearch/query/BoundedLinearFeatureQuery.java
```diff
@@ -0,0 +1,230 @@
...
+ * to accelerate the execution. The WAND algorithm leverage the score upper bound of sub-queries to
+ * skip non-competitive tokens. However, origin lucene FeatureQuery use Float.MAX_VALUE as the score
+ * upper bound, and this invalidates WAND.
+ *
+ * To mitigate this issue, we rewrite the FeatureQuery to BoundedLinearFeatureQuery. The caller can
+ * set the token score upperbound of this query. And according to our use case, we use LinearFunction
+ * as the score function.
+ */
+
+public final class BoundedLinearFeatureQuery extends Query {
```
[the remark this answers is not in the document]
zhichao-aws: Sure

zane-neo (approved):

src/main/java/org/opensearch/neuralsearch/query/SparseEncodingQueryBuilder.java
zhichao-aws: The print is not logging, but output to XContent. The boost and query name are also parameters and should be included in the XContent

navneet1v:
@zane-neo @zhichao-aws the unit tests have not covered some lines leading to failures of GH workflow, can we fix it

src/main/java/org/opensearch/neuralsearch/query/BoundedLinearFeatureQuery.java
navneet1v: Yes lets add comments over lines which are copied so that we are aware of from where the code is coming from

src/main/java/org/apache/lucene/BoundedLinearFeatureQuery.java
```diff
@@ -0,0 +1,235 @@
...
+ * To mitigate this issue, we rewrite the FeatureQuery to BoundedLinearFeatureQuery. The caller can
+ * set the token score upperbound of this query. And according to our use case, we use LinearFunction
+ * as the score function.
+ *
+ * This class combines both <a href="https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/document/FeatureQuery.java">FeatureQuery</a>
+ * and <a href="https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/document/FeatureField.java">FeatureField</a> together
+ * and will be deprecated after OpenSearch upgraded lucene to version 9.8.
+ */
+
+public final class BoundedLinearFeatureQuery extends Query {
```
navneet1v, line 75: There are no unit test for this class. Can we add the UTs for this class.

src/main/java/org/apache/lucene/BoundedLinearFeatureQuery.java
```diff
@@ -0,0 +1,235 @@
...
+
+    private final String fieldName;
+    private final String featureName;
+    private final Float scoreUpperBound;
+
+    public BoundedLinearFeatureQuery(String fieldName, String featureName, Float scoreUpperBound) {
+        this.fieldName = Objects.requireNonNull(fieldName);
+        this.featureName = Objects.requireNonNull(featureName);
+        this.scoreUpperBound = Objects.requireNonNull(scoreUpperBound);
```
navneet1v, line 84: have you considered using @NonNull in params for checking if input params are null or not?

src/main/java/org/apache/lucene/BoundedLinearFeatureQuery.java
```diff
@@ -0,0 +1,235 @@
...
+    @Override
+    public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost) throws IOException {
+        if (!scoreMode.needsScores()) {
+            // We don't need scores (e.g. for faceting), and since features are stored as terms,
+            // allow TermQuery to optimize in this case
+            TermQuery tq = new TermQuery(new Term(fieldName, featureName));
+            return searcher.rewrite(tq).createWeight(searcher, scoreMode, boost);
+        }
+
+        return new Weight(this) {
```
navneet1v, line 122: can we move this weight class as inner class or a separate class?

src/main/java/org/apache/lucene/BoundedLinearFeatureQuery.java
```diff
@@ -0,0 +1,235 @@
...
+
+            @Override
+            public boolean isCacheable(LeafReaderContext ctx) {
+                return false;
```
navneet1v, line 126: why this is not cachable? can you add comments around this?

src/main/java/org/apache/lucene/BoundedLinearFeatureQuery.java
```diff
@@ -0,0 +1,235 @@
...
+                    score,
+                    "Linear function on the " + fieldName + " field for the " + featureName + " feature, computed as w * S from:",
+                    Explanation.match(boost, "w, weight of this function"),
+                    Explanation.match(featureValue, "S, feature value")
+                );
+            }
+
+            @Override
+            public Scorer scorer(LeafReaderContext context) throws IOException {
+                Terms terms = Terms.getTerms(context.reader(), fieldName);
```
navneet1v, line 160: what will happen in the case when the field with `feildName` is not present

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -51,7 +51,7 @@ public abstract class NLPProcessor extends AbstractProcessor {
 
     private final Environment environment;
 
-    public NLPProcessor(
+    public InferenceProcessor(
```
navneet1v, line 54: check comments related to this in the main branch PR.

src/main/java/org/apache/lucene/BoundedLinearFeatureQuery.java
```diff
@@ -0,0 +1,235 @@
...
+            && Objects.equals(featureName, that.featureName)
+            && Objects.equals(scoreUpperBound, that.scoreUpperBound);
+    }
+
+    @Override
+    public int hashCode() {
+        int h = getClass().hashCode();
+        h = 31 * h + fieldName.hashCode();
+        h = 31 * h + featureName.hashCode();
+        h = 31 * h + scoreUpperBound.hashCode();
```
jmazanec15, line 109: What does 31 do in these cases?

src/main/java/org/apache/lucene/BoundedLinearFeatureQuery.java
jmazanec15: Also, what methods are different from the Lucene FeatureQuery implementation?

zane-neo: This method is the core different: https://github.com/opensearch-project/neural-search/pull/348/files/ddb5d69d16b88da62f864146d9d29252c2d3773e#diff-fa6b12823de991fcb735f2da7e1ad61e3b58829a4291bcfa276e2a1d685b16a6R227.

src/main/java/org/apache/lucene/BoundedLinearFeatureQuery.java
zane-neo: I think this is copied from lucene so not changed this, and they both have exact same functionality.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
model-collapse: Please checkout the comments in 353.

src/main/java/org/apache/lucene/BoundedLinearFeatureQuery.java
zane-neo: This is copied from lucene FeatureQuery, I think we can save the UT effort since this feature will only live a very short time(one version) and the added method is already covered.

src/main/java/org/apache/lucene/BoundedLinearFeatureQuery.java
zane-neo: When creating this new object, the fieldName is required non null.

src/main/java/org/apache/lucene/BoundedLinearFeatureQuery.java
zane-neo: This is also copied code.

src/main/java/org/apache/lucene/BoundedLinearFeatureQuery.java
zane-neo: This is copied code, but the line 109 are changed with our own new parameter. My opinion is 31 is a prime number and it can map the result into a bigger bucket without increasing the possibility of collision, but I think other prime number also can achieve this but maybe 31 is a big enough number for the distribution range.

model-collapse (approved):

PR #353, opened 2023-09-28
Change query clause name to neural_sparse

zane-neo:
### Description
Change query clause name. This doesn't need to be backported to 2.x.

### Issues Resolved

### Check List
    - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
@@ -32,7 +32,7 @@ import com.google.common.collect.ImmutableMap;
  * and set the target fields according to the field name map.
  */
 @Log4j2
-public abstract class NLPProcessor extends AbstractProcessor {
+public abstract class InferenceProcessor extends AbstractProcessor {
 
     public static final String MODEL_ID_FIELD = "model_id";
     public static final String FIELD_MAP_FIELD = "field_map";
@@ -51,7 +51,7 @@ public abstract class NLPProcessor extends AbstractProcessor {
 
     private final Environment environment;
 
-    public NLPProcessor(
+    public InferenceProcessor(
         String tag,
         String description,
         String type,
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -51,7 +51,7 @@ public abstract class NLPProcessor extends AbstractProcessor {
 
     private final Environment environment;
 
-    public NLPProcessor(
+    public InferenceProcessor(
```
navneet1v, line 54: Better naming would be TextInferenceProcessor as we dealing with Texts in these processors.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -55,8 +55,8 @@
 @Accessors(chain = true, fluent = true)
 @NoArgsConstructor
 @AllArgsConstructor
-public class SparseEncodingQueryBuilder extends AbstractQueryBuilder<SparseEncodingQueryBuilder> {
-    public static final String NAME = "sparse_encoding";
+public class NeuralSparseQueryBuilder extends AbstractQueryBuilder<NeuralSparseQueryBuilder> {
+    public static final String NAME = "neural_sparse";
```
navneet1v, line 59: what is `neural` about this query clause?

navneet1v:
@zane-neo lets remove the skip-changelog tag and provide add the info in changelog around what this change is about.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
zane-neo: There are several considerations:
1. This is an abstract parent class and it's better to be generic.
2. Most of the code in this class are to create corresponding data structure to ml-commons which is generic logic.
3. ml-commons is supporting more inference inputs like image(base64), when neural search going to support this, we can create a new class to extend this since this is generic.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -55,8 +55,8 @@
 @Accessors(chain = true, fluent = true)
 @NoArgsConstructor
 @AllArgsConstructor
-public class SparseEncodingQueryBuilder extends AbstractQueryBuilder<SparseEncodingQueryBuilder> {
-    public static final String NAME = "sparse_encoding";
+public class NeuralSparseQueryBuilder extends AbstractQueryBuilder<NeuralSparseQueryBuilder> {
```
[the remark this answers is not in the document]
zane-neo: This is because we call the query clause `neural_sparse`.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zane-neo: We use a neural network model to expand the query text to token/weight, we also had an email to discuss this name and there's no objections so this should be the final decision.

navneet1v:
@zane-neo are you actively working on this PR?

zane-neo:
This is merged in another PR, closing it now

PR #356, opened 2023-09-28
Consumed latest changes from core, use QueryPhaseSearcherWrapper as parent class for Hybrid QPS

martin-gaievski:
### Description
Need to use QueryPhaseSearcherWrapper as this is core new public implementation of the QueryPhaseSearcher. This will be released in core 2.11, came with [this PR](https://github.com/opensearch-project/OpenSearch/pull/7956).

### Check List
- [X] All tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,4 +20,5 @@ Support sparse semantic retrieval by introducing `sparse_encoding` ingest proces
 ### Infrastructure
 ### Documentation
 ### Maintenance
+Consumed latest changes from core, use QueryPhaseSearcherWrapper as parent class for Hybrid QPS ([#356](https://github.com/opensearch-project/neural-search/pull/356))
 ### Refactoring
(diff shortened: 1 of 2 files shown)


vamshin (approved):
LGTM! Thanks

navneet1v:
@martin-gaievski should we have a check with concurrent segment search enabled? what do you think?

martin-gaievski:
> @martin-gaievski should we have a check with concurrent segment search enabled? what do you think?

we definitely need to check that, do you mean check manually or add a test for this?

heemin32:
> @martin-gaievski should we have a check with concurrent segment search enabled? what do you think?

I believe, it should be safe to assume that concurrent search is disabled.

heemin32 (approved):

navneet1v:
> > @martin-gaievski should we have a check with concurrent segment search enabled? what do you think?
> 
> we definitely need to check that, do you mean check manually or add a test for this?

The thing is concurrent search is not GA, I think we can skip for now. But Last when I checked by enabling the concurrent search our tests failed. Now I am hoping when concurrent search will become GA they will ask all plugins to validate.

navneet1v (approved):

Issue #318, opened 2023-09-12
[FEATURE] Support multi-modal semantic search 


vamshin:
### Is your feature request related to a problem?
This feature aims to enhance semantic search use cases by enabling Multimodal support in Neural Search plugin.  Multimodal improves the relevancy of the results by combining text and other forms of inputs like image, audio, video etc. 
As of today OpenSearch supports Semantic search use cases purely based on text based embedding models. While this works for majority of usecases, it does not scale well for applications that need to embed other forms of input. For example, consider a query “Give me bright colored blue shoes”. This can get more relevant results if image properties such as color, intensity is captured along with text. Multimodal solves this problem

### What solution would you like?
- We can make use of [Binary field](http://a%20binary%20field%20type%20contains%20a%20binary%20value%20in%20base64%20encoding%20that%20is%20not%20searchable./) in OpenSearch to accept other forms of inputs like images, audio, video.
- Enhance existing text embedding processor/create new processor to accept the binary fields and pass to ml-commons to get relevant embedding
- Enhance neural query clause to accept binary along with string to perform semantic search

martin-gaievski:
@vamshin for 2.11 we can do only limited support due to tight deadline. I see that is added to the roadmap, do you want to keep it as is and we create a new feature request for a extended solution, or we can remove 2.11 label and keep using this issue.

vamshin:
@martin-gaievski I think we should create new one for tracking extended solutions and tag this issue as a parent issue. This way community knows this feature is already shipped.

PR #359, opened 2023-09-29
Added Multimodal semantic search feature

martin-gaievski:
### Description
Adding support for multimodal search. 

This PR covers phase 1 of the implementation that is a simplified version that supports only models with text and image (text, image or both). 
Main parts involved:
- data ingestion. new ingestion processor created for getting embeddings for documents during ingestion. Interface is:
```
{
  "description": "An example neural search pipeline",
  "processors" : [
    {
      "text-image-embedding": {
        "model_id": "model1234567",
        "embedding": "multimodal_embedding"
        "field_map": {
            "text": "my_text_field", 
            "image": "my_image_field" 
        }
      }
    }
  ]
} 
```
- search query. existing neural-search query enhanced by adding keyword for base64 encoded image. New interface looks like this:
```
{
  "query": {
        "neural": {
              "passage_vector": { 
                  "query_text": "Hello world", 
                  "query_image": "base64forimage_1234567890",
                  "model_id": "xzy76xswsd",
                  "k": 100
                }
           }
        }
    }
}
```

That is done in 2.x to be in sync with ml-commons that are also doings changes, and to be align better with 2.11 release that is coming soon. It will be upported to `main` later.

### Issues Resolved
https://github.com/opensearch-project/neural-search/issues/318
https://github.com/opensearch-project/neural-search/issues/362

### Check List
    - [X] All tests pass


src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,227 @@
...
+            String originalKey = fieldMapEntry.getValue(); // field from ingest document that we need to sent as model input, part of
+                                                           // processor definition
+
+            if (!sourceAndMetadataMap.containsKey(originalKey)) {
+                continue;
+            }
+            if (!(sourceAndMetadataMap.get(originalKey) instanceof String)) {
+                throw new IllegalArgumentException("Unsupported format of the field in the document, value must be a string");
+            }
+            if (((String) sourceAndMetadataMap.get(originalKey)).length() > MAX_CONTENT_LENGTH_IN_BYTES) {
```
navneet1v, line 155: we should not add this check of max content length.

martin-gaievski: I see, that is valid point, we may not know the size model will accept, so should not limit user. For AWS we may rely on network quotas which are even smaller for majority of instances https://docs.aws.amazon.com/opensearch-service/latest/developerguide/limits.html

navneet1v: if this is a mandatory check we can add a parameter in the processor which defines the max size and we can put the default value as the current value which is provided in this PR.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -191,4 +218,41 @@ private List<List<Float>> buildVectorFromResponse(MLOutput mlOutput) {
         return resultMaps;
     }
 
+    private List<Float> buildSingleVectorFromResponse(MLOutput mlOutput) {
+        final List<List<Float>> vector = buildVectorFromResponse(mlOutput);
+        return vector.isEmpty() ? new ArrayList<>() : vector.get(0);
+    }
+
+    private void inferenceSentencesWithRetry(
+        @NonNull final List<String> targetResponseFilters,
```
navneet1v, line 227: lets remove @Nonnull from private functions, null checks should be done on the public functions

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: it's not a must for now, I removed the check, we can add processor param or other way to configure that in next versions

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -191,4 +212,41 @@ private List<List<Float>> buildVectorFromResponse(MLOutput mlOutput) {
...
+            } else {
+                listener.onFailure(e);
+            }
+        }));
+    }
+
+    private MLInput createMLMultimodalInput(final List<String> targetResponseFilters, Map<String, String> input) {
+        List<String> inputText = new ArrayList<>();
+        inputText.add(input.get(INPUT_TEXT));
+        if (input.containsKey(INPUT_IMAGE)) {
```
heemin32, line 245: If there is no key `INPUT_IMAGE` in input, will be be same as text embedding processor or it is still different?
Just wonder if we can generalize it to handle text only, text + image, and text + other forms in a single processor.

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+    public static final String TYPE = "text_image_embedding";
+    public static final String MODEL_ID_FIELD = "model_id";
+    public static final String EMBEDDING_FIELD = "embedding";
+    public static final String FIELD_MAP_FIELD = "field_map";
+    public static final String TEXT_FIELD_NAME = "text";
+    public static final String IMAGE_FIELD_NAME = "image";
+    public static final String INPUT_TEXT = "inputText";
+    public static final String INPUT_IMAGE = "inputImage";
+
+    @VisibleForTesting
```
heemin32, line 45: What is this?

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+    private final String modelId;
+    private final String embedding;
+    private final Map<String, String> fieldMap;
+
+    private final MLCommonsClientAccessor mlCommonsClientAccessor;
+    private final Environment environment;
+
+    public TextImageEmbeddingProcessor(
+        String tag,
```
heemin32, line 54: final

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+        validateEmbeddingConfiguration(fieldMap);
+
+        this.modelId = modelId;
+        this.embedding = embedding;
+        this.fieldMap = fieldMap;
+        this.mlCommonsClientAccessor = clientAccessor;
+        this.environment = environment;
+    }
+
+    private void validateEmbeddingConfiguration(Map<String, String> fieldMap) {
```
heemin32, line 73:
```suggestion
    private void validateEmbeddingConfiguration(final Map<String, String> fieldMap) {
```

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+        if (fieldMap == null
+            || fieldMap.isEmpty()
+            || fieldMap.entrySet()
+                .stream()
+                .anyMatch(x -> StringUtils.isBlank(x.getKey()) || Objects.isNull(x.getValue()) || StringUtils.isBlank(x.getValue()))) {
```
heemin32, line 78: `isBlank` should handle null case
```suggestion
                .anyMatch(x -> StringUtils.isBlank(x.getKey()) || StringUtils.isBlank(x.getValue()))) {
```

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+            throw new IllegalArgumentException("Unable to create the TextImageEmbedding processor as field_map has invalid key or value");
+        }
+
+        if (fieldMap.entrySet().stream().anyMatch(entry -> !Set.of(TEXT_FIELD_NAME, IMAGE_FIELD_NAME).contains(entry.getKey()))) {
```
heemin32, line 82: We can make Set.of (TEXT_FIELD_NAME, IMAGE_FIELD_NAME) as constant.
```suggestion
        if (fieldMap.entrySet().stream().allMatch(entry -> VALID_FIELD_NAME.contains(entry.getKey())) == false) {
```

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+            throw new IllegalArgumentException("Unable to create the TextImageEmbedding processor as field_map has unsupported field name");
+        }
+    }
+
+    @Override
+    public IngestDocument execute(IngestDocument ingestDocument) {
+        return ingestDocument;
```
heemin32, line 89:
```suggestion
        throw new IllegalStateException("Not implemented");
```

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+     * @param ingestDocument {@link IngestDocument} which is the document passed to processor.
+     * @param handler {@link BiConsumer} which is the handler which can be used after the inference task is done.
+     */
+    @Override
+    public void execute(final IngestDocument ingestDocument, final BiConsumer<IngestDocument, Exception> handler) {
+        try {
+            validateEmbeddingFieldsValue(ingestDocument);
+            Map<String, String> knnMap = buildMapWithKnnKeyAndOriginalValue(ingestDocument);
+            Map<String, String> inferenceMap = createInferences(knnMap);
+            if (inferenceMap.isEmpty()) {
```
heemin32, line 104: This is not an error case but valid one?

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+                handler.accept(ingestDocument, null);
+            } else {
+                mlCommonsClientAccessor.inferenceSentences(this.modelId, inferenceMap, ActionListener.wrap(vectors -> {
```
heemin32, line 107: Just FYI. If you make async call, it won't throttle ingestion upfront because it will release ingest thread. That means, ingest request will get accepted but the document conversion will fail due to to many requests. Anyway, I guess it is how other embedding processor is implemented so I guess it won't be consistant if I ask to make a sync call here only.

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+                    handler.accept(ingestDocument, null);
+                }, e -> { handler.accept(null, e); }));
+            }
+        } catch (Exception e) {
+            handler.accept(null, e);
+        }
+
+    }
+
+    private void setVectorFieldsToDocument(IngestDocument ingestDocument, List<Float> vectors) {
```
heemin32, line 118:
```suggestion
    private void setVectorFieldsToDocument(final IngestDocument ingestDocument, final List<Float> vectors) {
```

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+        Objects.requireNonNull(vectors, "embedding failed, inference returns null result!");
+        log.debug("Text embedding result fetched, starting build vector output!");
+        Map<String, Object> textEmbeddingResult = buildTextEmbeddingResult(this.embedding, vectors);
+        textEmbeddingResult.forEach(ingestDocument::setFieldValue);
+    }
+
+    @SuppressWarnings({ "unchecked" })
+    private Map<String, String> createInferences(Map<String, String> knnKeyMap) {
```
heemin32, line 126:
```suggestion
    private Map<String, String> createInferences(final Map<String, String> knnKeyMap) {
```

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+            texts.put(INPUT_TEXT, knnKeyMap.get(fieldMap.get(TEXT_FIELD_NAME)));
+        }
+        if (fieldMap.containsKey(IMAGE_FIELD_NAME) && knnKeyMap.containsKey(fieldMap.get(IMAGE_FIELD_NAME))) {
+            texts.put(INPUT_IMAGE, knnKeyMap.get(fieldMap.get(IMAGE_FIELD_NAME)));
+        }
+        return texts;
+    }
+
+    @VisibleForTesting
+    Map<String, String> buildMapWithKnnKeyAndOriginalValue(IngestDocument ingestDocument) {
```
heemin32, line 138:
```suggestion
    Map<String, String> buildMapWithKnnKeyAndOriginalValue(final IngestDocument ingestDocument) {
```

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+                throw new IllegalArgumentException("Unsupported format of the field in the document, value must be a string");
+            }
+            mapWithKnnKeys.put(originalKey, (String) sourceAndMetadataMap.get(originalKey));
+        }
+        return mapWithKnnKeys;
+    }
+
+    @SuppressWarnings({ "unchecked" })
+    @VisibleForTesting
+    Map<String, Object> buildTextEmbeddingResult(String knnKey, List<Float> modelTensorList) {
```
heemin32, line 158:
```suggestion
    Map<String, Object> buildTextEmbeddingResult(final String knnKey, final List<Float> modelTensorList) {
```

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+        Map<String, Object> result = new LinkedHashMap<>();
+        result.put(knnKey, modelTensorList);
+        return result;
+    }
+
+    private void validateEmbeddingFieldsValue(IngestDocument ingestDocument) {
```
heemin32, line 164:
```suggestion
    private void validateEmbeddingFieldsValue(final IngestDocument ingestDocument) {
```

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+        Map<String, Object> sourceAndMetadataMap = ingestDocument.getSourceAndMetadata();
+        for (Map.Entry<String, String> embeddingFieldsEntry : fieldMap.entrySet()) {
+            Object sourceValue = sourceAndMetadataMap.get(embeddingFieldsEntry.getKey());
+            if (sourceValue != null) {
```
heemin32, line 168: nit. Reduce level of nested block.
```suggestion
            if (sourceValue == null) {
              continue;
            }
```

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+                    throw new IllegalArgumentException("field [" + sourceKey + "] is neither string nor nested type, can not process it");
+                } else if (StringUtils.isBlank(sourceValue.toString())) {
+                    throw new IllegalArgumentException("field [" + sourceKey + "] has empty string value, can not process it");
+                }
+            }
+        }
+    }
+
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    private void validateNestedTypeValue(String sourceKey, Object sourceValue, Supplier<Integer> maxDepthSupplier) {
```
heemin32, line 183:
```suggestion
    private void validateNestedTypeValue(final String sourceKey, final Object sourceValue, final Supplier<Integer> maxDepthSupplier) {
```

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+        int maxDepth = maxDepthSupplier.get();
+        if (maxDepth > MapperService.INDEX_MAPPING_DEPTH_LIMIT_SETTING.get(environment.settings())) {
```
heemin32, line 185: Is the setting dynamic? Reading setting from environment might not get updated dynamically. Need to check.

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+                .forEach(x -> validateNestedTypeValue(sourceKey, x, () -> maxDepth + 1));
+        } else if (!String.class.isAssignableFrom(sourceValue.getClass())) {
+            throw new IllegalArgumentException("map type field [" + sourceKey + "] has non-string type, can not process it");
+        } else if (StringUtils.isBlank(sourceValue.toString())) {
+            throw new IllegalArgumentException("map type field [" + sourceKey + "] has empty string, can not process it");
+        }
+    }
+
+    @SuppressWarnings({ "rawtypes" })
+    private static void validateListTypeValue(String sourceKey, Object sourceValue) {
```
heemin32, line 202: Can be `List<Object>` instead of `Object`
```suggestion
    private static void validateListTypeValue(final String sourceKey, final List<Object> sourceValue) {
```

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+        }
+    }
+
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    private void validateNestedTypeValue(String sourceKey, Object sourceValue, Supplier<Integer> maxDepthSupplier) {
+        int maxDepth = maxDepthSupplier.get();
+        if (maxDepth > MapperService.INDEX_MAPPING_DEPTH_LIMIT_SETTING.get(environment.settings())) {
+            throw new IllegalArgumentException("map type field [" + sourceKey + "] reached max depth limit, can not process it");
+        } else if ((List.class.isAssignableFrom(sourceValue.getClass()))) {
+            validateListTypeValue(sourceKey, sourceValue);
```
heemin32, line 188: With update in `validateListTypeValue` mentioned later
```suggestion
            validateListTypeValue(sourceKey, (List)sourceValue);
```

src/main/java/org/opensearch/neuralsearch/processor/factory/TextImageEmbeddingProcessorFactory.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Factory for text_image embedding ingest processor for ingestion pipeline. Instantiates processor based on user provided input.
+ */
+public class TextImageEmbeddingProcessorFactory implements Factory {
+
+    private final MLCommonsClientAccessor clientAccessor;
+
+    private final Environment environment;
+
+    public TextImageEmbeddingProcessorFactory(MLCommonsClientAccessor clientAccessor, Environment environment) {
```
heemin32, line 31:
```suggestion
    public TextImageEmbeddingProcessorFactory(final MLCommonsClientAccessor clientAccessor, final Environment environment) {
```

src/main/java/org/opensearch/neuralsearch/processor/factory/TextImageEmbeddingProcessorFactory.java
```diff
@@ -0,0 +1,48 @@
...
+        this.clientAccessor = clientAccessor;
+        this.environment = environment;
+    }
+
+    @Override
+    public TextImageEmbeddingProcessor create(
+        Map<String, Factory> registry,
```
heemin32, line 38: final?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -113,13 +116,31 @@ public void inferenceSentencesWithMapResult(
         retryableInferenceSentencesWithMapResult(modelId, inputText, 0, listener);
     }
 
+    /**
+     * Abstraction to call predict function of api of MLClient with provided targetResponse filters. It uses the
+     * custom model provided as modelId and run the {@link FunctionName#TEXT_EMBEDDING}. The return will be sent
+     * using the actionListener which will have a {@link List} of {@link List} of {@link Float} in the order of
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -191,4 +212,41 @@ private List<List<Float>> buildVectorFromResponse(MLOutput mlOutput) {
         return resultMaps;
     }
 
+    private List<Float> buildSingleVectorFromResponse(MLOutput mlOutput) {
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -191,4 +212,41 @@ private List<List<Float>> buildVectorFromResponse(MLOutput mlOutput) {
...
+            if (RetryUtil.shouldRetry(e, retryTime)) {
+                final int retryTimeAdd = retryTime + 1;
+                retryableInferenceSentencesWithSingleVectorResult(targetResponseFilters, modelId, inputObjects, retryTimeAdd, listener);
+            } else {
+                listener.onFailure(e);
+            }
+        }));
+    }
+
+    private MLInput createMLMultimodalInput(final List<String> targetResponseFilters, Map<String, String> input) {
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
martin-gaievski: it's same. From my understanding project management is leaning to opposite idea of having specialized processors, so in case we'll be adding more types of content it's ok to keep specialized processors and add new one that are more generic

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: oops, not needed, will remove it

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: I'm not sure that is required, in case of processor with current implementation we just let the even flow, but with exception this will fail the ingestion/data upload. I checked it's same implementation in existing TextEmbeddingProcessor

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: yes, this happens in case mapping is configured in processor but the document that user uploading does not have those fields. In such case embedding field for that document will be empty, and that doc will not appear in the search results

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: yes, that's dynamic

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/factory/TextImageEmbeddingProcessorFactory.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: ack

navneet1v:
@martin-gaievski Haven't reviewed the code completely. Will try to do it by tonight

src/main/java/org/opensearch/neuralsearch/processor/factory/TextImageEmbeddingProcessorFactory.java
martin-gaievski: ack

build.gradle
```diff
@@ -151,6 +151,7 @@ dependencies {
     runtimeOnly group: 'org.reflections', name: 'reflections', version: '0.9.12'
     runtimeOnly group: 'org.javassist', name: 'javassist', version: '3.29.2-GA'
     runtimeOnly group: 'org.opensearch', name: 'common-utils', version: "${opensearch_build}"
+    runtimeOnly group: 'org.apache.commons', name: 'commons-text', version: '1.10.0'
```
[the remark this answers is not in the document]
martin-gaievski: yes, that's needed when we do call ml-commons client as otherwise we will fail in runtime

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,218 @@
...
+        if (fieldMap == null
+            || fieldMap.isEmpty()
+            || fieldMap.entrySet()
+                .stream()
+                .anyMatch(x -> StringUtils.isBlank(x.getKey()) || Objects.isNull(x.getValue()) || StringUtils.isBlank(x.getValue()))) {
+            throw new IllegalArgumentException("Unable to create the TextImageEmbedding processor as field_map has invalid key or value");
+        }
+
+        if (fieldMap.entrySet().stream().anyMatch(entry -> !Set.of(TEXT_FIELD_NAME, IMAGE_FIELD_NAME).contains(entry.getKey()))) {
+            throw new IllegalArgumentException("Unable to create the TextImageEmbedding processor as field_map has unsupported field name");
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/test/java/org/opensearch/neuralsearch/common/BaseNeuralSearchIT.java
```diff
@@ -65,14 +65,18 @@ public abstract class BaseNeuralSearchIT extends OpenSearchSecureRestTestCase {
...
-    protected String PIPELINE_CONFIGURATION_NAME = "processor/PipelineConfiguration.json";
+    protected static final Map<ProcessorType, String> PIPELINE_CONFIGS_BY_TYPE = Map.of(
+        ProcessorType.TEXT_EMBEDDING,
+        "processor/PipelineConfiguration.json",
+        ProcessorType.SPARSE_ENCODING,
+        "processor/SparseEncodingPipelineConfiguration.json",
+        ProcessorType.TEXT_IMAGE_EMBEDDING,
+        "processor/PipelineForTextImageEmbeddingProcessorConfiguration.json"
+    );
+    // protected String PIPELINE_CONFIGURATION_NAME = "processor/PipelineConfiguration.json";
```
navneet1v, line 76: remove this

navneet1v:
@martin-gaievski overall code looks good to me. Lets make sure that we add enough tests to make the github action pass.

This file seems missing UTs. TextImageEmbeddingProcessor

src/test/java/org/opensearch/neuralsearch/common/BaseNeuralSearchIT.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -191,4 +211,41 @@ private List<List<Float>> buildVectorFromResponse(MLOutput mlOutput) {
...
+        final List<String> targetResponseFilters,
+        final String modelId,
+        final Map<String, String> inputObjects,
+        final int retryTime,
+        final ActionListener<List<Float>> listener
+    ) {
+        MLInput mlInput = createMLMultimodalInput(targetResponseFilters, inputObjects);
+        mlClient.predict(modelId, mlInput, ActionListener.wrap(mlOutput -> {
+            final List<Float> vector = buildSingleVectorFromResponse(mlOutput);
+            log.debug("Inference Response for input sentence {} is : {} ", inputObjects, vector);
```
jmazanec15, line 229: Brought this up in another review, but we shouldnt log inputObjects I dont think if its user input.

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,225 @@
...
+@Log4j2
+public class TextImageEmbeddingProcessor extends AbstractProcessor {
+
+    public static final String TYPE = "text_image_embedding";
+    public static final String MODEL_ID_FIELD = "model_id";
+    public static final String EMBEDDING_FIELD = "embedding";
+    public static final String FIELD_MAP_FIELD = "field_map";
+    public static final String TEXT_FIELD_NAME = "text";
+    public static final String IMAGE_FIELD_NAME = "image";
+    public static final String INPUT_TEXT = "inputText";
```
jmazanec15, line 43: Why camelcase?

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -0,0 +1,225 @@
...
+import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
+
+import com.google.common.annotations.VisibleForTesting;
+
+/**
+ * This processor is used for user input data text and image embedding processing, model_id can be used to indicate which model user use,
+ * and field_map can be used to indicate which fields needs embedding and the corresponding keys for the embedding results.
+ */
+@Log4j2
+public class TextImageEmbeddingProcessor extends AbstractProcessor {
```
jmazanec15, line 35: Can we extend NLPProcessor?

src/test/java/org/opensearch/neuralsearch/processor/NormalizationProcessorIT.java
```diff
@@ -94,7 +94,15 @@ public void testResultProcessor_whenOneShardAndQueryMatches_thenSuccessful() {
         createSearchPipelineWithResultsPostProcessor(SEARCH_PIPELINE);
         String modelId = getDeployedModelId();
 
-        NeuralQueryBuilder neuralQueryBuilder = new NeuralQueryBuilder(TEST_KNN_VECTOR_FIELD_NAME_1, "", modelId, 5, null, null);
+        NeuralQueryBuilder neuralQueryBuilder = new NeuralQueryBuilder(
```
jmazanec15, line 97: Do we need to change this in this PR?

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: This should be mapped by ml-commons in model connector to a model input, they expecting these field names

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: I was checking that, but now NLPProcessor is restricted to a certain input and output format that are stick to a text embeddings. We should refactor that to a more generic code so other processors can re-use it, but per my estimations it goes beyond the 2.11 release timeline.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
martin-gaievski: what is the concern with logging that? We definitely don't want to return user input back as that raises serious security concerns, but here we just logging that, and at the `debug` level. It may help us in troubleshooting.

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
heemin32: Could you verify that dynamic update is actually getting applied to this processor? I think you should read setting from `ClusterService`?

src/test/java/org/opensearch/neuralsearch/processor/NormalizationProcessorIT.java
martin-gaievski: yes, we're changing a constructor by adding `image` field. Not directly related to functionality, but it will fail otherwise.

navneet1v (approved):

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
jmazanec15: This could lead to unencrypted user data in log files.

martin-gaievski: oh, that's a valid point, I'll remove user input

src/test/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessorIT.java
```diff
@@ -0,0 +1,110 @@
...
+import org.junit.After;
+import org.opensearch.client.Response;
+import org.opensearch.common.xcontent.XContentHelper;
+import org.opensearch.common.xcontent.XContentType;
+import org.opensearch.neuralsearch.common.BaseNeuralSearchIT;
+
+import com.google.common.collect.ImmutableList;
+
+/**
+ * Testing text_and_image_embedding ingest processor. We can only test text in integ tests, none of pre-built models
```
jmazanec15, line 26: How are we confirming that this works?

martin-gaievski: We have tested that offline by setting up one of [BedRock titan](https://aws.amazon.com/bedrock/titan/) models and registering it using [remote connector feature](https://opensearch.org/docs/latest/ml-commons-plugin/extensibility/connectors/) in ml-commons.

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
martin-gaievski: Good catch, with current code we are not getting dynamic updates for settings. Just pushed a commit to address that, we're using ClusterService to read than in runtime for each query

heemin32 (approved):

navneet1v (approved):

PR #404, opened 2023-10-06
Fixed error for case when mltensor has data as null

martin-gaievski:
### Description
Found that we're not handling properly case when MLTensor has data == `null` from model/connector on `predict` call. Need to throw proper exception in that case

### Check List
- [X] All tests pass


heemin32 (approved):

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -40,6 +42,7 @@
 public class MLCommonsClientAccessor {
     private static final List<String> TARGET_RESPONSE_FILTERS = List.of("sentence_embedding");
     private final MachineLearningNodeClient mlClient;
+    private static final String EXCEPTION_MESSAGE_MODEL_PROCESSING_FAILED = "the system encountered an unexpected error during processing";
```
navneet1v, line 45: we need to fix this, lets make it more informative.

navneet1v (approved):

PR #438, opened 2023-10-11
[bug fix] Fix async actions are left in neural_sparse query

zhichao-aws:
### Description
The bug exists in multi-nodes and multi-shards environment. For neural_search query, the cluster will throw error "async actions are left".

The search action contains 2 phase, query and fetch. In multi nodes & multi shards case, nodes will finish the query phase normally. But there will be a serialization and deserialization at the start of fetch shard phase, and after deserialization the doRewrite method of constructed QueryBuilder(i.e. NeuralSparseQueryBuilder) will be called. However for this rewrite the QueryBuilder can not have async action otherwise an exception will be thrown.

In our current implementation, the model inference result is not included in the serialization/deserialization process, and it will register an async action when rewritten. And this will break the search.

Add queryTokensSupplier to the process of serialization/deserialization, as well as doEquals and doHashCode

### Issues Resolved

### Check List
    - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
@@ -89,6 +89,10 @@ public class NeuralSparseQueryBuilder extends AbstractQueryBuilder<NeuralSparseQ
         this.queryText = in.readString();
         this.modelId = in.readString();
         this.maxTokenScore = in.readOptionalFloat();
+        if (in.readBoolean()) {
+            Map<String, Float> queryTokens = in.readMap(StreamInput::readString, StreamInput::readFloat);
+            this.queryTokensSupplier = () -> queryTokens;
+        }
     }
 
     @Override
@@ -97,6 +101,12 @@ public class NeuralSparseQueryBuilder extends AbstractQueryBuilder<NeuralSparseQ
         out.writeString(queryText);
         out.writeString(modelId);
         out.writeOptionalFloat(maxTokenScore);
+        if (queryTokensSupplier != null && queryTokensSupplier.get() != null) {
+            out.writeBoolean(true);
+            out.writeMap(queryTokensSupplier.get(), StreamOutput::writeString, StreamOutput::writeFloat);
+        } else {
+            out.writeBoolean(false);
+        }
     }
 
     @Override
--- a/src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
@@ -6,7 +6,6 @@
 package org.opensearch.neuralsearch.query;
 
 import java.io.IOException;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
--- a/src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
@@ -6,6 +6,7 @@
 package org.opensearch.neuralsearch.query;
 
 import java.io.IOException;
+import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


zhichao-aws:
should be in 2.11

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -276,16 +286,25 @@ private static void validateQueryTokens(Map<String, Float> queryTokens) {
     protected boolean doEquals(NeuralSparseQueryBuilder obj) {
         if (this == obj) return true;
         if (obj == null || getClass() != obj.getClass()) return false;
+        if (queryTokensSupplier == null && obj.queryTokensSupplier != null) return false;
```
martin-gaievski, line 289: nit: we need to change formatting in this method, every if case body must be in curly braces, plus for null/not null checks we can utilize Objects static methods. I'm fine with doing this is future, not a blocker for this PR

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilderTests.java
```diff
@@ -294,6 +295,23 @@ public void testStreams() {
 
         NeuralSparseQueryBuilder copy = new NeuralSparseQueryBuilder(filterStreamInput);
         assertEquals(original, copy);
+
+        SetOnce<Map<String, Float>> queryTokensSetOnce = new SetOnce<>();
+        queryTokensSetOnce.set(Map.of("hello", 1.0f, "world", 2.0f));
+        original.queryTokensSupplier(queryTokensSetOnce::get);
+
+        streamOutput = new BytesStreamOutput();
```
martin-gaievski, line 303: can we create a new instance of ByteStreamOutput instead of re-using existent one. That improves readability and lower chance of error. Same for line 306, instance of NamedWriteableAwareStreamInput

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -89,6 +89,10 @@ public NeuralSparseQueryBuilder(StreamInput in) throws IOException {
         this.queryText = in.readString();
         this.modelId = in.readString();
         this.maxTokenScore = in.readOptionalFloat();
+        if (in.readBoolean()) {
```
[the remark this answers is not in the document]
martin-gaievski: to clarify my previous comment - that can be not backward compatible in terms of a rolling cluster upgrades. With that different cluster nodes may run different versions, say cluster is now on 2.8 and user wants to do rolling upgrade to 2.11 (some [general documentation on this](https://opensearch.org/docs/latest/install-and-configure/upgrade-opensearch/rolling-upgrade/)). 
In such case query may be serialized at 2.8 node (without new field at all) and deserialized an a 2.11 node. If we just try to read the boolean there will be a runtime exception. We need to check minimal supported version before even reading boolean flag. 
The reverse scenario is also possible: query serialized on 2.11 node and read on 2.8 node, in such case we shouldn't write to the stream.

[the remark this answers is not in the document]
heemin32: This class is newly introduced in 2.11. No issue with backward compatibility.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
heemin32: Let's handle the issue when we backport it to 2.x

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilderTests.java
heemin32: Let's handle the issue when we backport it to 2.x or as a follow up pr.

martin-gaievski (approved):
Looks good to me, BWC is not an issue as it's in new class, for others we can take care of them in next PRs

Issue #386, opened 2023-10-04
Add integration test for reindex 


ylwu-amzn:
We found some bug when reindex a non-KNN index to KNN index with neural-search ingest pipeline.  We fixed this issue https://github.com/opensearch-project/ml-commons/pull/1418.  

It will be best if neural-search can have some integration test to cover reindex. So we can catch such issue earlier.

heemin32:
The bug was in ml-commons. Ideally, test should be in ml-common imo. We don't expect every ml-common dependency plugin to add reindex integ test to catch a bug in ml-common?

ylwu-amzn:
It will be challenging for ml-commons , as run reindex needs Knn and neural-search plugins

heemin32:
Understand the difficulty to have reindex test in ml-commons but in a long term, we should consider implementing some very basic ingest pipeline inside ml-commons for integ test purpose.

navneet1v:
@heemin32 I would beg to differ little bit here. As a ingest processor is defined in neural search plugin, ideally it is the job of the neural search plugin to validate all the cases in which that processor can be run. Now running the processor on re-index is another way to calling that processor. Hence the integration tests should be in neural search for re-index, because ML Commons cannot cover all the cases in which how the processor can be invoked.

Having said all the things above, one thing we definitely need from ML Commons Plugin or MLCommons client side, in terms of security how the apis can be used. This will ensure that its not the clients who are letting MLCommons know that something is broken.

heemin32:
> Having said all the things above, one thing we definitely need from ML Commons Plugin or MLCommons client side, in terms of security how the apis can be used. This will ensure that its not the clients who are letting MLCommons know that something is broken.

My point align with this comment. MLCommons should be able to find if something is broken without relying on integ test in other plugins. Neural search should have re-index integ test to increase test coverage but MLCommons also should have a test case to cover the bug regarding this PR https://github.com/opensearch-project/ml-commons/pull/1418.

navneet1v:
@heemin32 

> implementing some very basic ingest pipeline inside ml-commons for integ test purpose.

My comment was mainly around this.

heemin32:
> @heemin32
> 
> > implementing some very basic ingest pipeline inside ml-commons for integ test purpose.
> 
> My comment was mainly around this.

Just one of option to catch the bug inside ml-common itself.

q-andy:
Hi, could you assign this to me?

PR #446, opened 2023-10-13
Add reindex ITs

zane-neo:
### Description
Add reindex ITs in neural search

### Issues Resolved
Issue: https://github.com/opensearch-project/neural-search/issues/386

### Check List
    - [ ] All tests pass

### DIFF
--- a/src/test/java/org/opensearch/neuralsearch/common/BaseNeuralSearchIT.java
+++ b/src/test/java/org/opensearch/neuralsearch/common/BaseNeuralSearchIT.java
@@ -772,6 +772,60 @@ public abstract class BaseNeuralSearchIT extends OpenSearchSecureRestTestCase {
         return modelGroupId;
     }
 
+    protected void createIndexWithPipeline(String indexName, String indexMappingFileName, String pipelineName) throws Exception {
+        createIndexWithConfiguration(
+            indexName,
+            Files.readString(Path.of(classLoader.getResource("processor/" + indexMappingFileName).toURI())),
+            pipelineName
+        );
+    }
+
+    /**
+     *  Ingest a document to index.
+     * @param indexName
+     * @param ingestDocument
+     * @throws Exception
+     */
+    protected String ingestDocument(String indexName, String ingestDocument) throws Exception {
+        Response response = makeRequest(
+            client(),
+            "POST",
+            indexName + "/_doc?refresh",
+            null,
+            toHttpEntity(ingestDocument),
+            ImmutableList.of(new BasicHeader(HttpHeaders.USER_AGENT, "Kibana"))
+        );
+        Map<String, Object> map = XContentHelper.convertToMap(
+            XContentType.JSON.xContent(),
+            EntityUtils.toString(response.getEntity()),
+            false
+        );
+        return (String) map.get("result");
+    }
+
+    /**
+     * Reindex from one index to another
+     * @param fromIndexName
+     * @param toIndexName
+     * @throws Exception
+     */
+    protected void reindex(String fromIndexName, String toIndexName) throws Exception {
+        Response response = makeRequest(
+            client(),
+            "POST",
+            "/_reindex?refresh",
+            null,
+            toHttpEntity("{\"source\":{\"index\":\""+ fromIndexName +"\"},\"dest\":{\"index\":\"" + toIndexName + "\"}}"),
+            ImmutableList.of(new BasicHeader(HttpHeaders.USER_AGENT, "Kibana"))
+        );
+        Map<String, Object> map = XContentHelper.convertToMap(
+            XContentType.JSON.xContent(),
+            EntityUtils.toString(response.getEntity()),
+            false
+        );
+        assertEquals(0, ((List)map.get("failures")).size());
+    }
+
     /**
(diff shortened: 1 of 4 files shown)


navneet1v:
@zane-neo can we add reindex IT for other processors too?

navneet1v:
@zane-neo any update on this PR?

zane-neo:
> @zane-neo can we add reindex IT for other processors too?

The issue in ml-commons is not specific to individual processor, it's more like a generic issue(user info not restored after the listener ran), so one or two processor to cover this is enough I thought.

navneet1v:
> > @zane-neo can we add reindex IT for other processors too?
> 
> The issue in ml-commons is not specific to individual processor, it's more like a generic issue(user info not restored after the listener ran), so one or two processor to cover this is enough I thought.

I won't agree with this. It will be good if we have reindex ITs for atleast all the local model like splade and text embeddings. Text + image processor works on remote model so we cannot integrate here.

zane-neo:
> > > @zane-neo can we add reindex IT for other processors too?
> > 
> > 
> > The issue in ml-commons is not specific to individual processor, it's more like a generic issue(user info not restored after the listener ran), so one or two processor to cover this is enough I thought.
> 
> I won't agree with this. It will be good if we have reindex ITs for atleast all the local model like splade and text embeddings. Text + image processor works on remote model so we cannot integrate here.

This PR already includes splade and text embedding processors.

martin-gaievski:
Using Text_image_embeddings processor with only text field should work with local model. Processor uses mlClient to call the model, and model type is abstracted by the client.

zane-neo:
> Using Text_image_embeddings processor with only text field should work with local model. Processor uses mlClient to call the model, and model type is abstracted by the client.

it's a text embedding case right?

martin-gaievski:
> > Using Text_image_embeddings processor with only text field should work with local model. Processor uses mlClient to call the model, and model type is abstracted by the client.
> 
> it's a text embedding case right?

I think it's not. Embeddings are only text, that's correct, but the processor is text_image. That processor can work with text, image and text+image. If we want to test that reindexing works with all processors it would not be correct to test it with text_embedding and assume that text_image_embedding processor is also tested.

zane-neo:
> > > Using Text_image_embeddings processor with only text field should work with local model. Processor uses mlClient to call the model, and model type is abstracted by the client.
> > 
> > 
> > it's a text embedding case right?
> 
> I think it's not. Embeddings are only text, that's correct, but the processor is text_image. That processor can work with text, image and text+image. If we want to test that reindexing works with all processors it would not be correct to test it with text_embedding and assume that text_image_embedding processor is also tested.

current tests covered local models which can prove the correctness of ml-commons code, no need to add other cases IMO.

navneet1v:
> current tests covered local models which can prove the correctness of ml-commons code, no need to add other cases IMO.

This is not the sole purpose of these tests. If the purpose is for correctness of ML code, the tests should be in MLCommons. 

We want to cover as many use cases as possible in which an ingestion processor can be used. Re-indexing is one of the use-case here. I know we started this PR as part of making sure that bug we got here: https://github.com/opensearch-project/neural-search/issues/386 can be caught in future, but that is not the sole purpose of the tests.

zane-neo:
> > current tests covered local models which can prove the correctness of ml-commons code, no need to add other cases IMO.
> 
> This is not the sole purpose of these tests. If the purpose is for correctness of ML code, the tests should be in MLCommons.
> 
> We want to cover as many use cases as possible in which an ingestion processor can be used. Re-indexing is one of the use-case here. I know we started this PR as part of making sure that bug we got here: #386 can be caught in future, but that is not the sole purpose of the tests.

There're quite a lot different flows/features in OpenSearch, e.g. shrink index, split index, restore index etc. Do we have a way to cover all these features when implementing a new feature to ensure they're not breaking? Is covering only reindex enough?

navneet1v:
> Do we have a way to cover all these features when implementing a new feature to ensure they're not breaking? Is covering only reindex enough?

Given the issue in hand we should start with re-index and then add more case which you provided after this. We can cut a github issue to make sure that we are covering the cases you mentioned too.

zane-neo:
> > Do we have a way to cover all these features when implementing a new feature to ensure they're not breaking? Is covering only reindex enough?
> 
> Given the issue in hand we should start with re-index and then add more case which you provided after this. We can cut a github issue to make sure that we are covering the cases you mentioned too.

Is there already a generic approach to address this? Like which can be used for all plugins so that no need to implement this in individual plugin separately?

navneet1v:
> > > Do we have a way to cover all these features when implementing a new feature to ensure they're not breaking? Is covering only reindex enough?
> > 
> > 
> > Given the issue in hand we should start with re-index and then add more case which you provided after this. We can cut a github issue to make sure that we are covering the cases you mentioned too.
> 
> Is there already a generic approach to address this? Like which can be used for all plugins so that no need to implement this in individual plugin separately?

As per my understanding there is no such thing. Main thing which I was trying to put here was, you are adding ITs only for sparse and text embedding processor. There are other processors too in the plugin. So it will be better if you can add re-index ITs for those processors too.

zane-neo:
Will do more test on open source part to reproduce this issue, will reopen this or create a new PR to this. Closing this now.

Issue #376, opened 2023-10-02
[FEATURE] Add z-score for the normalization processor


samuel-oci:
### Is your feature request related to a problem?
_Currently the normalization processor doesn't support z-score which is a popular technique and according in some instances produces superior results to min-max  see [blog](https://buff.ly/3EXazOb)_

### What solution would you like?
_Allow to specify z-score as a normalization technique in the [normalization processor](https://opensearch.org/docs/latest/search-plugins/search-pipelines/normalization-processor/)_

### What alternatives have you considered?
_Not much at the moment but suggestions are welcome._

### Do you have any additional context?
_see [blog](https://buff.ly/3EXazOb) mentioned earlier_

heemin32:
@samuel-oci Thanks for the feature request. Also welcome to take it and work on it.

samuel-oci:
Sounds good, I'll take it, feel free to assign to me.

samuel-oci:
Hi @heemin32 @vamshin @navneet1v I don't see any IT that tests normalization and was thinking to add those also as part of this PR. If I missed those or you already have those worked on somewhere else just let me know and we can consolidate.

navneet1v:
@martin-gaievski can you please provide the reference for the integration tests that has been added.

samuel-oci:
@navneet1v this one test hybrid but for some reason when I run it in my IDE to debug I can't see the normalization processor being triggered. Could also be something wrong with my setup.. if it is then feel free to ignore.

navneet1v:
> @navneet1v this one test hybrid but for some reason when I run it in my IDE to debug I can't see the normalization processor being triggered. Could also be something wrong with my setup.. if it is then feel free to ignore.

If you are using debugger it might not hit the code. I would recommend adding logs and check it. Debugger is tricky with IT tests, if you are just running `./gradlew integTest`

samuel-oci:
@navneet1v I added the following lines `build.gradle` under `IntegTest` task and also created a corresponding log4j2-test.xml:
```
    systemProperty 'log4j2.configurationFile', "${projectDir}/src/test/resources/log4j2-test.xml"

    // Set this to true this if you want to see the logs in the terminal test output.
    // note: if left false the log output will still show in your IDE
    testLogging.showStandardStreams = true
``` 

I do see logs of IT test and other core open search but don't see any logs showing from the normalization processor or NormalizationWorkflow. which is why I later run with debugger to double check.
[...]

PR #468, opened 2023-10-19
[FEATURE] Add z-score for the normalization processor #376

samuel-oci:
### Description
This change implements [https://github.com/opensearch-project/neural-search/issues/376](https://github.com/opensearch-project/neural-search/issues/376)
- Add z-score for hybrid query normalization processor
- Add IT that test normalization end to end

### Issues Resolved
Resolving [https://github.com/opensearch-project/neural-search/issues/376](https://github.com/opensearch-project/neural-search/issues/376)

### Check List
- [x ] New functionality includes testing.
    - [ x] All tests pass
- [x ] New functionality has been documented.
    - [ x] New functionality has javadoc added
- [ x] Commits are signed as per the DCO using --signoff

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
@@ -26,8 +26,6 @@ import org.opensearch.neuralsearch.processor.CompoundTopDocs;
 @Log4j2
 public class ScoreCombiner {
 
-    private static final Float ZERO_SCORE = 0.0f;
-
     /**
      * Performs score combination based on input combination technique. Mutates input object by updating combined scores
      * Main steps we're doing for combination:
--- a/src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
@@ -5,15 +5,18 @@
 
 package org.opensearch.neuralsearch.processor.normalization;
 
-import com.google.common.primitives.Floats;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
 import lombok.ToString;
+
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TopDocs;
 import org.opensearch.neuralsearch.processor.CompoundTopDocs;
 
-import java.util.Arrays;
-import java.util.List;
-import java.util.Objects;
+import com.google.common.primitives.Floats;
 
 /**
  * Implementation of z-score normalization technique for hybrid query
@@ -24,24 +27,26 @@ import java.util.Objects;
 TODO: Some todo items that apply here but also on the original normalization techniques on which it is modeled {@link L2ScoreNormalizationTechnique} and {@link MinMaxScoreNormalizationTechnique}
 1. Random access to abstract list object is a bad practice both stylistically and from performance perspective and should be removed
 2. Identical sub queries and their distribution between shards is currently completely implicit based on ordering and should be explicit based on identifier
-3. Weird calculation of numOfSubQueries instead of having a more explicit indicator
+3. Implicit calculation of numOfSubQueries instead of having a more explicit upstream indicator/metadata regarding it
  */
 @ToString(onlyExplicitlyIncluded = true)
 public class ZScoreNormalizationTechnique implements ScoreNormalizationTechnique {
     @ToString.Include
     public static final String TECHNIQUE_NAME = "z_score";
     private static final float SINGLE_RESULT_SCORE = 1.0f;
+
     @Override
-    public void normalize(List<CompoundTopDocs> queryTopDocs) {
-        // why are we doing that? is List<CompoundTopDocs> the list of subqueries for a single shard? or a global list of all subqueries across shards?
-        // If a subquery comes from each shard then when is it combined? that seems weird that combination will do combination of normalized results that each is normalized just based on shard level result
-        int numOfSubQueries = queryTopDocs.stream()
-                .filter(Objects::nonNull)
-                .filter(topDocs -> topDocs.getTopDocs().size() > 0)
-                .findAny()
-                .get()
-                .getTopDocs()
-                .size();
+    public void normalize(final List<CompoundTopDocs> queryTopDocs) {
+        /*
+           TODO: There is an implicit assumption in this calculation that probably need to be made clearer by passing some metadata with the results.
+           Currently assuming that finding a single non empty shard result will contain all sub query results with 0 hits.
+         */
+        final Optional<CompoundTopDocs> maybeCompoundTopDocs = queryTopDocs.stream()
+            .filter(Objects::nonNull)
+            .filter(topDocs -> topDocs.getTopDocs().size() > 0)
+            .findAny();
+
+        final int numOfSubQueries = maybeCompoundTopDocs.map(compoundTopDocs -> compoundTopDocs.getTopDocs().size()).orElse(0);
 
         // to be done for each subquery
         float[] sumPerSubquery = findScoreSumPerSubQuery(queryTopDocs, numOfSubQueries);
@@ -67,9 +72,7 @@ public class ZScoreNormalizationTechnique implements ScoreNormalizationTechnique
     static private float[] findScoreSumPerSubQuery(final List<CompoundTopDocs> queryTopDocs, final int numOfScores) {
         final float[] sumOfScorePerSubQuery = new float[numOfScores];
         Arrays.fill(sumOfScorePerSubQuery, 0);
-        //TODO: make this better, currently
-        // this is a horrible implementation in particular when it comes to the topDocsPerSubQuery.get(j)
-        // which does a random search on an abstract list type.
+        // TODO: make this syntactically clearer regarding performance by avoiding List.get(j) with an abstract List type
         for (CompoundTopDocs compoundQueryTopDocs : queryTopDocs) {
             if (Objects.isNull(compoundQueryTopDocs)) {
                 continue;
@@ -86,9 +89,7 @@ public class ZScoreNormalizationTechnique implements ScoreNormalizationTechnique
     static private long[] findNumberOfElementsPerSubQuery(final List<CompoundTopDocs> queryTopDocs, final int numOfScores) {
         final long[] numberOfElementsPerSubQuery = new long[numOfScores];
         Arrays.fill(numberOfElementsPerSubQuery, 0);
-        //TODO: make this better, currently
-        // this is a horrible implementation in particular when it comes to the topDocsPerSubQuery.get(j)
-        // which does a random search on an abstract list type.
+        // TODO: make this syntactically clearer regarding performance by avoiding List.get(j) with an abstract List type
         for (CompoundTopDocs compoundQueryTopDocs : queryTopDocs) {
             if (Objects.isNull(compoundQueryTopDocs)) {
                 continue;
@@ -108,21 +109,22 @@ public class ZScoreNormalizationTechnique implements ScoreNormalizationTechnique
             if (elementsPerSubquery[i] == 0) {
                 meanPerSubQuery[i] = 0;
             } else {
-                meanPerSubQuery[i] = sumPerSubquery[i]/elementsPerSubquery[i];
+                meanPerSubQuery[i] = sumPerSubquery[i] / elementsPerSubquery[i];
             }
         }
 
         return meanPerSubQuery;
     }
 
-    static private float[] findStdPerSubquery(final List<CompoundTopDocs> queryTopDocs, final float[] meanPerSubQuery, final long[] elementsPerSubquery, final int numOfScores) {
+    static private float[] findStdPerSubquery(
+        final List<CompoundTopDocs> queryTopDocs,
+        final float[] meanPerSubQuery,
+        final long[] elementsPerSubquery,
+        final int numOfScores
+    ) {
         final double[] deltaSumPerSubquery = new double[numOfScores];
         Arrays.fill(deltaSumPerSubquery, 0);
-
-
-        //TODO: make this better, currently
-        // this is a horrible implementation in particular when it comes to the topDocsPerSubQuery.get(j)
-        // which does a random search on an abstract list type.
+        // TODO: make this syntactically clearer regarding performance by avoiding List.get(j) with an abstract List type
         for (CompoundTopDocs compoundQueryTopDocs : queryTopDocs) {
             if (Objects.isNull(compoundQueryTopDocs)) {
                 continue;
@@ -147,7 +149,7 @@ public class ZScoreNormalizationTechnique implements ScoreNormalizationTechnique
         return stdPerSubQuery;
     }
 
-    static private float sumScoreDocsArray(ScoreDoc[] scoreDocs) {
+    static private float sumScoreDocsArray(final ScoreDoc[] scoreDocs) {
         float sum = 0;
         for (ScoreDoc scoreDoc : scoreDocs) {
             sum += scoreDoc.score;
@@ -161,6 +163,6 @@ public class ZScoreNormalizationTechnique implements ScoreNormalizationTechnique
         if (Floats.compare(mean, score) == 0) {
             return SINGLE_RESULT_SCORE;
         }
-        return  (score - mean) / standardDeviation;
+        return (score - mean) / standardDeviation;
     }
 }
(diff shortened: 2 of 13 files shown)


navneet1v:
@samuel-oci thanks for creating the PR. But given that this is a new feature, the recommendation is to put this code in feature branch. This is to ensure that main branch is not blocked.

I will go ahead and create a feature branch for this change. 

As you have already written down the full code, its a good time to start doing the performance testing and search relevancy testing for this feature.

navneet1v:
Created a feature branch from main, for this feature: feature/z-score-normalization

Please raise the PR against that branch. Also can you add a entry in the CHANGELOG.md file for this change.

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
```diff
@@ -0,0 +1,166 @@
...
+import java.util.Objects;
+
+/**
+ * Implementation of z-score normalization technique for hybrid query
+ * This is currently modeled based on the existing normalization techniques {@link L2ScoreNormalizationTechnique} and {@link MinMaxScoreNormalizationTechnique}
+ * However, this class as well as the original ones require a significant work to improve style and ease of use, see TODO items below
+ */
+/*
+TODO: Some todo items that apply here but also on the original normalization techniques on which it is modeled {@link L2ScoreNormalizationTechnique} and {@link MinMaxScoreNormalizationTechnique}
+1. Random access to abstract list object is a bad practice both stylistically and from performance perspective and should be removed
```
navneet1v, line 25: Can you please provide an alternative what should be used? 

As per my understanding, random access on the List is bad if List concrete implementation is LinkedList. But what I have seen generally is we use ArrayList which is backed by arrays, hence random access is done in constant time.

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
```diff
@@ -0,0 +1,166 @@
...
+2. Identical sub queries and their distribution between shards is currently completely implicit based on ordering and should be explicit based on identifier
```
navneet1v, line 26: This is really a good thought, but problem is none of the query clauses in Opensearch supports identifiers. During the implementation this was discussed. The problem is the way after QueryPhase the results are returned. They are returned in a ScoreDocs array which doesn't support identifiers.

We can go around that but it will require changes in interface of OpenSearch Core. Hence we decided against it to make sure that we are compatible with OpenSearch core.

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
```diff
@@ -0,0 +1,166 @@
...
+3. Weird calculation of numOfSubQueries instead of having a more explicit indicator
```
navneet1v, line 27: same as above.

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
```diff
@@ -0,0 +1,166 @@
...
+ */
+@ToString(onlyExplicitlyIncluded = true)
+public class ZScoreNormalizationTechnique implements ScoreNormalizationTechnique {
+    @ToString.Include
+    public static final String TECHNIQUE_NAME = "z_score";
+    private static final float SINGLE_RESULT_SCORE = 1.0f;
+    @Override
+    public void normalize(List<CompoundTopDocs> queryTopDocs) {
+        // why are we doing that? is List<CompoundTopDocs> the list of subqueries for a single shard? or a global list of all subqueries across shards?
+        // If a subquery comes from each shard then when is it combined? that seems weird that combination will do combination of normalized results that each is normalized just based on shard level result
```
navneet1v, line 37: Lets talk about these on the github issue and not on the PR.

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    static private float[] findStdPerSubquery(final List<CompoundTopDocs> queryTopDocs, final float[] meanPerSubQuery, final long[] elementsPerSubquery, final int numOfScores) {
+        final double[] deltaSumPerSubquery = new double[numOfScores];
+        Arrays.fill(deltaSumPerSubquery, 0);
+
+
+        //TODO: make this better, currently
+        // this is a horrible implementation in particular when it comes to the topDocsPerSubQuery.get(j)
+        // which does a random search on an abstract list type.
```
navneet1v, line 125: Please provide reason why this is bad and how it can be improved.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -26,8 +26,6 @@
 @Log4j2
 public class ScoreCombiner {
 
-    private static final Float ZERO_SCORE = 0.0f;
```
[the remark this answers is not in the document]
samuel-oci: It's not in use anywhere in the code

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
martin-gaievski: It should be fine if we know the exact implementation of List, as Navneet mentioned. But with list we can use functional style easier, without expensive conversion array -> stream, that was a reason why we switched to a List.

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
```diff
@@ -0,0 +1,166 @@
...
+2. Identical sub queries and their distribution between shards is currently completely implicit based on ordering and should be explicit based on identifier
+3. Weird calculation of numOfSubQueries instead of having a more explicit indicator
+ */
+@ToString(onlyExplicitlyIncluded = true)
+public class ZScoreNormalizationTechnique implements ScoreNormalizationTechnique {
+    @ToString.Include
+    public static final String TECHNIQUE_NAME = "z_score";
+    private static final float SINGLE_RESULT_SCORE = 1.0f;
+    @Override
+    public void normalize(List<CompoundTopDocs> queryTopDocs) {
```
martin-gaievski, line 35: please make all args of all public methods `final`

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
samuel-oci: Usually it is highly discouraged to do `List.get()` for an abstract List object because it could be an implementation that doesn't support random access efficiently (e.g. LinkedList). Suggested alternative is to enforce that this is explicitly declared as an ArrayList object throughout the hot path that require random access.

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
samuel-oci: ack, I think this comment is no longer relevant I put it there and forgot to remove so feel free to ignore this one.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryZScoreIT.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    private Map<String, Object> getTotalHits(Map<String, Object> searchResponseAsMap) {
+        Map<String, Object> hitsMap = (Map<String, Object>) searchResponseAsMap.get("hits");
+        return (Map<String, Object>) hitsMap.get("total");
+    }
+
+    private Optional<Float> getMaxScore(Map<String, Object> searchResponseAsMap) {
+        Map<String, Object> hitsMap = (Map<String, Object>) searchResponseAsMap.get("hits");
+        return hitsMap.get("max_score") == null ? Optional.empty() : Optional.of(((Double) hitsMap.get("max_score")).floatValue());
```
martin-gaievski, line 209: there 3 methods are copied from https://github.com/opensearch-project/neural-search/blob/main/src/test/java/org/opensearch/neuralsearch/query/HybridQueryIT.java#L271-L283, can we refactor code and pull them to a base class or a utility class for tests

src/test/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechniqueTests.java
```diff
@@ -0,0 +1,172 @@
...
+                new TotalHits(3, TotalHits.Relation.EQUAL_TO),
+                List.of(
+                        new TopDocs(
+                                new TotalHits(2, TotalHits.Relation.EQUAL_TO),
+                                new ScoreDoc[] { new ScoreDoc(2, 1.0f), new ScoreDoc(4, -1.0f) }
+                        ),
+                        new TopDocs(new TotalHits(0, TotalHits.Relation.EQUAL_TO), new ScoreDoc[0]),
+                        new TopDocs(
+                                new TotalHits(3, TotalHits.Relation.EQUAL_TO),
+                                new ScoreDoc[] { new ScoreDoc(3, 0.98058068f), new ScoreDoc(4, 0.39223227f), new ScoreDoc(2, -1.37281295f) }
```
martin-gaievski, line 82: can you please add a function or description with a formula of how those expected scores are calculated? 

Also why we have a negative score value for one of ScoreDocs?

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
martin-gaievski: I'm ok to switch from using general `List` to `ArrayList`, that still works with stream API and keep our requirements to a caller code cleaner. I expect that change will affect a lot of classes, thus I prefer to see it as a separate refactoring PR.

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
heemin32: Also, let's avoid using subject word like 'horrible'.

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
```diff
@@ -0,0 +1,166 @@
...
+                stdPerSubQuery[i] = 0;
+            } else {
+                stdPerSubQuery[i] = (float) Math.sqrt(deltaSumPerSubquery[i] / elementsPerSubquery[i]);
+            }
+        }
+
+        return stdPerSubQuery;
+    }
+
+    static private float sumScoreDocsArray(ScoreDoc[] scoreDocs) {
```
heemin32, line 150:
```suggestion
    static private float sumScoreDocsArray(final ScoreDoc[] scoreDocs) {
```

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
```diff
@@ -0,0 +1,166 @@
...
+            for (int j = 0; j < topDocsPerSubQuery.size(); j++) {
+                TopDocs subQueryTopDoc = topDocsPerSubQuery.get(j);
+                for (ScoreDoc scoreDoc : subQueryTopDoc.scoreDocs) {
+                    scoreDoc.score = normalizeSingleScore(scoreDoc.score, stdPerSubquery[j], meanPerSubQuery[j]);
+                }
+            }
+        }
+    }
+
+    static private float[] findScoreSumPerSubQuery(final List<CompoundTopDocs> queryTopDocs, final int numOfScores) {
```
heemin32, line 67: nit. private would be better unless you have specific reason this to be static. Better way would be moving all these methods to another class to make it easier to write unit test.

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
samuel-oci: @martin-gaievski same here, I added the comment out of intention to propose as a separate refactoring PR.

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
samuel-oci: sounds good @navneet1v I will give it some thought and will come up with suggestion. In any case not planning to do as part of this change. Can keep it for now and can suggest refactor or just remove if not achievable.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryZScoreIT.java
```diff
@@ -0,0 +1,211 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.neuralsearch.query;
```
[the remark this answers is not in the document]
samuel-oci: sounds good, will move to processor

src/test/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechniqueTests.java
samuel-oci: @martin-gaievski will add documentation. Regarding negatives, z-scores can also be negative:
https://www.z-table.com/
Btw, I should have mentioned in the comments that I wanted to bring up that the combiner is not good at dealing with negative values, but that would be adding additional scope.

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
samuel-oci: convention I was following is that if method is not dependent on any instance object it should be static.
Regarding refactoring method out to utility class, are there any other classes that can use it right now or in the future? Ideally I would like to avoid creating unnecessary abstraction.

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
samuel-oci: @heemin32 my apologies, will avoid it in the future.

samuel-oci:
> Created a feature branch from main, for this feature: feature/z-score-normalization
> 
> Please raise the PR against that branch. Also can you add a entry in the CHANGELOG.md file for this change.

Thank you @navneet1v @heemin32 @martin-gaievski for reviewing, I will create a new PR against the feature branch with your comments addressed.

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
```diff
@@ -0,0 +1,166 @@
...
+    public void normalize(List<CompoundTopDocs> queryTopDocs) {
+        // why are we doing that? is List<CompoundTopDocs> the list of subqueries for a single shard? or a global list of all subqueries across shards?
+        // If a subquery comes from each shard then when is it combined? that seems weird that combination will do combination of normalized results that each is normalized just based on shard level result
+        int numOfSubQueries = queryTopDocs.stream()
+                .filter(Objects::nonNull)
+                .filter(topDocs -> topDocs.getTopDocs().size() > 0)
+                .findAny()
+                .get()
+                .getTopDocs()
+                .size();
```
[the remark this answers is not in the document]
samuel-oci: yeah agreed, I will add checks, currently it's modeled on existing normalization techniques (MinMax/L2) which use similar code.
Also I'm not sure if there could also be a potential issue with implicit assumption that is made regarding the way we discover the number of sub queries. Currently it assumes that all none empty shard results will have all subqueries returned with 0 hits. I didn't confirm if this assumption always hold, but it would be nice if we had someway of passing it as metadata from upstream instead of making implicit assumptions. So just placing it here as a food for thought in case there is interest to explore that.

samuel-oci:
> As you have already written down the full code, its a good time to start doing the performance testing and search relevancy testing for this feature.

@navneet1v can you point me to how to run the performance and search relevancy tests? Is there a ready workflow with existing benchmarks?

PR #478, opened 2023-10-30
Deprecate max_token_score field of neural_sparse query

zhichao-aws:
### Description
In `neural_sparse` query, max_token_score field was used for sub-clause pruning of WAND scorer (lucene 9.7). Since we'll upgrade to lucene 9.8 in the next release, the inner logic in lucene changed and we don't need this field any more. We need to deprecate this field. To be more specific, in 2.x, user can still set this field but we'll ignore this and give a warning log. In 3.x, we don't parse this field.

### Check List
    - [x] All tests pass


zane-neo (approved):

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -238,14 +236,9 @@ protected Query doToQuery(QueryShardContext context) throws IOException {
...
 
-        final Float scoreUpperBound = maxTokenScore != null ? maxTokenScore : Float.MAX_VALUE;
-
         BooleanQuery.Builder builder = new BooleanQuery.Builder();
         for (Map.Entry<String, Float> entry : queryTokens.entrySet()) {
-            builder.add(
-                new BoostQuery(new BoundedLinearFeatureQuery(fieldName, entry.getKey(), scoreUpperBound), entry.getValue()),
-                BooleanClause.Occur.SHOULD
-            );
+            builder.add(FeatureField.newLinearQuery(fieldName, entry.getKey(), entry.getValue()), BooleanClause.Occur.SHOULD);
```
model-collapse, line 241: Please check the unit test, ensure this is covered.

model-collapse (commented):
Check the coverage thing.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: Sure.

navneet1v:
> Since we'll upgrade to lucene 9.8 in the next release, the inner logic in lucene changed and we don't need this field any more.

@zhichao-aws can you share some light around how lucene is taking care of this? Does lucene has added this capability? how does this impact the customer queries and search relevancy who is using this field.

navneet1v:
@zhichao-aws can you also add results of the tests which has been done, to ensure that if customer doesn't provide this deprecated field the queries are not impacted.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -62,8 +61,10 @@ public class NeuralSparseQueryBuilder extends AbstractQueryBuilder<NeuralSparseQ
     static final ParseField QUERY_TEXT_FIELD = new ParseField("query_text");
     @VisibleForTesting
     static final ParseField MODEL_ID_FIELD = new ParseField("model_id");
+    // We use max_token_score field to help WAND scorer prune query clause in lucene 9.7. But in lucene 9.8 the inner
+    // logics change, this field is not needed any more.
     @VisibleForTesting
-    static final ParseField MAX_TOKEN_SCORE_FIELD = new ParseField("max_token_score");
+    static final ParseField MAX_TOKEN_SCORE_FIELD = new ParseField("max_token_score").withAllDeprecated();
```
navneet1v, line 67: can we add the @Deprecated annotation on top of this.

navneet1v:
@zhichao-aws can you also add details on how we tested the upgrades?

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilderTests.java
```diff
@@ -88,7 +93,6 @@ public void testFromXContent_whenBuiltWithOptionals_thenBuildSuccessfully() {
               "VECTOR_FIELD": {
                 "query_text": "string",
                 "model_id": "string",
-                "max_token_score": 123.0,
```
navneet1v, line 91: can we keep a unit test where we are providing the deprecated field and no impact on queries are happening.

zhichao-aws:
> > Since we'll upgrade to lucene 9.8 in the next release, the inner logic in lucene changed and we don't need this field any more.
> 
> @zhichao-aws can you share some light around how lucene is taking care of this? Does lucene has added this capability? how does this impact the customer queries and search relevancy who is using this field.

Please refer this lucene [PR](https://github.com/apache/lucene/pull/12490) for more details. In short words,  Lucene used `ImpactsDISI`  to skip hits based on minimum competitive score. But lucene contributors found  _ImpactsDISI  does hurt more than it helps on average, because ImpactsDISI adds quite some overhead and the per-clauses minimum scores are usually so low that they don't actually enable skipping hits._ Now only top level scoring clause will use ImpactsDISI for pruning. FeatureQuery is not the top level scoring clause, and it won't use the ImpactsDISI. https://github.com/apache/lucene/pull/12490/files#diff-6ca6d673f9d09efdca430f2d5a381fbd862fd74385b778e45ade26fe112bca85 

So in our plugin, we don't need to provide the score upperbound for FeatureQuery after lucene 9.8. We have tested that this do reduces neural_sparse latency at a large margin. For 8.8 million docs case the latency was reduced to about 1/3 of lucene 9.7 case. This optimization doesn't affect the search result, only improves the speed of searching in shards. So it won't hurt the search relevance.  Since we still keep the api compatibility in 2.x, users can still use existing queries, we'll ignore the `max_token_score` field and give a warning log.

zhichao-aws:
> @zhichao-aws can you also add results of the tests which has been done, to ensure that if customer doesn't provide this deprecated field the queries are not impacted.

The deprecated field was optional in 2.11 release, so customers can always not privide this field.

zhichao-aws:
> @zhichao-aws can you also add details on how we tested the upgrades?

We have integ test for setting and doesn't setting this field:
https://github.com/zhichao-aws/neural-search/blob/dbda2c45638d141042793eeab0614742b898db3c/src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryIT.java#L97
https://github.com/zhichao-aws/neural-search/blob/dbda2c45638d141042793eeab0614742b898db3c/src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryIT.java#L68

We also have unit test to check we still parse this field, but will log a warning.
https://github.com/zhichao-aws/neural-search/blob/dbda2c45638d141042793eeab0614742b898db3c/src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilderTests.java#L123
When user provides this field, we ignore this and get the lucene query we want
https://github.com/zhichao-aws/neural-search/blob/dbda2c45638d141042793eeab0614742b898db3c/src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilderTests.java#L507

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: Sure.

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilderTests.java
zhichao-aws: I think we already have this.

zhichao-aws:
Hi @navneet1v , I've added some comments and commits, could you please help review again?

zhichao-aws:
BTW, this should be merged before we bump version to 2.12.0. For the old code there will be compilation errors due to the changes of lucene internal interface.

zhichao-aws:
force push to rebase 2.x

model-collapse (approved):
LGTM

PR #499, opened 2023-11-29
Bwc test

vibrantvarun:
### Description

### Issues Resolved

### Check List
    - [ ] All tests pass


src/testFixtures/java/org/opensearch/neuralsearch/TestUtils.java
```diff
@@ -27,15 +27,84 @@
...
 import org.opensearch.search.SearchHits;
 import org.opensearch.search.fetch.FetchSearchResult;
 import org.opensearch.search.query.QuerySearchResult;
 
 public class TestUtils {
 
-    private static final String RELATION_EQUAL_TO = "eq";
+    public static final String RELATION_EQUAL_TO = "eq";
     public static final float DELTA_FOR_SCORE_ASSERTION = 0.001f;
+    public static final String CLIENT_TIMEOUT_VALUE = "90s";
```
martin-gaievski, line 40: I think most of the constants that were moved here belong to exact integ test classes. In this shared util class we need to have something very generic and/or applicable to multiple tests, like ` NODES_BWC_CLUSTER` or `UPGRADED_CLUSTER`.

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
martin-gaievski, line 1: Some of the settings for ml-commons we're using for integ tests are not supported in earlier versions. we need  be a mechanism to handle that, otherwise bwc may start fail if we add some new setting in future. Example is "plugins.ml_commons.allow_registering_model_via_url", added in 2.8

martin-gaievski (commented):
I'm curios did you test anything with deployed models? In ml-commons there is this feature of automatic model undeploy/deploy, problem with that is that model id changes, that was one of the reasons we deploy fresh model for each integ test. Same thing can happen when nodes or all cluster got upgraded or/and restarted

Issue #202, opened 2023-06-26
Add backwards compatibility tests for automation


zelinh:
Add backwards compatibility tests for the plugin to automate end-to-end upgrade paths for faster releases.
Developer documentation for implementing bwc tests and hooking them to CI: https://github.com/opensearch-project/opensearch-plugins/blob/main/TESTING.md#backwards-compatibility-testing

This would include the BWC tests on both plugin level and distribution bundle level.

navneet1v:
@vamshin can you assign an owner for this task?

zelinh:
Hey @vibrantvarun, It seems like you have added the BWC tests for neural search on the plugin level with only necessary dependency plugins that neural search needs. 

We also have a framework for BWC tests on distribution level as well which will set up the test cluster with the bundle distribution including all plugins engineering effectiveness team builds. Are you plan to add that part as well? 

I believe it will be beneficial during the release cycle since we will be testing the exact distribution that we would ship to the customers.

It's a straightforward step based on what you already worked on the BWC tests. More onboarding info can be found [here](https://github.com/opensearch-project/opensearch-plugins/blob/main/TESTING.md#bwc-tests-on-distribution-bundle-level). 

cc: @vamshin @bbarani

vibrantvarun:
BWC tests at plugin level and distribution level are added now. Therefore closing the issue.

PR #515, opened 2023-12-07
BWC tests neural search

vibrantvarun:
### Description
Neural search transforms text into vectors and facilitates vector search both at ingestion time and at search time. During ingestion, neural search transforms document text into vector embeddings and indexes both the text and its vector embeddings in a vector index. When you use a neural query during search, neural search converts the query text into vector embeddings, uses vector search to compare the query and document embeddings, and returns the closest results.
Before you ingest documents into an index, documents are passed through a machine learning (ML) model, which generates vector embeddings for the document fields. When you send a search request, the query text or image is also passed through the ML model, which generates the corresponding vector embeddings. Then neural search performs a vector search on the embeddings and returns matching documents.

BWC stands for Backward Compatibility. 

Since, neural search feature is launched there are many features added to it. Therefore, there is a need to write BWC tests for the plugin to maintain compatibility with older versions.

This PR contains 

- [x] Setting up architecture for BWC tests
- [x] Optimizing the location of Common files used in BWC tests and Integ test framework
- [x] Semantic search

Moreover, in this PR I have added work to format the entire project by apply ./gradlew spotlessApply to all files and remove empty new lines new lines added below. This will do the following

- [x] User cannot add a new file without license header
- [x] No wildcard imports
- [x] Remove existing whitespaces and extra lines from the project

### Issues Resolved
[202](https://github.com/opensearch-project/neural-search/issues/202)

### Check List
    - [x] All tests pass


qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/AbstractRollingUpgradeTestCase.java
```diff
@@ -0,0 +1,90 @@
...
+import static org.opensearch.neuralsearch.TestUtils.ROLLING_UPGRADE_FIRST_ROUND;
+import static org.opensearch.neuralsearch.TestUtils.BWCSUITE_CLUSTER;
+import static org.opensearch.neuralsearch.TestUtils.NEURAL_SEARCH_BWC_PREFIX;
+
+public abstract class AbstractRollingUpgradeTestCase extends BaseNeuralSearchIT {
+    protected String testIndex;
+
+    @Before
+    protected void setIndex() {
+        // Creating index name by concatenating "knn-bwc-" prefix with test method name
```
naveentatikonda, line 28: Change the comment 
```suggestion
        // Creating index name by concatenating "neuralsearch-bwc-" prefix with test method name
```

qa/rolling-upgrade/build.gradle
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+}
+
+
+// Task to run BWC tests against the old cluster
+task testAgainstOldCluster(type: StandaloneRestIntegTestTask) {
+    dependsOn "zipBwcMlCommonsPlugin"
+    dependsOn "zipBwcKnnPlugin"
+    dependsOn "zipBwcPlugin"
```
naveentatikonda, line 69: same here

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/AbstractRestartUpgradeRestTestCase.java
```diff
@@ -0,0 +1,62 @@
...
+import static org.opensearch.neuralsearch.TestUtils.BWC_VERSION;
+import static org.opensearch.neuralsearch.TestUtils.NEURAL_SEARCH_BWC_PREFIX;
+import org.opensearch.test.rest.OpenSearchRestTestCase;
+
+public abstract class AbstractRestartUpgradeRestTestCase extends BaseNeuralSearchIT {
+    protected static String testIndex;
+
+    @Before
+    protected void setIndex() {
+        // Creating index name by concatenating "knn-bwc-" prefix with test method name
```
naveentatikonda, line 24:
```suggestion
        // Creating index name by concatenating "neuralsearch-bwc-" prefix with test method name
```

qa/rolling-upgrade/build.gradle
```diff
@@ -0,0 +1,135 @@
...
+        plugin(project.tasks.zipBwcMlCommonsPlugin.archiveFile)
+        plugin(project.tasks.zipBwcKnnPlugin.archiveFile)
+        plugin(project.tasks.zipBwcPlugin.archiveFile)
+        jvmArgs("-Xms1g", "-Xmx4g")
+        setting 'path.repo', "${buildDir}/cluster/shared/repo/${baseName}"
+        setting 'http.content_type.required', 'true'
+        environment "LD_LIBRARY_PATH", "${buildDir}/testclusters/${baseName}-0/distro/${knn_bwc_version_no_qualifier}-ARCHIVE/plugins/opensearch-knn/knnlib;${buildDir}/testclusters/${baseName}-0/distro/${knn_bwc_version_no_qualifier}-ARCHIVE/plugins/opensearch-knn/lib"
+        if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+            // While running on Windows OS, setting the PATH environment variable to include the paths to dlls of JNI libraries and windows dependencies
+            environment('PATH', System.getenv('PATH') + ";$rootDir/jni/release" + ";$rootDir/src/main/resources/windowsDependencies")
```
naveentatikonda, line 56: update the comment and remove jni path like mentioned in above comment

qa/rolling-upgrade/build.gradle
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// Part of rolling upgrade. Upgrades one node of the old cluster to new OpenSearch version with upgraded plugin version
+// This results in a mixed cluster with 2 nodes on the old version and 1 upgraded node.
+task testAgainstOneThirdUpgradedCluster(type: StandaloneRestIntegTestTask) {
+    useCluster testClusters."${baseName}"
+    dependsOn rootProject.tasks.assemble
+    dependsOn "testAgainstOldCluster"
+    doFirst {
+        testClusters."${baseName}".getNodes().getAt("${baseName}" + "-0").environment("LD_LIBRARY_PATH", "$rootDir/jni/release")
```
naveentatikonda, line 86: Also, remove the jni paths everywhere

.github/workflows/backwards_compatibility_tests_workflow.yml
```diff
@@ -0,0 +1,140 @@
...
+        with:
+          java-version: ${{ matrix.java }}
+
+      - if: startsWith(matrix.os,'ubuntu')
+        name: Install dependencies on ubuntu
+        run: |
+          sudo apt-get install libopenblas-dev gfortran -y
+
+      - if: startsWith(matrix.os,'windows')
+        name: Install MinGW Using Scoop on Windows
```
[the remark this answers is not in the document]
vibrantvarun: Yeah I was wondering the same. I will remove it

gradle.properties
```diff
@@ -3,6 +3,9 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
+version=2.0.0
```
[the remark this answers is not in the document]
vibrantvarun: I will remove it. Probably left as a typo while taking reference from KNN

.github/workflows/backwards_compatibility_tests_workflow.yml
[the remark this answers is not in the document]
vibrantvarun: done

gradle.properties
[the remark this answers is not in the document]
vibrantvarun: done

qa/build.gradle
```diff
@@ -0,0 +1,163 @@
...
+task pullMlCommonsBwcPlugin {
+    doLast {
+        copy {
+            from(java.nio.file.Path.of(tmp_dir.absolutePath, "opensearch-${neural_search_bwc_version_no_qualifier}", "plugins", "opensearch-ml"))
+            into java.nio.file.Path.of(tmp_dir.absolutePath, "opensearch-ml")
+        }
+    }
+}
+
+task pullKnnBwcPlugin {
```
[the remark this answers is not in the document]
vibrantvarun: done

qa/build.gradle
```diff
@@ -0,0 +1,163 @@
...
+}
+
+def knnJarDirectory = "$rootDir/build/dependencies/opensearch-knn"
+
+dependencies {
+    api "org.opensearch:opensearch:${opensearch_version}"
+    zipArchive group: 'org.opensearch.plugin', name:'opensearch-knn', version: "${opensearch_build}"
+    zipArchive group: 'org.opensearch.plugin', name:'opensearch-ml-plugin', version: "${opensearch_build}"
+    compileOnly fileTree(dir: knnJarDirectory, include: '*.jar')
+    //api group: 'commons-lang', name: 'commons-lang', version: '2.6'
```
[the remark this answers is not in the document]
vibrantvarun: done

qa/restart-upgrade/build.gradle
```diff
@@ -0,0 +1,98 @@
...
+        jvmArgs("-Xms1g", "-Xmx4g")
+        plugin(project.tasks.zipBwcMlCommonsPlugin.archiveFile)
+        plugin(project.tasks.zipBwcKnnPlugin.archiveFile)
+        plugin(project.tasks.zipBwcPlugin.archiveFile)
+        setting 'path.repo', "${buildDir}/cluster/shared/repo/${baseName}"
+        setting 'http.content_type.required', 'true'
+        environment "LD_LIBRARY_PATH", "${buildDir}/testclusters/${baseName}-0/distro/${knn_bwc_version_no_qualifier}-ARCHIVE/plugins/opensearch-knn/knnlib;${buildDir}/testclusters/${baseName}-0/distro/${knn_bwc_version_no_qualifier}-ARCHIVE/plugins/opensearch-knn/lib"
+        if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+            // While running on Windows OS, setting the PATH environment variable to include the paths to dlls of JNI libraries and windows dependencies
+            environment('PATH', System.getenv('PATH') + ";$rootDir/jni/release" + ";$rootDir/src/main/resources/windowsDependencies")
```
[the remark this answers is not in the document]
vibrantvarun: done

qa/restart-upgrade/build.gradle
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+    }
+}
+
+// Task to run BWC tests against the old cluster
+task testAgainstOldCluster(type: StandaloneRestIntegTestTask) {
+    dependsOn "zipBwcMlCommonsPlugin"
+    dependsOn "zipBwcKnnPlugin"
+    dependsOn "zipBwcPlugin"
```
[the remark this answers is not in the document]
vibrantvarun: done

qa/rolling-upgrade/build.gradle
vibrantvarun: done

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/AbstractRestartUpgradeRestTestCase.java
vibrantvarun: done

qa/rolling-upgrade/build.gradle
vibrantvarun: done

qa/rolling-upgrade/build.gradle
vibrantvarun: done

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearch.java
```diff
@@ -0,0 +1,91 @@
...
+                    Files.readString(Path.of(classLoader.getResource("processor/IndexMappings.json").toURI())),
+                    PIPELINE_NAME
+            );
+            addDocument(testIndex, DOC_ID,TEST_FIELD,TEXT);
+        }else {
+            Map<String,Object> pipeline= getIngestionPipeline(PIPELINE_NAME);
+            assertNotNull(pipeline);
+            String modelId=getModelId(pipeline, TEXT_EMBEDDING_PROCESSOR);
+            validateTestIndex(modelId);
+            deleteIndex(testIndex);
```
martin-gaievski, line 50: pipelines and models are global, please make sure they are not conflicting in case of multiple tests. We either need to wipe them after each test of check before creating if that resource exists. For models I think delete is the better option as models are taking memory, so with certain number of tests we'll start having CB exceptions of model load.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearch.java
```diff
@@ -0,0 +1,91 @@
...
+            createIndexWithConfiguration(
+                    testIndex,
+                    Files.readString(Path.of(classLoader.getResource("processor/IndexMappings.json").toURI())),
+                    PIPELINE_NAME
+            );
+            addDocument(testIndex, DOC_ID,TEST_FIELD,TEXT);
+        }else {
+            Map<String,Object> pipeline= getIngestionPipeline(PIPELINE_NAME);
+            assertNotNull(pipeline);
+            String modelId=getModelId(pipeline, TEXT_EMBEDDING_PROCESSOR);
```
martin-gaievski, line 48: Why we're retrieving model id from index and not reusing it from the previous step (runningAgainstOldCluster)? Did you check if model id is same or it changes between those two stages of test?

gradle.properties
```diff
@@ -3,6 +3,8 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
+systemProp.bwc.version=2.12.0-SNAPSHOT
```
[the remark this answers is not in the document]
vibrantvarun: We should keep latest version here because in wire dependency between older versions 3.x is compatible with 2.(latest-major)(i.e. 2.12-SNAPSHOT). Let say if we set 2.9 and If someone clones the project just hit ./gradlew :qa:bwcTestSuite then rolling upgrade tests will fail. He has to manually update the version here with the latest compatible version. This has been kept just to run tests easily on local machine with zero effort.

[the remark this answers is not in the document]
vibrantvarun: Still if you suggest to keep it 2.9 then I will update it.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearch.java
vibrantvarun: Ack

qa/restart-upgrade/src/test/resources/processor/IndexMappings.json
```diff
@@ -0,0 +1,33 @@
+{
+  "settings": {
+    "index": {
+      "knn": true,
+      "knn.algo_param.ef_search": 100,
+      "refresh_interval": "30s",
+      "default_pipeline": "%s"
+    },
+    "number_of_shards": 1,
```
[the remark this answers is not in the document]
vibrantvarun: Good point. I will update it.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearch.java
vibrantvarun: When the cluster is running against older version then we create a ingestion pipeline with a text embedding processor in it. That time I add model id into it.


Now when the cluster gets updated with a new version of OS then I cannot persist the model Id in a global variable. Similarly, I cannot store a task Id In which I can persist it and later use it to fetch model Id. So the only way I could fetch model id is by text embedding processor.

vibrantvarun: In Knn they can store model id but for neural we can't.

gradle.properties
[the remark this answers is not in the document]
martin-gaievski: That makes sense, I think we're good with 2.12 then

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearch.java
martin-gaievski: I such case it's possible to have unexpected model id. In neural search integ tests there are some flaky tests where problem is exactly that - model has not been either deployed or deleted properly. We can leave it like this for now, just need to keep this issue in mind when more tests are added in future.

vibrantvarun: I have taken care of it now

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearch.java
```diff
@@ -0,0 +1,93 @@
...
+            validateTestIndex(modelId);
+            deletePipeline(PIPELINE_NAME);
+            deleteModel(modelId);
+            deleteIndex(testIndex);
+        }
+    }
+
+
+    private void validateTestIndex(String modelId) throws Exception {
```
martin-gaievski, line 57: nit: for tests you can use SneakyThrows annotation instead of putting exact Exception to the method signature

qa/restart-upgrade/src/test/resources/processor/IndexMappings.json
```diff
@@ -0,0 +1,33 @@
...
+  "mappings": {
+    "properties": {
+      "passage_embedding": {
+        "type": "knn_vector",
+        "dimension": 768,
+        "method": {
+          "name": "hnsw",
+          "space_type": "l2",
+          "engine": "lucene",
+          "parameters": {
```
martin-gaievski, line 21: I think defaults are fine for engine params, for neural search it not that important

martin-gaievski (commented):
Could you please run `spotlessApply` for your changes, there are many small formatting issues like missing spaces etc.?

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearch.java
vibrantvarun: For all tests it is not applicable as there is an issue with lombok. It cannot rectify checked exception. There is checked exception (URI syntax exception) coming from [here](https://github.com/opensearch-project/neural-search/pull/515/files#diff-e4fb9812973229e1ebaad95042281ab554d59b3c3d16155af075823fe29b8caaR41). And If If add @Sneakythrows then I get below error 
```
                                                                                                                           ^
/Users/varunudr/neuralsearch/neural-search/qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearch.java:93: error: unreported exception URISyntaxException; must be caught or declared to be thrown
        String requestBody = Files.readString(Path.of(classLoader.getResource("processor/PipelineConfiguration.json").toURI()));

```

vibrantvarun:
**I have configured ./gradlew spotlessApply to scan every file in the project and remote extra white spaces.**

formatter/license-header.txt
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
```
martin-gaievski, line 5: are you sure we need anything other than first two lines in the license header? I think we normally use just those two 2 lines.

vibrantvarun: Yeah I think I can remove that. Just took reference from legacy plugin job scheduler . There they use it so I used it as well

qa/rolling-upgrade/src/test/resources/processor/UploadModelRequestBody.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "traced_small_model",
+  "version": "1.0.0",
+  "model_format": "TORCH_SCRIPT",
+  "model_task_type": "text_embedding",
+  "model_content_hash_value": "e13b74006290a9d0f58c1376f9629d4ebc05a0f9385f40db837452b167ae9021",
+  "model_group_id": "<MODEL_GROUP_ID>",
```
jmazanec15, line 7: Should this be "%s" instead of "<MODEL_GROUP)ID>"?

qa/restart-upgrade/src/test/resources/processor/UploadModelRequestBody.json
```diff
@@ -0,0 +1,15 @@
+{
```
jmazanec15, line 1: Are configurations duplicated across restart and rolling? If so, can we avoid duplicating?

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearch.java
```diff
@@ -0,0 +1,91 @@
...
+import java.nio.file.Path;
+import java.util.Map;
+
+import static org.opensearch.neuralsearch.TestUtils.NODES_BWC_CLUSTER;
+
+import static org.opensearch.neuralsearch.TestUtils.getModelId;
+import static org.opensearch.neuralsearch.TestUtils.TEXT_EMBEDDING_PROCESSOR;
+import org.opensearch.neuralsearch.query.NeuralQueryBuilder;
+
+public class SemanticSearch extends AbstractRestartUpgradeRestTestCase {
```
jmazanec15, line 18: Can this be `SemanticSearchIT` to be consistent with out test names?

qa/build.gradle
```diff
@@ -0,0 +1,174 @@
+/*
+ *  Copyright OpenSearch Contributors
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import org.apache.tools.ant.taskdefs.condition.Os
+import org.w3c.dom.Text
```
jmazanec15, line 7: Do we need this?

qa/build.gradle
```diff
@@ -0,0 +1,174 @@
...
+            }
+        }
+        ant.get(
+                src: srcUrl,
+                dest: tmp_dir.absolutePath,
+                httpusecaches: false
+        )
+        copy {
+            if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+                from zipTree(java.nio.file.Path.of(tmp_dir.absolutePath, "opensearch-${neural_search_bwc_version_no_qualifier}-${os_platform}-x64.${file_ext}"))
```
jmazanec15, line 93: Can we import Path so that we dont need to add `java.nio.file` prefix?

qa/build.gradle
```diff
@@ -0,0 +1,174 @@
...
+   }
+}
+
+// Task to pull ml plugin from archive
+task pullMlCommonsBwcPlugin {
+    doLast {
+        copy {
+            from(java.nio.file.Path.of(tmp_dir.absolutePath, "opensearch-${neural_search_bwc_version_no_qualifier}", "plugins", "opensearch-ml"))
+            into java.nio.file.Path.of(tmp_dir.absolutePath, "opensearch-ml")
+        }
```
jmazanec15, line 108: Should we delete the file like we do for pullBwcPlugin?

qa/build.gradle
```diff
@@ -0,0 +1,174 @@
...
+    doLast {
+        copy {
+            from(java.nio.file.Path.of(tmp_dir.absolutePath, "opensearch-${neural_search_bwc_version_no_qualifier}", "plugins", "opensearch-neural-search"))
+            into java.nio.file.Path.of(tmp_dir.absolutePath, "opensearch-neural-search")
+        }
+        delete java.nio.file.Path.of(tmp_dir.absolutePath, "opensearch-${neural_search_bwc_version_no_qualifier}"), java.nio.file.Path.of(tmp_dir.absolutePath, "opensearch-${neural_search_bwc_version_no_qualifier}-${os_platform}-x64.${file_ext}")
+    }
+}
+
+// Task to zip ml-commons plugin from archive
```
jmazanec15, line 135: Unzip, right?

qa/build.gradle
```diff
@@ -0,0 +1,174 @@
...
+        ant.get(
+                src: srcUrl,
+                dest: tmp_dir.absolutePath,
+                httpusecaches: false
+        )
+        copy {
+            if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+                from zipTree(java.nio.file.Path.of(tmp_dir.absolutePath, "opensearch-${neural_search_bwc_version_no_qualifier}-${os_platform}-x64.${file_ext}"))
+            } else {
+                from tarTree(java.nio.file.Path.of(tmp_dir.absolutePath, "opensearch-${neural_search_bwc_version_no_qualifier}-${os_platform}-x64.${file_ext}"))
```
jmazanec15, line 95: This will fail on arm. Is that expected?

qa/build.gradle
```diff
@@ -0,0 +1,174 @@
...
+String artifact_type = "tar"
+String file_ext = "tar.gz"
+
+if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+    os_platform = "windows"
+    artifact_type = "zip"
+    file_ext = "zip"
+}
+
+task deletetempDirectories {
```
jmazanec15, line 63: deleteTempDirectories

qa/restart-upgrade/build.gradle
```diff
@@ -0,0 +1,87 @@
+/*
+ *  Copyright OpenSearch Contributors
+ *  SPDX-License-Identifier: Apache-2.0
+ */
+
+import org.opensearch.gradle.testclusters.StandaloneRestIntegTestTask
+import org.apache.tools.ant.taskdefs.condition.Os
+import java.util.concurrent.Callable
+
+apply from : "$rootDir/qa/build.gradle"
```
jmazanec15, line 10: Do we need to call this? Or will it be automatically inherited?

qa/restart-upgrade/build.gradle
```diff
@@ -0,0 +1,87 @@
...
+
+String default_bwc_version = System.getProperty("bwc.version")
+String neural_search_bwc_version = System.getProperty("tests.bwc.version", default_bwc_version)
+boolean isSnapshot = neural_search_bwc_version.contains("-SNAPSHOT")
```
jmazanec15, line 14: Lets stick to `snake_case` for var names.

qa/restart-upgrade/build.gradle
```diff
@@ -0,0 +1,87 @@
...
+    }
+}), rootProject.tasks.bundlePlugin.archiveFile]
+
+// Creates a test cluster of previous version and loads k-NN plugin of bwcVersion
+testClusters {
+    "${baseName}" {
+        testDistribution = "ARCHIVE"
+        versions = [neural_search_bwc_version, opensearch_version]
+        numberOfNodes = 3
+        jvmArgs("-Xms1g", "-Xmx4g")
```
jmazanec15, line 46: Why these args?

qa/restart-upgrade/build.gradle
```diff
@@ -0,0 +1,87 @@
...
+    systemProperty 'tests.skip_delete_model_index', 'true'
+    systemProperty 'tests.plugin_bwc_version', neural_search_bwc_version
+
+    nonInputProperties.systemProperty('tests.rest.cluster', "${-> testClusters."${baseName}".allHttpSocketURI.join(",")}")
+    nonInputProperties.systemProperty('tests.clustername', "${-> testClusters."${baseName}".getName()}")
+    systemProperty 'tests.security.manager', 'false'
+}
+
+// All nodes are upgraded to latest version and run the tests
+task testRestartUpgrade(type: StandaloneRestIntegTestTask) {
```
jmazanec15, line 72: can we call this `testAgainstNewCluster` to parallel `testAgainstOldCluster`?

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/AbstractRestartUpgradeRestTestCase.java
```diff
@@ -0,0 +1,65 @@
...
+import org.opensearch.common.settings.Settings;
+import org.opensearch.neuralsearch.BaseNeuralSearchIT;
+import static org.opensearch.neuralsearch.TestUtils.CLIENT_TIMEOUT_VALUE;
+import static org.opensearch.neuralsearch.TestUtils.RESTART_UPGRADE_OLD_CLUSTER;
+import static org.opensearch.neuralsearch.TestUtils.BWC_VERSION;
+import static org.opensearch.neuralsearch.TestUtils.NEURAL_SEARCH_BWC_PREFIX;
+import org.opensearch.test.rest.OpenSearchRestTestCase;
+
+public abstract class AbstractRestartUpgradeRestTestCase extends BaseNeuralSearchIT {
+    protected static String testIndex;
```
jmazanec15, line 19: does this need to be static?

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/AbstractRestartUpgradeRestTestCase.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Before
+    protected void setIndex() {
```
jmazanec15, line 22: Could we just make this a getter function that tests can call - `getIndexNameForTest()`?

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearch.java
```diff
@@ -0,0 +1,91 @@
...
+        NeuralQueryBuilder neuralQueryBuilder = new NeuralQueryBuilder();
+        neuralQueryBuilder.fieldName("passage_embedding");
+        neuralQueryBuilder.modelId(modelId);
+        neuralQueryBuilder.queryText(TEXT);
+        neuralQueryBuilder.k(1);
+        Map<String, Object> response = search(testIndex, neuralQueryBuilder, 1);
+        assertNotNull(response);
+    }
+
+    private String uploadTextEmbeddingModel() throws Exception {
```
jmazanec15, line 65: Are we able to abstract these methods to a abstract class that can be used across all different integ tests?

qa/rolling-upgrade/build.gradle
```diff
@@ -0,0 +1,121 @@
...
+
+apply from : "$rootDir/qa/build.gradle"
+
+String default_bwc_version = System.getProperty("bwc.version")
+String neural_search_bwc_version = System.getProperty("tests.bwc.version", default_bwc_version)
+boolean isSnapshot = neural_search_bwc_version.contains("-SNAPSHOT")
+String knn_bwc_version_no_qualifier = isSnapshot ? neural_search_bwc_version - "-SNAPSHOT" : neural_search_bwc_version
+String baseName = "neuralSearchBwcCluster-rolling"
+
+List<Provider<RegularFile>> plugins = [provider(new Callable<RegularFile>(){
```
jmazanec15, line 19: Can we move these up to qa/build.gradle so we dont have to repeat code between restart and rolling?

qa/rolling-upgrade/src/test/resources/processor/UploadModelRequestBody.json
vibrantvarun: In Integ test similar kind of work is done. If this has to change then I can raise a new PR for that. Already this PR became heavy. Probably in next iteration for multimodal test I will add it. Does it work for now?

qa/restart-upgrade/src/test/resources/processor/UploadModelRequestBody.json
vibrantvarun: So, the files are kept at 2 places because the the test folder only picks resources if it is present under the same path. I have tried to keep it common but it does not work. 

For example
```
qa:
      restart-upgrade:
                     test
                            code
                            resources
       rolling-upgrade:
                     test
                            code
                            resources
```

I have tried all combinations like 

```
qa:
      restart-upgrade:
                     test
                            code
       rolling-upgrade:
                     test
                            code
       src
            test
                 resources
```

```
qa:
      resources:
      restart-upgrade:
                     test
                            code
       rolling-upgrade:
                     test
                            code
```

But it didn't work. 

Even I tried to change the path from gradle from where it picks the resources but it didn't work because after ./gradlew build it creates a file location in resources and from there it picks the resource. So i cannot change it

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearch.java
vibrantvarun: ACK

qa/build.gradle
vibrantvarun: Yeah we don't. Removing it.

qa/build.gradle
vibrantvarun: In pullBWCPlugin it deletes opensearch artifact.

The execution flow is
1) pullOpensearchArtifact 
2) Pull knn plugin
3) pull ml plugin
4) pull neural plugin
5) Delete opensearch artifact

So the reason it is kept in pullBwc plugin is because it the last one to be called. Therefore it deletes it after the use.

To delete individual package we use
https://github.com/opensearch-project/neural-search/pull/515/files#diff-07e82fba5bdc43702a0da2ef075f370622dd8ca817b16329d33fa12b9816a2e5R142

qa/build.gradle
vibrantvarun: yes, I will make the change

qa/build.gradle
vibrantvarun: Yes, This is consistent with [KNN behaviour ](https://github.com/opensearch-project/k-NN/blob/feb7968f25a35ecd1b73793584a5ebd2a36d9d72/qa/build.gradle#L73)

vibrantvarun: We support Ubuntu, Windows and Mac

qa/restart-upgrade/build.gradle
vibrantvarun: We need to call this

qa/build.gradle
vibrantvarun: ACK

qa/restart-upgrade/build.gradle
vibrantvarun: It will maintain consistency with our other plugin KNN. If still you want I can change this.

qa/restart-upgrade/build.gradle
vibrantvarun: Can you point to the variable you are referring to?

qa/restart-upgrade/build.gradle
vibrantvarun: There is an issue with ml-commons. It sends Circuit breaker exceptions due to out of memory. This has been given so that nodes can change memory requirements on runtime. This issue is consistent with integ tests too. So @martin-gaievski made a change [here](https://github.com/opensearch-project/neural-search/blob/5daddfd2f10486ad4688742aa26db82461b6a73c/build.gradle#L218)

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/AbstractRestartUpgradeRestTestCase.java
vibrantvarun: Yes it should be static. As the name for testIndex is set once and need not to be changed. If any issues comes then it is easy to debug. Otherwise changing it to non static will again and again require to create object and assign a new value in every test.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/AbstractRestartUpgradeRestTestCase.java
vibrantvarun: I think there is no need of changing the index name here. Therefore before running the tests it will set the name and will increase the reusability of the variable here. Therefore if I make one extra method getIndexNameForTest() and call it will be addition of new code with a very less impact. I can easily use testIndex without creating extra object, without assigning any new value. Anyways I always delete the resources after the test is executed.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearch.java
vibrantvarun: No, I tried that. But, there is a file handling issue which comes because it is not able to find the resources. I have discussed this issue with @navneet1v and therefore came to conclusion of putting it seperately. At the first place I wanted abstraction to happen and keep it in BaseNeuralSearchIT but gradle is not able to read the files from resources in cross module dependency. 

TestFixtures can only read file from its resources
Similarly qa: restart-upgrade can only read file from its resources
and qa build.gradle can't read resources from src:test because it is altogether a different module

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearch.java
martin-gaievski: got it, no issues

qa/rolling-upgrade/build.gradle
vibrantvarun: ACK

qa/restart-upgrade/build.gradle
jmazanec15: lets change. It will make it more readable

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/AbstractRestartUpgradeRestTestCase.java
jmazanec15: With the @Before annotation, it will be reset for each test. So Im not sure it needs to be static. I dont think there is any benefit.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/AbstractRestartUpgradeRestTestCase.java
jmazanec15: In general, I prefer to avoid adding additional state to classes if its not needed. For this case, we could get rid of the testIndex variable and just have a single getter function name that can be called to get a test specific index name.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearchIT.java
```diff
@@ -0,0 +1,87 @@
...
+        if (isRunningAgainstOldCluster()) {
+            String modelId = uploadTextEmbeddingModel();
+            loadModel(modelId);
+            createPipelineProcessor(modelId, PIPELINE_NAME);
+            createIndexWithConfiguration(
+                testIndex,
+                Files.readString(Path.of(classLoader.getResource("processor/IndexMappings.json").toURI())),
+                PIPELINE_NAME
+            );
+            addDocument(testIndex, DOC_ID, TEST_FIELD, TEXT);
```
jmazanec15, line 40: just checking this will refresh immediately correct? There may be race conditions if it is not refreshed.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearchIT.java
```diff
@@ -0,0 +1,87 @@
...
+        } else {
+            Map<String, Object> pipeline = getIngestionPipeline(PIPELINE_NAME);
+            assertNotNull(pipeline);
+            String modelId = getModelId(pipeline, TEXT_EMBEDDING_PROCESSOR);
+            validateTestIndex(modelId);
```
jmazanec15, line 45: Should we also test that we can add another document to the index - this would ensure the processor created on the old version works on the new version.

qa/restart-upgrade/src/test/resources/processor/UploadModelRequestBody.json
jmazanec15: ah I see makes sense.

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearchIT.java
```diff
@@ -0,0 +1,107 @@
...
+import static org.opensearch.neuralsearch.TestUtils.TEXT_EMBEDDING_PROCESSOR;
+import org.opensearch.neuralsearch.query.NeuralQueryBuilder;
+
+public class SemanticSearchIT extends AbstractRollingUpgradeTestCase {
+    private static final String PIPELINE_NAME = "nlp-pipeline";
+    private static final String TEST_FIELD = "test-field";
+    private static final String TEXT = "Hello world";
+    private static final String TEXT_MIXED = "Hello world mixed";
+    private static final String TEXT_UPGRADED = "Hello world upgraded";
+    private static final int NUM_DOCS = 1;
```
jmazanec15, line 22: call this NUM_DOCS_PER_ROUND to show that it is not the total number of docs

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -139,7 +138,10 @@ protected String uploadModel(String requestBody) throws Exception {
         String modelGroupId = registerModelGroup();
         // model group id is dynamically generated, we need to update model update request body after group is registered
         requestBody = requestBody.replace("<MODEL_GROUP_ID>", modelGroupId);
+        return uploadModelId(requestBody);
+    }
 
+    protected String uploadModelId(String requestBody) throws Exception {
```
jmazanec15, line 144: What does updload model id mean?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -812,6 +817,63 @@ private String registerModelGroup() {
...
+     * @param text to be added
+     */
+    protected void addDocument(String index, String docId, String fieldName, String text) throws IOException {
+        Request request = new Request("PUT", "/" + index + "/_doc/" + docId + "?refresh=true");
+
+        XContentBuilder builder = XContentFactory.jsonBuilder().startObject().field(fieldName, text).endObject();
+        request.setJsonEntity(builder.toString());
+        client().performRequest(request);
+
+        request = new Request("POST", "/" + index + "/_refresh");
```
jmazanec15, line 843: Do we need this if we call refresh in the put request?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -812,6 +817,63 @@ private String registerModelGroup() {
...
+        Response response = client().performRequest(request);
+        assertEquals(request.getEndpoint() + ": failed", RestStatus.OK, RestStatus.fromCode(response.getStatusLine().getStatusCode()));
+    }
+
+    /**
+    * Get ingest pipeline
+    * @param pipelineName of the ingest pipeline
```
jmazanec15, line 850: Add what is the return value as well

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
vibrantvarun: No we don't need it

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearchIT.java
vibrantvarun: yes it does

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearchIT.java
vibrantvarun: Yes I can

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
vibrantvarun: I already have uploadModel method signature in the BaseNeuralSearchIT. And due to file handling issues between cross modules I need to segregate it into 2 methods. So the best possible synonym I could find is uploadModelId

jmazanec15: I would probably call it `uploadModel(String requestBody, String groupId)` and move `requestBody = requestBody.replace("<MODEL_GROUP_ID>", modelGroupId);` into this function.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearchIT.java
```diff
@@ -0,0 +1,87 @@
...
+
+    private static final String PIPELINE_NAME = "nlp-pipeline";
+    private static final String TEST_FIELD = "passage_text";
+    private static final String TEXT = "Hello world";
+    private static final String TEXT_1 = "Hello world a";
+
+    // Test restart-upgrade Semantic Search
+    // Create Text Embedding Processor, Ingestion Pipeline and add document
+    // Validate process , pipeline and document count in restart-upgrade scenario
+    public void testSemanticSearch() throws Exception {
```
navneet1v, line 28:
```suggestion
    @SneakyThrows
    public void testSemanticSearch_E2EFlow() {
```

qa/restart-upgrade/src/test/resources/processor/IndexMappings.json
```diff
@@ -0,0 +1,32 @@
+{
+  "settings": {
+    "index": {
+      "knn": true,
+      "knn.algo_param.ef_search": 100,
+      "refresh_interval": "30s",
+      "default_pipeline": "%s"
+    },
+    "number_of_shards": 6,
+    "number_of_replicas": 0
```
navneet1v, line 10: should we go with 3 primaries and 1 replica?

qa/restart-upgrade/src/test/resources/processor/PipelineConfiguration.json
```diff
@@ -0,0 +1,13 @@
+{
+  "description": "text embedding pipeline for hybrid",
+  "processors": [
+    {
+      "text_embedding": {
+        "model_id": "%s",
+        "field_map": {
+          "passage_text": "passage_embedding"
+        }
```
navneet1v, line 9: in the follow up let's make sure that we are adding the tests for array fields and map fields in fieldMap

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearchIT.java
```diff
@@ -0,0 +1,105 @@
...
+    private static final String TEST_FIELD = "test-field";
+    private static final String TEXT = "Hello world";
+    private static final String TEXT_MIXED = "Hello world mixed";
+    private static final String TEXT_UPGRADED = "Hello world upgraded";
+    private static final int NUM_DOCS_PER_ROUND = 1;
+
+    // Test rolling-upgrade Semantic Search
+    // Create Text Embedding Processor, Ingestion Pipeline and add document
+    // Validate process , pipeline and document count in rolling-upgrade scenario
+    public void testSemanticSearch() throws Exception {
```
navneet1v, line 27:
```suggestion
    @SneakyThrows
    public void testSemanticSearch_E2ETest() {
```

qa/rolling-upgrade/src/test/resources/processor/IndexMappings.json
```diff
@@ -0,0 +1,32 @@
+{
+  "settings": {
+    "index": {
+      "knn": true,
+      "knn.algo_param.ef_search": 100,
+      "refresh_interval": "30s",
+      "default_pipeline": "%s"
+    },
+    "number_of_shards": 6,
+    "number_of_replicas": 0
```
navneet1v, line 10: same comment as of in RestartUpgrade

navneet1v (commented):
Few minor comments. Overall code looks good. 

Great effort in adding the spotlessCheck this was long pending. Code looks in awesome shape now.

build.gradle
```diff
@@ -302,6 +322,7 @@ jacocoTestReport {
     }
 }
 
+check.dependsOn spotlessCheck
```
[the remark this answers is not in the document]
vibrantvarun: Currently it works for build.

gradle Build depends on gradle Check and gradle Check depends on spotlessApply.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearchIT.java
vibrantvarun: https://github.com/opensearch-project/neural-search/pull/515#discussion_r1437834192

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearchIT.java
vibrantvarun: testSemanticSearch_E2Etest is little weird name as it has test on both sides start and end.

vibrantvarun: I can add E2E if you want but testSemanticSearch looks more adequate to it

qa/restart-upgrade/src/test/resources/processor/IndexMappings.json
vibrantvarun: https://github.com/opensearch-project/neural-search/pull/515#discussion_r1437284537

qa/rolling-upgrade/src/test/resources/processor/IndexMappings.json
vibrantvarun: https://github.com/opensearch-project/neural-search/pull/515#discussion_r1437284537

build.gradle
```diff
@@ -123,6 +125,26 @@ allprojects {
...
+    apply from: rootProject.file('repositories.gradle').absoluteFile
+    plugins.withId('java') {
+        sourceCompatibility = targetCompatibility = "11"
+    }
+
+    afterEvaluate {
+        project.dependencyLicenses.enabled = false
+        project.thirdPartyAudit.enabled = false
+        project.loggerUsageCheck.enabled = false
+        project.forbiddenApis.ignoreFailures = true
```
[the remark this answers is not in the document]
vibrantvarun: Typo

build.gradle
```diff
@@ -123,6 +125,26 @@ allprojects {
...
+        project.forbiddenPatterns {
+            setEnabled(false)
+        }
+        project.testingConventions.enabled = false
+        project.validateNebulaPom.enabled = false
+        project.licenseFile = rootProject.file('LICENSE.txt')
+        project.noticeFile = rootProject.file('NOTICE.txt')
+        project.forbiddenApis.ignoreFailures = true
```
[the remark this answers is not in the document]
vibrantvarun: Typo

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
vibrantvarun: I have made names more readable now.

qa/restart-upgrade/src/test/resources/processor/IndexMappings.json
vibrantvarun: I am getting below error for 3 / 1 configration of shards
```
org.opensearch.neuralsearch.bwc.SemanticSearchIT > testSemanticSearch_E2EFlow FAILED
    org.opensearch.client.ResponseException: method [PUT], host [http://127.0.0.1:40221/], URI [/neuralsearch-bwc-testsemanticsearch_e2eflow/_doc/2?refresh=true], status line [HTTP/1.1 400 Bad Request]
    {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Model not ready yet. Please run this first: POST /_plugins/_ml/models/eJaKyYwBOutJBPQEmqxh/_deploy"}],"type":"illegal_argument_exception","reason":"Model not ready yet. Please run this first: POST /_plugins/_ml/models/eJaKyYwBOutJBPQEmqxh/_deploy"},"status":400}
        at __randomizedtesting.SeedInfo.seed([7FBAAD9E29825EEB:C9A802068E91C12C]:0)
        at app//org.opensearch.client.RestClient.convertResponse(RestClient.java:385)
        
       ```

jmazanec15 (approved):
LGTM thanks!

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearchIT.java
navneet1v: actually testSemantic Search doesn't provide any information. Semantic Search is pretty broad term. What you are actually validating is textEmbedding Processor. So better name here could be

`testTextEmbeddingProcessor`

qa/restart-upgrade/src/test/resources/processor/IndexMappings.json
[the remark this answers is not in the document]
navneet1v: 1 shard use-case is a valid use case for hybrid search but not for processors. I would say lets go with 3 shards and 1 replica. This will ensure that we are handling all the valid cases for text embedding processor. For Hybrid Search we should create test for different shard configurations.

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/SemanticSearchIT.java
vibrantvarun: Done

navneet1v (approved):

vibrantvarun:
This PR also adds the feature request from [https://github.com/opensearch-project/neural-search/pull/515](https://github.com/opensearch-project/neural-search/pull/515)

PR #524, opened 2023-12-27
Fixing multiple issues reported in #497

martin-gaievski:
### Description
Fixed multiple issue in scope of this PR:

- Allow multiple identical sub-queries in hybrid query
- Allow total_hits flag usage
- Move check for mismatch of number of docs in fetch and query results to earlier stage, so normalization is not even called. current behavior is to throw an exception and fail search query

### Issues Resolved
https://github.com/opensearch-project/neural-search/issues/497

### Check List
    - [X] All tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Features
 ### Enhancements
 ### Bug Fixes
-- Multiple identical subqueries in Hybrid query  ([#524](https://github.com/opensearch-project/neural-search/pull/524))
+- Fixing multiple issues reported in #497 ([#524](https://github.com/opensearch-project/neural-search/pull/524))
 ### Infrastructure
 ### Documentation
 ### Maintenance
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Features
 ### Enhancements
 ### Bug Fixes
+- Multiple identical subqueries in Hybrid query  ([#524](https://github.com/opensearch-project/neural-search/pull/524))
 ### Infrastructure
 ### Documentation
 ### Maintenance
--- a/src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessor.java
@@ -7,9 +7,7 @@ package org.opensearch.neuralsearch.processor;
 
 import static org.opensearch.neuralsearch.search.util.HybridSearchResultFormatUtil.isHybridQueryStartStopElement;
 
-import java.util.Arrays;
 import java.util.List;
-import java.util.Locale;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -20,8 +18,6 @@ import org.opensearch.action.search.SearchPhaseName;
 import org.opensearch.action.search.SearchPhaseResults;
 import org.opensearch.neuralsearch.processor.combination.ScoreCombinationTechnique;
 import org.opensearch.neuralsearch.processor.normalization.ScoreNormalizationTechnique;
-import org.opensearch.search.SearchHit;
-import org.opensearch.search.SearchHits;
 import org.opensearch.search.SearchPhaseResult;
 import org.opensearch.search.fetch.FetchSearchResult;
 import org.opensearch.search.internal.SearchContext;
@@ -102,16 +98,7 @@ public class NormalizationProcessor implements SearchPhaseResultsProcessor {
         }
 
         QueryPhaseResultConsumer queryPhaseResultConsumer = (QueryPhaseResultConsumer) searchPhaseResult;
-        if (queryPhaseResultConsumer.getAtomicArray().asList().stream().filter(Objects::nonNull).noneMatch(this::isHybridQuery)) {
-            return true;
-        }
-        List<QuerySearchResult> querySearchResults = getQueryPhaseSearchResults(searchPhaseResult);
-        Optional<FetchSearchResult> fetchSearchResult = getFetchSearchResults(searchPhaseResult);
-        if (shouldSkipProcessorDueToIncompatibleQueryAndFetchResults(querySearchResults, fetchSearchResult)) {
-            log.debug("Query and fetch results do not match, normalization processor is skipped");
-            return true;
-        }
-        return false;
+        return queryPhaseResultConsumer.getAtomicArray().asList().stream().filter(Objects::nonNull).noneMatch(this::isHybridQuery);
     }
 
     /**
@@ -143,42 +130,5 @@ public class NormalizationProcessor implements SearchPhaseResultsProcessor {
     ) {
         Optional<Result> optionalFirstSearchPhaseResult = searchPhaseResults.getAtomicArray().asList().stream().findFirst();
         return optionalFirstSearchPhaseResult.map(SearchPhaseResult::fetchResult);
-    }
-
-    private boolean shouldSkipProcessorDueToIncompatibleQueryAndFetchResults(
-        final List<QuerySearchResult> querySearchResults,
-        final Optional<FetchSearchResult> fetchSearchResultOptional
-    ) {
-        if (fetchSearchResultOptional.isEmpty()) {
-            return false;
-        }
-        final List<Integer> docIds = unprocessedDocIds(querySearchResults);
-        SearchHits searchHits = fetchSearchResultOptional.get().hits();
-        SearchHit[] searchHitArray = searchHits.getHits();
-        // validate the both collections are of the same size
-        if (Objects.isNull(searchHitArray)) {
-            log.info("array of search hits in fetch phase results is null");
-            return true;
-        }
-        if (searchHitArray.length != docIds.size()) {
-            log.info(
-                String.format(
-                    Locale.ROOT,
-                    "number of documents in fetch results [%d] and query results [%d] is different",
-                    searchHitArray.length,
-                    docIds.size()
-                )
-            );
-            return true;
-        }
-        return false;
-    }
-
-    private List<Integer> unprocessedDocIds(final List<QuerySearchResult> querySearchResults) {
-        return querySearchResults.isEmpty()
-            ? List.of()
-            : Arrays.stream(querySearchResults.get(0).topDocs().topDocs.scoreDocs)
-                .map(scoreDoc -> scoreDoc.doc)
-                .collect(Collectors.toList());
     }
 }
--- a/src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessor.java
@@ -9,6 +9,7 @@ import static org.opensearch.neuralsearch.search.util.HybridSearchResultFormatUt
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Locale;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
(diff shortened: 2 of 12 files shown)


navneet1v:
Update the PR description to:
Fixing multiple issues reported in #497.

same for the change log.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Features
 ### Enhancements
 ### Bug Fixes
+- Multiple identical subqueries in Hybrid query  ([#524](https://github.com/opensearch-project/neural-search/pull/524))
```
navneet1v, line 23:
```suggestion
- Fixing multiple issues as reported in #497.  ([#524](https://github.com/opensearch-project/neural-search/pull/524))
```

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessor.java
```diff
@@ -98,7 +101,16 @@ private <Result extends SearchPhaseResult> boolean shouldSkipProcessor(SearchPha
...
-        return queryPhaseResultConsumer.getAtomicArray().asList().stream().filter(Objects::nonNull).noneMatch(this::isHybridQuery);
+        if (queryPhaseResultConsumer.getAtomicArray().asList().stream().filter(Objects::nonNull).noneMatch(this::isHybridQuery)) {
+            return true;
+        }
+        List<QuerySearchResult> querySearchResults = getQueryPhaseSearchResults(searchPhaseResult);
+        Optional<FetchSearchResult> fetchSearchResult = getFetchSearchResults(searchPhaseResult);
+        if (shouldSkipProcessorDueToIncompatibleQueryAndFetchResults(querySearchResults, fetchSearchResult)) {
+            log.debug("Query and fetch results do not match, normalization processor is skipped");
+            return true;
+        }
```
navneet1v, line 112: is this the case when Query and Fetch results are not compatible but hybrid query clause is there?

if yes, then I think we should throw the exception rather than just doing pass through. This will lead users getting confused if normalization is working or not.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessor.java
```diff
@@ -131,4 +143,29 @@ private <Result extends SearchPhaseResult> Optional<FetchSearchResult> getFetchS
...
+        if (fetchSearchResultOptional.isEmpty()) {
+            return false;
+        }
+        final List<Integer> docIds = unprocessedDocIds(querySearchResults);
+        SearchHits searchHits = fetchSearchResultOptional.get().hits();
+        SearchHit[] searchHitArray = searchHits.getHits();
+        // validate the both collections are of the same size
+        if (Objects.isNull(searchHitArray) || searchHitArray.length != docIds.size()) {
+            return true;
+        }
```
navneet1v, line 160: Lets add logs for this. Make it info logs so that we know what happened.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -173,8 +173,15 @@ private SearchHit[] getSearchHits(final List<Integer> docIds, final FetchSearchR
         SearchHits searchHits = fetchSearchResult.hits();
         SearchHit[] searchHitArray = searchHits.getHits();
         // validate the both collections are of the same size
-        if (Objects.isNull(searchHitArray) || searchHitArray.length != docIds.size()) {
-            throw new IllegalStateException("Score normalization processor cannot produce final query result");
+        if (Objects.isNull(searchHitArray)) {
+            throw new IllegalStateException(
```
navneet1v, line 177: do we need this? when we are doing the same thing in this function shouldSkipProcessorDueToIncompatibleQueryAndFetchResults?

src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
```diff
@@ -111,24 +112,34 @@ public float[] hybridScores() throws IOException {
...
-            float subScore = disiWrapper.scorer.score();
-            scores[queryToIndex.get(disiWrapper.scorer.getWeight().getQuery())] = subScore;
+            Query query = scorer.getWeight().getQuery();
+            List<Integer> indexes = queryToIndex.get(query);
+            // we need to find the index of first sub-query that hasn't been updated yet
+            int index = indexes.stream()
+                .mapToInt(idx -> idx)
+                .filter(index1 -> Float.compare(scores[index1], 0.0f) == 0)
+                .findFirst()
+                .orElseThrow(() -> new IllegalStateException("cannot collect score for subquery"));
```
navneet1v, line 126: Please make the exception error message more verbose.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
```diff
@@ -111,24 +112,34 @@ public float[] hybridScores() throws IOException {
...
-            if (disiWrapper.scorer.docID() == DocIdSetIterator.NO_MORE_DOCS) {
+            Scorer scorer = disiWrapper.scorer;
+            if (scorer.docID() == DocIdSetIterator.NO_MORE_DOCS) {
                 continue;
             }
-            float subScore = disiWrapper.scorer.score();
-            scores[queryToIndex.get(disiWrapper.scorer.getWeight().getQuery())] = subScore;
+            Query query = scorer.getWeight().getQuery();
+            List<Integer> indexes = queryToIndex.get(query);
+            // we need to find the index of first sub-query that hasn't been updated yet
```
navneet1v, line 121: who is updating the scores and how?

navneet1v:
Please add an IT for case when there are same queries in hyrbid query clause.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessor.java
martin-gaievski: Yes, at this moment we already checked if there is at least one result from hybrid query. 
Maybe I misunderstood the problem reported [in this issue](https://github.com/opensearch-project/neural-search/issues/497#issuecomment-1863499052). My understanding was that in such scenario with hybrid query there is a chance that fetch results have different number of docs comparing to query results. We currently are throwing exception in such case, like one that posted in that issue:

```
    "error": {
        "root_cause": [],
        "type": "search_phase_execution_exception",
        "reason": "The phase has failed",
        "phase": "query",
        "grouped": true,
        "failed_shards": [],
        "caused_by": {
            "type": "illegal_state_exception",
            "reason": "Score normalization processor cannot produce final query result"
        }
    },
    "status": 500
}
```
so the whole idea is still to throw an exception, just earlier?

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessor.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
martin-gaievski: yes, it's the same check. One thing is different - earlier in the processor code we're skipping execution, here we're throwing exception. Let's finalize on the approach mentioned in the earlier comment, do we skip processor or throw error. If we throw error I'll move these exceptions to processor

src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
martin-gaievski: score is set on the next line:
```
scores[index] = scorer.score()
```
let me add more on this to the comment

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessor.java
navneet1v: Yes, we are throwing exception but we don't know the root cause of the error. So when we do noop here, is the search getting failed going forward? When we were discussing this, I was under the impression that if there is some errors from fetch phase in that case we will fail after the processor execution. Is that happening  with this change?

martin-gaievski: I tried multiple variations of possible errors in fetch phase, like explicitly throwing exception and updating collection of hits with irrelevant data. 
In case of exception system was failing but it was before even hitting our code. I think that correlates with your expectations, but that happened even without any change from our side. 
In case of inconsistent fetch and query data there were no exception and result was always updated to match results of fetch phase. Before this change there was an exception reported by user, with this change we're silently skipping normalization and results are returned to user

martin-gaievski:
> Please add an IT for case when there are same queries in hyrbid query clause.

added IT as requested

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
martin-gaievski: removed the check from workflow class, let me know if we need to throw exceptions in processor class

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -173,8 +173,15 @@ private SearchHit[] getSearchHits(final List<Integer> docIds, final FetchSearchR
...
+                "score normalization processor cannot produce final query result, fetch query phase returns empty results"
+            );
+        }
+        if (searchHitArray.length != docIds.size()) {
+            throw new IllegalStateException(
+                "score normalization processor cannot produce final query result, the number of documents returned by fetch and query phases does not match"
```
navneet1v, line 183: can we add the lengths of the both the arrays in the message so that we know what is the mismatch.

src/test/java/org/opensearch/neuralsearch/search/HitsTresholdCheckerTests.java
```diff
@@ -27,7 +27,13 @@ public void testTresholdLimit_whenThresholdNegative_thenFail() {
         expectThrows(IllegalArgumentException.class, () -> new HitsThresholdChecker(-1));
     }
 
-    public void testTresholdLimit_whenThresholdMaxValue_thenFail() {
-        expectThrows(IllegalArgumentException.class, () -> new HitsThresholdChecker(Integer.MAX_VALUE));
+    public void testTrackThreshold_whenTrackThresholdSet_thenSuccessful() {
```
navneet1v, line 30: There is typo in the class name it should be HitsThresholdChecker and not HitsTresholdChecker

martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
martin-gaievski: ack

navneet1v (approved):

src/test/java/org/opensearch/neuralsearch/search/HitsTresholdCheckerTests.java
martin-gaievski: ack

src/test/java/org/opensearch/neuralsearch/query/HybridQueryIT.java
```diff
@@ -174,6 +174,75 @@ public void testComplexQuery_whenMultipleSubqueries_thenSuccessful() {
         assertEquals(RELATION_EQUAL_TO, total.get("relation"));
     }
 
+    /**
+     * Tests complex query with multiple nested sub-queries, where soem sub-queries are same
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
```diff
@@ -5,7 +5,10 @@
 
 package org.opensearch.neuralsearch.query;
 
+import static java.util.Locale.ROOT;
```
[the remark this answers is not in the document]
martin-gaievski: ack

heemin32 (approved):

PR #531, opened 2024-01-03
Update Developer guide with steps to run BWC Test Suite

vibrantvarun:
### Description
Update developer guide with steps to run BWC tests.

### Check List
    - [x] All tests pass


DEVELOPER_GUIDE.md
```diff
@@ -184,6 +186,25 @@ Additionally, it is possible to attach one debugger to the cluster JVM and anoth
 ./gradlew :integTest -Dtest.debug=1 -Dcluster.debug=1
 ```
 
+## Backwards Compatibility Testing
```
navneet1v, line 189: as per my understanding for Apple silicon chips and even for intel chips the BWC will not run. Can we add these details in the Developer guide.

DEVELOPER_GUIDE.md
```diff
@@ -184,6 +186,25 @@ Additionally, it is possible to attach one debugger to the cluster JVM and anoth
...
+
+The purpose of Backwards Compatibility Testing and different types of BWC tests are explained [here](https://github.com/opensearch-project/opensearch-plugins/blob/main/TESTING.md#backwards-compatibility-testing)
+
+Use these commands to run BWC tests for k-NN:
```
martin-gaievski, line 193: should be `neural-search`, not `knn`

DEVELOPER_GUIDE.md
```diff
@@ -184,6 +186,25 @@ Additionally, it is possible to attach one debugger to the cluster JVM and anoth
...
+1. Rolling upgrade tests: `./gradlew :qa:rolling-upgrade:testRollingUpgrade`
+2. Full restart upgrade tests: `./gradlew :qa:restart-upgrade:testAgainstNewCluster`
+3. `./gradlew :qa:bwcTestSuite` is used to run all the above bwc tests together.
```
martin-gaievski, line 196: for examples above we should mention what is the `bwc.version`. In this case this is default one, maybe put url of the file where that default version is defined

DEVELOPER_GUIDE.md
```diff
@@ -184,6 +186,25 @@ Additionally, it is possible to attach one debugger to the cluster JVM and anoth
...
+
+Use this command to run BWC tests for a given Backwards Compatibility Version:
+```
+./gradlew :qa:bwcTestSuite -Dbwc.version=2.9.0
+```
+Here, we are testing BWC Tests with BWC version of plugin as 2.9.0.
+
+### Adding new tests
+
+Before adding any new tests to Backward Compatibility Tests, we should be aware that the tests in BWC are not independent. While creating an index, a test cannot use the same index name if it is already used in other tests.
```
martin-gaievski, line 206: Maybe add a note that we do support only full restart and rolling upgrade tests, and when it makes sense to add test of each of those types.

vibrantvarun: We also support mixed cluster test. It is just I have combined it with rolling upgrade tests.

martin-gaievski: I think we should state this, that mixed cluster tests are combined with upgrade, and also some basic guidance when each type of test is needed

vibrantvarun: Basic guidance of when each type of test is needed is already been mentioned [here](https://github.com/opensearch-project/neural-search/pull/531/files#diff-d5555724c2ebca1b1f9be92a74c90e84c2dcbc76631148801024e24372b7edc6R191)

vibrantvarun: The mixed tests are a part of rolling upgrade as 
1) testAgainstOneThirdUpgraded Cluster
2) testAgainstTwoThiredUpgraded Cluster
3) Then Rolling upgrade tests runs

and the task of rolling upgrade depends on testAgainstTwoThiredUpgraded. So it is already quite understanding. Mentioning it seperately will confuse the people reading it.

martin-gaievski: To me that explanation in core is quite vague: `plugin functionalities work as expected after each node upgrade`. 
The whole purpose of adding a doc is explain the process and make it simpler for contributors. I'm trying to think from a new external contributor perspective, I would like to see something along these lines:
"When you're adding a new feature think of adding bwc test along with it. Ideally you need to add tests of all three types: mixed, upgrade, restart. With current design mixed and upgrade tests are combined and are part of the same test class (example). In some cases some test types are not possible (example), in such case try to do your best and add what's possible."

martin-gaievski (approved):

navneet1v (approved):

Issue #539, opened 2024-01-12
[BUG] Enable support for default model id on HybridQueryBuilder 


vibrantvarun:
### What is the bug?
When a neural query enricher processor is created to enable the support for default model id and the neural query clause is present at inner level of hybrid search query then the processor cannot add the model id in the neural clause.

### How can one reproduce the bug?
1) Create a Neural Query Enricher
2) Create a index and add the document
3) Create a search pipeline for (Normalization Processor)
4) Trigger hybrid query in the below shape

```
query{
     hybrid{
            queries[
                  neural {
                       // do not pass model id here as you expect that neural query enricher will add it here.
                  }
           ]  
     }
}

```
### What is the expected behavior?
Neural Query Enricher processor should be able to add the model id at inner level of hybrid query clause.

### Do you have any additional context?
The fix to the bug is 
Override the visit method from Querybuilder interface in HybridQueryBuilder and define it to the way that it parses the inner query builder.

PR #541, opened 2024-01-17
Enable support for default model id on HybridQueryBuilder

vibrantvarun:
### Description
This PR enables the support for Default model Id in Hybrid Query Builder.

Premise:
A customer is facing an issue while using the feature of default model id when neural query is a part of hybrid query.
See example below
```
query{
     hybrid{
            queries[
                  neural {
                       // do not pass model id here as you expect that neural query enricher will add it here.
                  }
           ]  
     }
}
```

```
{
  "from" : 0,
  "size" : 1000,
  "query": {
    "hybrid": {
      "queries": [
        {
          "match": {
            "passage": {
              "query": "how many camels are there in the world."
            }
          }
        },
        {
          "neural": {
            "content_vector": {
              "query_text": "how many camels are there in the world.",
              "k": 10,
              "filter" :
              {
                "bool" :
                {
                  "must": [
                    {
                      "term": {
                      "publisher_id": {
                        "value": "abc"
                        }
                      }
                    },
                    {
                      "term": {
                      "published": {
                        "value": true
                        }
                      }
                    }
                  ]
                }
              }
            }
          }
        }
      ]
    }
  },
  "_source": true
}
```
Reason:
NeuralQueryEnricher processor triggers a neural query visitor to visit all the querybuilders present in the search query and find the neuralquerybuilder and update the model id in it if not present. Visitor finds the visit method in the querybuilder and parses them. However, in hybridquerybuilder the definition of visit method was not there, therefore it was not able to parse the inner sub queries in it.

### Issues Resolved
[https://github.com/opensearch-project/neural-search/issues/539](https://github.com/opensearch-project/neural-search/issues/539)

### Check List
    - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
@@ -13,6 +13,7 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.lucene.search.BooleanClause.Occur;
 import org.apache.lucene.search.Query;
 import org.opensearch.common.lucene.search.Queries;
 import org.opensearch.core.ParseField;
@@ -27,6 +28,7 @@ import org.opensearch.index.query.QueryRewriteContext;
 import org.opensearch.index.query.QueryShardContext;
 import org.opensearch.index.query.QueryShardException;
 import org.opensearch.index.query.Rewriteable;
+import org.opensearch.index.query.QueryBuilderVisitor;
 
 import lombok.Getter;
 import lombok.NoArgsConstructor;
@@ -294,5 +296,18 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
             }
         }).filter(Objects::nonNull).collect(Collectors.toList());
         return queries;
+    }
+
+    /**
+     * visit method to parse the HybridQueryBuilder by a visitor
+     * @return
+     */
+    @Override
+    public void visit(QueryBuilderVisitor visitor) {
+        visitor.accept(this);
+        QueryBuilderVisitor subVisitor = visitor.getChildVisitor(Occur.MUST);
+        for (QueryBuilder subQueryBuilder : queries) {
+            subQueryBuilder.visit(subVisitor);
+        }
     }
 }
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
@@ -304,6 +304,8 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
     @Override
     public void visit(QueryBuilderVisitor visitor) {
         visitor.accept(this);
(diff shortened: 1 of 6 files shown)


vibrantvarun:
Test method name testVisit has been kept in hybridQueryBuilder to maintain name consistency with OpenSearch
[https://github.com/opensearch-project/OpenSearch/pull/10110/files](https://github.com/opensearch-project/OpenSearch/pull/10110/files)

martin-gaievski:
can we add integ test that would fail without changes in this PR? It can be something like: processor is set with the model id, we use hybrid query with neural query as one of sub-queries, do not specify model id in the query. My understanding without this change that is failing with a "missing model id" exception

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -295,4 +297,16 @@ private Collection<Query> toQueries(Collection<QueryBuilder> queryBuilders, Quer
...
         return queries;
     }
+
+    /**
+     * visit method to parse the HybridQueryBuilder by a visitor
+     */
+    @Override
+    public void visit(QueryBuilderVisitor visitor) {
+        visitor.accept(this);
+        QueryBuilderVisitor subVisitor = visitor.getChildVisitor(Occur.MUST);
```
heemin32, line 307: Could you tell why only `MUST` but not others like `SHOULD` and `MUST_NOT`?

martin-gaievski:
> can we add integ test that would fail without changes in this PR? It can be something like: processor is set with the model id, we use hybrid query with neural query as one of sub-queries, do not specify model id in the query. My understanding without this change that is failing with a "missing model id" exception. We can even extend this to a re-indexing scenario, where I suppose similar error will be in reindexing call

discussed offline with @vibrantvarun, test is already a [part of this PR](https://github.com/vibrantvarun/neural-search/blob/HybridQuery_SupportForEnablingDefaultModelId/src/test/java/org/opensearch/neuralsearch/processor/NeuralQueryEnricherProcessorIT.java#L67)

martin-gaievski (approved):

vibrantvarun:
As per discussion with @martin-gaievski offline, 
I have tested the scenario where the test(testNeuralQueryEnricherProcessor_whenHybridQueryBuilderAndNoModelIdPassed_thenSuccess) added in NeuralQueryEnricherProcessorIT was failing with an error("Model cannot be non-null" on my local machine If I do not add a visit method in the HybridQueryBuilder. 

Therefore after adding the visit method the processor is able to parse inner subqueries of the hybrid query builder.

Thanks

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
heemin32: Could you add that in comment that The value can be anything because `NeuralSearchQueryVisitor.getChildVisitor` return `this` regardless the value?

heemin32 (approved):

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
vibrantvarun: Sure

vibrantvarun:
This PR merge is dependent on [https://github.com/opensearch-project/security/pull/3959](https://github.com/opensearch-project/security/pull/3959)

Issue #384, opened 2023-10-03
[BUG] Flaky integ test testBooleanQuery_withNeuralAndBM25Queries, testBasicQuery


martin-gaievski:
### What is the bug?
Flaky integ test

### How can one reproduce the bug?
In github CI for the plugin you can see that tests for windows are not stable and failing at random. 

Example of such failed run:

https://github.com/opensearch-project/neural-search/actions/runs/6397635644/job/17366751285?pr=359

```
Tests with failures:
 - org.opensearch.neuralsearch.query.NeuralQueryIT.testBooleanQuery_withNeuralAndBM25Queries
 - org.opensearch.neuralsearch.query.NeuralQueryIT.testBasicQuery


=== Standard output of node `node{::integTest-0}` ===
�    ? errors and warnings from D:\a\neural-search\neural-search\build\testclusters\integTest-0\logs\opensearch.stdout.log ?
� WARN ][o.o.g.DanglingIndicesState] [integTest-0] gateway.auto_import_dangling_indices is disabled, dangling indices will not be automatically detected or imported and must be managed manually
� WARN ][o.o.d.FileBasedSeedHostsProvider] [integTest-0] expected, but did not find, a dynamic hosts list at [D:\a\neural-search\neural-search\build\testclusters\integTest-0\config\unicast_hosts.txt]
� WARN ][r.suppressed             ] [integTest-0] path: /_plugins/_ml/models/3XAr94oB_1pNJlKRXFYm, params: {model_id=3XAr94oB_1pNJlKRXFYm}
�  java.lang.Exception: Model cannot be deleted in deploying or deployed state. Try undeploy model first then delete
�  	at org.opensearch.ml.action.models.DeleteModelTransportAction.lambda$doExecute$2(DeleteModelTransportAction.java:128) [opensearch-ml-2.11.0.0-SNAPSHOT.jar:2.11.0.0-SNAPSHOT]
�  	at org.opensearch.core.action.ActionListener$1.onResponse(ActionListener.java:82) [opensearch-core-2.11.0-SNAPSHOT.jar:2.11.0-SNAPSHOT]
�  	at org.opensearch.ml.helper.ModelAccessControlHelper.validateModelGroupAccess(ModelAccessControlHelper.java:79) [opensearch-ml-2.11.0.0-SNAPSHOT.jar:2.11.0.0-SNAPSHOT]
[...]

PR #559, opened 2024-01-26
Fix for Flaky test for issue 384

vibrantvarun:
### Description
Currently, we are facing falkiness in the integ test behaviour of neural search because of following reasons.
1) We try to findAllDeployedModels at the end of the single test and try to delete it. However, if at the same time any other test from some other test class starts deploying model than we see below error due to multithreading environment
```
�  java.lang.Exception: Model cannot be deleted in deploying or deployed state. Try undeploy model first then delete
```
2) Moreover, we use the same index name in few classes such that it causes problems in multithreading environment.
3) Also, at some places we are not deleting the index after the test execution.

This PR address all the above issues.

### Issues Resolved
[384](https://github.com/opensearch-project/neural-search/issues/384)

### Check List
    - [x] All tests pass


martin-gaievski:
One BWC check is failing, please check if that's not for the same reason as we're trying to fix here

src/test/java/org/opensearch/neuralsearch/processor/NeuralQueryEnricherProcessorIT.java
```diff
@@ -29,20 +29,21 @@ public class NeuralQueryEnricherProcessorIT extends BaseNeuralSearchIT {
...
+        modelId = prepareModel();
     }
 
     @After
     @SneakyThrows
     public void tearDown() {
         super.tearDown();
         deleteSearchPipeline(search_pipeline);
-        findDeployedModels().forEach(this::deleteModel);
+        deleteModel(modelId);
```
martin-gaievski, line 46: do we need methods for deleting all model ids?

src/test/java/org/opensearch/neuralsearch/processor/ScoreCombinationIT.java
```diff
@@ -33,8 +33,8 @@
 import lombok.SneakyThrows;
 
 public class ScoreCombinationIT extends BaseNeuralSearchIT {
-    private static final String TEST_MULTI_DOC_INDEX_ONE_SHARD_NAME = "test-neural-multi-doc-one-shard-index";
-    private static final String TEST_MULTI_DOC_INDEX_THREE_SHARDS_NAME = "test-neural-multi-doc-three-shards-index";
+    private static final String TEST_MULTI_DOC_INDEX_ONE_SHARD_NAME = "test-score-combination-neural-multi-doc-one-shard-index";
+    private static final String TEST_MULTI_DOC_INDEX_THREE_SHARDS_NAME = "test-score-combination-neural-multi-doc-three-shards-index";
```
martin-gaievski, line 37: good catch, requests might collide in case different runners doing multiple tests in parallel

src/test/java/org/opensearch/neuralsearch/processor/NeuralQueryEnricherProcessorIT.java
```diff
@@ -29,20 +29,21 @@ public class NeuralQueryEnricherProcessorIT extends BaseNeuralSearchIT {
     private static final String ingest_pipeline = "nlp-pipeline";
     private static final String TEST_KNN_VECTOR_FIELD_NAME_1 = "test-knn-vector-1";
     private final float[] testVector = createRandomVector(TEST_DIMENSION);
+    private static String modelId;
 
     @Before
```
heemin32, line 34: Shouldn't this be `@BeforeClass`?

martin-gaievski (commented):
I think we're missing one piece in this - we're not using that model id that we read in setUp section. I see that in tests like this one we're reading model id: https://github.com/vibrantvarun/neural-search/blob/Flaky_test_384/src/test/java/org/opensearch/neuralsearch/processor/NormalizationProcessorIT.java#L94. We need to use that id we get in setUp method instead

vibrantvarun:
No it is not. It is due to sometimes when we push multiple commits back to back then due to resources sometimes tests fails.

src/test/java/org/opensearch/neuralsearch/processor/NeuralQueryEnricherProcessorIT.java
vibrantvarun: @BeforeClass runs once before the entire test fixture. If your test class has ten tests, @Before code will be executed ten times, but @BeforeClass will be executed only once. Here I need model Id to reinitialize after every test and a new value will be assigned.

src/test/java/org/opensearch/neuralsearch/processor/NeuralQueryEnricherProcessorIT.java
vibrantvarun: yeah I am removing it in next revision.

src/test/java/org/opensearch/neuralsearch/processor/NeuralQueryEnricherProcessorIT.java
heemin32: Then, static variable is dangerous. It shares the value across tests.

vibrantvarun: Okay then I will remove static and just declare the variable.
private String modelId;

heemin32: Same. You need to define the variable in each and every test method itself.

vibrantvarun: I see then I have to remove the setup method all together and and remove the teardown method as well.
I have to shift model Id code in tests and use wipeofTestResources method to clean up after every test.

heemin32: Yes.

src/test/java/org/opensearch/neuralsearch/processor/NeuralQueryEnricherProcessorIT.java
```diff
@@ -60,13 +49,13 @@ public void testNeuralQueryEnricherProcessor_whenNoModelIdPassed_thenSuccess() {
         Map<String, Object> response = search(index, neuralQueryBuilder, 2);
 
         assertFalse(response.isEmpty());
-
+        wipeOfTestResources(index, ingest_pipeline, modelId, search_pipeline);
```
heemin32, line 52: Should it be under `finally` block?

navneet1v:
@vibrantvarun what is pending on this PR?

src/test/java/org/opensearch/neuralsearch/processor/NeuralQueryEnricherProcessorIT.java
vibrantvarun: Added now

vibrantvarun:
@navneet1v @heemin32 advised to add finally block in adding wipeoftestResources. Therefore I added it and now raised a revision.

heemin32 (approved):

martin-gaievski (approved):
Nice rework for tests Varun

PR #562, opened 2024-01-26
Add validations from appsec

### LINKED ISSUE #555: Add validation for reranker processor parameters
For the [reranker processor](https://github.com/opensearch-project/neural-search/issues/248), there are no checks of number of context fields user can define for the processor when:

- defining processor as part of the search pipeline
- using `query_text_path` as part of search request

This brings risks of system instability in case very long string is passed in any of those two places.

I suggest we use some limit on number of fields for `document_fields`, let's start from 50 but that number is discussible. For the `query_text_path` we can limit the max level of nested field to not exceed one that set by `index.mapping.depth.limit` [setting](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/mapper/MapperService.java#L149).

HenryL27:
### Description
Adding validation for parameters of re-ranker processor.

### Issues Resolved
#555 

### Check List
    - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
@@ -11,7 +11,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.StringJoiner;
 
-import org.opensearch.env.Environment;
+import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.ingest.ConfigurationUtils;
 import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
 import org.opensearch.neuralsearch.processor.rerank.MLOpenSearchRerankProcessor;
@@ -38,7 +38,7 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
     public static final String CONTEXT_CONFIG_FIELD = "context";
 
     private final MLCommonsClientAccessor clientAccessor;
-    private final Environment environment;
+    private final ClusterService clusterService;
 
     @Override
     public SearchResponseProcessor create(
@@ -55,7 +55,7 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
             config,
             includeQueryContextFetcher,
             tag,
-            environment
+            clusterService
         );
         switch (type) {
             case ML_OPENSEARCH:
@@ -117,7 +117,7 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
             Map<String, Object> config,
             boolean includeQueryContextFetcher,
             String tag,
-            final Environment environment
+            final ClusterService clusterService
         ) {
             Map<String, Object> contextConfig = ConfigurationUtils.readMap(RERANK_PROCESSOR_TYPE, tag, config, CONTEXT_CONFIG_FIELD);
             List<ContextSourceFetcher> fetchers = new ArrayList<>();
@@ -125,14 +125,14 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
                 Object cfg = contextConfig.get(key);
                 switch (key) {
                     case DocumentContextSourceFetcher.NAME:
-                        fetchers.add(DocumentContextSourceFetcher.create(cfg, environment));
+                        fetchers.add(DocumentContextSourceFetcher.create(cfg, clusterService));
                         break;
                     default:
                         throw new IllegalArgumentException(String.format(Locale.ROOT, "unrecognized context field: %s", key));
                 }
             }
             if (includeQueryContextFetcher) {
-                fetchers.add(new QueryContextSourceFetcher(environment));
+                fetchers.add(new QueryContextSourceFetcher(clusterService));
             }
             return fetchers;
         }
--- a/src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
@@ -11,6 +11,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.StringJoiner;
 
+import org.opensearch.env.Environment;
 import org.opensearch.ingest.ConfigurationUtils;
 import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
 import org.opensearch.neuralsearch.processor.rerank.MLOpenSearchRerankProcessor;
@@ -37,6 +38,7 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
     public static final String CONTEXT_CONFIG_FIELD = "context";
 
     private final MLCommonsClientAccessor clientAccessor;
+    private final Environment environment;
 
     @Override
     public SearchResponseProcessor create(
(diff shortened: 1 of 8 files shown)


src/main/java/org/opensearch/neuralsearch/processor/rerank/context/QueryContextSourceFetcher.java
```diff
@@ -107,4 +124,11 @@ private static Map<String, Object> requestToMap(final SearchRequest request) thr
         Map<String, Object> map = parser.map();
         return map;
     }
+
+    private boolean validatePath(final String path) {
```
martin-gaievski, line 128: can we add a validation on a string length, something like >= 1000 characters is not valid? With current logic we first split the string into tokens, then validate. If there is in malicious input such operation may cause overflow before being stopped by a validator

src/main/java/org/opensearch/neuralsearch/processor/rerank/context/QueryContextSourceFetcher.java
```diff
@@ -76,9 +78,10 @@ public void fetchContext(
                     throw new IllegalArgumentException(
                         String.format(
                             Locale.ROOT,
-                            "%s exceeded the maximum path length of %d",
+                            "%s exceeded the maximum path length of %d nested fields or %d characters",
```
martin-gaievski, line 81: I think we need to keep one exception per one failed validation. Can we make throwing exception a part of `validatePath` method, and throw exception with specific reason for each of two cases? 
One more ask - please add log at `error` level with exact value entered by user.

src/main/java/org/opensearch/neuralsearch/processor/rerank/context/QueryContextSourceFetcher.java
```diff
@@ -125,10 +120,32 @@ private static Map<String, Object> requestToMap(final SearchRequest request) thr
...
+                String.format(
+                    Locale.ROOT,
+                    "%s exceeded the maximum path length of %d nested fields",
+                    QUERY_TEXT_PATH_FIELD,
+                    MapperService.INDEX_MAPPING_DEPTH_LIMIT_SETTING.get(environment.settings())
+                )
+            );
         }
-        return path.split("\\.").length <= MapperService.INDEX_MAPPING_DEPTH_LIMIT_SETTING.get(environment.settings());
+        return;
```
martin-gaievski, line 149: nit: we don't need this return

martin-gaievski (approved):

heemin32:
Could you add which validations is this PR added in the description?

src/test/java/org/opensearch/neuralsearch/processor/rerank/MLOpenSearchRerankProcessorTests.java
```diff
@@ -223,6 +230,49 @@ public void testRerankContext_whenQueryTextPathIsBadPointer_thenFail() throws IO
             .equals(QueryContextSourceFetcher.QUERY_TEXT_PATH_FIELD + " must point to a string field"));
     }
 
+    public void testRerankContext_whenQueryTextPathIsExceeedinglyManyCharacters_thenFail() throws IOException {
```
heemin32, line 233: Could you use `@SneakyThrows` for test case? We haven't documented publicly but we want to use `@SneakyThrows` in test case and `throws *Exception` in others.
```suggestion
    @SneakyThrows
    public void testRerankContext_whenQueryTextPathIsExceeedinglyManyCharacters_thenFail() {
```

src/test/java/org/opensearch/neuralsearch/processor/rerank/MLOpenSearchRerankProcessorTests.java
```diff
@@ -223,6 +230,49 @@ public void testRerankContext_whenQueryTextPathIsBadPointer_thenFail() throws IO
...
+                    QueryContextSourceFetcher.QUERY_TEXT_PATH_FIELD,
+                    QueryContextSourceFetcher.MAX_QUERY_PATH_STRLEN
+                )
+            ));
+    }
+
+    public void textRerankContext_whenQueryTextPathIsExceeedinglyDeeplyNested_ThenFail() throws IOException {
+        setupParams(Map.of(QueryContextSourceFetcher.QUERY_TEXT_PATH_FIELD, "a.b.c.d.e.f.g.h.i.j.k.l.m.n.o.p.q.r.s.t.w.x.y.z"));
+        setupSearchResults();
+        @SuppressWarnings("unchecked")
```
heemin32, line 258: Why do we need this annotation? I don't see `ActionListener<Map<String, Object>> listener = mock(ActionListener.class);` is causing any warning in my local.

src/test/java/org/opensearch/neuralsearch/processor/rerank/MLOpenSearchRerankProcessorTests.java
```diff
@@ -223,6 +230,49 @@ public void testRerankContext_whenQueryTextPathIsBadPointer_thenFail() throws IO
...
+                String.format(
+                    Locale.ROOT,
+                    "%s exceeded the maximum path length of %d characters",
+                    QueryContextSourceFetcher.QUERY_TEXT_PATH_FIELD,
+                    QueryContextSourceFetcher.MAX_QUERY_PATH_STRLEN
+                )
+            ));
+    }
+
+    public void textRerankContext_whenQueryTextPathIsExceeedinglyDeeplyNested_ThenFail() throws IOException {
```
heemin32, line 255:
```suggestion
    @SneakyThrows
    public void textRerankContext_whenQueryTextPathIsExceeedinglyDeeplyNested_thenFail() {
```

src/test/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactoryTests.java
```diff
@@ -187,4 +197,26 @@ public void testCrossEncoder_whenEmptyContextDocField_thenFail() {
         );
     }
 
+    public void testCrossEncoder_whenTooManyDocFields_ThenFail() {
```
heemin32, line 200:
```suggestion
    public void testCrossEncoder_whenTooManyDocFields_thenFail() {
```

src/main/java/org/opensearch/neuralsearch/processor/rerank/context/DocumentContextSourceFetcher.java
```diff
@@ -29,8 +32,10 @@ public class DocumentContextSourceFetcher implements ContextSourceFetcher {
 
     public static final String NAME = "document_fields";
     public static final String DOCUMENT_CONTEXT_LIST_FIELD = "document_context_list";
+    public static final int MAX_DOCUMENT_FIELDS = 50;
```
heemin32, line 35: I think this is not being used?
```suggestion

```

src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
```diff
@@ -37,6 +38,7 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
     public static final String CONTEXT_CONFIG_FIELD = "context";
 
     private final MLCommonsClientAccessor clientAccessor;
+    private final Environment environment;
```
heemin32, line 41: Please check if you can dynamically update the plugins.neural_search.reranker_max_document_fields. 
I think you need to use `ClusterSetting` rather than environment. Environment only reads setting from opensearch.yml file.

HenryL27: ack

heemin32 (approved):

HenryL27:
@martin-gaievski do we need to backport this?

martin-gaievski:
Yes, it will be backported later, along with other PRs for this feature. Idea is that all changes go to feature branch, once all steps are finalized it's merged to main in a separate PR and then backported from there to other branches like 2.x.

HenryL27:
oh right, forgot this was on a feature branch lol

PR #614, opened 2024-02-26
[FEATURE] support default model id in neural_sparse query

### LINKED ISSUE #610: [FEATURE] support default model id in `neural_sparse` query
Currently, users need to provide the `model_id` in `neural_sparse` query. This is not convinient and requires more effort to maintain the model id. We should support default model id in `neural_sparse` query just like neural query and hybrid query.

ref: #70 #337

zhichao-aws:
### Description
see #610 
In this PR we support default model_id for neural_sparse query. Existing code use check the class of queryBuilder in the visitor([ref](https://github.com/opensearch-project/neural-search/blob/ea49d3c5006efff9dfa36e69791ae9a8e468d25a/src/main/java/org/opensearch/neuralsearch/query/visitor/NeuralSearchQueryVisitor.java#L31)), use modelId method to set the default model id for all NeuralQueryBuilder. This PR create a new interface called ModelInferenceQueryBuilder, it has the common method for setting default model id. In this PR, we let the NeuralQueryBuilder and NeuralSparseQueryBuilder implements the ModelInferenceQueryBuilder, and in the visitor we'll check and set default model id for ModelInferenceQueryBuilder. The unit test, integration test and bwc test can cover the changes in this PR, and make sure it doesn't have impact on existing functionality.

### Issues Resolved
#610 

### Check List
    - [ ] All tests pass


zhichao-aws:
This PR also contains the content in #615. To make a clear commit history, please merge #615 first and I'll rebase main again.

vibrantvarun:
@zhichao-aws We have fixed the build on main. Can you rebase and try running failing tests again? Thanks

zhichao-aws:
rebase on #615

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -94,7 +101,11 @@ public NeuralSparseQueryBuilder(StreamInput in) throws IOException {
     protected void doWriteTo(StreamOutput out) throws IOException {
         out.writeString(fieldName);
         out.writeString(queryText);
-        out.writeString(modelId);
+        if (isClusterOnOrAfterMinReqVersionForDefaultModelIdSupport()) {
+            out.writeOptionalString(this.modelId);
+        } else {
+            out.writeString(this.modelId);
```
zane-neo, line 107: If the cluster version smaller than min version, shouldn't here be `writeOptionalString` since the modelId is null?

zhichao-aws: If the cluster version is before min version (i.e. 2.13), the model id field is required, it's not optional. So it can not be null and we can use writeString here

src/main/java/org/opensearch/neuralsearch/query/ModelInferenceQueryBuilder.java
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.query;
+
+/**
+ * Query builders which calls ml-commons API to do model inference.
+ * The model inference result is used for search on target field.
+ */
+
```
chishui, line 11: nit: empty line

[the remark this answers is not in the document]
zhichao-aws: ack

src/main/java/org/opensearch/neuralsearch/query/ModelInferenceQueryBuilder.java
```diff
@@ -0,0 +1,27 @@
...
+public interface ModelInferenceQueryBuilder {
+    /**
+     * Get the model id used by ml-commons model inference. Return null if the model id is absent.
+     */
+    public String modelId();
+
+    /**
+    * Set a new model id for the query builder.
```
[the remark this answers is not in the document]
zhichao-aws: ack

src/test/java/org/opensearch/neuralsearch/processor/NeuralQueryEnricherProcessorIT.java
```diff
@@ -9,24 +9,34 @@
...
+import org.opensearch.common.xcontent.XContentType;
+import org.opensearch.core.rest.RestStatus;
 import org.opensearch.neuralsearch.BaseNeuralSearchIT;
 import org.opensearch.neuralsearch.query.HybridQueryBuilder;
 import org.opensearch.neuralsearch.query.NeuralQueryBuilder;
 
 import com.google.common.primitives.Floats;
 
 import lombok.SneakyThrows;
+import org.opensearch.neuralsearch.query.NeuralSparseQueryBuilder;
```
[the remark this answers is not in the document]
zhichao-aws: ack

src/main/java/org/opensearch/neuralsearch/query/ModelInferenceQueryBuilder.java
```diff
@@ -0,0 +1,27 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.query;
+
+/**
+ * Query builders which calls ml-commons API to do model inference.
+ * The model inference result is used for search on target field.
+ */
+
+public interface ModelInferenceQueryBuilder {
```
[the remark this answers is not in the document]
zhichao-aws: Makes sense. @navneet1v @jmazanec15 @zane-neo Do you have any preferences or concerns for these two implementation?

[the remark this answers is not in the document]
zhichao-aws: If we create a new abstract class, we can extract the common attributes of fieldName, queryText, and modelId. And we can also wrap the serialize/deserialize of these fields. It seems more close to the mental model of these 2 query clauses.

The interface way requires less effort, and it has smaller risk in implementation. 

I'm open to both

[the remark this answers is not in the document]
zane-neo: Agree with the first option since the interface isn't add any concrete functionality to the class, in this case we shouldn't use interface and all the methods are retrieving attributes of a class(Builder), so we should have a class to hold these attributes and make it reusable.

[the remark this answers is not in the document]
martin-gaievski: We do have mechanism for handling default model id via visitor in the QueryBuilder. Please check [this PR](https://github.com/opensearch-project/neural-search/pull/541) where support for default model id has been added to hybrid query. Can we use same approach here, or if not - please explain why.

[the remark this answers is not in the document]
zhichao-aws: This PR support default model id of neural query nested inside hybrid query. The prerequisite is neural query default model id has already been supported. Current code check the class and support NeuralQueryBuilder only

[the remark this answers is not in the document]
zhichao-aws: https://github.com/opensearch-project/neural-search/blob/ea49d3c5006efff9dfa36e69791ae9a8e468d25a/src/main/java/org/opensearch/neuralsearch/query/visitor/NeuralSearchQueryVisitor.java#L31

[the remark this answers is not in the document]
vibrantvarun: Hey @zhichao-aws I think we do not need this interface. Neither we need an abstract class. I would suggest you to just create a new visitor.

Just like NeuralSearchQueryVisitor
Create one NeuralSparseSearchQueryVisitor.

NeuralQueryBuilder and NeuralSparseQueryBuilder has different purposes. In future, it might happen parameters change  in either of them. So it should be kept independent.

We have throughly tested NeuralSearchQueryVisitor. Making changes in that design might affect functionality.

So it is better if we create a NeuralSparseQueryVisitor. 

Also there are certain things to note

Every visitor has a getChildVisitor method which has Occur occur as a argument.

Tomorrow just for NeuralQueryBuilder if cx asks for a certain thing to happen at Occur.Must then we can restrict that change to just NeuralQuery. This will not impact NeuralSparseQueryBuilder. Therefore, keeping them independent is for greater good.


cc: @martin-gaievski @navneet1v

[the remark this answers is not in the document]
zhichao-aws: Hi @vibrantvarun , I have some concerns to implement a dedicated new visitor will share all codes with the current one.
1. For current implementation, a dedicated new visitor will share all codes with the current one, except the class checker. This will make our code base redundant. And bring more maintaining efforts. If we want to introduce new feature or fix bugs, we need to add code in both places.
2. Although NeuralQueryBuilder and NeuralSparseQueryBuilder has different purposes, they have much in common. Using an interface to extract the shared part is consistent with the mental model. The parameters may change for each query builder, but I don't think we'll change the most basic fields (in this PR, modelId and fieldName). Even if they are changed, we can still add more logics here to distinguish these two query builders.
For the risk point, I think the interface way is safe enough. We can regard it as a marker of these 2 classes, this will minimize the impact to existing logics. We have many test cases to ensure the functionality of existing features (I can also add new tests if necessary).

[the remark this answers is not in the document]
vibrantvarun: Please add BWC tests.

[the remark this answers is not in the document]
navneet1v: +1

zhichao-aws:
@vibrantvarun @navneet1v Have added bwc tests for neural_query_enricher. Please help review thanks!

The test covers the version from 2.9 to 2.13.0-SNAPSHOT. I don't check the behavior of neural_query_enricher in 2.13.0-SNAPSHOT now, because the behavior will change after this PR get backport to 2.x. Will raise another PR after backport.

navneet1v:
@vibrantvarun please review the PR as you have most context on this PR.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
```diff
@@ -0,0 +1,150 @@
...
+    private static final String TEST_TEXT_FIELD = "passage_text";
+    private static final String TEXT_1 = "Hello world a b";
+
+    // Test restart-upgrade neural_query_enricher in restart-upgrade scenario
+    public void testNeuralQueryEnricherProcessor_NeuralSparseSearch_E2EFlow() throws Exception {
+        waitForClusterHealthGreen(NODES_BWC_CLUSTER);
+        Version bwcVersion = parseVersionFromString(getBWCVersion().get());
+
+        // skip this test before we have neural_sparse search in 2.11
+        if (bwcVersion.before(Version.V_2_11_0)) {
```
vibrantvarun, line 38: We don't need this check as this can be covered here 

1.) https://github.com/opensearch-project/neural-search/blob/ea49d3c5006efff9dfa36e69791ae9a8e468d25a/qa/restart-upgrade/build.gradle#L69

2.) https://github.com/opensearch-project/neural-search/blob/ea49d3c5006efff9dfa36e69791ae9a8e468d25a/qa/restart-upgrade/build.gradle#L102

3)https://github.com/opensearch-project/neural-search/blob/ea49d3c5006efff9dfa36e69791ae9a8e468d25a/qa/rolling-upgrade/build.gradle#L103

4) https://github.com/opensearch-project/neural-search/blob/ea49d3c5006efff9dfa36e69791ae9a8e468d25a/qa/rolling-upgrade/build.gradle#L136

5)https://github.com/opensearch-project/neural-search/blob/ea49d3c5006efff9dfa36e69791ae9a8e468d25a/qa/rolling-upgrade/build.gradle#L169

Please remove all the version checks in the restart and rolling upgrade tests as you just add 1 entry of NeuralQueryEnricherProcessorIT in the places mentioned above.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
```diff
@@ -0,0 +1,150 @@
...
+            return;
+        }
+
+        String modelId = null;
```
vibrantvarun, line 42: You can avoid declaring modelId here. Instead declare it at line 50 of this line and later at line 78 you can again declare as String modelId= TestUtils.getModelId(getIngestionPipeline(SPARSE_INGEST_PIPELINE_NAME), SPARSE_ENCODING_PROCESSOR);

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
```diff
@@ -0,0 +1,150 @@
...
+
+            addSparseEncodingDoc(getIndexNameForTest(), "0", List.of(), List.of(), List.of(TEST_TEXT_FIELD), List.of(TEXT_1));
+
+            createSearchRequestProcessor(modelId, SPARSE_SEARCH_PIPELINE_NAME);
+            updateIndexSettings(
+                getIndexNameForTest(),
+                Settings.builder().put("index.search.default_pipeline", SPARSE_SEARCH_PIPELINE_NAME)
+            );
+
+            if (bwcVersion.onOrAfter(Version.V_2_13_0)) {
```
vibrantvarun, line 68: Please remove version and refer the comment above

vibrantvarun:
Please add a clear cut description of solution in detail on what exactly we are trying to achieve in this PR.

1.) What is the solution?
2.) Testability
3.) What existing work you are making a change in.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
zhichao-aws: Ack. Will add checks in build script and remove the versions check in test cases.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
zhichao-aws: This version check is different from the above one. The above version check is used to avoid running tests before we introduce neural_query_enricher and can be moved to build.gradle.
This version check is used to check the different behavior of neural_sparse default model id in 2.13-SNAPSHOT, but this can only happen after we backport this PR to 2.x. I'll create another PR to add checks after we finish backport.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -58,7 +58,7 @@
 @Accessors(chain = true, fluent = true)
 @NoArgsConstructor
 @AllArgsConstructor
-public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder> {
+public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder> implements ModelInferenceQueryBuilder {
```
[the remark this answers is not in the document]
zhichao-aws: Here we use lombok.Setter and lombok.Getter for NeuralQueryBuilder and NeuralSparseQueryBuilder. It will generate the set&get method automatically for all fields.

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilderTests.java
```diff
@@ -177,6 +187,30 @@ public void testFromXContent_whenBuildWithMissingModelId_thenFail() {
...
+    @SneakyThrows
+    public void testFromXContent_whenBuildWithMissingModelIdInOldVersion_thenFail() {
+        /*
+          {
+              "VECTOR_FIELD": {
+                "query_text": "string"
+              }
+          }
+        */
+        setUpClusterService(Version.V_2_12_0);
```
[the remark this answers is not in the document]
zhichao-aws: Here I want to test the xContentParser for NeuralSparseQueryBuilder, and it's behavior for modelId is different in different versions. The cluster version exposed to  xContentParser is from the mocked ClusterService in NeuralSearchClusterUtil, and the concrete version number can be set here.

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -1057,13 +1058,28 @@ protected void wipeOfTestResources(
...
+                deleteModel(modelId);
+            }
         }
         if (indexName != null) {
             deleteIndex(indexName);
         }
     }
 
+    // parse the version from string like: "2.12.0","2.13.0-SNAPSHOT"
+    protected Version parseVersionFromString(String version) {
```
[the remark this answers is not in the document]
zhichao-aws: We'll need this to test the behavior of default model_id in neural_sparse query in 2.13-SNAPSHOT.

zhichao-aws:
> Please add a clear cut description of solution in detail on what exactly we are trying to achieve in this PR.
> 
> 1.) What is the solution? 2.) Testability 3.) What existing work you are making a change in.

In this PR we support default model_id for neural_sparse query. Existing code use check the class of queryBuilder in the visitor([ref](https://github.com/opensearch-project/neural-search/blob/ea49d3c5006efff9dfa36e69791ae9a8e468d25a/src/main/java/org/opensearch/neuralsearch/query/visitor/NeuralSearchQueryVisitor.java#L31)), use `modelId` method to set the default model id for all `NeuralQueryBuilder`. This PR create a new interface called `ModelInferenceQueryBuilder`, it has the common method for setting default model id. In this PR, we let the `NeuralQueryBuilder` and `NeuralSparseQueryBuilder` implements the `ModelInferenceQueryBuilder`, and in the visitor we'll check and set default model id for `ModelInferenceQueryBuilder`.  The unit test, integration test and bwc test can cover the changes in this PR, and make sure it doesn't have impact on existing functionality.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
```diff
@@ -0,0 +1,125 @@
...
+                // after we support default model id in neural_sparse query
+                // do nothing here. need to add test codes after finishing backport
+                ;
+            } else {
+                // before we support default model id in neural_sparse query
+                expectThrows(ResponseException.class, () -> search(getIndexNameForTest(), sparseEncodingQueryBuilderWithoutModelId, 1));
+            }
+        } else {
+            String modelId = "";
```
vibrantvarun, line 70: You have declared modelId with an empty string here. In wipeoftestresources method we check if modelId!=null we delete it. Here as per your code it will always try to delete it. Please declare this null here. Take reference from other similar test files

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
vibrantvarun: Can you explain the scenario of what exactly that test case would be? AFAIK, you have made change in NeuralSparseQueryBuilder for reading and writing the Xcontent object so it should work fine.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
zhichao-aws: ack

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
zhichao-aws: Assume we have finished backport here.
In restart scenerio, if the bwc-version is 2.13, we should be able to use default model_id for old version cluster. While it should throw exception for 2.11 and 2.12. 
In rolling-upgrade scenerio, if the bwc-version is 2.13, we should always be able to use default model_id for all cluster states. But for 2.11 and 2.12, we can only use default model id for updated cluster.

We can make a double check in BWC test to make sure the behavior is same as our expectation. And for future  code changes, unexpected modifications can be avoided by sufficient tests.

vibrantvarun: I think this double check is not needed. Because 

Consider bwc version is 2.13 and and the current version is 2.14
then if you don't create neural query enricher processor to enable default model id then it will return model Id cannot be null as per below code added in NeuralSparseQueryBuilder

```
if (!isClusterOnOrAfterMinReqVersionForDefaultModelIdSupport()) {
            requireValue(
                sparseEncodingQueryBuilder.modelId(),
                String.format(Locale.ROOT, "%s field must be provided for [%s] query", MODEL_ID_FIELD.getPreferredName(), NAME)
            );
        }
```

Now consider bwc version is 2.11 and the current version is 2.13

when the cluster is running at 2.11 and you created a neural query enricher processor then 
if restart upgrade tests are running then it will upgrade nodes to 2.13 and therefore it will support default model id functionality.

if rolling upgrade scenario is there then it might work or it might not work as it totally depends on which node it makes the search request.

Now consider bwc version is 2.11 and the current version is 2.12
then if you don't create neural query enricher processor to enable default model id then it will return model Id cannot be null as per below code added in NeuralSparseQueryBuilder

```
if (!isClusterOnOrAfterMinReqVersionForDefaultModelIdSupport()) {
            requireValue(
                sparseEncodingQueryBuilder.modelId(),
                String.format(Locale.ROOT, "%s field must be provided for [%s] query", MODEL_ID_FIELD.getPreferredName(), NAME)
            );
        }
```

Therefore, all you have to do is when the current Version is on or after 2.13 and the bwc version is below 2.13 then you have not run this particular test in bwc. Just return. I that version check should be just in rolling upgrade tests 

     if (bwcVersion.onOrBefore(Version.V_2_13_0)) {
        then return;
   }

vibrantvarun: Even you can add a check in build.gradle of rolling upgrade tests  that to not run this particular test when bwc version is 2.11 and 2.12

vibrantvarun: You can avoid the check altogether

zhichao-aws: Current version can not be version before 2.13, because we introduce these test code to main and backport to 2.13.

> Consider bwc version is 2.13 and and the current version is 2.14
then if you don't create neural query enricher processor to enable default model id then it will return model Id cannot be null as per below code added in NeuralSparseQueryBuilder
But we do have created  neural query enricher processor. The processor is introduced from 2.11, and it cannot for neural sparse in 2.11 and 2.12, while in 2.13 it works for neural sparse. And the version check here is used to distinguish this behavior difference.

> Now consider bwc version is 2.11 and the current version is 2.13
If the current version is 2.13, then the bwc version can not be onOrAfter 2.13, and this line will not be triggered. This line is designed for current version after 2.13.

zhichao-aws: > Even you can add a check in build.gradle of rolling upgrade tests that to not run this particular test when bwc version is 2.11 and 2.12

But we need a check to ensure the query enricher for `neural` clause works well, right? We don't have bwc tests for query enricher now, and we add test cases for neural and neural_sparse in this PR.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
```diff
@@ -0,0 +1,115 @@
...
+        NeuralSparseQueryBuilder sparseEncodingQueryBuilderWithoutModelId = new NeuralSparseQueryBuilder().fieldName(TEST_ENCODING_FIELD)
+            .queryText(TEXT_1);
+        // will set the model_id after we obtain the id
+        NeuralSparseQueryBuilder sparseEncodingQueryBuilderWithModelId = new NeuralSparseQueryBuilder().fieldName(TEST_ENCODING_FIELD)
+            .queryText(TEXT_1);
+
+        if (isRunningAgainstOldCluster()) {
+            String modelId = uploadSparseEncodingModel();
+            loadModel(modelId);
+            sparseEncodingQueryBuilderWithModelId.modelId(modelId);
```
vibrantvarun, line 42: Here there is no need of setting model Id .
Because in BWC tests our main purpose is 

When running against the old cluster create model Id, create enricher processor

When cluster gets updated to new version and when you execute a search then it will pick model id from the processor and set it internally. So you don't have to set it here

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
```diff
@@ -0,0 +1,115 @@
...
+    private static final String TEST_TEXT_FIELD = "passage_text";
+    private static final String TEXT_1 = "Hello world a b";
+
+    // Test restart-upgrade neural_query_enricher in restart-upgrade scenario
+    public void testNeuralQueryEnricherProcessor_NeuralSparseSearch_E2EFlow() throws Exception {
+        waitForClusterHealthGreen(NODES_BWC_CLUSTER);
+        NeuralSparseQueryBuilder sparseEncodingQueryBuilderWithoutModelId = new NeuralSparseQueryBuilder().fieldName(TEST_ENCODING_FIELD)
+            .queryText(TEXT_1);
+        // will set the model_id after we obtain the id
+        NeuralSparseQueryBuilder sparseEncodingQueryBuilderWithModelId = new NeuralSparseQueryBuilder().fieldName(TEST_ENCODING_FIELD)
```
vibrantvarun, line 36: This object is not needed

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
```diff
@@ -0,0 +1,115 @@
...
+    private static final String SPARSE_SEARCH_PIPELINE_NAME = "nqep-nlp-search-pipeline-sparse";
+    private static final String DENSE_SEARCH_PIPELINE_NAME = "nqep-nlp-search-pipeline-dense";
+    private static final String TEST_ENCODING_FIELD = "passage_embedding";
+    private static final String TEST_TEXT_FIELD = "passage_text";
+    private static final String TEXT_1 = "Hello world a b";
+
+    // Test restart-upgrade neural_query_enricher in restart-upgrade scenario
+    public void testNeuralQueryEnricherProcessor_NeuralSparseSearch_E2EFlow() throws Exception {
+        waitForClusterHealthGreen(NODES_BWC_CLUSTER);
+        NeuralSparseQueryBuilder sparseEncodingQueryBuilderWithoutModelId = new NeuralSparseQueryBuilder().fieldName(TEST_ENCODING_FIELD)
```
vibrantvarun, line 33: This object is also not needed.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
```diff
@@ -0,0 +1,115 @@
...
+                Settings.builder().put("index.search.default_pipeline", SPARSE_SEARCH_PIPELINE_NAME)
+            );
+
+            // do nothing here. need to add test codes after finishing backport
+        } else {
+            String modelId = null;
+            try {
+                modelId = TestUtils.getModelId(getIngestionPipeline(SPARSE_INGEST_PIPELINE_NAME), SPARSE_ENCODING_PROCESSOR);
+                loadModel(modelId);
+                sparseEncodingQueryBuilderWithModelId.modelId(modelId);
```
vibrantvarun, line 64: No need to set modelId here.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
```diff
@@ -0,0 +1,115 @@
...
+                    search(getIndexNameForTest(), sparseEncodingQueryBuilderWithoutModelId, 1).get("hits"),
+                    search(getIndexNameForTest(), sparseEncodingQueryBuilderWithModelId, 1).get("hits")
+                );
+            } finally {
+                wipeOfTestResources(getIndexNameForTest(), SPARSE_INGEST_PIPELINE_NAME, modelId, SPARSE_SEARCH_PIPELINE_NAME);
+            }
+        }
+    }
+
+    public void testNeuralQueryEnricherProcessor_NeuralSearch_E2EFlow() throws Exception {
```
vibrantvarun, line 75: Same comments which I did for the above tests applies here as well.

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
```diff
@@ -0,0 +1,147 @@
...
+    private static final String DENSE_SEARCH_PIPELINE_NAME = "nqep-nlp-search-pipeline-dense";
+    private static final String TEST_ENCODING_FIELD = "passage_embedding";
+    private static final String TEST_TEXT_FIELD = "passage_text";
+    private static final String TEXT_1 = "Hello world a b";
+    private String sparseModelId = "";
+    private String denseModelId = "";
+
+    public void testNeuralQueryEnricherProcessor_NeuralSparseSearch_E2EFlow() throws Exception {
+        waitForClusterHealthGreen(NODES_BWC_CLUSTER);
+        NeuralSparseQueryBuilder sparseEncodingQueryBuilderWithoutModelId = new NeuralSparseQueryBuilder().fieldName(TEST_ENCODING_FIELD)
```
vibrantvarun, line 34: Same comments from restart-upgrade applies here. You expect that when the model Id is to be set internally by neuralQueryEnricher processor then why are we setting it explicity everywhere?

qa/restart-upgrade/build.gradle
```diff
@@ -65,12 +65,21 @@ task testAgainstOldCluster(type: StandaloneRestIntegTestTask) {
...
         filter {
             excludeTestsMatching "org.opensearch.neuralsearch.bwc.MultiModalSearchIT.*"
             excludeTestsMatching "org.opensearch.neuralsearch.bwc.HybridSearchIT.*"
             excludeTestsMatching "org.opensearch.neuralsearch.bwc.NeuralSparseSearchIT.*"
+            excludeTestsMatching "org.opensearch.neuralsearch.bwc.NeuralQueryEnricherProcessorIT.*"
+        }
+    }
+
+    // Excluding the test because we introduce this feature in 2.13
+    if (ext.neural_search_bwc_version.startsWith("2.11") || ext.neural_search_bwc_version.startsWith("2.12")){
```
vibrantvarun, line 80: There are 3 cases

1) bwc version is 2.11 and current version is 2.12
For this the check should be
if (ext.neural_search_bwc_version.startsWith("2.11")  && ext.opensearch_version.startsWith("2.12")){
something on the above lines

2) bwc version is 2.11 and current version is on or after 2.13
This condition is not needed because processor will be created in the older version and search will executed in the newer version


3) bwc version is 2.13 and current version is on after 2.14

qa/restart-upgrade/build.gradle
```diff
@@ -98,12 +107,21 @@ task testAgainstNewCluster(type: StandaloneRestIntegTestTask) {
...
```
vibrantvarun, line 122: Refer the comment above

vibrantvarun:
I am thinking more on this

I think this functionality will break the scenario's if a user 
wants a different modeId with neural search query and neural sparse search query.

Also if a user doesn't want default model id support on neural query or neural sparse query either of them he cannot do that as by default if we create a processor it will be applied on both the query builder.

Therfore, there should be a different solution.

Here is my thinking 
This work can be done in possibly 2 ways
1) Create a new processor NeuralSparseQueryEnricherProcessor and a new visitor

or 

2) Create new fields in the existing neuralqueryenricher processor like default_sparse_modelId and set it there and also the field map. But there would be lot of ifs and elses in the NeuralQueryEnricherProcessor. 
Also create new visitor because if you use the current visitor then there will be a if check for neural sparse query also and it will break some scenarios in NeuralQuery model id support.

Therefore my suggestion is to go with Option 1.

cc: @navneet1v @martin-gaievski

zhichao-aws:
> I am thinking more on this
> 
> I think this functionality will break the scenario's if a user wants a different modeId with neural search query and neural sparse search query.
> 
> Also if a user doesn't want default model id support on neural query or neural sparse query either of them he cannot do that as by default if we create a processor it will be applied on both the query builder.
> 
> Therefore, there should be a different solution.
> 
> Here is my thinking This work can be done in possibly 2 ways
> 
> 1. Create a new processor NeuralSparseQueryEnricherProcessor and a new visitor
> 
> or
> 
> 2. Create new fields in the existing neuralqueryenricher processor like default_sparse_modelId and set it there and also the field map. But there would be lot of ifs and elses in the NeuralQueryEnricherProcessor.
>    Also create new visitor because if you use the current visitor then there will be a if check for neural sparse query also and it will break some scenarios in NeuralQuery model id support.
> 
> Therefore my suggestion is to go with Option 1. Maintanence with option 1 is easy as it gives flexibility to extend the functionality for multiple use cases.
> 
> cc: @navneet1v @martin-gaievski

This scenerio is similar to users have 2 knn fields and needs 2 default model_id for neural query. In this case, users can set the `neural_field_default_id` to set the default model_id for each field. And for the scenerio you mentioned, it works the same.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
zhichao-aws: Here we set the NeuralSparseQueryBuilder with model id, because we want to check the search result for default model id is same with the search result with a model id. This is to make sure the functionality of default model id.

qa/restart-upgrade/build.gradle
zhichao-aws: For case 1 I think we can ignore because the codes in this PR won't be in 2.12.

for case 2, do you mean we don't need to do search in old cluster? This will introduce a behavior different in old cluster( if bwcVersion before 2.13 we just create processor, if after 2.13, we also do search). And we'll need to add bwcVersion check in the IT, like the code I removed yesterday.

src/main/java/org/opensearch/neuralsearch/query/ModelInferenceQueryBuilder.java
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * Get the model id used by ml-commons model inference. Return null if the model id is absent.
+     */
+    public String modelId();
+
+    /**
+     * Set a new model id for the query builder.
+     */
+    public ModelInferenceQueryBuilder modelId(String modelId);
```
zane-neo, line 20: Is this returned value used anywhere?

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
```diff
@@ -0,0 +1,115 @@
...
+    private static final String SPARSE_INGEST_PIPELINE_NAME = "nqep-nlp-ingest-pipeline-sparse";
+    private static final String DENSE_INGEST_PIPELINE_NAME = "nqep-nlp-ingest-pipeline-dense";
+    private static final String SPARSE_SEARCH_PIPELINE_NAME = "nqep-nlp-search-pipeline-sparse";
+    private static final String DENSE_SEARCH_PIPELINE_NAME = "nqep-nlp-search-pipeline-dense";
+    private static final String TEST_ENCODING_FIELD = "passage_embedding";
+    private static final String TEST_TEXT_FIELD = "passage_text";
+    private static final String TEXT_1 = "Hello world a b";
+
+    // Test restart-upgrade neural_query_enricher in restart-upgrade scenario
+    public void testNeuralQueryEnricherProcessor_NeuralSparseSearch_E2EFlow() throws Exception {
```
zane-neo, line 31: Can we split this method to different ones and each one for only a single purpose, or at least add comments on this method since it seems this is rather than only a neural sparse search e2e test.

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilderTests.java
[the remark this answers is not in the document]
zane-neo: This is to test this part: https://github.com/opensearch-project/neural-search/pull/614/files#diff-c2736fa58eb7d17649ddcad13edab06a2f89cf888d8165d1fb568d09a948596dR89, right?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -1057,7 +1057,13 @@ protected void wipeOfTestResources(
             deleteSearchPipeline(searchPipeline);
         }
         if (modelId != null) {
-            deleteModel(modelId);
+            try {
+                deleteModel(modelId);
```
zane-neo, line 1061: Can we make this retryable with max_retry time, e.g. 5, one retry might not helpful.

src/main/java/org/opensearch/neuralsearch/query/ModelInferenceQueryBuilder.java
zhichao-aws: Yes, it's used by the NeuralSearchQueryVisitor to check whether the model id field is null

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
zhichao-aws: I think the flaky test is caused by some status sync issue in ml-commons. The model is undeployed but it's state is still deployed. And I've create a ticket about this: https://github.com/opensearch-project/ml-commons/issues/1986. For this case, one more call will help the model sync the model state. And I haven't get this exception anymore locally or in github CI after making the changes here.
How about changing this retry to 5 times if we meet this flaky test again? And we should also find the root cause in ml-commons.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java
zhichao-aws: Yes, it's bwc test for NeuralQueryEnricherProcessor, and it includes default model_id for both neural and neural_sparse. Add comments about this.

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilderTests.java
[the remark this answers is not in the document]
vibrantvarun: yes exactly, @zhichao-aws please remove that check

vibrantvarun:
> > I am thinking more on this
> > I think this functionality will break the scenario's if a user wants a different modeId with neural search query and neural sparse search query.
> > Also if a user doesn't want default model id support on neural query or neural sparse query either of them he cannot do that as by default if we create a processor it will be applied on both the query builder.
> > Therefore, there should be a different solution.
> > Here is my thinking This work can be done in possibly 2 ways
> > 
> > 1. Create a new processor NeuralSparseQueryEnricherProcessor and a new visitor
> > 
> > or
> > 
> > 2. Create new fields in the existing neuralqueryenricher processor like default_sparse_modelId and set it there and also the field map. But there would be lot of ifs and elses in the NeuralQueryEnricherProcessor.
> >    Also create new visitor because if you use the current visitor then there will be a if check for neural sparse query also and it will break some scenarios in NeuralQuery model id support.
> > 
> > Therefore my suggestion is to go with Option 1. Maintanence with option 1 is easy as it gives flexibility to extend the functionality for multiple use cases.
> > cc: @navneet1v @martin-gaievski
> 
> This scenerio is similar to users have 2 knn fields and needs 2 default model_id for neural query. In this case, users can set the `neural_field_default_id` to set the default model_id for each field. And for the scenerio you mentioned, it works the same.

Hey @zhichao-aws 

Consider you created a processor

```
UT /_search/pipeline/default_model_pipeline 
{
  "request_processors": [
    {
      "neural_query_enricher" : {
        "default_model_id": "bQ1J8ooBpBj3wT4HVUsb",
        "neural_field_default_id": {
           "my_field_1": "uZj0qYoBMtvQlfhaYeud",
           "my_field_2": "upj0qYoBMtvQlfhaZOuM"
        }
      }
    }
  ]
}
```
By the code definition the first priority is given to neural_field_default_id to set the model Id in the querybuilder. Then if the neural_field_default_id is not present then it will cater default_model_id


Q ) Consider an example where cx want to do a neural search with default_model_id  which applies to all field regardless of any field specific model id. 

A) Cx can set a default_model_id which applies it to all fields.
```
UT /_search/pipeline/default_model_pipeline 
{
  "request_processors": [
    {
      "neural_query_enricher" : {
        "default_model_id": "bQ1J8ooBpBj3wT4HVUsb",
        "neural_field_default_id": {
        }
      }
    }
  ]
}
```


Q) The work which you did Now if Cx want a specific model Id for neural sparse search which is field related (lets say field name is passage_text) and for neural search cx want a default_model_id which applies to all fields.The work which you did will work like

A) Cx created a processor which has default_model_id for neural search and field specific model Id for neural sparse search

```
UT /_search/pipeline/default_model_pipeline 
{
  "request_processors": [
    {
      "neural_query_enricher" : {
        "default_model_id": "bQ1J8ooBpBj3wT4HVUsb",
        "neural_field_default_id": {
             "passage_text": "uZj0qYoBMtvQlfhaYeud"
        }
      }
    }
  ]
}
```

Now as per your code it will always select field specific model id because field map will never be empty and the code you did in neuralsearchqueryvisitor will always apply same logic for neural search and neural sparse search.


Moreover, the (reply)[https://github.com/opensearch-project/neural-search/pull/614#issuecomment-1986586819] you gave here says here to create a different model Id in for each field in the neural_field_default_id map. 

```
```
UT /_search/pipeline/default_model_pipeline 
{
  "request_processors": [
    {
      "neural_query_enricher" : {
        "default_model_id": null,
        "neural_field_default_id": {
           "passage_text": "uZj0qYoBMtvQlfhaYeud",
           "passage_text_1": " LuZj0qYoBMtvQlfhaYeud",
        }
      }
    }
  ]
}
```

1) The problem with above scenario's are code will fetch model Id with whatever field is passed in the map. But, most of the cx rely heavily on default_model_id as they prefer to apply 1 model id for all fields rather than giving it seperatly in the map.

2) One more problem is what if on the same field name i want to use 2 model id's one in neural search query and neural sparse query. 

Also there might most of the scenarios which might be missing. I have already discussed with @navneet1v and would advise to create a new visitor. Also for more readability add one more map in the neural_query_enricher processor i.e. neural_sparse_field_default_id and use it in new visitor.


Also consider 1 more scenario

```
"query": {
    "hybrid": {
      "queries": [
        {
          "neural_sparse": {
                 "passage_embedding": {
                        "query_text": "Hi world",
                 }
        },
        {
          "neural": {
            "passage_embedding": {
              "query_text": "Hi world",
              "k": 5
            }
          }
        }
      ]
    }
  }
```

How can someone give a different model id for a specific field in both the clauses above when having same field name by using your code implementation in the PR?

**Comment by zhichao-aws** [comment]
Hi @vibrantvarun 
> How can someone give a different model id for a specific field in both the clauses above when having same field name by using your code implementation in the PR?

This is an impossible scenerio. Because `neural` query works only on `knn` field, and `neural_sparse` query works only on `rank_features` field. One field can not be `knn` and `rank_features` field. There must be 2 fields name for each.

> Now as per your code it will always select field specific model id because field map will never be empty and the code you did in neuralsearchqueryvisitor will always apply same logic for neural search and neural sparse search.

Now the code implementation is as follows:
```
       if (modelInferenceQueryBuilder.modelId() == null) {
            if (neuralFieldMap != null
                && modelInferenceQueryBuilder.fieldName() != null
                && neuralFieldMap.get(modelInferenceQueryBuilder.fieldName()) != null) {
                String fieldDefaultModelId = (String) neuralFieldMap.get(modelInferenceQueryBuilder.fieldName());
                modelInferenceQueryBuilder.modelId(fieldDefaultModelId);
            } else if (modelId != null) {
                modelInferenceQueryBuilder.modelId(modelId);
            } else {
                    "model id must be provided in neural query or a default model id must be set in search request processor"
                );
            }
        }
```
Consider we have a default model id for `neural` and field model id for `neural_sparse` and field name is `passage_embedding`. For neural sparse query on `passage_embedding`, it works fine. And for `neural` on other fields, for the first `if` statement, the check `neuralFieldMap.get(modelInferenceQueryBuilder.fieldName()) != null` will fail, because we only have `passage_embedding` in the map and the field name for neural query can not be `passage_embedding`. So it will check the second `if` statement and assign the default model_id.

File: src/main/java/org/opensearch/neuralsearch/query/ModelInferenceQueryBuilder.java (hunk 21 above)
Comment (zane-neo) [reply src/main/java/org/opensearch/neuralsearch/query/ModelInferenceQueryBuilder.java:20]: Can you point the line using this return value?

Comment (zhichao-aws) [reply src/main/java/org/opensearch/neuralsearch/query/ModelInferenceQueryBuilder.java:20]: Sure. https://github.com/zhichao-aws/neural-search/blob/1ade82f5566b96fe617ab1f6c5933ebad3258c6d/src/main/java/org/opensearch/neuralsearch/query/visitor/NeuralSearchQueryVisitor.java#L33

File: src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilderTests.java (hunk 10 above)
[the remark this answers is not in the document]
Comment (zhichao-aws) [reply src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilderTests.java:206]: It's testing this part: https://github.com/zhichao-aws/neural-search/blob/1ade82f5566b96fe617ab1f6c5933ebad3258c6d/src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java#L168

File: src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java (hunk 23 above)
Comment (zhichao-aws) [reply src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java:1061]: > We don't need this work as if you see the definition of deleteModel then it already checks for undeployed state 5 times.

From our side we do check the model states multiple times. But from the ml-commons side, the model management system sometimes doesn't sync the model state and need another `undeploy` API call to refresh the model state. You can check the ticket I posted here for reference, the response for getTask API was frozen, no field was updated even includes the `last_update_time`. In this case no matter how many times we call the `getTask` API, we can not get the latest state.

File: src/main/java/org/opensearch/neuralsearch/query/ModelInferenceQueryBuilder.java (hunk 21 above)
Comment (zane-neo) [reply src/main/java/org/opensearch/neuralsearch/query/ModelInferenceQueryBuilder.java:20]: The line you're pointing at seems the getter method without parameter, am I missing anything?

Comment (zhichao-aws) [reply src/main/java/org/opensearch/neuralsearch/query/ModelInferenceQueryBuilder.java:20]: Sorry, I misunderstood your question. The  invokation is here:
https://github.com/zhichao-aws/neural-search/blob/1ade82f5566b96fe617ab1f6c5933ebad3258c6d/src/main/java/org/opensearch/neuralsearch/query/visitor/NeuralSearchQueryVisitor.java#L38
https://github.com/zhichao-aws/neural-search/blob/1ade82f5566b96fe617ab1f6c5933ebad3258c6d/src/main/java/org/opensearch/neuralsearch/query/visitor/NeuralSearchQueryVisitor.java#L40

Comment (zhichao-aws) [reply src/main/java/org/opensearch/neuralsearch/query/ModelInferenceQueryBuilder.java:20]: We just set the modelId, and don't use the return value. But this setter method is generated automatically by lombok @Setter annotation. And the return value for setter is the class itself by default.

**Comment by zane-neo** [comment]
LGTM, approving.

**Review by zane-neo — APPROVED** [review]

File: qa/restart-upgrade/build.gradle (hunk 19 above)
Comment (vibrantvarun) [reply qa/restart-upgrade/build.gradle:80]: Got it.

**Comment by zhichao-aws** [comment]
> Looks much better now. LGTM , but before I approve this merge request can you add a documentation-website issue to add the documentation for cx to understand the feature.
> 
> Sample issue: [opensearch-project/documentation-website#5060](https://github.com/opensearch-project/documentation-website/issues/5060)
> 
> Reference: https://opensearch.org/docs/latest/search-plugins/semantic-search/#setting-a-default-model-on-an-index-or-field

Good point. The document issue link: https://github.com/opensearch-project/documentation-website/issues/6652

**Review by vibrantvarun — APPROVED** [review]
LGTM

File: qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/NeuralQueryEnricherProcessorIT.java [hunk 24]
```diff
@@ -0,0 +1,151 @@
...
+                    Settings.builder().put("index.search.default_pipeline", SPARSE_SEARCH_PIPELINE_NAME)
+                );
+
+                // do nothing here. need to add test codes after finishing backport
+                break;
+            case MIXED:
+                sparseModelId = TestUtils.getModelId(getIngestionPipeline(SPARSE_INGEST_PIPELINE_NAME), SPARSE_ENCODING_PROCESSOR);
+                loadModel(sparseModelId);
+                sparseEncodingQueryBuilderWithModelId.modelId(sparseModelId);
+                // do nothing here. need to add test codes after finishing backport
```
vibrantvarun, line 67: Nit: Remove this comment

zhichao-aws: ack

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 3.0](https://github.com/opensearch-project/neural-search/compare/2.x...HEAD)
 ### Features
+- Enabled support for applying default modelId in neural sparse query ([#614](https://github.com/opensearch-project/neural-search/pull/614)
```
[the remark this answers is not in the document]
zhichao-aws: ack

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -108,7 +119,9 @@ protected void doXContent(XContentBuilder xContentBuilder, Params params) throws
         xContentBuilder.startObject(NAME);
         xContentBuilder.startObject(fieldName);
         xContentBuilder.field(QUERY_TEXT_FIELD.getPreferredName(), queryText);
-        xContentBuilder.field(MODEL_ID_FIELD.getPreferredName(), modelId);
+        if (modelId != null) {
```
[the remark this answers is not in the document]
zhichao-aws: Ack

src/main/java/org/opensearch/neuralsearch/query/ModelInferenceQueryBuilder.java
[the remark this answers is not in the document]
zhichao-aws: In NeuralSparseQueryBuilder and NeuralQueryBuilder we have a `modelId` setter method generated automatically by lombok `@Setter` annotation. The auto generated `modelId` will take the class itself as the return type.

Do you prefer we add a `setModelId` method with void return value in NeuralSparseQueryBuilder and NeuralQueryBuilder?

src/main/java/org/opensearch/neuralsearch/query/ModelInferenceQueryBuilder.java
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Query builders which calls ml-commons API to do model inference.
+ * The model inference result is used for search on target field.
+ */
+public interface ModelInferenceQueryBuilder {
+    /**
+     * Get the model id used by ml-commons model inference. Return null if the model id is absent.
+     */
+    public String modelId();
```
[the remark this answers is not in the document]
zhichao-aws: This getter method is also generated automatically by lombok.

Do you prefer we define a `getModelId` with Optional return type in NeuralSparseQueryBuilder and NeuralQueryBuilder?

src/main/java/org/opensearch/neuralsearch/query/visitor/NeuralSearchQueryVisitor.java
```diff
@@ -28,16 +28,16 @@ public class NeuralSearchQueryVisitor implements QueryBuilderVisitor {
      */
     @Override
     public void accept(QueryBuilder queryBuilder) {
-        if (queryBuilder instanceof NeuralQueryBuilder) {
-            NeuralQueryBuilder neuralQueryBuilder = (NeuralQueryBuilder) queryBuilder;
-            if (neuralQueryBuilder.modelId() == null) {
+        if (queryBuilder instanceof ModelInferenceQueryBuilder) {
+            ModelInferenceQueryBuilder modelInferenceQueryBuilder = (ModelInferenceQueryBuilder) queryBuilder;
+            if (modelInferenceQueryBuilder.modelId() == null) {
```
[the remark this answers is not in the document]
zhichao-aws: ack

src/main/java/org/opensearch/neuralsearch/query/visitor/NeuralSearchQueryVisitor.java
```diff
@@ -28,16 +28,16 @@ public class NeuralSearchQueryVisitor implements QueryBuilderVisitor {
...
                 if (neuralFieldMap != null
-                    && neuralQueryBuilder.fieldName() != null
-                    && neuralFieldMap.get(neuralQueryBuilder.fieldName()) != null) {
-                    String fieldDefaultModelId = (String) neuralFieldMap.get(neuralQueryBuilder.fieldName());
-                    neuralQueryBuilder.modelId(fieldDefaultModelId);
+                    && modelInferenceQueryBuilder.fieldName() != null
+                    && neuralFieldMap.get(modelInferenceQueryBuilder.fieldName()) != null) {
```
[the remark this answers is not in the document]
zhichao-aws: ack

vibrantvarun:
Please check the codecoverage and fix the GH action

Issue #621, opened 2024-03-01
[BUG] Exception "e_o_f_exception read past EOF" returned for sub-query with `multi_match` nested into `dis_max`


martin-gaievski:
### What is the bug?
Error below returned in response

```
"type": "e_o_f_exception",
"reason": "read past EOF (pos=4250): MemorySegmentIndexInput(path=\"/home/ec2-user/opensearch/data/
```
error appears when query is quite complex and for certain query phases. Example of the query structure:

```
"hybrid": {
          "queries": [
                {
                    "bool": {
                          "boost": 1,
                          "should": {
                                "dis_max": {
                                      "queries": [
                                          "bool": {
                                                    "boost": 1.1,
                                                    "filter": [],
                                                    "minimum_should_match": "1",
                                                    "should": {
                                                        "dis_max": {
                                                            "queries": [
                                                                {
                                                                    "multi_match": { }
                                                                },
                                                                {
                                                                    "multi_match": { }
                                                                },
                                                                {
                                                                    "match": { }
                                                                }
                                                            ],
                                                            "tie_breaker": 0
                                                        }
                                                    }
                                                }
```

Below is example of zoomed in section of `multi_match` sub-query:

```
                                             "should": {
                                                        "dis_max": {
                                                            "queries": [
                                                                {
                                                                    "multi_match": {
                                                                        "_name": "new/timeline_match/match_in_phrases",
                                                                        "boost": 5,
                                                                        "fields": [
                                                                            "sentences.in_english"
                                                                        ],
                                                                        "fuzziness": "AUTO:3,11",
                                                                        "fuzzy_transpositions": true,
                                                                        "max_expansions": 50,
                                                                        "minimum_should_match": "60%",
                                                                        "operator": "AND",
                                                                        "prefix_length": 0,
                                                                        "query": "{{query_text}}",
                                                                        "type": "most_fields"
                                                                    }
                                                                },
```

### What is the expected behavior?
No errors, query should return meaningful results with documetns and scores

### What is your host/environment?
Issue present in 2.12 release

### Do you have any additional context?
1. In a local test cluster error can manifestate itself differently. There is an exception below in a log and cluster crushes.

```
» ERROR][o.o.b.OpenSearchUncaughtExceptionHandler] [integTest-0] fatal error in thread [opensearch[integTest-0][search][T#14]], exiting
»  java.lang.AssertionError: null
[...]

PR #498, opened 2023-11-22
Hybrid query and nested type fields

martin-gaievski:
### Description
Nested fields have few problems that we're addressing in this PR:
-  query is wrapped into the Bool query inside the core code ([ref](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/DefaultSearchContext.java#L367-L370)), that happening behind the scenes and depends only on presents of the field in the index mapping. This case should work.

- wrapping up hybrid search query into another query in general. That is not how hybrid query were designed, it must be a top level query. We need to track and block such queries.

We're mainly addressing case 1, and trying our best for case 2. 

Below is snippet of example from original github issue to illustrate case 1. Without this PR search query returns wrong scores:

index mapping 

```
PUT /index-test
{
    "settings": {
        "index.knn": true,
        "number_of_shards": 2,
        "default_pipeline": "pipeline-test"
    },
    "mappings": {
        "properties": {
            "id": {
                "type": "text"
            },
            "passage_embedding": {
                "type": "knn_vector",
                "dimension": 768,
                "method": {
                    "name": "hnsw",
                    "space_type": "cosinesimil",
                    "engine": "lucene",
                    "parameters": {
                        "ef_construction": 512,
                        "m": 8
                    }
                }
            },
            "name": {
                "type": "text"
            },
            "passage_text": {
                "type": "text"
            },
            "test": {
                "type": "text"
            },
            "user": {
                "type": "nested"
            }
        }
    }
}
```

add doc to index

```
POST /index-test/_doc/1
{
    "name": "A West Virginia university women 's basketball team , officials , and a small gathering of fans are in a West Virginia arena .",
    "id": "4319130149.jpg"
}
```

search request
```
GET /index-test/_search?search_pipeline=nlp-search-pipeline

{
    "_source": {
        "exclude": [
            "passage_embedding"
        ]
    },
    "query": {
        "hybrid": {
            "queries": [
                {
                    "match": {
                        "name": {
                            "query": "wild west"
                        }
                    }
                },
                {
                    "neural": {
                        "passage_embedding": {
                            "query_text": "wild west",
                            "model_id": "{{model_id}}",
                            "k": 20
                        }
                    }
                }
            ]
        }
    }
}
```

### Issues Resolved
https://github.com/opensearch-project/neural-search/issues/466

### Check List
- [x] New functionality includes testing.
    - [x] All tests pass
- [x] Commits are signed as per the DCO using --signoff

[...]

PR #624, opened 2024-03-03
Fix runtime exceptions in hybrid query for case when sub-query scorer return TwoPhase iterator that is incompatible with DISI iterator

martin-gaievski:
### Description
Adding approximation two phase iterator to hybrid query. That helps in scenarios when sub-queries are complex and direct iteration over all scorers is expensive and may lead to instability, like runtime exceptions. 

Example of such query is `bool` with `filter` some complex queries in `should` like `dis_max`. 

```
"query": {
        "hybrid": {
            "queries": [
                {
                    "bool": {
                        "filter": [
                            {
                                "terms": {
                                    "field1": [
                                        "value1",
                                        "value2"
                                    ]
                                }
                            }
                        ],
                        "should": {
                            "dis_max": {
                                "queries": [
                                    {
                                        "multi_match": {
                                            "query": "{{query_text}}",
                                            "type": "phrase_prefix"
                                        }
                                    },
                                    {
                                        "multi_match": {
                                            "query": "{{query_text}}",
                                            "type": "most_fields"
                                        }
                                    }
                                ]
}}}}]}}
```

System uses [ReqOptSumScorer](https://github.com/apache/lucene/blob/branch_9_9/lucene/core/src/java/org/apache/lucene/search/ReqOptSumScorer.java) that is based on approximation iterator for scores and two phase iterator for docs. In underlying query clause is complex enough and data set is large avoiding such iterations and using scorers directly may lead to runtime exceptions described in https://github.com/opensearch-project/neural-search/issues/621

Key changes in this PR:
 - added two phase iterator for scorer
 - add collector manager, inject it at preProcess of aggregationProcessor phase in search
 - make query phase search light, only validate for hybrid query special cases
 - change the way sub-queries are rewritten - instead of static Rewriteable.rewrite switch to QueryBuild.rewrite

### Issues Resolved
https://github.com/opensearch-project/neural-search/issues/621

### Check List
    - [x] All tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Features
 ### Enhancements
 ### Bug Fixes
+- Adding two phase iterator for hybrid query ([#624](https://github.com/opensearch-project/neural-search/pull/624))
 ### Infrastructure
 ### Documentation
 ### Maintenance
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Features
 ### Enhancements
 ### Bug Fixes
-- Adding two phase iterator for hybrid query ([#624](https://github.com/opensearch-project/neural-search/pull/624))
+- Fix runtime exceptions in hybrid query for case when sub-query scorer return TwoPhase iterator that is incompatible with DISI iterator ([#624](https://github.com/opensearch-project/neural-search/pull/624))
 ### Infrastructure
 ### Documentation
 ### Maintenance
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
@@ -12,7 +12,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
 
-import org.apache.lucene.index.IndexReader;
 import org.apache.lucene.search.BooleanClause;
 import org.apache.lucene.search.BooleanQuery;
 import org.apache.lucene.search.IndexSearcher;
@@ -77,12 +76,12 @@ public final class HybridQuery extends Query implements Iterable<Query> {
     /**
      * Re-writes queries into primitive queries. Callers are expected to call rewrite multiple times if necessary,
      * until the rewritten query is the same as the original query.
-     * @param reader
+     * @param indexSearcher
      * @return
      * @throws IOException
      */
     @Override
-    public Query rewrite(IndexReader reader) throws IOException {
+    public Query rewrite(IndexSearcher indexSearcher) throws IOException {
         if (subQueries.isEmpty()) {
             return new MatchNoDocsQuery("empty HybridQuery");
         }
@@ -90,7 +89,7 @@ public final class HybridQuery extends Query implements Iterable<Query> {
         boolean actuallyRewritten = false;
         List<Query> rewrittenSubQueries = new ArrayList<>();
         for (Query subQuery : subQueries) {
-            Query rewrittenSub = subQuery.rewrite(reader);
+            Query rewrittenSub = subQuery.rewrite(indexSearcher);
             /* we keep rewrite sub-query unless it's not equal to itself, it may take multiple levels of recursive calls
                queries need to be rewritten from high-level clauses into lower-level clauses because low-level clauses
                perform better. For hybrid query we need to track progress of re-write for all sub-queries */
@@ -102,7 +101,7 @@ public final class HybridQuery extends Query implements Iterable<Query> {
             return new HybridQuery(rewrittenSubQueries);
         }
 
-        return super.rewrite(reader);
+        return super.rewrite(indexSearcher);
     }
 
     /**
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
@@ -27,8 +27,6 @@ import org.apache.lucene.search.Weight;
 import lombok.Getter;
 import org.apache.lucene.util.PriorityQueue;
 
-import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
-
 /**
  * Class abstracts functionality of Scorer for hybrid query. When iterating over documents in increasing
  * order of doc id, this class fills up array of scores per sub-query for each doc id. Order in array of scores
@@ -47,7 +45,7 @@ public final class HybridQueryScorer extends Scorer {
     private final Map<Query, List<Integer>> queryToIndex;
 
     private final DocIdSetIterator approximation;
-    HybridScorePropagator disjunctionBlockPropagator;
+    HybridScoreBlockBoundaryPropagator disjunctionBlockPropagator;
     private final TwoPhase twoPhase;
 
     public HybridQueryScorer(Weight weight, List<Scorer> subScorers) throws IOException {
@@ -56,23 +54,19 @@ public final class HybridQueryScorer extends Scorer {
 
     public HybridQueryScorer(Weight weight, List<Scorer> subScorers, ScoreMode scoreMode) throws IOException {
         super(weight);
-        // max
         this.subScorers = Collections.unmodifiableList(subScorers);
-        // custom
         subScores = new float[subScorers.size()];
         this.queryToIndex = mapQueryToIndex();
-        // base
         this.subScorersPQ = initializeSubScorersPQ();
-        // base
         boolean needsScores = scoreMode != ScoreMode.COMPLETE_NO_SCORES;
-        this.approximation = new HybridDisjunctionDISIApproximation(this.subScorersPQ);
-        // max
+
+        this.approximation = new HybridSubqueriesDISIApproximation(this.subScorersPQ);
         if (scoreMode == ScoreMode.TOP_SCORES) {
-            this.disjunctionBlockPropagator = new HybridScorePropagator(subScorers);
+            this.disjunctionBlockPropagator = new HybridScoreBlockBoundaryPropagator(subScorers);
         } else {
             this.disjunctionBlockPropagator = null;
         }
-        // base
+
         boolean hasApproximation = false;
         float sumMatchCost = 0;
         long sumApproxCost = 0;
@@ -116,7 +110,7 @@ public final class HybridQueryScorer extends Scorer {
         float totalScore = 0.0f;
         for (DisiWrapper disiWrapper = topList; disiWrapper != null; disiWrapper = disiWrapper.next) {
             // check if this doc has match in the subQuery. If not, add score as 0.0 and continue
-            if (disiWrapper.scorer.docID() == NO_MORE_DOCS) {
+            if (disiWrapper.scorer.docID() == DocIdSetIterator.NO_MORE_DOCS) {
                 continue;
             }
             totalScore += disiWrapper.scorer.score();
@@ -187,7 +181,7 @@ public final class HybridQueryScorer extends Scorer {
     @Override
     public int docID() {
         if (subScorersPQ.size() == 0) {
-            return NO_MORE_DOCS;
+            return DocIdSetIterator.NO_MORE_DOCS;
         }
         return subScorersPQ.top().doc;
     }
@@ -269,6 +263,10 @@ public final class HybridQueryScorer extends Scorer {
         return children;
     }
 
+    /**
+     *  Object returned by Scorer.twoPhaseIterator() to provide an approximation of a DocIdSetIterator.
+     *  After calling nextDoc() or advance(int) on the iterator returned by approximation(), you need to check matches() to confirm if the retrieved document ID is a match.
+     */
     static class TwoPhase extends TwoPhaseIterator {
         private final float matchCost;
         // list of verified matches on the current doc
@@ -292,11 +290,10 @@ public final class HybridQueryScorer extends Scorer {
         }
 
         DisiWrapper getSubMatches() throws IOException {
-            // iteration order does not matter
-            for (DisiWrapper w : unverifiedMatches) {
-                if (w.twoPhaseView.matches()) {
-                    w.next = verifiedMatches;
-                    verifiedMatches = w;
+            for (DisiWrapper wrapper : unverifiedMatches) {
+                if (wrapper.twoPhaseView.matches()) {
+                    wrapper.next = verifiedMatches;
+                    verifiedMatches = wrapper;
                 }
             }
             unverifiedMatches.clear();
@@ -308,39 +305,38 @@ public final class HybridQueryScorer extends Scorer {
             verifiedMatches = null;
             unverifiedMatches.clear();
 
-            for (DisiWrapper w = subScorers.topList(); w != null;) {
-                DisiWrapper next = w.next;
+            for (DisiWrapper wrapper = subScorers.topList(); wrapper != null;) {
+                DisiWrapper next = wrapper.next;
 
-                if (w.twoPhaseView == null) {
+                if (Objects.isNull(wrapper.twoPhaseView)) {
                     // implicitly verified, move it to verifiedMatches
-                    w.next = verifiedMatches;
-                    verifiedMatches = w;
+                    wrapper.next = verifiedMatches;
+                    verifiedMatches = wrapper;
 
                     if (!needsScores) {
                         // we can stop here
                         return true;
                     }
                 } else {
-                    unverifiedMatches.add(w);
+                    unverifiedMatches.add(wrapper);
                 }
-                w = next;
+                wrapper = next;
             }
 
-            if (verifiedMatches != null) {
+            if (Objects.nonNull(verifiedMatches)) {
                 return true;
             }
 
             // verify subs that have an two-phase iterator
             // least-costly ones first
             while (unverifiedMatches.size() > 0) {
-                DisiWrapper w = unverifiedMatches.pop();
-                if (w.twoPhaseView.matches()) {
-                    w.next = null;
-                    verifiedMatches = w;
+                DisiWrapper wrapper = unverifiedMatches.pop();
+                if (wrapper.twoPhaseView.matches()) {
+                    wrapper.next = null;
+                    verifiedMatches = wrapper;
                     return true;
                 }
             }
-
             return false;
         }
 
@@ -350,18 +346,22 @@ public final class HybridQueryScorer extends Scorer {
         }
     }
 
-    static class HybridDisjunctionDISIApproximation extends DocIdSetIterator {
-        final DocIdSetIterator delegate;
+    /**
+     * A DocIdSetIterator which is a disjunction of the approximations of the provided iterators and supports
+     * sub iterators that return empty results
+     */
+    static class HybridSubqueriesDISIApproximation extends DocIdSetIterator {
+        final DocIdSetIterator docIdSetIterator;
         final DisiPriorityQueue subIterators;
 
-        public HybridDisjunctionDISIApproximation(DisiPriorityQueue subIterators) {
-            delegate = new DisjunctionDISIApproximation(subIterators);
+        public HybridSubqueriesDISIApproximation(final DisiPriorityQueue subIterators) {
+            docIdSetIterator = new DisjunctionDISIApproximation(subIterators);
             this.subIterators = subIterators;
         }
 
         @Override
         public long cost() {
-            return delegate.cost();
+            return docIdSetIterator.cost();
         }
 
         @Override
@@ -369,7 +369,7 @@ public final class HybridQueryScorer extends Scorer {
             if (subIterators.size() == 0) {
                 return NO_MORE_DOCS;
             }
-            return delegate.docID();
+            return docIdSetIterator.docID();
         }
 
         @Override
@@ -377,15 +377,15 @@ public final class HybridQueryScorer extends Scorer {
             if (subIterators.size() == 0) {
                 return NO_MORE_DOCS;
             }
-            return delegate.nextDoc();
+            return docIdSetIterator.nextDoc();
         }
 
         @Override
-        public int advance(int target) throws IOException {
+        public int advance(final int target) throws IOException {
             if (subIterators.size() == 0) {
                 return NO_MORE_DOCS;
             }
-            return delegate.advance(target);
+            return docIdSetIterator.advance(target);
         }
     }
 }
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
@@ -264,8 +264,11 @@ public final class HybridQueryScorer extends Scorer {
     }
 
     /**
-     *  Object returned by Scorer.twoPhaseIterator() to provide an approximation of a DocIdSetIterator.
-     *  After calling nextDoc() or advance(int) on the iterator returned by approximation(), you need to check matches() to confirm if the retrieved document ID is a match.
+     *  Object returned by {@link Scorer#twoPhaseIterator()} to provide an approximation of a {@link DocIdSetIterator}.
+     *  After calling {@link DocIdSetIterator#nextDoc()} or {@link DocIdSetIterator#advance(int)} on the iterator
+     *  returned by approximation(), you need to check {@link TwoPhaseIterator#matches()} to confirm if the retrieved
+     *  document ID is a match. Implementation inspired by identical class for
+     *  <a href="https://github.com/apache/lucene/blob/branch_9_10/lucene/core/src/java/org/apache/lucene/search/DisjunctionScorer.java">DisjunctionScorer</a>
      */
     static class TwoPhase extends TwoPhaseIterator {
         private final float matchCost;
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
@@ -6,6 +6,7 @@ package org.opensearch.neuralsearch.query;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -18,10 +19,15 @@ import org.apache.lucene.search.DisiWrapper;
 import org.apache.lucene.search.DisjunctionDISIApproximation;
 import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.Query;
+import org.apache.lucene.search.ScoreMode;
 import org.apache.lucene.search.Scorer;
+import org.apache.lucene.search.TwoPhaseIterator;
 import org.apache.lucene.search.Weight;
 
 import lombok.Getter;
+import org.apache.lucene.util.PriorityQueue;
+
+import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
 
 /**
  * Class abstracts functionality of Scorer for hybrid query. When iterating over documents in increasing
@@ -40,12 +46,60 @@ public final class HybridQueryScorer extends Scorer {
 
     private final Map<Query, List<Integer>> queryToIndex;
 
+    private final DocIdSetIterator approximation;
+    HybridScorePropagator disjunctionBlockPropagator;
+    private final TwoPhase twoPhase;
+
     public HybridQueryScorer(Weight weight, List<Scorer> subScorers) throws IOException {
+        this(weight, subScorers, ScoreMode.TOP_SCORES);
+    }
+
+    public HybridQueryScorer(Weight weight, List<Scorer> subScorers, ScoreMode scoreMode) throws IOException {
         super(weight);
+        // max
         this.subScorers = Collections.unmodifiableList(subScorers);
+        // custom
         subScores = new float[subScorers.size()];
         this.queryToIndex = mapQueryToIndex();
+        // base
         this.subScorersPQ = initializeSubScorersPQ();
+        // base
+        boolean needsScores = scoreMode != ScoreMode.COMPLETE_NO_SCORES;
+        this.approximation = new HybridDisjunctionDISIApproximation(this.subScorersPQ);
+        // max
+        if (scoreMode == ScoreMode.TOP_SCORES) {
+            this.disjunctionBlockPropagator = new HybridScorePropagator(subScorers);
+        } else {
+            this.disjunctionBlockPropagator = null;
+        }
+        // base
+        boolean hasApproximation = false;
+        float sumMatchCost = 0;
+        long sumApproxCost = 0;
+        // Compute matchCost as the average over the matchCost of the subScorers.
+        // This is weighted by the cost, which is an expected number of matching documents.
+        for (DisiWrapper w : subScorersPQ) {
+            long costWeight = (w.cost <= 1) ? 1 : w.cost;
+            sumApproxCost += costWeight;
+            if (w.twoPhaseView != null) {
+                hasApproximation = true;
+                sumMatchCost += w.matchCost * costWeight;
+            }
+        }
+        if (!hasApproximation) { // no sub scorer supports approximations
+            twoPhase = null;
+        } else {
+            final float matchCost = sumMatchCost / sumApproxCost;
+            twoPhase = new TwoPhase(approximation, matchCost, subScorersPQ, needsScores);
+        }
+    }
+
+    @Override
+    public int advanceShallow(int target) throws IOException {
+        if (disjunctionBlockPropagator != null) {
+            return disjunctionBlockPropagator.advanceShallow(target);
+        }
+        return super.advanceShallow(target);
     }
 
     /**
@@ -55,16 +109,27 @@ public final class HybridQueryScorer extends Scorer {
      */
     @Override
     public float score() throws IOException {
-        DisiWrapper topList = subScorersPQ.topList();
+        return score(getSubMatches());
+    }
+
+    private float score(DisiWrapper topList) throws IOException {
         float totalScore = 0.0f;
         for (DisiWrapper disiWrapper = topList; disiWrapper != null; disiWrapper = disiWrapper.next) {
             // check if this doc has match in the subQuery. If not, add score as 0.0 and continue
-            if (disiWrapper.scorer.docID() == DocIdSetIterator.NO_MORE_DOCS) {
+            if (disiWrapper.scorer.docID() == NO_MORE_DOCS) {
                 continue;
             }
             totalScore += disiWrapper.scorer.score();
         }
         return totalScore;
+    }
+
+    DisiWrapper getSubMatches() throws IOException {
+        if (twoPhase == null) {
+            return subScorersPQ.topList();
+        } else {
+            return twoPhase.getSubMatches();
+        }
     }
 
     /**
@@ -73,7 +138,16 @@ public final class HybridQueryScorer extends Scorer {
      */
     @Override
     public DocIdSetIterator iterator() {
-        return new DisjunctionDISIApproximation(this.subScorersPQ);
+        if (twoPhase != null) {
+            return TwoPhaseIterator.asDocIdSetIterator(twoPhase);
+        } else {
+            return approximation;
+        }
+    }
+
+    @Override
+    public TwoPhaseIterator twoPhaseIterator() {
+        return twoPhase;
     }
 
     /**
@@ -93,12 +167,28 @@ public final class HybridQueryScorer extends Scorer {
         }).max(Float::compare).orElse(0.0f);
     }
 
+    @Override
+    public void setMinCompetitiveScore(float minScore) throws IOException {
+        if (disjunctionBlockPropagator != null) {
+            disjunctionBlockPropagator.setMinCompetitiveScore(minScore);
+        }
+
+        for (Scorer scorer : subScorers) {
+            if (Objects.nonNull(scorer)) {
+                scorer.setMinCompetitiveScore(minScore);
+            }
+        }
+    }
+
     /**
      * Returns the doc ID that is currently being scored.
      * @return document id
      */
     @Override
     public int docID() {
+        if (subScorersPQ.size() == 0) {
+            return NO_MORE_DOCS;
+        }
         return subScorersPQ.top().doc;
     }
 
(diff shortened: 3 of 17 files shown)


martin-gaievski:
@msfroh @reta - can you please take a look if there are some issues in how we've added approximation and two phase iterators?

jmazanec15:
@martin-gaievski Could you provide additional context on why rewrite needs to be changed?

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
```diff
@@ -77,20 +76,20 @@ public String toString(String field) {
...
      * Re-writes queries into primitive queries. Callers are expected to call rewrite multiple times if necessary,
      * until the rewritten query is the same as the original query.
-     * @param reader
+     * @param indexSearcher
      * @return
      * @throws IOException
      */
     @Override
-    public Query rewrite(IndexReader reader) throws IOException {
+    public Query rewrite(IndexSearcher indexSearcher) throws IOException {
```
jmazanec15, line 84: Why do we need to switch? Isnt this an overriden method - what happens to the other method if called?

src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
```diff
@@ -169,4 +253,139 @@ private DisiPriorityQueue initializeSubScorersPQ() {
...
+        ArrayList<ChildScorable> children = new ArrayList<>();
+        for (DisiWrapper scorer = getSubMatches(); scorer != null; scorer = scorer.next) {
+            children.add(new ChildScorable(scorer.scorer, "SHOULD"));
+        }
+        return children;
+    }
+
+    /**
+     *  Object returned by Scorer.twoPhaseIterator() to provide an approximation of a DocIdSetIterator.
+     *  After calling nextDoc() or advance(int) on the iterator returned by approximation(), you need to check matches() to confirm if the retrieved document ID is a match.
```
jmazanec15, line 268: Link to reference implementation

src/main/java/org/opensearch/neuralsearch/query/HybridScoreBlockBoundaryPropagator.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.lucene.search.Scorer;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.Objects;
+
+/**
+ * This class functions as a utility for propagating block boundaries within disjunctions.
```
jmazanec15, line 17: What is a block boundary?

jmazanec15:
> System uses [ReqOptSumScorer](https://github.com/apache/lucene/blob/branch_9_9/lucene/core/src/java/org/apache/lucene/search/ReqOptSumScorer.java) that is based on approximation iterator for scores and two phase iterator for docs.

Are we actually using ReqOptSumScorer in this PR? I didnt see it. Seems like we model the new scorer off of it.

msfroh:
I suspect that this is very similar to https://github.com/opensearch-project/OpenSearch/issues/8155.

I'm going to take a closer look at `ReqOptSumScorer`, as it was implicated in that issue as well. As @noCharger and I were digging into that issue, we found that normally DISI and TwoPhaseIterator are "adaptable" into each other (where a DISI as a TPI is basically an "exact approximation" and a TPI as a DISI advances until it finds a doc where `matches()` is true).

I suspect that `ReqOptSumScorer` may be returning a TwoPhaseIterator that somehow doesn't give proper DISI behavior.

Anyway, I will take a look at that after I review this PR.

martin-gaievski:
> @martin-gaievski Could you provide additional context on why rewrite needs to be changed?

With existing code does the rewrite of sub-queries via [Rewriteable.rewrite](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/Rewriteable.java#L150) method, with new approach that is switched to [AbstractQueryBuilder.doRewrite](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/AbstractQueryBuilder.java#L295). For many queries that `doRewrite` method has query specific logic that is simply skipped with today's approach, for example [Term query](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/TermQueryBuilder.java#L185) or [Dismax query](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/DisMaxQueryBuilder.java#L209).

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
martin-gaievski: Lucene has [some logic](https://github.com/apache/lucene/blob/branch_9_10/lucene/core/src/java/org/apache/lucene/search/Query.java#L55-L60) and [different code paths](https://github.com/apache/lucene/blob/branch_9_10/lucene/core/src/java/org/apache/lucene/search/Query.java#L98-L100) depending on what exact method has been overriden. With today's code (where we override `rewrite(IndexReader)`) in certain scenarios doc iterator may have empty results (no doc ids). In particular I noticed that for Term query in scenarios when there are multiple term sub-queries and more than one query has matches. I suspect that happens because of rewrite. We'd stop doing rewrites too early on a first sub-query with matched docs, while we should continue rewrite cycles for other term sub-queries.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
martin-gaievski: ack

martin-gaievski:
> > System uses [ReqOptSumScorer](https://github.com/apache/lucene/blob/branch_9_9/lucene/core/src/java/org/apache/lucene/search/ReqOptSumScorer.java) that is based on approximation iterator for scores and two phase iterator for docs.
> 
> Are we actually using ReqOptSumScorer in this PR? I didnt see it. Seems like we model the new scorer off of it.

You're correct, we're not using that scorer neither in this PR or anywhere in neural-search plugin. When issue happens it's part of the stack trace:

```
"Caused by: java.io.EOFException: read past EOF (pos=13125): MemorySegmentIndexInput(path=\"/home/ec2-user/opensearch/data/nodes/0/indices/OOSE9kYNQL2RBK0UCWxBuA/0/index/_12.nvd\") [slice=randomaccess]",
"at org.apache.lucene.store.MemorySegmentIndexInput.handlePositionalIOOBE(MemorySegmentIndexInput.java:99) ~[lucene-core-9.9.2.jar:9.9.2 a2939784c4ca60bc28bf488b5479c02fc2e5e22c - 2024-01-25 09:51:09]",
"at org.apache.lucene.store.MemorySegmentIndexInput$SingleSegmentImpl.readByte(MemorySegmentIndexInput.java:507) ~[lucene-core-9.9.2.jar:9.9.2 a2939784c4ca60bc28bf488b5479c02fc2e5e22c - 2024-01-25 09:51:09]",
"at org.apache.lucene.codecs.lucene90.Lucene90NormsProducer$8.longValue(Lucene90NormsProducer.java:443) ~[lucene-core-9.9.2.jar:9.9.2 a2939784c4ca60bc28bf488b5479c02fc2e5e22c - 2024-01-25 09:51:09]",
"at org.apache.lucene.search.LeafSimScorer.getNormValue(LeafSimScorer.java:47) ~[lucene-core-9.9.2.jar:9.9.2 a2939784c4ca60bc28bf488b5479c02fc2e5e22c - 2024-01-25 09:51:09]",
"at org.apache.lucene.search.LeafSimScorer.score(LeafSimScorer.java:60) ~[lucene-core-9.9.2.jar:9.9.2 a2939784c4ca60bc28bf488b5479c02fc2e5e22c - 2024-01-25 09:51:09]",
"at org.apache.lucene.search.TermScorer.score(TermScorer.java:86) ~[lucene-core-9.9.2.jar:9.9.2 a2939784c4ca60bc28bf488b5479c02fc2e5e22c - 2024-01-25 09:51:09]",
"at org.apache.lucene.search.DisjunctionSumScorer.score(DisjunctionSumScorer.java:41) ~[lucene-core-9.9.2.jar:9.9.2 a2939784c4ca60bc28bf488b5479c02fc2e5e22c - 2024-01-25 09:51:09]",
"at org.apache.lucene.search.DisjunctionScorer.score(DisjunctionScorer.java:178) ~[lucene-core-9.9.2.jar:9.9.2 a2939784c4ca60bc28bf488b5479c02fc2e5e22c - 2024-01-25 09:51:09]",
"at org.apache.lucene.search.DisjunctionMaxScorer.score(DisjunctionMaxScorer.java:65) ~[lucene-core-9.9.2.jar:9.9.2 a2939784c4ca60bc28bf488b5479c02fc2e5e22c - 2024-01-25 09:51:09]",
"at org.apache.lucene.search.DisjunctionScorer.score(DisjunctionScorer.java:178) ~[lucene-core-9.9.2.jar:9.9.2 a2939784c4ca60bc28bf488b5479c02fc2e5e22c - 2024-01-25 09:51:09]",
"at org.apache.lucene.search.ConjunctionScorer.score(ConjunctionScorer.java:61) ~[lucene-core-9.9.2.jar:9.9.2 a2939784c4ca60bc28bf488b5479c02fc2e5e22c - 2024-01-25 09:51:09]",
"at org.apache.lucene.search.DisjunctionMaxScorer.score(DisjunctionMaxScorer.java:65) ~[lucene-core-9.9.2.jar:9.9.2 a2939784c4ca60bc28bf488b5479c02fc2e5e22c - 2024-01-25 09:51:09]",
"at org.apache.lucene.search.DisjunctionScorer.score(DisjunctionScorer.java:178) ~[lucene-core-9.9.2.jar:9.9.2 a2939784c4ca60bc28bf488b5479c02fc2e5e22c - 2024-01-25 09:51:09]",
"at org.apache.lucene.search.ReqOptSumScorer.score(ReqOptSumScorer.java:266) ~[lucene-core-9.9.2.jar:9.9.2 a2939784c4ca60bc28bf488b5479c02fc2e5e22c - 2024-01-25 09:51:09]","at org.opensearch.neuralsearch.query.HybridQueryScorer.hybridScores(HybridQueryScorer.java:136) ~[?:?]","at org.opensearch.neuralsearch.search.HybridTopScoreDocCollector$1.collect(HybridTopScoreDocCollector.java:64) ~[?:?]",
"at org.opensearch.neuralsearch.query.HybridQueryScorer.hybridScores(HybridQueryScorer.java:136) ~[?:?]","at org.opensearch.neuralsearch.search.HybridTopScoreDocCollector$1.collect(HybridTopScoreDocCollector.java:64) 
```

this is part of the [original GH issue](https://github.com/opensearch-project/neural-search/issues/621) where the error has been reported

src/main/java/org/opensearch/neuralsearch/query/HybridScoreBlockBoundaryPropagator.java
martin-gaievski: There is a logic in Lucene that iterates on range of docs, e.g. [DocIdIterator.range](https://github.com/apache/lucene/blob/branch_9_10/lucene/core/src/java/org/apache/lucene/search/DocIdSetIterator.java#L96). For that doc ids are split into intervals and we need to know where in those intervals our range starts and/or ends. I understand that block is a pieces of range that is part of one interval, and boundary is the doc id that marks start/end of that block. 
This class is actually inspired by [similar class in Lucene](https://github.com/apache/lucene/blob/branch_9_10/lucene/core/src/java/org/apache/lucene/search/DisjunctionScoreBlockBoundaryPropagator.java), we just cannot call it directly as it's not public

msfroh: Essentially, term postings (matching doc IDs) are grouped into blocks of docs (I forget the number) and the start of the block holds some metadata ("impacts") including maximum term frequency of the term in any doc in the block and the minimum "norm" (doc length of a matching doc for any of the terms) for any doc in the block.

Using those values gives you a (potentially loose, if the max term frequency and minimum norm don't come from the same doc) upper bound on the possible score contribution of a doc in the block. Once you compute that across all terms, you can compute another (possibly even looser) upper bound on all scores for docs in this block. If that upper bound is below the score of the Nth hit when you're collecting the top N, you know you can safely skip the block -- none of its docs could possibly be good enough to win the title of TopDocs.

jmazanec15:
> > @martin-gaievski Could you provide additional context on why rewrite needs to be changed?
> 
> With existing code does the rewrite of sub-queries via [Rewriteable.rewrite](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/Rewriteable.java#L150) method, with new approach that is switched to [AbstractQueryBuilder.doRewrite](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/AbstractQueryBuilder.java#L295). For many queries that `doRewrite` method has query specific logic that is simply skipped with today's approach, for example [Term query](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/TermQueryBuilder.java#L185) or [Dismax query](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/DisMaxQueryBuilder.java#L209).

Makes sense. But is this required for 2 phase iterator for hybrid query or is it unrelated to that component of this change?

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -0,0 +1,248 @@
...
+
+        @Override
+        public Collector newCollector() {
+            if (Objects.isNull(maxScoreCollector)) {
+                maxScoreCollector = getCollector();
+                return maxScoreCollector;
+            } else {
+                Collector toReturnCollector = maxScoreCollector;
+                maxScoreCollector = null;
+                return toReturnCollector;
```
jmazanec15, line 222: Are we sure we can do this? From lucene, it seems that it must return a new collector each time: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/CollectorManager.java#L27-L28.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -0,0 +1,248 @@
...
+                .get();
+        }
+    }
+
+    private DocValueFormat[] getSortValueFormats(final SortAndFormats sortAndFormats) {
+        return sortAndFormats == null ? null : sortAndFormats.formats;
+    }
+
+    /**
+     * Implementation of the HybridCollector that reuses instance of collector on each even call. This allows caller to
```
jmazanec15, line 198: What is special about even calls?

martin-gaievski:
> > > @martin-gaievski Could you provide additional context on why rewrite needs to be changed?
> > 
> > 
> > With existing code does the rewrite of sub-queries via [Rewriteable.rewrite](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/Rewriteable.java#L150) method, with new approach that is switched to [AbstractQueryBuilder.doRewrite](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/AbstractQueryBuilder.java#L295). For many queries that `doRewrite` method has query specific logic that is simply skipped with today's approach, for example [Term query](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/TermQueryBuilder.java#L185) or [Dismax query](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/DisMaxQueryBuilder.java#L209).
> 
> Makes sense. But is this required for 2 phase iterator for hybrid query or is it unrelated to that component of this change?

It's needed in kind of indirect form - because rewrite is not completed for some queries AND we get scorers for all sub-queries [via scorerSupplier](https://github.com/martin-gaievski/neural-search/blob/hybrid_query_scorers_with_approximations/src/main/java/org/opensearch/neuralsearch/query/HybridQueryWeight.java#L75) that leads to some scorers being incorrectly set to `null`, meaning results of those sub-queries are ignored. With today's approach we still have those correct results from scorer

src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
```diff
@@ -40,12 +44,56 @@ public final class HybridQueryScorer extends Scorer {
 
     private final Map<Query, List<Integer>> queryToIndex;
 
+    private final DocIdSetIterator approximation;
+    HybridScoreBlockBoundaryPropagator disjunctionBlockPropagator;
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -113,16 +103,17 @@ private boolean isHybridQuery(final Query query, final SearchContext searchConte
...
 
-    private boolean hasNestedFieldOrNestedDocs(final Query query, final SearchContext searchContext) {
+    private static boolean hasNestedFieldOrNestedDocs(final Query query, final SearchContext searchContext) {
         return searchContext.mapperService().hasNested() && new NestedHelper(searchContext.mapperService()).mightMatchNestedDocs(query);
     }
 
-    private boolean isWrappedHybridQuery(final Query query) {
+    private static boolean isWrappedHybridQuery(final Query query) {
         return query instanceof BooleanQuery
             && ((BooleanQuery) query).clauses().stream().anyMatch(clauseQuery -> clauseQuery.getQuery() instanceof HybridQuery);
```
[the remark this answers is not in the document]
martin-gaievski: it's not recursive, we're checking 1 level only. This is needed for case when index has `nested` fields. In such case any incoming query is wrapped into bool query. We're unwrapping hybrid query for such case and use unwrapped instance in all next steps. There is another logic in the same area - we're not allow hybrid query to be wrapped into another compound query. All these changes are done under this PR https://github.com/opensearch-project/neural-search/pull/498

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -180,152 +171,68 @@ private void validateNestedBooleanQuery(final Query query, final int level) {
...
-            queryResult.terminatedEarly(false);
+            if (isHybridQuery(context.query(), context)) {
+                // adding collector manager for hybrid query
+                CollectorManager collectorManager;
+                try {
+                    collectorManager = HybridCollectorManager.createHybridCollectorManager(context);
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
+                }
+                Map<Class<?>, CollectorManager<? extends Collector, ReduceableSearchResult>> collectorManagersByManagerClass = context
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -180,152 +171,68 @@ private void validateNestedBooleanQuery(final Query query, final int level) {
...
-            List<ScoreDoc> result = new ArrayList<>();
-            result.add(createStartStopElementForHybridSearchResults(delimiterDocId));
-            for (TopDocs topDoc : topDocs) {
-                if (Objects.isNull(topDoc) || Objects.isNull(topDoc.scoreDocs)) {
-                    result.add(createDelimiterElementForHybridSearchResults(delimiterDocId));
-                    continue;
+        @Override
+        public void postProcess(SearchContext context) {
+            if (isHybridQuery(context.query(), context)) {
+                if (!context.shouldUseConcurrentSearch()) {
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -180,152 +171,68 @@ private void validateNestedBooleanQuery(final Query query, final int level) {
...
-        final HybridTopScoreDocCollector collector = new HybridTopScoreDocCollector(
-            numDocs,
-            new HitsThresholdChecker(Math.max(numDocs, searchContext.trackTotalHitsUpTo()))
-        );
+        private final AggregationProcessor delegateAggsProcessor;
 
-        searcher.search(query, collector);
+        @Override
+        public void preProcess(SearchContext context) {
+            delegateAggsProcessor.preProcess(context);
```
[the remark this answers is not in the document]
martin-gaievski: yes, it's needed as initial map of collector managers is set in that core aggregation processor

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -66,15 +54,17 @@ public boolean searchWith(
         final boolean hasFilterCollector,
         final boolean hasTimeout
     ) throws IOException {
-        if (isHybridQuery(query, searchContext)) {
+        if (!isHybridQuery(query, searchContext)) {
+            validateQuery(searchContext, query);
+            return super.searchWith(searchContext, searcher, query, collectors, hasFilterCollector, hasTimeout);
+        } else {
```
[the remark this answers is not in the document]
martin-gaievski: there can be a valid case when hybrid query is wrapped into another bool query, in case of nested fields. Please check my previous comment

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -113,16 +103,17 @@ private boolean isHybridQuery(final Query query, final SearchContext searchConte
...
 
-    private boolean isWrappedHybridQuery(final Query query) {
+    private static boolean isWrappedHybridQuery(final Query query) {
         return query instanceof BooleanQuery
             && ((BooleanQuery) query).clauses().stream().anyMatch(clauseQuery -> clauseQuery.getQuery() instanceof HybridQuery);
     }
 
-    private Query extractHybridQuery(final SearchContext searchContext, final Query query) {
+    @VisibleForTesting
+    protected Query extractHybridQuery(final SearchContext searchContext, final Query query) {
```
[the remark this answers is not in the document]
martin-gaievski: at this point we know that hybrid query must be present as `isHybridQuery` method in the same class returned `true`

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: great point. So we only do it for non concurrent search case because current queryPhaseSearcher in core does not use collector managers in search in `concurrent search` feature is not enabled ([ref](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/query/QueryPhase.java#L356)). In contrast for concurrent search they [do it properly with collector managers](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/query/ConcurrentQueryPhaseSearcher.java#L82)
It's important because `reduce` of collector manager is not get called for non concurrent search. We have to do it manually and for that we're using state, basically returning same collector that already has scores. 
This is suggested by reta and done in a same fashion as [core does it for global aggregations](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/aggregations/GlobalAggCollectorManager.java#L35-L43)

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: I probably formulate this in unclear way, basically the idea is that on every odd call we will create new collector and on every even call we will return that collector instead of creating a new one.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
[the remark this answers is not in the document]
martin-gaievski: correct, I just fixed this after Jack's comment

src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
```diff
@@ -40,12 +44,56 @@ public final class HybridQueryScorer extends Scorer {
...
+    private final TwoPhase twoPhase;
+
     public HybridQueryScorer(Weight weight, List<Scorer> subScorers) throws IOException {
+        this(weight, subScorers, ScoreMode.TOP_SCORES);
+    }
+
+    public HybridQueryScorer(Weight weight, List<Scorer> subScorers, ScoreMode scoreMode) throws IOException {
```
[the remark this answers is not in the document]
martin-gaievski: nope, changing to package protected

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
```diff
@@ -47,16 +46,35 @@ public HybridTopScoreDocCollector(int numHits, HitsThresholdChecker hitsThreshol
...
+                if (scorer instanceof HybridQueryScorer) {
+                    return (HybridQueryScorer) scorer;
+                }
+                for (Scorable.ChildScorable childScorable : scorer.getChildren()) {
+                    HybridQueryScorer hybridQueryScorer = getHybridQueryScorer(childScorable.child);
+                    if (hybridQueryScorer != null) {
+                        return hybridQueryScorer;
+                    }
+                }
+                return null;
```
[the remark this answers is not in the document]
martin-gaievski: this will definitely cause a NPE in that line. Let me add a code where we throw a runtime exception in this case, as we cannot proceed with score collections if all scorers are empty/null.

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -180,152 +171,68 @@ private void validateNestedBooleanQuery(final Query query, final int level) {
...
-        }
+    @Override
+    public AggregationProcessor aggregationProcessor(SearchContext searchContext) {
+        AggregationProcessor coreAggProcessor = super.aggregationProcessor(searchContext);
+        return new HybridAggregationProcessor(coreAggProcessor);
+    }
 
-        final QuerySearchResult queryResult = searchContext.queryResult();
+    @AllArgsConstructor
+    public static class HybridAggregationProcessor implements AggregationProcessor {
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -0,0 +1,248 @@
...
+        for (final Collector collector : collectors) {
+            if (collector instanceof MultiCollectorWrapper) {
+                for (final Collector sub : (((MultiCollectorWrapper) collector).getCollectors())) {
+                    if (sub instanceof HybridTopScoreDocCollector) {
+                        hybridTopScoreDocCollectors.add((HybridTopScoreDocCollector) sub);
+                    }
+                }
+            } else if (collector instanceof HybridTopScoreDocCollector) {
+                hybridTopScoreDocCollectors.add((HybridTopScoreDocCollector) collector);
+            }
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -0,0 +1,248 @@
...
+    @Override
+    abstract public Collector newCollector();
+
+    Collector getCollector() {
+        Collector hybridcollector = new HybridTopScoreDocCollector(numHits, hitsThresholdChecker);
+        return hybridcollector;
+    }
+
+    @Override
+    public ReduceableSearchResult reduce(Collection<Collector> collectors) {
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/query/HybridQueryWeight.java
```diff
@@ -5,36 +5,39 @@
...
  */
 public final class HybridQueryWeight extends Weight {
 
-    private final HybridQuery queries;
     // The Weights for our subqueries, in 1-1 correspondence
     private final List<Weight> weights;
 
     private final ScoreMode scoreMode;
 
+    static final int BOOLEAN_REWRITE_TERM_COUNT_THRESHOLD = 16;
```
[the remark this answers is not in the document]
martin-gaievski: ack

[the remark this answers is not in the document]
martin-gaievski: looks like we can replace this by existing constant of  max number of subqueries

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Features
 ### Enhancements
 ### Bug Fixes
+- Adding two phase iterator for hybrid query ([#624](https://github.com/opensearch-project/neural-search/pull/624))
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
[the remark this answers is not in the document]
navneet1v: Lets add enough logs to ensure that we understand why the scorer was null.

On the other side do we know in which cases the scorer can be null? and is it a valid case?

src/main/java/org/opensearch/neuralsearch/search/query/HybridAggregationProcessor.java
```diff
@@ -0,0 +1,77 @@
...
+                throw new RuntimeException(e);
+            }
+            context.queryCollectorManagers().put(HybridCollectorManager.class, collectorManager);
+        }
+    }
+
+    @Override
+    public void postProcess(SearchContext context) {
+        if (isHybridQuery(context.query(), context)) {
+            // for case when concurrent search is not enabled (default as of 2.12 release) reduce for collector
```
jmazanec15, line 49: Can you add link for reference?

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
jmazanec15: right, but why not reuse each time?

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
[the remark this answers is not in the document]
martin-gaievski: I don't have info on a cases when `scorer` can be null and it's a valid scenario. I'll add more logs

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: For concurrent search scenario it may cause problems. Core uses thread pool of runners to execute the search, in case we use state that state may be picked up and used in search in completely different slice of segment (I think that's the term they are using for blocks that one thread runner is using). To summarize this:

concurrent search - we cannot use state. reduce will be called so manual reduce call and collector's state both are not needed
non concurrent search - reduce must be called manually. it's safe to use state because it's single threaded execution

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: yes, that's correct. So the flow is:
- call 1: maxScoreCollector is null, we get new collector instance, maxScoreCollector = new collector instance, return maxScoreCollector
- call 2: maxScoreCollector != null, save instance referenced by maxScoreCollector, set maxScoreCollector to `null`, return saved instance
- call 3: same logic as in call 1, collector instance will be new (not same as in call 1)
- call 4: same logic as in call 2, collector instance will be the one that we've initialized in call 3

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: Discussed offline, initial question was more focused on non-concurrent search and if we can reuse collector for multiple calls, not in every second call.

We can, just the actual logic is to make max of two calls in that flow. 
[In here](https://github.com/opensearch-project/OpenSearch/blob/2.12/server/src/main/java/org/opensearch/search/query/QueryPhase.java#L352) collectors are created, and then used for search. And later in our code in aggregation processor we call [collector manager.reduce](https://github.com/martin-gaievski/neural-search/blob/hybrid_query_scorers_with_approximations/src/main/java/org/opensearch/neuralsearch/search/query/HybridAggregationProcessor.java#L71).
For every new search request [new collector manager is created](https://github.com/martin-gaievski/neural-search/blob/hybrid_query_scorers_with_approximations/src/main/java/org/opensearch/neuralsearch/search/query/HybridAggregationProcessor.java#L38) and this flow repeats.

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
```diff
@@ -47,20 +47,55 @@ public HybridTopScoreDocCollector(int numHits, HitsThresholdChecker hitsThreshol
...
+                    compoundQueryScorer = getHybridQueryScorer(scorer);
+                    if (Objects.isNull(compoundQueryScorer)) {
+                        log.error(
+                            String.format(Locale.ROOT, "cannot find scorer of type HybridQueryScorer in a hierarchy of scorer %s", scorer)
+                        );
+                    }
+                }
+            }
+
+            private HybridQueryScorer getHybridQueryScorer(final Scorable scorer) throws IOException {
```
msfroh, line 71: I'm a little worried about this logic, because it lifts the `HybridQueryScorer` up to the top, which doesn't necessarily sound right.

If you have a query like:

```
{
  "bool": {
    "should": [
      {
        "match": { ... }
      },
      {
        "hybrid": { ... }
      }
    ]
}
```

The score should be based on the sum of the match query and the hybrid query scores. From the logic in `collect`, it looks like it will just use the hybrid query score, right?

navneet1v: @msfroh this usecase is not supported. Hybrid query clause should be a top level query clause. We have such query we throw exceptions out.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -0,0 +1,248 @@
...
+
+import static org.opensearch.neuralsearch.search.util.HybridSearchResultFormatUtil.createDelimiterElementForHybridSearchResults;
+import static org.opensearch.neuralsearch.search.util.HybridSearchResultFormatUtil.createStartStopElementForHybridSearchResults;
+
+/**
+ * Collector manager based on HybridTopScoreDocCollector that allows users to parallelize counting the number of hits.
+ * In most cases it will be wrapped in MultiCollectorManager.
+ */
+@RequiredArgsConstructor
+public abstract class HybridCollectorManager implements CollectorManager<Collector, ReduceableSearchResult> {
```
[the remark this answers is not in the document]
navneet1v: its actually both. but @martin-gaievski can add more here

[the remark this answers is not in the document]
martin-gaievski: correct, that's both. It's needed in scope of two phase iterator because of the query search context, with today's implementation we're creating new collector specific to hybrid query and that effectively drops some logic from [QueryPhase.executeInternal](https://github.com/opensearch-project/OpenSearch/blob/2.12/server/src/main/java/org/opensearch/search/query/QueryPhase.java#L191) related to construction of that context.

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
martin-gaievski: [This is plugin code](https://github.com/martin-gaievski/neural-search/blob/hybrid_query_scorers_with_approximations/src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java#L149-L150) that does this validation

src/main/java/org/opensearch/neuralsearch/search/query/HybridAggregationProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        delegateAggsProcessor.postProcess(context);
+    }
+
+    private void reduceCollectorResults(SearchContext context) {
+        CollectorManager<?, ReduceableSearchResult> collectorManager = context.queryCollectorManagers().get(HybridCollectorManager.class);
+        try {
+            final Collection collectors = List.of(collectorManager.newCollector());
+            collectorManager.reduce(collectors).reduce(context.queryResult());
```
reta, line 71:
```suggestion
collectorManager.reduce(List.of()).reduce(context.queryResult());
```

The `HybridCollectorManager` for non-concurent case keeps single collector (see please `GlobalAggCollectorManagerWithSingleCollector` as an example)

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -0,0 +1,260 @@
...
+                isSingleShard,
+                trackTotalHitsUpTo,
+                searchContext.sort()
+            );
+    }
+
+    @Override
+    abstract public Collector newCollector();
+
+    Collector getCollector() {
```
reta, line 83: That should be a default impl for `newCollector()`, the `getCollector` is not needed here at all I believe

src/main/java/org/opensearch/neuralsearch/search/query/HybridAggregationProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+    public void preProcess(SearchContext context) {
+        delegateAggsProcessor.preProcess(context);
+
+        if (isHybridQuery(context.query(), context)) {
+            // adding collector manager for hybrid query
+            CollectorManager collectorManager;
+            try {
+                collectorManager = HybridCollectorManager.createHybridCollectorManager(context);
+            } catch (IOException e) {
+                throw new RuntimeException(e);
```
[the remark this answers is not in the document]
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridAggregationProcessor.java
martin-gaievski: ack

reta:
Thanks @martin-gaievski , the part related to HybridCollectorManager looks good to me, thank you for making the changes!

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
```diff
@@ -47,20 +47,55 @@ public HybridTopScoreDocCollector(int numHits, HitsThresholdChecker hitsThreshol
...
+                        return hybridQueryScorer;
+                    }
+                }
+                return null;
             }
 
             @Override
             public void collect(int doc) throws IOException {
+                if (Objects.isNull(compoundQueryScorer)) {
+                    throw new IllegalArgumentException("scorers are null for all sub-queries in hybrid query");
```
navneet1v, line 97: if all scorers are null, it means that there are no results. Can you try putting 2 text queries in Hybrid query clause where both results in not docs. and see if we are hitting this case.

@martin-gaievski

navneet1v (commented):
One comment. Overall code is good.

navneet1v (approved):

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
martin-gaievski: In such case this code is never executed and overall result has 0 total hits and empty collection of docs, something like

```
{
    "took": 226,
    "timed_out": false,
    "_shards": {
        "total": 2,
        "successful": 2,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "total": {
            "value": 0,
            "relation": "eq"
        },
        "max_score": 0.0,
        "hits": []
    }
}
```

navneet1v: Great. its just that code like this scares me a bit. But code is approved from my side.

jmazanec15 (approved):

Issue #508, opened 2023-11-03
[FEATURE] post_filter to be supported with Hybrid Search


ankitas3:
**Is your feature request related to a problem?**
Currently post_filter sent in query is not supported with Hybrid Query. The results are not reflected as per the filter applied.

**What solution would you like?**
post_filter should be supported and the results should be filtered from the hybrid query results.

ulan-yisaev:
Please implement this feature, it is very needed!

vamshin:
@ulan-yisaev we are actively working on this and hoping to make it to 2.13 release

vibrantvarun:
PR has been merged and we can close the issue now

PR #633, opened 2024-03-13
Support for post filter in hybrid query

vibrantvarun:
### Description
This PR adds Support for Post filter in hybrid query. 

The crux of the logic implemented in this PR is when the search request has post filter query in that, then instead of returning plain HybridTopDocCollector Object we would be returning FilteredCollectorObject which will have HybridTopDocCollector object wrapped underneath it.

Also This PR has unit tests and Integration tests for the post_filter scenario.

Testing example on local machine

1) Create index

```
http://localhost:9200/movies

{
  "settings": {
    "index": {
      "number_of_shards": 1,
      "number_of_replicas": 0
    },
    "index.search.concurrent_segment_search.enabled": false
  },
  "mappings": {
    "properties": {
      "name": {
        "type": "text"
      },
      "stock":{
        "type": "integer"
      }
    }
  }
}
```

2) Create pipeline 

```
http://localhost:9200/_search/pipeline/nlp-search-pipeline

{
  "phase_results_processors": [
    {
      "normalization-processor": {
        "normalization": {
          "technique": "min_max"
        },
        "combination": {
          "parameters": {
            "weights": [
              0.3,
              0.6,
              0.1
            ]
          }
        }
      }
    }
  ]
}

```

3) Add documents

```
{
  "name": "Dunes part 1",
  "stock":2
}
```

```
{
  "name": "Dunes part 2",
  "stock":26
}
```
```
{
  "name": "Mission Impossible 1",
  "stock":234
}
```
```
{
  "name": "Mission Impossible 2",
  "stock":238
}
```
```
{
  "name": "Avengers",
  "stock":300
}
```

4) search without post_filter
```
http://localhost:9200/movies/_search?search_pipeline=nlp-search-pipeline

{
   "query": {
      "hybrid":{
          "queries":[
              { 
                "match":{
                    "name": "mission"
                }
              },
              {
                "term":{
                    "name":{
                        "value":"part"
                    }
                }
              },
              {
              "range": {
                 "stock": {
                      "gte": 200,
                      "lte": 400
                  }
                }
              }
          ]
      }

  }
}


{
    "took": 12,
    "timed_out": false,
    "_shards": {
        "total": 1,
        "successful": 1,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "total": {
            "value": 5,
            "relation": "eq"
        },
        "max_score": 0.70000005,
        "hits": [
            {
                "_index": "movies",
                "_id": "2",
                "_score": 0.70000005,
                "_source": {
                    "name": "Mission Impossible 1",
                    "stock": 234
                }
            },
            {
                "_index": "movies",
                "_id": "3",
                "_score": 0.70000005,
                "_source": {
                    "name": "Mission Impossible 2",
                    "stock": 238
                }
            },
            {
                "_index": "movies",
                "_id": "4",
                "_score": 0.4,
                "_source": {
                    "name": "Avengers",
                    "stock": 300
                }
            },
            {
                "_index": "movies",
                "_id": "1",
                "_score": 0.3,
                "_source": {
                    "name": "Dunes part 1",
                    "stock": 2
                }
            },
            {
                "_index": "movies",
                "_id": "5",
                "_score": 0.3,
                "_source": {
                    "name": "Dunes part 2",
                    "stock": 26
                }
            }
        ]
    }
}
```

5) Search with post_filter

```
{
   "query": {
      "hybrid":{
          "queries":[
              { 
                "match":{
                    "name": "mission"
                }
              },
              {
                "term":{
                    "name":{
                        "value":"part"
                    }
                }
              },
              {
              "range": {
                 "stock": {
                      "gte": 200,
                      "lte": 400
                  }
                }
              }
          ]
      }

  },
  "post_filter":{
      "range": {
                 "stock": {
                      "gte": 230,
                      "lte": 400
                  }
       }
  }
}

{
    "took": 12,
    "timed_out": false,
    "_shards": {
        "total": 1,
        "successful": 1,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "total": {
            "value": 3,
            "relation": "eq"
        },
        "max_score": 0.70000005,
        "hits": [
            {
                "_index": "movies",
                "_id": "2",
                "_score": 0.70000005,
                "_source": {
                    "name": "Mission Impossible 1",
                    "stock": 234
                }
            },
            {
                "_index": "movies",
                "_id": "3",
                "_score": 0.70000005,
                "_source": {
                    "name": "Mission Impossible 2",
                    "stock": 238
                }
            },
            {
                "_index": "movies",
                "_id": "4",
                "_score": 0.4,
                "_source": {
                    "name": "Avengers",
                    "stock": 300
                }
            }
        ]
    }
}
```

### Issues Resolved
[508](https://github.com/opensearch-project/neural-search/issues/508)

### Check List
    - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
+++ b/src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
@@ -106,7 +106,7 @@ public abstract class HybridCollectorManager implements CollectorManager<Collect
         Collector hybridcollector = new HybridTopScoreDocCollector(numHits, hitsThresholdChecker);
         // Check if filterWeight is present. If it is present then return wrap Hybrid collector object underneath the FilteredCollector
         // object and return it.
-        return filterWeight == null ? hybridcollector : new FilteredCollector(hybridcollector, filterWeight);
+        return Objects.nonNull(filterWeight) ? new FilteredCollector(hybridcollector, filterWeight) : hybridcollector;
     }
 
     /**
--- a/src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
+++ b/src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
@@ -11,10 +11,15 @@ import org.apache.lucene.search.CollectorManager;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TopDocs;
 import org.apache.lucene.search.TotalHits;
+import org.apache.lucene.search.Weight;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.ScoreMode;
+import org.opensearch.common.lucene.search.FilteredCollector;
 import org.opensearch.common.lucene.search.TopDocsAndMaxScore;
 import org.opensearch.neuralsearch.search.HitsThresholdChecker;
 import org.opensearch.neuralsearch.search.HybridTopScoreDocCollector;
 import org.opensearch.search.DocValueFormat;
+import org.opensearch.search.internal.ContextIndexSearcher;
 import org.opensearch.search.internal.SearchContext;
 import org.opensearch.search.query.MultiCollectorWrapper;
 import org.opensearch.search.query.QuerySearchResult;
@@ -30,6 +35,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.Optional;
 
 import static org.opensearch.neuralsearch.search.util.HybridSearchResultFormatUtil.createDelimiterElementForHybridSearchResults;
 import static org.opensearch.neuralsearch.search.util.HybridSearchResultFormatUtil.createStartStopElementForHybridSearchResults;
@@ -46,6 +52,7 @@ public abstract class HybridCollectorManager implements CollectorManager<Collect
     private final boolean isSingleShard;
     private final int trackTotalHitsUpTo;
     private final SortAndFormats sortAndFormats;
+    private final Optional<Weight> filterWeight;
 
     /**
      * Create new instance of HybridCollectorManager depending on the concurrent search beeing enabled or disabled.
@@ -60,27 +67,42 @@ public abstract class HybridCollectorManager implements CollectorManager<Collect
         int numDocs = Math.min(searchContext.from() + searchContext.size(), totalNumDocs);
         int trackTotalHitsUpTo = searchContext.trackTotalHitsUpTo();
 
+        Weight filteringWeight = null;
+        // Check for post filter to create weight
+        if (Objects.nonNull(searchContext.parsedPostFilter())) {
+            Query filterQuery = searchContext.parsedPostFilter().query();
+            ContextIndexSearcher searcher = searchContext.searcher();
+            filteringWeight = searcher.createWeight(searcher.rewrite(filterQuery), ScoreMode.COMPLETE_NO_SCORES, 1f);
+        }
+
         return searchContext.shouldUseConcurrentSearch()
             ? new HybridCollectorConcurrentSearchManager(
                 numDocs,
                 new HitsThresholdChecker(Math.max(numDocs, searchContext.trackTotalHitsUpTo())),
                 isSingleShard,
                 trackTotalHitsUpTo,
-                searchContext.sort()
+                searchContext.sort(),
+                filteringWeight
             )
             : new HybridCollectorNonConcurrentManager(
                 numDocs,
                 new HitsThresholdChecker(Math.max(numDocs, searchContext.trackTotalHitsUpTo())),
                 isSingleShard,
                 trackTotalHitsUpTo,
-                searchContext.sort()
+                searchContext.sort(),
+                filteringWeight
             );
     }
 
     @Override
     public Collector newCollector() {
         Collector hybridcollector = new HybridTopScoreDocCollector(numHits, hitsThresholdChecker);
-        return hybridcollector;
+        // Check if filterWeight is present. If it is present then return wrap Hybrid collector object underneath the FilteredCollector
+        // object and return it.
+        if (filterWeight.isEmpty()) {
+            return hybridcollector;
+        }
+        return new FilteredCollector(hybridcollector, filterWeight.get());
     }
 
     /**
@@ -108,7 +130,10 @@ public abstract class HybridCollectorManager implements CollectorManager<Collect
                 }
             } else if (collector instanceof HybridTopScoreDocCollector) {
                 hybridTopScoreDocCollectors.add((HybridTopScoreDocCollector) collector);
-            }
+            } else if (collector instanceof FilteredCollector
+                && ((FilteredCollector) collector).getCollector() instanceof HybridTopScoreDocCollector) {
+                    hybridTopScoreDocCollectors.add((HybridTopScoreDocCollector) ((FilteredCollector) collector).getCollector());
+                }
         }
 
         if (!hybridTopScoreDocCollectors.isEmpty()) {
@@ -216,9 +241,10 @@ public abstract class HybridCollectorManager implements CollectorManager<Collect
             HitsThresholdChecker hitsThresholdChecker,
             boolean isSingleShard,
             int trackTotalHitsUpTo,
-            SortAndFormats sortAndFormats
+            SortAndFormats sortAndFormats,
+            Weight filteringWeight
         ) {
-            super(numHits, hitsThresholdChecker, isSingleShard, trackTotalHitsUpTo, sortAndFormats);
+            super(numHits, hitsThresholdChecker, isSingleShard, trackTotalHitsUpTo, sortAndFormats, Optional.ofNullable(filteringWeight));
             scoreCollector = Objects.requireNonNull(super.newCollector(), "collector for hybrid query cannot be null");
         }
 
@@ -245,9 +271,10 @@ public abstract class HybridCollectorManager implements CollectorManager<Collect
             HitsThresholdChecker hitsThresholdChecker,
             boolean isSingleShard,
             int trackTotalHitsUpTo,
-            SortAndFormats sortAndFormats
+            SortAndFormats sortAndFormats,
+            Weight filteringWeight
         ) {
-            super(numHits, hitsThresholdChecker, isSingleShard, trackTotalHitsUpTo, sortAndFormats);
+            super(numHits, hitsThresholdChecker, isSingleShard, trackTotalHitsUpTo, sortAndFormats, Optional.ofNullable(filteringWeight));
         }
     }
 }
--- a/src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
+++ b/src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
@@ -39,6 +39,7 @@ import java.util.Optional;
 
 import static org.opensearch.neuralsearch.search.util.HybridSearchResultFormatUtil.createDelimiterElementForHybridSearchResults;
 import static org.opensearch.neuralsearch.search.util.HybridSearchResultFormatUtil.createStartStopElementForHybridSearchResults;
+import static org.opensearch.neuralsearch.util.HybridQueryUtil.boost_factor;
 
 /**
  * Collector manager based on HybridTopScoreDocCollector that allows users to parallelize counting the number of hits.
@@ -52,7 +53,7 @@ public abstract class HybridCollectorManager implements CollectorManager<Collect
     private final boolean isSingleShard;
     private final int trackTotalHitsUpTo;
     private final SortAndFormats sortAndFormats;
-    private final Optional<Weight> filterWeight;
+    private final Optional<Weight> optionalFilterWeight;
 
     /**
      * Create new instance of HybridCollectorManager depending on the concurrent search beeing enabled or disabled.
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -60,27 +67,42 @@ public static CollectorManager createHybridCollectorManager(final SearchContext
         int numDocs = Math.min(searchContext.from() + searchContext.size(), totalNumDocs);
         int trackTotalHitsUpTo = searchContext.trackTotalHitsUpTo();
 
+        Weight filteringWeight = null;
+        // Check for post filter to create weight
```
martin-gaievski, line 71: comment needs to be more detailed

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -60,27 +67,42 @@ public static CollectorManager createHybridCollectorManager(final SearchContext
...
+        if (Objects.nonNull(searchContext.parsedPostFilter())) {
+            Query filterQuery = searchContext.parsedPostFilter().query();
```
martin-gaievski, line 73: we should check if query object isn't null

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -60,27 +67,42 @@ public static CollectorManager createHybridCollectorManager(final SearchContext
...
+            ContextIndexSearcher searcher = searchContext.searcher();
+            filteringWeight = searcher.createWeight(searcher.rewrite(filterQuery), ScoreMode.COMPLETE_NO_SCORES, 1f);
```
martin-gaievski, line 75: why these params are used for create weight: COMPLETE_NO_SCORES score mode, and boost factor 1? Additionally, boost factor can be referenced via some constant I believe.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -46,6 +52,7 @@ public abstract class HybridCollectorManager implements CollectorManager<Collect
     private final boolean isSingleShard;
     private final int trackTotalHitsUpTo;
     private final SortAndFormats sortAndFormats;
+    private final Optional<Weight> filterWeight;
```
martin-gaievski, line 55: if you want to stick with Optional approach add Optional suffix to the variable name

src/test/java/org/opensearch/neuralsearch/query/HybridQueryAggregationsIT.java
```diff
@@ -182,6 +182,57 @@ public void testAggregationNotSupportedConcurrentSearch_whenUseSamplerAgg_thenSu
         }
     }
 
+    @SneakyThrows
+    public void testPostFilter_WhenConcurrentSearchNotEnabled_thenSuccessful() {
```
martin-gaievski, line 186: can we add one more test case for case when concurrent search is enabled. It's GA since 2.12, probably will be enabled by default in coming versions

src/test/java/org/opensearch/neuralsearch/query/HybridQueryAggregationsIT.java
```diff
@@ -182,6 +182,57 @@ public void testAggregationNotSupportedConcurrentSearch_whenUseSamplerAgg_thenSu
...
+        updateClusterSettings("search.concurrent_segment_search.enabled", true);
+        try {
+            prepareResources(TEST_MULTI_DOC_INDEX_WITH_TEXT_AND_INT_MULTIPLE_SHARDS, SEARCH_PIPELINE);
+
+            AggregationBuilder aggsBuilder = AggregationBuilders.sum(SUM_AGGREGATION_NAME).field(INTEGER_FIELD_1);
```
martin-gaievski, line 191: why we're adding aggs here, test supposed to test post filter?

martin-gaievski: I also like to see more complex filter query, maybe bool with should, your current query for `range` can be one clause, also add one more, for example `match`

src/test/java/org/opensearch/neuralsearch/search/query/HybridCollectorManagerTests.java
```diff
@@ -120,6 +124,90 @@ public void testNewCollector_whenConcurrentSearch_thenSuccessful() {
...
+        when(searchContext.queryCollectorManagers()).thenReturn(classCollectorManagerMap);
+        when(searchContext.shouldUseConcurrentSearch()).thenReturn(true);
+
+        CollectorManager hybridCollectorManager = HybridCollectorManager.createHybridCollectorManager(searchContext);
+        assertNotNull(hybridCollectorManager);
+        assertTrue(hybridCollectorManager instanceof HybridCollectorManager.HybridCollectorConcurrentSearchManager);
+
+        Collector collector = hybridCollectorManager.newCollector();
+        assertNotNull(collector);
+        assertTrue(collector instanceof FilteredCollector);
```
martin-gaievski, line 205: we should also check that hybrid collector is wrapped by this filtered collector

src/test/java/org/opensearch/neuralsearch/query/HybridQueryAggregationsIT.java
```diff
@@ -182,6 +182,57 @@ public void testAggregationNotSupportedConcurrentSearch_whenUseSamplerAgg_thenSu
...
+                Map<String, Object> source = (Map<String, Object>) oneHit.get("_source");
+                int docIndex = (int) source.get(INTEGER_FIELD_1);
+                docIndexes.add(docIndex);
+            }
+            assertEquals(0, docIndexes.stream().filter(docIndex -> docIndex < 3000 || docIndex > 5000).count());
+        } finally {
+            wipeOfTestResources(TEST_MULTI_DOC_INDEX_WITH_TEXT_AND_INT_MULTIPLE_SHARDS, null, null, SEARCH_PIPELINE);
+        }
+    }
+
```
martin-gaievski, line 235: We also need negative tests, e.g. one where filter should have some matches but hybrid query does return empty results and one more where hybrid query return something but filter cuts off all documents.

I'm ok if tests are added in another PR, but you need to tests these scenarios at least manually. Maybe post results of manual testing to PR description

martin-gaievski (commented):
Overall code looks good, left just few comments

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
vibrantvarun: This check would become redundant as when parsedPostFilter is set and if query is null then below line will handle that situation https://github.com/opensearch-project/OpenSearch/blob/77a119319c14b0c1fa45dc85a618bf2429576b1e/server/src/main/java/org/opensearch/search/SearchService.java#L1255

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
vibrantvarun: ScoreMode COMPLETE_NO_SCORES will be passed as post_filter does not contribute in scoring. COMPLETE_NO_SCORES means it is not a scoring clause. Boost factor 1f is taken because if boost is multiplicative of 1 then it means "no boost". Previously this code in OpenSearch looked like https://github.com/opensearch-project/OpenSearch/commit/36a5cf8f35e5cbaa1ff857b5a5db8c02edc1a187

src/test/java/org/opensearch/neuralsearch/query/HybridQueryAggregationsIT.java
vibrantvarun: Check PR description

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: To me we always need to do basic validation for cases like this, this searchContext is constructed not even in the plugin, and logic that you've referenced can change anytime, or new logic can be added. Let's not rely on good intensions in other components

src/test/java/org/opensearch/neuralsearch/query/HybridQueryAggregationsIT.java
martin-gaievski: checked, nothing from my request is in description. Can you please be specific, which section of description answers my requests?

src/main/java/org/opensearch/neuralsearch/util/HybridQueryUtil.java
```diff
@@ -21,6 +21,8 @@
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 public class HybridQueryUtil {
 
+    public static final float boost_factor = 1f;
```
navneet1v, line 24: lets move this to HybridCollectorManagerClass, as it is only getting used there.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -216,9 +247,10 @@ public HybridCollectorNonConcurrentManager(
             HitsThresholdChecker hitsThresholdChecker,
             boolean isSingleShard,
             int trackTotalHitsUpTo,
-            SortAndFormats sortAndFormats
+            SortAndFormats sortAndFormats,
+            Weight filteringWeight
```
navneet1v, line 251: we should make this parameter as Optional<Weight> optionalFilterWeight.

navneet1v (commented):
Overall code looks good to me. Just resolve the comments and then its ready to be shipped.

navneet1v:
@vibrantvarun lets ensure the GH actions are successful.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
vibrantvarun: If i do that then in newCollector method 

```
 if (optionalFilterWeight.isEmpty()) {
            return hybridcollector;
        }
```

This code breaks with below error

```
{
    "error": {
        "root_cause": [
            {
                "type": "null_pointer_exception",
                "reason": "Cannot invoke \"java.util.Optional.isEmpty()\" because \"this.optionalFilterWeight\" is null"
            }
        ],
        "type": "search_phase_execution_exception",
        "reason": "all shards failed",
        "phase": "query",
        "grouped": true,
        "failed_shards": [
            {
                "shard": 0,
                "index": "movies",
                "node": "MpUK-BFmRRiigVIII84uIQ",
                "reason": {
                    "type": "null_pointer_exception",
                    "reason": "Cannot invoke \"java.util.Optional.isEmpty()\" because \"this.optionalFilterWeight\" is null"
                }
            }
        ],
        "caused_by": {
            "type": "null_pointer_exception",
            "reason": "Cannot invoke \"java.util.Optional.isEmpty()\" because \"this.optionalFilterWeight\" is null",
            "caused_by": {
                "type": "null_pointer_exception",
                "reason": "Cannot invoke \"java.util.Optional.isEmpty()\" because \"this.optionalFilterWeight\" is null"
            }
        }
    },
    "status": 500
}
```

src/test/java/org/opensearch/neuralsearch/query/HybridQueryAggregationsIT.java
vibrantvarun: Synced with @martin-gaievski offline and verified all the test cases

navneet1v: @vibrantvarun lets add as java doc on the tests where it is not clear what test scenarios a test case is handling

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: @vibrantvarun did you construct optional with [ofNullable](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#ofNullable-T-) method?

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -46,6 +52,9 @@ public abstract class HybridCollectorManager implements CollectorManager<Collect
     private final boolean isSingleShard;
     private final int trackTotalHitsUpTo;
     private final SortAndFormats sortAndFormats;
+    @Nullable
+    private final Weight filterWeight;
+    public static final float boost_factor = 1f;
```
navneet1v, line 57: this should be private

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
vibrantvarun: yes. @navneet1v and me hopped on call and determined that it is kind of potiential bug and not the right practice to use Optional in class variables.

martin-gaievski: wonder what is the potential bug, it's not a blocker for this PR but using optional definitely isn't a bug

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -60,27 +69,44 @@ public static CollectorManager createHybridCollectorManager(final SearchContext
...
             );
     }
 
     @Override
     public Collector newCollector() {
         Collector hybridcollector = new HybridTopScoreDocCollector(numHits, hitsThresholdChecker);
-        return hybridcollector;
+        // Check if filterWeight is present. If it is present then return wrap Hybrid collector object underneath the FilteredCollector
+        // object and return it.
+        return filterWeight == null ? hybridcollector : new FilteredCollector(hybridcollector, filterWeight);
```
martin-gaievski, line 109: for same checks we used Objects.isNull is other places in the code

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
vibrantvarun: @navneet1v would you like to answer martin question

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
vibrantvarun: As discussed with @navneet1v, this static call is call made to Objects class for just a simple check. Therefore, we decided to go with simple null check. Initially my plan was to add Objects.nonNull check but @navneet1v insisted to use simple check to avoid extra call.

navneet1v: @vibrantvarun lets make things consistent in this PR for now.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
navneet1v: It was not a bug. The implementation was not correct due to which above exception happened.

but overall when we used optional on class member the IDE was adding warnings. Hence I suggested we can skip that. It was more of a coding practice and not a serious issue

navneet1v (approved):

PR #670, opened 2024-04-04
Allowing execution of hybrid query on index alias with filters

martin-gaievski:
### Description
Allowing scenario when hybrid query is executed for index alias that has filter. Today we're blocking such queries because of high level bool query that wraps the hybrid query to incorporate filtering logic. By design we are blocking such queries, hybrid query must be a high level query. 

To change this we doing following:
- extract hybrid query from compound query with the alias filter
- wrap each sub query of the hybrid query into bool query in a following form:
 ```
   bool : {
      must: [
        <original_sub_query>
      ],
      filter: {
        <alias_filter>
      }
   }
```

- rewrite hybrid query to following form:
```
 hybrid: {
      queries: [
          { sub_query1_wrapped_into_bool },
          { sub_query2_wrapped_into_bool }
      ]
 }
```

Part of this change is enhanced logic of query phase searcher for checking if the incoming query is  "hybrid query". We need to add a special case when index alias filter is present and incoming query is a bool query that wraps hybrid query.

We're changing logic for nested fields case by making it similar to alias filters. Instead of simply removing filter query we're rewriting hybrid query and adding filter query to every sub-query 

In addition to new integ tests I've run few scenarios manually. Below is example that is similar to one reported in original GH issue:

Create index with keyword and integer fields, ingest following 8 documents:

```
POST /_bulk

{ "index": { "_index": "my-nlp-index" } }
{ "category": "permission", "doc_keyword": "workable", "doc_index": 4976, "doc_price": 100}
{ "index": { "_index": "my-nlp-index" } }
{ "category": "sister", "doc_keyword": "angry", "doc_index": 2231, "doc_price": 200 }
{ "index": { "_index": "my-nlp-index" } }
{ "category": "hair", "doc_keyword": "likeable", "doc_price": 25 }
{ "index": { "_index": "my-nlp-index" } }
{ "category": "editor", "doc_index": 9871, "doc_price": 30 }
{ "index": { "_index": "my-nlp-index" } }
{ "category": "statement", "doc_keyword": "entire", "doc_index": 8242, "doc_price": 350  } 
{ "index": { "_index": "my-nlp-index" } }
{ "category": "statement", "doc_keyword": "idea", "doc_index": 5212, "doc_price": 200  } 
{ "index": { "_index": "my-nlp-index" } }
{ "category": "editor", "doc_keyword": "bubble", "doc_index": 1298, "doc_price": 130 } 
{ "index": { "_index": "my-nlp-index" } }
{ "category": "editor", "doc_keyword": "bubble", "doc_index": 521, "doc_price": 75  } 
```

create new index alias with filter:

```
POST /_aliases
{
    "actions": [
        {
            "add": {
                "index": "my-nlp-index",
                "alias": "alias_filter_1",
                "filter": {
                    "bool": {
                        "must_not": {
                            "term": {
                                "category": "statement"
                            }
                        }
                    }
                }
            }
        }
    ]
}
```
run following query against index alias:
```
GET /alias_filter_1/_search?search_pipeline=nlp-search-pipeline
{
    "query": {
        "hybrid": {
            "queries": [
                {
                    "range": {
                        "doc_index": {
                            "gte": 20,
                            "lte": 100
                        }
                    }
                },
                {
                    "bool": {
                        "should": [
                            {
                                "match": {
                                    "doc_keyword": "likeable"
                                }
                            },
                            {
                                "term": {
                                    "category": "statement"
                                }
                            }
                        ]
                    }
                }
            ]
        }
    }
}
```
following is result of the query. filter has been applied, but it's not part of the query itself, it's  pulled based on the alias name:
```
{
    "took": 5,
    "timed_out": false,
    "_shards": {
        "total": 1,
        "successful": 1,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "total": {
            "value": 1,
            "relation": "eq"
        },
        "max_score": 0.5,
        "hits": [
            {
                "_index": "my-nlp-index",
                "_id": "jNJ7p44B1ntc6ZjxQK8q",
                "_score": 0.5,
                "_source": {
                    "category": "hair",
                    "doc_keyword": "likeable",
                    "doc_price": 25
                }
            }
        ]
    }
}
```

### Issues Resolved
https://github.com/opensearch-project/neural-search/issues/627

### Check List
    - [X] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
+++ b/src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
@@ -69,7 +69,6 @@ public class HybridQueryPhaseSearcher extends QueryPhaseSearcherWrapper {
             }
             HybridQuery hybridQuery = (HybridQuery) booleanClauses.stream().findFirst().get().getQuery();
             List<Query> filterQueries = booleanClauses.stream()
-                .skip(1)
                 .filter(clause -> BooleanClause.Occur.FILTER == clause.getOccur())
                 .map(BooleanClause::getQuery)
                 .collect(Collectors.toList());
--- a/src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
+++ b/src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
@@ -7,6 +7,7 @@ package org.opensearch.neuralsearch.search.query;
 import java.io.IOException;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
 import org.apache.lucene.search.BooleanClause;
@@ -59,22 +60,20 @@ public class HybridQueryPhaseSearcher extends QueryPhaseSearcherWrapper {
 
     @VisibleForTesting
     protected Query extractHybridQuery(final SearchContext searchContext, final Query query) {
-        if (hasNestedFieldOrNestedDocs(query, searchContext)
+        if ((hasAliasFilter(query, searchContext) || hasNestedFieldOrNestedDocs(query, searchContext))
             && isWrappedHybridQuery(query)
             && !((BooleanQuery) query).clauses().isEmpty()) {
-            // extract hybrid query and replace bool with hybrid query
-            List<BooleanClause> booleanClauses = ((BooleanQuery) query).clauses();
-            if (booleanClauses.isEmpty() || booleanClauses.get(0).getQuery() instanceof HybridQuery == false) {
-                throw new IllegalStateException("cannot process hybrid query due to incorrect structure of top level bool query");
-            }
-            return booleanClauses.get(0).getQuery();
-        } else if (hasAliasFilter(query, searchContext) && isWrappedHybridQuery(query) && !((BooleanQuery) query).clauses().isEmpty()) {
             List<BooleanClause> booleanClauses = ((BooleanQuery) query).clauses();
             if (!(booleanClauses.get(0).getQuery() instanceof HybridQuery)) {
                 throw new IllegalStateException("cannot process hybrid query due to incorrect structure of top level query");
             }
-            HybridQuery hybridQuery = (HybridQuery) booleanClauses.get(0).getQuery();
-            HybridQuery hybridQueryWithFilter = new HybridQuery(hybridQuery.getSubQueries(), searchContext.aliasFilter());
+            HybridQuery hybridQuery = (HybridQuery) booleanClauses.stream().findFirst().get().getQuery();
+            List<Query> filterQueries = booleanClauses.stream()
+                .skip(1)
+                .filter(clause -> BooleanClause.Occur.FILTER == clause.getOccur())
+                .map(BooleanClause::getQuery)
+                .collect(Collectors.toList());
+            HybridQuery hybridQueryWithFilter = new HybridQuery(hybridQuery.getSubQueries(), filterQueries);
             return hybridQueryWithFilter;
         }
         return query;
--- a/src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
+++ b/src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
@@ -14,7 +14,6 @@ import org.apache.lucene.search.BooleanQuery;
 import org.apache.lucene.search.Query;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.index.mapper.MapperService;
-import org.opensearch.index.search.NestedHelper;
 import org.opensearch.neuralsearch.query.HybridQuery;
 import org.opensearch.search.aggregations.AggregationProcessor;
 import org.opensearch.search.internal.ContextIndexSearcher;
@@ -25,6 +24,8 @@ import org.opensearch.search.query.QueryPhaseSearcherWrapper;
 
 import lombok.extern.log4j.Log4j2;
 
+import static org.opensearch.neuralsearch.util.HybridQueryUtil.hasAliasFilter;
+import static org.opensearch.neuralsearch.util.HybridQueryUtil.hasNestedFieldOrNestedDocs;
 import static org.opensearch.neuralsearch.util.HybridQueryUtil.isHybridQuery;
 
 /**
@@ -51,10 +52,6 @@ public class HybridQueryPhaseSearcher extends QueryPhaseSearcherWrapper {
         }
     }
 
-    private static boolean hasNestedFieldOrNestedDocs(final Query query, final SearchContext searchContext) {
-        return searchContext.mapperService().hasNested() && new NestedHelper(searchContext.mapperService()).mightMatchNestedDocs(query);
-    }
-
     private static boolean isWrappedHybridQuery(final Query query) {
         return query instanceof BooleanQuery
             && ((BooleanQuery) query).clauses().stream().anyMatch(clauseQuery -> clauseQuery.getQuery() instanceof HybridQuery);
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by martin-gaievski** [comment]
BWC will keep failing unless all dependent repos are switched to 2.14 snapshot version, in particular: knn, ml-commons, common-utils. Also this PR should be merged for neural-search repo: https://github.com/opensearch-project/neural-search/pull/653

File: src/test/java/org/opensearch/neuralsearch/query/HybridQueryIT.java [hunk 1]
```diff
@@ -578,6 +579,54 @@ public void testRequestCache_whenMultipleShardsQueryReturnResults_thenSuccessful
         }
     }
 
+    @SneakyThrows
+    public void testIndexAlias_whenHybridQueryAndIndexAliasHasFilter_thenSuccess() {
```
navneet1v, line 583: @martin-gaievski can we add a test when there is both nested fields and index alias filters.

src/main/java/org/opensearch/neuralsearch/util/HybridQueryUtil.java
```diff
@@ -24,48 +26,58 @@ public class HybridQueryUtil {
...
+                // we have already checked if query in instance of Boolean in higher level else if condition
+                return ((BooleanQuery) query).clauses()
+                    .stream()
+                    .filter(clause -> clause.getQuery() instanceof HybridQuery == false)
+                    .allMatch(
+                        clause -> clause.getOccur() == BooleanClause.Occur.FILTER
+                            && clause.getQuery() instanceof FieldExistsQuery
+                            && SeqNoFieldMapper.PRIMARY_TERM_NAME.equals(((FieldExistsQuery) clause.getQuery()).getField())
+                    );
+            } else if (hasAliasFilter(query, searchContext)) {
```
navneet1v, line 62: rather than if else I think we should have if cases for different conditions. I want to know the case when we have both nested fields and alias filter in the query what we will happen?

src/test/java/org/opensearch/neuralsearch/query/HybridQueryIT.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/util/HybridQueryUtil.java
martin-gaievski: I didn't test this scenario, let me check how the query will be constructed by core. I hope it's not double wrapped, but both filters are added to one parent bool query.

martin-gaievski: You were right @navneet1v , we're not handling scenario with both nested fields and index alias filter properly. Basically nested fields that is first check that catches the query, kicks in and then query is failing. I'm reworking that logic now, in case of either of nested fields or alias filter or both we'll be doing same: get all filters from the parent bool and apply all those filters to every sub-query of the hybrid query. 
Example: if both nested and alias filter are there system sends us this query
```
  bool : {
     must: [
       hybrid_query
     ],
     filter: {
       alias_filter1,
       alias_filter2,
       field_exists (added for nested field)
     }
  }
```
we'll rewrite it to following form:
```
 hybrid: {
      queries: [
          bool : {
              must: [
                 <sub_query_1>
               ],
               filter: {
                  alias_filter1,
                  alias_filter2,
                  field_exists (added for nested field)
                }
          },
          bool : {
              must: [
                 <sub_query_2>
               ],
               filter: {
                  alias_filter1,
                  alias_filter2,
                  field_exists (added for nested field)
                }
          }
      ]
 }
```

navneet1v: Good we are able to catch this. :) Lets fix this. Can we somehow build our code in a fashion that we don't face this issue again. :D

src/test/java/org/opensearch/neuralsearch/query/HybridQueryIT.java
martin-gaievski: Extend this test case, now both criteria are there - nested fields in index mapping and alias with filter

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -51,26 +53,28 @@ public boolean searchWith(
...
-
     private static boolean isWrappedHybridQuery(final Query query) {
         return query instanceof BooleanQuery
             && ((BooleanQuery) query).clauses().stream().anyMatch(clauseQuery -> clauseQuery.getQuery() instanceof HybridQuery);
     }
 
     @VisibleForTesting
     protected Query extractHybridQuery(final SearchContext searchContext, final Query query) {
-        if (hasNestedFieldOrNestedDocs(query, searchContext)
+        if ((hasAliasFilter(query, searchContext) || hasNestedFieldOrNestedDocs(query, searchContext))
```
martin-gaievski, line 63: processing is similar for both cases - we need to rewrite query and add filter(s) to each sub query

navneet1v: cool..

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -51,26 +53,28 @@ public boolean searchWith(
...
             List<BooleanClause> booleanClauses = ((BooleanQuery) query).clauses();
-            if (booleanClauses.isEmpty() || booleanClauses.get(0).getQuery() instanceof HybridQuery == false) {
-                throw new IllegalStateException("cannot process hybrid query due to incorrect structure of top level bool query");
+            if (!(booleanClauses.get(0).getQuery() instanceof HybridQuery)) {
+                throw new IllegalStateException("cannot process hybrid query due to incorrect structure of top level query");
             }
-            return booleanClauses.get(0).getQuery();
+            HybridQuery hybridQuery = (HybridQuery) booleanClauses.stream().findFirst().get().getQuery();
+            List<Query> filterQueries = booleanClauses.stream()
+                .skip(1)
```
navneet1v, line 72: why this?

martin-gaievski: first clause in this case will be hybrid query, we've checked in this is wrapped hybrid query earlier. the rest of clauses that are of type `filter` we should use as filters and rewrite sub-queries

martin-gaievski: the structure of query is something like this:
```
  bool : {
     must: [
       hybrid_query
     ],
     filter: {
       alias_filter1,
       alias_filter2,
       field_exists (added for nested field)
     }
  }
```
in java that is:
```
  bool : {
     clause1: {
       hybrid_query,
       occur: must
     },
     clause2: {
       alias_filter1,
       occur: filter
     },
     clause3: {
       alias_filter2,
       occur: filter
     },
     clause4: {
       field_exists,
       occur: filter
     }
  }
```

navneet1v: can we use this occur to get all the fiters out rather than doing skip(1)

martin-gaievski: it's actually there on the next line `.filter(clause -> BooleanClause.Occur.FILTER == clause.getOccur())`, that skip(1) is like a small optimization that only removes one element that we know isn't a filter for sure. But deleting that `skip` function will not harm, let me create new commit

martin-gaievski: pushed corresponding code change

navneet1v: its just good that we don't skip like this.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryIT.java
```diff
@@ -578,6 +579,54 @@ public void testRequestCache_whenMultipleShardsQueryReturnResults_thenSuccessful
         }
     }
 
+    @SneakyThrows
+    public void testWrappedQueryWithFilter_whenIndexAliasHasFilterAndIndexWithNestedFields_thenSuccess() {
```
navneet1v, line 583: can we have an integ test where we just alias filters.

martin-gaievski: sure, added test specifically for that scenario

navneet1v (approved):

jmazanec15 (approved):

PR #687, opened 2024-04-11
fix map type validation issue in processors

zane-neo:
### Description
When user uses map type configuration in processors, the validation will validate all other fields instead of only the configured fields, sometimes if other fields has unsupported types, user will get exception which is not expected.

Multiple processors uses similar validation logic but code are duplicated, in this PR a new Util is been added to extract the duplicated code to a common place and reused by different processors.

### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/2309

### Check List
    - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
+++ b/src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.util;
+
+import org.apache.commons.lang3.StringUtils;
+import org.opensearch.cluster.metadata.IndexMetadata;
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.env.Environment;
+import org.opensearch.index.mapper.IndexFieldMapper;
+import org.opensearch.index.mapper.MapperService;
+
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Objects;
+
+public class ProcessorDocumentUtils {
+
+    public static long getMaxDepth(Map<String, Object> sourceAndMetadataMap, ClusterService clusterService, Environment environment) {
+        String indexName = sourceAndMetadataMap.get(IndexFieldMapper.NAME).toString();
+        IndexMetadata indexMetadata = clusterService.state().metadata().index(indexName);
+        if (indexMetadata != null) {
+            Settings settings = indexMetadata.getSettings();
+            return MapperService.INDEX_MAPPING_DEPTH_LIMIT_SETTING.get(settings);
+        }
+        return MapperService.INDEX_MAPPING_DEPTH_LIMIT_SETTING.get(environment.settings());
+    }
+
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    public static void validateMapTypeValue(
+        final String sourceKey,
+        final Map<String, Object> sourceValue,
+        final Object fieldMap,
+        final int depth,
+        final long maxDepth
+    ) {
+        if (sourceValue == null) return; // allow map type value to be null.
+        validateDepth(sourceKey, depth, maxDepth);
+        if (!(fieldMap instanceof Map)) { // source value is map type means configuration has to be map type
+            throw new IllegalArgumentException(
+                String.format(
+                    Locale.getDefault(),
+                    "[%s] configuration doesn't match actual value type, configuration type is: %s, actual value type is: %s",
+                    sourceKey,
+                    fieldMap.getClass().getName(),
+                    sourceValue.getClass().getName()
+                )
+            );
+        }
+        // next level validation, only validate the keys in configuration.
+        ((Map<String, Object>) fieldMap).forEach((key, nextFieldMap) -> {
+            Object nextSourceValue = sourceValue.get(key);
+            if (nextSourceValue != null) {
+                if (nextSourceValue instanceof List) {
+                    validateListTypeValue(key, (List) nextSourceValue, fieldMap, depth + 1, maxDepth);
+                } else if (nextSourceValue instanceof Map) {
+                    validateMapTypeValue(key, (Map<String, Object>) nextSourceValue, nextFieldMap, depth + 1, maxDepth);
+                } else if (!(nextSourceValue instanceof String)) {
+                    throw new IllegalArgumentException("map type field [" + key + "] is neither string nor nested type, cannot process it");
+                } else if (StringUtils.isBlank((String) nextSourceValue)) {
+                    throw new IllegalArgumentException("map type field [" + key + "] has empty string value, cannot process it");
+                }
+            }
+        });
+    }
+
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    private static void validateListTypeValue(String sourceKey, List sourceValue, Object fieldMap, int depth, long maxDepth) {
+        validateDepth(sourceKey, depth, maxDepth);
+        if (sourceValue == null || sourceValue.isEmpty()) return;
+        Object firstNonNullElement = sourceValue.stream().filter(Objects::nonNull).findFirst().orElse(null);
+        if (firstNonNullElement == null) return;
+        for (Object element : sourceValue) {
+            if (firstNonNullElement instanceof List) { // nested list case.
+                validateListTypeValue(sourceKey, (List) element, fieldMap, depth + 1, maxDepth);
+            } else if (firstNonNullElement instanceof Map) {
+                validateMapTypeValue(sourceKey, (Map<String, Object>) element, ((Map) fieldMap).get(sourceKey), depth + 1, maxDepth);
+            } else if (!(firstNonNullElement instanceof String)) {
+                throw new IllegalArgumentException("list type field [" + sourceKey + "] has non string value, cannot process it");
+            } else {
+                if (element == null) {
+                    throw new IllegalArgumentException("list type field [" + sourceKey + "] has null, cannot process it");
+                } else if (!(element instanceof String)) {
+                    throw new IllegalArgumentException("list type field [" + sourceKey + "] has non string value, cannot process it");
+                } else if (StringUtils.isBlank(element.toString())) {
+                    throw new IllegalArgumentException("list type field [" + sourceKey + "] has empty string, cannot process it");
+                }
+            }
+        }
+    }
+
+    private static void validateDepth(String sourceKey, int depth, long maxDepth) {
+        if (depth > maxDepth) {
+            throw new IllegalArgumentException("map type field [" + sourceKey + "] reached max depth limit, cannot process it");
+        }
+    }
+}
--- a/src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
+++ b/src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
@@ -35,7 +35,8 @@ public class ProcessorDocumentUtils {
         final Map<String, Object> sourceValue,
         final Object fieldMap,
         final int depth,
-        final long maxDepth
+        final long maxDepth,
+        final boolean allowEmpty
     ) {
         if (sourceValue == null) return; // allow map type value to be null.
         validateDepth(sourceKey, depth, maxDepth);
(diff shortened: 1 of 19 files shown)


src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
```diff
@@ -0,0 +1,115 @@
...
+                    ((Map) fieldMap).get(sourceKey),
+                    depth + 1,
+                    maxDepth,
+                    allowEmpty
+                );
+            } else if (!(firstNonNullElement instanceof String)) {
+                throw new IllegalArgumentException("list type field [" + sourceKey + "] has non string value, cannot process it");
+            } else {
+                if (element == null) {
+                    throw new IllegalArgumentException("list type field [" + sourceKey + "] has null, cannot process it");
```
chishui, line 100: when you check `if (firstNonNullElement instanceof List)` and `else if (firstNonNullElement instanceof Map)`, `element` could also be null, but no exception thrown, the logic seems inconsistent. Why we check both firstNonNullElement and element, it's a bit confusing

src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
```diff
@@ -0,0 +1,115 @@
...
+import org.opensearch.env.Environment;
+import org.opensearch.index.mapper.IndexFieldMapper;
+import org.opensearch.index.mapper.MapperService;
+
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Objects;
+
+public class ProcessorDocumentUtils {
```
chishui, line 20: Can we add UT for this class?

chishui (commented):
Thanks for the work to refactor and reduce duplicate code

src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
zane-neo: Make sense, I'll change this part to follow a consistent principle: for list type, we don't allow null element in it.

yuye-aws:
Happy to see that this PR extracts all validation function into file `src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java`? Currently, all the functions except `getMaxDepth` is about validation, can we simply rename this file to `DocumentValidator`?

yuye-aws:
We still have parsing methods like `buildMapWithTargetKeyAndOriginalValue` and `createInferenceList`. Can we also extract them into another class named "DocumentParser"?

src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorTests.java
```diff
@@ -630,15 +649,18 @@ public void testExecute_withFixedTokenLength_andFieldMapNestedMap_sourceDataList
...
-        assert (listResult instanceof List);
-        assertEquals(((List<?>) listResult).size(), 0);
+        IllegalArgumentException illegalArgumentException = assertThrows(
+            IllegalArgumentException.class,
+            () -> processor.execute(ingestDocument)
+        );
+        assertEquals(
+            "[body] configuration doesn't match actual value type, configuration type is: java.lang.String, actual value type is: com.google.common.collect.RegularImmutableMap",
+            illegalArgumentException.getMessage()
+        );
```
yuye-aws, line 663: String.fromat()

src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorTests.java
```diff
@@ -630,15 +649,18 @@ public void testExecute_withFixedTokenLength_andFieldMapNestedMap_sourceDataList
     }
 
     @SneakyThrows
-    public void testExecute_withFixedTokenLength_andSourceDataListWithHybridType_thenSucceed() {
+    public void testExecute_withFixedTokenLength_andSourceDataListWithHybridType_thenFail() {
```
yuye-aws, line 652: Why this test case should fail?

src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
zane-neo: In fact, these code are extracted from old code and there're already a lot of tests to cover these code, from the code coverage ran from local, all lines are covered already.

src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorTests.java
```diff
@@ -600,7 +619,7 @@ public void testExecute_withFixedTokenLength_andFieldMapNestedMap_thenFail() {
             () -> processor.execute(ingestDocument)
         );
         assertEquals(
-            String.format(Locale.ROOT, "map type field [%s] has non-string type, cannot process it", INPUT_NESTED_FIELD_KEY),
+            "[body] configuration doesn't match actual value type, configuration type is: java.lang.String, actual value type is: java.util.ImmutableCollections$Map1",
```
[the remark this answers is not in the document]
zane-neo: The error message is deterministic not dynamic, I don't think we need to use string.format. Also, use plain text is a little bit more straightforward than string.format.

src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorTests.java
zane-neo: Current code we didn't have bi-directional validation of document source and configuration map. A map type configuration can't fit both string and map type value, e.g. 
```
{
  "field_map": {
    "a": {
      "b": {
        "c": "c_target"
      }
    }
  }
}
```
Can not fit to both(case 1)
```
{
  "a": {
    "b": {
      "c": "text to embedding/chunk"
    }
  }
}
```
and(case 2)
{
  "a": {
    "b": "text to embedding/chunk"
  }
}
```
For case 2, text_embedding and text_image_embedding processor needs to build a temporary map with target key, and `b`'s type will be treated as map based on configuration and then cause class cast exception. 
For text chunking, current code doesn't throw exception but the result is not correct, based on user's configuration, `c` is the target chunking field but the actual chunking field in this case is `b`, and the generated result is:
```
{
  "c_target": "b's chunking result"
}
```
Which is unexpected.

So now we enforced the bi-directional to avoid this case, and the base principle we'll follow in the future is: list type always have same type/data structure object in it.

File: src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorTests.java (hunk 3 above)
Comment (zane-neo) [reply src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorTests.java:663]: Same above.

File: src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java [hunk 6]
```diff
@@ -0,0 +1,115 @@
...
+
+    public static long getMaxDepth(Map<String, Object> sourceAndMetadataMap, ClusterService clusterService, Environment environment) {
```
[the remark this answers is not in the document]
zane-neo: Make sense, I was trying to reduce the method's parameter so split them to different methods, but making it a private method is the correct direction.

src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+        return MapperService.INDEX_MAPPING_DEPTH_LIMIT_SETTING.get(environment.settings());
+    }
+
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    public static void validateMapTypeValue(
+        final String sourceKey,
+        final Map<String, Object> sourceValue,
+        final Object fieldMap,
+        final int depth,
```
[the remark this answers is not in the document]
zane-neo: long is from OpenSearch Core(I don't think it make sense BTW), but I've changed the int to long to fit it.

zane-neo:
> Happy to see that this PR extracts all validation function into file `src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java`? Currently, all the functions except `getMaxDepth` is about validation, can we simply rename this file to `DocumentValidator`?

No, this file will be adding more and more common methods like parsing data structure etc, so will name as what it is now.

zane-neo:
> We still have parsing methods like `buildMapWithTargetKeyAndOriginalValue` and `createInferenceList`. Can we also extract them into another class named "DocumentParser"?

We will supports more complex cases in the future and do this extraction that time, we'll put these common code in the new class added this time instead of creating another class.

yuye-aws:
> > We still have parsing methods like `buildMapWithTargetKeyAndOriginalValue` and `createInferenceList`. Can we also extract them into another class named "DocumentParser"?
> 
> We will supports more complex cases in the future and do this extraction that time, we'll put these common code in the new class added this time instead of creating another class.

Makes sense.

src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorTests.java
yuye-aws: Agree with `list type always have same type/data structure object in it`. However, it maybe a breaking change for not supporting hybrid list. Users previously pass in hybrid input in a list will now receive an error. Can we avoid throwing an error the user?

zane-neo: Breaking change means an expected behavior changed, like above mentioned, this in fact is a bug since the result is not what user expected, by adding this validation user can realize the issue and fix their inputs.

zane-neo:
The BWC test failure is not caused by this change and there's already an issue tracking it: https://github.com/opensearch-project/neural-search/issues/688

vibrantvarun:
Why BWC tests are failing in the build with the same issue?

yuye-aws:
This PR looks good to me

vibrantvarun:
Why are the BWC tests failing @yuye-aws ?

yuye-aws:
Hi @vibrantvarun ! The BWC tests not get passed. What's our next step plan to merge this PR?

src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
```diff
@@ -0,0 +1,151 @@
...
+     * @param  environment   environment passed from OpenSearch core.
+     * @param  indexName     the maximum allowed depth for recursion
+     * @param  allowEmpty   flag to allow empty values in map type validation.
+     */
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    public static void validateMapTypeValue(
+        final String sourceKey,
+        final Map<String, Object> sourceValue,
+        final Object fieldMap,
+        final long depth,
```
martin-gaievski, line 44: looks like in business logic code we're always calling this method with `depth = 1`. Better encapsulate this parameter by making this method private, and create one more public method without `depth` param. This new method will call the private method and pass `depth` as `1`.

zhichao-aws: +1

src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
```diff
@@ -0,0 +1,151 @@
...
+        final String indexName,
+        final ClusterService clusterService,
+        final Environment environment,
```
zhichao-aws, line 47: if clusterService and environment are only used here for those processors; how about make the 2 variables static field for ProcessorDocumentUtils? Otherwise for every new processor we need to create 2 fields for them

src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
zane-neo: done

src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
zane-neo: It's true that we can do this to reduce code repetition, but the current repetition isn't significant, and we're adding new processors at a very slow pace. It's fine to keep things as they are now, we can optimize when more processors are added.

zhichao-aws (approved):

src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
```diff
@@ -0,0 +1,162 @@
...
+                        (Map<String, Object>) nextSourceValue,
+                        nextFieldMap,
+                        depth + 1,
+                        indexName,
+                        clusterService,
+                        environment,
+                        allowEmpty
+                    );
+                } else if (!(nextSourceValue instanceof String)) {
+                    throw new IllegalArgumentException("map type field [" + key + "] is neither string nor nested type, cannot process it");
```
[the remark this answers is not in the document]
zane-neo: Done

src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
```diff
@@ -0,0 +1,162 @@
...
+                } else if (!allowEmpty && StringUtils.isBlank((String) nextSourceValue)) {
+                    throw new IllegalArgumentException("map type field [" + key + "] has empty string value, cannot process it");
```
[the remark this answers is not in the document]
zane-neo: Done

src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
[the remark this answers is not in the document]
martin-gaievski: Why default() locale? We typically use Locale.ROOT, it refers to empty location rather then picking up locale of OS JVM process on data node.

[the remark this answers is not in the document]
zane-neo: Done

martin-gaievski (approved):

Issue #608, opened 2024-02-21
[FEATURE] Support for raw sparse vectors input in the neural sparse query 


brusic:
### Is your feature request related to a problem?

[Neural sparse search](https://opensearch.org/docs/latest/search-plugins/neural-sparse-search/)

Currently the neural search query only accepts the model id alongside the text to be encoded, which requires a model to be registered into a pipeline. The query should also support passing in the vector directly, bypassing the pipeline phase. It can be beneficial for clients to do the encoding for several reasons: ad hoc analysis, unit testing, custom/unsupported models.

### What solution would you like?
Accept a vector, similar to [knn search](https://opensearch.org/docs/latest/search-plugins/knn/index/)
```
GET /my-nlp-index/_search
{
  "_source": {
    "excludes": [
      "passage_embedding"
    ]
  },
  "query": {
    "neural_sparse": {
      "passage_embedding": {
        "vector": ['a':2, 'b':3, 'c':5, 'd':6],        
        "k": 5
      }
    }
  }
}


```

### What alternatives have you considered?
`rank_features` is a close alternative, but can only rank (boost) other query clauses.

### Do you have any additional context?

ES will soon have a [weighted_tokens query](https://github.com/elastic/elasticsearch/blob/main/docs/reference/query-dsl/weighted-tokens-query.asciidoc), which is analogous to their [text_expansion query](https://github.com/elastic/elasticsearch/blob/main/docs/reference/query-dsl/text-expansion-query.asciidoc).

zhichao-aws:
Do you mean the `neural_sparse` query clause? Seems you're using query tokens and weights to search, and paste a link about sparse search. But `neural` query is using dense models

brusic:
Correct, sorry for the confusion. Used the wrong query in my example, probably due to never having used the `neural_sparse` query. Updated the example and add a link to the sparse search.

model-collapse:
@brusic I changed the title into an accurate one.

zhichao-aws:
Hi @brusic , our enhancements has been merged now and will be released at 2.14 version. Now users can just use neural sparse query with raw tokens. Sample query:
```
"neural_sparse": {
  "<vector_field>": {
    "query_tokens": {
       "token 1": 1.1,
       "token 2": 2.2,
       "token 3": 3.3
    }
  }
}
```

zhichao-aws:
Close this issue as we have finished the feature. Feel free to re-open it if there is more discussion

PR #693, opened 2024-04-16
enhancements: support neural_sparse query by tokens

zhichao-aws:
### Description
support raw tokens query in neural_sparse search

### Issues Resolved
https://github.com/opensearch-project/neural-search/issues/608

### Check List
    - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
@@ -66,19 +66,16 @@ public class NeuralSparseQueryBuilder extends AbstractQueryBuilder<NeuralSparseQ
     static final ParseField QUERY_TOKENS_FIELD = new ParseField("query_tokens");
     @VisibleForTesting
     static final ParseField MODEL_ID_FIELD = new ParseField("model_id");
-    private static final String EMPTY_STRING = "";
-
     private static MLCommonsClientAccessor ML_CLIENT;
-
-    public static void initialize(MLCommonsClientAccessor mlClient) {
-        NeuralSparseQueryBuilder.ML_CLIENT = mlClient;
-    }
-
     private String fieldName;
     private String queryText;
     private String modelId;
     private Supplier<Map<String, Float>> queryTokensSupplier;
     private static final Version MINIMAL_SUPPORTED_VERSION_DEFAULT_MODEL_ID = Version.V_2_13_0;
+
+    public static void initialize(MLCommonsClientAccessor mlClient) {
+        NeuralSparseQueryBuilder.ML_CLIENT = mlClient;
+    }
 
     /**
      * Constructor from stream input
@@ -101,28 +98,28 @@ public class NeuralSparseQueryBuilder extends AbstractQueryBuilder<NeuralSparseQ
         }
         // to be backward compatible with previous version, we need to use writeString/readString API instead of optionalString API
         // after supporting query by tokens, queryText and modelId can be null. here we write an empty String instead
-        if (EMPTY_STRING.equals(queryText)) {
+        if (StringUtils.EMPTY.equals(this.queryText)) {
             this.queryText = null;
         }
-        if (EMPTY_STRING.equals(modelId)) {
+        if (StringUtils.EMPTY.equals(this.modelId)) {
             this.modelId = null;
         }
     }
 
     @Override
     protected void doWriteTo(StreamOutput out) throws IOException {
-        out.writeString(fieldName);
+        out.writeString(this.fieldName);
         // to be backward compatible with previous version, we need to use writeString/readString API instead of optionalString API
         // after supporting query by tokens, queryText and modelId can be null. here we write an empty String instead
-        out.writeString(Objects.isNull(queryText) ? EMPTY_STRING : queryText);
+        out.writeString(StringUtils.defaultString(this.queryText, StringUtils.EMPTY));
         if (isClusterOnOrAfterMinReqVersionForDefaultModelIdSupport()) {
-            out.writeOptionalString(modelId);
+            out.writeOptionalString(this.modelId);
         } else {
-            out.writeString(Objects.isNull(modelId) ? EMPTY_STRING : modelId);
+            out.writeString(StringUtils.defaultString(this.modelId, StringUtils.EMPTY));
         }
-        if (!Objects.isNull(queryTokensSupplier) && !Objects.isNull(queryTokensSupplier.get())) {
+        if (!Objects.isNull(this.queryTokensSupplier) && !Objects.isNull(this.queryTokensSupplier.get())) {
             out.writeBoolean(true);
-            out.writeMap(queryTokensSupplier.get(), StreamOutput::writeString, StreamOutput::writeFloat);
+            out.writeMap(this.queryTokensSupplier.get(), StreamOutput::writeString, StreamOutput::writeFloat);
         } else {
             out.writeBoolean(false);
         }
@@ -213,6 +210,15 @@ public class NeuralSparseQueryBuilder extends AbstractQueryBuilder<NeuralSparseQ
                     )
                 );
             }
+        }
+
+        if (Objects.nonNull(sparseEncodingQueryBuilder.queryText) && StringUtils.EMPTY.equals(sparseEncodingQueryBuilder.queryText)) {
+            throw new IllegalArgumentException(
+                String.format(Locale.ROOT, "%s field can not be empty", QUERY_TEXT_FIELD.getPreferredName())
+            );
+        }
+        if (Objects.nonNull(sparseEncodingQueryBuilder.modelId) && StringUtils.EMPTY.equals(sparseEncodingQueryBuilder.modelId)) {
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "%s field can not be empty", MODEL_ID_FIELD.getPreferredName()));
         }
 
         return sparseEncodingQueryBuilder;
--- a/src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
@@ -186,8 +186,9 @@ public class NeuralSparseQueryBuilder extends AbstractQueryBuilder<NeuralSparseQ
                 )
             );
         }
+
+        requireValue(sparseEncodingQueryBuilder.fieldName(), "Field name must be provided for " + NAME + " query");
         if (Objects.isNull(sparseEncodingQueryBuilder.queryTokensSupplier)) {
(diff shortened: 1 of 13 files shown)


src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -95,16 +99,26 @@ public NeuralSparseQueryBuilder(StreamInput in) throws IOException {
             Map<String, Float> queryTokens = in.readMap(StreamInput::readString, StreamInput::readFloat);
             this.queryTokensSupplier = () -> queryTokens;
         }
+        // to be backward compatible with previous version, we need to use writeString/readString API instead of optionalString API
+        // after supporting query by tokens, queryText and modelId can be null. here we write an empty String instead
+        if (EMPTY_STRING.equals(queryText)) {
```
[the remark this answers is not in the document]
chishui: Can we treat null and "" the same? If so, we don't need to assign it to null and only need to change checking logic below

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -95,16 +99,26 @@ public NeuralSparseQueryBuilder(StreamInput in) throws IOException {
...
+        }
     }
 
     @Override
     protected void doWriteTo(StreamOutput out) throws IOException {
         out.writeString(fieldName);
-        out.writeString(queryText);
+        // to be backward compatible with previous version, we need to use writeString/readString API instead of optionalString API
+        // after supporting query by tokens, queryText and modelId can be null. here we write an empty String instead
+        out.writeString(Objects.isNull(queryText) ? EMPTY_STRING : queryText);
```
chishui, line 117: `StringUtils.defaultString` can replace ternary operator

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -118,10 +132,15 @@ protected void doWriteTo(StreamOutput out) throws IOException {
     protected void doXContent(XContentBuilder xContentBuilder, Params params) throws IOException {
         xContentBuilder.startObject(NAME);
         xContentBuilder.startObject(fieldName);
-        xContentBuilder.field(QUERY_TEXT_FIELD.getPreferredName(), queryText);
+        if (Objects.nonNull(queryText)) {
```
chishui, line 135: "" is acceptable here?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -118,10 +132,15 @@ protected void doWriteTo(StreamOutput out) throws IOException {
...
+            xContentBuilder.field(QUERY_TEXT_FIELD.getPreferredName(), queryText);
+        }
         if (Objects.nonNull(modelId)) {
             xContentBuilder.field(MODEL_ID_FIELD.getPreferredName(), modelId);
         }
+        if (Objects.nonNull(queryTokensSupplier) && Objects.nonNull(queryTokensSupplier.get())) {
```
chishui, line 141: just curious, why we need supplier for queryTokens?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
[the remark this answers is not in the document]
chishui: could you help move `initialize` function below variable definition?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -62,7 +63,10 @@ public class NeuralSparseQueryBuilder extends AbstractQueryBuilder<NeuralSparseQ
     @VisibleForTesting
     static final ParseField QUERY_TEXT_FIELD = new ParseField("query_text");
     @VisibleForTesting
+    static final ParseField QUERY_TOKENS_FIELD = new ParseField("query_tokens");
+    @VisibleForTesting
     static final ParseField MODEL_ID_FIELD = new ParseField("model_id");
+    private static final String EMPTY_STRING = "";
```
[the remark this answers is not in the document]
zhichao-aws: ack

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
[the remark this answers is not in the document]
zhichao-aws: > Can we treat null and "" the same? If so, we don't need to assign it to null and only need to change checking logic below
In the core logics (reWrite part) they are treated same. Both null and empty will throw exception. But in some detailed parts they are treated different. like doXContent, if value is null we don't include the field; but if it's empty, we still put a field there with empty string value. So strictly we should transform it back to null

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: yes

zhichao-aws: Although it's an invalid query to use an empty string and we'll throw exception. Here we accept empty string

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: It's due to the async model inference in doRewrite

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: will add check logics in `fromXContent` to make sure the queryText and modelId is nonNull from user

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: The doRewrite will be called multiple times. At the first time we register the async model inference and set the supplier non-null (although the model inference is not finished). Before the async inference end, when doRewrite is called again, we know we already have a registered model inference and will skip it.
Otherwise we'll register multiple inference for one query

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -161,16 +187,40 @@ public static NeuralSparseQueryBuilder fromXContent(XContentParser parser) throw
...
+                        "using %s, %s field must be provided for [%s] query",
+                        QUERY_TEXT_FIELD.getPreferredName(),
+                        MODEL_ID_FIELD.getPreferredName(),
+                        NAME
+                    )
+                );
+            }
+        }
+
+        if (Objects.nonNull(sparseEncodingQueryBuilder.queryText) && StringUtils.EMPTY.equals(sparseEncodingQueryBuilder.queryText)) {
```
chishui, line 215: 1. nit: use getter instead of reference to member variable.
2. If we only want to check sparseEncodingQueryBuilder.queryText() is empty string, `StringUtils.EMPTY.equals` is enough, if we want to check null or "", then we should use `||` instead

chishui (commented):
Overall, LGTM

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -95,20 +96,30 @@ public NeuralSparseQueryBuilder(StreamInput in) throws IOException {
             Map<String, Float> queryTokens = in.readMap(StreamInput::readString, StreamInput::readFloat);
             this.queryTokensSupplier = () -> queryTokens;
         }
+        // to be backward compatible with previous version, we need to use writeString/readString API instead of optionalString API
+        // after supporting query by tokens, queryText and modelId can be null. here we write an empty String instead
+        if (StringUtils.EMPTY.equals(this.queryText)) {
```
zane-neo, line 101: Not getting it why can't using readOptionalString? I didn't see it breaks BWC, can you elaborate more?

zhichao-aws: readString will directly readString from byte stream, while readOptionalString will first read a boolean from the byte Stream, then read a String.
Imagine we have 2 node, one is 2.14, the other is 2.13. In 2.13 we call writeString to stream, and in 2.14 we use readOptionalString to read it. This will break the serialization and deserialization process, making the behavior unpredictable. 

> I didn't see it breaks BWC
Because now we use a BWC-safe way to write/read it. Furthermore, an un-safe way will make bwc un-predictable, make it flaky test. But we can not make sure every API call will throw exception

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: ack

zane-neo (approved):

chishui (approved):

model-collapse (approved):

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -102,21 +103,31 @@ public NeuralSparseQueryBuilder(StreamInput in) throws IOException {
...
+            this.queryText = null;
+        }
+        if (StringUtils.EMPTY.equals(this.modelId)) {
```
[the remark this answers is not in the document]
zhichao-aws: In previous release we'll throw exceptions at doRewrite. Now we throw the exception when parsing them

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -102,21 +103,31 @@ public NeuralSparseQueryBuilder(StreamInput in) throws IOException {
...
     }
 
     @Override
     protected void doWriteTo(StreamOutput out) throws IOException {
-        out.writeString(fieldName);
-        out.writeString(queryText);
+        out.writeString(this.fieldName);
+        // to be backward compatible with previous version, we need to use writeString/readString API instead of optionalString API
+        // after supporting query by tokens, queryText and modelId can be null. here we write an empty String instead
+        out.writeString(StringUtils.defaultString(this.queryText, StringUtils.EMPTY));
```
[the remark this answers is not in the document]
zhichao-aws: Yes

PR #695, opened 2024-04-17
[FEATURE] Add rescoreContext for neuralsparse query's two-phase speed up.

### LINKED ISSUE #646: [RFC] Enhancing  Neural Sparse  Query Speed with a Two-Phase Approach
# Overview

**OpenSearch Neural Search** is an OpenSearch plugin that adds semantic retrieval into the OpenSearch ecosystem. The plugin provides the capability for indexing documents and doing neural search on the indexed documents. 

In this plugin, **Neural Sparse query** is a semantic search API in which the storage is implemented by native Lucene inverted index. It reduces the query costs, and improves the search relevance of query. Compared with BM25, neural sparse search expands the tokens in the document, has larger inverted indexes. And bi-encoder mode will significantly expand the query tokens, results in longer retrieval time. 

We investigated that two-phase **Rescore Search** can speed up retrieval by splitting a query into two steps. According to our experiments, two-phase rescore search can speeds up the search at most 9.8 times with negligible impacts on search relevance.

# Description
## Problem Statement

Neural sparse search is a semantic search method supported by Opensearch, which inference a series of tokens containing relevance scores from text and query statements respectively, and then performs relevance operations to get the most semantically relevant documents to the query text.

It tends to compute a large number of low-scoring tokens that do not too much contribute to the search results, but these low-scoring tokens can slow down the search. For example, Assuming the text we're querying is "what's the weather in NYC now?", then some tokens will be generated. Among them, "weather" and "NYC" are tokens with higher scores and more importance, while "what", "is", "the", "in", etc., are tokens with lower scores.

## Two Phase Search

Two Phase  search refers to the process of re-evaluating and re-ranking search results after initial ranking. This is often used to enhance the relevance of search results. In the first phase we use high score tokens to retrieve N top docs, where N is larger than the original search size M. In the second phase we use low score tokens to rescore these N top docs, and returns the top M docs after rescore.

## tokens

In neural sparse search bi-encoder mode, query text will be encoded to tokens and corresponding weights, which are semantically close to the original text. But the number of expanded tokens is large, so it increases the searching time cost. 

The principle of the neural sparse query is to match all generated (token, score) pairs with the inverted index of documents. Therefore, each additional token in query makes the searcher go through one more inverted index.

At the same time, we find that the model generates a large number of low-score tokens that contribute less to the search score. And the low-score tokens has large overlap with tokens considered having low IDF score. It is mainly the higher-scored tokens that have the primary impact in the query.

### high score tokens

Using a portion of the high-scoring tokens obtained from the model computation will make the search faster.
high score tokens means token score is larger than rescore_ratio*max_token_score_in_query.

### low score tokens

Low-scoring tokens are used for a second phase scoring after obtaining a relatively small number of TopDocs.

### token and query

Now the neural sparse query use all tokens to query, we tend to split this to two phase - high score tokens and low score tokens. Based on neural sparse score algorithm, the score of neural sparse equals to the sum of **highscoretokenquery** and **lowscoretokenquery**.

## target workflow

prog_rescore drawio

Initially, we generate all tokens derived from the query text, then select the High-scoring tokens for the Initial Query. Subsequently, we utilize low-scoring tokens within the TopDocs of  initial query to conduct the second rescore stage. Then we sum up the scores of two phases and return.

## Feasibility Experiment

In our previous investigation, with a rescore factor of 0.4, the search relevance loss is less than 0.04%, but can speeds up 9.8 times at most on BEIR.

The following figure shows the acceleration effect under doc-only and bi-encoder. The speed up number is using P90_latency@rescore_ratio=0 divide P90_latency@rescore_ratio=x.
- bi-encoder speed up

diyige

- doc-only speed up

doconly

# Use Case

## Top Level Query

In this scenario, we only need to consider processing a top-level `NeuralSparseQuery` without discussing its combination with other queries for compression or its application after being optimized.

```
  "query": {
    "neural_sparse": {
      "passage_embedding": {
        "query_text": "Hi world",
        "model_id": <model_id>,
      }
  }
```

### Wrapped Neural Sparse Query in Boolean Query

In this scenario, the `NeuralSparseQuery` will be optimized into a clause of a Boolean query. In such cases, because of fine-tune by opensearch core and lucene, we need to manually split the Boolean Query before we can determine if there is a query clause originating from NeuralSpare inside.

```
{
  "query": {
    "bool": {
      "should": [
        {
          "neural_sparse": {
            "passage_embedding": {
              "query_text": "Hi world",
              "model_id": "<model_id>",
              "boost": 1.2
            }
          }
          "match":{
                "text": "Hi"
           }
        }
      ],
      "must": [
        {
          "match": {
            "text": "world"
          }
        }
      ],
    "boost": 1.2
    }
  }
}
```

# Two Phase Design Options

## Current WorkFlow of the QueryPhase

curworkflow

The diagram below is the sequence diagram of the current system. As shown in the diagram, when a Query arrives at the Search Service, it is sent to the corresponding QueryBuilder for construction. After successful construction, it is processed and executed by the corresponding searcher , and the result is finally returned.

In short, what we need to do is to preserve the ability to query using different tokens in NeuralSparse Query as well as applying those lower scoring tokens in TopDocs.

Thus, we have several possible solutions to choose from: 

1. Implement a special Custom Query and change its scoring function logic, allowing it to selectively use tokens during the Scorer's scoring process. 
2. Change the search logic by implementing our own QueryPhase, dividing the original search logic into two stages for execution. 

### Custom Query in Lucene

In Apache Lucene, the processing flow of a custom query involves several key components, including the Query Builder, Weight, Scorer, and possibly the PhaseSearcher. And we can make our own implementation of them all.

#### Weight

When IndexSearcher is ready to execute a query, it calls the createWeight method of the Query object to create a corresponding Weight object. This Weight object is the basis for calculating document relevance scores and other search-related operations during the search process.

#### Scorer

Scorer is a core component used to calculate the relevance score of a document relative to a query.The Scorer object is created for each index segment (LeafReaderContext) by the Weight object. It is the basis for the implementation of the Lucene scoring mechanism, which is designed to assign each document that matches a query a score that reflects how well the document matches the query criteria.

### Developer-implemented specific Query

As previously mentioned, the scoring of each document is conducted through the score function in Score. Therefore, we focus on changing the logic of the score. Each Neural Sparse query generates a Neural Sparse Weight, and during the query process, the scoring of documents mainly relies on the score function in the Scorer object within the Neural Sparse Weight.

Therefore, we consider changing the data structure of the Weight to carry a priority queue of the current TopDocs' scores. For documents with scores lower than the minimum value in this priority queue, it can be proven that they will inevitably not appear in the final TopDocs. Thus, we only apply the faster HighScoreTokenQuery to them. For documents with higher scores, we update the priority queue and apply scoring with all tokens to these documents.

### Trade Off

Pros

1. It can support almost any query type.
2. It does not bring in much coupling in the Neural Search Plugin.

Cons

1. It's performance gains won't be as noticeable as expected, the reason is that we need to maintain a priority queue in the search process, which is based on a massive number of documents, and its overhead is not negligible; moreover, the actual reduction in computation does not reach as much as ideal, especially in the first few hundred calculations, where the priority queue is frequently updated and often needs to perform an AllTokenQuery.
2. On top of combining multiple query scoring weights, its correctness decreases.

### Target WorkFlow

currentstru

As target workflow illustrated, we introduce the NeuralSparseQueryPhaseSearcher module to replace the DefaultQueryPhaseSearcher, allowing us to adjust the TopDocs scores through a two-phase search after executing the first search. Simultaneously, for nested queries, the NestedQueryParser is used to extract the correct two-stage Query.

### Description

This module is tasked with interpreting incoming queries and executing targeted queries. Within this module, we will implement the specific query logic for the NeuralSparseQuery.

In short, if we implement a  NeuralSparseQueryPhaseSearcher which implemented QueryPhaseSearcher and register it to OpenSearch through the getQueryPhaseSearcher method, the search logic which in DefaultQueryPhaseSearcher can be override to NeuralSparseQueryPhaseSearcher.searchWith().

After receiving a NestedQuery containing NeuralSparseQuery, our searchWith logic will proceed in the following steps:

1. Transform all NeuralSparse query objects into queries that only contain High-Score Tokens .Return the simplified NestedQuery as the first phase query.
2. For each NeuralSparse object, extract the queries containing Low-Score Tokens and record the corresponding weights. These queries are combined into an equivalent query, serving as the second phase query.“
3. Execute the first query to score all documents and obtain TopDocs.
4. For each document in TopDocs, modify its score to be the existing score plus the score from the LowTokenQuery.

### Nested Query Parse Module

It is hard to capture the NeuralSparseQuery after embedding through a custom PhaseSearcher, because the rewritten top-level Query contains all optimized query statements, when neuralsparsequery may be only a part. However, we can apply specific logic to split the Compound queries.

#### Description

In a single NeuralSparseQuery, our goal is to extract Low-score Tokens for the second-stage query evaluation. However, in composite queries, considering the impact of weights and conditions, we need to determine the extraction method based on the type of query. We plan to go live and support two types of composite queries, BoostQuery and BooleanQuery, in version 2.14. Next, I will introduce how to extract the correct second-stage queries from them.

##### Boost Query

BoostQuery means that each query will have different weights. When extracting Low-Score Tokens from the NeuralSparseQuery nested within, we will record their corresponding weights.

##### For Different Boolean Clause

Boolean query have multiple clauses: MUST, MUST NOT, FILTER and SHOULD, these clauses have different scoring rules and, therefore, require different processing strategies.

- Must
Documents must fulfill all the conditions of the MUST clause to be considered as a match for the query and MUST clauses contribute to the score, which means they affect the final score of the document.
This means that we cannot simply disassemble the NeuralSparse clause in the Must statement, as it may result in a small portion of the document being lost on the result.

- Should
If a MUST clause is present, a document is considered a match only if it satisfies all MUST clauses. Also, matches of SHOULD clauses increase the score of the document. But they are not necessary to determine document unless there are no MUST clauses and minShouldMatch is set a upright value.
This means that we can disassemble part of the NeuralSparse in Should when there is a must clause of the same level as should or when minShouldMatch is zero.

- Must not
Any document that satisfies the MUST_NOT clause will not be included in the query results, so I think it's safe to leave the NeuralSparse clause out of it for now.

- Filter
FILTER clauses are similar to MUST clauses in that documents must fulfill the conditions of these clauses in order to be considered as a match.
Also, unlike MUST clauses, FILTER clauses do not contribute to the score; they are only used to include or exclude documents without affecting the score.
Therefore, I think Two-Phase should not be optimized in FILTER clauses.

#### Example

Assume that we build a query from this structure.

```yaml
query:
  - booleanQueryA(1.3): 
      should:
        - neural_sparse(1.4)
        - booleanQueryB(1.1):
            must:
              - term(1.0):
            should:
              - term(1.5):
```

The query score will be below.

```
Score(booleanQueryA) = 1.3 * { 1.4 * score(neural_sparse) }+ 1.1*(booleanQueryB)
```

The query after this module’s parse and change wii be like follow:

```yaml
new_nested_query:
  - booleanQueryA(1.3): 
      should:
        - highTokenScoreQuery(1.4)
        - booleanQueryB(1.1):
            must:
              - term(1.0):
            should:
              - term(1.5):
```

And We will get a lowScoreTokenQuery and whose weight equals 1.3*1.4 = 1.82. And use this query to get a score to add to the first phase‘s result. Now the TopDocs’s score will be 1.82*lowScoreTokenQuery + firstQueryScore that equals for the initial query.

### Trade off

Pros:

1. Support Top-Level and Nested Boolean\Boost Query.
2. High performance.

Cons:

1. Doesn’t support all compound query. (But support can be provided in future)
2. Need more time to development.

### SearchPipeline
SearchRequestProcessor is a SearchPipeline can  take the SearchRequest before the rewrite and build.
From the SearchRequest, we can get the 1. whole query builder 2. rescores.

Based on that, we can get the NeuralSparseQueryBuilder object as the originNeuralSparseQueryBuilder,
We make a twoPhaseRescoreQueryBuilder which is copied from originNeuralSparseQueryBuilder.
And set the twoPhaseRescoreQueryBuilder as a member variable of originNeuralSparseQueryBuilder.
Finally, add the twoPhaseRescoreQueryBuilder to the searchRequest.

### Resolve the model  inference

The NeuralSpareQueryBuilder get the queryTokens in the rewrite, and each searchRequest will be rewrite, to avoid repeat model inference time cost, add a flag to perform different logic in the originNeuralSparseQueryBuilder and twoPhaseNeuralSparseQueryBuilder.

For originNeuralSparseQueryBuilder:
    The originNeuralSparseQueryBuilder use register a model inference as a asyncAction to get queryToken and their score.

For twoPhaseQueryBuilder:
    The twoPhaseQueryBuilder just wait the originNeuralSparseQueryBuilder to set.
### FlowChart
seachpipeline

### Parameter

SearchPipeLine can carry the parameter.
For different index, user can use different settings by search pipeline.
```json
{
  "request_processors": [
    {
      "neural_sparse_two_phase_processor" : {
        "tag":'tag1',
        "description": '',
        "two_phase_ratio": 0.4,
        "two_phase_size_expandation":5.0,
        "two_phase_window_size": 1000
      }
    }
  ]
}
```
### Trade off
Pros
- No impact on other use cases. We only go through queries for indexes using 2-phase search processor. Also less risky.
- We can set the parameters in search processor parameter, no need to config index settings
- No need to migrate from short-term to long-term, so it doesn’t have BWC issues.
Cons
- Needs more effort for users to setup.
## Poc code

https://github.com/opensearch-project/neural-search/compare/main...conggguan:neural-search:search-pipeline

# API Design

## setting API

To use rescore for each time in neural sparse query, 

```
// enable the rescore
PUT my_index
{
    "plugins.neural_search.neural_sparse_two_phase_settings":{
        "should_two_phase":true,(default false)
     }
}
// disable the rescore
PUT my_index
{
    "plugins.neural_search.neural_sparse_two_phase_settings":{
        "should_two_phase":false
     }
}
```

## search API

**One-off  query API**
In each search of neural sparse index, user can do a single rescore query by specify the rescore_parameter.
In this example, the query will retrieve for Top 50 docs by important tokens first and get the 50*0.2=10 docs by all tokens next.

```
# Search with index
GET my_index/_search
{
  "query": {
    "neural_sparse": {
      "passage_embedding": {
        "query_text" : "Hi world",
        "model_id" : <model_id>,
        "two_phase_parameter" : {
          "two_phase_ratio" : 0.2,
          "two_phase_window_size" : 50
       }
      }
    }
  }
}
```

**Standard query API**
After enabling Two-Phase search acceleration through Settings API, Two-Phase acceleration will be activated in the default search without any other parameters.

```
# Search with index
GET my_index/_search
{
  "query": {
    "neural_sparse": {
      "passage_embedding": {
        "query_text" : "Hi world",
        "model_id" : <model_id>
      }
  }
}
```
# BenchMark
## The benchmark Env
-opensearch cluster
	- node_number: 3 
	- ec2_instance_type: m5.4xlarge
- benchmark tool
	- opensearch benchmark
- benchmark config
	- client number: 20
	- warmup-iterations: 50
	- iterations: 200
### The doc-only P90 table
doc-only
### The bi-encoder P90 table
bi-encoder

conggguan:
### Description
This change implement for #646 
- Enhance the speed of neuralsparse query by two-phase.
- Now support top-level, boolean and boost compound query; for other compound query will degrade into origin logic and speed.

### Feature support query
- [x] NeuralSparseQuery
- [x] NeuralSparseQuery nested in BoostQuery
- [x] NeuralSparseQuery nested in BooleanQuery
#### Intro
Now implement two-phase feature by adding rescoreContext in the begin of HybridQueryPhaseSearcher's. It is harmless but little hack, so there is a long term plan to replace this implementation.
#### 
#### RFC in opensearch core
To add a correct rescoreContext to searchContext, we need both searchContext and whole compound query, but now if we can only get these together in the HybridQueryPhaseSearcher. In QueryBuilder, the searchContext is private, and whole query are not visible.
What we need to implement this feature is the whole query and searchContext, so if OpenSearch Core can open a API which can help me get it before the query performed, we can move the rescoreContext function to a more general interface of core.
#### RoadMap of this optimizing
1. [RFC in OpenSearch Core](https://github.com/opensearch-project/OpenSearch/issues/13320). 
2. A Incremental update in NeuralSearch plugin.

### Issues Resolved
Resolve #646 

### Check List
    - [ ] All tests pass


src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQuery.java
```diff
@@ -0,0 +1,92 @@
...
+import lombok.NonNull;
+import org.apache.lucene.search.BooleanClause;
+import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.QueryVisitor;
+import org.apache.lucene.search.ScoreMode;
+import org.apache.lucene.search.Weight;
+
+/**
+ * Implementation of Query interface for type "hybrid". It allows execution of multiple sub-queries and collect individual
```
navneet1v, line 21: what is the meaning of hybrid here?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseTwoPhaseParameters.java
```diff
@@ -0,0 +1,170 @@
...
+import java.io.IOException;
+import java.util.Locale;
+import java.util.Objects;
+
+@Getter
+@Setter
+@Accessors(chain = true, fluent = true)
+@NoArgsConstructor
+@AllArgsConstructor
+public class NeuralSparseTwoPhaseParameters implements Writeable {
```
navneet1v, line 36: add documentation on all public classes and public function.

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -44,6 +45,7 @@ public boolean searchWith(
         final boolean hasFilterCollector,
         final boolean hasTimeout
     ) throws IOException {
+        addTwoPhaseNeuralSparseQuery(query, searchContext);
```
navneet1v, line 48: What we are doing in this function is not clear from the function name. Are we updating the query clause or we are updating the rescore context? 

Earlier also we discussed that hybrid queryphase searcher is not related to NeuralSparseQuery. Better way to do this exposing an interface from Opensearch core to override the search context.

src/main/java/org/opensearch/neuralsearch/search/util/NeuralSparseTwoPhaseUtil.java
```diff
@@ -0,0 +1,96 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static java.lang.Float.max;
+import static java.lang.Integer.min;
+import static org.opensearch.index.IndexSettings.MAX_RESCORE_WINDOW_SETTING;
+
+public class NeuralSparseTwoPhaseUtil {
```
navneet1v, line 25: Format all the classes to have all the public functions at start and then all the protected functions followed by private functions. doing this enhance the maintainability and readability of classes.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQuery.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Override
+    public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost) throws IOException {
+        return currentQuery.createWeight(searcher, scoreMode, boost);
+    }
+
+    public void extractLowScoreToken() {
+        this.currentQuery = highScoreTokenQuery;
+    }
```
navneet1v, line 91: name of the function doesn't justify the operation performed by the function. Rename the function.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
```diff
@@ -34,4 +33,37 @@ public final class NeuralSearchSettings {
...
+        "plugins.neural_search.neural_sparse.two_phase.default_pruning_ratio",
+        0.4f,
+        0.0f,
+        1.0f,
+        Setting.Property.NodeScope,
+        Setting.Property.Dynamic
+    );
+
+    // the default value is consistent with core settings MAX_RESCORE_WINDOW_SETTING
+    public static final Setting<Integer> NEURAL_SPARSE_TWO_PHASE_MAX_WINDOW_SIZE = Setting.intSetting(
```
navneet1v, line 62: add java docs on all the settings.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Enhancements
 - BWC tests for text chunking processor ([#661](https://github.com/opensearch-project/neural-search/pull/661))
 - Allowing execution of hybrid query on index alias with filters ([#670](https://github.com/opensearch-project/neural-search/pull/670))
+- Allowing query by raw tokens in neural_sparse query ([#693](https://github.com/opensearch-project/neural-search/pull/693))
```
navneet1v, line 23: changelog is not justfiying the changes which are present in this PR.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQuery.java
```diff
@@ -0,0 +1,92 @@
...
+        return new NeuralSparseQuery(rewrittenCurrentQuery, rewrittenFirstStepQuery, rewrittenSecondPhaseQuery, rescoreWindowSizeExpansion);
+    }
+
+    @Override
+    public void visit(QueryVisitor queryVisitor) {
+        QueryVisitor v = queryVisitor.getSubVisitor(BooleanClause.Occur.SHOULD, this);
+        currentQuery.visit(v);
+        highScoreTokenQuery.visit(v);
+        lowScoreTokenQuery.visit(v);
+    }
```
navneet1v, line 62: why we need this?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQuery.java
```diff
@@ -0,0 +1,92 @@
...
+import org.apache.lucene.search.Weight;
+
+/**
+ * Implementation of Query interface for type "hybrid". It allows execution of multiple sub-queries and collect individual
+ * scores for each sub-query.
+ */
+@AllArgsConstructor
+@Getter
+@NonNull
+public final class NeuralSparseQuery extends Query {
```
navneet1v, line 27: earlier NeuralSparseQuery was getting executed via RankFeature Query. Now we are changing to some other query clause? How we are ensuring that the changes are backward compatible?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -95,33 +99,57 @@ public NeuralSparseQueryBuilder(StreamInput in) throws IOException {
...
+        }
+        if (StringUtils.EMPTY.equals(this.modelId)) {
+            this.modelId = null;
+        }
     }
 
     @Override
     protected void doWriteTo(StreamOutput out) throws IOException {
-        out.writeString(fieldName);
-        out.writeString(queryText);
```
navneet1v, line 103: This will cause backward incompatibility, as are removing things from what has been getting written down during serialization.

navneet1v:
> #### Intro
> Now implement two-phase feature by adding rescoreContext in the begin of HybridQueryPhaseSearcher's. It is harmless but little hack, so there is a long term plan to replace this implementation.
> 
> #### RFC in opensearch core
> What we need to implement this feature is the whole query and searchContext, so if OpenSearch Core can open a API which can help me get it before the query performed, we can move the rescoreContext function to a more general interface of core.
> 
> #### RoadMap of this optimizing
> 1. RFC in OpenSearch Core.
> 2. A Incremental update in NeuralSearch plugin.

Please create a RFC on the opensearch core side and provide details why we cannot open up an interface from core to implement this functionality?

navneet1v:
I was not able to complete the full review of the PR as the PR contains more than 1 feature. The description of the PR says it has these 2 feature:

> Enhance the speed of neuralsparse query by two-phase.
> Now support top-level, boolean and boost compound query; for other compound query will degrade into origin logic and speed.

But in the PR we have code add the feature to use tokens in the NeuralSparseQuery clause.

zhichao-aws:
> I was not able to complete the full review of the PR as the PR contains more than 1 feature. The description of the PR says it has these 2 feature:
> 
> > Enhance the speed of neuralsparse query by two-phase.
> > Now support top-level, boolean and boost compound query; for other compound query will degrade into origin logic and speed.
> 
> But in the PR we have code add the feature to use tokens in the NeuralSparseQuery clause.

Yes we merge #693 manually in advance because the query by tokens help us to write IT cases more flexible. Will update this PR after #693 get merged (will be merged soon)

zhichao-aws:
> I was not able to complete the full review of the PR as the PR contains more than 1 feature. The description of the PR says it has these 2 feature:
> 
> > Enhance the speed of neuralsparse query by two-phase.
> > Now support top-level, boolean and boost compound query; for other compound query will degrade into origin logic and speed.
> 
> But in the PR we have code add the feature to use tokens in the NeuralSparseQuery clause.

Updated now.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQuery.java
conggguan: Now change the name to setCurrentQueryToHighScoreTokenQuery().

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
conggguan: Now change the name to addRescoreContextFromNeuralSparseQuery.

As earlier discuss, we decided to remove this from hybridQueryPhaseSearcher asap, RFC draft in opensearch core  are on the way.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
conggguan: Added java docs on these.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQuery.java
conggguan: Sorry, it's a mistake, I added some proper comments on neuralsparse query.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQuery.java
conggguan: I made a update.
Basically, each custom Query which extends Query need to implement this abstract to visit all child queries.
But at now, I think we can visit the current query only, because other sub queries in NeuralSparseQuery are subset of the current query at first.
It's happen on the DefaultAggregationProcessor.proProcess() and have no influence on the search progress now. So just visit the currentQuery will keep the same logic as this update before.
As we concerned, after building a interface on opensearch core we can remove this custom query class also.

conggguan:
Some BWC test failed because of there is a version check of stream.
```java
// stream in 
if (NeuralSparseTwoPhaseParameters.isClusterOnOrAfterMinReqVersionForTwoPhaseSearchSupport()) {
    this.neuralSparseTwoPhaseParameters = in.readOptionalWriteable(NeuralSparseTwoPhaseParameters::new);
}
// stream out
if (NeuralSparseTwoPhaseParameters.isClusterOnOrAfterMinReqVersionForTwoPhaseSearchSupport()) {
    out.writeOptionalWriteable(this.neuralSparseTwoPhaseParameters);
}
```
But at now the V_2_14_0 hasn't been loaded, so after merge will success.
Or we can set Version.CURRENT to MinReqVersion and make a further PR to change it to 2.14 again.

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryIT.java
```diff
@@ -114,6 +141,47 @@ public void testBasicQueryUsingQueryTokens() {
...
+     *                     "c": float
+     *                 },
+     *                 "boost": 2
+     *             }
+     *         }
+     *     }
+     * }
+     */
+    @SneakyThrows
+    public void testBasicQueryUsingQueryTokens_whenTwoPhaseEnabled() {
```
zhichao-aws, line 164: In these test cases, we always get same scores no matter 2-phase is enabled or disabled. And this can't help us find out when there is problem in 2-phase.
Can we design some test cases where we have different scores for 2-phase enabled/disabled? We can just write raw token scores and use raw token scores to search.

zhichao-aws: After we have different behavior for 2-phase enabled/disabled, we should add test cases about update cluster settings to control the behavior, and use DSL to control the behavior (and DSL should overwrite cluster settings)

src/main/java/org/opensearch/neuralsearch/search/util/NeuralSparseTwoPhaseUtil.java
```diff
@@ -0,0 +1,104 @@
...
+    ) {
+        if (query instanceof BoostQuery) {
+            BoostQuery boostQuery = (BoostQuery) query;
+            weight *= boostQuery.getBoost();
+            windoSizeExpansion = max(
+                windoSizeExpansion,
+                populateQueryWeightsMapAndGetWindowSizeExpansion(boostQuery.getQuery(), query2Weight, weight, windoSizeExpansion)
+            );
+        } else if (query instanceof BooleanQuery) {
+            for (BooleanClause clause : (BooleanQuery) query) {
```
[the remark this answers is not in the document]
conggguan: Yes, actually, it will be determine as a boostQuery including a booleanQuery first, and then carry the right boosted weight and go to next step.

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryIT.java
conggguan: Add a test case for this: testNeuralSparseQuery_whenDifferentTwoPhaseParameter_thenGetDifferentResult.
Use different parameter to get expected different result.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ## [Unreleased 2.x](https://github.com/opensearch-project/neural-search/compare/2.13...2.x)
 ### Features
 - Support k-NN radial search parameters in neural search([#697](https://github.com/opensearch-project/neural-search/pull/697))
+- Enhance neural_sparse query's latency performance with two-phase rescore query([#695](https://github.com/opensearch-project/neural-search/pull/695/files)).
```
zhichao-aws, line 23: should change the link 
https://github.com/opensearch-project/neural-search/pull/695/files
to 
https://github.com/opensearch-project/neural-search/pull/695

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilderTests.java
```diff
@@ -346,14 +495,45 @@ public void testFromXContent_whenBuildWithEmptyModelId_thenFail() {
...
 
+    @SneakyThrows
+    public void testFromXContent_whenBuiltWithEmptyTwoPhaseParams_thenThrowException() {
+        /*
+          {
+              "VECTOR_FIELD": {
+                "query_text": "string",
+                "model_id": "string",
+                "two_phase_settings":{
+                      "window_size_expansion": "",
```
zhichao-aws, line 506: the comment is not correct here. It's null instead of empty string

src/main/java/org/opensearch/neuralsearch/search/util/NeuralSparseTwoPhaseUtil.java
```diff
@@ -0,0 +1,106 @@
...
+        if (query2weight.isEmpty()) {
+            return;
+        } else if (query2weight.size() == 1) {
+            Map.Entry<Query, Float> entry = query2weight.entrySet().stream().findFirst().get();
+            twoPhaseQuery = new BoostQuery(entry.getKey(), entry.getValue());
+        } else {
+            twoPhaseQuery = getNestedTwoPhaseQuery(query2weight);
+        }
+        int curWindowSize = (int) (searchContext.size() * windowSizeExpansion);
+        if (curWindowSize < 0 || curWindowSize > NeuralSparseTwoPhaseParameters.MAX_WINDOW_SIZE) {
```
zhichao-aws, line 45: how about change to curWindowSize <= 0?

conggguan: Actually, searchContext.size() is not a IllegalArgumentException on most query.
Such as, use a query with size=0 to get the return information which include hits and other metadata; and do a profile neuralsparse query with size=0 is also possible. So I think maybe there a little help to keep not throwing exception when curWindowsize=0, but make a explanation in doc?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQuery.java
```diff
@@ -0,0 +1,114 @@
...
+    private final Float rescoreWindowSizeExpansion;
+
+    /**
+     *
+     * @param field The field of this query.
+     * @return String of NeuralSparseQuery object.
+     */
+    @Override
+    public String toString(String field) {
+        return "NeuralSparseQuery("
```
martin-gaievski, line 41: can we use String.format or StringBuilder for string concatenation?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -376,4 +430,38 @@ public String getWriteableName() {
...
+
+    private Map<String, Float> getFilteredScoreTokens(Map<String, Float> queryTokens, boolean aboveThreshold, float ratio) {
+        float max = queryTokens.values().stream().max(Float::compare).orElse(0f);
+        float threshold = ratio * max;
+        if (max == 0) {
+            return Collections.emptyMap();
+        }
+        return queryTokens.entrySet()
+            .stream()
+            .filter(entry -> (aboveThreshold == (entry.getValue() >= threshold)))
```
martin-gaievski, line 456: please add comment for logic behind this call

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseTwoPhaseParameters.java
```diff
@@ -0,0 +1,206 @@
...
+    private Float pruning_ratio;
+    private Boolean enabled;
+
+    /**
+     * Initialize when start a cluster.
+     *
+     * @param clusterService The opensearch clusterService.
+     * @param settings       The env settings to initialize.
+     */
+    public static void initialize(ClusterService clusterService, Settings settings) {
```
martin-gaievski, line 66: please make params of public method as `final`. Same for other methods in this class

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseTwoPhaseParameters.java
```diff
@@ -0,0 +1,206 @@
...
+            return false;
+        }
+        NeuralSparseTwoPhaseParameters other = (NeuralSparseTwoPhaseParameters) obj;
+        return Objects.equals(enabled, other.enabled)
+            && Objects.equals(window_size_expansion, other.window_size_expansion)
+            && Objects.equals(pruning_ratio, other.pruning_ratio);
+    }
+
+    public int hashcode() {
+        HashCodeBuilder builder = new HashCodeBuilder().append(enabled).append(window_size_expansion).append(pruning_ratio);
```
martin-gaievski, line 181: if possible can we init this object as part of constructor

src/main/java/org/opensearch/neuralsearch/search/util/NeuralSparseTwoPhaseUtil.java
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Util class for do two phase preprocess for the NeuralSparseQuery.
+ * Include adding the second phase query to searchContext and set the currentQuery to highScoreTokenQuery.
+ */
+public class NeuralSparseTwoPhaseUtil {
+    /**
+     * @param query         The whole query include neuralSparseQuery to executed.
+     * @param searchContext The searchContext with this query.
+     */
+    public static void addRescoreContextFromNeuralSparseQuery(final Query query, SearchContext searchContext) {
```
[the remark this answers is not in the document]
martin-gaievski: Please add comment for this method

src/main/java/org/opensearch/neuralsearch/search/util/NeuralSparseTwoPhaseUtil.java
```diff
@@ -0,0 +1,106 @@
...
+        } else if (query2weight.size() == 1) {
+            Map.Entry<Query, Float> entry = query2weight.entrySet().stream().findFirst().get();
+            twoPhaseQuery = new BoostQuery(entry.getKey(), entry.getValue());
+        } else {
+            twoPhaseQuery = getNestedTwoPhaseQuery(query2weight);
+        }
+        int curWindowSize = (int) (searchContext.size() * windowSizeExpansion);
+        if (curWindowSize < 0 || curWindowSize > NeuralSparseTwoPhaseParameters.MAX_WINDOW_SIZE) {
+            throw new IllegalArgumentException(
+                "Two phase final windowSize "
```
martin-gaievski, line 47: Please use String.format for the exception message

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryIT.java
```diff
@@ -206,6 +282,55 @@ public void testBooleanQuery_withMultipleSparseEncodingQueries() {
...
+     *         }
+     *     }
+     * }
+     */
+    @SneakyThrows
+    public void testBooleanQuery_withMultipleSparseEncodingQueries_whenTwoPhaseEnabled() {
+        String modelId = null;
+        try {
+            initializeIndexIfNotExist(TEST_MULTI_NEURAL_SPARSE_FIELD_INDEX_NAME);
+            modelId = prepareSparseEncodingModel();
```
martin-gaievski, line 310: We need to rework this integ test. There are 12 new tests where we're uploading a model. Recently we have found that in 2.13+ for jdk21 that causes exception from ml-commons CB when tests are running in infra CI. For now the only working solution is to minimize number of model deployments/undeployments. 

Please consider moving most of test cases to unit test. Integ tests should be used to test end to end flow, not to test all permutations and negative cases.

martin-gaievski (changes requested):
My big concern is high number of integ tests with model deployment, the rest is less critical

src/main/java/org/opensearch/neuralsearch/search/util/NeuralSparseTwoPhaseUtil.java
```diff
@@ -0,0 +1,106 @@
...
+                    + "[plugins.neural_search.neural_sparse.two_phase.max_window_size] to a integer at least 50."
+            );
+        }
+        QueryRescorer.QueryRescoreContext rescoreContext = new QueryRescorer.QueryRescoreContext(curWindowSize);
+        rescoreContext.setQuery(twoPhaseQuery);
+        rescoreContext.setRescoreQueryWeight(getOriginQueryWeightAfterRescore(searchContext.rescore()));
+        searchContext.addRescore(rescoreContext);
+    }
+
+    private static float populateQueryWeightsMapAndGetWindowSizeExpansion(
```
navneet1v, line 67: This function needs to be split into 2 parts.
1. Where are are just changing the rescore query/recore context.
2. Where we are changing the actual Query clause provided by user to set what should be the currentQuery that needs to be executed.

We should also add some warnings which should tell us that query clauses was not boost, bool or neuralsparse hence this two phase optimization is not applied.

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryIT.java
navneet1v: +1 on this comment.

src/main/java/org/opensearch/neuralsearch/search/util/NeuralSparseTwoPhaseUtil.java
conggguan: For First comments, I refactor the function addRescoreContextFromNeuralSparseQuery. You can take a look for further optimization.

For the warn logs, I think, If we want to detect all of compound queries that if there has a neuralSparseQuery which couldn't be speed up, maybe it cost some performance loss, such as when a hybrid query, constant score query, dismax query  of a filter boolean query come in, we must to traversal all the sub queries of them and make a warn log.
So I think maybe make a illustration in OpenSearch Document is better for neural-search plugin's performance and log's simplicity.
If you think there is a very strong need to add warn log ?

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
```diff
@@ -34,4 +33,49 @@ public final class NeuralSearchSettings {
         50,
         Setting.Property.NodeScope
     );
+
+    /**
+     * Use this setting to manage if a neuralSparseQuery build a two-phase query of not.
+     */
+    public static final Setting<Boolean> NEURAL_SPARSE_TWO_PHASE_DEFAULT_ENABLED = Setting.boolSetting(
```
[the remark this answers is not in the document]
conggguan: The reason for setting it as a node-scope setting is that we believe the feature's settings are relatively stable. In our initial investigations, we observed great performance and accuracy with the default settings. We also think that significant differences in two-phase parameters might cause variations in accuracy across different indexes. Therefore, we made this a node-scope setting. 
Additionally, for customization in a single query, we provide a one-time two-phase parameter that allows users to temporarily override this setting in an individual query."
```
{
  "query": {
    "neural_sparse":{
      "VECTOR_FIELD":
      {
        "model_id": "model_id",
        "query_text": "abcd",
        "two_phase_settings":
         {					
            "enabled": true
            "window_size_expansion": 5,  
            "pruning_ratio": 0.4,  
         }
      }
    }
  }
}
```

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryIT.java
conggguan: I replaced the model inference with a fixed query token in most of the newly added test cases.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseTwoPhaseParameters.java
conggguan: Change all public method para as final , besides euqals(Object object)

src/main/java/org/opensearch/neuralsearch/search/util/NeuralSparseTwoPhaseUtil.java
```diff
@@ -0,0 +1,106 @@
...
+     * @param searchContext The searchContext with this query.
+     */
+    public static void addRescoreContextFromNeuralSparseQuery(final Query query, SearchContext searchContext) {
+        Map<Query, Float> query2weight = new HashMap<>();
+        float windowSizeExpansion = populateQueryWeightsMapAndGetWindowSizeExpansion(query, query2weight, 1.0f, 1.0f);
+        Query twoPhaseQuery;
+        if (query2weight.isEmpty()) {
+            return;
+        } else if (query2weight.size() == 1) {
+            Map.Entry<Query, Float> entry = query2weight.entrySet().stream().findFirst().get();
```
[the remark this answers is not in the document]
conggguan: delete the extra else,
replaced the findFirst.

src/main/java/org/opensearch/neuralsearch/search/util/NeuralSparseTwoPhaseUtil.java
conggguan: Replaced this with format

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQuery.java
conggguan: Replaced with Format.

src/main/java/org/opensearch/neuralsearch/search/util/NeuralSparseTwoPhaseUtil.java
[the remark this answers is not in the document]
conggguan: Add a comment for this method, changed the searchContext to final argument.

zhichao-aws:
> Lets ensure that all BWC tests are successful before this change can be merged.

Some BWC test failed because of there is a version check of stream.

// stream in 
if (NeuralSparseTwoPhaseParameters.isClusterOnOrAfterMinReqVersionForTwoPhaseSearchSupport()) {
    this.neuralSparseTwoPhaseParameters = in.readOptionalWriteable(NeuralSparseTwoPhaseParameters::new);
}
// stream out
if (NeuralSparseTwoPhaseParameters.isClusterOnOrAfterMinReqVersionForTwoPhaseSearchSupport()) {
    out.writeOptionalWriteable(this.neuralSparseTwoPhaseParameters);
}
But at now the PR hasn't been backport to V_2_14_0, so the 2.14 node can not deserialize the stream correctly. It can only work after backport success.
One workaround is using Version.Current, but we need to modify it to 2.14 after backport finished.

src/main/java/org/opensearch/neuralsearch/search/util/NeuralSparseTwoPhaseUtil.java
zhichao-aws: If we only consider the supported compound queires, we can early exit for unsupported cases.
However, if we want to log warn as you mentioned, we need to go through to the end for every compound queries. 
Because here is the common entry for all queries, if users are not using semantic search and need to do many complexed compound queries, going through all of them will have performance impact.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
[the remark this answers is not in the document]
navneet1v: if the settings are stable and we already provide user a way to set the settings in the query itself then we don't need to explicitly provide setting at cluster level because the way I look at this is these settings are of no use. They just taking space in cluster state. Hence I would recommend not having the settings. Plus more I look at this, these are not setting these are more of Neural Sparse query attributes. Hence, we should just remove the settings completely.

src/main/java/org/opensearch/neuralsearch/search/util/NeuralSparseTwoPhaseUtil.java
navneet1v: I got what you are trying to say. Now help me understand this, if I have a query with multiple level of nesting of bool -> should -> bool -> should -> dis_max -> neuralsparse in this case the optimization will not happen and we will do these iterations to check neural sparse query is present or not.


Moreover now this check is going to happen for every bool query(simple or complex), it has a bigger blast radius and can impact performance of every user query. As the call is recursive so it will iterate over every query clause if it is bool.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseTwoPhaseParameters.java
```diff
@@ -0,0 +1,207 @@
...
+@Getter
+@Setter
+@Accessors(chain = true, fluent = true)
+@NoArgsConstructor
+@AllArgsConstructor
+public class NeuralSparseTwoPhaseParameters implements Writeable {
+    public static volatile Float DEFAULT_WINDOW_SIZE_EXPANSION;
+    public static volatile Float DEFAULT_PRUNING_RATIO;
+    public static volatile Boolean DEFAULT_ENABLED;
+    public static volatile Integer MAX_WINDOW_SIZE;
```
navneet1v, line 45: why public and volatile?

navneet1v: and if field is static and final then the name should be all caps. But that is not the case here.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseTwoPhaseParameters.java
```diff
@@ -0,0 +1,207 @@
...
+    @VisibleForTesting
+    static final ParseField NAME = new ParseField("two_phase_settings");
+    @VisibleForTesting
+    static final ParseField WINDOW_SIZE_EXPANSION = new ParseField("window_size_expansion");
+    @VisibleForTesting
+    static final ParseField PRUNING_RATIO = new ParseField("pruning_ratio");
+    @VisibleForTesting
+    static final ParseField ENABLED = new ParseField("enabled");
+    private static final Version MINIMAL_SUPPORTED_VERSION_TWO_PHASE_SEARCH = Version.CURRENT;
```
navneet1v, line 54: Version.Current is not right way. Wondering how it worked for other features?

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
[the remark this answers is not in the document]
zhichao-aws: The settings is stable for **a concrete sparse model**. If the sparse model is trained with [thresholding in loss function](https://dl.acm.org/doi/pdf/10.1145/3539618.3592051), they can have different hyperparameters to work with. While the default settings (5, 0.4) works best for our sparse models.

Let's make the decision for keeping these settings based on pros and cons:
pros:
- make it more convinient for users with customized sparse models
- make it more convinient to disable the 2-phase search, there are use cases to use OpenSearch to benchmark the sparse model. For these cases we need accurate search relevance scores
cons:
- having 3 more cluster settings, taking dozens of bytes in memory

Many settings in OpenSearch are rarely changed. Most users won't feel this settings, but for users who need the settings, it is very important to be able to configure them. I don't think this is a critical for performance, hence one single embedding vector will take much more space than these settings.

src/main/java/org/opensearch/neuralsearch/search/util/NeuralSparseTwoPhaseUtil.java
zhichao-aws: > I got what you are trying to say. Now help me understand this, if I have a query with multiple level of nesting of bool -> should -> bool -> should -> dis_max -> neuralsparse in this case the optimization will not happen and we will do these iterations to check neural sparse query is present or not.

This is why we need to early exit. We iterate deeper only if its the query clause of specific type. I.e. for a complexed query clause tree, once we found current node is not of specific types we just exit to minimize the performance impact.

With this early exit, we just go through all boolean and boost, and we don't find some use case have deep nested boolean and boost but without nodes we can early exit from.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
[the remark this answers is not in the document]
navneet1v: If we are defining these settings based on the pros you mentioned then these settings are much more suited at index level and not at cluster level. Because if I have 2 indices using 2 different sparse models, then this setting will be of use. But if we make them at index level then atleast they are useful as I can change setting for 1 index. 

Moreover if you think little hard on that point, more than 1 model can be used in 2 different sparse fields on same index. In that case not even the index settings be useful. Hence if you see the settings are not very useful in that perspective because for most of the cases default works and for cases where default doesn't work user can update the query.

> cons:
having 3 more cluster settings, taking dozens of bytes in memory

Replicate this dozen of bytes over more than 100 nodes then you will see the impact. Nevertheless the point for that comment was, if the settings are not useful as I mentioned above then just adding them in cluster setting is just waste of resources. I cannot comment on other settings which customer don't change and still present in Opensearch. But all I will say is if something is not correct in current code is not a justification to follow the same pattern.

src/main/java/org/opensearch/neuralsearch/search/util/NeuralSparseTwoPhaseUtil.java
zhichao-aws: > Another point, if we are not adding any log around whether twoPhaseOptimization got added or not, how we are expecting admin/operators to Opensearch to figure out whether for a given query optimization worked or not?

This feature is not about functionality, but rather an enhancements to search experience. We expect most users can just benefits from the speeds up without learning how it works or configure it(except those who need to customize the settings based on their use case). We'll just pick the part we can speed up and operate it.(like the filter-while-search, we don't have such detailed log there) 

If users want to know whether its enabled, they can just compare the search speed with 2-phase enabled and disabled.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseTwoPhaseParameters.java
zhichao-aws: > why public and volatile?

Using a volatile field and settings update consumer is a common approach in OpenSearch to get real-time cluster settings from memory, without accessing the cluster settings object. You can just find many references in other plugins.

The public can be removed.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseTwoPhaseParameters.java
zhichao-aws: Please take a look at what I respond you yesterday.

zhichao-aws: https://github.com/opensearch-project/neural-search/pull/695#issuecomment-2076819753

navneet1v: I checked that, but my question is simple. This is not the first feature who is facing this problem. So can you tell me how this particular use case is special

src/main/java/org/opensearch/neuralsearch/search/util/NeuralSparseTwoPhaseUtil.java
navneet1v: > > I got what you are trying to say. Now help me understand this, if I have a query with multiple level of nesting of bool -> should -> bool -> should -> dis_max -> neuralsparse in this case the optimization will not happen and we will do these iterations to check neural sparse query is present or not.
> 
> This is why we need to early exit. We iterate deeper only if its the query clause of specific type. I.e. for a complexed query clause tree, once we found current node is not of specific types we just exit to minimize the performance impact.
> 
> With this early exit, we just go through all boolean and boost, and we don't find some use case have deep nested boolean and boost but without nodes we can early exit from.
> 
> But give an warning log for every query has large impact for performance.

So in my message there 2 points. Seems like you read only one. 

Here is the thing:
1. The code is not going to do any early exit if there is bool clause every where. I am hoping that is clear. 
2. Comparing the latencies of 2 different queries is not an acceptable user explanation here why the queries didn't speed up. So if want operators to understand why a speed up didn't happen we better figure out a way. Otherwise there can be a flurry of questions around why the query is not speeding up. Thats where helpful log message are useful. If there is a better way than log I would suggest to do that.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseTwoPhaseParameters.java
zhichao-aws: The bwc framework is introduced after 2.11, and it's test cases(neural query, neural sparse query, and hybrid query) are already exists when we write this framework.

navneet1v: > The bwc framework is introduced after 2.11, and it's test cases(neural query, neural sparse query, and hybrid query) are already exists when we write this framework.


So what difference does it make?

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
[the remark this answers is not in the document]
zhichao-aws: We evaluated implementing them as index settings, and it will have performance issue.
For cluster settings we can use volatile object to just fetch the value from memory. But for index settings we must access the settings object and go through the [get setting API](https://github.com/opensearch-project/OpenSearch/blob/81707dcc74fd8a66349c80c0c47254de110b9f5d/server/src/main/java/org/opensearch/common/settings/Setting.java#L480), for every neural sparse search request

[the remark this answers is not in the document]
navneet1v: If a setting is supposed to be an index setting and we making it as part of cluster settings just on the name of performance, do we know how much performance boost we are getting here ? Plus index settings are also stored on the node. So, when settings change you might get hit for 1 query but other queries will be fast.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseTwoPhaseParameters.java
zhichao-aws: Can you give an example of what you think the same case, and we take a look at whether they are difference.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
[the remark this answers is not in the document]
zhichao-aws: > If a setting is supposed to be an index setting and we making it as part of cluster settings just on the name of performance, do we know how much performance boost we are getting here?
The performance impact is very straightforward in code so we haven't benchmark it. For cluster settings, we just do value assignment; and for index settings, we need to iterate the environment settings. And this will happen for every neural sparse query.

> Plus index settings are also stored on the node. So, when settings change you might get hit for 1 query but other queries will be fast.
Do you mean we cache these settings for each index? We need to implement another complexed controller module to manage the cache for each index and do update even manage their lifecycle with index create/delete. I think this is completely overdesigned. You just mentioned the overhead to add one setting, implmenting index settings will make this overhead much larger because in every index we have these settings.

src/main/java/org/opensearch/neuralsearch/search/util/NeuralSparseTwoPhaseUtil.java
zhichao-aws: > So in my message there 2 points. Seems like you read only one.
I gave 2 response for the 2 points.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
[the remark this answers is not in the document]
navneet1v: > Do you mean we cache these settings for each index? We need to implement another complexed controller module to manage the cache for each index and do update even manage their lifecycle with index create/delete. I think this is completely overdesigned. You just mentioned the overhead to add one setting, implmenting index settings will make this overhead much larger because in every index we have these settings.

I am saying they are already cached at the node level. You don't need to cache it. So any time there is change in settings(index/cluster settings) for any index those settings are pushed to all the data nodes. 

> For cluster settings, we just do value assignment; and for index settings, we need to iterate the environment settings. And this will happen for every neural sparse query.

Not sure what you mean here.

src/main/java/org/opensearch/neuralsearch/search/util/NeuralSparseTwoPhaseUtil.java
navneet1v: The reply for both of them is added. Do you agree with \#1 or not?

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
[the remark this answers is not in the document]
zhichao-aws: OK I get your point, I checked the settings API and found I misunderstand how it works. 
So you put +1 to index settings now?

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryIT.java
zhichao-aws: After we changed to use query_tokens, we don't deploy new model for new neural sparse integ test.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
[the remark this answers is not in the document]
conggguan: Changed to indexscope.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
martin-gaievski: nit: you should put comment before the code, not a blocker but please in mind for next PRs

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseTwoPhaseParameters.java
martin-gaievski: got it, I'm ok with current implementation then

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -376,4 +417,58 @@ public String getWriteableName() {
...
+
+    private BooleanQuery buildFeatureFieldQueryFromTokens(Map<String, Float> tokens, String fieldName) {
+        BooleanQuery.Builder builder = new BooleanQuery.Builder();
+        for (Map.Entry<String, Float> entry : tokens.entrySet()) {
+            builder.add(FeatureField.newLinearQuery(fieldName, entry.getKey(), entry.getValue()), BooleanClause.Occur.SHOULD);
+        }
+        return builder.build();
+    }
+
+    private void completeTwoPhasePara(IndexSettings indexSettings) {
```
martin-gaievski, line 458: are there some letters missing in the method name, or it's a weird name? :)

martin-gaievski, line 458: Please fix formatting in this method, every method call should be inside the curly braces and on a new line

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseTwoPhaseParameters.java
martin-gaievski: I'm not sure minimal supported version should be current. It mean with each release it will be incremented and not supported from previous release? Say we merge this before 2.15, then in 2.16 branch 2.16 is minimal supported meaning 2.15 is not supported

PR #710, opened 2024-04-25
Add support for asymmetric embedding models

### LINKED ISSUE #620: [PROPOSAL] Add support for asymmetric embedding models to neural-search
### What are you proposing?
With https://github.com/opensearch-project/ml-commons/issues/1799 OpenSearch can now use asymmetric embedding models such as e5. Asymmetric embedding models differ from symmetric ones by embedding queries and passages differently. Many of them require the text to be prefixed by special strings (e.g. `query: ` and `passage: ` for the e5 model family). The aforementioned issue documents the work that led to ml-commons supporting serving these kinds of model. There, new model configurations have been added, a special new `AsymmetricTextEmbeddingParameters` type was introduced and the correct handling of inference for these models has been implemented. Here I want to propose to add support for this in the neural-search plugin. 

I'm a user and would like to have this feature. 😉
Many of the best performing embedding models are asymmetric ones (cf. MTEB board https://huggingface.co/spaces/mteb/leaderboard). For OpenSearch it is important to keep up with the latest developments in information retrieval research and give users the freedom to choose as many different models as possible.

Neural-search should be able to use the newly introduced capability in ml-commons.

My proposal will not affect any API. The idea is to have neural-search automatically detect that a particular model being used is asymmetric (it is possible to infer this information from the model configuration) and then call the inference method with the right parameters. In the embedding processor, we need to add the information that we are embedding a passage; in the query builder, that we are embedding a query. That's it.

No.

No.

The user experience will not change in any way.

No.

Not building this reduces the attractiveness of OS for dense retrieval. Other vector DBs already support these models.

The ground work has been done in ml-commons, here we only need to build on top of it. I will open a PR soon on which we can base the discussion.

I can't think of any.

br3no:
### Description
This PR adds support for asymmetric embedding models such as https://huggingface.co/intfloat/multilingual-e5-small to the neural-search plugin.

It builds on the work done in https://github.com/opensearch-project/ml-commons/issues/1799.

Asymmetric embedding models behave differently when embedding passages and queries. For that end, the model must "know" on inference time, what kind of data it is embedding.

The changes are:
#### 1. `src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java`
The processor signals it is embedding passages, by passing the new `AsymmetricTextEmbeddingParameters` using the content type `EmbeddingContentType.PASSAGE`.
#### 2. `src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java`
Analogously, the query builder uses `EmbeddingContentType.QUERY`.
#### 3. `src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java`
Here is where most of the work was done. The class has been extended in a backwards-compatible way with inference methods that allow one to pass `MLAlgoParams` objects. Usage of `AsymmetricTextEmbeddingParameters` (which implements `MLAlgoParams`) is mandatory for asymmetric models. At the same time symmetric models do not accept them. 
    
The only way to know whether a model is asymmetric or symmetric is by reading its model configuration (if the models' configuration contains a `passage_prefix` and/or a `query_prefix`, they are asymmetric, otherwise they are symmetric). 
    
The `src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java` class deals with this, keeping the complexity in one place and not requiring any API change to the neural-search plugin (as proposed in #620). When calling the inference methods, clients (such as the `TextEmbeddingProcessor`) may pass the `AsymmetricTextEmbeddingParameters` object without caring if the model they are using is symmetric or asymmetric. The accessor class will first read the model's configuration (by calling the `getModel` API of the `mlClient`) and deal appropriately. 
    
To avoid adding this extra roundtrip to every inference call, the asymmetry information is kept in a cache in memory.

### Issues Resolved
#620 

### Check List
    - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
@@ -11,16 +11,24 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
+import org.opensearch.common.CheckedConsumer;
+import org.opensearch.common.Nullable;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.common.util.CollectionUtils;
 import org.opensearch.ml.client.MachineLearningNodeClient;
 import org.opensearch.ml.common.FunctionName;
+import org.opensearch.ml.common.MLModel;
 import org.opensearch.ml.common.dataset.MLInputDataset;
 import org.opensearch.ml.common.dataset.TextDocsInputDataSet;
 import org.opensearch.ml.common.dataset.TextSimilarityInputDataSet;
 import org.opensearch.ml.common.input.MLInput;
+import org.opensearch.ml.common.input.parameter.MLAlgoParams;
+import org.opensearch.ml.common.model.MLModelConfig;
+import org.opensearch.ml.common.model.TextEmbeddingModelConfig;
 import org.opensearch.ml.common.output.MLOutput;
 import org.opensearch.ml.common.output.model.ModelResultFilter;
 import org.opensearch.ml.common.output.model.ModelTensor;
@@ -40,6 +48,7 @@ import lombok.extern.log4j.Log4j2;
 public class MLCommonsClientAccessor {
     private static final List<String> TARGET_RESPONSE_FILTERS = List.of("sentence_embedding");
     private final MachineLearningNodeClient mlClient;
+    private final Map<String, Boolean> modelAsymmetryCache = new ConcurrentHashMap<>();
 
     /**
      * Wrapper around {@link #inferenceSentences} that expected a single input text and produces a single floating
@@ -54,7 +63,29 @@ public class MLCommonsClientAccessor {
         @NonNull final String inputText,
         @NonNull final ActionListener<List<Float>> listener
     ) {
-        inferenceSentences(TARGET_RESPONSE_FILTERS, modelId, List.of(inputText), ActionListener.wrap(response -> {
+        inferenceSentence(modelId, inputText, null, listener);
+    }
+
+    /**
+     * Wrapper around {@link #inferenceSentences} that expected a single input text and produces a single floating
+     * point vector as a response. Supports passing {@link MLAlgoParams} to the inference. If the model is
+     * asymmetric, passing a
+     * {@link org.opensearch.ml.common.input.parameter.textembedding.AsymmetricTextEmbeddingParameters} is
+     * mandatory. This method will check whether the model being used is asymmetric and correctly handle the
+     * parameter, so it's okay to always pass the parameter (even if the model is symmetric).
+     *
+     * @param modelId {@link String}
+     * @param inputText {@link List} of {@link String} on which inference needs to happen
+     * @param mlAlgoParams {@link MLAlgoParams} which will be used to run the inference
+     * @param listener {@link ActionListener} which will be called when prediction is completed or errored out
+     */
+    public void inferenceSentence(
+        @NonNull final String modelId,
+        @NonNull final String inputText,
+        @Nullable final MLAlgoParams mlAlgoParams,
+        @NonNull final ActionListener<List<Float>> listener
+    ) {
+        inferenceSentences(TARGET_RESPONSE_FILTERS, modelId, List.of(inputText), mlAlgoParams, ActionListener.wrap(response -> {
             if (response.size() != 1) {
                 listener.onFailure(
                     new IllegalStateException(
@@ -69,35 +100,62 @@ public class MLCommonsClientAccessor {
     }
 
     /**
-     * Abstraction to call predict function of api of MLClient with default targetResponse filters. It uses the
-     * custom model provided as modelId and run the {@link FunctionName#TEXT_EMBEDDING}. The return will be sent
-     * using the actionListener which will have a {@link List} of {@link List} of {@link Float} in the order of
-     * inputText. We are not making this function generic enough to take any function or TaskType as currently we
-     * need to run only TextEmbedding tasks only.
+     * Abstraction to call predict function of api of MLClient with default targetResponse filters. It
+     * uses the custom model provided as modelId and run the {@link FunctionName#TEXT_EMBEDDING}. The
+     * return will be sent using the actionListener which will have a {@link List} of {@link List} of
+     * {@link Float} in the order of inputText. We are not making this function generic enough to take
+     * any function or TaskType as currently we need to run only TextEmbedding tasks only.
      *
-     * @param modelId {@link String}
+     * @param modelId   {@link String}
      * @param inputText {@link List} of {@link String} on which inference needs to happen
-     * @param listener {@link ActionListener} which will be called when prediction is completed or errored out
+     * @param listener  {@link ActionListener} which will be called when prediction is completed or
+     *                  errored out
      */
     public void inferenceSentences(
         @NonNull final String modelId,
         @NonNull final List<String> inputText,
         @NonNull final ActionListener<List<List<Float>>> listener
     ) {
-        inferenceSentences(TARGET_RESPONSE_FILTERS, modelId, inputText, listener);
+        inferenceSentences(TARGET_RESPONSE_FILTERS, modelId, inputText, null, listener);
     }
 
     /**
-     * Abstraction to call predict function of api of MLClient with provided targetResponse filters. It uses the
-     * custom model provided as modelId and run the {@link FunctionName#TEXT_EMBEDDING}. The return will be sent
-     * using the actionListener which will have a {@link List} of {@link List} of {@link Float} in the order of
-     * inputText. We are not making this function generic enough to take any function or TaskType as currently we
-     * need to run only TextEmbedding tasks only.
+     * Abstraction to call predict function of api of MLClient with default targetResponse filters. It
+     * uses the custom model provided as modelId and run the {@link FunctionName#TEXT_EMBEDDING}. The
+     * return will be sent using the actionListener which will have a {@link List} of {@link List} of
+     * {@link Float} in the order of inputText. We are not making this function generic enough to take
+     * any function or TaskType as currently we need to run only TextEmbedding tasks only. Supports
+     * passing {@link MLAlgoParams} to the inference. If the model is asymmetric, passing a
+     * {@link org.opensearch.ml.common.input.parameter.textembedding.AsymmetricTextEmbeddingParameters}
+     * is mandatory. This method will check whether the model being used is asymmetric and correctly
+     * handle the parameter, so it's okay to always pass the parameter (even if the model is symmetric).
      *
-     * @param targetResponseFilters {@link List} of {@link String} which filters out the responses
      * @param modelId {@link String}
      * @param inputText {@link List} of {@link String} on which inference needs to happen
-     * @param listener {@link ActionListener} which will be called when prediction is completed or errored out.
+     * @param mlAlgoParams {@link MLAlgoParams} which will be used to run the inference
+     * @param listener {@link ActionListener} which will be called when prediction is completed or
+     */
+    public void inferenceSentences(
+        @NonNull final String modelId,
+        @NonNull final List<String> inputText,
+        @Nullable final MLAlgoParams mlAlgoParams,
+        @NonNull final ActionListener<List<List<Float>>> listener
+    ) {
+        inferenceSentences(TARGET_RESPONSE_FILTERS, modelId, inputText, mlAlgoParams, listener);
+    }
+
+    /**
+     * Abstraction to call predict function of api of MLClient with provided targetResponse filters.
+     * It uses the custom model provided as modelId and run the {@link FunctionName#TEXT_EMBEDDING}.
+     * The return will be sent using the actionListener which will have a {@link List} of {@link List}
+     * of {@link Float} in the order of inputText. We are not making this function generic enough to
+     * take any function or TaskType as currently we need to run only TextEmbedding tasks only.
+     *
+     * @param targetResponseFilters {@link List} of {@link String} which filters out the responses
+     * @param modelId               {@link String}
+     * @param inputText             {@link List} of {@link String} on which inference needs to happen
+     * @param listener              {@link ActionListener} which will be called when prediction is
+     *                              completed or errored out.
      */
     public void inferenceSentences(
         @NonNull final List<String> targetResponseFilters,
@@ -105,7 +163,35 @@ public class MLCommonsClientAccessor {
         @NonNull final List<String> inputText,
         @NonNull final ActionListener<List<List<Float>>> listener
     ) {
-        retryableInferenceSentencesWithVectorResult(targetResponseFilters, modelId, inputText, 0, listener);
+        inferenceSentences(targetResponseFilters, modelId, inputText, null, listener);
+    }
+
+    /**
+     * Abstraction to call predict function of api of MLClient with provided targetResponse filters.
+     * It uses the custom model provided as modelId and run the {@link FunctionName#TEXT_EMBEDDING}.
+     * The return will be sent using the actionListener which will have a {@link List} of {@link List}
+     * of {@link Float} in the order of inputText. We are not making this function generic enough to
+     * take any function or TaskType as currently we need to run only TextEmbedding tasks only. Supports
+     * passing {@link MLAlgoParams} to the inference. If the model is asymmetric, passing a
+     * {@link org.opensearch.ml.common.input.parameter.textembedding.AsymmetricTextEmbeddingParameters}
+     * is mandatory. This method will check whether the model being used is asymmetric and correctly
+     * handle the parameter, so it's okay to always pass the parameter (even if the model is symmetric).
+     *
+     * @param targetResponseFilters {@link List} of {@link String} which filters out the responses
+     * @param modelId               {@link String}
+     * @param inputText             {@link List} of {@link String} on which inference needs to happen
+     * @param mlAlgoParams          {@link MLAlgoParams} which will be used to run the inference
+     * @param listener              {@link ActionListener} which will be called when prediction is
+     *                              completed or errored out.
+     */
+    public void inferenceSentences(
+        @NonNull final List<String> targetResponseFilters,
+        @NonNull final String modelId,
+        @NonNull final List<String> inputText,
+        @Nullable final MLAlgoParams mlAlgoParams,
+        @NonNull final ActionListener<List<List<Float>>> listener
+    ) {
+        retryableInferenceSentencesWithVectorResult(targetResponseFilters, modelId, inputText, mlAlgoParams, 0, listener);
     }
 
     public void inferenceSentencesWithMapResult(
@@ -113,35 +199,65 @@ public class MLCommonsClientAccessor {
         @NonNull final List<String> inputText,
         @NonNull final ActionListener<List<Map<String, ?>>> listener
     ) {
-        retryableInferenceSentencesWithMapResult(modelId, inputText, 0, listener);
+        retryableInferenceSentencesWithMapResult(modelId, inputText, null, 0, listener);
     }
 
     /**
-     * Abstraction to call predict function of api of MLClient with provided targetResponse filters. It uses the
-     * custom model provided as modelId and run the {@link FunctionName#TEXT_EMBEDDING}. The return will be sent
-     * using the actionListener which will have a list of floats in the order of inputText.
+     * Abstraction to call predict function of api of MLClient with provided targetResponse filters.
+     * It uses the custom model provided as modelId and run the {@link FunctionName#TEXT_EMBEDDING}.
+     * The return will be sent using the actionListener which will have a list of floats in the order
+     * of inputText.
      *
-     * @param modelId {@link String}
-     * @param inputObjects {@link Map} of {@link String}, {@link String} on which inference needs to happen
-     * @param listener {@link ActionListener} which will be called when prediction is completed or errored out.
+     * @param modelId      {@link String}
+     * @param inputObjects {@link Map} of {@link String}, {@link String} on which inference needs to
+     *                     happen
+     * @param listener     {@link ActionListener} which will be called when prediction is completed or
+     *                     errored out.
      */
     public void inferenceSentences(
         @NonNull final String modelId,
         @NonNull final Map<String, String> inputObjects,
         @NonNull final ActionListener<List<Float>> listener
     ) {
-        retryableInferenceSentencesWithSingleVectorResult(TARGET_RESPONSE_FILTERS, modelId, inputObjects, 0, listener);
+        inferenceSentences(modelId, inputObjects, null, listener);
     }
 
     /**
-     * Abstraction to call predict function of api of MLClient. It uses the custom model provided as modelId and the
-     * {@link FunctionName#TEXT_SIMILARITY}. The return will be sent via actionListener as a list of floats representing
-     * the similarity scores of the texts w.r.t. the query text, in the order of the input texts.
+     * Abstraction to call predict function of api of MLClient with provided targetResponse filters.
+     * It uses the custom model provided as modelId and run the {@link FunctionName#TEXT_EMBEDDING}.
+     * The return will be sent using the actionListener which will have a list of floats in the order
+     * of inputText. Supports passing {@link MLAlgoParams} to the inference. If the model is asymmetric,
+     * passing a
+     * {@link org.opensearch.ml.common.input.parameter.textembedding.AsymmetricTextEmbeddingParameters}
+     * is mandatory. This method will check whether the model being used is asymmetric and correctly
+     * handle the parameter, so it's okay to always pass the parameter (even if the model is symmetric).
      *
-     * @param modelId {@link String} ML-Commons Model Id
+     * @param modelId      {@link String}
+     * @param inputObjects {@link Map} of {@link String}, {@link String} on which inference needs to
+     *                     happen
+     * @param mlAlgoParams {@link MLAlgoParams} which will be used to run the inference
+     * @param listener     {@link ActionListener} which will be called when prediction is completed or
+     *                     errored out.
+     */
+    public void inferenceSentences(
+        @NonNull final String modelId,
+        @NonNull final Map<String, String> inputObjects,
+        @Nullable final MLAlgoParams mlAlgoParams,
+        @NonNull final ActionListener<List<Float>> listener
+    ) {
+        retryableInferenceSentencesWithSingleVectorResult(TARGET_RESPONSE_FILTERS, modelId, inputObjects, null, 0, listener);
+    }
+
+    /**
+     * Abstraction to call predict function of api of MLClient. It uses the custom model provided as
+     * modelId and the {@link FunctionName#TEXT_SIMILARITY}. The return will be sent via
+     * actionListener as a list of floats representing the similarity scores of the texts w.r.t. the
+     * query text, in the order of the input texts.
+     *
+     * @param modelId   {@link String} ML-Commons Model Id
      * @param queryText {@link String} The query to compare all the inputText to
      * @param inputText {@link List} of {@link String} The texts to compare to the query
-     * @param listener {@link ActionListener} receives the result of the inference
+     * @param listener  {@link ActionListener} receives the result of the inference
      */
     public void inferenceSimilarity(
         @NonNull final String modelId,
@@ -155,42 +271,95 @@ public class MLCommonsClientAccessor {
     private void retryableInferenceSentencesWithMapResult(
         final String modelId,
         final List<String> inputText,
+        final MLAlgoParams mlAlgoParams,
         final int retryTime,
         final ActionListener<List<Map<String, ?>>> listener
     ) {
-        MLInput mlInput = createMLTextInput(null, inputText);
-        mlClient.predict(modelId, mlInput, ActionListener.wrap(mlOutput -> {
-            final List<Map<String, ?>> result = buildMapResultFromResponse(mlOutput);
-            listener.onResponse(result);
-        }, e -> {
-            if (RetryUtil.shouldRetry(e, retryTime)) {
-                final int retryTimeAdd = retryTime + 1;
-                retryableInferenceSentencesWithMapResult(modelId, inputText, retryTimeAdd, listener);
-            } else {
-                listener.onFailure(e);
-            }
-        }));
+
+        Consumer<Boolean> runPrediction = isAsymmetricModel -> {
+            MLInput mlInput = createMLTextInput(null, inputText, isAsymmetricModel ? mlAlgoParams : null);
+            mlClient.predict(modelId, mlInput, ActionListener.wrap(mlOutput -> {
+                final List<Map<String, ?>> result = buildMapResultFromResponse(mlOutput);
+                listener.onResponse(result);
+            }, e -> {
+                if (RetryUtil.shouldRetry(e, retryTime)) {
+                    final int retryTimeAdd = retryTime + 1;
+                    retryableInferenceSentencesWithMapResult(modelId, inputText, mlAlgoParams, retryTimeAdd, listener);
+                } else {
+                    listener.onFailure(e);
+                }
+            }));
+        };
+
+        checkModelAsymmetryAndThenPredict(modelId, listener::onFailure, runPrediction);
     }
 
     private void retryableInferenceSentencesWithVectorResult(
         final List<String> targetResponseFilters,
         final String modelId,
         final List<String> inputText,
+        final MLAlgoParams mlAlgoParams,
         final int retryTime,
         final ActionListener<List<List<Float>>> listener
     ) {
-        MLInput mlInput = createMLTextInput(targetResponseFilters, inputText);
-        mlClient.predict(modelId, mlInput, ActionListener.wrap(mlOutput -> {
-            final List<List<Float>> vector = buildVectorFromResponse(mlOutput);
-            listener.onResponse(vector);
-        }, e -> {
-            if (RetryUtil.shouldRetry(e, retryTime)) {
-                final int retryTimeAdd = retryTime + 1;
-                retryableInferenceSentencesWithVectorResult(targetResponseFilters, modelId, inputText, retryTimeAdd, listener);
-            } else {
-                listener.onFailure(e);
+
+        Consumer<Boolean> runPrediction = isAsymmetricModel -> {
+            MLInput mlInput = createMLTextInput(targetResponseFilters, inputText, isAsymmetricModel ? mlAlgoParams : null);
+            mlClient.predict(modelId, mlInput, ActionListener.wrap(mlOutput -> {
+                final List<List<Float>> vector = buildVectorFromResponse(mlOutput);
+                listener.onResponse(vector);
+            }, e -> {
+                if (RetryUtil.shouldRetry(e, retryTime)) {
+                    final int retryTimeAdd = retryTime + 1;
+                    retryableInferenceSentencesWithVectorResult(
+                        targetResponseFilters,
+                        modelId,
+                        inputText,
+                        mlAlgoParams,
+                        retryTimeAdd,
+                        listener
+                    );
+                } else {
+                    listener.onFailure(e);
+                }
+            }));
+        };
+
+        checkModelAsymmetryAndThenPredict(modelId, listener::onFailure, runPrediction);
+    }
+
+    /**
+     * Check if the model is asymmetric and then run the prediction. Model asymmetry is a concept
+     * that is specific to TextEmbeddingModelConfig. If the model is not a TextEmbeddingModel, then
+     * this check is not applicable.
+     *
+     * The asymmetry of a model is static for a given model. To avoid repeated checks for the same
+     * model, we cache the model asymmetry status. Non-TextEmbeddingModels are cached as false.
+     *
+     * @param modelId    The model id to check
+     * @param onFailure The action to take if the model cannot be retrieved
+     * @param runPrediction The action to take if the model is successfully retrieved
+     */
+    private void checkModelAsymmetryAndThenPredict(String modelId, Consumer<Exception> onFailure, Consumer<Boolean> runPrediction) {
+        CheckedConsumer<MLModel, ? extends Exception> checkModelAsymmetryListener = model -> {
+            MLModelConfig modelConfig = model.getModelConfig();
+            if (!(modelConfig instanceof TextEmbeddingModelConfig)) {
+                modelAsymmetryCache.putIfAbsent(modelId, false);
+                return;
             }
-        }));
+            final TextEmbeddingModelConfig textEmbeddingModelConfig = (TextEmbeddingModelConfig) modelConfig;
+            final boolean isAsymmetricModel = textEmbeddingModelConfig.getPassagePrefix() != null
+                || textEmbeddingModelConfig.getQueryPrefix() != null;
+            modelAsymmetryCache.putIfAbsent(modelId, isAsymmetricModel);
+        };
+        if (modelAsymmetryCache.containsKey(modelId)) {
+            runPrediction.accept(modelAsymmetryCache.get(modelId));
+        } else {
+            mlClient.getModel(modelId, ActionListener.<MLModel>wrap(mlModel -> {
+                checkModelAsymmetryListener.accept(mlModel);
+                runPrediction.accept(modelAsymmetryCache.get(modelId));
+            }, onFailure));
+        }
     }
 
     private void retryableInferenceSimilarityWithVectorResult(
@@ -213,10 +382,10 @@ public class MLCommonsClientAccessor {
         }));
     }
 
-    private MLInput createMLTextInput(final List<String> targetResponseFilters, List<String> inputText) {
+    private MLInput createMLTextInput(final List<String> targetResponseFilters, List<String> inputText, MLAlgoParams mlAlgoParams) {
         final ModelResultFilter modelResultFilter = new ModelResultFilter(false, true, targetResponseFilters, null);
         final MLInputDataset inputDataset = new TextDocsInputDataSet(inputText, modelResultFilter);
-        return new MLInput(FunctionName.TEXT_EMBEDDING, null, inputDataset);
+        return new MLInput(FunctionName.TEXT_EMBEDDING, mlAlgoParams, inputDataset);
     }
 
     private MLInput createMLTextPairsInput(final String query, final List<String> inputText) {
@@ -264,25 +433,42 @@ public class MLCommonsClientAccessor {
         final List<String> targetResponseFilters,
         final String modelId,
         final Map<String, String> inputObjects,
+        final MLAlgoParams mlAlgoParams,
         final int retryTime,
         final ActionListener<List<Float>> listener
     ) {
-        MLInput mlInput = createMLMultimodalInput(targetResponseFilters, inputObjects);
-        mlClient.predict(modelId, mlInput, ActionListener.wrap(mlOutput -> {
-            final List<Float> vector = buildSingleVectorFromResponse(mlOutput);
-            log.debug("Inference Response for input sentence is : {} ", vector);
-            listener.onResponse(vector);
-        }, e -> {
-            if (RetryUtil.shouldRetry(e, retryTime)) {
-                final int retryTimeAdd = retryTime + 1;
-                retryableInferenceSentencesWithSingleVectorResult(targetResponseFilters, modelId, inputObjects, retryTimeAdd, listener);
-            } else {
-                listener.onFailure(e);
-            }
-        }));
+
+        Consumer<Boolean> predictConsumer = isAsymmetricModel -> {
+            MLInput mlInput = createMLMultimodalInput(targetResponseFilters, inputObjects, isAsymmetricModel ? mlAlgoParams : null);
+            mlClient.predict(modelId, mlInput, ActionListener.wrap(mlOutput -> {
+                final List<Float> vector = buildSingleVectorFromResponse(mlOutput);
+                log.debug("Inference Response for input sentence is : {} ", vector);
+                listener.onResponse(vector);
+            }, e -> {
+                if (RetryUtil.shouldRetry(e, retryTime)) {
+                    final int retryTimeAdd = retryTime + 1;
+                    retryableInferenceSentencesWithSingleVectorResult(
+                        targetResponseFilters,
+                        modelId,
+                        inputObjects,
+                        mlAlgoParams,
+                        retryTimeAdd,
+                        listener
+                    );
+                } else {
+                    listener.onFailure(e);
+                }
+            }));
+        };
+
+        checkModelAsymmetryAndThenPredict(modelId, listener::onFailure, predictConsumer);
     }
 
-    private MLInput createMLMultimodalInput(final List<String> targetResponseFilters, final Map<String, String> input) {
+    private MLInput createMLMultimodalInput(
+        final List<String> targetResponseFilters,
+        final Map<String, String> input,
+        MLAlgoParams mlAlgoParams
+    ) {
         List<String> inputText = new ArrayList<>();
         inputText.add(input.get(INPUT_TEXT));
         if (input.containsKey(INPUT_IMAGE)) {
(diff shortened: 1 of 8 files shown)


navneet1v:
@br3no can you add an entry in the changelog.

navneet1v:
@br3no Thanks for raising the PR. I am wondering do we require this change? In MLCommons repository a generic MLInference processor is getting launched which is supposed to do the inference of any kind of model both during ingestion and search. RFC: https://github.com/opensearch-project/ml-commons/issues/2173

If that gets launched do we really need this feature?

br3no:
@navneet1v I have been loosely following the discussions in the mentioned RFC. It's a large change that I don't expect to be stable soon – the PR is very much in flux. Also, I don't see the use-case of asymmetric embedding models being addressed.

This PR here is much smaller in comparison and is not in any way in conflict with the RFC work. If once the work on the ML Inference Processors is finished and the use-case is addressed there as well, we can deprecate and eventually remove the functionality again.

Until then, this PR offers users the chance to use more modern local embeddings. I'm eager to put this to spin, tbh.

navneet1v:
> Also, I don't see the use-case of asymmetric embedding models being addressed.

If that is the case I would recommend posting the same on the RFC to ensure that your use case is handled.


On the other hand, I do agree this is an interesting feature. I would like to get some eyes on this change mainly in terms of should this be added or not given a more generic processor is around the corner. As I am of my opinion is concerned the main reason of generic processor was to avoid creating new/updating processors to support new model types which is happening in this PR.

Thoughts? @jmazanec15 , @martin-gaievski , @vamshin , @vibrantvarun . 

Let me add some PMs too for Opensearch-project to know their thoughts. @dylan-tong-aws

br3no:
@navneet1v I have added a comment earlier today to the RFC (cf. https://github.com/opensearch-project/ml-commons/issues/2173#issuecomment-2078796787).

Sure, let's open the discussion and get some PMs into it. 

I really don't mind leaving this out **if** the support is introduced in another PR in 2.14. I'm concerned https://github.com/opensearch-project/ml-commons/issues/2173 is a much larger effort, that won't be ready that quickly...

It's not about my contribution – I need the feature. 🙃

navneet1v:
> I really don't mind leaving this out if the support is introduced in another PR in 2.14. I'm concerned https://github.com/opensearch-project/ml-commons/issues/2173 is a much larger effort, that won't be ready that quickly...

I can see the feature is marked for 2.14 release of Opensearch. Let me add maintainers from ML team too. @mingshl , @ylwu-amzn

br3no:
@mingshl @ylwu-amzn, I'd really like to have this feature in 2.14. 

Do you think this use-case will be fully supported with https://github.com/opensearch-project/ml-commons/issues/2173? Cf. https://github.com/opensearch-project/ml-commons/issues/2173#issuecomment-2078796787

If not, I'd be happy to help this PR get merged as an interim solution! Let me know what you think!

mingshl:
@br3no ml inference processor is targeting at first supporting remote model only. How did you usually connect this model? is it local or remote? 

if remote, can you please provide a SageMaker deployment code piece then I can quickly test it in 2.14 test cluster. Thanks

br3no:
@mingshl sorry for taking so long to answer!

The use-case **for now** is to use a local, asymmetric model such as https://huggingface.co/intfloat/multilingual-e5-small. 

This PR here is the last puzzle piece to allow one to use these kinds of model and should in principle also work with remote models. It makes sure that the neural-search plugin uses the correct inference parameters when embedding passages and queries with asymmetric models. Regardless of whether the model is local or remote, if you are using asymmetric models, you will need to provide this information anyway.

The thing is that asymmetric models need to know at inference time **what** exactly they are embedding. OpenSearch currently treats embedding models as symmetric, meaning that regardless of whether the text being embedded is a query or a passage, the embedding will be always the same. Asymmetric models require content "hints" to the text being embedded; the model exemplified above uses the string prefixes `passage: ` and `query: `. These models perform better than similarly sized symmetric models.

In https://github.com/opensearch-project/ml-commons/issues/1799 we have added the concept of asymmetric models into ml-commons, introducing the `AsymmetricTextEmbeddingParameters` class, used at inference time to signal if the text being embedded is a query or a passage. So this PR is only using this new infrastructure.

I would really be happy to get this merged as an interim solution until the ml inference processor fully supports this use-case.

reuschling:
I also vote for this PR in need for this functionality.

navneet1v:
@br3no will it possible if you can contribute back in MLInference processor for local model support? Is that even an option?

navneet1v:
> I'd propose then to merge this PR now and then start the work to eventually replace this once the MLInference processor supports this use case...

The problem is once this is released it cannot be deprecated till a major version release. Hence I am bit hesitant to have this feature in neural search plugin.

ylwu-amzn:
> This PR has a very small surface. It doesn't change any APIs. So I believe this is not something to worry about, actually. Once the MLInferece processor supports asymmetric models the neural-search plugin can be changed to start using it instead of what I built here. This would be not a breaking change, only an internal implementation detail.

Thanks @br3no , Can you show some examples of neural search query with asymmetric embedding model?  ML inference processor can support asymmetric embedding model. We are working on unifying the interface so we can also support local model in ML inference processor.  I think @navneet1v concern is valid. We should consider deprecation effort. Let's check how user experience looks like with this PR.

br3no:
@ylwu-amzn as I said in the comment above, there is no API change in this PR. You would use the neural-search plugin in the same way it is used today.

**What is implemented here?**
The main work is done in `MLCommonsClientAccessor`. Whenever neural-search is embedding text (either in the `TextEmbeddingProcessor` or in the `NeuralQueryBuilder`) the plugin will ensure that **if** the embedding model is asymmetric, the parameters will be passed on correctly. More details can be found in the description and code.

ylwu-amzn:
@br3no , sorry that I didn't have enough time to read the details. Took a quick look, seems the code will specify the "QUERY" type when run query, and set  "PASSAGE" type when ingest. That seems no API change, cx can use same ingest pipeline and neural search, switching to asymmetric model is seamless. I think this is a good design, so it can support BWC and seems can be deprecated/migrated together with current text embedding APIs.   I'm good for merge this.

br3no:
Exactly! Great!

br3no:
@ylwu-amzn could you drive the review process forward then?

ylwu-amzn:
> @ylwu-amzn could you drive the review process forward then?

Sorry , missed your comment. Asking neural-search maintainers if they have other concerns

src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
```diff
@@ -161,6 +196,62 @@ public void testInferenceSentences_whenNotConnectionException_thenNoRetry() {
...
+            .predict(
+                Mockito.eq(TestCommonConstants.MODEL_ID),
+                Mockito.argThat((MLInput input) -> input.getParameters() != null),
+                Mockito.isA(ActionListener.class)
+            );
+        Mockito.verify(singleSentenceResultListener).onResponse(vector);
+        Mockito.verifyNoMoreInteractions(singleSentenceResultListener);
+    }
+
+    public void testInferenceSentences_whenGetModelException_thenFailure() {
```
martin-gaievski, line 225: we need to test scenario when we're retrying 1-2 times. I see scenario when first request has failed with error that isn't retryable, this isn't a full coverage

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -54,7 +63,29 @@ public void inferenceSentence(
...
+     * {@link org.opensearch.ml.common.input.parameter.textembedding.AsymmetricTextEmbeddingParameters} is
+     * mandatory. This method will check whether the model being used is asymmetric and correctly handle the
+     * parameter, so it's okay to always pass the parameter (even if the model is symmetric).
+     *
+     * @param modelId {@link String}
+     * @param inputText {@link List} of {@link String} on which inference needs to happen
+     * @param mlAlgoParams {@link MLAlgoParams} which will be used to run the inference
+     * @param listener {@link ActionListener} which will be called when prediction is completed or errored out
+     */
+    public void inferenceSentence(
```
martin-gaievski, line 82: We need to refactor this method to accept single POJO object that can be built with builder pattern, similar to ml-commons class with algorithm params. Reason - we cannot add new method every time we need to add new parameter to inference ml client.
I'm fine if we do it in a separate PR, please create GH issue and post link here.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -40,6 +48,7 @@
 public class MLCommonsClientAccessor {
     private static final List<String> TARGET_RESPONSE_FILTERS = List.of("sentence_embedding");
     private final MachineLearningNodeClient mlClient;
+    private final Map<String, Boolean> modelAsymmetryCache = new ConcurrentHashMap<>();
```
martin-gaievski, line 51: please add comment around this cache behavior and usage. You can start from following:
- it's local for the data node
- how we invalidate and evict if we ever going to do this
- if cache miss, what is the latency to do retrieve value via API call and put it to cache
- how big is a single object, that also implies some eviction strategy as we cannot grow indefinitely 
- what is the behavior is case of node restart/drop/model got redeployed

src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
br3no: @martin-gaievski thanks for pointing this out.

I'm wondering though, if we should make this retryable at all. Let me elaborate:

In my understanding, inference requests are retried because they tend to fail more often than regular operations in OpenSearch. I don't know the history and complete reasoning behind this, so I speculate it has to do with the fact that the inference is done natively and that many things can go wrong there.

With my change, if fetching the model information fails (`mlClient.getModel(modelId, ...`), there is no retry. Model information is fetched the first time inference is requested with a particular model. After that, the result is cached and the method behaves exactly as before the PR.

So my argument is: should we really add a retry logic to this relatively simple operation? If `getModel` fails, it is most likely to fail again, so retrying wouldn't make sense. If so, one could argue that all operations in OpenSearch should be wrapped in a retry logic.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
br3no: Okay. Cf. https://github.com/opensearch-project/neural-search/issues/790

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
br3no: * Yes, it is local for the data node.
* We are never evicting entries.
* The latency is that of fetching the model configuration and parsing it.
* The cache will map `modelId` to `Boolean`; the size requirement is 20 chars + 1 Boolean per entry.
* If the node restarts the cache will be empty; the first inference request will lead to the cache entry being populated
* if the node drops the cluster, the inference request will fail. The cache will not be emptied.
* if the model gets redeployed the user will need to request inference with a new `modelId`; which will lead to a new cache entry. The old one will continue there.

The only scenario I can think of where this design could be problematic is if a malicious actor floods the cluster with billions of inference requests with inexistent models. This would lead to an increase in heap usage that would never be GC'd. 

Let me know if you think this should be changed.

reuschling:
Is there some progress or news regarding this PR? It is still a small thing without Api change which is the last puzzle piece to enable asymmetric embedding models.

br3no:
I have no news. We've moved to prefixing the strings outside of OpenSearch.

reuschling:
@ylwu-amzn, @martin-gaievski, are there some things left that speaks against merging this functionality? It is a blocker right now, and it is a marginal change only that doesn't influence other functionalities.

model-collapse:
I am going to pick this PR up right now, @br3no @reuschling sorry for the late reply. Let me check your code first.

zane-neo:
@br3no I read through the context and I believe this is a valid use case and this PR provides user the flexibility to integrate the feature, thank you very much! But current the PR needs rebase with master and in case you're not monitoring the PR,  I've rebased your code with main and created this PR to your repo: https://github.com/br3no/neural-search/pull/1, please take time to check the PR, you can fix this PR or cherry-pick my commit to this PR, I would like to push this forward to make it released in 2.18 which is 5th Nov, thanks again.

br3no:
@zane-neo great to hear this!

I have merged your PR. Is there anything else I need to do?

zane-neo:
> @zane-neo great to hear this!
> 
> I have merged your PR. Is there anything else I need to do?

@br3no The main blocker for now is the current PR has conflicts needs to be resolved, so I was saying there are two approaches to fix this:
1. You fix the conflicts by rebasing the main branch.
2. You cherry-pick my commit in the PR sent to your repo to your `asymmetric-embeddings-620` branch and push the commit.
To be honest I prefer the first approach, but since you're merging my PR so you can continue with the left ones: cherry-pick my commit to the branch and push. Thanks!

br3no:
@zane-neo I rebased the changes and adapted new tests that were affected by my changes. I think this is now good to go,

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
zane-neo: @br3no It's fine since this not accessible from public API, to use huge amount of inexistent models, the malicious actor has to first got the permission to update the pipeline, and if that's true the malicious actor has much more straightforward means to attack the system.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -155,42 +271,95 @@ public void inferenceSimilarity(
...
+     * model, we cache the model asymmetry status. Non-TextEmbeddingModels are cached as false.
+     *
+     * @param modelId    The model id to check
+     * @param onFailure The action to take if the model cannot be retrieved
+     * @param runPrediction The action to take if the model is successfully retrieved
+     */
+    private void checkModelAsymmetryAndThenPredict(String modelId, Consumer<Exception> onFailure, Consumer<Boolean> runPrediction) {
+        CheckedConsumer<MLModel, ? extends Exception> checkModelAsymmetryListener = model -> {
+            MLModelConfig modelConfig = model.getModelConfig();
+            if (!(modelConfig instanceof TextEmbeddingModelConfig)) {
```
zane-neo, line 346: Add null check for modelConfig as it could be null and throws NPE

src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
zane-neo: @br3no I think Martin was talking about adding more test method to increase the test coverage, e.g. add a new test method `testInferenceSentences_whenGetModelNodeDisconnectedException_thenRetryToSuccess`, with this method you only need to change the 
```
RuntimeException exception = new RuntimeException("Bam!");
setupMocksForTextEmbeddingModelAsymmetryCheck(exception);
```
to
```
NodeDisconnectedException exception = new NodeDisconnectedException("node disconnected!");
setupMocksForTextEmbeddingModelAsymmetryCheck(exception);
```
in this method, and by that the new method should cover the retry part.

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
```diff
@@ -252,6 +257,58 @@ private void createTextEmbeddingIndex() throws Exception {
...
+            createPipelineProcessor(modelId, PIPELINE_NAME, ProcessorType.TEXT_EMBEDDING, 2);
+            createTextEmbeddingIndex();
+            ingestDocument();
+            assertEquals(1, getDocCount(INDEX_NAME));
+        } finally {
+            wipeOfTestResources(INDEX_NAME, PIPELINE_NAME, modelId, null);
+        }
+    }
+
+    private void ingestDocument() throws Exception {
```
zane-neo, line 274: Is it possible to reuse the below `ingestDocument(String, String)`? Or at least we can put this document to a json file for better readability and maintenance, WDYT?

zane-neo:
> @zane-neo I rebased the changes and adapted new tests that were affected by my changes. I think this is now good to go.

Left several comments and please take a look, other part LGTM, thanks.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
br3no: I believe the null check is not necessary, since the `instanceof` operator is null-safe. Cf. https://docs.oracle.com/javase/specs/jls/se23/html/jls-15.html#jls-15.20.2

src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
br3no: @zane-neo if I replace the exception, the behavior will not change, i.e. there will not be a successful retry. 

This is because the `checkModelAsymmetryAndThenPredict` method in `MLCommonsClientAccessor` will not retry `mlClient.getModel` if there is a failure.

The reasoning was that if there is an exception in `mlClient.getModel`, which is a very simple operation (it's just fetching the model information), then most likely retrying it will lead to the same failure. If we think this operation should be retried, then, I argued above, all operations in OpenSearch would be equally good candidates to be wrapped in a retry logic.

I'm happy to implement the retry in this operation, but I don't believe it's worth it. It will introduce more complexity to an already quite intricate code path, making it less readable. At the same time I believe it will almost never really be useful, as a failing `mlClient.getModel` will almost always fail again if repeated.

src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java
br3no: Yes, sure. Good catch.

br3no:
@zane-neo I have pushed a small refactoring of a test, as you suggested. Thanks!

I have also added comments to the issues you raised in the other threads.

src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
zane-neo: Agreed, it's fine to not retry the `client.getModel` since it's essentially different with `client.predict`, for predict request, ml-commons has internal routing logic to dispatch the request to different nodes and it's possible that the node disconnected from cluster right after received the request, then retry is useful as next predict request will be dispatched to a different node in round robin approach. But for `client.getMode` it internally uses the `client.get` operation which is more deterministic, e.g. user might specify to read the result from primary shard or read from the only one replica, retry won't help in these cases. And it's a valid argue that in OpenSearch not all the operations are wrapped with retry logic.

zane-neo (approved):

zane-neo:
> @zane-neo I have pushed a small refactoring of a test, as you suggested. Thanks!
> 
> I have also added comments to the issues you raised in the other threads.

Thanks @br3no currently it LGTM, approved the PR, I'll push for another approval and then we can merge it. Sadly it's not possible to release it in 2.18 as the release flow has been started and code has been frozen, but I'm sure we can make it released in 2.19.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -69,79 +100,164 @@ public void inferenceSentence(
...
+     * @param listener  {@link ActionListener} which will be called when prediction is completed or
+     *                  errored out
      */
     public void inferenceSentences(
         @NonNull final String modelId,
         @NonNull final List<String> inputText,
         @NonNull final ActionListener<List<List<Float>>> listener
     ) {
-        inferenceSentences(TARGET_RESPONSE_FILTERS, modelId, inputText, listener);
+        inferenceSentences(TARGET_RESPONSE_FILTERS, modelId, inputText, null, listener);
```
vibrantvarun, line 119: Why null is passed here?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
martin-gaievski: @br3no thank you for response, it all makes sense. 
I think we need to:
- use existing OpenSearch component for cache 
- set a limit on either size or time in the cache, that should lower the chance of data node going to the critical level of heap memory usage.

please take a look on these classes: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/cache/Cache.java and https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/cache/CacheBuilder.java

it has many nice things like multiple modes for eviction and stats

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
vibrantvarun: Can you create a DTO object here. And add the values which are absolutely necessary to pass. That way you can avoid sending null here.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -69,79 +100,164 @@ public void inferenceSentence(
...
+     * @param listener              {@link ActionListener} which will be called when prediction is
+     *                              completed or errored out.
+     */
+    public void inferenceSentences(
+        @NonNull final List<String> targetResponseFilters,
+        @NonNull final String modelId,
+        @NonNull final List<String> inputText,
+        @NonNull final ActionListener<List<List<Float>>> listener
+    ) {
+        inferenceSentences(targetResponseFilters, modelId, inputText, null, listener);
```
vibrantvarun, line 166: Create DTO to pass in argument.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -69,79 +100,164 @@ public void inferenceSentence(
...
      */
     public void inferenceSentences(
         @NonNull final List<String> targetResponseFilters,
         @NonNull final String modelId,
         @NonNull final List<String> inputText,
+        @Nullable final MLAlgoParams mlAlgoParams,
         @NonNull final ActionListener<List<List<Float>>> listener
     ) {
-        retryableInferenceSentencesWithVectorResult(targetResponseFilters, modelId, inputText, 0, listener);
+        retryableInferenceSentencesWithVectorResult(targetResponseFilters, modelId, inputText, mlAlgoParams, 0, listener);
```
vibrantvarun, line 194: DTO object as argument size is exceeding 5 now.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -69,79 +100,164 @@ public void inferenceSentence(
...
+     * @param listener     {@link ActionListener} which will be called when prediction is completed or
+     *                     errored out.
      */
     public void inferenceSentences(
         @NonNull final String modelId,
         @NonNull final Map<String, String> inputObjects,
         @NonNull final ActionListener<List<Float>> listener
     ) {
-        retryableInferenceSentencesWithSingleVectorResult(TARGET_RESPONSE_FILTERS, modelId, inputObjects, 0, listener);
+        inferenceSentences(modelId, inputObjects, null, listener);
```
vibrantvarun, line 222: DTO object please.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -47,10 +49,15 @@ public void doExecute(
...
         BiConsumer<IngestDocument, Exception> handler
     ) {
-        mlCommonsClientAccessor.inferenceSentences(this.modelId, inferenceList, ActionListener.wrap(vectors -> {
-            setVectorFieldsToDocument(ingestDocument, ProcessMap, vectors);
-            handler.accept(ingestDocument, null);
-        }, e -> { handler.accept(null, e); }));
+        mlCommonsClientAccessor.inferenceSentences(
+            this.modelId,
+            inferenceList,
+            AsymmetricTextEmbeddingParameters.builder().embeddingContentType(EmbeddingContentType.PASSAGE).build(),
```
martin-gaievski, line 55: this can be pulled to a constant level

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -333,10 +335,15 @@ protected QueryBuilder doRewrite(QueryRewriteContext queryRewriteContext) {
...
         }
         queryRewriteContext.registerAsyncAction(
-            ((client, actionListener) -> ML_CLIENT.inferenceSentences(modelId(), inferenceInput, ActionListener.wrap(floatList -> {
-                vectorSetOnce.set(vectorAsListToArray(floatList));
-                actionListener.onResponse(null);
-            }, actionListener::onFailure)))
+            ((client, actionListener) -> ML_CLIENT.inferenceSentences(
+                modelId(),
+                inferenceInput,
+                AsymmetricTextEmbeddingParameters.builder().embeddingContentType(EmbeddingContentType.QUERY).build(),
```
martin-gaievski, line 341: this is effectively a constant, no need to compute on every call

vibrantvarun, line 341: nit: always add new parameters in the end of argument list so that it is easy to understand the changes.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
vibrantvarun: For reference check CombineScoreDTO class

src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
martin-gaievski: @zane-neo @br3no please help me to understand the flow for reading model config. Is there a chance that this call is remote, meaning going outside the opensearch cluster if model deployed remotely and connected via ml-commons remote connector? 

If that's not the case then I'm fine with logic without retry, it's a transport request within single OS domain. In this case where the mode configuration is stored, is it in ml-commons memory or cluster metadata?

If it can be a remote call then we need a retry. Doesn't matter how simple operation is, there can be a transient network issue or anything of this sort. In this case it's different from most of OS APIs, those are primality case 1, transport requests inside cluster.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
yuye-aws: > Can you create a DTO object here. And add the values which are absolutely necessary to pass. That way you can avoid sending null here.

Let's do it in the next PR.

yuye-aws:
Hi @vibrantvarun ! Since this PR is targeted towards 2.19, we still have time to raise next PRs before code freeze date of 2.19 (Jan 28, 2025). I think we can leave the BWC tests for the next PR.

@br3no Can you discuss with @vibrantvarun on whether a DTO object is needed for this PR? I understand DTO object is a best practice, but I don't want it takes too long before merging this PR.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
zane-neo: @vibrantvarun This is been tracked in this issue: https://github.com/opensearch-project/neural-search/issues/790

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
zane-neo: https://github.com/opensearch-project/neural-search/issues/790

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
zane-neo: https://github.com/opensearch-project/neural-search/issues/790

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
zane-neo: same above

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
zane-neo: I'm open to make the change but I don't think it's a mandatory, the cache is to store the model info and it only consumes very little space comparing to a model, so the heap memory could reach throttle when registering a new model instead of writing the little metadata into this cache. The only case is there's a way to bypass the model registration and put huge amount of info to the cache directly which is not a case as above explanation.

src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
zane-neo: No case is to fetching model metadata from remote services, cause they are all stored in OS cluster.

zane-neo (approved):

zane-neo:
> @br3no How are we ensuring BWC test compatibility?

This is a valid concern, I took a look on the code and it's could have BWC issue during seder. The latest version serializes the [AsymmetricTextEmbeddingParameters](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/input/parameter/textembedding/AsymmetricTextEmbeddingParameters.java#L98), and if a node is deployed with a legacy version OS and when [deserialization](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/input/MLInput.java#L119), it fetches the class from the internal cache map [here](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/MLCommonsClassLoader.java#L236), and since a legacy version doesn't have this class, the IllegalArgumentException will be thrown. @br3no Can you do a test on this case to double confirm if this is true? Thanks.

martin-gaievski:
@zane-neo for some reason github ui glitching for me so I cannot put my answer under your comments

1. replacing map with cache object - it still worth it before we don't have many references to hashmap, besides API for our cache is very simple
2. nature of calls for requesting model metadata - that's good then, we can move forward without adding a retry logic then

br3no:
I'll try to find time to work on this tomorrow. I'll replace the map with the OS cache and I'll try to have a look at the BWC tests.

Regarding the DTO, there's already #790 opened as a follow-up task. @yuye-aws, is this fine by you?

vibrantvarun:
@br3no DTO object helps to make look code more readable and understandable.  Issue 790 is currently not prioritized and is in backlog. We don't know when will that issue will be prioritized. Therefore, I would request you to add it in this PR. It will not be a design level change or something completely changes the purpose of this PR.

cc: @martin-gaievski @yuye-aws @zhichao-aws

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
yuye-aws: Agreeing with @zane-neo , `modelAsymmetryCache` is just a map from model_id to boolean. The size is very limited. @br3no I think you just need to add some comments to make the code more readable.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -155,42 +271,95 @@ public void inferenceSimilarity(
...
+            MLModelConfig modelConfig = model.getModelConfig();
+            if (!(modelConfig instanceof TextEmbeddingModelConfig)) {
+                modelAsymmetryCache.putIfAbsent(modelId, false);
+                return;
             }
-        }));
+            final TextEmbeddingModelConfig textEmbeddingModelConfig = (TextEmbeddingModelConfig) modelConfig;
+            final boolean isAsymmetricModel = textEmbeddingModelConfig.getPassagePrefix() != null
+                || textEmbeddingModelConfig.getQueryPrefix() != null;
+            modelAsymmetryCache.putIfAbsent(modelId, isAsymmetricModel);
```
yuye-aws, line 353: Can you simplify with
```
MLModelConfig modelConfig = model.getModelConfig();
if (!(modelConfig instanceof TextEmbeddingModelConfig textEmbeddingModelConfig)) {
    modelAsymmetryCache.putIfAbsent(modelId, false);
    return;
}
final boolean isAsymmetricModel = textEmbeddingModelConfig.getPassagePrefix() != null
    || textEmbeddingModelConfig.getQueryPrefix() != null;
modelAsymmetryCache.putIfAbsent(modelId, isAsymmetricModel);
```

src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
```diff
@@ -59,15 +63,34 @@ public void testInferenceSentence_whenValidInput_thenSuccess() {
             actionListener.onResponse(createModelTensorOutput(TestCommonConstants.PREDICT_VECTOR_ARRAY));
             return null;
         }).when(client).predict(Mockito.eq(TestCommonConstants.MODEL_ID), Mockito.isA(MLInput.class), Mockito.isA(ActionListener.class));
+        setupMocksForTextEmbeddingModelAsymmetryCheck(false);
 
-        accessor.inferenceSentence(TestCommonConstants.MODEL_ID, TestCommonConstants.SENTENCES_LIST.get(0), singleSentenceResultListener);
+        accessor.inferenceSentence(
```
yuye-aws, line 68: I understand you are not using the old `inferenceSentence` method. Can you double check whether this method is still need? We don't wish to see redundant function.

yuye-aws (commented):
Excellent code! Just a few comments

yuye-aws:
> I'll try to find time to work on this tomorrow. I'll replace the map with the OS cache and I'll try to have a look at the BWC tests.
> 
> Regarding the DTO, there's already #790 opened as a follow-up task. @yuye-aws, is this fine by you?

It's fine for me as long as the DTO gets merged before the code freeze date of the **2.19**.

yuye-aws:
> @br3no DTO object helps to make look code more readable and understandable. Issue 790 is currently not prioritized and is in backlog. We don't know when will that issue will be prioritized. Therefore, I would request you to add it in this PR. It will not be a design level change or something completely changes the purpose of this PR.
> 
> cc: @martin-gaievski @yuye-aws @zhichao-aws

Let's label the issue 790 with 2.19 tag. The DTO is a small fix. If @br3no is too busy to create a PR, I can help him instead. As @reuschling has stated, this PR is the last puzzle for the asymmetric embedding feature. I don't want it be too long before we gets it merged.

yuye-aws:
> @br3no @yuye-aws some of my comments still are not addressed, the major one is switching from hashmap to existing core class for cache. in regards to comments for using unbounded map - I do not agree with mentioned arguments. We need to have a limit of that cache size because the lifespan of the cluster may be long (e.g. in a managed OS service) and instance type for data node can be tiny, like t2 with 1Gb. Switching to existing cache component that provides that limit and keeps system stable is close to zero effort, why not to do this.

Seems that the [cache class](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/cache/Cache.java) will evict outdated elements with LRU policy. @br3no Do you think it's fine to deprecate some model cache?

br3no:
Sorry folks, I didn't have the time to look into the latest comments yet. It's been busy weeks. I'll try to have a look today and address all remaining points.

br3no:
I have addressed all open issues with the exception of the BWC test. I'll try to have a look in the coming days.

yuye-aws:
Pinging @martin-gaievski to review the cache.

yuye-aws:
> I have addressed all open issues with the exception of the BWC test. I'll try to have a look in the coming days.

For the BWC test, you just need to create two classes for `TextEmbeddingProcessorIT.java`. One for rolling upgrade and another for restart upgrade:
1. qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/
2. qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/TextChunkingProcessorIT.java

br3no:
@yuye-aws aren't the test cases covered completely by the `HybridSearchIT` classes? There the text embedding processor is created, documents are indexed and a hybrid query is executed.

br3no:
I think I will need some assistance in understanding what to do exactly regarding the BWC tests. I'm not even sure I understand where to look for errors.

The error scenario you folks are seeing is:
* a cluster is updated
* a query is sent to a node running an older version of OS
* the query contains an `AsymmetricTextEmbeddingParameters`
* the legacy version cannot deserialize this class

Did I get this right? Or is it another scenario you are concerned with?

yuye-aws:
> The class `AsymmetricTextEmbeddingParameters` has been part of ml-commons for about 3 months. This would then be a problem for all versions older than 2.14.

You can exclude versions older than 2.14 just like [here](https://github.com/opensearch-project/neural-search/blob/0316dc41c639e14071125ab91165d62a1cd6899f/qa/rolling-upgrade/build.gradle#L288-L294).

yuye-aws:
> @yuye-aws aren't the test cases covered completely by the `HybridSearchIT` classes? There the text embedding processor is created, documents are indexed and a hybrid query is executed.

Do you mean the BWC tests for `HybridSearchIT`? If so, you can fill in the gap between BWC tests and integrations tests for `TextEmbeddingProcessorIT`.

br3no:
@yuye-aws 

What I meant with [this comment](https://github.com/opensearch-project/neural-search/pull/710#issuecomment-2460163754) is that I don't see the need to implement a different BWC Test class for the change in this PR. The existing BWC tests (e.g. the `HybridQueryIT`, but others as well) test the complete process of creating an index, creating a pipeline for document embedding and issuing neural queries. There is no new process introduced with this PR that requires different BWC test code, from my point of view. Please correct me if I'm wrong.

BTW the rolling-upgrade BWC tests are failing on the main branch.

I ran 
```
./gradlew :qa:rolling-upgrade:testRollingUpgrade -D'tests.bwc.version=2.18.0-SNAPSHOT'
```
without success.

This makes it hard to add new tests.

yuye-aws:
> ./gradlew :qa:rolling-upgrade:testRollingUpgrade

@vibrantvarun The comment makes sense to me. Can you check whether BWC tests are needed and help him?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -38,23 +46,181 @@
...
+        }
+
+        public String getQueryText() {
+            return queryText;
+        }
+
+        /**
+         * Builder for {@link InferenceRequest}. Supports fluent construction of the request object.
+         */
+        public static class Builder {
```
martin-gaievski, line 107: Lots of this builder code can be collapsed if we use Builder annotation from Lombok. Something like this:

```
@Getter
@Builder
public class InferenceRequest {

    private final String modelId; // required
    @Singular("inputText")
    private List<String> inputTexts;
    private MLAlgoParams mlAlgoParams;
    private List<String> targetResponseFilters;
    private Map<String, String> inputObjects;
    private String queryText;

    private InferenceRequest(String modelId, List<String> inputTexts, MLAlgoParams mlAlgoParams, 
                             List<String> targetResponseFilters, Map<String, String> inputObjects, String queryText) {
        this.modelId = modelId;
        this.inputTexts = inputTexts != null ? inputTexts : new ArrayList<>();
        this.mlAlgoParams = mlAlgoParams;
        this.targetResponseFilters = targetResponseFilters;
        this.inputObjects = inputObjects;
        this.queryText = queryText;
    }

    public void addInputText(String inputText) {
        if (this.inputTexts == null) {
            this.inputTexts = new ArrayList<>();
        }
        this.inputTexts.add(inputText);
    }
}
```

and call it in the similar way:
```
InferenceRequest request2 = InferenceRequest.builder()
        .modelId("model123")
        .inputText("Sample text 1")
        .inputText("Sample text 2")
        .inputText("Sample text 3")
        .build();
```

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -38,23 +46,181 @@
...
+             */
+            public InferenceRequest build() {
+                return new InferenceRequest(modelId, inputTexts, mlAlgoParams, targetResponseFilters, inputObjects, queryText);
+            }
+
+        }
+    }
+
     private final MachineLearningNodeClient mlClient;
+    private final Cache<String, Boolean> modelAsymmetryCache = CacheBuilder.<String, Boolean>builder().setMaximumWeight(10_000).build();
```
martin-gaievski, line 196: please pull the max weight number to a constant, it would be easy to find/change if needed

br3no:
@martin-gaievski I have addressed all your latest comments. Hope this PR can now be approved.

martin-gaievski (approved):
It looks good to me, thank you

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -264,32 +403,49 @@ private void retryableInferenceSentencesWithSingleVectorResult(
...
-            if (RetryUtil.shouldRetry(e, retryTime)) {
-                final int retryTimeAdd = retryTime + 1;
-                retryableInferenceSentencesWithSingleVectorResult(targetResponseFilters, modelId, inputObjects, retryTimeAdd, listener);
-            } else {
-                listener.onFailure(e);
-            }
-        }));
+
+        Consumer<Boolean> predictConsumer = isAsymmetricModel -> {
+            MLInput mlInput = createMLMultimodalInput(targetResponseFilters, inputObjects, isAsymmetricModel ? mlAlgoParams : null);
```
heemin32, line 412: There will be a case where we might want to pass mlAlgoParams for SymmetricModel in the future. Shouldn't we check if the model is asymmetric or not before constructing the request?
```
// Check here if model is symmetric or asymmetric
InferenceRequest.builder().modelId(this.modelId).inputTexts(inferenceList).mlAlgoParams(PASSAGE_PARAMETERS).build(),
```

Modifying the request internally will lead to a confusion later when we need to pass mlAlgoParams to symmetric model but it silently omit it before calling the model.

zane-neo:
> > @br3no How are we ensuring BWC test compatibility?
> 
> This is a valid concern, I took a look on the code and it's could have BWC issue during seder. The latest version serializes the [AsymmetricTextEmbeddingParameters](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/input/parameter/textembedding/AsymmetricTextEmbeddingParameters.java#L98), and if a node is deployed with a legacy version OS and when [deserialization](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/input/MLInput.java#L119), it fetches the class from the internal cache map [here](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/MLCommonsClassLoader.java#L236), and since a legacy version doesn't have this class, the IllegalArgumentException will be thrown. @br3no Can you do a test on this case to double confirm if this is true? Thanks.

@br3no Please take a look on this comment, the seder is the only risk I see between old nodes and new nodes, and I don't think HybridSearchIT has covered asymmetric case as it only tested with text embedding model(symmetric model), they don't have the new configuration introduced in asymmetric model so no seder issue either. I would suggest you create a old cluster with two nodes(ml-node) and replace one nodes with latest code, and test two cases:
1. Send request to old node and make the old node dispatch the request to new node.
2. Send request to new node and make the new node dispatch request to old node.
You don't need to manually configure anything to enable the dispatch as ml-commons automatically dispatch the requests to different nodes in a round-robin fashion, by controlling which node receives the request and trigger request twice, one of the request will be dispatched to another node. If you don't see any seder exception then it's good to merge the PR, thanks.

br3no:
@zane-neo so if I get this right, I should create a new test that uses the asymmetric model feature. This test should only run for OS versions >= 2.19. Is this right?

br3no:
Ping.

zane-neo:
> @zane-neo so if I get this right, I should create a new test that uses the asymmetric model feature. This test should only run for OS versions >= 2.19. Is this right?
> 
> Your concern is about making sure future releases will not break compatibility with this feature.

@br3no That's right to use the asymmetric model feature for OS>=2.19, but the thing is a little bit different. you should test a mixed cluster with OS 2.19 and OS 2.18, since 2.18 has different code base of neural search, so a request to 2.19 node and then being dispatched to 2.18 node could encounter seder issues. My guess is you can test on this and based on the result:
1. No seder error, you can add BWC tests with OS >= 2.19
2. Has seder error, fix that and add BWC tests with OS >= 2.19

martin-gaievski:
btw, if you're working on BWC for 2.19+ and want to check how the results are, better rebase on latest main, we've switched 2.18-snapshot to 2.18/2.19-snapshot. This branch will keep failing on running BWCs

martin-gaievski:
@br3no the CI is in good shape now, you can work on BWC tests.

zane-neo:
> > @zane-neo so if I get this right, I should create a new test that uses the asymmetric model feature. This test should only run for OS versions >= 2.19. Is this right?
> > Your concern is about making sure future releases will not break compatibility with this feature.
> 
> @br3no That's right to use the asymmetric model feature for OS>=2.19, but the thing is a little bit different. you should test a mixed cluster with OS 2.19 and OS 2.18, since 2.18 has different code base of neural search, so a request to 2.19 node and then being dispatched to 2.18 node could encounter seder issues. My guess is you can test on this and based on the result:
> 
> 1. No seder error, you can add BWC tests with OS >= 2.19
> 2. Has seder error, fix that and add BWC tests with OS >= 2.19

@br3no I tested on this and I didn't find seder issues for OS version >= 2.13 so there's no concern on this, you can work on creating BWC tests for OS version >= 2.19, thanks.

br3no:
Sorry folks, I didn't have time lately to invest here. I'll try to do this in the holiday season.

mingshl:
> Sorry folks, I didn't have time lately to invest here. I'll try to do this in the holiday season.

Hi @br3no,  @brianf-aws is planning on adding a tutorial for using asymmetric local model with ml inference processors during ingest and search. https://github.com/opensearch-project/ml-commons/pull/3258 you can watch for the new updates and post comments through the pr.

br3no:
@mingshl does this make this PR here obsolete?

brianf-aws:
> @mingshl does this make this PR here obsolete?

Hey Breno, I did researching and thanks to your contribution (Asymmetric Embedding support) within ML-Commons its possible to get embeddings on demand using the ML Inference processor (MLinferenceSearchProcessor & MLInferenceIngestProcessor^). I didn't read fully through the comment thread but I believe what you are going after can be done. 

Hopefully the tutorial (https://github.com/opensearch-project/ml-commons/pull/3258) being created matches your use case. If not can you let us know what might be missing so we can understand better? 

^There is a bit of a caveat, (There is a slight code change required to allow Asymmetric Embeddings with MLInferenceIngest Processor please see here https://github.com/opensearch-project/ml-commons/pull/3281).

yuye-aws:
Hi @br3no! Can you update the conflicting files?

heemin32:
I believe this feature should be included in the text embedding processor, even though users could achieve a similar outcome using the ML processor. This is because the feature is closely tied to the text embedding processor. The ML processor is designed for more general use cases and might be more complex to utilize compared to the text embedding processor. That said, I would like my comments to be addressed before this PR is merged. https://github.com/opensearch-project/neural-search/pull/710#discussion_r1835282405

martin-gaievski:
@mingshl can you please answer this comment https://github.com/opensearch-project/neural-search/pull/710#issuecomment-2565629431, not sure if we need to keep this PR alive

ylwu-amzn:
Sounds not bad to support this in text embedding processor, should be minor effort since this PR is specifically for that purpose

martin-gaievski:
@ryanbogan please start working on this implementation

PR #711, opened 2024-04-26
Removed map of subquery to subquery index in favor of storing index as part of DISI wrapper to improve hybrid query latencies by 20%

martin-gaievski:
In this PR we're continue to improve latency of hybrid query as part of meta issue https://github.com/opensearch-project/neural-search/issues/705. 

Based on following flamegraph next area that may give high boost is a lookup of sub-query index by the query. That is needed to get index of sub-query and store score of that sub-query for one document ([code ref](https://github.com/opensearch-project/neural-search/blob/2.13/src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java#L204-L224)).

profile_hybqr_15_04_no_concurrent

Most time (~28%) is taken by store and lookup of the index based on query as a key. Depending on the exact sub-query calculation of its hash code can be slow (hybrid query works with any type of OpenSearch query). That is a problem on large datasets as this is done for each doc by each sub-query.

We can avoid creation and usage of that query to index map by storing sub-query index at [time we create collection of DISIWrapers](https://github.com/opensearch-project/neural-search/blob/2.13/src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java#L253-L260). 

As per benchmark results that gives about 20% performance boost. I've run it on 2.13 using [noaa OSB workload](https://github.com/opensearch-project/opensearch-benchmark-workloads/tree/main/noaa), all times are in ms:

Before the change (baseline)

```

Bool: p50 88.2863 | p90 103.777
Hybrid: p50 299.427 | p90 319.797

Bool: p50 92.7222 | p90 98.6847
Hybrid: p50 94.5511 | p90 111.645

Bool: p50 98.0301 | p90 108.948
Hybrid: p50 475.319 | p90 515.999
```

After the change:

```

Bool: p50 97.9306 | p90 116.299
Hybrid: p50 228.696 | p90 249.665

Bool: p50 87.3152 | p90 89.3061
Hybrid: p50 89.9654 | p90 92.349

Bool: p50 97.9891 | p90 114.396
Hybrid: p50 353.631 | p90 377.527
```

### Issues Resolved
https://github.com/opensearch-project/neural-search/issues/705

### Check List
    - [X] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
@@ -190,7 +190,8 @@ public final class HybridQueryScorer extends Scorer {
      */
     public float[] hybridScores() throws IOException {
         float[] scores = new float[subScores.length];
-        if (subScorersPQ.topList() instanceof HybridDisiWrapper == false) {
+        DisiWrapper topList = subScorersPQ.topList();
+        if (topList instanceof HybridDisiWrapper == false) {
             log.error(
                 String.format(
                     Locale.ROOT,
@@ -199,10 +200,12 @@ public final class HybridQueryScorer extends Scorer {
                     subScorersPQ.topList().getClass().getSimpleName()
                 )
             );
-            throw new IllegalStateException();
+            throw new IllegalStateException(
+                "Unable to collect scores for one of the sub-queries, encountered an unexpected type of score iterator."
+            );
         }
-        HybridDisiWrapper topList = (HybridDisiWrapper) subScorersPQ.topList();
-        for (HybridDisiWrapper disiWrapper = topList; disiWrapper != null; disiWrapper = (HybridDisiWrapper) disiWrapper.next) {
+        for (HybridDisiWrapper disiWrapper = (HybridDisiWrapper) topList; disiWrapper != null; disiWrapper =
+            (HybridDisiWrapper) disiWrapper.next) {
             // check if this doc has match in the subQuery. If not, add score as 0.0 and continue
             Scorer scorer = disiWrapper.scorer;
             if (scorer.docID() == DocIdSetIterator.NO_MORE_DOCS) {
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
@@ -8,18 +8,13 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Locale;
-import java.util.Map;
 import java.util.Objects;
 
-import com.google.common.primitives.Ints;
 import org.apache.lucene.search.DisiPriorityQueue;
 import org.apache.lucene.search.DisiWrapper;
 import org.apache.lucene.search.DisjunctionDISIApproximation;
 import org.apache.lucene.search.DocIdSetIterator;
-import org.apache.lucene.search.Query;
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
```diff
@@ -194,70 +187,30 @@ public int docID() {
      */
     public float[] hybridScores() throws IOException {
         float[] scores = new float[subScores.length];
-        DisiWrapper topList = subScorersPQ.topList();
-        for (DisiWrapper disiWrapper = topList; disiWrapper != null; disiWrapper = disiWrapper.next) {
+        HybridDisiWrapper topList = (HybridDisiWrapper) subScorersPQ.topList();
```
VijayanB, line 190: nit: can you check instance type before type case?

martin-gaievski: yes, added the check. My concern was added latency, but seems it's not that big. If we assume 39 ns per `instance of` call, then for my config in experiment 2 it will be 0.3ms, I think that's ok

navneet1v: Great... deep-dive again

src/main/java/org/opensearch/neuralsearch/search/HybridDisiWrapper.java
```diff
@@ -0,0 +1,23 @@
...
+import org.apache.lucene.search.DisiWrapper;
+import org.apache.lucene.search.Scorer;
+
+/**
+ * Wrapper for DisiWrapper, saves state of sub-queries for performance reasons
+ */
+@Getter
+public class HybridDisiWrapper extends DisiWrapper {
+    // index of disi wrapper sub-query object when its part of the hybrid query
+    int subQueryIndex = -1;
```
VijayanB, line 17: nit: private final int subQueryIndex ;

src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
VijayanB: It would have been nice if PriorityQueue uses Generics .

martin-gaievski: we're using not Java PriorityQueue but [Lucene implementation](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/DisiPriorityQueue.java#L30), it's already uses DisiWrapper as type

src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
```diff
@@ -194,70 +190,41 @@ public int docID() {
...
+        if (subScorersPQ.topList() instanceof HybridDisiWrapper == false) {
+            log.error(
+                String.format(
+                    Locale.ROOT,
+                    "Unexpected type of DISI wrapper, expected [%s] but found [%s]",
+                    HybridDisiWrapper.class.getSimpleName(),
+                    subScorersPQ.topList().getClass().getSimpleName()
+                )
+            );
+            throw new IllegalStateException();
```
navneet1v, line 202: add the exception message in this exception too.

navneet1v (approved):

src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
martin-gaievski: ack

VijayanB (approved):
LGTM.

Issue #729, opened 2024-05-02
In hybrid query allow to skip parallel score collection by core TopDocsCollector 


martin-gaievski:
As part of performance optimization for Hybrid query we need to find a way to minimize time taken by parallel score collecting processes that is running in core with TopDocsCollector. As per following information collected during profiling these calls take 40-80% of CPU time.
 
![325698784-4a1336dd-923a-4f07-bb22-5fb37720dfd5](https://github.com/opensearch-project/neural-search/assets/94878159/b3e390be-2dec-4314-9ab2-19eacc89549a)

![325698823-74ac1ec0-f400-4c27-97c4-0228618bf88f](https://github.com/opensearch-project/neural-search/assets/94878159/ceab2295-92e8-4758-bb6b-5c307f30824a)

As a baseline we're taking results from previous PR related to hybrid query optimization, those are based on 2.13 version and [noaa OSB workload](https://github.com/opensearch-project/opensearch-benchmark-workloads/tree/main/noaa), all times are in ms:

```
One sub-query that selects 11M documents

Bool: p50 97.9306 | p90 116.299
Hybrid: p50 228.696 | p90 249.665

One sub-query that selects 1.6K documents

Bool: p50 87.3152 | p90 89.3061
Hybrid: p50 89.9654 | p90 92.349

Three sub-query that select 15M documents

Bool: p50 97.9891 | p90 114.396
Hybrid: p50 353.631 | p90 377.527
```

Most likely that will be a compound change in both core OpenSearch and neural-search plugin. Preferred suggestion is - provide capability to skip or ignore TopDocsCollector in core QueryPhaseSearcher (Core side) and by using that new option call only HybridQueryDocsCollector (plugin side).

- https://github.com/opensearch-project/OpenSearch/issues/13170 - feature request in core for allowing caller to pass empty query collector context

Issue #704, opened 2024-04-23
[META] Improve Hybrid query latency


martin-gaievski:
Hybrid query has high latency comparing to other compound queries like Boolean query. Based on results collected for 2.13 and depending on the dataset and exact query it may be up to 10 times slower than Bool. Another reason for this issue is degradation in performance of hybrid query comparing to initial release e.g. in OpenSearch 2.11.

Following are goals for this work:
- bring performance of hybrid query to a level when it's comparable with bool query: 
- For small datasets and sub-sets it should much Bool with deviation within 20% for p90
- For large datasets (10M+ documents) and if a sub-queries return large sub-set of documents (1M+ documents in sub-query result) hybrid query should perform no worse than 2x of Bool query
- Multiple sub-queries can add additional overhead of no more than 20% of overall query time for p90
- reach the level of performance of hybrid query released in 2.11

There were some GH issues in the past that are related to the same problem, e.g. https://github.com/opensearch-project/neural-search/issues/281. In addition to that, based on analysis of the source code and some profiling I can think of following list of items:

- don't execute TopDocsCollector core collector as it takes compute and results are ignored   
- optimize plugin code for better performance: check for sub-optimal initializations, loops, type conversions etc.
- for cases when some of sub-queries are rewritten to the same lucene form - execute only one query and copy scores

Github issues for each child item:
- https://github.com/opensearch-project/neural-search/issues/279 parallel execution of sub-queries
- https://github.com/opensearch-project/neural-search/issues/705 replace streaming API calls
- https://github.com/opensearch-project/OpenSearch/issues/13170 allow empty query collector context (skip TopDocsCollector)
- https://github.com/opensearch-project/neural-search/issues/729 enable empty query collector context for hybrid query scenario
- https://github.com/opensearch-project/neural-search/issues/745 optimize the way we iterate over results and collect scores of sub queries

PR #731, opened 2024-05-03
Pass empty QueryCollectorContext in case of hybrid query to improve latencies by 20%

martin-gaievski:
### Description
In this change we're improving hybrid query latencies by avoiding redundant document score collection.

As per flamegraphs posted in https://github.com/opensearch-project/neural-search/issues/729 from 40 to 80% of CPU time of hybrid query processing is taken by TopDocsCollector in core. Those scores are not needed and should be removed from the execution flow in case this is hybrid query.

At the hight level the idea is to set empty doc collector instead of top docs collector in case incoming query is of type hybrid. 

Ability to pass empty doc collector has been [added to core in earlier PR](https://github.com/opensearch-project/OpenSearch/pull/13481). In this PR we're using this feature in neural plugin by passing new empty query collector context to query phase searcher. This is done by overriding `searchWith` method of both default and concurrent query phase searchers. The only collector and collector manager that will be executed for hybrid query are custom plugin implementations: [HybridTopScoreDocCollector](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java) and [HybridCollectorManager](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.jav).

Below are metrics for hybrid query latency that I've collected after this change. There are based on 2.x (2.14) version and [noaa OSB workload](https://github.com/opensearch-project/opensearch-benchmark-workloads/tree/main/noaa), all times are in ms:

```

Bool: p50 80.0931 | p90 80.4647
Hybrid: p50 182.314 | p90 204.263

Bool: p50 73.1325 | p90 73.7713
Hybrid: p50 72.9788 | p90 73.211

Bool: p50 85.8744 | p90 90.3619
Hybrid: p50 274.728 | p90 293.725
```

following is the baseline results, measurements are done before this change
```

Bool: p50 97.9306 | p90 116.299
Hybrid: p50 228.696 | p90 249.665

Bool: p50 87.3152 | p90 89.3061
Hybrid: p50 89.9654 | p90 92.349

Bool: p50 97.9891 | p90 114.396
Hybrid: p50 353.631 | p90 377.527
```

Following flamegraph (taken after this change) shows that there is no call to TopDocsCollector. More deep dive is needed to find next focus area for optimizations:

profile_hybrq_baseline_before_no_topdocscoll_5_2_1820

### Issues Resolved
https://github.com/opensearch-project/neural-search/issues/729
https://github.com/opensearch-project/neural-search/issues/704

### Check List
    - [ ] All tests pass


martin-gaievski:
BWC rolling upgrade tests will fail for 2.14 while the release is ongoing and core is already switched to 2.15 for 2.x branch, restart BWCs are fine

src/main/java/org/opensearch/neuralsearch/search/query/DefaultQueryPhaseSearcherWithEmptyQueryCollectorContext.java
```diff
@@ -0,0 +1,41 @@
...
+import org.opensearch.search.query.QueryPhase;
+
+import java.io.IOException;
+import java.util.LinkedList;
+
+/**
+ * Class that inherits DefaultQueryPhaseSearcher implementation but calls its search with only
+ * empty query collector context
+ */
+public class DefaultQueryPhaseSearcherWithEmptyQueryCollectorContext extends QueryPhase.DefaultQueryPhaseSearcher {
```
navneet1v, line 20: Let make these classes package private and final. As they should not be extended.

navneet1v, line 20: also see if we can get rid of these files and move classes inside our queryphase searcher

martin-gaievski: ack

martin-gaievski: ack

navneet1v (approved):

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -132,4 +151,60 @@ public AggregationProcessor aggregationProcessor(SearchContext searchContext) {
         AggregationProcessor coreAggProcessor = super.aggregationProcessor(searchContext);
         return new HybridAggregationProcessor(coreAggProcessor);
     }
+
+    /**
+     * Class that inherits ConcurrentQueryPhaseSearcher implementation but calls its search with only
+     * empty query collector context
+     */
+    @NoArgsConstructor(access = AccessLevel.PRIVATE)
+    final class ConcurrentQueryPhaseSearcherWithEmptyQueryCollectorContext extends ConcurrentQueryPhaseSearcher {
```
chishui, line 160: (refactor suggestion) better to move these two searcher classes out to independent files

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -36,6 +40,14 @@
 @Log4j2
 public class HybridQueryPhaseSearcher extends QueryPhaseSearcherWrapper {
 
+    private final QueryPhaseSearcher defaultQueryPhaseSearcherWithEmptyCollectorContext;
```
chishui, line 43: Seems like it's better to use `QueryPhaseSearcherWrapper` as composition instead of inheritance. We can add a constructor in `QueryPhaseSearcherWrapper` to accept both`QueryPhase.DefaultQueryPhaseSearcher` and `ConcurrentQueryPhaseSearcher`. Then `HybridQueryPhaseSearcher` can have two `QueryPhaseSearcherWrapper` instances, one is `defaultQueryPhaseSearcherWrapper`, the other as one is `queryPhaseSearcherWrapperWithEmptyCollectorContext`, then in `searchWith`, it's either `defaultQueryPhaseSearcherWrapper.searchWith` or `queryPhaseSearcherWrapperWithEmptyCollectorContext.searchWith`

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
martin-gaievski: I prefer inner classes approach, we don't need those to be extended and even visible to the users. As of now there is no useful payload in those classes except for the method override. Once that changed we can think of refactoring you've mentioned

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
martin-gaievski: For this we need change in OS core, but owners from core side are against it, method override is their preferred approach, mainly to keep changes on core side to minimum. You can check the discussion details in corresponding core PR https://github.com/opensearch-project/OpenSearch/issues/13170

VijayanB (approved):
LGTM.

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
chishui: The only change on core side from my suggestion is to add an additional constructor so that client can provide default and concurrent QueryPhaseSearcher. I'm afraid it's not the solution #13170 describes and core team is against.

PR #717, opened 2024-04-30
Fix: change max chunk limit exception

yuye-aws:
### Description
Fix the issue: https://github.com/opensearch-project/neural-search/issues/716

### Issues Resolved
Fix the issue: https://github.com/opensearch-project/neural-search/issues/716

### Check List
- [ ] New functionality includes testing.
    - [ ] All tests pass
- [ ] New functionality has been documented.
    - [ ] New functionality has javadoc added
- [ ] Commits are signed as per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #733, opened 2024-05-06
Optimize parameter parsing in text chunking processor

yuye-aws:
### Description
There PR optimizes the parameter parsing step in chunking algorithms. For both algorithms, all runtime parameters for text chunking processor is always available. Therefore, default value is not needed.

### Issues Resolved

Optimize parameter parsing in text chunking processor. 

### Check List
    - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
@@ -30,7 +30,8 @@ import static org.opensearch.neuralsearch.processor.chunker.Chunker.MAX_CHUNK_LI
 import static org.opensearch.neuralsearch.processor.chunker.Chunker.DEFAULT_MAX_CHUNK_LIMIT;
 import static org.opensearch.neuralsearch.processor.chunker.Chunker.DISABLED_MAX_CHUNK_LIMIT;
 import static org.opensearch.neuralsearch.processor.chunker.Chunker.CHUNK_STRING_COUNT_FIELD;
-import static org.opensearch.neuralsearch.processor.chunker.ChunkerParameterParser.parseIntegerParameter;
+import static org.opensearch.neuralsearch.processor.chunker.ChunkerParameterParser.parseInteger;
+import static org.opensearch.neuralsearch.processor.chunker.ChunkerParameterParser.parseIntegerWithDefault;
 
 /**
  * This processor is used for text chunking.
@@ -115,8 +116,8 @@ public final class TextChunkingProcessor extends AbstractProcessor {
         }
         Map<String, Object> chunkerParameters = (Map<String, Object>) algorithmValue;
         // parse processor level max chunk limit
-        this.maxChunkLimit = parseIntegerParameter(chunkerParameters, MAX_CHUNK_LIMIT_FIELD, DEFAULT_MAX_CHUNK_LIMIT);
-        if (maxChunkLimit < 0 && maxChunkLimit != DISABLED_MAX_CHUNK_LIMIT) {
+        this.maxChunkLimit = parseIntegerWithDefault(chunkerParameters, MAX_CHUNK_LIMIT_FIELD, DEFAULT_MAX_CHUNK_LIMIT);
+        if (maxChunkLimit <= 0 && maxChunkLimit != DISABLED_MAX_CHUNK_LIMIT) {
             throw new IllegalArgumentException(
                 String.format(
                     Locale.ROOT,
@@ -309,10 +310,10 @@ public final class TextChunkingProcessor extends AbstractProcessor {
         }
         List<String> contentResult = chunker.chunk(content, runTimeParameters);
         // update chunk_string_count for each string
-        int chunkStringCount = parseIntegerParameter(runTimeParameters, CHUNK_STRING_COUNT_FIELD, 1);
+        int chunkStringCount = parseInteger(runTimeParameters, CHUNK_STRING_COUNT_FIELD);
         runTimeParameters.put(CHUNK_STRING_COUNT_FIELD, chunkStringCount - 1);
         // update runtime max_chunk_limit if not disabled
-        int runtimeMaxChunkLimit = parseIntegerParameter(runTimeParameters, MAX_CHUNK_LIMIT_FIELD, maxChunkLimit);
+        int runtimeMaxChunkLimit = parseInteger(runTimeParameters, MAX_CHUNK_LIMIT_FIELD);
         if (runtimeMaxChunkLimit != DISABLED_MAX_CHUNK_LIMIT) {
             runTimeParameters.put(MAX_CHUNK_LIMIT_FIELD, runtimeMaxChunkLimit - contentResult.size());
         }
--- a/src/main/java/org/opensearch/neuralsearch/processor/chunker/ChunkerParameterParser.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/chunker/ChunkerParameterParser.java
@@ -20,13 +20,10 @@ public final class ChunkerParameterParser {
 
     /**
      * Parse String type parameter.
+     * Return default value if the parameter is missing.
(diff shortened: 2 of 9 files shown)


yuye-aws:
@zane-neo This PR resolves your comment in https://github.com/opensearch-project/neural-search/pull/717: 
> In fact, with field_map and document, we can always calculate a chunk_string_count, we don't need a default value here. Suggesting optimize this code in next PR
Please help review thanks.

yuye-aws:
We need to rerun the CI checks to check if there are any flakey tests.

yuye-aws:
Hi maintainer, please attach backport-2.x label to this PR.

src/main/java/org/opensearch/neuralsearch/processor/chunker/ChunkerParameterParser.java
```diff
@@ -39,14 +39,36 @@ public static String parseStringParameter(final Map<String, Object> parameters,
         return fieldValue.toString();
     }
 
+    /**
+     * Parse integer type parameter without default value
+     * Throw IllegalArgumentException if parameter is missing or is not an integer.
+     */
+    public static int parseIntegerParameter(final Map<String, Object> parameters, final String fieldName) {
+        if (!parameters.containsKey(fieldName)) {
```
martin-gaievski, line 47:
```suggestion
       return parseIntegerParameter(parameters, fieldName, null);
```

martin-gaievski: I think this design doesn't follow conventions in other place in Java API.
If you calling parse(argument) than it should execute parsing logic
If you calling parse(argument, defaultValue) than it should try to execute parsing logic, and return default value in case that parsing logic return empty value

In your code everywhere where this new method called with null you can call your new method without default value.
And in that new method implementation you can just call this method with defaultValue = null, see my previous comment with suggestion

src/main/java/org/opensearch/neuralsearch/processor/chunker/ChunkerParameterParser.java
```diff
@@ -41,12 +41,16 @@ public static String parseStringParameter(final Map<String, Object> parameters,
...
-     * Throw IllegalArgumentException if parameter is not an integer.
+     * Throw IllegalArgumentException if both parameter and default value is missing or parameter is not an integer.
      */
-    public static int parseIntegerParameter(final Map<String, Object> parameters, final String fieldName, final int defaultValue) {
+    public static int parseIntegerParameter(final Map<String, Object> parameters, final String fieldName, final Integer defaultValue) {
         if (!parameters.containsKey(fieldName)) {
-            // all integer parameters are optional
-            return defaultValue;
+            // return the default value when parameter is not existing
+            if (defaultValue == null) {
```
zane-neo, line 49: I suggest splitting this method to three new methods: 
1. parseInteger: Simply parse integer with catching number format exception.
2. parseIntegerWithDefault: if field not shown, return default value, else return `parseInteger`.
3. parseIntegerWithException: if field not shown, throw IllegalStateException, else return `parseInteger`.
This won't add repeated code also can make sure the method 2 have concise semantic by making the default value type to `int`, currently default value accepts `null` which is not in a correct semantic. For runtime parameters, there isn't default value, so throwing exception is the correct semantic for them.

yuye-aws: Oh, I don't think the function is necessary. I have removed it.

src/main/java/org/opensearch/neuralsearch/processor/chunker/ChunkerParameterParser.java
yuye-aws: Oh, I don't think the function is necessary. I have removed it.

src/main/java/org/opensearch/neuralsearch/processor/chunker/ChunkerParameterParser.java
yuye-aws: Since all integer parameters are optional. I will not keep the `parseIntegerWithException` method.

yuye-aws: I mean, runtime parameters are always available while non-runtime parameters have default values.

yuye-aws: Updated.

src/main/java/org/opensearch/neuralsearch/processor/chunker/ChunkerParameterParser.java
yuye-aws: > If you calling parse(argument, defaultValue) than it should try to execute parsing logic, and return default value in case that parsing logic return empty value

Not very sure what you mean here. We should throw illegal argument exception if necessary.

yuye-aws: > In your code everywhere where this new method called with null you can call your new method without default value.

Updated by Zan's comment. You can review now.

src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
```diff
@@ -115,8 +116,8 @@ private void parseAlgorithmMap(final Map<String, Object> algorithmMap) {
         }
         Map<String, Object> chunkerParameters = (Map<String, Object>) algorithmValue;
         // parse processor level max chunk limit
-        this.maxChunkLimit = parseIntegerParameter(chunkerParameters, MAX_CHUNK_LIMIT_FIELD, DEFAULT_MAX_CHUNK_LIMIT);
-        if (maxChunkLimit < 0 && maxChunkLimit != DISABLED_MAX_CHUNK_LIMIT) {
+        this.maxChunkLimit = parseIntegerWithDefault(chunkerParameters, MAX_CHUNK_LIMIT_FIELD, DEFAULT_MAX_CHUNK_LIMIT);
+        if (maxChunkLimit <= 0 && maxChunkLimit != DISABLED_MAX_CHUNK_LIMIT) {
```
yuye-aws, line 120: I am making this change to keep it consistent with the error message:
`Parameter [%s] must be positive or %s to disable this parameter"`

yuye-aws: Do you think we need to allow maxChunkLimit to be 0?

zane-neo: It could be, it's fine to add = 0 here.

zane-neo (approved):

martin-gaievski (commented):
Overall code looks good to me. One comment - can you add a unit test for ChunkerParameterParser class? As we're adding more methods to it having specialized test class will give more confidence

yuye-aws:
> Overall code looks good to me. One comment - can you add a unit test for ChunkerParameterParser class? As we're adding more methods to it having specialized test class will give more confidence

Sure. I will add unit tests today.

yuye-aws:
HI @zane-neo and @martin-gaievski ! I have implemented the unit tests for chunker parameter parser. Can you help review again?

src/test/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcherTests.java
```diff
@@ -24,7 +24,6 @@
 import java.util.LinkedList;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ExecutorService;
```
yuye-aws, line 27: This change is due to spotless apply

martin-gaievski (approved):
looks good to me, thank you

yuye-aws:
@martin-gaievski Can you merge the PR?

yuye-aws:
The last approver is expected to merge the PR.

Issue #743, opened 2024-05-09
[FEATURE] Support batch ingestion in TextEmbeddingProcessor & SparseEncodingProcessor


chishui:
### Is your feature request related to a problem?
RFC: https://github.com/opensearch-project/OpenSearch/issues/12457

We have implemented batch ingestion logic in OpenSearch core in version 2.14, now we want to enable the batch ingestion capability in neural-search processors: TextEmbeddingProcessor & SparseEncodingProcessor so that we can better utilize the remote ML server's GPU capacity and accelerate the ingestion process, based our benchmark, batch can reduce total ingestion time by 77% without seeing throttling error (P90, SageMaker), please refer to [here](https://github.com/opensearch-project/OpenSearch/issues/12457#issuecomment-2012437250) to see the benchmark results. 

### What solution would you like?
1. In `InferenceProcessor`, override `Processor`'s `batchExecute` API, add a default implementation to combine `List<String> inferenceText` from multiple docs, then reuse the `mlCommonsClientAccessor.inferenceSentences` and `mlCommonsClientAccessor.inferenceSentencesWithMapResult`. After getting inference results, map them to each doc and update the docs.
2. We'll sort the docs by length before sending for inference to achieve better performance. And inference results will be restored to original order before being processed.
**(This was original proposed in ml-commons. But as @ylwu-amzn suggested that we can reuse `input_docs_processed_step_size` as max batch size, then it makes more sense to sort the docs in neural-search as we can ensure that we won't sort docs from TextImageEmbeddingProcessor)**

### What alternatives have you considered?
N/A

### Do you have any additional context?
N/A

zhichao-aws:
Hi @chishui , could you please provide an example API request body to create a batched ingest processor?

martin-gaievski:
We are not adding this feature for `TextImageEmbeddingProcessor`, is there a plan to do it later?

chishui:
@zhichao-aws there is no changes to how TextEmbeddingProcessor and SparseEncodingProcessor are created. It's only when user uses `_bulk` API with `batch_size` parameter, then the processors will see documents in batches.

chishui:
@martin-gaievski we don't have a plan to support `TextImageEmbeddingProcessor` as it requires text and image to be grouped in one request.

chishui:
Updated description in "What solution would you like?" section.

navneet1v:
[...]

PR #744, opened 2024-05-09
[FEATURE] Support batch ingestion in TextEmbeddingProcessor & SparseEncodingProcessor

chishui:
### Description
Add support for batch ingestion in TextEmbeddingProcessor & SparseEncodingProcessor to improve ingestion performance.

https://github.com/opensearch-project/neural-search/issues/743

### Issues Resolved
https://github.com/opensearch-project/neural-search/issues/743

### Check List
    - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
@@ -125,6 +125,12 @@ public abstract class InferenceProcessor extends AbstractProcessor {
         }
     }
 
+    /**
+     * This is the function which does actual inference work for batchExecute interface.
+     * @param inferenceList a list of String for inference.
+     * @param handler a callback handler to handle inference results which is a list of objects.
+     * @param onException an exception callback to handle exception.
+     */
     abstract void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler, Consumer<Exception> onException);
 
     @Override
@@ -157,6 +163,8 @@ public abstract class InferenceProcessor extends AbstractProcessor {
             handler.accept(ingestDocumentWrappers);
         }, exception -> {
             for (IngestDocumentWrapper ingestDocumentWrapper : ingestDocumentWrappers) {
+                // The IngestDocumentWrapper might already run into exception and not sent for inference. So here we only
+                // set exception to IngestDocumentWrapper which doesn't have exception before.
                 if (ingestDocumentWrapper.getException() == null) {
                     ingestDocumentWrapper.update(ingestDocumentWrapper.getIngestDocument(), exception);
                 }
--- a/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
@@ -5,20 +5,26 @@
 package org.opensearch.neuralsearch.processor;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.function.BiConsumer;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+import lombok.AllArgsConstructor;
+import lombok.Getter;
 import org.apache.commons.lang3.StringUtils;
+import org.opensearch.core.common.util.CollectionUtils;
 import org.opensearch.env.Environment;
 import org.opensearch.index.mapper.MapperService;
 import org.opensearch.ingest.AbstractProcessor;
 import org.opensearch.ingest.IngestDocument;
+import org.opensearch.ingest.IngestDocumentWrapper;
 import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -117,6 +123,83 @@ public abstract class InferenceProcessor extends AbstractProcessor {
         } catch (Exception e) {
             handler.accept(null, e);
         }
+    }
+
+    abstract void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler, Consumer<Exception> onException);
+
+    @Override
+    public void batchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers, Consumer<List<IngestDocumentWrapper>> handler) {
+        if (CollectionUtils.isEmpty(ingestDocumentWrappers)) {
+            handler.accept(Collections.emptyList());
+            return;
+        }
+
+        List<DataForInference> dataForInferences = getDataForInference(ingestDocumentWrappers);
+        List<String> inferenceList = constructInferenceTexts(dataForInferences);
+        if (inferenceList.isEmpty()) {
+            handler.accept(ingestDocumentWrappers);
+            return;
+        }
+        doBatchExecute(inferenceList, results -> {
+            int startIndex = 0;
+            for (DataForInference dataForInference : dataForInferences) {
+                if (dataForInference.getIngestDocumentWrapper().getException() != null || dataForInference.getInferenceList().isEmpty()) {
+                    continue;
+                }
+                List<?> inferenceResults = results.subList(startIndex, startIndex + dataForInference.getInferenceList().size());
+                startIndex += dataForInference.getInferenceList().size();
+                setVectorFieldsToDocument(
+                    dataForInference.getIngestDocumentWrapper().getIngestDocument(),
+                    dataForInference.getProcessMap(),
+                    inferenceResults
+                );
+            }
+            handler.accept(ingestDocumentWrappers);
+        }, exception -> {
+            for (IngestDocumentWrapper ingestDocumentWrapper : ingestDocumentWrappers) {
+                if (ingestDocumentWrapper.getException() == null) {
+                    ingestDocumentWrapper.update(ingestDocumentWrapper.getIngestDocument(), exception);
+                }
+            }
+            handler.accept(ingestDocumentWrappers);
+        });
+    }
+
+    private List<String> constructInferenceTexts(List<DataForInference> dataForInferences) {
+        List<String> inferenceTexts = new ArrayList<>();
+        for (DataForInference dataForInference : dataForInferences) {
+            if (dataForInference.getIngestDocumentWrapper().getException() != null) {
+                continue;
+            }
+            inferenceTexts.addAll(dataForInference.getInferenceList());
+        }
+        return inferenceTexts;
+    }
+
+    private List<DataForInference> getDataForInference(List<IngestDocumentWrapper> ingestDocumentWrappers) {
+        List<DataForInference> dataForInferences = new ArrayList<>();
+        for (IngestDocumentWrapper ingestDocumentWrapper : ingestDocumentWrappers) {
+            Map<String, Object> processMap = null;
+            List<String> inferenceList = null;
+            try {
+                validateEmbeddingFieldsValue(ingestDocumentWrapper.getIngestDocument());
+                processMap = buildMapWithProcessorKeyAndOriginalValue(ingestDocumentWrapper.getIngestDocument());
+                inferenceList = createInferenceList(processMap);
+            } catch (Exception e) {
+                ingestDocumentWrapper.update(ingestDocumentWrapper.getIngestDocument(), e);
+            } finally {
+                dataForInferences.add(new DataForInference(ingestDocumentWrapper, processMap, inferenceList));
+            }
+        }
+        return dataForInferences;
+    }
+
+    @Getter
+    @AllArgsConstructor
+    private static class DataForInference {
+        private final IngestDocumentWrapper ingestDocumentWrapper;
+        private final Map<String, Object> processMap;
+        private final List<String> inferenceList;
     }
 
     @SuppressWarnings({ "unchecked" })
--- a/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
@@ -5,6 +5,7 @@
 package org.opensearch.neuralsearch.processor;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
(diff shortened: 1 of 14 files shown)


src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -119,6 +125,83 @@ public void execute(IngestDocument ingestDocument, BiConsumer<IngestDocument, Ex
...
+                List<?> inferenceResults = results.subList(startIndex, startIndex + dataForInference.getInferenceList().size());
+                startIndex += dataForInference.getInferenceList().size();
+                setVectorFieldsToDocument(
+                    dataForInference.getIngestDocumentWrapper().getIngestDocument(),
+                    dataForInference.getProcessMap(),
+                    inferenceResults
+                );
+            }
+            handler.accept(ingestDocumentWrappers);
+        }, exception -> {
```
zhichao-aws, line 158: Is it possible for parts of documents failed but others success? For example, for local deployment case, if we do the inference one by one (I didn't confirm it but it should be), is it possible?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -119,6 +125,83 @@ public void execute(IngestDocument ingestDocument, BiConsumer<IngestDocument, Ex
...
+            for (IngestDocumentWrapper ingestDocumentWrapper : ingestDocumentWrappers) {
+                if (ingestDocumentWrapper.getException() == null) {
+                    ingestDocumentWrapper.update(ingestDocumentWrapper.getIngestDocument(), exception);
+                }
+            }
+            handler.accept(ingestDocumentWrappers);
+        });
+    }
+
+    private List<String> constructInferenceTexts(List<DataForInference> dataForInferences) {
```
zhichao-aws, line 168: How about make this a member-method of  DataForInference class?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -119,6 +125,83 @@ public void execute(IngestDocument ingestDocument, BiConsumer<IngestDocument, Ex
         }
     }
 
+    abstract void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler, Consumer<Exception> onException);
```
zhichao-aws, line 128: Do we need to implement a sync-style doBatchExecute here like the `execute(IngestDocument ingestDocument)` and `execute(IngestDocument ingestDocument, BiConsumer<IngestDocument, Exception> handler)`

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -48,4 +49,13 @@ public void doExecute(
             handler.accept(ingestDocument, null);
         }, e -> { handler.accept(null, e); }));
     }
+
+    @Override
+    public void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler, Consumer<Exception> onException) {
+        mlCommonsClientAccessor.inferenceSentences(
+            this.modelId,
+            inferenceList,
+            ActionListener.wrap(vectors -> { handler.accept(vectors); }, exception -> onException.accept(exception))
```
zhichao-aws, line 58: Just use `onException` here instead of `exception -> onException.accept(exception)`

src/test/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessorTests.java
```diff
@@ -170,6 +172,28 @@ public void testExecute_withMapTypeInput_successful() {
 
     }
 
+    public void test_batchExecute_successful() {
```
zhichao-aws, line 175: Can we add test case of failure?

zhichao-aws (commented):
Great job! I don't have critical concern on this PR, just some minor points.
Please add integ tests about batch ingestion, thanks

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
chishui: Here is the code for local model, https://github.com/opensearch-project/ml-commons/blob/caf1d65fb3372c4a39c8d21545016ffdbf6c1c9a/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/TextEmbeddingModel.java#L43

we predict them one by one and if any of the prediction throws exception, the whole batch predict throw exception, so it's either success or all fail.

For remote inference, partial failure will cause inference exception callback triggered https://github.com/opensearch-project/ml-commons/blob/94a113da8a2d1e28a84ba7a422b43287ac00448e/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/remote/MLSdkAsyncHttpResponseHandler.java#L183.

Since the logic of how processors invoke ml-commons remains the same, the error handling logic also remains the same.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
chishui: I feel logic of these methods have stronger relations with `InferenceProcessor` than with `DataForInference` and `DataForInference` is just of a data wrapper. If there is a need to reuse this method outside of this class in the future, we can make it a separate utility class, for now, it's not very necessary.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
chishui: We didn't create the version without callback of `batchExecute` in `Processor` as it's only the callback version which is triggered from pipeline.

zhichao-aws (commented):
LGTM. After we bump version in this repo and pass the bwc rolling-upgrade test I can give an approval

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -119,6 +125,83 @@ public void execute(IngestDocument ingestDocument, BiConsumer<IngestDocument, Ex
...
+                setVectorFieldsToDocument(
+                    dataForInference.getIngestDocumentWrapper().getIngestDocument(),
+                    dataForInference.getProcessMap(),
+                    inferenceResults
+                );
+            }
+            handler.accept(ingestDocumentWrappers);
+        }, exception -> {
+            for (IngestDocumentWrapper ingestDocumentWrapper : ingestDocumentWrappers) {
+                if (ingestDocumentWrapper.getException() == null) {
```
martin-gaievski, line 160: what will happen with `exception` if ingestDocumentWrapper.getException() is not null? Can you add a code comment that explains why we're updating document wrapper only in case it doesn't have stored exception, and ignoring it otherwise?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
martin-gaievski, line 128: please add comment for method and params

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
chishui: "ingestDocumentWrapper.getException() is not null" means there was exception already and we don't overwrite it. Will add comment

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -165,6 +180,31 @@ public void batchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers, Con
...
+            originalOrderMap.put(i, docsWithIndex.get(i).v1());
+        }
+        return Tuple.tuple(sortedInferenceList, originalOrderMap);
+    }
+
+    private List<?> restoreToOriginalOrder(List<?> results, Map<Integer, Integer> originalOrder) {
+        List<Object> sortedResults = new ArrayList<>();
+        sortedResults.addAll(results);
+        for (int i = 0; i < results.size(); ++i) {
+            if (!originalOrder.containsKey(i)) continue;
```
zhichao-aws, line 201: This condition won't be triggered, right?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -165,6 +180,31 @@ public void batchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers, Con
...
     }
 
+    private Tuple<List<String>, Map<Integer, Integer>> sortByLengthAndReturnOriginalOrder(List<String> inferenceList) {
+        List<Tuple<Integer, String>> docsWithIndex = new ArrayList<>();
+        for (int i = 0; i < inferenceList.size(); ++i) {
+            docsWithIndex.add(Tuple.tuple(i, inferenceList.get(i)));
+        }
+        docsWithIndex.sort(Comparator.comparingInt(t -> t.v2().length()));
+        List<String> sortedInferenceList = docsWithIndex.stream().map(Tuple::v2).collect(Collectors.toList());
+        Map<Integer, Integer> originalOrderMap = new HashMap<>();
```
zhichao-aws, line 190: suggestion: use an array or array list to store the map of order. i.e. element index -> value in array
It is more efficient than hashmap

zhichao-aws (commented):
Please merge or rebase main branch to update the bwc workflow

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
chishui: Actually, ml-commons could return more than what we send (which I will fix later) so it could be triggered. For example, we have docs [a, b, c, d, e], the return could be [ra, rb, rc, rd, re, rc, rd, re, re], the good thing is it wont' cause incorrect response to user.

zhichao-aws: I also noticed this problem, and fixed it in https://github.com/opensearch-project/ml-commons/pull/2462. After that I think this won't be triggered

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
chishui: > It is more efficient than hashmap

I ran a test, read in List is more performant than Map, but for write, Map is better. The time difference for 10000 times run with data size 1000, for write, Map takes 0.08s less, for read, List takes 0.05s less. For this logic read and write is 1:1, it's really hard to tell which one is better than the other.

zhichao-aws (approved):
lgtm

chishui:
@martin-gaievski do you have other comments? Can I get an approval from you?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -119,6 +128,120 @@ public void execute(IngestDocument ingestDocument, BiConsumer<IngestDocument, Ex
...
+            return;
+        }
+        Tuple<List<String>, Map<Integer, Integer>> sortedResult = sortByLengthAndReturnOriginalOrder(inferenceList);
+        inferenceList = sortedResult.v1();
+        Map<Integer, Integer> originalOrder = sortedResult.v2();
+        doBatchExecute(inferenceList, results -> {
+            int startIndex = 0;
+            results = restoreToOriginalOrder(results, originalOrder);
+            for (DataForInference dataForInference : dataForInferences) {
+                if (dataForInference.getIngestDocumentWrapper().getException() != null || dataForInference.getInferenceList().isEmpty()) {
```
martin-gaievski, line 159: should we have check for null for `dataForInference.getInferenceList()`, seems that can potentially produce NPE

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
martin-gaievski: I would double check on performance, basically based on:
- list still is for wrapper type Integer, that boxing/unboxing takes cycles. I think we can use array of ints here
- 1.000 is not that big size, datasets can be in tens of hundreds of millions

martin-gaievski (commented):
In this approach do we accumulate documents in the memory buffer. Do we have some sort of limit for potential memory amount that takes to keep it? Maybe it's discussed in the RFC, I couldn't find it.
It's possible that user's doc is huge and if the size of the batch is high then system will crush as we don't have any sort of memory circuit breaker.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -119,6 +128,120 @@ public void execute(IngestDocument ingestDocument, BiConsumer<IngestDocument, Ex
...
+        Map<Integer, Integer> originalOrderMap = new HashMap<>();
+        for (int i = 0; i < docsWithIndex.size(); ++i) {
+            originalOrderMap.put(i, docsWithIndex.get(i).v1());
+        }
+        return Tuple.tuple(sortedInferenceList, originalOrderMap);
+    }
+
+    private List<?> restoreToOriginalOrder(List<?> results, Map<Integer, Integer> originalOrder) {
+        List<Object> sortedResults = new ArrayList<>();
+        sortedResults.addAll(results);
```
zane-neo, line 199: Seems this can be optimized to using array to avoid the `addAll` operation

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
chishui: > In this approach do we accumulate documents in the memory buffer. Do we have some sort of limit for potential memory amount that takes to keep it? Maybe it's discussed in the RFC, I couldn't find it.
It's possible that user's doc is huge and if the size of the batch is high then system will crush as we don't have any sort of memory circuit breaker.

We don't deep copy documents, it's only references we keep, they refer to the same document object in memory.

> should we have check for null for dataForInference.getInferenceList(), seems that can potentially produce NPE

Will add null check

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
chishui: Will change

zane-neo (approved):

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
chishui: I tested the performance of `new XX()` vs `toArrays()` with code below, `new XX()` is slightly performant with 0.1 micro second, but the reason I'm copying results to new list is that I only want to change the ones with orders in the `originalOrder` and keep remaining as they are to limit impact.

Issue #745, opened 2024-05-09
In hybrid query optimize the way we iterate over results and collect scores of sub queries


martin-gaievski:
As part of performance optimization for Hybrid query we need to find a way to minimize time taken by getting next matching doc for sun query and collection sun query scores. As per following information collected during profiling these calls take ~85% of CPU time.

![327627242-9901f27b-3399-4a21-a543-6a21e541a1fc (1)](https://github.com/opensearch-project/neural-search/assets/94878159/ce3eee86-bc06-4210-99e7-76f887c2a19b)

As a baseline we're taking results from previous PR related to hybrid query optimization, those are based on 2.13 version and [noaa OSB workload](https://github.com/opensearch-project/opensearch-benchmark-workloads/tree/main/noaa), all times are in ms:

```
One sub-query that selects 11M documents

Bool: p50 77.8893 | p90 78.1916
Hybrid: p50 186.709 | p90 197.739

One sub-query that selects 1.6K documents

Bool: p50 71.0947 | p90 71.691
Hybrid: p50 71.5156 | p90 72.8801

Three sub-query that select 15M documents

Bool: p50 87.0556 | p90 90.9105
Hybrid: p50 287.255 | p90 313.868
```

The current logic of iterating over doc and collecting scores is following:

- iterators for individual sub queries are wrapped into compound iterator. Wrapper provides ability to position all individual iterators  to the same doc id
- bulk scorer iterate over results by requesting next doc id, wrapper doc iterator return collection of sub query iterators that are positioned on the next (same) doc id
- for doc id collector collecting scores. that's done by using collection of priority queues of hits for every sub query. priority queue is needed to sort scores in desc order as iterators cannot do it by design - different sub queries may have different order of scores and different doc ids.

PR #746, opened 2024-05-10
Use lazy initialization for priority queue of hits and scores to improve latencies by 20%

martin-gaievski:
### Description
In this change we're improving hybrid query latencies by changing allocation of query hits objects for hybrid scores. Currently those objects are pre-populated, then for new element top of the priority queue is updated and heap is rebalanced. We're changing this to lazy initialization - priority queue is initially empty, for each hit we're inserting element. If we reach the capacity then element with low score is pushed out of the heap. As per benchmark results this alone gives 10% improvement.

With several other minor changes like pre calculate and store size instead of calling size() for each doc and each sub-query we can get 10% percent. 

Below are metrics for hybrid query latency that I've collected after this change. There are based on 2.x (2.14) version and [noaa OSB workload](https://github.com/opensearch-project/opensearch-benchmark-workloads/tree/main/noaa), all times are in ms:

```

Bool: p50 76.5531 | p90 77.8108
Hybrid: p50 133.644 | p90 151.903

Bool: p50 69.9585 | p90 70.2855
Hybrid: p50 70.1703 | p90 70.5498

Bool: p50 87.1001 | p90 87.7131
Hybrid: p50 234.802 | p90 262.876
```

following are the baseline results
```

Bool: p50 77.8893 | p90 78.1916
Hybrid: p50 186.709 | p90 197.739

Bool: p50 71.0947 | p90 71.691
Hybrid: p50 71.5156 | p90 72.8801

Bool: p50 87.0556 | p90 90.9105
Hybrid: p50 287.255 | p90 313.868
```

Following framegraph has been captured after this change:

profile_hybrq_step4v2_59_1610

### Issues Resolved
https://github.com/opensearch-project/neural-search/issues/745
https://github.com/opensearch-project/neural-search/issues/704

### Check List
- [X] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
@@ -4,14 +4,7 @@
  */
 package org.opensearch.neuralsearch.query;
 
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
-import java.util.Locale;
-import java.util.Objects;
-
+import lombok.Getter;
 import lombok.extern.log4j.Log4j2;
 import org.apache.lucene.search.DisiPriorityQueue;
 import org.apache.lucene.search.DisiWrapper;
@@ -21,10 +14,15 @@ import org.apache.lucene.search.ScoreMode;
 import org.apache.lucene.search.Scorer;
 import org.apache.lucene.search.TwoPhaseIterator;
 import org.apache.lucene.search.Weight;
-
-import lombok.Getter;
 import org.apache.lucene.util.PriorityQueue;
 import org.opensearch.neuralsearch.search.HybridDisiWrapper;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
 
 /**
  * Class abstracts functionality of Scorer for hybrid query. When iterating over documents in increasing
@@ -40,11 +38,10 @@ public final class HybridQueryScorer extends Scorer {
 
     private final DisiPriorityQueue subScorersPQ;
 
-    private final float[] subScores;
-
     private final DocIdSetIterator approximation;
     private final HybridScoreBlockBoundaryPropagator disjunctionBlockPropagator;
     private final TwoPhase twoPhase;
+    private final int numSubqueries;
 
     public HybridQueryScorer(final Weight weight, final List<Scorer> subScorers) throws IOException {
         this(weight, subScorers, ScoreMode.TOP_SCORES);
@@ -53,7 +50,7 @@ public final class HybridQueryScorer extends Scorer {
     HybridQueryScorer(final Weight weight, final List<Scorer> subScorers, final ScoreMode scoreMode) throws IOException {
         super(weight);
         this.subScorers = Collections.unmodifiableList(subScorers);
-        subScores = new float[subScorers.size()];
+        this.numSubqueries = subScorers.size();
         this.subScorersPQ = initializeSubScorersPQ();
         boolean needsScores = scoreMode != ScoreMode.COMPLETE_NO_SCORES;
 
@@ -100,13 +97,8 @@ public final class HybridQueryScorer extends Scorer {
      */
     @Override
     public float score() throws IOException {
-        return score(getSubMatches());
-    }
-
-    private float score(DisiWrapper topList) throws IOException {
         float totalScore = 0.0f;
-        for (DisiWrapper disiWrapper = topList; disiWrapper != null; disiWrapper = disiWrapper.next) {
-            // check if this doc has match in the subQuery. If not, add score as 0.0 and continue
+        for (DisiWrapper disiWrapper : subScorersPQ) {
             if (disiWrapper.scorer.docID() == DocIdSetIterator.NO_MORE_DOCS) {
                 continue;
             }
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
```diff
@@ -113,10 +112,10 @@ public void collect(int doc) throws IOException {
                     }
                     totalHits[i]++;
                     PriorityQueue<ScoreDoc> pq = compoundScores[i];
-                    ScoreDoc topDoc = pq.top();
-                    topDoc.doc = doc + docBase;
-                    topDoc.score = score;
-                    pq.updateTop();
+                    ScoreDoc topDoc = new ScoreDoc(doc + docBase, score);
```
navneet1v, line 115: rename topDoc variable to may be doc or currentDoc etc.

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
```diff
@@ -113,10 +112,10 @@ public void collect(int doc) throws IOException {
                     }
                     totalHits[i]++;
                     PriorityQueue<ScoreDoc> pq = compoundScores[i];
-                    ScoreDoc topDoc = pq.top();
-                    topDoc.doc = doc + docBase;
-                    topDoc.score = score;
-                    pq.updateTop();
```
navneet1v, line 119: I have question on the old implementation, even if top of the pq has a score better than current doc we were still updating the top of the doc. this looks to me like a bug. @martin-gaievski how it was working earlier?

martin-gaievski: that PriorityQueue is a [lucene implementation](https://github.com/yintaoxue/read-open-source-code/blob/master/solr-4.10.4/src/org/apache/lucene/util/PriorityQueue.java) of min heap, so top is actually the lowest score. In previous implementation heap got pre-populated with zero score objects, then we iterate over doc scores, put them to heap and push down to the bottom of the heap. This is they call [downHeap in lucene class](https://github.com/yintaoxue/read-open-source-code/blob/master/solr-4.10.4/src/org/apache/lucene/util/PriorityQueue.java#L233) 
Example:
```
- initialization [0, 0, 0, 0, 0] 
- step 1, score 0.5 -> [0, 0, 0, 0, 0.5]
- step 2, score 0.3 -> [0, 0, 0, 0.3, 0.5]
- step 3, score 0.6 -> [0, 0, 0.3, 0.6, 0.5]
- step 4, score 0.2 -> [0, 0.2, 0.3, 0.6, 0.5]
```

When we fill final collection of scores we:
- pop all empty (score 0) elements 
- pop all elements and fill the collection starting from the end going to element with index 0. part of the pop() method is to rebuild the heap so new top element is the min element

Example
```
- we know there were 4 hits, so left only last 4 elements [0, 0.2, 0.3, 0.6, 0.5] -> [0.2, 0.3, 0.6, 0.5]
- [0.2, 0.3, 0.6, 0.5] -> final scores [0.2]  min heap [0.3, 0.6, 0.5]
- [0.3, 0.6, 0.5] -> final scores [0.2, 0.3]  min heap [0.5, 0.6] here we're doing downHeap and pushing min element to the top
- [0.5, 0.6] -> final scores [0.2, 0.3, 0.5]  min heap [0.6]
- [0.6] -> final scores [0.2, 0.3, 0.5, 0.6]  min heap []
```

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
martin-gaievski: discussed with Navneet offline, there was a bug in old implementation. 

The size of the heap is[ less of equal max number of hits among all sub queries](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java#L69). This means we're not storing all the scores, but keep only sub set of it in the heap. 
we were updating top element blindly without checking if the element we're adding is less or greater than that element. In case when heap is full that will affect min element. If affected that's alway one element.
With new implementation the bug is not there, this is taken care of by [lessThan](https://github.com/yintaoxue/read-open-source-code/blob/master/solr-4.10.4/src/org/apache/lucene/util/PriorityQueue.java#L151) check. This lessThan is implemented for ScoreDoc in [lucene HitQueue](https://github.com/yintaoxue/read-open-source-code/blob/master/lucene-4.7.2/src/org/apache/lucene/search/HitQueue.java#L76-L82)

navneet1v (approved):

src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java
```diff
@@ -53,7 +50,7 @@ public HybridQueryScorer(final Weight weight, final List<Scorer> subScorers) thr
     HybridQueryScorer(final Weight weight, final List<Scorer> subScorers, final ScoreMode scoreMode) throws IOException {
         super(weight);
         this.subScorers = Collections.unmodifiableList(subScorers);
-        subScores = new float[subScorers.size()];
+        this.numSubqueries = subScorers.size();
```
VijayanB, line 53: why do we need this variable? Why can't we use  this.subScorers.size() wherever it is required?

martin-gaievski: Mainly for saving on CPU cycles. That call to size() poped up in the flame graph as a contributor to the CPU cycles time, single digit percentage.

PR #747, opened 2024-05-14
[Feature] Neural Sparse Query Two Phase Search pipeline

### LINKED ISSUE #646: [RFC] Enhancing  Neural Sparse  Query Speed with a Two-Phase Approach

conggguan:
### Description
This change implement for #646 
- Enhance the speed of neuralsparse query by two-phase.
- Now support top-level, boolean and boost compound query; for other compound query will degrade into origin logic and speed.

### Feature support query
- [x] NeuralSparseQuery
- [x] NeuralSparseQuery nested in BoostQuery
- [x] NeuralSparseQuery nested in BooleanQuery

### Issues Resolved
Resolve #646 

### Check List
    - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
@@ -6,14 +6,13 @@ package org.opensearch.neuralsearch.processor;
 
 import lombok.Getter;
 import lombok.Setter;
-import org.apache.lucene.search.BooleanClause;
 import org.opensearch.common.SetOnce;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.index.query.BoolQueryBuilder;
 import org.opensearch.index.query.QueryBuilder;
-import org.opensearch.index.query.QueryBuilderVisitor;
 import org.opensearch.ingest.ConfigurationUtils;
 import org.opensearch.neuralsearch.query.NeuralSparseQueryBuilder;
+import org.opensearch.search.builder.SearchSourceBuilder;
 import org.opensearch.search.pipeline.AbstractProcessor;
 import org.opensearch.search.pipeline.Processor;
 import org.opensearch.search.pipeline.SearchRequestProcessor;
@@ -25,20 +24,24 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+/**
+ * A SearchRequestProcessor to generate two-phase NeuralSparseQueryBuilder,
+ * and add it to the Rescore of a searchRequest.
+ */
 @Setter
 @Getter
 public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements SearchRequestProcessor {
 
-    public static String TYPE = "neural_sparse_two_phase_processor";
-    boolean enable;
-    float ratio;
-    float window_expansion;
-    int max_window_size;
-    static final String PARAMETER_KEY = "two_phase_parameter";
-    static final String RATIO_KEY = "prune_ratio";
-    static final String ENABLE_KEY = "enabled";
-    static final String EXPANSION_KEY = "expansion";
-    static final String MAX_WINDOW_SIZE_KEY = "max_window_size";
+    public static final String TYPE = "neural_sparse_two_phase_processor";
+    private boolean enabled;
+    private float ratio;
+    private float window_expansion;
+    private int max_window_size;
+    private static final  String PARAMETER_KEY = "two_phase_parameter";
+    private static final  String RATIO_KEY = "prune_ratio";
+    private static final  String ENABLE_KEY = "enabled";
+    private static final  String EXPANSION_KEY = "expansion_rate";
+    private static final  String MAX_WINDOW_SIZE_KEY = "max_window_size";
 
     protected NeuralSparseTwoPhaseProcessor(
         String tag,
@@ -50,7 +53,7 @@ public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements 
         int max_window_size
     ) {
         super(tag, description, ignoreFailure);
-        this.enable = enabled;
+        this.enabled = enabled;
         if (ratio < 0f || ratio > 1f) {
             throw new IllegalArgumentException(String.format(Locale.ROOT, "The prune ratio must be within [0, 1]. Received: %f", ratio));
         }
@@ -69,17 +72,32 @@ public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements 
         this.max_window_size = max_window_size;
     }
 
+    /**
+     * Process the search request of neural_sparse_two_phase_processor
+     * @param request the search request (which may have been modified by an earlier processor)
+     * @return request the search request that add the two-phase rescore query of neural sparse query.
+     */
     @Override
     public SearchRequest processRequest(SearchRequest request) throws Exception {
-        if (!enable) return request;
+        if (!enabled) return request;
         QueryBuilder queryBuilder = request.source().query();
         Map<NeuralSparseQueryBuilder, Float> queryBuilderMap = new HashMap<>();
-        QueryBuilderVisitor queryBuilderVisitor = new TwoPhaseQueryBuilderVisitor(queryBuilderMap, 1.0f);
-        queryBuilder.visit(queryBuilderVisitor);
-        BoolQueryBuilder boolQueryBuilder = new BoolQueryBuilder();
-        queryBuilderMap.forEach((key, value) -> boolQueryBuilder.should(((QueryBuilder) key).boost(value)));
+        processQueryBuilder(queryBuilder, queryBuilderMap, 1.0f);
+        BoolQueryBuilder boolQueryBuilder = getBoolQueryBuilderFromNeuralSparseQueryBuilderMap(queryBuilderMap);
+        boolQueryBuilder.boost(getOriginQueryWeightAfterRescore(request.source()));
         RescorerBuilder<QueryRescorerBuilder> rescorerBuilder = new QueryRescorerBuilder(boolQueryBuilder);
-        rescorerBuilder.windowSize((int) (request.source().size() == -1 ? 10 : request.source().size() * window_expansion));
+        int windowSize = (int) ((request.source().size() == -1 ? 10 : request.source().size()) * window_expansion);
+        if (windowSize > max_window_size || windowSize < 0) {
+            throw new IllegalArgumentException(
+                String.format(
+                    Locale.ROOT,
+                    "The two-phase window size of neural_sparse_two_phase_processor should be [0,%d], but get the value of %d",
+                    max_window_size,
+                    windowSize
+                )
+            );
+        }
+        rescorerBuilder.windowSize(windowSize);
         request.source().addRescorer(rescorerBuilder);
         return request;
     }
@@ -89,11 +107,17 @@ public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements 
         return TYPE;
     }
 
-    public static Map<Boolean, SetOnce<Map<String, Float>>> getSplitSetOnceByScoreThreshold(Map<String, Float> queryTokens) {
+    /**
+     * Based on ratio, split a Map into two map by the value.
+     * @param queryTokens the queryTokens map, key is the token String, value is the score.
+     * @param ratio The ratio that control how tokens map be split.
+     * @return A map has two element, {[True, token map whose value above threshold],[False, token map whose value below threshold]}
+     */
+    public static Map<Boolean, SetOnce<Map<String, Float>>> getSplitSetOnceByScoreThreshold(Map<String, Float> queryTokens, float ratio) {
         float max = 0f;
         for (Float value : queryTokens.values())
             max = value > max ? value : max;
-        float threshold = max * 0.4f;
+        float threshold = max * ratio;
         Map<Boolean, Map<String, Float>> queryTokensByScore = queryTokens.entrySet()
             .stream()
             .collect(
@@ -135,35 +159,55 @@ public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements 
             return new NeuralSparseTwoPhaseProcessor(tag, description, ignoreFailure, enabled, ratio, window_expansion, max_window_size);
 
         }
-
     }
 
-    @Setter
-    static private class TwoPhaseQueryBuilderVisitor implements QueryBuilderVisitor {
-        @Setter
-        float boost;
-        float subBoost;
-        Map<NeuralSparseQueryBuilder, Float> queryBuilderMap;
-
-        public TwoPhaseQueryBuilderVisitor(Map<NeuralSparseQueryBuilder, Float> queryBuilderMap, float boost) {
-            this.queryBuilderMap = queryBuilderMap;
-            this.boost = boost;
-            this.subBoost = boost;
-        }
-
-        @Override
-        public void accept(QueryBuilder qb) {
-            if (qb instanceof NeuralSparseQueryBuilder) {
-                this.queryBuilderMap.put(((NeuralSparseQueryBuilder) qb).copyForTwoPhase(), boost * qb.boost());
-            } else {
-                subBoost *= qb.boost();
-            }
-        }
-
-        @Override
-        public QueryBuilderVisitor getChildVisitor(BooleanClause.Occur occur) {
-            if (occur.equals(BooleanClause.Occur.SHOULD)) return new TwoPhaseQueryBuilderVisitor(queryBuilderMap, subBoost);
-            return NO_OP_VISITOR;
+    private void processQueryBuilder(
+        QueryBuilder queryBuilder,
+        Map<NeuralSparseQueryBuilder, Float> neuralSparseQueryBuilderFloatMap,
+        float baseBoost
+    ) {
+        if (queryBuilder instanceof BoolQueryBuilder) {
+            processQueryBuilder((BoolQueryBuilder) queryBuilder, neuralSparseQueryBuilderFloatMap, baseBoost);
+        } else if (queryBuilder instanceof NeuralSparseQueryBuilder) {
+            processQueryBuilder((NeuralSparseQueryBuilder) queryBuilder, neuralSparseQueryBuilderFloatMap, baseBoost);
         }
     }
+
+    private void processQueryBuilder(
+        BoolQueryBuilder queryBuilder,
+        Map<NeuralSparseQueryBuilder, Float> neuralSparseQueryBuilderFloatMap,
+        float baseBoost
+    ) {
+        baseBoost *= queryBuilder.boost();
+        for (QueryBuilder subShouldQueryBuilder : queryBuilder.should()) {
+            processQueryBuilder(subShouldQueryBuilder, neuralSparseQueryBuilderFloatMap, baseBoost);
+        }
+    }
+
+    private void processQueryBuilder(
+        NeuralSparseQueryBuilder queryBuilder,
+        Map<NeuralSparseQueryBuilder, Float> neuralSparseQueryBuilderFloatMap,
+        float baseBoost
+    ) {
+        float finalBaseBoost = baseBoost * queryBuilder.boost();
+        neuralSparseQueryBuilderFloatMap.put(queryBuilder.copyForTwoPhase(ratio), finalBaseBoost);
+    }
+
+    private BoolQueryBuilder getBoolQueryBuilderFromNeuralSparseQueryBuilderMap(Map<NeuralSparseQueryBuilder, Float> queryBuilderFloatMap) {
+        BoolQueryBuilder boolQueryBuilder = new BoolQueryBuilder();
+        queryBuilderFloatMap.forEach((neuralSparseQueryBuilder, boost) -> {
+            final float finalBoost = neuralSparseQueryBuilder.boost() * boost;
+            boolQueryBuilder.should(neuralSparseQueryBuilder.boost(finalBoost));
+        });
+        return boolQueryBuilder;
+    }
+
+    private static float getOriginQueryWeightAfterRescore(SearchSourceBuilder searchSourceBuilder) {
+        if (searchSourceBuilder.rescores() == null) return 1.0f;
+        return searchSourceBuilder.rescores()
+            .stream()
+            .map(rescorerBuilder -> ((QueryRescorerBuilder) rescorerBuilder).getQueryWeight())
+            .reduce(1.0f, (a, b) -> a * b);
+    }
+
 }
--- a/src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
@@ -122,13 +122,15 @@ public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements 
      *
      * @param queryTokens    the queryTokens map, key is the token String, value is the score.
      * @param thresholdRatio The ratio that control how tokens map be split.
-     * @return A map has two element, {[True, token map whose value above threshold],[False, token map whose value below threshold]}
+     * @return A tuple has two element, { token map whose value above threshold, token map whose value below threshold }
      */
     public static Tuple<Map<String, Float>, Map<String, Float>> splitQueryTokensByRatioedMaxScoreAsThreshold(
         final Map<String, Float> queryTokens,
         final float thresholdRatio
     ) {
-
+        if (Objects.isNull(queryTokens)) {
+            throw new IllegalArgumentException("Query tokens cannot be null or empty.");
+        }
         float max = 0f;
         for (Float value : queryTokens.values()) {
             max = Math.max(value, max);
@@ -143,8 +145,8 @@ public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements 
 
         Map<String, Float> highScoreTokens = queryTokensByScore.get(Boolean.TRUE);
         Map<String, Float> lowScoreTokens = queryTokensByScore.get(Boolean.FALSE);
-        if (Objects.isNull(highScoreTokens) || highScoreTokens.isEmpty()) {
-            throw new IllegalArgumentException("Query tokens cannot be null or empty.");
+        if (Objects.isNull(highScoreTokens)) {
+            highScoreTokens = Collections.emptyMap();
         }
         if (Objects.isNull(lowScoreTokens)) {
             lowScoreTokens = Collections.emptyMap();
--- a/src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
@@ -4,6 +4,8 @@
  */
 package org.opensearch.neuralsearch.processor;
 
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
 import lombok.Getter;
 import lombok.Setter;
 import org.opensearch.common.SetOnce;
@@ -19,7 +21,6 @@ import org.opensearch.search.pipeline.SearchRequestProcessor;
 import org.opensearch.search.rescore.QueryRescorerBuilder;
 import org.opensearch.search.rescore.RescorerBuilder;
 
-import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -37,11 +38,11 @@ public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements 
     private float ratio;
     private float window_expansion;
     private int max_window_size;
-    private static final  String PARAMETER_KEY = "two_phase_parameter";
-    private static final  String RATIO_KEY = "prune_ratio";
-    private static final  String ENABLE_KEY = "enabled";
-    private static final  String EXPANSION_KEY = "expansion_rate";
-    private static final  String MAX_WINDOW_SIZE_KEY = "max_window_size";
+    private static final String PARAMETER_KEY = "two_phase_parameter";
+    private static final String RATIO_KEY = "prune_ratio";
+    private static final String ENABLE_KEY = "enabled";
+    private static final String EXPANSION_KEY = "expansion_rate";
+    private static final String MAX_WINDOW_SIZE_KEY = "max_window_size";
 
     protected NeuralSparseTwoPhaseProcessor(
         String tag,
@@ -55,18 +56,20 @@ public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements 
         super(tag, description, ignoreFailure);
         this.enabled = enabled;
         if (ratio < 0f || ratio > 1f) {
-            throw new IllegalArgumentException(String.format(Locale.ROOT, "The prune ratio must be within [0, 1]. Received: %f", ratio));
+            throw new IllegalArgumentException(
+                String.format(Locale.ROOT, "The two_phase_parameter.prune_ratio must be within [0, 1]. Received: %f", ratio)
+            );
         }
         this.ratio = ratio;
-        if (window_expansion < 1.0) {
+        if (window_expansion < 1.0f) {
             throw new IllegalArgumentException(
-                String.format(Locale.ROOT, "The window expansion must be greater than or equal to 1.0. Received: %f", window_expansion)
+                String.format(Locale.ROOT, "The two_phase_parameter.expansion_rate must >= 1.0. Received: %f", window_expansion)
             );
         }
         this.window_expansion = window_expansion;
         if (max_window_size < 50) {
             throw new IllegalArgumentException(
-                String.format(Locale.ROOT, "The maximum window size must be greater than or equal to 50. Received: %n" + max_window_size)
+                String.format(Locale.ROOT, "The two_phase_parameter.max_window_size must >= 50. Received: %n" + max_window_size)
             );
         }
         this.max_window_size = max_window_size;
@@ -79,14 +82,18 @@ public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements 
      */
     @Override
     public SearchRequest processRequest(SearchRequest request) throws Exception {
-        if (!enabled) return request;
+        if (!enabled || ratio == 0f) {
+            return request;
+        }
         QueryBuilder queryBuilder = request.source().query();
-        Map<NeuralSparseQueryBuilder, Float> queryBuilderMap = new HashMap<>();
-        processQueryBuilder(queryBuilder, queryBuilderMap, 1.0f);
-        BoolQueryBuilder boolQueryBuilder = getBoolQueryBuilderFromNeuralSparseQueryBuilderMap(queryBuilderMap);
-        boolQueryBuilder.boost(getOriginQueryWeightAfterRescore(request.source()));
-        RescorerBuilder<QueryRescorerBuilder> rescorerBuilder = new QueryRescorerBuilder(boolQueryBuilder);
-        int windowSize = (int) ((request.source().size() == -1 ? 10 : request.source().size()) * window_expansion);
+        Multimap<NeuralSparseQueryBuilder, Float> queryBuilderMap = ArrayListMultimap.create();
+        collectNeuralSparseQueryBuilder(queryBuilder, queryBuilderMap, 1.0f);
+        if (queryBuilderMap.isEmpty()) return request;
+        QueryBuilder nestedTwoPhaseQueryBuilder = getNestedQueryBuilderFromNeuralSparseQueryBuilderMap(queryBuilderMap);
+        nestedTwoPhaseQueryBuilder.boost(getOriginQueryWeightAfterRescore(request.source()));
+        RescorerBuilder<QueryRescorerBuilder> twoPhaseRescorer = new QueryRescorerBuilder(nestedTwoPhaseQueryBuilder);
+        int requestSize = request.source().size();
+        int windowSize = (int) ((requestSize == -1 ? 10 : requestSize) * window_expansion);
         if (windowSize > max_window_size || windowSize < 0) {
             throw new IllegalArgumentException(
                 String.format(
@@ -97,8 +104,8 @@ public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements 
                 )
             );
         }
-        rescorerBuilder.windowSize(windowSize);
-        request.source().addRescorer(rescorerBuilder);
+        twoPhaseRescorer.windowSize(windowSize);
+        request.source().addRescorer(twoPhaseRescorer);
         return request;
     }
 
@@ -113,11 +120,15 @@ public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements 
      * @param ratio The ratio that control how tokens map be split.
      * @return A map has two element, {[True, token map whose value above threshold],[False, token map whose value below threshold]}
      */
-    public static Map<Boolean, SetOnce<Map<String, Float>>> getSplitSetOnceByScoreThreshold(Map<String, Float> queryTokens, float ratio) {
+    public static Map<Boolean, SetOnce<Map<String, Float>>> getSplitSetOnceByScoreThreshold(
+        final Map<String, Float> queryTokens,
+        final float ratio
+    ) {
         float max = 0f;
         for (Float value : queryTokens.values())
-            max = value > max ? value : max;
+            max = Math.max(value, max);
         float threshold = max * ratio;
+
         Map<Boolean, Map<String, Float>> queryTokensByScore = queryTokens.entrySet()
             .stream()
             .collect(
@@ -129,6 +140,10 @@ public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements 
         return Map.of(Boolean.TRUE, highScoreTksSetOnce, Boolean.FALSE, lowScoreTksSetOnce);
     }
 
+    /**
+     * Factory to create NeuralSparseTwoPhaseProcessor, provide default parameter,
+     *
+     */
     public static class Factory implements Processor.Factory<SearchRequestProcessor> {
         @Override
         public NeuralSparseTwoPhaseProcessor create(
@@ -142,72 +157,74 @@ public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements 
 
             boolean enabled = ConfigurationUtils.readBooleanProperty(TYPE, tag, config, ENABLE_KEY, true);
             Map<String, Object> map = ConfigurationUtils.readOptionalMap(TYPE, tag, config, PARAMETER_KEY);
+
             float ratio = 0.4f;
             float window_expansion = 5.0f;
             int max_window_size = 10000;
             if (map != null) {
-                if (map.containsKey(RATIO_KEY)) {
-                    ratio = ((Number) map.get(RATIO_KEY)).floatValue();
-                }
-                if (map.containsKey(EXPANSION_KEY)) {
-                    window_expansion = ((Number) map.get(EXPANSION_KEY)).floatValue();
-                }
-                if (map.containsKey(MAX_WINDOW_SIZE_KEY)) {
-                    max_window_size = ((Number) map.get(MAX_WINDOW_SIZE_KEY)).intValue();
-                }
+                ratio = ((Number) map.getOrDefault(RATIO_KEY, ratio)).floatValue();
+                window_expansion = ((Number) map.getOrDefault(EXPANSION_KEY, window_expansion)).floatValue();
+                max_window_size = ((Number) map.getOrDefault(MAX_WINDOW_SIZE_KEY, max_window_size)).intValue();
             }
+
             return new NeuralSparseTwoPhaseProcessor(tag, description, ignoreFailure, enabled, ratio, window_expansion, max_window_size);
-
         }
     }
 
-    private void processQueryBuilder(
-        QueryBuilder queryBuilder,
-        Map<NeuralSparseQueryBuilder, Float> neuralSparseQueryBuilderFloatMap,
-        float baseBoost
+    private QueryBuilder getNestedQueryBuilderFromNeuralSparseQueryBuilderMap(
+        final Multimap<NeuralSparseQueryBuilder, Float> queryBuilderFloatMap
     ) {
-        if (queryBuilder instanceof BoolQueryBuilder) {
-            processQueryBuilder((BoolQueryBuilder) queryBuilder, neuralSparseQueryBuilderFloatMap, baseBoost);
-        } else if (queryBuilder instanceof NeuralSparseQueryBuilder) {
-            processQueryBuilder((NeuralSparseQueryBuilder) queryBuilder, neuralSparseQueryBuilderFloatMap, baseBoost);
-        }
-    }
-
-    private void processQueryBuilder(
-        BoolQueryBuilder queryBuilder,
-        Map<NeuralSparseQueryBuilder, Float> neuralSparseQueryBuilderFloatMap,
-        float baseBoost
-    ) {
-        baseBoost *= queryBuilder.boost();
-        for (QueryBuilder subShouldQueryBuilder : queryBuilder.should()) {
-            processQueryBuilder(subShouldQueryBuilder, neuralSparseQueryBuilderFloatMap, baseBoost);
-        }
-    }
-
-    private void processQueryBuilder(
-        NeuralSparseQueryBuilder queryBuilder,
-        Map<NeuralSparseQueryBuilder, Float> neuralSparseQueryBuilderFloatMap,
-        float baseBoost
-    ) {
-        float finalBaseBoost = baseBoost * queryBuilder.boost();
-        neuralSparseQueryBuilderFloatMap.put(queryBuilder.copyForTwoPhase(ratio), finalBaseBoost);
-    }
-
-    private BoolQueryBuilder getBoolQueryBuilderFromNeuralSparseQueryBuilderMap(Map<NeuralSparseQueryBuilder, Float> queryBuilderFloatMap) {
         BoolQueryBuilder boolQueryBuilder = new BoolQueryBuilder();
-        queryBuilderFloatMap.forEach((neuralSparseQueryBuilder, boost) -> {
-            final float finalBoost = neuralSparseQueryBuilder.boost() * boost;
+        queryBuilderFloatMap.asMap().forEach((neuralSparseQueryBuilder, boosts) -> {
+            float reduceBoost = boosts.stream().reduce(0.0f, Float::sum);
+            final float finalBoost = neuralSparseQueryBuilder.boost() * reduceBoost;
             boolQueryBuilder.should(neuralSparseQueryBuilder.boost(finalBoost));
         });
         return boolQueryBuilder;
     }
 
-    private static float getOriginQueryWeightAfterRescore(SearchSourceBuilder searchSourceBuilder) {
+    private float getOriginQueryWeightAfterRescore(final SearchSourceBuilder searchSourceBuilder) {
         if (searchSourceBuilder.rescores() == null) return 1.0f;
         return searchSourceBuilder.rescores()
             .stream()
             .map(rescorerBuilder -> ((QueryRescorerBuilder) rescorerBuilder).getQueryWeight())
             .reduce(1.0f, (a, b) -> a * b);
+    }
+
+    private void collectNeuralSparseQueryBuilder(
+        final QueryBuilder queryBuilder,
+        final Multimap<NeuralSparseQueryBuilder, Float> neuralSparseQueryBuilderFloatMap,
+        float baseBoost
+    ) {
+        if (queryBuilder instanceof BoolQueryBuilder) {
+            collectNeuralSparseQueryBuilder((BoolQueryBuilder) queryBuilder, neuralSparseQueryBuilderFloatMap, baseBoost);
+        } else if (queryBuilder instanceof NeuralSparseQueryBuilder) {
+            collectNeuralSparseQueryBuilder((NeuralSparseQueryBuilder) queryBuilder, neuralSparseQueryBuilderFloatMap, baseBoost);
+        }
+    }
+
+    private void collectNeuralSparseQueryBuilder(
+        final BoolQueryBuilder queryBuilder,
+        final Multimap<NeuralSparseQueryBuilder, Float> neuralSparseQueryBuilderFloatMap,
+        float baseBoost
+    ) {
+        baseBoost *= queryBuilder.boost();
+        for (QueryBuilder subQuery : queryBuilder.should()) {
+            if (subQuery instanceof BoolQueryBuilder) {
+                collectNeuralSparseQueryBuilder(subQuery, neuralSparseQueryBuilderFloatMap, baseBoost);
+            } else if (subQuery instanceof NeuralSparseQueryBuilder) {
+                collectNeuralSparseQueryBuilder((NeuralSparseQueryBuilder) subQuery, neuralSparseQueryBuilderFloatMap, baseBoost);
+            }
+        }
+    }
+
+    private void collectNeuralSparseQueryBuilder(
+        final NeuralSparseQueryBuilder queryBuilder,
+        final Multimap<NeuralSparseQueryBuilder, Float> neuralSparseQueryBuilderFloatMap,
+        float baseBoost
+    ) {
+        baseBoost *= queryBuilder.boost();
+        neuralSparseQueryBuilderFloatMap.put(queryBuilder.getCopyNeuralSparseQueryBuilderForTwoPhase(ratio), baseBoost);
     }
 
 }
--- a/src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
@@ -236,15 +236,15 @@ public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements 
             Map<String, Object> twoPhaseConfigMap = ConfigurationUtils.readOptionalMap(TYPE, tag, config, PARAMETER_KEY);
 
             float ratio = DEFAULT_RATIO;
-            float window_expansion = DEFAULT_WINDOW_EXPANSION;
-            int max_window_size = DEFAULT_MAX_WINDOW_SIZE;
+            float windowExpansion = DEFAULT_WINDOW_EXPANSION;
+            int maxWindowSize = DEFAULT_MAX_WINDOW_SIZE;
             if (Objects.nonNull(twoPhaseConfigMap)) {
                 ratio = ((Number) twoPhaseConfigMap.getOrDefault(RATIO_KEY, ratio)).floatValue();
-                window_expansion = ((Number) twoPhaseConfigMap.getOrDefault(EXPANSION_KEY, window_expansion)).floatValue();
-                max_window_size = ((Number) twoPhaseConfigMap.getOrDefault(MAX_WINDOW_SIZE_KEY, max_window_size)).intValue();
+                windowExpansion = ((Number) twoPhaseConfigMap.getOrDefault(EXPANSION_KEY, windowExpansion)).floatValue();
+                maxWindowSize = ((Number) twoPhaseConfigMap.getOrDefault(MAX_WINDOW_SIZE_KEY, maxWindowSize)).intValue();
             }
 
-            return new NeuralSparseTwoPhaseProcessor(tag, description, ignoreFailure, enabled, ratio, window_expansion, max_window_size);
+            return new NeuralSparseTwoPhaseProcessor(tag, description, ignoreFailure, enabled, ratio, windowExpansion, maxWindowSize);
         }
     }
 
--- a/src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
@@ -188,6 +188,19 @@ public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements 
         } else if (queryBuilder instanceof NeuralSparseQueryBuilder) {
             NeuralSparseQueryBuilder neuralSparseQueryBuilder = (NeuralSparseQueryBuilder) queryBuilder;
             float updatedBoost = baseBoost * neuralSparseQueryBuilder.boost();
+            /*
+             * We obtain a copied modifiedQueryBuilder from the valid origin NeuralSparseQueryBuilder. After this,
+             * when the original NeuralSparseQueryBuilder starts to rewrite, it will only retain the tokens that
+             * have higher scores (controlled by the maxScore * ratio). The tokens with lower scores will be
+             * passed to the modifiedQueryBuilder's queryTokenSupplier.
+             *
+             * By doing this, we reduce the score computation time for the original NeuralSparseQueryBuilder,
+             * and use the modifiedQueryBuilder to make a score increment on TopDocs.
+             *
+             * When 2-phase is enabled:
+             *     - Docs besides TopDocs: Score = HighScoreToken's score
+             *     - Final TopDocs: Score = HighScoreToken's score + LowScoreToken's score
+             */
             NeuralSparseQueryBuilder modifiedQueryBuilder = neuralSparseQueryBuilder.getCopyNeuralSparseQueryBuilderForTwoPhase(ratio);
             result.put(modifiedQueryBuilder, updatedBoost);
         }
--- a/src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
@@ -8,8 +8,8 @@ import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.Multimap;
 import lombok.Getter;
 import lombok.Setter;
-import org.opensearch.common.SetOnce;
 import org.opensearch.action.search.SearchRequest;
+import org.opensearch.common.collect.Tuple;
 import org.opensearch.index.query.BoolQueryBuilder;
 import org.opensearch.index.query.QueryBuilder;
 import org.opensearch.ingest.ConfigurationUtils;
@@ -21,6 +21,7 @@ import org.opensearch.search.pipeline.SearchRequestProcessor;
 import org.opensearch.search.rescore.QueryRescorerBuilder;
 import org.opensearch.search.rescore.RescorerBuilder;
 
+import java.util.Collections;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
@@ -35,8 +36,6 @@ import java.util.stream.Collectors;
 public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements SearchRequestProcessor {
 
     public static final String TYPE = "neural_sparse_two_phase_processor";
-    public static final String UP_THRESHOLD = "UP_THRESHOLD";
-    public static final String DOWN_THRESHOLD = "DOWN_THRESHOLD";
     private boolean enabled;
     private float ratio;
     private float windowExpansion;
@@ -93,7 +92,7 @@ public class NeuralSparseTwoPhaseProcessor extends AbstractProcessor implements 
      * @return request the search request that add the two-phase rescore query of neural sparse query.
      */
     @Override
-    public SearchRequest processRequest(final SearchRequest request) throws Exception {
+    public SearchRequest processRequest(final SearchRequest request) {
         if (!enabled || ratio == 0f) {
             return request;
         }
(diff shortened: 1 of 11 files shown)


src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,230 @@
...
+        this.max_window_size = max_window_size;
+    }
+
+    /**
+     * Process the search request of neural_sparse_two_phase_processor
+     * @param request the search request (which may have been modified by an earlier processor)
+     * @return request the search request that add the two-phase rescore query of neural sparse query.
+     */
+    @Override
+    public SearchRequest processRequest(SearchRequest request) throws Exception {
```
martin-gaievski, line 84: make `request` param `final`

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,230 @@
...
+        QueryBuilder queryBuilder = request.source().query();
+        Multimap<NeuralSparseQueryBuilder, Float> queryBuilderMap = ArrayListMultimap.create();
+        collectNeuralSparseQueryBuilder(queryBuilder, queryBuilderMap, 1.0f);
+        if (queryBuilderMap.isEmpty()) return request;
+        QueryBuilder nestedTwoPhaseQueryBuilder = getNestedQueryBuilderFromNeuralSparseQueryBuilderMap(queryBuilderMap);
+        nestedTwoPhaseQueryBuilder.boost(getOriginQueryWeightAfterRescore(request.source()));
+        RescorerBuilder<QueryRescorerBuilder> twoPhaseRescorer = new QueryRescorerBuilder(nestedTwoPhaseQueryBuilder);
+        int requestSize = request.source().size();
+        int windowSize = (int) ((requestSize == -1 ? 10 : requestSize) * window_expansion);
+        if (windowSize > max_window_size || windowSize < 0) {
```
martin-gaievski, line 97: move everything about creating twoPhaseRescorer and setting its value into a separate method

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,230 @@
...
+            PipelineContext pipelineContext
+        ) throws IllegalArgumentException {
+
+            boolean enabled = ConfigurationUtils.readBooleanProperty(TYPE, tag, config, ENABLE_KEY, true);
+            Map<String, Object> map = ConfigurationUtils.readOptionalMap(TYPE, tag, config, PARAMETER_KEY);
+
+            float ratio = 0.4f;
+            float window_expansion = 5.0f;
+            int max_window_size = 10000;
+            if (map != null) {
```
martin-gaievski, line 164:
```suggestion
            if (Objects.nonNull(map)) {
```

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,230 @@
...
+    private void collectNeuralSparseQueryBuilder(
+        final QueryBuilder queryBuilder,
+        final Multimap<NeuralSparseQueryBuilder, Float> neuralSparseQueryBuilderFloatMap,
+        float baseBoost
+    ) {
+        if (queryBuilder instanceof BoolQueryBuilder) {
+            collectNeuralSparseQueryBuilder((BoolQueryBuilder) queryBuilder, neuralSparseQueryBuilderFloatMap, baseBoost);
+        } else if (queryBuilder instanceof NeuralSparseQueryBuilder) {
+            collectNeuralSparseQueryBuilder((NeuralSparseQueryBuilder) queryBuilder, neuralSparseQueryBuilderFloatMap, baseBoost);
+        }
```
martin-gaievski, line 203: can you add comment on what will happen if all if conditions are not met

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,230 @@
...
+
+    private void collectNeuralSparseQueryBuilder(
+        final BoolQueryBuilder queryBuilder,
+        final Multimap<NeuralSparseQueryBuilder, Float> neuralSparseQueryBuilderFloatMap,
+        float baseBoost
+    ) {
+        baseBoost *= queryBuilder.boost();
+        for (QueryBuilder subQuery : queryBuilder.should()) {
+            if (subQuery instanceof BoolQueryBuilder) {
+                collectNeuralSparseQueryBuilder(subQuery, neuralSparseQueryBuilderFloatMap, baseBoost);
```
martin-gaievski, line 214: which one of your two methods will be called here? I think for clarity it's better to rename them, so it's clear from the name if it accepts BooleanQuery or generic Query

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -77,6 +80,10 @@ public class NeuralSparseQueryBuilder extends AbstractQueryBuilder<NeuralSparseQ
     private String modelId;
     private Float maxTokenScore;
     private Supplier<Map<String, Float>> queryTokensSupplier;
+    // A parameter that can detect if twoPhase are enabled, when twoPhasePruneRatio equals -1f, it means it two-phase rescoreQueryBuilder's
+    // subQueryBuilder.
+    private float twoPhasePruneRatio = 0F;
+    private NeuralSparseQueryBuilder twoPhaseNeuralSparseQueryBuilder = null;
```
martin-gaievski, line 86: initialization with null is not needed here

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,230 @@
...
+        }
+    }
+
+    private void collectNeuralSparseQueryBuilder(
+        final NeuralSparseQueryBuilder queryBuilder,
+        final Multimap<NeuralSparseQueryBuilder, Float> neuralSparseQueryBuilderFloatMap,
+        float baseBoost
+    ) {
+        baseBoost *= queryBuilder.boost();
+        neuralSparseQueryBuilderFloatMap.put(queryBuilder.getCopyNeuralSparseQueryBuilderForTwoPhase(ratio), baseBoost);
```
martin-gaievski, line 227: can we rewrite it to a form when input arguments are not mutated inside the method that is called recursively. We can return pair of objects to the caller.

martin-gaievski, line 227: Another comment - seems we're creating copy of the NeuralSparceQueryBuilder every time we reach here. In worse case how many times we can hit this place? Do we really need a new copy every time?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -270,40 +299,54 @@ private static void parseQueryParams(XContentParser parser, NeuralSparseQueryBui
     }
 
     @Override
-    protected QueryBuilder doRewrite(QueryRewriteContext queryRewriteContext) throws IOException {
-        // We need to inference the sentence to get the queryTokens. The logic is similar to NeuralQueryBuilder
-        // If the inference is finished, then rewrite to self and call doToQuery, otherwise, continue doRewrite
-        if (null != queryTokensSupplier) {
+    protected QueryBuilder doRewrite(QueryRewriteContext queryRewriteContext) {
+        if (queryTokensSupplier != null) {
```
martin-gaievski, line 303: Objects.nonNull

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -270,40 +299,54 @@ private static void parseQueryParams(XContentParser parser, NeuralSparseQueryBui
...
     @Override
     protected Query doToQuery(QueryShardContext context) throws IOException {
         final MappedFieldType ft = context.fieldMapper(fieldName);
         validateFieldType(ft);
-
         Map<String, Float> queryTokens = queryTokensSupplier.get();
-        validateQueryTokens(queryTokens);
-
+        if (null == queryTokens) {
+            if (twoPhasePruneRatio == -1f) return new MatchNoDocsQuery();
```
martin-gaievski, line 347: formatting, please use braces for if case.

martin-gaievski (commented):
General comments for this PR:
- please think of simpler and more optimal low level design, logic is a bit overcomplicated
- formatting, please put calls under if or else conditions into a braces even if it's one liner

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
conggguan: Added.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
conggguan: Moved.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
conggguan: Changed.

conggguan: Indeed, consolidating all query tokens to construct a single NeuralSparseQueryBuilder is a viable strategy. This approach could potentially increase the overall efficiency of the queries, but it might also complicate the logic. Queries involving a large number of NeuralSparseQuery components may not be very common. Therefore, do you think make a more complicate but more effective logic is essential?

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
conggguan: I integrated them into a one more simple function now.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,223 @@
...
+        float threshold = max * ratio;
+
+        Map<Boolean, Map<String, Float>> queryTokensByScore = queryTokens.entrySet()
+            .stream()
+            .collect(
+                Collectors.partitioningBy(entry -> entry.getValue() >= threshold, Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue))
+            );
+        SetOnce<Map<String, Float>> highScoreTksSetOnce = new SetOnce<>(queryTokensByScore.get(Boolean.TRUE));
+        SetOnce<Map<String, Float>> lowScoreTksSetOnce = new SetOnce<>(queryTokensByScore.get(Boolean.FALSE));
+        if (highScoreTksSetOnce.get() == null) throw new IllegalArgumentException();
```
martin-gaievski, line 130:
```suggestion
        if (Objects.isNull(highScoreTksSetOnce.get())) {
                 throw new IllegalArgumentException();
        }         
```

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,223 @@
...
+            Map<String, Object> config,
+            PipelineContext pipelineContext
+        ) throws IllegalArgumentException {
+
+            boolean enabled = ConfigurationUtils.readBooleanProperty(TYPE, tag, config, ENABLE_KEY, true);
+            Map<String, Object> twoPhaseConfigMap = ConfigurationUtils.readOptionalMap(TYPE, tag, config, PARAMETER_KEY);
+
+            float ratio = 0.4f;
+            float window_expansion = 5.0f;
+            int max_window_size = 10000;
```
martin-gaievski, line 154: please make constants for those 3 values you're assigning to variable initially

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,223 @@
...
+        queryBuilderFloatMap.asMap().forEach((neuralSparseQueryBuilder, boosts) -> {
+            float reduceBoost = boosts.stream().reduce(0.0f, Float::sum);
+            final float finalBoost = neuralSparseQueryBuilder.boost() * reduceBoost;
+            boolQueryBuilder.should(neuralSparseQueryBuilder.boost(finalBoost));
+        });
+        return boolQueryBuilder;
+    }
+
+    private float getOriginQueryWeightAfterRescore(final SearchSourceBuilder searchSourceBuilder) {
+        if (searchSourceBuilder.rescores() == null) return 1.0f;
```
martin-gaievski, line 178:
```suggestion
        if (Objects.isNull(searchSourceBuilder.rescores())) {
              return 1.0f;
        }
```

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,223 @@
...
+    @Override
+    public SearchRequest processRequest(final SearchRequest request) throws Exception {
+        if (!enabled || ratio == 0f) {
+            return request;
+        }
+        QueryBuilder queryBuilder = request.source().query();
+        // Collect the nested NeuralSparseQueryBuilder in the whole query.
+        Multimap<NeuralSparseQueryBuilder, Float> queryBuilderMap = ArrayListMultimap.create();
+        queryBuilderMap = collectNeuralSparseQueryBuilder(queryBuilder, 1.0f);
+        if (queryBuilderMap.isEmpty()) return request;
```
martin-gaievski, line 93:
```suggestion
        if (queryBuilderMap.isEmpty()) {
              return request;
        }
```

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -113,6 +120,28 @@ public NeuralSparseQueryBuilder(StreamInput in) throws IOException {
...
+            .queryText(this.queryText)
+            .modelId(this.modelId)
+            .maxTokenScore(this.maxTokenScore)
+            .twoPhasePruneRatio(-1f);
+        if (this.queryTokensSupplier != null) {
+            Map<String, Float> tokens = queryTokensSupplier.get();
+            Map<Boolean, SetOnce<Map<String, Float>>> splitTokens = getSplitSetOnceByScoreThreshold(tokens, ratio);
+            this.queryTokensSupplier(splitTokens.get(true)::get);
+            copy.queryTokensSupplier(splitTokens.get(false)::get);
+        } else copy.queryTokensSupplier(new SetOnce<Map<String, Float>>()::get);
```
martin-gaievski, line 140:
```suggestion
        } else {
            copy.queryTokensSupplier(new SetOnce<Map<String, Float>>()::get);
        }    
```

src/test/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessorIT.java
```diff
@@ -0,0 +1,536 @@
...
+        }
+    }
+
+    @SneakyThrows
+    private void initializeTwoPhaseProcessor() {
+        createNeuralSparseTwoPhaseSearchProcessor(search_pipeline);
+    }
+
+    @SneakyThrows
+    private void updateIndexTwoPhaseProcessor(String indexName) {
```
martin-gaievski, line 106:
```suggestion
    private void setDefaultSearchPipelineForIndex(String indexName) {
```

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
martin-gaievski: if can keep correctness and improve performance I would vote for such approach. Also that aligns with one of the PRs goals which is improve query latency. Is it possible for you to run a benchmark to see how much latency improvement we can get if we reuse a single object?

martin-gaievski (commented):
Apart from discussion on new object vs single object and few minor comments related to formatting the code looks good to me.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,223 @@
...
+        String description,
+        boolean ignoreFailure,
+        boolean enabled,
+        float ratio,
+        float window_expansion,
+        int max_window_size
+    ) {
+        super(tag, description, ignoreFailure);
+        this.enabled = enabled;
+        if (ratio < 0f || ratio > 1f) {
```
zane-neo, line 59: Can we make these configuration optional with default value?

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,223 @@
...
+            .reduce(1.0f, (a, b) -> a * b);
+    }
+
+    private Multimap<NeuralSparseQueryBuilder, Float> collectNeuralSparseQueryBuilder(final QueryBuilder queryBuilder, float baseBoost) {
+        Multimap<NeuralSparseQueryBuilder, Float> result = ArrayListMultimap.create();
+
+        if (queryBuilder instanceof BoolQueryBuilder) {
+            BoolQueryBuilder boolQueryBuilder = (BoolQueryBuilder) queryBuilder;
+            float updatedBoost = baseBoost * boolQueryBuilder.boost();
+            for (QueryBuilder subQuery : boolQueryBuilder.should()) {
```
zane-neo, line 191: Do we need to consider other cases like must or filter?

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,223 @@
...
+     * @return request the search request that add the two-phase rescore query of neural sparse query.
+     */
+    @Override
+    public SearchRequest processRequest(final SearchRequest request) throws Exception {
+        if (!enabled || ratio == 0f) {
+            return request;
+        }
+        QueryBuilder queryBuilder = request.source().query();
+        // Collect the nested NeuralSparseQueryBuilder in the whole query.
+        Multimap<NeuralSparseQueryBuilder, Float> queryBuilderMap = ArrayListMultimap.create();
```
zane-neo, line 91: This line is redundant since in the `collectNeuralSparseQueryBuilder` there'll be a map been created in the beginning.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,223 @@
...
+        return TYPE;
+    }
+
+    /**
+     * Based on ratio, split a Map into two map by the value.
+     * @param queryTokens the queryTokens map, key is the token String, value is the score.
+     * @param ratio The ratio that control how tokens map be split.
+     * @return A map has two element, {[True, token map whose value above threshold],[False, token map whose value below threshold]}
+     */
+    public static Map<Boolean, SetOnce<Map<String, Float>>> getSplitSetOnceByScoreThreshold(
```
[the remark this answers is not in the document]
zane-neo: The method name is not self-explaining, please change to a better name.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
zane-neo: In which case the highScoreTks will be null? And why it's an IllegalArgumentException, also add error message to exceptions.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,223 @@
...
+
+        Map<Boolean, Map<String, Float>> queryTokensByScore = queryTokens.entrySet()
+            .stream()
+            .collect(
+                Collectors.partitioningBy(entry -> entry.getValue() >= threshold, Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue))
+            );
+        SetOnce<Map<String, Float>> highScoreTksSetOnce = new SetOnce<>(queryTokensByScore.get(Boolean.TRUE));
+        SetOnce<Map<String, Float>> lowScoreTksSetOnce = new SetOnce<>(queryTokensByScore.get(Boolean.FALSE));
+        if (highScoreTksSetOnce.get() == null) throw new IllegalArgumentException();
+        return Map.of(Boolean.TRUE, highScoreTksSetOnce, Boolean.FALSE, lowScoreTksSetOnce);
```
zane-neo, line 131: Consider if it's better to use meaningful string as key to make the code easier to read.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,223 @@
...
+        }
+        // We only support BoostQuery, BooleanQuery and NeuralSparseQuery now. For other compound query type which are not support now, will
+        // do nothing and just quit.
+        return result;
+    }
+
+    private RescorerBuilder<QueryRescorerBuilder> buildRescoreQueryBuilderForTwoPhase(final QueryBuilder nestedTwoPhaseQueryBuilder, final SearchRequest searchRequest){
+        RescorerBuilder<QueryRescorerBuilder> twoPhaseRescorer = new QueryRescorerBuilder(nestedTwoPhaseQueryBuilder);
+        int requestSize = searchRequest.source().size();
+        int windowSize = (int) ((requestSize == -1 ? 10 : requestSize) * window_expansion);
```
zane-neo, line 209: Looks like 10 is a magic number, making it a constant and add comments to it.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
conggguan: The optional and default value are in the Factor class.
```
            boolean enabled = ConfigurationUtils.readBooleanProperty(TYPE, tag, config, ENABLE_KEY, DEFAULT_ENABLED);
            Map<String, Object> twoPhaseConfigMap = ConfigurationUtils.readOptionalMap(TYPE, tag, config, PARAMETER_KEY);

            float ratio = DEFAULT_RATIO;
            float window_expansion = DEFAULT_WINDOW_EXPANSION;
            int max_window_size = DEFAULT_MAX_WINDOW_SIZE;
            if (Objects.nonNull(twoPhaseConfigMap)) {
                ratio = ((Number) twoPhaseConfigMap.getOrDefault(RATIO_KEY, ratio)).floatValue();
                window_expansion = ((Number) twoPhaseConfigMap.getOrDefault(EXPANSION_KEY, window_expansion)).floatValue();
                max_window_size = ((Number) twoPhaseConfigMap.getOrDefault(MAX_WINDOW_SIZE_KEY, max_window_size)).intValue();
            }
```

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
conggguan: Removed.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,223 @@
...
+     * @param ratio The ratio that control how tokens map be split.
+     * @return A map has two element, {[True, token map whose value above threshold],[False, token map whose value below threshold]}
+     */
+    public static Map<Boolean, SetOnce<Map<String, Float>>> getSplitSetOnceByScoreThreshold(
+        final Map<String, Float> queryTokens,
+        final float ratio
+    ) {
+        float max = 0f;
+        for (Float value : queryTokens.values())
+            max = Math.max(value, max);
```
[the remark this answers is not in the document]
conggguan: Done.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
conggguan: I think only when the model inference error or the parameter is null or empty, the highScoreTks will be null.
And a simple error msg now.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
conggguan: Use DOWN_THRESHOLD and UP_THRESHOLD

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
conggguan: Use constants for default value and camel case for variables.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
conggguan: Done.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
conggguan: As a discuss with @zhichao-aws earlier, we think reduce the other cases like must or filter will change the meaning of a query, so we just focus the should clause now.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
conggguan: Not a magic number, it's the default value of a query in opensearch now, make it to constant now.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
conggguan: Made change.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
conggguan: Change Done.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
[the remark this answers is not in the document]
conggguan: > This looks like a helper method, can we move it to utility class, create that class if needed

Do we need a helper class for only one function? I think this function will only run in the case of two-phase now.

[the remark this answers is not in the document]
conggguan: > The method name is not self-explaining, please change to a better name.

Change the name to splitQueryTokensByRatioedMaxScoreAsThreshold.

conggguan (commented):
For martin's concern about nested multiple neural sparse query.
I make a poc and do a benchmark test on a ec2 opensearch cluster.
The number 1,3,5 means I divided one neural sparse query into 1/3/5 neural sparse query nested in a boolean's should clauses, and get the P90 latency of them.
It shows that only a very high number case, we need to concern the cost of new object of neuralSparseQueryBuilder.
Besides, if we make a change the collect into 1 builder, we only have to consider the influence of query filename, and the case of same token with different score in different query. So I think keep the way that create a new sub queryBuilder for each valid neuralSparse query is more proper now.

| Way                                  | 1       | 3       | 5       |
| ------------------------------------ | ------- | ------- | ------- |
| Origin                               | 250.905 | 253.211 | 257.216 |
| TwoPhase with each have a sub object | 159.491 | 178.789 | 191.345 |
| TwoPhase all nested have 1 object    | 160.962 | 190.459 | 199.683 |

## Update
Later I will do a more accuracy test on multiple neural sparse query without model inference and get a final result.

## test result without model inference time
| Way                                  | 1       | 3       | 5       |
| ------------------------------------ | ------- | ------- | ------- |
| Origin                               |         |         |         |
| TwoPhase with each have a sub object | 163.836 | 176.413 | 191.046 |
| TwoPhase all nested have 1 object    | 159.885 | 189.109 | 197.972 |

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -270,40 +304,56 @@ private static void parseQueryParams(XContentParser parser, NeuralSparseQueryBui
...
+            List.of(queryText),
+            ActionListener.wrap(mapResultList -> {
+                Map<String, Float> queryTokens = TokenWeightUtil.fetchListOfTokenWeightMap(mapResultList).get(0);
+                if (Objects.nonNull(twoPhaseSharedQueryToken)) {
+                    Map<String, SetOnce<Map<String, Float>>> splitSetOnce = splitQueryTokensByRatioedMaxScoreAsThreshold(
+                        queryTokens,
+                        twoPhasePruneRatio
+                    );
+                    setOnce.set(splitSetOnce.get(UP_THRESHOLD).get());
+                    twoPhaseSharedQueryToken.putAll(splitSetOnce.get(DOWN_THRESHOLD).get());
```
zhichao-aws, line 337: Can we just assign the result Map to twoPhaseSharedQueryToken? The putAll needs to put all elements with o(N) complexity

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -76,7 +81,12 @@ public class NeuralSparseQueryBuilder extends AbstractQueryBuilder<NeuralSparseQ
     private String queryText;
     private String modelId;
     private Float maxTokenScore;
+    private Map<String, Float> twoPhaseSharedQueryToken;
     private Supplier<Map<String, Float>> queryTokensSupplier;
+    // A parameter that can detect if twoPhase are enabled, when twoPhasePruneRatio equals -1f, it means it two-phase rescoreQueryBuilder's
+    // subQueryBuilder.
+    private float twoPhasePruneRatio = 0F;
```
zhichao-aws, line 88: It's twoPhaseSharedQueryToken that used to determine if twoPhase are enabled, right? The comments here seems not very precise

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -270,40 +304,56 @@ private static void parseQueryParams(XContentParser parser, NeuralSparseQueryBui
...
     protected Query doToQuery(QueryShardContext context) throws IOException {
         final MappedFieldType ft = context.fieldMapper(fieldName);
         validateFieldType(ft);
-
         Map<String, Float> queryTokens = queryTokensSupplier.get();
-        validateQueryTokens(queryTokens);
-
+        if (Objects.isNull(queryTokens)) {
+            if (twoPhasePruneRatio == -1f) {
+                return new MatchNoDocsQuery();
```
zhichao-aws, line 353: Why for 2-phase query builder the query tokens can be null? From my understanding, if there is no eligible tokens, the query tokens should be empty instead of null

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -330,41 +380,42 @@ private static void validateFieldType(MappedFieldType fieldType) {
...
-        }
-    }
-
     @Override
     protected boolean doEquals(NeuralSparseQueryBuilder obj) {
         if (this == obj) return true;
         if (Objects.isNull(obj) || getClass() != obj.getClass()) return false;
         if (Objects.isNull(queryTokensSupplier) && Objects.nonNull(obj.queryTokensSupplier)) return false;
         if (Objects.nonNull(queryTokensSupplier) && Objects.isNull(obj.queryTokensSupplier)) return false;
+        if (Objects.nonNull(twoPhaseSharedQueryToken) && Objects.isNull(obj.twoPhaseSharedQueryToken)) return false;
```
zhichao-aws, line 389: We don't need to check null here. Line 401 is enough. EqualsBuilder.append checks the case that one of the object is null

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
zhichao-aws: In last PR we also supported BoostQuery, and use clause.isScoring() to determine whether to support the boolean clause. Are they same here?

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,246 @@
...
+            return 1.0f;
+        }
+        return searchSourceBuilder.rescores()
+            .stream()
+            .map(rescorerBuilder -> ((QueryRescorerBuilder) rescorerBuilder).getQueryWeight())
+            .reduce(1.0f, (a, b) -> a * b);
+    }
+
+    private Multimap<NeuralSparseQueryBuilder, Float> collectNeuralSparseQueryBuilder(final QueryBuilder queryBuilder, float baseBoost) {
+        Multimap<NeuralSparseQueryBuilder, Float> result = ArrayListMultimap.create();
```
zhichao-aws, line 206: Why we use Multimap here? Do we have multiple values for one field?

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,246 @@
...
+            throw new IllegalArgumentException("Query tokens cannot be null.");
+        }
+        return Map.of(UP_THRESHOLD, highScoreTokensSetOnce, DOWN_THRESHOLD, lowScoreTokensSetOnce);
+    }
+
+    /**
+     * Factory to create NeuralSparseTwoPhaseProcessor, provide default parameter,
+     *
+     */
+    public static class Factory implements Processor.Factory<SearchRequestProcessor> {
```
zhichao-aws, line 156: Can we move the Factory to the bottom of this class, and move other methods before where they are invoked?

src/test/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessorIT.java
```diff
@@ -0,0 +1,536 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Supplier;
+
+import static org.opensearch.neuralsearch.util.TestUtils.createRandomTokenWeightMap;
+import static org.opensearch.neuralsearch.util.TestUtils.objectToFloat;
+
+public class NeuralSparseTwoPhaseProcessorIT extends BaseNeuralSearchIT {
```
zhichao-aws, line 28: I think we should involve at least one IT test case that use query text. Because now we have different logics for query by text and query by tokens

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
conggguan: There's 3 case , main querybuilldr in 2-phase, sub querybuilder in 2-phase, main querybuilder without 2-phase, 
which 2 case will have a null twoPhaseSharedQueryToken. So twoPhasePruneRatio will help to determine.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
conggguan: Moved， but other method is private, maybe should after the invoke?

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
conggguan: If there has two same neuralsparse querybuilder in a compound query, will have same key.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
conggguan: Not same, what you see is a draft in last PR, the final version is determine which type. Because we don't want to change the must/must not/filter, but must isScoreing=True.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -113,6 +123,30 @@ public NeuralSparseQueryBuilder(StreamInput in) throws IOException {
...
     }
 
+    /**
+     * Copy this QueryBuilder for two phase rescorer, set the copy one's twoPhasePruneRatio to -1.
+     * @param ratio the parameter of the NeuralSparseTwoPhaseProcessor, control how to split the queryTokens to two phase.
+     * @return A copy NeuralSparseQueryBuilder for twoPhase, it will be added to the rescorer.
+     */
+    public NeuralSparseQueryBuilder getCopyNeuralSparseQueryBuilderForTwoPhase(float ratio) {
+        this.twoPhasePruneRatio(ratio);
+        NeuralSparseQueryBuilder copy = new NeuralSparseQueryBuilder().fieldName(this.fieldName)
```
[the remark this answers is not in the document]
conggguan: Add queryname, it's a miss.
For boost I compute out of this function for simple.
```
            NeuralSparseQueryBuilder neuralSparseQueryBuilder = (NeuralSparseQueryBuilder) queryBuilder;
            float updatedBoost = baseBoost * neuralSparseQueryBuilder.boost();
            NeuralSparseQueryBuilder modifiedQueryBuilder = neuralSparseQueryBuilder.getCopyNeuralSparseQueryBuilderForTwoPhase(ratio);
            result.put(modifiedQueryBuilder, updatedBoost);
```

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -270,40 +304,56 @@ private static void parseQueryParams(XContentParser parser, NeuralSparseQueryBui
...
-                ActionListener.wrap(mapResultList -> {
-                    queryTokensSetOnce.set(TokenWeightUtil.fetchListOfTokenWeightMap(mapResultList).get(0));
-                    actionListener.onResponse(null);
-                }, actionListener::onFailure)
-            ))
-        );
+        queryRewriteContext.registerAsyncAction(getModelInferenceAsync(queryTokensSetOnce));
         return new NeuralSparseQueryBuilder().fieldName(fieldName)
             .queryText(queryText)
+            .queryName(queryName)
```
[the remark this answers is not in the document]
conggguan: Indeed, I remove them.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
conggguan: Ok, I make it into assign now.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
conggguan: thanks, removed.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,246 @@
...
+        return TYPE;
+    }
+
+    /**
+     * Based on ratio, split a Map into two map by the value.
+     * @param queryTokens the queryTokens map, key is the token String, value is the score.
+     * @param thresholdRatio The ratio that control how tokens map be split.
+     * @return A map has two element, {[True, token map whose value above threshold],[False, token map whose value below threshold]}
+     */
+    public static Map<String, SetOnce<Map<String, Float>>> splitQueryTokensByRatioedMaxScoreAsThreshold(
```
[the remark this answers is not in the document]
conggguan: Change to map now,
For the String, zan said maybe it's easy to understand,
But tuple are more light, so I change to tuple and add comments to get is more clear.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
conggguan: But if we make low score tokens always not null, we can not consider using this filed to determine any more.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
conggguan: Because the splitquerytoken function will generate null if any token's score are high.
I make a check to that functon, now when all token are high score, it will return empty map.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: > which 2 case will have a null twoPhaseSharedQueryToken. So twoPhasePruneRatio will help to determine.

Do you mean sub querybuilder in 2-phase, main querybuilder without 2-phase? Do we have any logics handle them differently here?

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
zhichao-aws: OK, I think putting the private methods after the public ones is also fine

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
conggguan: Yes, twoPhasePruneRatio
1.0f-> 2-phase disabled
(0f,1f) -> 2-phase enabled and this is the main neural sparse querybuilder
(-1f,-0f) -> 2-phase enabled and this is the sub neural sparse querybuilder.

zhichao-aws: > 2-phase enabled and this is the sub neural sparse querybuilder
For this case it's always -1 right? Why can it has value from (-1f,-0f)

I understand it can distinguish this case, but why we need to distinguish them here? From the two phase processor we can distinguish them without this twoPhasePruneRatio. From the NeuralSparseQueryBuilder I don't  find the logics we need to handle them differently

conggguan: Yes, we don't need to distinguish now, but we need this field to split query tokens in rewrite, and I think make it carry more information maybe better? 
For example, if in a future version, we can find neural sparse query which not be accelerated, whose ratio is 1.0, can make log very easy. And for further need, I think carry more information is better?
Do you think it may make confuse for developer?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -270,40 +304,56 @@ private static void parseQueryParams(XContentParser parser, NeuralSparseQueryBui
     }
 
     @Override
-    protected QueryBuilder doRewrite(QueryRewriteContext queryRewriteContext) throws IOException {
-        // We need to inference the sentence to get the queryTokens. The logic is similar to NeuralQueryBuilder
```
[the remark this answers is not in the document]
conggguan: Sorry, because I made some change in this function and miss the origin comments, now add a new comments.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: I think we can add more comments about this. e.g. the meaning of each cases, and we don't need these information now, but keep them for future extension etc

zhichao-aws: Current comments only mentions twoPhasePruneRatio, but in line 333 it checks twoPhaseSharedQueryToken for whether 2-phase is enbaled. It's confusing for me

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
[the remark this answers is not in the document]
zhichao-aws: Seems we still remove one line of comments.

And we'd better add more comments about the new logics in rewrite for 2-phase search

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: > 1.0f-> 2-phase disabled

But the default value is 0F?

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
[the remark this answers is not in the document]
zhichao-aws: Please also update the comments for this method

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: Seems unneccessary to wrap it in a brace like `(splitQueryTokens.v2())`

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,250 @@
...
+
+        Map<Boolean, Map<String, Float>> queryTokensByScore = queryTokens.entrySet()
+            .stream()
+            .collect(
+                Collectors.partitioningBy(entry -> entry.getValue() >= threshold, Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue))
+            );
+
+        Map<String, Float> highScoreTokens = queryTokensByScore.get(Boolean.TRUE);
+        Map<String, Float> lowScoreTokens = queryTokensByScore.get(Boolean.FALSE);
+        if (Objects.isNull(highScoreTokens) || highScoreTokens.isEmpty()) {
```
zhichao-aws, line 146: I think we only need to check the highScoreTokens is null, and set to empty when it's null. We don't need to throw exception when it's empty.

Because we can't exclude the case the model can generate empty sparse vector for input text. Although it's rare, but still possible

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: But the check is not changed here, right?
I think we may need to restore to the origin check `validateQueryTokens`

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: I mean we can directly use this
```
EqualsBuilder equalsBuilder = new EqualsBuilder().append(fieldName, obj.fieldName)
            .append(queryText, obj.queryText)
            .append(modelId, obj.modelId)
            .append(maxTokenScore, obj.maxTokenScore)
            .append(twoPhasePruneRatio, obj.twoPhasePruneRatio)
            .append(twoPhaseSharedQueryToken, obj.twoPhaseSharedQueryToken);
```

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
conggguan: > > 1.0f-> 2-phase disabled
> 
> But the default value is 0F?

0.0f-> 2-phase disabled
(0f,1f) -> 2-phase enabled and this is the main neural sparse querybuilder
-1f -> 2-phase enabled and this is the sub neural sparse querybuilder.
Sorry, this is the logic,
And I make some change to let it carry more information.
0.0f-> 2-phase disabled
ratio value (0f,1f) -> 2-phase enabled and this is the main neural sparse querybuilder
-1*ratio value (-1f,0f) -> 2-phase enabled and this is the sub neural sparse querybuilder.

And for your confusing, these information are not required now, the ratio just work for split tokens.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
conggguan: remove the redundant.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
conggguan: OK, made change  as this.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
conggguan: I think the origin validateQueryTokens is a little redundant because of there has a check on newLinearQuery().
```
public static Query newLinearQuery(String fieldName, String featureName, float weight) {
    if (!(weight <= 0.0F) && !(weight > 64.0F)) {
        Query q = new FeatureQuery(fieldName, featureName, new LinearFunction());
        if (weight != 1.0F) {
            q = new BoostQuery((Query)q, weight);
        }

        return (Query)q;
    } else {
        throw new IllegalArgumentException("weight must be in (0, 64.0], got: " + weight);
    }
}
```
So maybe a null check is enough?

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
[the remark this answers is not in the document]
conggguan: Updated. 
```
 @return A tuple has two element, { token map whose value above threshold, token map whose value below threshold }
```

zhichao-aws (approved):
lgtm

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -361,10 +421,17 @@ protected boolean doEquals(NeuralSparseQueryBuilder obj) {
...
-        HashCodeBuilder builder = new HashCodeBuilder().append(fieldName).append(queryText).append(modelId).append(maxTokenScore);
+        HashCodeBuilder builder = new HashCodeBuilder().append(fieldName)
+            .append(queryText)
+            .append(modelId)
+            .append(maxTokenScore)
+            .append(twoPhasePruneRatio);
         if (queryTokensSupplier != null) {
             builder.append(queryTokensSupplier.get());
         }
+        if (twoPhaseSharedQueryToken != null) {
```
zhichao-aws, line 432: We don't need to do the null check here

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -77,6 +80,17 @@ public class NeuralSparseQueryBuilder extends AbstractQueryBuilder<NeuralSparseQ
     private String modelId;
     private Float maxTokenScore;
     private Supplier<Map<String, Float>> queryTokensSupplier;
+    // A filed that for neural_sparse_two_phase_processor, if twoPhaseSharedQueryToken is not null,
```
martin-gaievski, line 83: nit:
```suggestion
    // A field that for neural_sparse_two_phase_processor, if twoPhaseSharedQueryToken is not null,
```

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -0,0 +1,251 @@
...
+            Map<String, Object> config,
+            PipelineContext pipelineContext
+        ) throws IllegalArgumentException {
+
+            boolean enabled = ConfigurationUtils.readBooleanProperty(TYPE, tag, config, ENABLE_KEY, DEFAULT_ENABLED);
+            Map<String, Object> twoPhaseConfigMap = ConfigurationUtils.readOptionalMap(TYPE, tag, config, PARAMETER_KEY);
+
+            float ratio = DEFAULT_RATIO;
+            float window_expansion = DEFAULT_WINDOW_EXPANSION;
+            int max_window_size = DEFAULT_MAX_WINDOW_SIZE;
```
martin-gaievski, line 240: please use camelcase for variable names

martin-gaievski (commented):
We need a bwc for this change, are we covered by anything that already exists or we need a new test?

conggguan:
> We need a bwc for this change, are we covered by anything that already exists or we need a new test?

I think we need a BWC  test, and it would be better to perform it after this code is merged. Currently, I can't build a BWC test to invoke the code from this PR since it hasn't been merged yet. I will add a BWC test as soon as this PR is merged.

Is this a good solution?

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
zane-neo: I thought this feature was a performance improvement, can you add comments to explain why it's changing the meaning of a query in this PR?

conggguan: Before invoking getCopyNeuralSparseQueryBuilderForTwoPhase I add some comments, you can take a look for it's and give me a feedback if this comments are proper.

martin-gaievski:
> > We need a bwc for this change, are we covered by anything that already exists or we need a new test?
> 
> I think we need a BWC test, and it would be better to perform it after this code is merged. Currently, I can't build a BWC test to invoke the code from this PR since it hasn't been merged yet. I will add a BWC test as soon as this PR is merged.
> 
> Is this a good solution?

That works, although I'm not sure what issue you're facing as bwc should be able to use code from active PR. Please make sure the PR with BWC is merged in the same release, which in case of this change is 2.15.

martin-gaievski (approved):

PR #749, opened 2024-05-15
Parallelize hybrid query processor

### LINKED ISSUE #279: [FEATURE] Implement parallel execution of sub-queries for hybrid search
Currently individual queries of Hybrid query (aka sub-queries) are executed sequentially (Initial implementation that is done under https://github.com/opensearch-project/neural-search/issues/123). That may be sub-optimal as system may waste time on waiting.

Obvious approach for such execution is to run every sub-query execution using parallel thread and wait for results from all thread, this will be at shard level. Results then will be combined and sent to coordinator node all at once, as it's done today.

Sub-queries may be executed in parallel. Actual improvement must be verified by running benchmarks on sequential and parallel approaches.

VijayanB:
### Description
1. Add new thread pool to schedule tasks that are related to hybrid query execution
2. Register executor builders with Plugin
3. Use Lucene's Task Executor to execute and collect results
4. Parallelize Query re-write
5. Parallelize score supplier creation

### Issues Resolved
#279 

### Check List
    - [x] All tests pass


src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutor.java
```diff
@@ -0,0 +1,63 @@
...
+import lombok.NonNull;
+import org.apache.lucene.search.TaskExecutor;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.common.util.concurrent.OpenSearchExecutors;
+import org.opensearch.threadpool.ExecutorBuilder;
+import org.opensearch.threadpool.FixedExecutorBuilder;
+import org.opensearch.threadpool.ThreadPool;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class HybridQueryExecutor {
```
navneet1v, line 19: if this class should not be extended by anyone, lets make this class as final

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
```diff
@@ -106,22 +111,40 @@ public Query rewrite(IndexSearcher indexSearcher) throws IOException {
             return new MatchNoDocsQuery("empty HybridQuery");
         }
 
-        boolean actuallyRewritten = false;
-        List<Query> rewrittenSubQueries = new ArrayList<>();
+        HybridQueryRewriteCollectorManager manager = new HybridQueryRewriteCollectorManager(indexSearcher);
+        List<Callable<Void>> queryRewriteTasks = new ArrayList<>();
+        List<HybridQueryExecutorCollector<IndexSearcher, Map.Entry<Query, Boolean>>> collectors = new ArrayList<>();
```
navneet1v, line 116: lets make all the variable final here. We should not allow anyone to change references of these variables.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryRewriteCollectorManager.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Returns true if any of the {@link Query} from collector were actually rewritten.
+     * @param collectors
+     * @return at least one query is rewritten by any of the collectors
+     */
+    public Boolean anyQueryRewrite(List<HybridQueryExecutorCollector<IndexSearcher, Map.Entry<Query, Boolean>>> collectors) {
+        // return true if at least one query is rewritten
+        return collectors.stream().anyMatch(collector -> collector.getResult().getValue());
+    }
```
navneet1v, line 56: I would recommend avoiding streams here as we have seen earlier streams on complex objects tend to take more time during execution. 

Ref: https://github.com/opensearch-project/neural-search/pull/706

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
```diff
@@ -106,22 +111,40 @@ public Query rewrite(IndexSearcher indexSearcher) throws IOException {
...
 
-        if (actuallyRewritten) {
-            return new HybridQuery(rewrittenSubQueries);
+        HybridQueryExecutor.getExecutor().invokeAll(queryRewriteTasks);
+
+        final boolean anyQueryRewrite = manager.anyQueryRewrite(collectors);
+        if (anyQueryRewrite == false) {
+            return super.rewrite(indexSearcher);
         }
+        final List<Query> rewrittenSubQueries = manager.getRewriteQueries(collectors);
```
navneet1v, line 129: how we are ensuring that all the queries are rewritten?

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
```diff
@@ -106,22 +111,40 @@ public Query rewrite(IndexSearcher indexSearcher) throws IOException {
...
+    private Void rewriteQuery(Query query, HybridQueryExecutorCollector<IndexSearcher, Map.Entry<Query, Boolean>> collector) {
+        collector.collect(indexSearcher -> {
+            try {
+                Query rewrittenQuery = query.rewrite(indexSearcher);
+                /* we keep rewrite sub-query unless it's not equal to itself, it may take multiple levels of recursive calls
+                queries need to be rewritten from high-level clauses into lower-level clauses because low-level clauses
+                perform better. For hybrid query we need to track progress of re-write for all sub-queries */
+
+                boolean actuallyRewritten = rewrittenQuery != query;
+                return new AbstractMap.SimpleEntry(rewrittenQuery, actuallyRewritten);
```
navneet1v, line 142: can we avoid storing queries in a map? and who is going to store this simple map entry?

navneet1v (commented):
Looked PR at high level and added comments. Please resolve them after that I will do 1 more review of the PR.

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutorCollectorManager.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.executors;
+
+/**
+ * {@link HybridQueryExecutorCollectorManager} is responsible for creating new {@link HybridQueryExecutorCollector} instances
+ */
+public interface HybridQueryExecutorCollectorManager<C extends HybridQueryExecutorCollector, R> {
```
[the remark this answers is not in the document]
VijayanB: Good catch. I previously added another method and used R as return type and then later i removed the method but forgot to remove that.

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutor.java
```diff
@@ -0,0 +1,63 @@
...
+        );
+    }
+
+    /**
+     * Initialize @{@link TaskExecutor} to run tasks concurrently using {@link ThreadPool}
+     * @param threadPool OpenSearch's thread pool instance
+     */
+    public static void initialize(@NonNull ThreadPool threadPool) {
+        if (threadPool == null) {
+            throw new IllegalArgumentException("Argument thread-pool cannot be null");
```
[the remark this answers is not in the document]
VijayanB: Ack

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutor.java
VijayanB: Ack

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
VijayanB: Ack

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutor.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Return TaskExecutor Wrapper that helps runs tasks concurrently
+     * @return TaskExecutor instance to help run search tasks in parallel
+     */
+    public static TaskExecutor getExecutor() {
+        return taskExecutor != null ? taskExecutor : new TaskExecutor(Runnable::run);
+    }
+
+    @VisibleForTesting
+    public static String getThreadPoolName() {
```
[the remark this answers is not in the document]
VijayanB: Ack

src/main/java/org/opensearch/neuralsearch/query/HybridQueryRewriteCollectorManager.java
VijayanB: Ack

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
VijayanB: if query rewrite is not done, you will have Query as null object in the list.

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
VijayanB: HybridQueryExecutorCollector.collect() will store this entry. If we don't store map, we can't assemble back to caller once the execution is completed.

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutorCollector.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.function.Function;
+
+/**
+ * {@link HybridQueryExecutorCollector} is a generic Collector used by Hybrid Search Query during
+ * Query phase to parallelize sub query's action to improve latency
+ */
+@RequiredArgsConstructor(staticName = "newCollector")
+public class HybridQueryExecutorCollector<I, R> {
+
+    private final I param;
```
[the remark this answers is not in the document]
VijayanB: if collector requires inputs, this will be used. ex: IndexReader, LeafReader, etc ..

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutor.java
```diff
@@ -0,0 +1,63 @@
...
+     * Provide fixed executor builder to use for hybrid query executors
+     * @param settings Node level settings
+     * @return the executor builder for hybrid query's custom thread pool.
+     */
+    public static ExecutorBuilder getExecutorBuilder(final Settings settings) {
+        final int allocatedProcessors = OpenSearchExecutors.allocatedProcessors(settings);
+        return new FixedExecutorBuilder(
+            settings,
+            HYBRID_QUERY_EXEC_THREAD_POOL_NAME,
+            allocatedProcessors,
```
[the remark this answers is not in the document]
VijayanB: Good point. Changing to twice default allocated processor.

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
VijayanB: The below command make sure that all tasks are completed `HybridQueryExecutor.getExecutor().invokeAll(queryRewriteTasks);`

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
shatejas: You can use optional instead.
```
return rewrittenQuery != query ? Optional.empty() : Optional.of(rewrittenQuery)
```

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
```diff
@@ -106,22 +111,40 @@ public Query rewrite(IndexSearcher indexSearcher) throws IOException {
...
+        final List<Callable<Void>> queryRewriteTasks = new ArrayList<>();
+        final List<HybridQueryExecutorCollector<IndexSearcher, Map.Entry<Query, Boolean>>> collectors = new ArrayList<>();
         for (Query subQuery : subQueries) {
-            Query rewrittenSub = subQuery.rewrite(indexSearcher);
-            /* we keep rewrite sub-query unless it's not equal to itself, it may take multiple levels of recursive calls
-               queries need to be rewritten from high-level clauses into lower-level clauses because low-level clauses
-               perform better. For hybrid query we need to track progress of re-write for all sub-queries */
-            actuallyRewritten |= rewrittenSub != subQuery;
-            rewrittenSubQueries.add(rewrittenSub);
+            final HybridQueryExecutorCollector<IndexSearcher, Map.Entry<Query, Boolean>> collector = manager.newCollector();
```
shatejas, line 118: Just for my understanding, why is a collector chosen here? you could have plugged in TaskExecutor here and then simply use the list of results returned by invokeAll. The entire collector seems to be a round about way to get the results. Manager would have been useful if you wanted to return a different collector based on certain conditions IMO

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutor.java
```diff
@@ -0,0 +1,77 @@
...
+            );
+        }
+        taskExecutor = new TaskExecutor(threadPool.executor(HYBRID_QUERY_EXEC_THREAD_POOL_NAME));
+    }
+
+    /**
+     * Return TaskExecutor Wrapper that helps runs tasks concurrently
+     * @return TaskExecutor instance to help run search tasks in parallel
+     */
+    public static TaskExecutor getExecutor() {
```
chishui, line 69: Do we expect `initialize()` to always be called before this `getExecutor()`? If so, should you create an instance when it's null? Why not just return null and fail fast?

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutorCollector.java
```diff
@@ -0,0 +1,26 @@
...
+import lombok.RequiredArgsConstructor;
+
+import java.util.function.Function;
+
+/**
+ * {@link HybridQueryExecutorCollector} is a generic Collector used by Hybrid Search Query during
+ * Query phase to parallelize sub query's action to improve latency
+ */
+@RequiredArgsConstructor(staticName = "newCollector")
+public final class HybridQueryExecutorCollector<I, R> {
```
chishui, line 17: Is `HybridQueryExecutorCollector` thread safe?

src/main/java/org/opensearch/neuralsearch/query/HybridQueryRewriteCollectorManager.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Returns list of {@link Query} that were rewritten by collectors
+     * @param collectors
+     * @return
+     */
+    public List<Query> getRewriteQueries(List<HybridQueryExecutorCollector<IndexSearcher, Map.Entry<Query, Boolean>>> collectors) {
+        List<Query> rewrittenQueries = new ArrayList<>();
+        for (HybridQueryExecutorCollector<IndexSearcher, Map.Entry<Query, Boolean>> collector : collectors) {
+            Query key = collector.getResult().getKey();
```
chishui, line 45: `getResult()` could be null, please do a null check before using it (also apply to usages from other places)

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
VijayanB: Good question. The main idea is to move logic of collecting and reducing/merging into its own manager. By introducing TaskExecutor inside Scorer/Weight we are coupling how to parallelize with what needs to be parallelized. With introduction of Collector and Collector Manager, we decoupled parallelization from logic. IMO, this abstracts parallelization implementation from all of its caller.

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutor.java
```diff
@@ -0,0 +1,77 @@
...
+import lombok.experimental.PackagePrivate;
+import org.apache.lucene.search.TaskExecutor;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.common.util.concurrent.OpenSearchExecutors;
+import org.opensearch.threadpool.ExecutorBuilder;
+import org.opensearch.threadpool.FixedExecutorBuilder;
+import org.opensearch.threadpool.ThreadPool;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class HybridQueryExecutor {
```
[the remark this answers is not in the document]
VijayanB: Ack

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutor.java
```diff
@@ -0,0 +1,77 @@
...
+    private static final String HYBRID_QUERY_EXEC_THREAD_POOL_NAME = "_plugin_neural_search_hybrid_query_executor";
+    private static final Integer HYBRID_QUERY_EXEC_THREAD_POOL_QUEUE_SIZE = 1000;
+    private static TaskExecutor taskExecutor;
+
+    /**
+     * Will allocate twice the default allocated processor. To avoid 0, we will return 2 as minimum
+     * processor count
+     */
+    private static int allocateTwiceDefaultProcessors(final Settings settings) {
```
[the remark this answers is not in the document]
VijayanB: Ack

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutor.java
```diff
@@ -0,0 +1,77 @@
...
+        final int allocatedProcessors = OpenSearchExecutors.allocatedProcessors(settings);
+        int processorCount = Math.max(2 * allocatedProcessors, 2);
+        return Math.min(processorCount, HYBRID_QUERY_EXEC_THREAD_POOL_QUEUE_SIZE);
```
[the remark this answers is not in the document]
VijayanB: Good point. Will change it to Integer.Max as maximum value.

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutor.java
```diff
@@ -0,0 +1,77 @@
...
+            HYBRID_QUERY_EXEC_THREAD_POOL_NAME
+        );
+    }
+
+    /**
+     * Initialize @{@link TaskExecutor} to run tasks concurrently using {@link ThreadPool}
+     * @param threadPool OpenSearch's thread pool instance
+     */
+    public static void initialize(@NonNull ThreadPool threadPool) {
+        if (threadPool == null) {
```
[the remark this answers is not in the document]
VijayanB: Ack. Removed NonNull.

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutor.java
```diff
@@ -0,0 +1,77 @@
...
+            throw new IllegalArgumentException(
+                "Argument thread-pool to Hybrid Query Executor cannot be null."
```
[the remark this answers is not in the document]
VijayanB: Not necessary. This is more of message to caller to understand what went wrong. Haven't seen where it is coupled with actual variable name and class instances.

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutor.java
VijayanB: Yes, we expect intialize() to be called to set Executor, since, it is good practice to use executor to reuse threads instead of creating one thread for every new task. However, if someone doesn't have thread-pool, we don't want to fail. Instead, we are allowing default implementation of where you create new thread for execution.

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
shatejas: Lets dissect this a little more. There are 3 classes here:
1. HybridQueryExecutor: responsible for initializing and holding a common threadpool for query, scorer and weight,
2. HybridQueryExecutorCollectorManager: Responsible for giving a new collector
3. HybridQueryExecutorCollector : Holds the query results

> With introduction of Collector and Collector Manager, we decoupled parallelization from logic. IMO, this abstracts parallelization implementation from all of its caller.

HybridQueryExecutor seems to be a good wrapper to make sure the same task executor is used. I am more curious about the additional value of `HybridQueryExecutorCollectorManager` and `HybridQueryExecutorCollector`

Why can't it be
```
final List<Callable<Entry<Query, Boolean>>> queryRewriteTasks = new ArrayList<>();
// for each subquery
queryRewriteTasks.add(() -> rewriteQuery(subQuery); //rewrite needs to return entry<query, boolean>;
List<Entry<Query, Boolean>> rewrittenQueries = HybridQueryExecutor.getExecutor().invokeAll(queryRewriteTasks);

// rest of the logic
```
This way you don't have to worry about the threadsafety of the collectors. There are some utility methods in collector manager which arent related to managing a collector but more of related to rewritten queries itself.

Overall while this works fine, I want to make sure these interfaces have a clear definition and its clear on how to use them. Currently it seems like we can simplify this by not having it unless I am misunderstanding the value it brings

src/main/java/org/opensearch/neuralsearch/query/HybridQueryRewriteCollectorManager.java
VijayanB: Ack

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutorCollector.java
VijayanB: There are no setters and only way to update result is by calling collect method. Do you have any particular concerns?

chishui: Since `collect()` could set `result`, if `collect()` and `getResults()` are called asynchronously, technically, there could be race condition, right?

src/main/java/org/opensearch/neuralsearch/query/HybridQueryRewriteCollectorManager.java
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Returns true if any of the {@link Query} from collector were actually rewritten.
+     * If any of the given collector doesn't have result, it will be ignored as if that
+     * instance did not exist. This method must be called after collection is finished
+     * on all provided collectors.
+     * @param collectors List of collectors to check any of their query was rewritten during
+     *                   collect step.
+     * @return at least one query is rewritten by any of the collectors
+     */
+    public Boolean anyQueryRewrite(List<HybridQueryExecutorCollector<IndexSearcher, Map.Entry<Query, Boolean>>> collectors) {
```
martin-gaievski, line 68: why wrapper `Boolean` class and not primitive type? At the caller you're unboxing that to `boolean`, so that's not need, plus it's slower.

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
martin-gaievski: and +1 to Navneet's point regarding map and specifically using Query as a key. I was using similar approach in initial implementation of hybrid query, problem is that for some queries `hashCode` can be slow. 
How much time this will be called - one time per sub-query x num of shards?

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutorCollector.java
VijayanB: You are right. Currently getResult() is called from collector manager, where the contract is that it can be called only after all collectors are finished collection. This was taken care in the implementation. I can add this note in collector  as well. i am reluctant to synchronize to result, since it will add additional latency for scenario that is not possible at this moment. Do you have any other suggestions?

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
VijayanB: Used optional to avoid null condition, but we need query irrespective of whether query is re-written or not.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryRewriteCollectorManager.java
VijayanB: Ack

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
VijayanB: @martin-gaievski We are using list to store the result. I don't believe hashcode will be calculated while adding it to list.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryRewriteCollectorManager.java
martin-gaievski: you probably can return `boolean`, will save some cycles and avoids potential NPE if return is `null`

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
martin-gaievski: understood, I mismatch the Map.Entry<Query, Boolean> with Map<Query, Boolean>, I guess the Map.Entry is used to operate on pair of objects. I'm good now, thanks Vijay

martin-gaievski: > Used optional to avoid null condition, but we need query irrespective of whether query is re-written or not.

We can explicitly check for null, my point was performance. Doing first `Optional.of(rewrittenQuery)`, and later `optional.get()` both will add some cycles of wrapping query object into Optional and then unwrapping it. If we have done benchmarks with Optional and without it then we can use it as datapoint to make a decision.

VijayanB: Sounds good. I will run benchmark with and without before merging into main. In the meantime, can we keep Optional? What do you think?

martin-gaievski: Yes, I'm not sure what is the delta in performance if any, let's keep it with Optional, run benchmark and act per per data

shatejas: the intent was never to have an Optional as a member variable, the intent was to only return optional to indicate the value is not rewritten. It got misunderstood.

I realize that the query is needed irrespective of if its rewritten so Optional cannot be used. An `Map.Entry` might be a good option here but will leave it up to you.

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutorCollector.java
shatejas: Adding a comment is a good intention. you won't be able to enforce it. You are relying on the hope that the code doesn't change in future to hit that race condition so there will always be a risk.

If you really want to enforce; your options here would be to synchronize or move away from `result collection` to `returning results` as pointed out in [this thread](https://github.com/opensearch-project/neural-search/pull/749/files#r1608714432)

chishui: > i am reluctant to add synchronize to result variable, since it will add additional latency for scenario that is not possible at this moment

Actually, the overhead of synchronization is way less than we thought, here are some synchronization benchmark articles I found, 
- https://baptiste-wicht.com/posts/2010/09/java-synchronization-mutual-exclusion-benchmark.html
- https://isuru-perera.blogspot.com/2016/05/benchmarking-java-locks-with-counters.html
One operation of "synchronized" is < 1 micro second, considering the time spending on actual query logic with intensive I/O, this is negligible and you have simple implementation with concurrent situation covered.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryRewriteCollectorManager.java
VijayanB: Ack

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutor.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * Will allocate twice the default allocated processor. To avoid 0, we will return 2 as minimum
+     * processor count
+     */
+    private static int allocateTwiceDefaultProcessors(final Settings settings) {
+        final int allocatedProcessors = OpenSearchExecutors.allocatedProcessors(settings);
+        int processorCount = Math.max(2 * allocatedProcessors, 2);
+        return Math.min(processorCount, Integer.MAX_VALUE);
```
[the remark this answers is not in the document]
VijayanB: 500 ( 2 * 250 ) as max seems less. If someone sets egregious processor settings, it is going to impact all the thread queue since we use this setting everywhere. How about 1000? I don't have data point for this number. We can add comment that this is chosen to limit size.

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutorCollector.java
VijayanB: @shatejas @chishui Fair enough. I am changing to synchronized, however while benchmarking i am planning to compare with and without synchronization to see that it is not adding any latency to this block.

VijayanB:
@navneet1v Can you take a look at one more time? Thanks

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutorCollector.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.Optional;
+import java.util.function.Function;
+
+/**
+ * {@link HybridQueryExecutorCollector} is a generic Collector used by Hybrid Search Query during
+ * Query phase to parallelize sub query's action to improve latency
+ */
+@RequiredArgsConstructor(staticName = "newCollector")
+public final class HybridQueryExecutorCollector<I, R> {
```
martin-gaievski, line 19: Did you explore other names for this abstraction? I have doubt about the 'Collector` suffix in the `HybridQueryExecutorCollector` name, it sounds similar to Lucene [Collector](https://lucene.apache.org/core/9_0_0/core/org/apache/lucene/search/Collector.html) which is in use for hybrid query, but your class is different. So to avoid ambiguity I would rename this interface and related classes with something without "Collector". Few possible examples for name prefix: HybridQueryExecutorJob or HybridQueryExecutorTask.

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutorCollectorManager.java
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.executors;
+
+/**
+ * {@link HybridQueryExecutorCollectorManager} is responsible for creating new {@link HybridQueryExecutorCollector} instances
+ */
+public interface HybridQueryExecutorCollectorManager<C extends HybridQueryExecutorCollector> {
+    C newCollector();
```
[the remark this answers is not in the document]
VijayanB: Ack

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutorCollector.java
```diff
@@ -0,0 +1,33 @@
...
+    // will be used as input for all instances of collector generated by newCollector method,
+    // if it is required for collect operation
+    private final I param;
+
+    // getResult should only be called after collector's collect method is invoked.
+    @Getter(onMethod_ = { @Synchronized })
+    private Optional<R> result = Optional.empty();
+
+    @Synchronized
+    public void collect(Function<I, R> action) {
```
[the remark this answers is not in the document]
VijayanB: Ack

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutorCollector.java
VijayanB: I choose this suffix so that it follows Lucene's Collector/Manager pattern. This suffix is commonly used by every feature that wants to convey Manager/Entity pattern where collector is an entity which will be used only to collect the result .

martin-gaievski: Understood, just in Lucene and thus OpenSearch Collector and CollectorManager is mainly used in context of collecting doc scores. Maybe I was looking into that doc collector code for too long :) I would change the class name, but I'm leaving the final decision to you.

src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutorCollector.java
```diff
@@ -0,0 +1,37 @@
...
+@RequiredArgsConstructor(staticName = "newCollector")
+public final class HybridQueryExecutorCollector<I, R> {
+
+    // will be used as input for all instances of collector generated by newCollector method,
+    // if it is required for collect operation
+    private final I param;
+
+    // getResult should only be called after collector's collect method is invoked.
+    @Getter(onMethod_ = { @Synchronized })
+    private Optional<R> result = Optional.empty();
```
martin-gaievski, line 27: we're saving some state in collector. Is it guaranteed that same collector instance is not used second time before this saved state (result) is read by the caller?
The way it's done in core for QueryContext and Collectors - they put Collector Manager into same package with Collector classes, everything in Collector is package private, only CollectorManager methods are public. This way they shielded from scenario when someone reuse same instance of Collector and state is corrupted. 
This also opens up possibility for us to reuse Collector instances in future, if some of them can be redesigned to be stateless.

VijayanB: Yes @martin-gaievski Every-time manager will create new instance when newCollector method is called. Let me move collector manager and collector in same package and make it package private if this helps.

martin-gaievski: I think separate package for executors is good, I think even before your latest commit we need to cleanup the `query` package. Can you please do the other way around - restore that `executors` package and keep (or move) executor specific classes in that package. Then we can have `newCollector` method of Manager class as public. It's ok to have Collector.getResult public, Optional is immutable so client should not be able to change it.

VijayanB: Ack.

martin-gaievski (approved):
Looks good to me, thank you Vijay.
Please plan to add unit tests for the new classes from the executor package. We can do it in a separate PR.

naveentatikonda (approved):

Issue #764, opened 2024-05-27
[PROPOSAL] Optimize max score calculation in Hybrid Search


vibrantvarun:
### What are you proposing?
The max score calculation per shard in the Query Phase of hybrid search is done by iterating over the score-docs (results), instead we can just keep a track of max score at the collector level and  update its value when the score is calculated. 

Current logic is 
```
private float getMaxScore(final List<TopDocs> topDocs) {
        if (topDocs.isEmpty()) {
            return 0.0f;
        } else {
            return topDocs.stream()
                .map(docs -> docs.scoreDocs.length == 0 ? new ScoreDoc(-1, 0.0f) : docs.scoreDocs[0])
                .map(scoreDoc -> scoreDoc.score)
                .max(Float::compare)
                .get();
        }
    }
```
where topDocs represent the result found from different subqueries in that particular shard.

Instead we can add a variable
```
private int maxScore=0.0f;
```
and later in the [collect](https://github.com/opensearch-project/neural-search/blob/afd1215a930a14cb889ea8751997a8a18cce5d1a/src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java#L94) method where hybrid scores are [calculated](https://github.com/opensearch-project/neural-search/blob/afd1215a930a14cb889ea8751997a8a18cce5d1a/src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java#L98).

Add a condition in [here](https://github.com/opensearch-project/neural-search/blob/afd1215a930a14cb889ea8751997a8a18cce5d1a/src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java#L118)
 ```
if(score>maxScore){
   maxScore=score;
}
```

By this way we will reduce the latency by

**(Number of Shards) x (Number of topDocs found on each shard)**

### What problems are you trying to solve?
Reducing the latency in Hybrid Search

### What is the developer experience going to be?
No changes in the developer experience.

#### Are there any security considerations? 
Nope

#### Are there any breaking changes to the API
No breaking changes

martin-gaievski:
It's good idea @vibrantvarun, we're calculating maxScore at the coordinator level today, we should be able to skip same computation at the shard level.
In your PR https://github.com/opensearch-project/neural-search/pull/765 you changed the logic at the shard level, not skipped it. Can you clarify why we kept the logic at shard level?

vibrantvarun:
[...]

PR #765, opened 2024-05-27
Optimize the max score tracking in the Query Phase of Hybrid Search

vibrantvarun:
### Description
Optimize the max score tracking in the QueryPhase of Hybrid Search. 
**Latency reduced**
```
(Number of Shards) x (Number of topDocs found on each shard)
```

### Issues Resolved
[764](https://github.com/opensearch-project/neural-search/issues/764)

### Check List
    - [ ] All tests pass


src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
```diff
@@ -115,6 +117,9 @@ public void collect(int doc) throws IOException {
                     collectedHitsPerSubQuery[i]++;
                     PriorityQueue<ScoreDoc> pq = compoundScores[i];
                     ScoreDoc currentDoc = new ScoreDoc(doc + docBase, score);
+                    if (score > maxScore) {
```
shatejas, line 120: nit: ` maxScore = Math.max(score, maxScore)` for readability

src/test/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollectorTests.java
```diff
@@ -564,4 +564,64 @@ public void testTotalHitsCountValidation_whenTotalHitsCollectedAtTopLevelInColle
         reader.close();
         directory.close();
     }
+
+    @SneakyThrows
+    public void testMaxScoreValidation_whenMaxScoreCollectedAtTopLevelInCollector_thenSuccessful() {
+        final Directory directory = newDirectory();
+        final IndexWriter w = new IndexWriter(directory, newIndexWriterConfig(new MockAnalyzer(random())));
+        FieldType ft = new FieldType(TextField.TYPE_NOT_STORED);
+        ft.setIndexOptions(random().nextBoolean() ? IndexOptions.DOCS : IndexOptions.DOCS_AND_FREQS);
```
shatejas, line 573: nit: randomness can introduce flakiness in the test. avoid if possible, seems simple enough to add two separate cases. Try and model your tests to be parametrized so you can easily add cases w/o copy pasting

src/test/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollectorTests.java
```diff
@@ -564,4 +564,64 @@ public void testTotalHitsCountValidation_whenTotalHitsCollectedAtTopLevelInColle
         reader.close();
         directory.close();
     }
+
+    @SneakyThrows
+    public void testMaxScoreValidation_whenMaxScoreCollectedAtTopLevelInCollector_thenSuccessful() {
```
shatejas, line 569: any possibility we can introduce multiple leaves in the test? The nature of the collector is to collect top results across leaves. And the tests should ideally reflect that behavior for it to be tight. See if you can have a case where you can have at least one more leaf

vibrantvarun: The [collect](https://github.com/opensearch-project/neural-search/pull/765/files#diff-efec3f3ef0d15cf4737a062d0f2b229eb984463e066013395a9a5b7e1867a40eL94) method runs for every doc Id found as a result of the query on a particular shard. What you are expecting is a step which is way before this method to execute. 

Let me explain how this thing works

1) Initialize leaf collectors from all the collectors at a shard level. A leaf collector is unique entity.

For example
Let say there are 2 collectors TopDocCollector and HybridTopScoreDocCollector. These 2 collectors will have a unique leaf collectors of there own. Those leaf collectors will run once per shard. 

2) Then execute those leaf collectors (In this case https://github.com/opensearch-project/neural-search/pull/765/files#diff-efec3f3ef0d15cf4737a062d0f2b229eb984463e066013395a9a5b7e1867a40eR51)
This will like
leafCollector.collect (docId1) ---> this call will be made from OpenSearch
leafCollector.collect (docId2)
leafCollector.collect (docId3)

docId1, docId2, docId3 are the results of HybridQuery
3) return the result to coordinator node.

src/test/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollectorTests.java
vibrantvarun: I am following the pattern above which we used earlier to add tests in this Class. Btw, it does not matter if we use random in this case or not. Because the core logic we are testing here is the max score calculated by previous logic and the max score calculated by new logic. If at any point the two scores doesn't match then that will be considered as failure not flakiness.

vibrantvarun: If we have to change it to modularized or maybe optimize the tests then that has to be done in the entire project, Probably we cater it in one complete issue. The score of this PR is to just validate the max score logic.

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
vibrantvarun: Sure I will add the change.

VijayanB:
@vibrantvarun Do you have comparison in latency before and after this change?

vibrantvarun:
> @vibrantvarun Do you have comparison in latency before and after this change?

Sure, I will add the latency graph. I am having some issues in the setup. Probably, once I resolve the issues I can add that flame graph in the issue of this PR. But, to unblock me on to raise my Sorting PR this PR needs to be merged.

src/test/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollectorTests.java
shatejas: >  am following the pattern above 
I would highly encourage to break anti-patterns when you see one.

> Btw, it does not matter if we use random in this case or not.
Lets remove it then? pick a value

> If we have to change the tests to be parameterized then that has to be done in the entire project
You don't have to revamp all tests. its hard to justify the return on it. You can always start with new tests though.

src/main/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollector.java
```diff
@@ -115,6 +117,7 @@ public void collect(int doc) throws IOException {
                     collectedHitsPerSubQuery[i]++;
                     PriorityQueue<ScoreDoc> pq = compoundScores[i];
                     ScoreDoc currentDoc = new ScoreDoc(doc + docBase, score);
+                    maxScore = Math.max(score, maxScore);
```
VijayanB, line 120: nit: May be move this close to score variable

vibrantvarun: The motive here is to compare the score which is added in the scoredoc with the maxScore. That is the  reason I have added it after the currentDoc. I can change this condition to currentDoc.score()

src/test/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollectorTests.java
vibrantvarun: The thing is the place where random() is used it is lucene method. And lucene encourages to use randomized tests https://lucene.apache.org/core/7_3_1/test-framework/org/apache/lucene/util/LuceneTestCase.html . The reason is that it is not possible to cover all the test cases by using parameterized approach. Therefore, lucene recommends Randomized tests to increase more coverage.

vibrantvarun: At any other place which is not a lucene method we have used the paramertized approach for example DOCID1, DOCID2 etc.

vibrantvarun: https://github.com/randomizedtesting/randomizedtesting/wiki/Core-Concepts

vibrantvarun: Therefore, using random() is not a anti pattern.

src/test/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollectorTests.java
shatejas: While there is a new leafCollector, [you still are updating the priority queue and max score](https://github.com/opensearch-project/neural-search/pull/765/files#diff-efec3f3ef0d15cf4737a062d0f2b229eb984463e066013395a9a5b7e1867a40eL41) across collectors because its a member variable. And thats where the nature of this collector comes that is to collect the top docs across leaves and its important to test it out.

src/test/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollectorTests.java
```diff
@@ -564,4 +564,64 @@ public void testTotalHitsCountValidation_whenTotalHitsCollectedAtTopLevelInColle
...
+            nextDoc = iterator.nextDoc();
+        }
+
+        List<TopDocs> topDocs = hybridTopScoreDocCollector.topDocs();
+        float maxScore = topDocs.stream()
+            .map(docs -> docs.scoreDocs.length == 0 ? new ScoreDoc(-1, 0.0f) : docs.scoreDocs[0])
+            .map(scoreDoc -> scoreDoc.score)
+            .max(Float::compare)
+            .get();
+        assertEquals(maxScore, hybridTopScoreDocCollector.getMaxScore(), 0.0f);
```
VijayanB, line 622: seems you are depending on objects that are not part of testing. If those dependencies introduce any bug, this test will fail not because of what you want to test. Hence, either mock those external dependencies or convert to integration test if you feel you need to cover the complete flow.

vibrantvarun: I am removing this unit test and will add a integ test instead. It will also address @shatejas and @VijayanB concerns.

src/test/java/org/opensearch/neuralsearch/search/HybridTopScoreDocCollectorTests.java
shatejas: Thanks for the article! 

Randomized testing seems to be having a use case when the complexity of the code is high and its not possible to cover cases everytime. Justifying a random input when it doesn't matter doesn't seems to be the right approach. I will leave it up to you how you want to keep this as it wasn't a blocker in the first place

VijayanB (approved):

junqiu-lei (approved):

jmazanec15 (commented):
> (Number of Shards) x (Number of topDocs found on each shard)

Id be interested to see end user impact on this. topDocs will be a max of 10k right?

martin-gaievski:
Final max score is calculated at the coordinator level as part of [normalization processor workflow](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java#L113-L118). I think we can skip it calculation at shard level, maybe put 0.0. 
@vibrantvarun can you check this theory, my concern is that 0.0 score doesn't trigger any unwanted code path in core logic, plus check edge cases?

src/test/java/org/opensearch/neuralsearch/query/HybridQueryIT.java
```diff
@@ -206,6 +206,36 @@ public void testTotalHits_whenResultSizeIsLessThenDefaultSize_thenSuccessful() {
         assertEquals(RELATION_EQUAL_TO, total.get("relation"));
     }
 
+    @SneakyThrows
+    public void testMaxScoreCalculation_whenMaxScoreIsTrackedAtCollectorLevel_thenSuccessful() {
+        initializeIndexIfNotExist(TEST_BASIC_VECTOR_DOC_FIELD_INDEX_NAME);
```
shatejas, line 211: The primary concern in the unit test was the behavior of collector during multi segment scenario. Does this initialization make sure of 2 segments at least

vibrantvarun: We mostly write multiple segments scenarios when a results has been impacted by concurrent search is enabled and disabled. In this case the results does not get impacted by that.

Issue #507, opened 2023-11-03
[FEATURE] Sorting support for Hybrid Search 


ankitas3:
**Is your feature request related to a problem?**
Hybrid Search always returns results in the descending order of score. Applying sorting as below has no affect on the results
```
{
  "query": {
    "hybrid":{
      "queries": [
        <text query>,
        <kNN query>
      ]
    }
  }
  "sort": [
    {
      "title.keyword": {
        "order": "desc"
      }
    }
  ]
}
```

**What solution would you like?**
Sorting should work as it does with other boolean queries. Results matching the Hybrid query should later be sorted on the basis of sorting applied.

vibrantvarun:
Hey @ankitas3 Can you share the use case of this problem? We are doing some brainstorming and we need to understand how sorting fits in your usecase.

tsilvafsl:
I am experiencing a problem that might be related.

When performing a hybrid search query with a `search_pipeline`, the `sort` tag does not work as expected.

In other words, the `sort` is being bypassed by `search_pipeline` and if you remove the `search pipeline`, the `sort` works fine in a hybrid query.

In this issue I describe the problem and a reproducible example.
https://forum.opensearch.org/t/unable-to-sort-query-results-with-search-pipeline-hybrid-query/19726

Is it possible to make it work with current opensearch?

navneet1v:
@vibrantvarun can we close this issue as the feature is merged?

PR #768, opened 2024-05-29
[Feature] Enable Sorting in Hybrid Search

vibrantvarun:
### Description
This PR is to enable sorting in Hybrid Search.
The following key areas which are covered by this PR
1. Enable sorting on single field and multi-field.
2. Enable `search_after` pagination support with sorting. The reason why `search_after` is included in this PR because without this the whole sorting feature is half baked. This feature exclusively works with sorting.
    -    The same methods which we have created for sorting are used by search_after. Therefore, the code can be more optimized when this feature is added with this PR.
3. Block `track_scores` when sorting is applied
4. Block the scenario when user tries to sort by `field` and `_score` both.

Order to review the PR
1. `HybridCollectorManager.createHybridCollectorManager` method
2. `HybridCollectorManager.newCollector` method
3. `HybridTopFieldDocSortCollector.java`
4. `MultiLeafFieldComparator.java` which is directly copy of from Lucene. The reason why we copied this class in neural search because this class is final in Lucene.
5. `HybridCollectorManager.reduce` method
6. `NormarlizationProcessorWorkFlow`
7. `CompoundTopDocs`
8. `ScoreCombiner`
9. `HybridQueryResultUtil`
10. Test Classes

### Issues Resolved
[507](https://github.com/opensearch-project/neural-search/issues/507)

### Check List
    - [ ] All tests pass


src/main/java/org/opensearch/neuralsearch/search/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,472 @@
...
+                    }
+                }
+                return true;
+            }
+            return false;
+        }
+
+    }
+
+    public static class SimpleFieldCollector extends HybridTopFieldDocSortCollector {
```
martin-gaievski, line 292: class level comment is missing

src/main/java/org/opensearch/neuralsearch/search/MultiLeafFieldComparator.java
```diff
@@ -0,0 +1,95 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.search;
+
+import java.io.IOException;
+import org.apache.lucene.search.DocIdSetIterator;
+import org.apache.lucene.search.LeafFieldComparator;
+import org.apache.lucene.search.Scorable;
+
+final class MultiLeafFieldComparator implements LeafFieldComparator {
```
martin-gaievski, line 12: class level comment is missing

src/main/java/org/opensearch/neuralsearch/search/MultiLeafFieldComparator.java
```diff
@@ -0,0 +1,95 @@
...
+    private final int[] reverseMul;
+    // we extract the first comparator to avoid array access in the common case
+    // that the first comparator compares worse than the bottom entry in the queue
+    private final LeafFieldComparator firstComparator;
+    private final int firstReverseMul;
+
+    MultiLeafFieldComparator(LeafFieldComparator[] comparators, int[] reverseMul) {
+        if (comparators.length != reverseMul.length) {
+            throw new IllegalArgumentException(
+                "Must have the same number of comparators and reverseMul, got " + comparators.length + " and " + reverseMul.length
```
martin-gaievski, line 24: please use proper formatting for complex error message - String.format + placeholders for parameters

src/main/java/org/opensearch/neuralsearch/search/MultiLeafFieldComparator.java
```diff
@@ -0,0 +1,95 @@
...
+    @Override
+    public void setBottom(int slot) throws IOException {
+        for (LeafFieldComparator comparator : comparators) {
+            comparator.setBottom(slot);
+        }
+    }
+
+    @Override
+    public int compareBottom(int doc) throws IOException {
+        int cmp = firstReverseMul * firstComparator.compareBottom(doc);
```
martin-gaievski, line 42: what is s semantic of this variable? maybe coma up with better name

src/main/java/org/opensearch/neuralsearch/search/MultiLeafFieldComparator.java
```diff
@@ -0,0 +1,95 @@
...
+            if (cmp != 0) {
+                return cmp;
+            }
+        }
+        return 0;
+    }
+
+    @Override
+    public int compareTop(int doc) throws IOException {
+        int cmp = firstReverseMul * firstComparator.compareTop(doc);
```
martin-gaievski, line 57: same question for variable name

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -86,21 +98,51 @@ public static CollectorManager createHybridCollectorManager(final SearchContext
...
+                    throw new IllegalArgumentException("after.fields wasn't set; you must pass fillFields=true for the previous search");
+                }
+
+                if (after.fields.length != sortAndFormats.sort.getSort().length) {
+                    throw new IllegalArgumentException(
+                        "after.fields has " + after.fields.length + " values but sort has " + sortAndFormats.sort.getSort().length
+                    );
+                }
+
+                hybridcollector = new HybridTopFieldDocSortCollector.PagingFieldCollector(
```
martin-gaievski, line 136: move validation logic to it's own method, this will make code easier to read

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -149,9 +210,71 @@ public ReduceableSearchResult reduce(Collection<Collector> collectors) {
...
+            HybridTopFieldDocSortCollector hybridSortedTopScoreDocCollector = hybridSortedTopDocCollectors.stream()
+                .findFirst()
+                .orElseThrow(() -> new IllegalStateException("cannot collect results of hybrid search query"));
+
+            HybridTopFieldDocSortCollector.SimpleFieldCollector simpleFieldCollector;
+            HybridTopFieldDocSortCollector.PagingFieldCollector pagingFieldCollector;
+            List<TopFieldDocs> topFieldDocs;
+            long maxTotalHits;
+            float maxScore;
+            if (hybridSortedTopScoreDocCollector instanceof HybridTopFieldDocSortCollector.SimpleFieldCollector) {
```
martin-gaievski, line 224: I'm not sure you need whole if/else construct, seems that all three fields you're trying to get are defined at the abstract class level, so you don'r need to cast to the subclass

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -127,14 +170,32 @@ public ReduceableSearchResult reduce(Collection<Collector> collectors) {
...
+                    && ((FilteredCollector) collector).getCollector() instanceof HybridTopFieldDocSortCollector.SimpleFieldCollector) {
+                        hybridSortedTopDocCollectors.add(
+                            (HybridTopFieldDocSortCollector.SimpleFieldCollector) ((FilteredCollector) collector).getCollector()
+                        );
+                    } else if (collector instanceof FilteredCollector
+                        && ((FilteredCollector) collector).getCollector() instanceof HybridTopFieldDocSortCollector.PagingFieldCollector) {
+                            hybridSortedTopDocCollectors.add(
+                                (HybridTopFieldDocSortCollector.PagingFieldCollector) ((FilteredCollector) collector).getCollector()
+                            );
+                        }
```
martin-gaievski, line 198: should we validate that both regular score and sorted score collectors are not empty?

martin-gaievski, line 198: Another idea -  do we need list of collectors? We are using only first meaning any suitable one that we have found. It should be enough just to maintain one variable for each collector type, you can override previous value if it has been set.

src/main/java/org/opensearch/neuralsearch/search/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,472 @@
...
+                    initializePriorityQueuesWithComparators(context, subScoresByQuery.length);
+                    incrementTotalHitCount();
+                    for (int i = 0; i < subScoresByQuery.length; i++) {
+                        float score = subScoresByQuery[i];
+                        // if score is 0.0 there is no hits for that sub-query
+                        if (score == 0) {
+                            continue;
+                        }
+                        collectedHits[i]++;
+                        maxScore = Math.max(score, maxScore);
```
martin-gaievski, line 322: this is not needed, we're calculating max score at the processor level

yuye-aws:
Why are we blocking `track_scores` when sorting is applied. According to [track_scores](https://opensearch.org/docs/latest/search-plugins/searching-data/sort/#tracking-scores), user can still set this to be true when sorting is applied.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -149,9 +210,71 @@ public ReduceableSearchResult reduce(Collection<Collector> collectors) {
...
+        boolean isSortByScore = false;
+        for (SortField sortField : sortFields) {
+            SortField.Type type = sortField.getType();
+            if (type.equals(SortField.Type.SCORE)) {
+                isSortByScore = true;
+            } else {
+                isSortByField = true;
+            }
+            if (isSortByScore && isSortByField) {
+                break;
```
yuye-aws, line 260: Why not throwing `IllegalArgumentException` immediately here?

src/main/java/org/opensearch/neuralsearch/search/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,472 @@
...
+    private TotalHits.Relation totalHitsRelation = TotalHits.Relation.EQUAL_TO;
+    private Boolean searchSortPartOfIndexSort = null;
+
+    private static final TopFieldDocs EMPTY_TOPDOCS = new TopFieldDocs(
+        new TotalHits(0, TotalHits.Relation.EQUAL_TO),
+        new ScoreDoc[0],
+        new SortField[0]
+    );
+
+    // Declaring the constructor private prevents extending this class by anyone
```
yuye-aws, line 68: From my understanding, private constructor does not prevent extending the class. Are we enforcing this class to be a singleton

PR #787, opened 2024-06-11
Adds method_parameters in neural search query to support ef_search

shatejas:
### Description
Adds method_parameters in neural search

### Check List
    - [x] All tests pass


src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
```diff
@@ -0,0 +1,47 @@
...
+import org.opensearch.knn.index.IndexUtil;
+import org.opensearch.neuralsearch.util.NeuralSearchClusterUtil;
+
+import java.util.Map;
+
+import static org.opensearch.knn.index.query.KNNQueryBuilder.MAX_DISTANCE_FIELD;
+import static org.opensearch.knn.index.query.KNNQueryBuilder.MIN_SCORE_FIELD;
+import static org.opensearch.neuralsearch.query.NeuralQueryBuilder.MODEL_ID_FIELD;
+
+public class MinClusterVersionUtil {
```
martin-gaievski, line 18: please add class level comments

src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private static final Version MINIMAL_SUPPORTED_VERSION_DEFAULT_MODEL_ID = Version.V_2_11_0;
+    private static final Version MINIMAL_SUPPORTED_VERSION_RADIAL_SEARCH = Version.V_2_14_0;
+
+    // Note this minimal version will act as a override
+    private static final Map<String, Version> MINIMAL_VERSION_NEURAL = ImmutableMap.<String, Version>builder()
```
martin-gaievski, line 24: nit: I think "NEURAL" prefix is redundant, in plugin repo everything is related to this plugin

src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
```diff
@@ -0,0 +1,47 @@
...
+    public static boolean isClusterOnOrAfterMinReqVersionForRadialSearch() {
+        return NeuralSearchClusterUtil.instance().getClusterMinVersion().onOrAfter(MINIMAL_SUPPORTED_VERSION_RADIAL_SEARCH);
+    }
+
+    public static boolean isClusterOnOrAfterMinReqVersion(String key) {
+        Version version;
+        if (MINIMAL_VERSION_NEURAL.containsKey(key)) {
+            version = MINIMAL_VERSION_NEURAL.get(key);
+        } else {
+            version = IndexUtil.minimalRequiredVersionMap.get(key);
```
martin-gaievski, line 43: IndexUtil is from knn repo, we shouldn't add/keep to minimal such dependencies on other plugins.

Also for the logic itself - why we're falling back to knn implementation in case we cannot find settings specific to neural search? I think we either add settings to neural or return 'false'

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -292,15 +301,14 @@ protected QueryBuilder doRewrite(QueryRewriteContext queryRewriteContext) {
...
-            KNNQueryBuilder knnQueryBuilder = new KNNQueryBuilder(fieldName(), vectorSupplier.get()).filter(filter());
-            if (maxDistance != null) {
-                knnQueryBuilder.maxDistance(maxDistance);
-            } else if (minScore != null) {
-                knnQueryBuilder.minScore(minScore);
-            } else {
-                knnQueryBuilder.k(k);
-            }
-            return knnQueryBuilder;
+            return KNNQueryBuilder.builder()
```
martin-gaievski, line 304: please add comment on why we can set all maxDistance, minScore and k even if some of them can be null, if that was enabled in knn recently please add PR ref

martin-gaievski (commented):
changelog entry is missing

shatejas:
> changelog entry is missing

Its being merged in feature branch. Was planning to add it when merging in main. does that work?

src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
shatejas: > IndexUtil is from knn repo, we shouldn't add/keep to minimal such dependencies on other plugins.

Isn't this a wrapper on k-NN and takes a dependency on it? Why can't there be class dependencies for reuse when its a direct dependency?

> Also for the logic itself - why we're falling back to knn implementation in case we cannot find settings specific to neural search? I think we either add settings to neural or return 'false'

To not have a mismatch for Knn related parameters. For instance, efsearch has a version associated with it. Inconsistencies might cause issues there. If we want it to be different we should add it in `MINIMAL_VERSION_NEURAL`.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
shatejas: Yes the builder will throw an exception if any if the parameters or their combinations are incorrect. This querybuilder shouldn't be responsible for k-NN validations. Here is the PR https://github.com/opensearch-project/k-NN/pull/1707

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -174,6 +178,9 @@ protected void doXContent(XContentBuilder xContentBuilder, Params params) throws
         if (Objects.nonNull(minScore)) {
             xContentBuilder.field(MIN_SCORE_FIELD.getPreferredName(), minScore);
         }
+        if (Objects.nonNull(methodParameters)) {
+            MethodParametersParser.doXContent(xContentBuilder, methodParameters);
```
navneet1v, line 182: is this class MethodParametersParser coming from k-NN?

shatejas: Yes

navneet1v: This is good. Lets ensure that we add some more docs on the MethodParametersParser function that it is now going beyond the boundaries of k-NN plugin. I would hate to see if we change signature of these functions and then neural plugin starts to break.

martin-gaievski (approved):
We need to think more on how to commonize logic that's shared between neural search and knn plugins. May be a too heavy lift for this change. At the moment this PR is good, and I see we do have an issue to follow up on later: https://github.com/opensearch-project/k-NN/issues/1756

martin-gaievski:
Please check why CI is failing, probably need to rebase on latest main

shatejas:
> Please check why CI is failing, probably need to rebase on latest main

The build is failing because the knn changes aren't in main yet. It is expected to fail as it cannot identify `MethodParameterParser`. Locally I published to local maven repo for neural to build

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
shatejas: https://github.com/opensearch-project/k-NN/pull/1758/files

navneet1v:
@vibrantvarun can you check

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/HybridSearchIT.java
```diff
@@ -69,6 +70,7 @@ private void validateNormalizationProcessor(final String fileName, final String
                 loadModel(modelId);
                 addDocuments(getIndexNameForTest(), false);
                 validateTestIndex(modelId, getIndexNameForTest(), searchPipelineName);
+                validateTestIndex(modelId, getIndexNameForTest(), searchPipelineName, Map.of("ef_search", 100));
```
vibrantvarun, line 73: Instead of adding this validation in next line, we should pass this Map in line 72. Otherwise, it will repeat the same validation again. Instead of creating a copy method validateTestIndex ,Can we add extra parameter in the same method definition?

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/HybridSearchIT.java
```diff
@@ -73,6 +73,7 @@ public void testNormalizationProcessor_whenIndexWithMultipleShards_E2EFlow() thr
                     loadModel(modelId);
                     addDocument(getIndexNameForTest(), "2", TEST_FIELD, TEXT_UPGRADED, null, null);
                     validateTestIndexOnUpgrade(totalDocsCountUpgraded, modelId);
+                    validateTestIndexOnUpgrade(totalDocsCountUpgraded, modelId, Map.of("ef_search", 100));
```
vibrantvarun, line 76: Same comment as above.

qa/rolling-upgrade/src/test/resources/processor/IndexMappings.json
```diff
@@ -16,7 +16,7 @@
         "method": {
           "name": "hnsw",
           "space_type": "l2",
-          "engine": "lucene",
+          "engine": "faiss",
```
vibrantvarun, line 19: Last when I wrote BWC tests with "faiss" engine tests were not passing. Because test framework works well only with "lucene" engine. Can you run the test standalone and verify it once?

vibrantvarun:
There are lot of compilation errors which are not related to KNN. @shatejas  Can you check why it is happening?

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/HybridSearchIT.java
shatejas: > Instead of adding this validation in next line, we should pass this Map in line 72. Otherwise, it will repeat the same validation again

The intent is to do the validation again, with and without method parameters. so its called twice, The results should be the same

> Instead of creating a copy method validateTestIndex ,Can we add extra parameter in the same method definition?

Its not a copy method, its an overloaded method. The logic is being reused see lines 101 and 105

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/HybridSearchIT.java
shatejas: Same reply as above

qa/rolling-upgrade/src/test/resources/processor/IndexMappings.json
shatejas: This was changed before we had support for method parameters for lucene (it would have thrown errors with lucene). can roll it back now that we have support for lucene

shatejas:
> There are lot of compilation errors which are not related to KNN. @shatejas Can you check why it is happening?

See [this comment](https://github.com/opensearch-project/neural-search/pull/787#issuecomment-2168453319)

vibrantvarun (approved):

PR #797, opened 2024-06-17
[Part 1] Collector for Sorting Results

vibrantvarun:
### Description
This PR is the first PR of Sorting feature. It covers the collector which will be responsible for getting the sorted results on the basis of the sort criteria.

1) The HybridTopFieldDocSortingCollector.SimpleFieldCollector is responsible get the individual query results in the sorted form. 
2) The HybridTopFieldDocSortingCollector.PagingFieldCollector is responsible to applying `search_after` filter pagination on the sorted results of individual query.

The inspiration of the collector is from https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java

The MultileaffieldComparator is from lucene
https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/MultiLeafFieldComparator.java

### Issues Resolved
[https://github.com/opensearch-project/neural-search/issues/507](https://github.com/opensearch-project/neural-search/issues/507)

### Check List
    - [ ] All tests pass


src/main/java/org/opensearch/neuralsearch/search/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,485 @@
+/*
```
Vikasht34, line 1: At high level , this class says Abstract but it contains 500 lines of code , it's very hard to maintain and read from one file, Can u break this class in logical part.

src/main/java/org/opensearch/neuralsearch/search/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,485 @@
...
+                    }
+                }
+                return true;
+            }
+            return false;
+        }
+
+    }
+
+    public static class SimpleFieldCollector extends HybridTopFieldDocSortCollector {
```
martin-gaievski, line 305: can we pull this class to its own file? Same for PagingFieldCollector class

src/main/java/org/opensearch/neuralsearch/search/HybridTopFieldDocSortCollector.java
shatejas: +1. Seems like this PR is favoring inheritance for code reuse. Can we keep inheritance/polymorphism for enforcing behavior and composition for code reuse? This has added a advantages and you will find quite a few articles around it. Here is a pointer to start with https://en.wikipedia.org/wiki/Composition_over_inheritance 

As a guideline don't use inheritance for code reuse until absolutely required

vibrantvarun:
Integ tests in secure environment cluster are failing due to [https://github.com/opensearch-project/security-dashboards-plugin/issues/2009](https://github.com/opensearch-project/security-dashboards-plugin/issues/2009)

src/main/java/org/opensearch/neuralsearch/search/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,368 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.search;
```
navneet1v, line 5: can we move all collector related classes to package org.opensearch.neuralsearch.search.collector

Vikasht34 (commented):
At 1000ft overview , here are the high level comments that needs to be addressed !! 

The whole idea was move away from inheritance and use composition. Right now we have two classes one Abstract Class which inherits from collector and other two class which inhertis from this .
Instead of this , Write all the common logic in sperate component classes like 

1.CollectorState: Manage state variables like totalHits, maxScore, etc.
2.PriorityQueueManager: Handle priority queue operations.
3. ScorerManager: Handle scorer operations and initialization.


these are indiviadual components that needs to be HAS-A relation with PagingFieldCollector and the other one, like below. 

public class PagingFieldCollector implements Collector {
    private final CollectorState state;
    private final PriorityQueueManager queueManager;
    private final ScorerManager scorerManager;
}

if I look at current implementation also , Abstarct class is doing lots of stuff , If we can extract most of the functionality from this class and put in helper and then have "HAS-A" relation with this abstract class would also help. 

I have not looked into other code part, let's fix on structure of classes and then we can have low -level review .

vibrantvarun:
> At 1000ft overview , here are the high level comments I have
> 
> The whole idea was move away from inheritance and use composition. Right now we have two classes one Abstract Class which inherits from collector and other two class which inhertis from this . Instead of this , Write all the common logic in sperate component classes like
> 
> 1.CollectorState: Manage state variables like totalHits, maxScore, etc. 2.PriorityQueueManager: Handle priority queue operations. 3. ScorerManager: Handle scorer operations and initialization.
> 
> these are indiviadual components that needs to be HAS-A relation with PagingFieldCollector and the other one, like below.
> 
> public class PagingFieldCollector implements Collector { private final CollectorState state; private final PriorityQueueManager queueManager; private final ScorerManager scorerManager; }
> 
> if I look at current implementation also , Abstarct class is doing lots of stuff , If we can extract most of the functionality from this class and put in helper and then have "HAS-A" relation with this abstract class would also help.
> 
> I have not looked into other code part, let's fix on structure of classes and then we can have low -level review .

@Vikasht34 The idea is here is to replicate what is there in lucene and not create a whole new level of sort collector. They have given a significant effort in maintaining this class [https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java). I would not like to move away from this pattern as it may break many functionlities.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,411 @@
...
+                    compoundScores[i],
+                    collectedHits[i],
+                    sort.getSort()
+                )
+            )
+            .collect(Collectors.toList());
+        return topFieldDocs;
+    }
+
+    public abstract class HybridTopDocSortLeafCollector implements LeafCollector {
```
navneet1v, line 163: This class being a public class can be moved out of this file to a separate file with correct access modifiers

src/main/java/org/opensearch/neuralsearch/search/MultiLeafFieldComparator.java
```diff
@@ -0,0 +1,100 @@
...
+
+import java.io.IOException;
+import org.apache.lucene.search.DocIdSetIterator;
+import org.apache.lucene.search.LeafFieldComparator;
+import org.apache.lucene.search.Scorable;
+
+/*
+MultiLeafFieldComparator holds information when sort criteria is applied on more that one field.
+This class is same as of lucene. Because lucene implementation does not have public access we have to add it in neural search plugin.
+https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/MultiLeafFieldComparator.java
```
navneet1v, line 15: this is exact copy of Lucene file? if yes then we might need to reconsider what we are trying to do here.

@vibrantvarun if there are classes which are exact copy from Lucene with access modifiers changed then I would probably put them in package with namespace of Lucene in it. This way atleast we know what are the classes which are from Lucene

src/main/java/org/opensearch/neuralsearch/search/collector/PagingFieldCollector.java
```diff
@@ -0,0 +1,93 @@
...
+                    if (resultsFoundOnPreviousPage) {
+                        return;
+                    }
+                    maxScore = Math.max(score, maxScore);
+                    if (queueFull[i]) {
+                        collectCompetitiveHit(doc, i);
+                    } else {
+                        collectedHits[i]++;
+                        collectHit(doc, collectedHits[i], i, score);
+                    }
```
navneet1v, line 73: need more details what we are trying to do here.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,411 @@
...
+        new ScoreDoc[0],
+        new SortField[0]
+    );
+
+    // Declaring the constructor private prevents extending this class by anyone
+    // else. Note that the class cannot be final since it's extended by the
+    // internal versions. If someone will define a constructor with any other
+    // visibility, then anyone will be able to extend the class, which is not what
+    // we want.
+    public HybridTopFieldDocSortCollector(final int numHits, final HitsThresholdChecker hitsThresholdChecker) {
```
navneet1v, line 100: since this is an abstract class lets make the constructor package private so that classes in this package only can use this constructor.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,411 @@
...
+            if (fields1.length > fields2.length) {
+                return false;
+            }
+            return Arrays.asList(fields1).equals(Arrays.asList(fields2).subList(0, fields1.length));
+        } else {
+            return false;
+        }
+    }
+
+    public List<TopFieldDocs> topDocs(final FieldValueHitQueue<FieldValueHitQueue.Entry>[] compoundScores, final Sort sort) {
```
navneet1v, line 145: add documentation who is going to call this function? and when

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,411 @@
...
+        if (compoundScores == null) {
+            return new ArrayList<>();
+        }
+        final List<TopFieldDocs> topFieldDocs = IntStream.range(0, compoundScores.length)
```
navneet1v, line 149: given the impact we have seen of using streams on performance I would make this as a simple for loop. To ensure that we are not doing another round of optimizations on these streams.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,411 @@
...
+
+    @Override
+    public ScoreMode scoreMode() {
+        return hitsThresholdChecker.scoreMode();
+    }
+
+    /*
+      TopFieldDocs per subquery
+     */
+    protected TopFieldDocs topDocsPerQuery(
```
navneet1v, line 355: any reason for making this protected? I see that it not getting used in base classes. plus do we recommend overriding this function in future by base classes if no, then we should make this as private.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,411 @@
...
+            return EMPTY_TOPDOCS;
+        }
+
+        int size = howMany - start;
+        ScoreDoc[] results = new ScoreDoc[size];
+        // pq's pop() returns the 'least' element in the queue, therefore need
+        // to discard the first ones, until we reach the requested range.
+        for (int i = pq.size() - start - size; i > 0; i--) {
+            pq.pop();
+        }
```
navneet1v, line 384: help me understand what you are trying to do here:
You are trying to reduce the size of the PQ by discarding the elements but one thing I am not getting is why we are doing that?

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: Actually no, this class is operates on the state of the collector by using instance variable. If we think of moving this class out then the constructor of the class will become super heavy as more than 10-12 variables. Also functionality might break. What I can do is I can make this class protected.

src/main/java/org/opensearch/neuralsearch/search/collector/SimpleFieldCollector.java
```diff
@@ -0,0 +1,68 @@
...
+                    } else {
+                        collectHit(doc, collectedHits[i], i, score);
+                    }
+
+                }
+            }
+        };
+    }
+
+    public List<TopFieldDocs> topDocs() {
```
navneet1v, line 65: We can avoid this function by having the base class always calling right topDocs function.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -86,21 +100,33 @@ public static CollectorManager createHybridCollectorManager(final SearchContext
...
     public Collector newCollector() {
-        Collector hybridcollector = new HybridTopScoreDocCollector(numHits, hitsThresholdChecker);
+        Collector hybridcollector;
+        if (sortAndFormats != null) {
+            if (after == null) {
+                hybridcollector = new SimpleFieldCollector(numHits, hitsThresholdChecker, sortAndFormats.sort);
+            } else {
+                validateSearchAfterFieldAndSortFormats();
+                hybridcollector = new PagingFieldCollector(numHits, hitsThresholdChecker, sortAndFormats.sort, after);
+            }
```
navneet1v, line 126: do we support pagination currently with Hybrid Query clause? My understanding was no.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -127,14 +154,28 @@ public ReduceableSearchResult reduce(Collection<Collector> collectors) {
...
                 && ((FilteredCollector) collector).getCollector() instanceof HybridTopScoreDocCollector) {
                     hybridTopScoreDocCollectors.add((HybridTopScoreDocCollector) ((FilteredCollector) collector).getCollector());
-                }
+                } else if (collector instanceof FilteredCollector
+                    && ((FilteredCollector) collector).getCollector() instanceof SimpleFieldCollector) {
+                        hybridSortedTopDocCollectors.add((SimpleFieldCollector) ((FilteredCollector) collector).getCollector());
+                    } else if (collector instanceof FilteredCollector
+                        && ((FilteredCollector) collector).getCollector() instanceof PagingFieldCollector) {
+                            hybridSortedTopDocCollectors.add((PagingFieldCollector) ((FilteredCollector) collector).getCollector());
+                        }
```
navneet1v, line 178: fix formatting here.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -149,9 +190,83 @@ public ReduceableSearchResult reduce(Collection<Collector> collectors) {
...
+                maxScore = pagingFieldCollector.getMaxScore();
+            }
+
+            TopDocs newTopDocs = getNewTopFieldDocs(
+                getTotalHits(this.trackTotalHitsUpTo, topFieldDocs, isSingleShard, maxTotalHits),
+                topFieldDocs,
+                sortAndFormats.sort.getSort()
+            );
+            TopDocsAndMaxScore topDocsAndMaxScore = new TopDocsAndMaxScore(newTopDocs, maxScore);
+            return (QuerySearchResult result) -> { result.topDocs(topDocsAndMaxScore, getSortValueFormats(sortAndFormats)); };
```
navneet1v, line 222: lets move this to a private function in this class. We can move the above if condition but that you can take iteratively.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
vibrantvarun: No we do not. This `search_after` works exclusively with sorting. I have added this functionality with this feature so that in future I do not have to touch this class again for adding any new functionality. Next thing in my roadmap is enable pagination in hybrid search. https://opensearch.org/docs/latest/search-plugins/searching-data/paginate/#the-search_after-parameter

src/main/java/org/opensearch/neuralsearch/search/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,368 @@
+/*
+ * Copyright OpenSearch Contributors
```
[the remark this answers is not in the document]
vibrantvarun: Ack

src/main/java/org/opensearch/neuralsearch/search/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,368 @@
...
+
+    private boolean canEarlyTerminateOnPrefix(Sort searchSort, Sort indexSort) {
+        if (indexSort != null) {
+            final SortField[] fields1 = searchSort.getSort();
+            final SortField[] fields2 = indexSort.getSort();
+            // early termination is possible if fields1 is a prefix of fields2
+            if (fields1.length > fields2.length) {
+                return false;
+            }
+            return Arrays.asList(fields1).equals(Arrays.asList(fields2).subList(0, fields1.length));
```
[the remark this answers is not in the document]
vibrantvarun: ack

src/main/java/org/opensearch/neuralsearch/search/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,368 @@
...
+
+/*
+ Collects the TopFieldDocs after executing hybrid query. Uses HybridQueryTopDocs as DTO to handle each sub query results.
+ The individual query results are sorted as per the sort criteria sent in the search request.
+ */
+@Log4j2
+public abstract class HybridTopFieldDocSortCollector implements Collector {
+    final int numHits;
+    final HitsThresholdChecker hitsThresholdChecker;
+    int docBase;
```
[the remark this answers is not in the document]
vibrantvarun: Working on it to address this issue

src/main/java/org/opensearch/neuralsearch/search/HitsThresholdChecker.java
```diff
@@ -26,15 +26,15 @@ public HitsThresholdChecker(int totalHitsThreshold) {
...
     }
 
-    protected boolean isThresholdReached() {
+    public boolean isThresholdReached() {
         return hitCount >= getTotalHitsThreshold();
     }
 
-    protected ScoreMode scoreMode() {
+    public ScoreMode scoreMode() {
         return ScoreMode.TOP_SCORES;
```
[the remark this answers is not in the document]
vibrantvarun: The reason I changed this because currently these methods are just being used by collectors as package private method. Having access modifier protected does not have any role in its utilization in the collector. Now as per your earlier comment I moved all the collectors together in search.collector package. Therefore, the methods from HitsThresholdChecker are not accessible from the collectors. That  is why I changed this.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: The code therefore is become difficult to maintain. That is why keeping in the same class is recommended.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: made this method private

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,411 @@
...
+                numberOfSubQueries = length;
+                compoundScores = new FieldValueHitQueue[length];
+                comparators = new LeafFieldComparator[length];
+                queueFull = new boolean[length];
+                collectedHits = new int[length];
+                for (int i = 0; i < length; i++) {
+                    initializeLeafFieldComparators(context, i);
+                }
+            }
+            if (initializePerSegment) {
```
martin-gaievski, line 230: when `initializePerSegment` can be false? We're setting it to `true` in constructor

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,411 @@
...
+                    return hybridQueryScorer;
+                }
+            }
+            return null;
+        }
+
+        /*
+        The method initializes once per search request.
+         */
+        void initializePriorityQueuesWithComparators(LeafReaderContext context, int length) throws IOException {
```
martin-gaievski, line 219:
```suggestion
        void initializePriorityQueuesWithComparators(LeafReaderContext context, int  numberOfSubQueries) throws IOException {
```

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,411 @@
...
+        }
+
+        /* This method initializes the comparators per segment
+         */
+        private void initializeComparators(LeafReaderContext context, int subQueryNumber) throws IOException {
+            // as all segments are sorted in the same way, enough to check only the 1st segment for indexSort
+            if (searchSortPartOfIndexSort == null) {
+                Sort indexSort = context.reader().getMetaData().getSort();
+                searchSortPartOfIndexSort = canEarlyTerminate(sort, indexSort);
+                log.info("searchSortPartOfIndexSort " + searchSortPartOfIndexSort);
```
martin-gaievski, line 260: why do we need this line?

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,411 @@
...
+            comparators[subQueryNumber].copy(slot, doc);
+            add(slot, doc, compoundScores[subQueryNumber], subQueryNumber, score);
+            if (queueFull[subQueryNumber]) {
+                comparators[subQueryNumber].setBottom(bottom.slot);
+            }
+        }
+
+        /*
+        // This hit is competitive - replace bottom element in queue & adjustTop
+         */
```
martin-gaievski, line 320: please fix comment style, use either block or line style

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,411 @@
...
+
+        boolean thresholdCheck(int doc, int subQueryNumber) throws IOException {
+            if (collectedAllCompetitiveHits || reverseMul * comparators[subQueryNumber].compareBottom(doc) <= 0) {
+                // since docs are visited in doc Id order, if compare is 0, it means
+                // this document is larger than anything else in the queue, and
+                // therefore not competitive.
+                if (searchSortPartOfIndexSort) {
+                    if (hitsThresholdChecker.isThresholdReached()) {
+                        setTotalHitsRelation(TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO);
+                        throw new CollectionTerminatedException();
```
martin-gaievski, line 335: add a log message so we have more debug info if this happens in prod

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchResultFormatUtil.java
```diff
@@ -52,4 +57,56 @@ public static boolean isHybridQueryStartStopElement(final ScoreDoc scoreDoc) {
...
+     * @return
+     */
+    public static FieldDoc createFieldDocDelimiterElementForHybridSearchResults(final int docId, final Object[] fields) {
+        return new FieldDoc(docId, MAGIC_NUMBER_DELIMITER, fields);
+    }
+
+    /**
+     * This method is for creating dummy sort object for the field docs having magic number scores which acts as delimiters.
+     *  The sort object should be in the same type of the field on which sorting criteria is applied.
+     * @param fields contains the information about the object type of the field on which sorting criteria is applied
```
martin-gaievski, line 77: can you add an example of the format with one of these types?

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchResultFormatUtil.java
```diff
@@ -16,6 +20,7 @@ public class HybridSearchResultFormatUtil {
     // and OpenSearch convention is that scores are positive numbers
     public static final Float MAGIC_NUMBER_START_STOP = -9549511920.4881596047f;
     public static final Float MAGIC_NUMBER_DELIMITER = -4422440593.9791198149f;
+    public static final Float MAX_SCORE_WHEN_NO_HITS_FOUND = 0.0f;
```
martin-gaievski, line 23: where is this used?

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: I would not code for pagination in this iteration, mainly to make complexity of this change less. If that's a reasonable amount of effort can be remove pagination specific code? We can add it together with the rest of changes for pagination

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: Removed this as we do not need it.

src/main/java/org/opensearch/neuralsearch/search/MultiLeafFieldComparator.java
```diff
@@ -0,0 +1,100 @@
...
+    public MultiLeafFieldComparator(LeafFieldComparator[] comparators, int[] reverseMul) {
+        if (comparators.length != reverseMul.length) {
+            throw new IllegalArgumentException(
+                "Must have the same number of comparators and reverseMul, got " + comparators.length + " and " + reverseMul.length
+            );
+        }
+        this.comparators = comparators;
+        this.reverseMul = reverseMul;
+        this.firstComparator = comparators[0];
+        this.firstReverseMul = reverseMul[0];
```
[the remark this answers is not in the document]
vibrantvarun: I have already checked this before calling MultiLeafFieldComparator. If in this class I put this sanity check then I have to initialize firstComparator and firstReverseMul variables, and these 2 variables are final variables.

src/main/java/org/opensearch/neuralsearch/search/collector/SimpleFieldCollector.java
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.search.LeafCollector;
+import org.apache.lucene.search.Sort;
+import org.opensearch.neuralsearch.search.HitsThresholdChecker;
+
+/*
+  SimpleFieldCollector collects the sorted results at the shard level for every individual query.
+  It collects the list of TopFieldDocs.
+ */
+public class SimpleFieldCollector extends HybridTopFieldDocSortCollector {
```
navneet1v, line 18: if we think no one should extend this class lets make this class as Final.

src/main/java/org/opensearch/neuralsearch/search/collector/SimpleFieldCollector.java
```diff
@@ -0,0 +1,59 @@
...
+                }
+                float[] subScoresByQuery = compoundQueryScorer.hybridScores();
+                initializePriorityQueuesWithComparators(context, subScoresByQuery.length);
+                incrementTotalHitCount();
+                for (int i = 0; i < subScoresByQuery.length; i++) {
+                    float score = subScoresByQuery[i];
+                    // if score is 0.0 there is no hits for that sub-query
+                    if (score == 0) {
+                        continue;
+                    }
```
navneet1v, line 44: Looking at this why we are incrementing the hitcount when in the for loop there is a possibility that score can be 0?

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: Consider there are 2 shards and each shard has 3 segments.

The leaf collector gets initialized per segment. 
Therefore, the code which is under initializePerSegment will run per segment once during initialization.

As soon as that code runs, it will set the variable to false.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
vibrantvarun: Consider in future when we roll out pagination feature then there are 4 types of different pagination. All cannot go into the same PR as they are altogether different as a individual. Now, the thing with `search_after` it is widely used with sorting, that is why its code looks almost same with simpleFieldCollector. Later, when we roll out pagination we need not have to touch sorting again for this purpose. That is the reason I am rolling `search_after` feature with this PR.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,403 @@
...
+            return null;
+        }
+
+        /*
+        Increment total hit count and validate if threshold is reached.
+         */
+        protected void incrementTotalHitCount() throws IOException {
+            totalHits++;
+            hitsThresholdChecker.incrementHitCount();
+            if (scoreMode().isExhaustive() == false
```
Vikasht34, line 175: Can we remove notifying other comparators from this function and keep this function only to increasecount, Right now this functionName and responsibility does not match . 
Other reason is , in If Statement , there is 3 checks we are performing and that's going to happen every time when we Increment . If we can early terminate also that could be beneficial with that we wil saving some operations in if statement.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,403 @@
...
+        }
+        return topFieldDocs;
+    }
+
+    @Override
+    public ScoreMode scoreMode() {
+        return hitsThresholdChecker.scoreMode();
+    }
+
+    protected abstract class HybridTopDocSortLeafCollector implements LeafCollector {
```
Vikasht34, line 117: Why this class has to be part of this file? and why this class has to be abstract , when there is no one extending this class and there is no concrete implementation from this class.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,403 @@
...
+
+        public HybridTopDocSortLeafCollector(Sort sort, @Nullable FieldDoc after) {
+            this.sort = sort;
+            this.after = after;
+            this.initializePerSegment = true;
+        }
+
+        @Override
+        public void setScorer(final Scorable scorer) throws IOException {
+            if (scorer instanceof HybridQueryScorer) {
```
Vikasht34, line 133: What if Scorer is null !! , it is going to break the flow.. Can we check NPE here

Vikasht34: I am bit confused on what we are checking here , 

If you see if Statement , it checks it is instance of HybridQueryScorer and again in else also we are checking the same thing in different function getHybridQueryScorer , 

if (scorer == null) {
                return null;
            }
            if (scorer instanceof HybridQueryScorer) {
                return (HybridQueryScorer) scorer;
            }
            
            Can we remove this multiple if else and reduce cyclometric complexicity of this code . 
            
            and It's fundamentally wrong becuase same check is happening in if and else.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,403 @@
...
+                if (Objects.isNull(compoundQueryScorer)) {
+                    log.error(
+                        String.format(Locale.ROOT, "cannot find scorer of type HybridQueryScorer in a hierarchy of scorer %s", scorer)
+                    );
+                }
+            }
+        }
+
+        private HybridQueryScorer getHybridQueryScorer(final Scorable scorer) throws IOException {
+            if (scorer == null) {
```
Vikasht34, line 147: this has to moved to parent function if scorer is null !! I mean in above function.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,403 @@
...
+                return null;
+            }
+            if (scorer instanceof HybridQueryScorer) {
+                return (HybridQueryScorer) scorer;
+            }
+            for (Scorable.ChildScorable childScorable : scorer.getChildren()) {
+                HybridQueryScorer hybridQueryScorer = getHybridQueryScorer(childScorable.child);
```
Vikasht34, line 154: so is this idea of getting one child score , is that reason of recursive call?

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,403 @@
...
+import org.opensearch.common.Nullable;
+import org.opensearch.neuralsearch.search.HitsThresholdChecker;
+import org.opensearch.neuralsearch.search.lucene.MultiLeafFieldComparator;
+
+/*
+ Collects the TopFieldDocs after executing hybrid query. Uses HybridQueryTopDocs as DTO to handle each sub query results.
+ The individual query results are sorted as per the sort criteria sent in the search request.
+ */
+@Log4j2
+public abstract class HybridTopFieldDocSortCollector implements Collector {
```
Vikasht34, line 41: I am failing to understand why we need this abstract class , Since Collector is already interface which has 3 functions 
1. getLeafCollector 2. scoreMode and 3. setWeight 

now Why dont' we have two class SingleFieldCollector and PagingField collector , directly implements collector infetrface and then a common helper class which implements leafCollector which will be intialized from above two classes to get leafCollector.

and when we are initalizing the class , intialize through interface not like new Class()

Vikasht34 (commented):
I will do s

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: This is the change in the design of how it is done in the current HybridQuery collector. If any change we seek in the Hybrid Query then it can be catered in other issue.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: Because collect method which is used in SimpleFieldCollector and PagingFieldCollector is defined in LeafCollector interface.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: No, it is not same. The scorer can be nested.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: It is the change in the design from TopFieldCollector from lucene. 
By your suggestion
A helper class will be an abstract class because collect method needs to be defined in the collector not in the helper class. That is what I have exactly done now.

I have created an abstract class which implements Collector. 
The collector method has 3 interface getLeafCollector, scoreMode and setWeight

getLeafCollector definition is different in case of PagingFieldColletor(sort, after) and in case of SimpleFieldCollector(sort, null) 

In the Leafcollector also the collect method implementation is different for both the cases. So there main collector class HybridTopFieldDocSortCollector will also be abstract and the leafCollector class HybridTopFieldDocLeafCollector class is also abstract.

Now the reason why HybridTopFieldDocLeafCollector cannot be a standalone class because it has lot of operations on the variables defined in the main collector class. Maintaining those variable states for every segment and shard is very difficult for every shard and segments.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: Yes because it can be nested.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: Can you check this class once https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java?

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: No, we can't do that because this operation has to performed after incrementing total hits. 
Check this out https://github.com/apache/lucene/blob/057cbf3c8665354cd21dac20570a7e088dc8ce3a/lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java#L74

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: This call is made from core that too at Shard level.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
martin-gaievski: @Vikasht34 income scorer object can be a compound  object that has multi level hierarchy of nested scorers. Some of them can be null, and this is a legit case, for example some scenarios of knn query.
We are looking for one scorer that is of type HybridQueryScorer, once that found we stop the search and return that scorer as global result of all recursive calls.
If in the end we couldn't find hybrid scorer we fail with exception, that's why that check is needed in the top level calling function.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
vibrantvarun: Discussed offline with @martin-gaievski and we are keeping this collector in this PR.

martin-gaievski: To share the context - wording in previous comment need to be reviewed, this is part of the sorting feature, fac that it can be re-used in "pagination" is just a nice bonus, not a goal

src/main/java/org/opensearch/neuralsearch/search/collector/SimpleFieldCollector.java
```diff
@@ -0,0 +1,59 @@
...
```
navneet1v, line 44: This condition is making me shiver, as what I know is there can be 0 score value too at-least for k-NN with script score I have seen it.

src/main/java/org/opensearch/neuralsearch/search/collector/PagingFieldCollector.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.lucene.search.LeafCollector;
+import org.apache.lucene.search.Sort;
+import org.opensearch.common.Nullable;
+import org.opensearch.neuralsearch.search.HitsThresholdChecker;
+
+/**
+ * PagingFieldCollector collects the sorted results at the shard level for every individual query
+ * as per search_after criteria applied in the search request.
+ * It collects the list of TopFieldDocs.
+ */
```
navneet1v, line 20: @vibrantvarun Given hybrid query doesn't support pagination and I know we want to build this pagination with sort, my worry is customer behavior will be broken where with normal hybrid search pagination doesn't work with then moment we add sort it starts to work. What is your thought?

src/main/java/org/opensearch/neuralsearch/search/collector/PagingFieldCollector.java
```diff
@@ -0,0 +1,101 @@
...
+        super(numHits, hitsThresholdChecker, sort);
+        this.sort = sort;
+        this.after = after;
+    }
+
+    @Override
+    public LeafCollector getLeafCollector(LeafReaderContext context) {
+        docBase = context.docBase;
+        final int afterDoc = after.doc - docBase;
+        return new HybridTopDocSortLeafCollector(sort, after) {
```
navneet1v, line 36: The logic of this HybridTopDocSortLeafCollector and the one in the SimpleFieldCollector is mostly same. Or at-least we can some part are pretty common, I think we can move some code back to Base class thoughts?

src/main/java/org/opensearch/neuralsearch/search/collector/PagingFieldCollector.java
```diff
@@ -0,0 +1,101 @@
...
+                    }
+
+                    // if queueFull[i] is true then it indicates
+                    // that we have found the results equal to the size sent in the search request.
+                    if (queueFull[i]) {
+                        // If threshold is reached then return. Default value of threshold is 10000.
+                        if (thresholdCheck(doc, i)) {
+                            return;
+                        }
+                    }
```
navneet1v, line 59: why we are doing this check here before knowing result is on previous page or not? plus this is not same as SimpleFieldCollector why is that?

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,410 @@
...
+        protected void collectCompetitiveHit(int doc, int subQueryNumber) throws IOException {
+            // This hit is competitive - replace bottom element in queue & adjustTop
+            if (numHits > 0) {
+                comparators[subQueryNumber].copy(bottom.slot, doc);
+                updateBottom(doc, compoundScores[subQueryNumber]);
+                comparators[subQueryNumber].setBottom(bottom.slot);
+            }
+        }
+
+        protected boolean thresholdCheck(int doc, int subQueryNumber) throws IOException {
```
navneet1v, line 223: why protected?

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,410 @@
...
+         2. Also, after initializing for every segment the comparators has to be refreshed.
+        Therefore, to do the above two things lazily we have to use a flag initializeComparatorsPerSegmentwhich is set to true when a leafCollector is initialized per segment.
+        Later, in the collect method when number of sub-queries has been found then initialize the comparators(1) or (2) refresh the comparators and set the flag to false.
+        */
+        private boolean initializeComparatorsPerSegment;
+
+        public HybridTopDocSortLeafCollector(Sort sort, @Nullable FieldDoc after) {
+            this.sort = sort;
+            this.after = after;
+            this.initializeComparatorsPerSegment = true;
```
navneet1v, line 135: I think you mean `initializeComparatorsPerShard` and not `initializeComparatorsPerSegment`.

Better name of consider, `isComparatorsIntialized`

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: Because in SimpleFieldCollector and PagingFieldCollector **getLeafCollector** method, we initialize the object of HybridTopFieldDocLeafCollector  by defining collect method in it. Now as HybridTopFieldDocLeafCollector is an **abstract** class we need this access modifier to be **protected** so that the places where we initialize the abstract leaf collector we can call this thresholdcheck method in the collect method.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,410 @@
...
+import org.opensearch.common.Nullable;
+import org.opensearch.neuralsearch.search.HitsThresholdChecker;
+import org.opensearch.neuralsearch.search.lucene.MultiLeafFieldComparator;
+
+/*
+ Collects the TopFieldDocs after executing hybrid query. Uses HybridQueryTopDocs as DTO to handle each sub query results.
+ The individual query results are sorted as per the sort criteria sent in the search request.
+ */
+@Log4j2
+public abstract class HybridTopFieldDocSortCollector implements Collector {
```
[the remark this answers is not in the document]
vibrantvarun: This class needs to be public because we are checking the instance of this class in HybridCollectorManger reduce method to get top docs.

src/main/java/org/opensearch/neuralsearch/search/collector/SimpleFieldCollector.java
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.lucene.search.LeafCollector;
+import org.apache.lucene.search.Sort;
+import org.opensearch.neuralsearch.search.HitsThresholdChecker;
+
+/**
+ * SimpleFieldCollector collects the sorted results at the shard level for every individual query.
+ * It collects the list of TopFieldDocs.
+ */
+public final class SimpleFieldCollector extends HybridTopFieldDocSortCollector {
+    private final Sort sort;
```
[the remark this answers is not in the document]
vibrantvarun: Yes it can be removed.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -211,6 +301,56 @@ private TotalHits getTotalHits(
...
+            }
+            // format scores using following template:
+            // doc_id | magic_number_1
+            // doc_id | magic_number_2
+            // ...
+            // doc_id | magic_number_2
+            // ...
+            // doc_id | magic_number_2
+            // ...
+            // doc_id | magic_number_1
```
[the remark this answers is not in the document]
vibrantvarun: Oh yes, making the change now

src/main/java/org/opensearch/neuralsearch/search/collector/PagingFieldCollector.java
vibrantvarun: Lets understand with an example
Consider there is a search_after condition to fetch results after value 40. Now from the total results 30 results fall after value 30. 
ie. 41......70

Now in the search query I have give the results size = 10

The final result should be from 41....50. 

Now answering your question

```
 if (queueFull[i]) {
                        // If threshold is reached then return. Default value of threshold is 10000.
                        if (thresholdCheck(doc, i)) {
                            return;
                        }
                    }
```

This code segment checks for that results size i.e I got 10 results already. 

Then we check whether all results has to be collected or not. i.e If I haven't found 10 results then continue the collection of docId's

Now if  result size is let say 100 and as I mentioned earlier the total results are 30. So it will continue iterating and the above condition of queue[i] will not fulfilled therefore we will keep on iterating.

Furthermore, answering your second concern the SimpleFieldCollector is just responsible for sorting the results. Therefore if the the queue is full it means I have found the desired number of results needed from size param sent in the search request. 


I hope that clears everything.

src/main/java/org/opensearch/neuralsearch/search/collector/PagingFieldCollector.java
vibrantvarun: No it is not the same, here in the pagingFieldCollector the things which are different 
1) the docBase on which the whole sorting operates.
2) the position of different types of checks like when to check queue[i] is full and whether results are found on previous page or not. 

Lot of if else would be there on the basis of what is sent in the after param of the search request. The code in the common place will become so difficult to understand and maintain.

src/main/java/org/opensearch/neuralsearch/search/collector/SimpleFieldCollector.java
vibrantvarun: @martin-gaievski can you share your thoughts on how hybrid query is catering this case currently?

src/main/java/org/opensearch/neuralsearch/search/collector/PagingFieldCollector.java
vibrantvarun: Let me put it this way,

Pagination is of 4 types
1) from and size parameters
2) Scroll
3) `search_after` exclusively used with sorting
4) Point in time with `search_after`

Here I have covered only 3rd point that just not as pagination but an extended use case for sort. The reason why I have covered it is because it is almost the similar amount of work. For customer, It will be like `search_after` with sorting. Moreover, the `search_after` is not used without sorting. So the customer will not try to use `search_after` with hybrid query without sort, therefore no  customer behavior would be there.

If `search_after` could be used with or without sorting then we can say there is a difference in customer behavior.

src/main/java/org/opensearch/neuralsearch/search/collector/SimpleFieldCollector.java
martin-gaievski: for hybrid query we do not adding such hit to sub-query result -> it will not be in the final list of hits. If score = 0. is a valid case we may change the logic and add treat such hit similar to others. 
Change should be simple, score = 0 chosen as a stop value because we initialize array of hybrid scores with 0 in [scorer](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/query/HybridQueryScorer.java#L183-L196). We can change it to something else and check for that new value instead of 0.

vibrantvarun: Okay then lets open an issue separately to cater this change as it would require changes in current hybrid query workflow.  When the change in the current workflow would come we can cater the same in SimpleFieldCollector and Paging FieldCollector

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,408 @@
...
+      This ensures that the compareBottom method adjusts the order based on whether you want ascending or descending sorting.
+     */
+    protected int reverseMul;
+    protected FieldValueHitQueue<FieldValueHitQueue.Entry>[] compoundScores;
+    protected boolean queueFull[];
+    @Getter
+    protected float maxScore = 0.0f;
+    protected int[] collectedHits;
+    /*
+       searchSortPartOfIndexSort is used to evaluate whether to perform index sort or not.
```
martin-gaievski, line 67: nit: can be a line comment

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -149,9 +183,70 @@ public ReduceableSearchResult reduce(Collection<Collector> collectors) {
             TopDocsAndMaxScore topDocsAndMaxScore = new TopDocsAndMaxScore(newTopDocs, hybridTopScoreDocCollector.getMaxScore());
             return (QuerySearchResult result) -> { result.topDocs(topDocsAndMaxScore, getSortValueFormats(sortAndFormats)); };
         }
+
+        if (!hybridSortedTopDocCollectors.isEmpty()) {
+            HybridTopFieldDocSortCollector hybridSortedTopScoreDocCollector = hybridSortedTopDocCollectors.stream()
```
martin-gaievski, line 188: please add a TODO note here, we already know that this logic is incorrect for case of multiple segment slices -> multiple collectors.

martin-gaievski (commented):
Can we have few basic unit tests for your new sorting collector and additions to hybrid collector manager?

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchResultFormatUtil.java
```diff
@@ -52,4 +56,62 @@ public static boolean isHybridQueryStartStopElement(final ScoreDoc scoreDoc) {
     public static boolean isHybridQueryDelimiterElement(final ScoreDoc scoreDoc) {
         return Objects.nonNull(scoreDoc) && scoreDoc.doc >= 0 && Float.compare(scoreDoc.score, MAGIC_NUMBER_DELIMITER) == 0;
     }
+
+    public static FieldDoc createFieldDocStartStopElementForHybridSearchResults(final int docId, final Object[] fields) {
```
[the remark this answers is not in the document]
vibrantvarun: Due to 4 reasons I am not in favor of combining them.
1. Keeping method different gives you more Readability and simplicity.
2. I need to pass an extra argument from the place where it is called. That argument will have same constant values MAGIC_NUMBER_START_STOP and MAGIC_NUMBER_DELIMITER which are declared in HybridSearchResultFormatUtil. Therefore, to do that I need to import both the variables at 2 places HybridCollectorManager, In Normalization process . It is an unoptimized way. 
3. I have to put if and else which will create overhead. 
4. Need to make changes for Hybrid query as well as in this class you see the earlier methods are used in the same way.

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchResultFormatUtil.java
```diff
@@ -52,4 +56,62 @@ public static boolean isHybridQueryStartStopElement(final ScoreDoc scoreDoc) {
...
+    }
+
+    /**
+     * This method is for creating dummy sort object for the field docs having magic number scores which acts as delimiters.
+     *  The sort object should be in the same type of the field on which sorting criteria is applied.
+     * @param fields contains the information about the object type of the field on which sorting criteria is applied
+     * @return
+     */
+    public static Object[] createSortFieldsForDelimiterResults(final Object[] fields) {
+        Object SORT_FIELDS_FOR_DELIMITER_RESULTS;
```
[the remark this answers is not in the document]
vibrantvarun: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -211,6 +302,58 @@ private TotalHits getTotalHits(
         return new TotalHits(maxTotalHits, relation);
     }
 
+    private TopDocs getNewTopFieldDocs(final TotalHits totalHits, final List<TopFieldDocs> topFieldDocs, final SortField sortFields[]) {
+        FieldDoc[] fieldDocs = new FieldDoc[0];
```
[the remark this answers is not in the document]
vibrantvarun: ack. Btw, topFieldDoc.scoreDocs.length==0 does not mean adding delimiter, it means no result found for that subquery.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -86,21 +100,33 @@ public static CollectorManager createHybridCollectorManager(final SearchContext
...
+                filteringWeight,
+                searchContext.searchAfter()
             );
     }
 
     @Override
     public Collector newCollector() {
-        Collector hybridcollector = new HybridTopScoreDocCollector(numHits, hitsThresholdChecker);
+        Collector hybridcollector;
+        if (sortAndFormats != null) {
```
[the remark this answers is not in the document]
vibrantvarun: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -149,9 +183,71 @@ public ReduceableSearchResult reduce(Collection<Collector> collectors) {
...
+            TopDocsAndMaxScore topDocsAndMaxScore = new TopDocsAndMaxScore(newTopDocs, maxScore);
+            return (QuerySearchResult result) -> { result.topDocs(topDocsAndMaxScore, getSortValueFormats(sortAndFormats)); };
+        }
+
         throw new IllegalStateException("cannot collect results of hybrid search query, there are no proper score collectors");
     }
 
+    private static void validateSortCriteria(SearchContext searchContext, boolean trackScores) {
+        SortField[] sortFields = searchContext.sort().sort.getSort();
+        boolean isSortByField = false;
```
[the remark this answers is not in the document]
vibrantvarun: ack

vibrantvarun:
> Can we have few basic unit tests for your new sorting collector and additions to hybrid collector manager?

@martin-gaievski added the basic unit tests.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,407 @@
...
+        protected boolean thresholdCheck(int doc, int subQueryNumber) throws IOException {
+            if (collectedAllCompetitiveHits || reverseMul * comparators[subQueryNumber].compareBottom(doc) <= 0) {
+                // since docs are visited in doc Id order, if compare is 0, it means
+                // this document is larger than anything else in the queue, and
+                // therefore not competitive.
+                if (searchSortPartOfIndexSort) {
+                    if (hitsThresholdChecker.isThresholdReached()) {
+                        setTotalHitsRelation(TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO);
+                        log.info("Terminating collection as hits threshold is reached");
+                        throw new CollectionTerminatedException();
```
martin-gaievski, line 229: why this exception? I don't see a code where we're handling this exception type in some special way. I would rather throw IllegalState where I can put some meaningful error message that gives some clues to users and engineers while troubleshooting.
Also why logging is at info level, if we're throwing runtime exception it should be an error level.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,407 @@
...
+            if (compoundScores == null) {
+                compoundScores = new FieldValueHitQueue[numberOfSubQueries];
+                comparators = new LeafFieldComparator[numberOfSubQueries];
+                queueFull = new boolean[numberOfSubQueries];
+                collectedHits = new int[numberOfSubQueries];
+                for (int i = 0; i < numberOfSubQueries; i++) {
+                    initializeLeafFieldComparators(context, i);
+                }
+            }
+            if (isComparatorsInitialized) {
```
martin-gaievski, line 252: shouldn't it be the other way around - check and initialize comparators if they are not yet initialized? Either the logic is wrong or variable names are not descriptive

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,407 @@
...
+    // Add the entry in the Priority queue
+    private void add(int slot, int doc, FieldValueHitQueue<FieldValueHitQueue.Entry> compoundScore, int subQueryNumber, float score) {
+        FieldValueHitQueue.Entry bottomEntry = new FieldValueHitQueue.Entry(slot, docBase + doc);
+        bottomEntry.score = score;
+        bottom = compoundScore.add(bottomEntry);
+        // The queue is full either when totalHits == numHits (in SimpleFieldCollector), in which case
+        // slot = totalHits - 1, or when hitsCollected == numHits (in PagingFieldCollector this is hits
+        // on the current page) and slot = hitsCollected - 1.
+        assert slot < numHits;
+        queueFull[subQueryNumber] = slot == numHits - 1;
```
martin-gaievski, line 372: nit: I would rather put this in a more expanded form for better readability

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,407 @@
...
+
+    private boolean canEarlyTerminateOnDocId(Sort searchSort) {
+        final SortField[] fields1 = searchSort.getSort();
+        return SortField.FIELD_DOC.equals(fields1[0]);
+    }
+
+    private boolean canEarlyTerminateOnPrefix(Sort searchSort, Sort indexSort) {
+        if (indexSort != null) {
+            final SortField[] fields1 = searchSort.getSort();
+            final SortField[] fields2 = indexSort.getSort();
```
martin-gaievski, line 392: please use more meaningful variable names instead of `fields1` and `fields2`

src/main/java/org/opensearch/neuralsearch/search/collector/SimpleFieldCollector.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public SimpleFieldCollector(int numHits, HitsThresholdChecker hitsThresholdChecker, Sort sort) {
+        super(numHits, hitsThresholdChecker, sort);
+    }
+
+    @Override
+    public LeafCollector getLeafCollector(LeafReaderContext context) {
+        docBase = context.docBase;
+
+        return new HybridTopDocSortLeafCollector(null) {
```
martin-gaievski, line 28: this looks weird, it should be possible to define no args constructor that passes param as `null` at parent level (HybridTopDocSortLeafCollector) extend it at the child level (SimpleFieldCollector).

src/main/java/org/opensearch/neuralsearch/search/lucene/MultiLeafFieldComparator.java
```diff
@@ -0,0 +1,119 @@
...
+    private final int[] reverseMul;
+    // we extract the first comparator to avoid array access in the common case
+    // that the first comparator compares worse than the bottom entry in the queue
+    private final LeafFieldComparator firstComparator;
+    private final int firstReverseMul;
+
+    public MultiLeafFieldComparator(LeafFieldComparator[] comparators, int[] reverseMul) {
+        if (comparators.length != reverseMul.length) {
+            throw new IllegalArgumentException(
+                "Must have the same number of comparators and reverseMul, got " + comparators.length + " and " + reverseMul.length
```
martin-gaievski, line 29: use String.format with params

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -86,24 +100,40 @@ public static CollectorManager createHybridCollectorManager(final SearchContext
...
+        }
+        // Hybrid Query with Sorting Applied
+        Collector hybridSortcollector;
+        if (after == null) {
+            hybridSortcollector = new SimpleFieldCollector(numHits, hitsThresholdChecker, sortAndFormats.sort);
+        } else {
+            validateSearchAfterFieldAndSortFormats();
+            hybridSortcollector = new PagingFieldCollector(numHits, hitsThresholdChecker, sortAndFormats.sort, after);
+        }
+        // Check if filterWeight is present. If it is present then return wrap Hybrid Sort collector object underneath the FilteredCollector
```
martin-gaievski, line 134: you can do whole this method simpler. Write one private method that will give you a low level collector that is either hybrid topscore or one of sorting specific collectors. Then with that collector do the old logic. It should be possible as all collectors are "isa" Collector interface

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -119,6 +149,7 @@ public Collector newCollector() {
     @Override
     public ReduceableSearchResult reduce(Collection<Collector> collectors) {
         final List<HybridTopScoreDocCollector> hybridTopScoreDocCollectors = new ArrayList<>();
+        final List<HybridTopFieldDocSortCollector> hybridSortedTopDocCollectors = new ArrayList<>();
```
martin-gaievski, line 152: getting these hybrid collectors can be in it's own method to keep `reduce` simple. Main question - do you see the scenario when both topdocs and sorting are possible? If not, then we can get away with 1 - simpler main method logic - either this or that, and 2 - maybe a single list of collectors, I hope that collector methods that we're calling here are common between all hybrid specific collectors

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -149,9 +187,71 @@ public ReduceableSearchResult reduce(Collection<Collector> collectors) {
...
+    }
+
+    private void validateSearchAfterFieldAndSortFormats() {
+        if (after.fields == null) {
+            throw new IllegalArgumentException("after.fields wasn't set; you must pass fillFields=true for the previous search");
+        }
+
+        if (after.fields.length != sortAndFormats.sort.getSort().length) {
+            throw new IllegalArgumentException(
+                "after.fields has " + after.fields.length + " values but sort has " + sortAndFormats.sort.getSort().length
```
martin-gaievski, line 250: use String.format for complex messages

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -211,6 +306,61 @@ private TotalHits getTotalHits(
         return new TotalHits(maxTotalHits, relation);
     }
 
+    private TopDocs getNewTopFieldDocs(final TotalHits totalHits, final List<TopFieldDocs> topFieldDocs, final SortField sortFields[]) {
```
martin-gaievski, line 309: nit: I feel like we need a new abstraction for getting these top docs, we can do it later in separate PR

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,407 @@
...
+    protected FieldValueHitQueue<FieldValueHitQueue.Entry>[] compoundScores;
+    protected boolean queueFull[];
+    @Getter
+    protected float maxScore = 0.0f;
+    protected int[] collectedHits;
+
+    // searchSortPartOfIndexSort is used to evaluate whether to perform index sort or not.
+    private Boolean searchSortPartOfIndexSort = null;
+
+    private static final TopFieldDocs EMPTY_TOPDOCS = new TopFieldDocs(
```
martin-gaievski, line 70:
```suggestion
    private static final TopFieldDocs EMPTY_TOP_FIELD_DOCS = new TopFieldDocs(
```

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: In opensearch this exception is caught https://github.com/opensearch-project/OpenSearch/blob/0d01d1755e282763e5ea020112617ba25459863c/server/src/main/java/org/opensearch/search/internal/ContextIndexSearcher.java#L328

vibrantvarun: Also this is not an error https://github.com/opensearch-project/OpenSearch/blob/0d01d1755e282763e5ea020112617ba25459863c/server/src/main/java/org/opensearch/search/internal/ContextIndexSearcher.java#L329 Check this comment out

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: No, it does not work that ways.
1. The code which is under the flag **isComparatorsInitialized** needs to be executed for every segment.
2. HybridTopDocSortLeafCollector constructor gets initialized for every segment. Therefore, comparators need to be refreshed for every segment.

The reason why we need this flag is because we need to lazy init this.

Reference:
https://github.com/apache/lucene/blob/7e31f56ea1130909948bcf9e5beefc161ceef137/lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java#L52

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,407 @@
...
+     */
+    protected int reverseMul;
+    protected FieldValueHitQueue<FieldValueHitQueue.Entry>[] compoundScores;
+    protected boolean queueFull[];
+    @Getter
+    protected float maxScore = 0.0f;
+    protected int[] collectedHits;
+
+    // searchSortPartOfIndexSort is used to evaluate whether to perform index sort or not.
+    private Boolean searchSortPartOfIndexSort = null;
```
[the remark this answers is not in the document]
vibrantvarun: The reason why it is kept as null because we need to make sure that for every shard this value has to be **null** and to avoid having any garbage value this condition is a good to keep.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
```diff
@@ -0,0 +1,407 @@
...
+ Collects the TopFieldDocs after executing hybrid query. Uses HybridQueryTopDocs as DTO to handle each sub query results.
+ The individual query results are sorted as per the sort criteria sent in the search request.
+ */
+@Log4j2
+public abstract class HybridTopFieldDocSortCollector implements Collector {
+    private final int numHits;
+    private final HitsThresholdChecker hitsThresholdChecker;
+    private final Sort sort;
+    private FieldComparator<?> firstComparator;
+    private FieldValueHitQueue.Entry bottom = null;
```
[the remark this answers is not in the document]
vibrantvarun: The reason why it is kept as null because we need to make sure that for every shard this value has to be null and to avoid having any garbage value this condition is a good to keep.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: ack

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: ack

src/main/java/org/opensearch/neuralsearch/search/collector/SimpleFieldCollector.java
vibrantvarun: ack

src/main/java/org/opensearch/neuralsearch/search/lucene/MultiLeafFieldComparator.java
vibrantvarun: ack

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java
vibrantvarun: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
vibrantvarun: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
vibrantvarun: Discussed offline. The refactoring will be catered once I will raise the bug fix Missing docs when concurrent search enabled in this branch.

martin-gaievski (approved):
Looks good to me, thanks!

Vikasht34 (approved):

src/main/java/org/opensearch/neuralsearch/search/collector/PagingFieldCollector.java
navneet1v: As per our discussion and looking at this feature that search_after is used only when sorting is present, lets make sure that we document this properly in the docs. We should also mention in the docs that search_after is only supported with sorting. This part is crirtical

navneet1v (approved):

Issue #799, opened 2024-06-19
[BUG] Hybrid Search results are missing when concurrent search is enabled


vibrantvarun:
### What is the bug?
Hybrid search results are missing when concurrent search is enabled in single shard and multiple shards scenario.  
**For example:**
For **single** shard I ingested 6 documents and did a matchAll query under the hybrid clause then I am expecting **6** results should be there but I am getting just **5** results.

For **multiple** shard I did the same by ingesting 100 documents and I am expecting **100** results but getting **95**.

### How can one reproduce the bug?
1) Just ingest the documents for single shard by following URL  http://localhost:9200/my-nlp-index/_doc/1?refresh=true and make sure update docId when ingesting new documents.
2) Enable the concurrent search in index settings
3) For multiple shard use code to ingest 100 documents in a same way
4) Do a match all query with maximum results size, you will find the results are missing.

### What is the expected behavior?
Results should not be missing and the right of results should be returned.

### Do you have any additional context?
When concurrent search is disabled everything is working fine.

vibrantvarun:
@martin-gaievski assigning this to you

vibrantvarun:
@martin-gaievski Can we close the bug?

martin-gaievski:
yup

PR #800, opened 2024-06-21
Fix for missing HybridQuery results when concurrent segment search is enabled

martin-gaievski:
### Description
Fixed gap in implementation for case when shard has multiple (6+) segments. In such case some hits were missing in final hybrid query result.
Issue caused by wrong assumption that collector manager can have only one hybrid query result collector in reduce phase. It's true for number of segments < 6, otherwise core passes multiple collectors with a portion of result each. We have to merge those results at shard level to have correct collection of hits.

[Lucene code](https://github.com/apache/lucene/blob/branch_9_11/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java#L108-L110) where they define limit per slice (block that is processed by one collector). It's max of 250.000 docs or 5 segments.

Logic for merge is a bit tricky because we have to deal with TopDocs that has been formatted to special hybrid query format. Each next collector result should be merge into query result one by one. On each merge we need to find results of one sub-query and merge then separately, then wrap into hybrid query result format.

Example:
TopDocs in query result:
```
query1: {doc1: 10, doc3: 5, doc5: 3}
query2: {doc2: 3, doc3: 1}
```
 
result from next block of segments:
```
query1: {doc2: 11, doc6: 2}
query2: {doc5: 2, doc6: 1}
```

merged result:
```
query1: {doc2:11, doc1: 10, doc3: 5, doc5: 3, doc6: 2}
query2: {doc2: 3, doc5: 2, doc3:1, doc6: 1}
```

Added extensive list of unit tests and integ test that would fail with old logic (assertion of total hits, actual number would be lower).

### Issues Resolved
https://github.com/opensearch-project/neural-search/issues/799

### Check List
    - [] All tests pass


src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -44,9 +49,9 @@
 @RequiredArgsConstructor
 public abstract class HybridCollectorManager implements CollectorManager<Collector, ReduceableSearchResult> {
 
+    private static final int MIN_NUMBER_OF_ELEMENTS_IN_SCORE_DOC = 3;
     private final int numHits;
     private final HitsThresholdChecker hitsThresholdChecker;
-    private final boolean isSingleShard;
```
martin-gaievski, line 49: not related to this change just a code cleanup, field isn't used anywhere in the code

martin-gaievski:
BWC for 2.15 will keep failing because main is still pointing to snapshot and as per our process we're changing that after the release.
Security check CI action will also fail for jdk 11 and 17, core has changed min requirement for security plugin to jdk 21. CI for jdk 21 is passing

vibrantvarun:
Great Job @martin-gaievski for raising this PR.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryAggregationsIT.java
```diff
@@ -100,43 +100,43 @@ protected boolean preserveClusterUponCompletion() {
 
     @SneakyThrows
     public void testPipelineAggs_whenConcurrentSearchEnabled_thenSuccessful() {
-        updateClusterSettings("search.concurrent_segment_search.enabled", true);
```
vibrantvarun, line 103: Can we also do the same change by using CONCURRENT_SEGMENT_SEARCH_ENABLED in other places also like HybridSearchPostFilterIT?

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -138,16 +139,36 @@ public ReduceableSearchResult reduce(Collection<Collector> collectors) {
...
+            List<ReduceableSearchResult> results = new ArrayList<>();
+            for (HybridTopScoreDocCollector hybridTopScoreDocCollector : hybridTopScoreDocCollectors) {
+                List<TopDocs> topDocs = hybridTopScoreDocCollector.topDocs();
+                TopDocs newTopDocs = getNewTopDocs(
+                    getTotalHits(this.trackTotalHitsUpTo, topDocs, hybridTopScoreDocCollector.getTotalHits()),
+                    topDocs
+                );
+                TopDocsAndMaxScore topDocsAndMaxScore = new TopDocsAndMaxScore(newTopDocs, hybridTopScoreDocCollector.getMaxScore());
+
+                results.add((QuerySearchResult result) -> {
```
vibrantvarun, line 151: Can we break results.add and the logic inside it into separate parts? It becomes too difficult to understand. Probably it can be moved to a new method.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -138,16 +139,36 @@ public ReduceableSearchResult reduce(Collection<Collector> collectors) {
...
+                    // this is case of first collector, query result object doesn't have any top docs set, so we can
+                    // just set new top docs without merge
+                    if (result.hasConsumedTopDocs()) {
+                        result.topDocs(topDocsAndMaxScore, getSortValueFormats(sortAndFormats));
+                        return;
+                    }
+                    // in this case top docs are already present in result, and we need to merge next result object with what we have.
```
vibrantvarun, line 158: I think the comment on line 158 should be present on line 162.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -138,16 +139,36 @@ public ReduceableSearchResult reduce(Collection<Collector> collectors) {
...
+                    // if collector doesn't have any hits we can just skip it and save some cycles by not doing merge
+                    if (newTopDocs.totalHits.value == 0) {
+                        return;
+                    }
+                    // we need to do actual merge because query result and current collector both have some score hits
+                    TopDocsAndMaxScore originalTotalDocsAndHits = result.topDocs();
+                    result.topDocs(
+                        mergeTopDocsAndMaxScores(originalTotalDocsAndHits, topDocsAndMaxScore),
+                        getSortValueFormats(sortAndFormats)
```
vibrantvarun, line 167: Can we make sure this computation getSortValueFormats is done once? Because it will not change with new collector instance. It will be same for all instances of HybridTopScoreDocCollector so just calculate it before line 142.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -215,6 +231,109 @@ private DocValueFormat[] getSortValueFormats(final SortAndFormats sortAndFormats
         return sortAndFormats == null ? null : sortAndFormats.formats;
     }
 
+    private ReduceableSearchResult reduceCollectorResults(List<ReduceableSearchResult> results) {
```
vibrantvarun, line 234: What exactly are we trying to do here? Can you add a comment. What exactly this r.reduce method will do?

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -215,6 +231,109 @@ private DocValueFormat[] getSortValueFormats(final SortAndFormats sortAndFormats
...
+    ) {
+        if (Objects.requireNonNull(sourceScoreDocs).length < MIN_NUMBER_OF_ELEMENTS_IN_SCORE_DOC
+            || Objects.requireNonNull(newScoreDocs).length < MIN_NUMBER_OF_ELEMENTS_IN_SCORE_DOC) {
+            throw new IllegalArgumentException("cannot merge top docs because it does not have enough elements");
+        }
+        // we overshoot and preallocate more than we need - length of both top docs combined.
+        // we will take only portion of the array at the end
+        List<ScoreDoc> mergedScoreDocs = new ArrayList<>(sourceScoreDocs.length + newScoreDocs.length);
+        int sourcePointer = 0;
+        mergedScoreDocs.add(sourceScoreDocs[sourcePointer]);
```
vibrantvarun, line 290: I am assuming this is the code which will add
<docId, MAGIC_NUMBER_START_STOP, score, shardID> entry?

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -215,6 +231,109 @@ private DocValueFormat[] getSortValueFormats(final SortAndFormats sortAndFormats
...
+        final ScoreDoc[] newScoreDocs,
+        final Comparator<ScoreDoc> scoreDocComparator
+    ) {
+        if (Objects.requireNonNull(sourceScoreDocs).length < MIN_NUMBER_OF_ELEMENTS_IN_SCORE_DOC
+            || Objects.requireNonNull(newScoreDocs).length < MIN_NUMBER_OF_ELEMENTS_IN_SCORE_DOC) {
+            throw new IllegalArgumentException("cannot merge top docs because it does not have enough elements");
+        }
+        // we overshoot and preallocate more than we need - length of both top docs combined.
+        // we will take only portion of the array at the end
+        List<ScoreDoc> mergedScoreDocs = new ArrayList<>(sourceScoreDocs.length + newScoreDocs.length);
```
vibrantvarun, line 288: Can we use a array here as they are faster? By this we can save toArray call at line 263.

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchResultFormatUtil.java
```diff
@@ -52,4 +52,28 @@ public static boolean isHybridQueryStartStopElement(final ScoreDoc scoreDoc) {
     public static boolean isHybridQueryDelimiterElement(final ScoreDoc scoreDoc) {
         return Objects.nonNull(scoreDoc) && scoreDoc.doc >= 0 && Float.compare(scoreDoc.score, MAGIC_NUMBER_DELIMITER) == 0;
     }
+
+    /**
+     * Checking if passed scoreDocs object is a special element (start/stop or delimiter) in the list of hybrid query result scores
+     * @param scoreDoc score doc object to check on
+     * @return true if it is a special element
+     */
+    public static boolean isHybridQuerySpecialElement(final ScoreDoc scoreDoc) {
```
vibrantvarun, line 61: nit: isHybridQueryResultDelimiter

src/test/java/org/opensearch/neuralsearch/query/HybridQueryIT.java
```diff
@@ -692,6 +698,101 @@ public void testWrappedQueryWithFilter_whenIndexAliasHasFilters_thenSuccess() {
...
+            wipeOfTestResources(TEST_INDEX_DOC_QTY_ONE_SHARD, null, null, SEARCH_PIPELINE);
+            updateClusterSettings(CONCURRENT_SEGMENT_SEARCH_ENABLED, false);
+        }
+    }
+
+    @SneakyThrows
+    public void testConcurrentSearchWithMultipleSlices_whenMultipleShardsIndex_thenSuccessful() {
+        try {
+            updateClusterSettings(CONCURRENT_SEGMENT_SEARCH_ENABLED, true);
+            int numberOfDocumentsInIndex = 2_000;
```
vibrantvarun, line 752: nit: 2000

src/test/java/org/opensearch/neuralsearch/query/HybridQueryIT.java
```diff
@@ -692,6 +698,101 @@ public void testWrappedQueryWithFilter_whenIndexAliasHasFilters_thenSuccess() {
         }
     }
 
+    @SneakyThrows
+    public void testConcurrentSearchWithMultipleSlices_whenSingleShardIndex_thenSuccessful() {
+        try {
+            updateClusterSettings(CONCURRENT_SEGMENT_SEARCH_ENABLED, true);
+            int numberOfDocumentsInIndex = 1_000;
```
vibrantvarun, line 705: nit: 1000

src/test/java/org/opensearch/neuralsearch/query/HybridQueryAggregationsIT.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: nope, that comment signifies the whole section where we do have something in query result and we may need to merge it with new object

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: ack, adding comment

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: yes, will add comment with these details

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: no, it cannot be done efficiently because we don't know length of resulting array because we do not know how many sub-queries are there. And if we overshoot and then copy only smaller part of the array then it's not better than using ArrayList. With ArrayList main problem is that it should be rebuild periodically when capacity is reached, that's why I passed upper bound of expected length in constructor

src/test/java/org/opensearch/neuralsearch/query/HybridQueryIT.java
martin-gaievski: To me syntax with delimiter is more readable

src/test/java/org/opensearch/neuralsearch/query/HybridQueryIT.java
martin-gaievski: same as above

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchResultFormatUtil.java
martin-gaievski: it's not only delimiter, but start/stop element too, `special element` has generic meaning for those two

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: Refactored code a bit

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -215,6 +219,139 @@ private DocValueFormat[] getSortValueFormats(final SortAndFormats sortAndFormats
...
+
+    /**
+     * Merge two score docs objects, result ScoreDocs[] object will have all hits per sub-query from both original objects.
+     * Logic is based on assumption that hits of every sub-query are sorted by score.
+     * Method returns new object and doesn't mutate original ScoreDocs arrays.
+     * @param sourceScoreDocs original score docs from query result
+     * @param newScoreDocs new score docs that we need to merge into existing scores
+     * @return merged array of ScoreDocs objects
+     */
+    private List<ScoreDoc> mergedScoreDocs(
```
vibrantvarun, line 293: Can we move all this code to a new class for ex: HybridQueryMergeTopDocs ?

As I for sorting I am planning to create the similar class where I will keep all the merge related code?

```
/*
 * Copyright OpenSearch Contributors
 * SPDX-License-Identifier: Apache-2.0
 */
package org.opensearch.neuralsearch.search;

import java.util.Comparator;
import org.apache.lucene.search.FieldComparator;
import org.apache.lucene.search.FieldDoc;
import org.apache.lucene.search.Pruning;
import org.apache.lucene.search.ScoreDoc;
import org.apache.lucene.search.SortField;

public class HybridQueryMergeTopFieldDocs {
    final SortField[] sortFields;
    final FieldComparator<?>[] comparators;
    final int[] reverseMul;
    final Comparator<ScoreDoc> tieBreaker;

    public HybridQueryMergeTopFieldDocs(SortField[] sortFields, Comparator<ScoreDoc> tieBreaker) {
        this.sortFields = sortFields;
        this.tieBreaker = tieBreaker;
        comparators = new FieldComparator[sortFields.length];
        reverseMul = new int[sortFields.length];
        for (int compIDX = 0; compIDX < sortFields.length; compIDX++) {
            final SortField sortField = sortFields[compIDX];
            comparators[compIDX] = sortField.getComparator(1, Pruning.NONE);
            reverseMul[compIDX] = sortField.getReverse() ? -1 : 1;
        }
    }

    public boolean lessThan(final FieldDoc firstFD, final FieldDoc secondFD) {
        for (int compIDX = 0; compIDX < comparators.length; compIDX++) {
            final FieldComparator comp = comparators[compIDX];

            final int cmp = reverseMul[compIDX] * comp.compareValues(firstFD.fields[compIDX], secondFD.fields[compIDX]);

            if (cmp != 0) {
                return cmp < 0;
            }
        }
        return tieBreakLessThan(firstFD, secondFD, tieBreaker);
    }

    private boolean tieBreakLessThan(ScoreDoc firstDoc, ScoreDoc secondDoc, Comparator<ScoreDoc> tieBreaker) {
        assert tieBreaker != null;
        int value = tieBreaker.compare(firstDoc, secondDoc);
        return value < 0;
    }
}
```

martin-gaievski: let's me think about proper LLD. Ideally we can do a strategy here and use implementation that is specific to the TopDocs type

src/main/java/org/opensearch/neuralsearch/search/query/ScoreDocsMerger.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.List;
+import java.util.Objects;
+
+import static org.opensearch.neuralsearch.search.util.HybridSearchResultFormatUtil.isHybridQueryScoreDocElement;
+
+/**
+ * Merges two ScoreDocs arrays into one
+ */
+@NoArgsConstructor
+public class ScoreDocsMerger {
```
vibrantvarun, line 22: nit: HybridQueryMergeOperations

src/main/java/org/opensearch/neuralsearch/search/query/ScoreDocsMerger.java
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * Merge two score docs objects, result ScoreDocs[] object will have all hits per sub-query from both original objects.
+     * Logic is based on assumption that hits of every sub-query are sorted by score.
+     * Method returns new object and doesn't mutate original ScoreDocs arrays.
+     * @param sourceScoreDocs original score docs from query result
+     * @param newScoreDocs new score docs that we need to merge into existing scores
+     * @return merged array of ScoreDocs objects
+     */
+    @VisibleForTesting
+    protected ScoreDoc[] mergedScoreDocs(
```
vibrantvarun, line 35: nit: mergeScoreDocs

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -215,6 +220,80 @@ private DocValueFormat[] getSortValueFormats(final SortAndFormats sortAndFormats
...
+        return (result) -> {
+            for (ReduceableSearchResult r : results) {
+                // call reduce for results of each single collector, this will update top docs in query result
+                r.reduce(result);
+            }
+        };
+    }
+
+    @VisibleForTesting
+    protected TopDocsAndMaxScore mergeTopDocsAndMaxScores(TopDocsAndMaxScore source, TopDocsAndMaxScore newTopDocs) {
```
vibrantvarun, line 260: Why access modifier of this method is protected?

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -215,6 +220,80 @@ private DocValueFormat[] getSortValueFormats(final SortAndFormats sortAndFormats
...
+        // format of results in both new and source TopDocs is following
+        // doc_id | magic_number_1
+        // doc_id | magic_number_2
+        // ...
+        // doc_id | magic_number_2
+        // ...
+        // doc_id | magic_number_2
+        // ...
+        // doc_id | magic_number_1
+        ScoreDoc[] mergedScoreDocs = scoreDocsMerger.mergedScoreDocs(
```
vibrantvarun, line 274: nit: name changes as suggested in ScoreDocsMerger class

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -215,6 +220,80 @@ private DocValueFormat[] getSortValueFormats(final SortAndFormats sortAndFormats
...
+        );
+        TotalHits mergedTotalHits = getMergedTotalHits(source, newTopDocs);
+        TopDocsAndMaxScore result = new TopDocsAndMaxScore(
+            new TopDocs(mergedTotalHits, mergedScoreDocs),
+            Math.max(source.maxScore, newTopDocs.maxScore)
+        );
+        return result;
+    }
+
+    private TotalHits getMergedTotalHits(TopDocsAndMaxScore source, TopDocsAndMaxScore newTopDocs) {
```
vibrantvarun, line 287: Can we move this method to HybridQueryMergeOperations class?

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: I pulled some code to a new Merger class. Not very match of abstracting is possible as ScoreDoc and SortField are not in the same class hierarchy, and case for general hybrid search is much simpler comparing to sorting. 
We can think and refactor more as a separate task some time later.

src/main/java/org/opensearch/neuralsearch/search/query/ScoreDocsMerger.java
martin-gaievski: I've pulled out merging logic for TopDocs and ScoreDoc into separate classes, this should address this comment and all underlying comment related to naming

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -46,12 +51,13 @@ public abstract class HybridCollectorManager implements CollectorManager<Collect
 
     private final int numHits;
     private final HitsThresholdChecker hitsThresholdChecker;
-    private final boolean isSingleShard;
     private final int trackTotalHitsUpTo;
     private final SortAndFormats sortAndFormats;
     @Nullable
     private final Weight filterWeight;
     private static final float boost_factor = 1f;
+    private final ScoreDocsMerger<ScoreDoc> scoreDocsMerger = new ScoreDocsMerger<>();
```
vibrantvarun, line 59: Can this initialization be moved to constructor and initialize only if concurrent segment search is enabled?

src/main/java/org/opensearch/neuralsearch/search/util/ScoreDocsMerger.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.List;
+import java.util.Objects;
+
+import static org.opensearch.neuralsearch.search.util.HybridSearchResultFormatUtil.isHybridQueryScoreDocElement;
+
+/**
+ * Merges two ScoreDoc arrays into one
+ */
+@NoArgsConstructor
+public class ScoreDocsMerger<T extends ScoreDoc> {
```
vibrantvarun, line 21: I think we should add name **HybridQuery** in the Name of the class as this operation are exclusive to the HybridQuery. like HybridQueryScoreDocsMerger

src/main/java/org/opensearch/neuralsearch/search/util/TopDocsMerger.java
```diff
@@ -0,0 +1,69 @@
...
+import org.opensearch.common.lucene.search.TopDocsAndMaxScore;
+
+import java.util.Comparator;
+import java.util.Objects;
+
+/**
+ * Utility class for merging TopDocs and MaxScore across multiple search queries
+ */
+@RequiredArgsConstructor
+public class TopDocsMerger {
```
vibrantvarun, line 21: nit: HybridQueryTopDocsMerger. This class is exclusive for Hybrid Query.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Features
 ### Enhancements
 ### Bug Fixes
+- Fixed merge logic for multiple collector result case ([#800](https://github.com/opensearch-project/neural-search/pull/800))
```
vibrantvarun, line 19: A better changelog would be "Fix for HybridQuery results are missing when concurrent segment search is enabled"

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -118,6 +121,27 @@ public Collector newCollector() {
      */
     @Override
     public ReduceableSearchResult reduce(Collection<Collector> collectors) {
+        final List<HybridTopScoreDocCollector> hybridTopScoreDocCollectors = getHybridScoreDocCollectors(collectors);
+        if (hybridTopScoreDocCollectors.isEmpty()) {
```
vibrantvarun, line 125: Can we move this validation into getHybridScoreDocCollectors method itself?

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: what benefit it will give? this code only creates java object, no heavy compute is there

src/main/java/org/opensearch/neuralsearch/search/util/ScoreDocsMerger.java
martin-gaievski: ack, we use format that is specific for hybrid query results

src/main/java/org/opensearch/neuralsearch/search/util/TopDocsMerger.java
martin-gaievski: it's actually generic, there isn't any special logic for hybrid query. Will leave name as it's now

CHANGELOG.md
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: I don't think it's a proper place for that validation. getter is responsible only for giving the collection of collectors, usage of that collection including validation is client responsibility

CHANGELOG.md
vibrantvarun: I cannot see the fix here.

src/main/java/org/opensearch/neuralsearch/search/util/TopDocsMerger.java
vibrantvarun: The merge method is exclusive for hybrid query.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
vibrantvarun: As per my understanding.
1. For normal hybrid search without concurrent segment search I do not these objects to be initialized.
2. Also it is not a good practice to initialize object globally. 

@navneet1v your thoughts?

vibrantvarun:
Overall looks good to me. Just waiting for @navneet1v review.

src/main/java/org/opensearch/neuralsearch/search/util/TopDocsMerger.java
martin-gaievski: Not really, it merges two TopDocsAndMaxScore object, only specific of hybrid query is in ScoreDocs merger.

CHANGELOG.md
martin-gaievski: corrected

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: 1. is not a benefit, initialization is cheap, let's not over engineer here
2.  not sure which practice you're referring to as "bad". One potential cons I can think of - for testing we would be able to inject mock merger for some test case, but we don't have such tests now and we can change it anytime if needed. For scope of this change there is only one mode of how we're going to use both merger classes, so it's ok to init at class property level. In case that changes in future we can change the code.

navneet1v:
> [Lucene code](https://github.com/apache/lucene/blob/branch_9_11/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java#L108-L110) where they define limit per slice (block that is processed by one collector). It's max of 250.000 docs or 5 segments.

@martin-gaievski Opensearch defines more control on top of this, to how to slice the segments. Ref: https://opensearch.org/docs/latest/search-plugins/concurrent-segment-search/#slicing-mechanisms

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -46,12 +49,13 @@ public abstract class HybridCollectorManager implements CollectorManager<Collect
...
     private final int numHits;
     private final HitsThresholdChecker hitsThresholdChecker;
-    private final boolean isSingleShard;
     private final int trackTotalHitsUpTo;
     private final SortAndFormats sortAndFormats;
     @Nullable
     private final Weight filterWeight;
     private static final float boost_factor = 1f;
+    private final HybridQueryScoreDocsMerger<ScoreDoc> scoreDocsMerger = new HybridQueryScoreDocsMerger<>();
+    private final TopDocsMerger topDocsMerger = new TopDocsMerger(scoreDocsMerger);
```
navneet1v, line 58: do we think we should init them via constructor? it can help us in unit testing by doing  dependency injection.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -118,6 +119,27 @@ public Collector newCollector() {
...
+            List<TopDocs> topDocs = hybridTopScoreDocCollector.topDocs();
+            TopDocs newTopDocs = getNewTopDocs(
+                getTotalHits(this.trackTotalHitsUpTo, topDocs, hybridTopScoreDocCollector.getTotalHits()),
+                topDocs
+            );
+            TopDocsAndMaxScore topDocsAndMaxScore = new TopDocsAndMaxScore(newTopDocs, hybridTopScoreDocCollector.getMaxScore());
+
+            results.add((QuerySearchResult result) -> reduceCollectorResults(result, topDocsAndMaxScore, docValueFormats, newTopDocs));
+        }
+        return reduceSearchResults(results);
```
navneet1v, line 139: So what I can see here is first for every collector results we are converting the to TopDocsAndMaxScore(the format what hybrid query needs) and then merging all of them.

Can we do this reverse, where we first merge the results and then we convert the results to TopDocsAndMaxScore. I feel in this way we can avoid a lot of duplicate iterations.

thoughts? @martin-gaievski , @vibrantvarun

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -215,6 +219,43 @@ private DocValueFormat[] getSortValueFormats(final SortAndFormats sortAndFormats
...
 
+    private void reduceCollectorResults(
+        QuerySearchResult result,
+        TopDocsAndMaxScore topDocsAndMaxScore,
+        DocValueFormat[] docValueFormats,
+        TopDocs newTopDocs
+    ) {
+        // this is case of first collector, query result object doesn't have any top docs set, so we can
+        // just set new top docs without merge
+        if (result.hasConsumedTopDocs()) {
```
navneet1v, line 230: what is this case can you please elaborate? I am confused here.

src/main/java/org/opensearch/neuralsearch/search/util/HybridQueryScoreDocsMerger.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.List;
+import java.util.Objects;
+
+import static org.opensearch.neuralsearch.search.util.HybridSearchResultFormatUtil.isHybridQueryScoreDocElement;
+
+/**
+ * Merges two ScoreDoc arrays into one
+ */
+@NoArgsConstructor
+public class HybridQueryScoreDocsMerger<T extends ScoreDoc> {
```
navneet1v, line 21: lets make this class package private, as the functions inside of this class should be limited to HybridQuery clause only

martin-gaievski:
> > [Lucene code](https://github.com/apache/lucene/blob/branch_9_11/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java#L108-L110) where they define limit per slice (block that is processed by one collector). It's max of 250.000 docs or 5 segments.
> 
> @martin-gaievski Opensearch defines more control on top of this, to how to slice the segments. Ref: https://opensearch.org/docs/latest/search-plugins/concurrent-segment-search/#slicing-mechanisms. are you taking any assumptions based on lucene slicing mechanism?

thanks for sharing that link, I was not aware of the OpenSearch specific approach. Anyhow, I'm not making any assumptions based on number of segments or any other condition of how the slices are constructed. Main principle will be same - it will be either one collector or many collectors with search results. Previously we were handling results from one collector, now we can handle scenario with multiple collectors.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: I had same idea initially, but unfortunately it's not feasible. Main reason is that every run of reduce method doesn't know about other reduces. We need to put collected results back to QueryResults object, so if we had such knowledge we can merge individual results and then update QueryResult at the end. But because such knowledge is missing and reduce is called b core (and differently depending on w/o concurrent segment search, and maybe also that different slicing mechanism that you've shared) then our reduce call should be stateless. 
I'm running benchmarks now to check if there is impact of performance

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: hasConsumedTopDocs() will return `true` if topDocs object is null in this QueryResult. Direct getter is weird - it will throw exception in case topDocs is null [code ref](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/query/QuerySearchResult.java#L159)

src/main/java/org/opensearch/neuralsearch/search/util/HybridQueryScoreDocsMerger.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: That's the question Varun has. I think it's fine for now to have it like this, don't see a use case when we want to init those classes differently. It can be changed in future if needed.

src/main/java/org/opensearch/neuralsearch/search/util/HybridQueryScoreDocsMerger.java
martin-gaievski: pushed commit, I have to move both merger classes to `query` package otherwise package private doesn't work.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: updated comment to make i more clear

martin-gaievski:
Ran a benchmark of https://github.com/opensearch-project/opensearch-benchmark-workloads/tree/main/noaa_semantic_search on main with and without this change, here are results for medium and large queries, where we have up to 8 segments and where the impact should be the visible the most:

With the change the performance is actually better. I'm taking one more round for baseline to make sure data isn't flaky, but these are preliminary results. 

baseline:
```
   {
    "task": "hybrid-query-only-range-medium-subset",
    "operation": "hybrid-query-only-range-medium-subset",
    "throughput": {
     "min": 2.011052370071411,
     "mean": 2.021946349143982,
     "median": 2.018242835998535,
     "max": 2.0519936084747314,
     "unit": "ops/s"
    },    
    "service_time": {
     "50_0": 91.03496170043945,
     "90_0": 96.44083023071289,
     "99_0": 98.4694938659668,
     "100_0": 98.7366943359375,
     "mean": 91.10263374328613,
     "unit": "ms"
    },
    "error_rate": 0.0
   },
-----
  {
    "task": "hybrid-query-only-range-large-subset",
    "operation": "hybrid-query-only-range-large-subset",
    "throughput": {
     "min": 2.0065014362335205,
     "mean": 2.01287588596344,
     "median": 2.010721802711487,
     "max": 2.0304625034332275,
     "unit": "ops/s"
    }
    "service_time": {
     "50_0": 188.2995376586914,
     "90_0": 204.8991470336914,
     "99_0": 218.81844329833984,
     "100_0": 222.81707763671875,
     "mean": 186.44740097045897,
     "unit": "ms"
    },
    "error_rate": 0.0
   },
```

with this change

```
   {
    "task": "hybrid-query-only-range",
    "operation": "hybrid-query-only-range",
    "throughput": {
     "min": 2.0113322734832764,
     "mean": 2.0225303030014037,
     "median": 2.0187013149261475,
     "max": 2.053476572036743,
     "unit": "ops/s"
    },
    "service_time": {
     "50_0": 72.53012466430664,
     "90_0": 73.0995979309082,
     "99_0": 89.77497482299805,
     "100_0": 105.39061737060547,
     "mean": 72.93198219299316,
     "unit": "ms"
    },
    "error_rate": 0.0
   },
---
  {
    "task": "hybrid-query-only-range-medium-subset",
    "operation": "hybrid-query-only-range-medium-subset",
    "throughput": {
     "min": 2.0104377269744873,
     "mean": 2.0207511043548583,
     "median": 2.0172306299209595,
     "max": 2.0493412017822266,
     "unit": "ops/s"
    },
    "service_time": {
     "50_0": 100.66146469116211,
     "90_0": 104.38032913208008,
     "99_0": 106.95694351196289,
     "100_0": 107.02909088134766,
     "mean": 100.12281967163086,
     "unit": "ms"
    },
    "error_rate": 0.0
   }
```

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
martin-gaievski: I've moved initialization to constructor and now topdocs merger is a constructor parameter, hope that's what you had in mind

src/main/java/org/opensearch/neuralsearch/search/util/TopDocsMerger.java
vibrantvarun: +1 to @martin-gaievski

vibrantvarun (approved):

vibrantvarun:
LGTM

navneet1v (approved):

martin-gaievski:
I re-ran the benchmark, it shows no change for large sub-set and 5% delta for medium sub-set. that looks more realistic as we only add some computation.

baseline:
```
{
    "task": "hybrid-query-only-range-medium-subset",
    "operation": "hybrid-query-only-range-medium-subset",
    "throughput": {
     "min": 2.0109918117523193,
     "mean": 2.0218198919296264,
     "median": 2.0181103944778442,
     "max": 2.051851272583008,
     "unit": "ops/s"
    },
    "service_time": {
     "50_0": 96.55980682373047,
     "90_0": 99.6461067199707,
     "99_0": 102.91871643066406,
     "100_0": 103.3681869506836,
     "mean": 95.76778579711915,
     "unit": "ms"
    },
    "duration": 62003.01305705216
   }
---
   {
    "task": "hybrid-query-only-range-large-subset",
    "operation": "hybrid-query-only-range-large-subset",
    "throughput": {
     "min": 2.0072591304779053,
     "mean": 2.014420256614685,
     "median": 2.0119868516921997,
     "max": 2.0340819358825684,
     "unit": "ops/s"
    },
    "service_time": {
     "50_0": 196.891845703125,
     "90_0": 209.67161560058594,
     "99_0": 224.50284576416016,
     "100_0": 227.4385986328125,
     "mean": 195.78350692749024,
     "unit": "ms"
    },
    "error_rate": 0.0
   }
```

after the change
```
{
    "task": "hybrid-query-only-range-medium-subset",
    "operation": "hybrid-query-only-range-medium-subset",
    "throughput": {
     "min": 2.0046942234039307,
     "mean": 2.0093229246139526,
     "median": 2.0077611207962036,
     "max": 2.0221428871154785,
     "unit": "ops/s"
    },
    "service_time": {
     "50_0": 88.03913879394531,
     "90_0": 93.1028938293457,
     "99_0": 104.01826095581055,
     "100_0": 106.14491271972656,
     "mean": 88.63995048522949,
     "unit": "ms"
    },
    "error_rate": 0.0
   },
---
   {
    "task": "hybrid-query-only-range-large-subset",
    "operation": "hybrid-query-only-range-large-subset",
    "throughput": {
     "min": 2.0080811977386475,
     "mean": 2.0161024808883665,
     "median": 2.0133992433547974,
     "max": 2.0380523204803467,
     "unit": "ops/s"
    },
    "service_time": {
     "50_0": 200.56720733642578,
     "90_0": 215.31163024902344,
     "99_0": 229.7359161376953,
     "100_0": 233.6544952392578,
     "mean": 196.9455467224121,
     "unit": "ms"
    },
    "error_rate": 0.0
   }
```

PR #802, opened 2024-06-21
[Part 2] Normalization Phase for Sorting

vibrantvarun:
### Description
This PR is part 2 of sorting feature which contains the changes required in Normalization Phase to enable Sorting.
**Key Highlights**
1. Make changes in CompoundTopDocs to cater TopFieldDocs
2. Cater changes in updateOrginalQueryResults method of NormalizationProcessorWorkflow to update Sorted results
3. While combining documents from multiple subqueries in a shard, make changes in ScoreCombiner to arrange them in the result as per sort criteria.

### Issues Resolved
[https://github.com/opensearch-project/neural-search/issues/507](https://github.com/opensearch-project/neural-search/issues/507)

### Check List
    - [ ] All tests pass


src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -24,32 +32,41 @@
 @Log4j2
 public class ScoreCombiner {
 
-    private static final Float ZERO_SCORE = 0.0f;
```
vibrantvarun, line 27: Removing this as it is not used anywhere.

src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
```diff
@@ -101,9 +101,13 @@ private static void validateMapTypeValue(
                         allowEmpty
                     );
                 } else if (!(nextSourceValue instanceof String)) {
-                    throw new IllegalArgumentException(String.format(Locale.ROOT, "map type field [%s] is neither string nor nested type, cannot process it", key));
+                    throw new IllegalArgumentException(
```
vibrantvarun, line 104: Just spotlessApply changes.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -59,32 +76,127 @@ private void combineShardScores(final ScoreCombinationTechnique scoreCombination
             scoreCombinationTechnique
         );
 
+        // If sorting is enabled then create the map of sort field values per docId
+        Map<Integer, Object[]> docIdSortFieldMap = null;
+        List<TopFieldDocs> topFieldDocs = null;
+        if (sort != null) {
+            final boolean isSortByScore = checkIfSortOrderByScore(sort);
+            topFieldDocs = new ArrayList<>();
+            for (int i = 0; i < topDocsPerSubQuery.size(); i++) {
```
martin-gaievski, line 85: looks like we can use `for` in iterative form, code will look a bit cleaner

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -59,32 +76,127 @@ private void combineShardScores(final ScoreCombinationTechnique scoreCombination
...
 
         // - update query search results with normalized scores
-        updateQueryTopDocsWithCombinedScores(compoundQueryTopDocs, topDocsPerSubQuery, combinedNormalizedScoresByDocId, sortedDocsIds);
+        updateQueryTopDocsWithCombinedScores(
+            compoundQueryTopDocs,
+            topDocsPerSubQuery,
+            combinedNormalizedScoresByDocId,
+            sortedDocsIds,
+            docIdSortFieldMap,
+            sort != null ? true : false
```
martin-gaievski, line 104:
```suggestion
            sort != null
```

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -59,32 +76,127 @@ private void combineShardScores(final ScoreCombinationTechnique scoreCombination
...
+        );
+    }
+
+    private boolean checkIfSortOrderByScore(Sort sort) {
+        if (sort != null) {
```
martin-gaievski, line 109: better fail fast - start with case if sort == null then return, then the rest of code doesn't need to be in inside the if condition

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchResultFormatUtil.java
```diff
@@ -16,6 +16,7 @@ public class HybridSearchResultFormatUtil {
     // and OpenSearch convention is that scores are positive numbers
     public static final Float MAGIC_NUMBER_START_STOP = -9549511920.4881596047f;
     public static final Float MAGIC_NUMBER_DELIMITER = -4422440593.9791198149f;
+    public static final Float MAX_SCORE_WHEN_NO_HITS_FOUND = 0.0f;
```
martin-gaievski, line 19: I would rather place this in score combiner class. This util is class is mainly responsible for the grammar type of formatting for the query, and this score is more related to the logic itself. See for all other methods we only use magic numbers, and real info like doc id passed as method params.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -136,4 +257,23 @@ private TotalHits getTotalHits(final List<TopDocs> topDocsPerSubQuery, final lon
...
+                return scoreComparison;
+            }
+
+            int docIdComparison = Integer.compare(o1.doc, o2.doc);
+            if (docIdComparison != 0) {
+                return docIdComparison;
+            }
+
+            // When duplicate result found then both score and doc ID are equal, return 1
+            return (o1.score == o2.score && o1.doc == o2.doc) ? 1 : 0;
```
martin-gaievski, line 275: I think in core that use shard id as a third params to run the sort in case doc id and score are same. Do we have shard id info here to do the same?

martin-gaievski (commented):
Looks like for unit tests we don't have anything new, sorting is not active. Can we add few tests that actually testing new code paths?

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
vibrantvarun: Actually No, at the normalization phase we do not have shard info. It comes in the fetch phase in multiple shard scenario.

To help better understand the situation let me put a example here.
Consider a result which is a part of subquery 1 and subquery 2 provided we have 2 subqueries under the hybrid clause.
The result before TopDocs.merge will look like
<docId, Magic number, -1, [1]>
<docId, Magic number, -1, [1]>
<docId, Magic number, -1, [1]>

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
```diff
@@ -88,17 +91,22 @@ public CompoundTopDocs(final TopDocs topDocs) {
             if (isHybridQueryDelimiterElement(scoreDoc) || isHybridQueryStartStopElement(scoreDoc)) {
                 ScoreDoc[] subQueryScores = scoreDocList.toArray(new ScoreDoc[0]);
                 TotalHits totalHits = new TotalHits(subQueryScores.length, TotalHits.Relation.EQUAL_TO);
-                TopDocs subQueryTopDocs = new TopDocs(totalHits, subQueryScores);
+                TopDocs subQueryTopDocs;
+                if (isSortEnabled) {
+                    subQueryTopDocs = new TopFieldDocs(totalHits, subQueryScores, ((TopFieldDocs) topDocs).fields);
```
VijayanB, line 96: please check for instance before type cast

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
```diff
@@ -113,7 +121,18 @@ private List<ScoreDoc> cloneLargestScoreDocs(final List<TopDocs> docs) {
...
+        List<ScoreDoc> scoreDocs = new ArrayList<>();
+        for (int scoreDocIndex = 0; scoreDocIndex < maxScoreDocs.length; scoreDocIndex++) {
+            if (isSortEnabled) {
+                FieldDoc fieldDoc = (FieldDoc) maxScoreDocs[scoreDocIndex];
+                scoreDocs.add(new FieldDoc(fieldDoc.doc, fieldDoc.score, fieldDoc.fields, fieldDoc.shardIndex));
+            } else {
+                ScoreDoc scoreDoc = maxScoreDocs[scoreDocIndex];
+                scoreDocs.add(new ScoreDoc(scoreDoc.doc, scoreDoc.score, scoreDoc.shardIndex));
+            }
+        }
```
VijayanB, line 135: nit: move logic to build FieldDoc/ScoreDoc to separate method and call appropriately if sortEnabled or not. The above code will check for isSortEnabled every time, though, it is fixed for all the score docs.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -110,13 +124,31 @@ private void updateOriginalQueryResults(final List<QuerySearchResult> querySearc
...
+            }
+            TopDocsAndMaxScore updatedTopDocsAndMaxScore;
+            if (sort == null) {
+                // create final version of top docs with all updated values
+                TopDocs topDocs = new TopDocs(updatedTopDocs.getTotalHits(), updatedTopDocs.getScoreDocs().toArray(new ScoreDoc[0]));
+                updatedTopDocsAndMaxScore = new TopDocsAndMaxScore(topDocs, maxScore);
+            } else {
+                final FieldDoc[] fieldDocs = new FieldDoc[updatedTopDocs.getScoreDocs().size()];
+                int i = 0;
+                if (updatedTopDocs.getTotalHits().value > 0) {
```
VijayanB, line 139: flip if condition to reduce nesting

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -110,13 +124,31 @@ private void updateOriginalQueryResults(final List<QuerySearchResult> querySearc
...
+                    for (ScoreDoc scoreDoc : updatedTopDocs.getScoreDocs()) {
+                        maxScore = Math.max(maxScore, scoreDoc.score);
+                        fieldDocs[i++] = (FieldDoc) scoreDoc;
+                    }
+                } else {
+                    maxScore = MAX_SCORE_WHEN_NO_HITS_FOUND;
```
VijayanB, line 145: can you add comment on why we have to update maxscore if sort != null and total hits > 0 ?

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -206,4 +238,77 @@ private List<Integer> unprocessedDocIds(final List<QuerySearchResult> querySearc
...
+    /**
+     * This method processes a list of query top documents and retrieved the top documents
+     * that match certain criteria. If a valid top document is found, it creates and returns a sort object based on the first FieldDoc in the scoreDocs array of the top document.
+     * Otherwise, it returns null.
+     * @param queryTopDocs List of query top documents to be evaluated.
+     * @return Sort object based on the first FieldDoc in the scoreDocs array, or null if no valid document is found.
+     */
+    private Sort evaluateSortCriteria(List<CompoundTopDocs> queryTopDocs) {
+        for (CompoundTopDocs compoundTopDocs : queryTopDocs) {
+            if (compoundTopDocs != null && compoundTopDocs.getTotalHits().value > 0) {
```
VijayanB, line 251: flip if condition to return null

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -206,4 +238,77 @@ private List<Integer> unprocessedDocIds(final List<QuerySearchResult> querySearc
...
+
+                if (foundTopDoc != null && foundTopDoc.scoreDocs[0] instanceof FieldDoc) {
+                    return createSort(topDocs.toArray(new TopFieldDocs[0]));
+                } else {
+                    return null;
+                }
+            }
+        }
+        return null;
+    }
```
VijayanB, line 269: hard to follow the logic because of too many nested, can you refactor to limit the nesting.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
martin-gaievski: got it, then we should be good with what you do have in PR now

src/test/java/org/opensearch/neuralsearch/processor/NormalizationProcessorTests.java
```diff
@@ -172,7 +171,7 @@ public void testSearchResultTypes_whenCompoundDocs_thenDoNormalizationCombinatio
                     createStartStopElementForHybridSearchResults(4) }
 
             );
-            querySearchResult.topDocs(new TopDocsAndMaxScore(topDocs, 0.5f), new DocValueFormat[0]);
+            querySearchResult.topDocs(new TopDocsAndMaxScore(topDocs, 0.5f), null);
```
vibrantvarun, line 174: DocValueFormat should be null when Sorting is applied.

src/test/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflowTests.java
```diff
@@ -108,7 +107,7 @@ public void testSearchResultTypes_whenNoMatches_thenReturnZeroResults() {
                     ),
                     0.0f
                 ),
-                new DocValueFormat[0]
```
vibrantvarun, line 111: DocValueFormat should be null when sorting is not applied.

src/main/java/org/opensearch/neuralsearch/processor/combination/dto/CombineScoresDTO.java
```diff
@@ -0,0 +1,34 @@
...
+import org.opensearch.common.Nullable;
+import org.opensearch.neuralsearch.processor.CompoundTopDocs;
+import org.opensearch.neuralsearch.processor.combination.ScoreCombinationTechnique;
+import org.opensearch.search.query.QuerySearchResult;
+
+/**
+ * CombineScoresDTO handles the request sent for score combination after normalization.
+ */
+@AllArgsConstructor
+public class CombineScoresDTO {
```
martin-gaievski, line 21: Please make it in a way the clients are "highly motivated" to use it in a builder fashion. Check out how it's done in say knn repo got [BaseQueryFactory](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/BaseQueryFactory.java).
This is how [client use it](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNQueryFactory.java#L48-L56)

src/test/java/org/opensearch/neuralsearch/processor/ScoreCombinationTechniqueTests.java
```diff
@@ -19,7 +21,7 @@ public class ScoreCombinationTechniqueTests extends OpenSearchTestCase {
 
     public void testEmptyResults_whenEmptyResultsAndDefaultMethod_thenNoProcessing() {
         ScoreCombiner scoreCombiner = new ScoreCombiner();
-        scoreCombiner.combineScores(List.of(), ScoreCombinationFactory.DEFAULT_METHOD);
+        scoreCombiner.combineScores(new CombineScoresDTO(List.of(), ScoreCombinationFactory.DEFAULT_METHOD, Collections.emptyList(), null));
```
martin-gaievski, line 24: Please use build style, check my other comment in DTO class. With the init through constructor it's not better than having all arguments passed directly, we just slightly moved burden from one place to another

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -59,36 +79,145 @@ private void combineShardScores(final ScoreCombinationTechnique scoreCombination
             scoreCombinationTechnique
         );
 
+        // If sorting is enabled then create the map of sort field values per docId
+        Map<Integer, Object[]> docIdSortFieldMap = null;
```
martin-gaievski, line 83: can we have a private method that just return the map

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -59,36 +79,145 @@ private void combineShardScores(final ScoreCombinationTechnique scoreCombination
...
+        } else {
+            if (Objects.isNull(topFieldDocs)) {
+                throw new IllegalArgumentException("topFieldDocs cannot be null when sorting is enabled.");
+            }
+            int topN = 0;
+            for (TopFieldDocs topFieldDoc : topFieldDocs) {
+                topN += topFieldDoc.scoreDocs.length;
+            }
+
+            // Merge the sorted results of individual queries to form a one final result per shard which is sorted.
```
martin-gaievski, line 172: can you add example of such result by shard for query with multiple sub-queries

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -59,36 +79,145 @@ private void combineShardScores(final ScoreCombinationTechnique scoreCombination
...
+            final TopDocs sortedTopDocs = TopDocs.merge(sort, 0, topN, topFieldDocs.toArray(new TopFieldDocs[0]), getTieBreaker());
+
+            // Remove duplicates from the sorted top docs.
+            Set<Integer> uniqueDocIds = new LinkedHashSet<>();
+            for (ScoreDoc scoreDoc : sortedTopDocs.scoreDocs) {
+                uniqueDocIds.add(scoreDoc.doc);
+            }
+            sortedDocsIds = new ArrayList<>(uniqueDocIds);
```
martin-gaievski, line 180: At the first glance it looks like lot of tossing of same info between different collections: List of topDocs -> array -> sorted topDocs -> set of doc ids -> list of unique ids
it may be a hot spot as we're on the search path. Can we eliminate some iterations? Why we do have duplicates in score docs?

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -136,4 +274,23 @@ private TotalHits getTotalHits(final List<TopDocs> topDocsPerSubQuery, final lon
         }
         return new TotalHits(maxHits, totalHits);
     }
+
+    // Tie-breaker to merge multiple top docs
+    private Comparator<ScoreDoc> getTieBreaker() {
+        final Comparator<ScoreDoc> SORTING_TIE_BREAKER = (o1, o2) -> {
```
martin-gaievski, line 280: can we make this constant at the class level?

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchSortUtil.java
```diff
@@ -0,0 +1,116 @@
...
+     * @return true if sort widening is required, false otherwise.
+     */
+    private static boolean isSortWideningRequired(TopFieldDocs[] topFieldDocs, int sortFieldindex) {
+        for (int i = 0; i < topFieldDocs.length - 1; i++) {
+            TopFieldDocs currentTopFieldDoc = topFieldDocs[i];
+            TopFieldDocs nextTopFieldDoc = topFieldDocs[i + 1];
+            if (currentTopFieldDoc == null || nextTopFieldDoc == null) {
+                throw new IllegalArgumentException("topFieldDocs cannot be null when sorting is applied");
+            }
+            if (!currentTopFieldDoc.fields[sortFieldindex].equals(nextTopFieldDoc.fields[sortFieldindex])) {
```
martin-gaievski, line 110: Is it check reliable? I check for TopDocs it's not overridden, so we're using default equals which essentially checking for the same ref in memory. Should we check based on the object fields instead?

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchSortUtil.java
```diff
@@ -0,0 +1,116 @@
...
+     * This method iterates through the provided topFieldDocs array and checks if any adjacent pairs of sort fields at the specified index are not equal.
+     * If any such pair is found, it returns true, indicating that sort widening is required. Otherwise, it returns false.
+     * @param topFieldDocs array of top field documents to be checked.
+     * @param sortFieldindex index of the sort field to be checked within each top field document.
+     * @return true if sort widening is required, false otherwise.
+     */
+    private static boolean isSortWideningRequired(TopFieldDocs[] topFieldDocs, int sortFieldindex) {
+        for (int i = 0; i < topFieldDocs.length - 1; i++) {
+            TopFieldDocs currentTopFieldDoc = topFieldDocs[i];
+            TopFieldDocs nextTopFieldDoc = topFieldDocs[i + 1];
```
martin-gaievski, line 106: if `topFieldDocs` array is always same array does it mean we're repeating same compute multiple times? Can be save on CPU cycles, maybe compute once

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchSortUtil.java
```diff
@@ -0,0 +1,116 @@
...
+import org.opensearch.neuralsearch.processor.CompoundTopDocs;
+import org.opensearch.search.query.QuerySearchResult;
+import org.opensearch.search.sort.SortedWiderNumericSortField;
+
+/**
+ * Utility class for evaluating and creating sort criteria
+ */
+public class HybridSearchSortUtil {
+
+    public static Sort evaluateSortCriteria(final List<QuerySearchResult> querySearchResults, final List<CompoundTopDocs> queryTopDocs) {
```
martin-gaievski, line 22: add comment, it's public method

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchSortUtil.java
vibrantvarun: No, we need to check for every field because
```
Creates Sort object from topFieldsDocs fields. It is necessary to widen the SortField.Type to maximum byte size for merging sorted docs. Different indices might have different types. This will avoid user to do re-index of data in case of mapping field change for newly indexed data. This will support Int to Long and Float to Double. Earlier widening of type was taken care in IndexNumericFieldData, but since we now want to support sort optimization, we removed type widening there and taking care here during merging.
```

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchSortUtil.java
vibrantvarun: Yes this check is reliable. The same is done in OpenSearch core. [https://github.com/opensearch-project/OpenSearch/blob/8ad199dac020c91c80180b4f8042f65c9994a81a/server/src/main/java/org/opensearch/action/search/SearchPhaseController.java#L630](https://github.com/opensearch-project/OpenSearch/blob/8ad199dac020c91c80180b4f8042f65c9994a81a/server/src/main/java/org/opensearch/action/search/SearchPhaseController.java#L630)

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
vibrantvarun: 1. List of TopFieldDocs is input of the method.
2. Method **TopDocs.merge**
   a. Input: Array of TopFieldDocs
   b. Output: Get list of sorted score docs which contains duplicates basically results from multiple queries.
3. Convert sortedTopDocs.scoreDocs to list and then pass it to set. 

As the list is sorted so all the top entries will be added first in the set.

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
```diff
@@ -88,17 +91,22 @@ public CompoundTopDocs(final TopDocs topDocs) {
             if (isHybridQueryDelimiterElement(scoreDoc) || isHybridQueryStartStopElement(scoreDoc)) {
                 ScoreDoc[] subQueryScores = scoreDocList.toArray(new ScoreDoc[0]);
                 TotalHits totalHits = new TotalHits(subQueryScores.length, TotalHits.Relation.EQUAL_TO);
-                TopDocs subQueryTopDocs = new TopDocs(totalHits, subQueryScores);
+                TopDocs subQueryTopDocs;
+                if (isSortEnabled && topDocs instanceof TopFieldDocs) {
```
martin-gaievski, line 95: are we handling properly condition when `isSortEnabled == false` and  `topDocs instanceof TopFieldDocs`. Not sure if that's possible but if it is we probably should fail with runtime exception

src/main/java/org/opensearch/neuralsearch/processor/combination/BaseScoreCombinationFactory.java
```diff
@@ -0,0 +1,41 @@
...
+import org.apache.lucene.search.Sort;
+import org.opensearch.common.Nullable;
+import org.opensearch.neuralsearch.processor.CompoundTopDocs;
+import org.opensearch.search.query.QuerySearchResult;
+
+/**
+ * Base class for creating score combination DTO.
+ */
+@Log4j2
+public abstract class BaseScoreCombinationFactory {
```
martin-gaievski, line 22: Why do we need this class? We can just have CombineScoresDTO as a class, without wrapping abstract class BaseScoreCombinationFactory

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -61,34 +97,177 @@ private void combineShardScores(final ScoreCombinationTechnique scoreCombination
...
+        // ShardId will not be -1 in when index has single shard because Fetch phase gets executed before Normalization
+        if (!compoundQueryTopDocs.getScoreDocs().isEmpty()) {
+            shardId = compoundQueryTopDocs.getScoreDocs().get(0).shardIndex;
+        }
         List<ScoreDoc> scoreDocs = new ArrayList<>();
-        int shardId = compoundQueryTopDocs.getScoreDocs().get(0).shardIndex;
-        for (int j = 0; j < maxHits && j < sortedScores.size(); j++) {
-            int docId = sortedScores.get(j);
-            scoreDocs.add(new ScoreDoc(docId, combinedNormalizedScoresByDocId.get(docId), shardId));
+        int j = 0;
```
martin-gaievski, line 246:
```suggestion
        int hitCount = 0;
```

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
vibrantvarun: If sortEnabled = false then topdocs cannot be the instance of TopFieldDocs. Because creation of top field docs depends upon sortfields. No sort fields then it will be a sure shot top docs object and not an instance of top field docs. Creation of top docs is being handled by query phase.

src/main/java/org/opensearch/neuralsearch/processor/combination/CombineScoresDTO.java
```diff
@@ -0,0 +1,34 @@
...
+import org.opensearch.neuralsearch.processor.CompoundTopDocs;
+import org.opensearch.search.query.QuerySearchResult;
+
+/**
+ * DTO object to hold data required for Score Combination.
+ */
+@AllArgsConstructor
+@Builder
+@Getter
+@Log4j2
```
martin-gaievski, line 24: nit: we probably don't need this

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
martin-gaievski: Then why are we doing "instance of"? It should be either just check for the `sortEnabled` or you need to handle both case when input object is and is not instance of. Currently in case it's not a TopFieldDocs the flow goes to TopDocs case, if that can handle such scenario (as it's a base class I think) I'm fine with it, otherwise we need to fail fast with Illegal Argument or something similar

src/test/java/org/opensearch/neuralsearch/query/HybridQuerySortIT.java
```diff
@@ -0,0 +1,705 @@
...
+    public static void setUpCluster() {
+        // we need new instance because we're calling non-static methods from static method.
+        // main purpose is to minimize network calls, initialization is only needed once
+        HybridQuerySortIT instance = new HybridQuerySortIT();
+        instance.initClient();
+        instance.updateClusterSettings();
+    }
+
+    @Override
+    protected boolean preserveClusterUponCompletion() {
```
martin-gaievski, line 70: if that is intentional why do we need to preserve cluster state for this test?

src/test/java/org/opensearch/neuralsearch/query/HybridQuerySortIT.java
```diff
@@ -0,0 +1,705 @@
...
+    // prepareResourcesBeforeTestExecution(SHARDS_COUNT_IN_SINGLE_NODE_CLUSTER);
+    // testSingleFieldSort_whenMultipleSubQueries_thenSuccessful(TEST_MULTI_DOC_INDEX_WITH_TEXT_AND_INT_SINGLE_SHARD);
+    // testMultipleFieldSort_whenMultipleSubQueries_thenSuccessful(TEST_MULTI_DOC_INDEX_WITH_TEXT_AND_INT_SINGLE_SHARD);
+    // } finally {
+    // wipeOfTestResources(TEST_MULTI_DOC_INDEX_WITH_TEXT_AND_INT_SINGLE_SHARD, null, null, SEARCH_PIPELINE);
+    // }
+    // }
+
+    @SneakyThrows
+    public void testSortOnSingleShard_WhenConcurrentSearchDisabled_thenSuccessful() {
```
martin-gaievski, line 88:
```suggestion
    public void testSortOnSingleShard_whenConcurrentSearchDisabled_thenSuccessful() {
```

please fix the case in other method names too

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
vibrantvarun: I have removed instance of check. I added after vijay's review. But it is not needed.

martin-gaievski (approved):

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -107,16 +136,43 @@ private void updateOriginalQueryResults(final List<QuerySearchResult> querySearc
...
+            return MAX_SCORE_WHEN_NO_HITS_FOUND;
+        }
+        if (isSortEnabled) {
+            float maxScore = MAX_SCORE_WHEN_NO_HITS_FOUND;
+            // In case of sorting iterate over score docs and deduce the max score
+            for (ScoreDoc scoreDoc : updatedTopDocs.getScoreDocs()) {
+                maxScore = Math.max(maxScore, scoreDoc.score);
+            }
+            return maxScore;
+        } else {
```
VijayanB, line 159: nit: You can remove the else

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -107,16 +136,43 @@ private void updateOriginalQueryResults(final List<QuerySearchResult> querySearc
...
+    /**
+     * Get Top Docs on Shard
+     * @param updatedTopDocs compound top docs on a shard
+     * @param sort  sort criteria
+     * @return TopDocs which will be instance of TopFieldDocs  if sort is enabled.
+     */
+    private TopDocs getTopDocsOnShard(CompoundTopDocs updatedTopDocs, Sort sort) {
+        if (sort != null) {
+            return new TopFieldDocs(updatedTopDocs.getTotalHits(), updatedTopDocs.getScoreDocs().toArray(new FieldDoc[0]), sort.getSort());
+        } else {
```
VijayanB, line 174: nit: same as above

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -107,16 +136,43 @@ private void updateOriginalQueryResults(final List<QuerySearchResult> querySearc
...
 
-            TopDocsAndMaxScore updatedTopDocsAndMaxScore = new TopDocsAndMaxScore(topDocs, maxScore);
-            querySearchResult.topDocs(updatedTopDocsAndMaxScore, null);
+    /**
+     * Get Top Docs on Shard
+     * @param updatedTopDocs compound top docs on a shard
+     * @param sort  sort criteria
+     * @return TopDocs which will be instance of TopFieldDocs  if sort is enabled.
+     */
+    private TopDocs getTopDocsOnShard(CompoundTopDocs updatedTopDocs, Sort sort) {
```
VijayanB, line 171: nit: buildTopDocs

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -23,33 +32,59 @@
...
+        }
 
-    private static final Float ZERO_SCORE = 0.0f;
+        int docIdComparison = Integer.compare(o1.doc, o2.doc);
+        if (docIdComparison != 0) {
+            return docIdComparison;
+        }
+
+        // When duplicate result found then both score and doc ID are equal, return 1
+        return (o1.score == o2.score && o1.doc == o2.doc) ? 1 : 0;
```
VijayanB, line 49: isn't this implicit? why not just return 1?

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -61,34 +96,177 @@ private void combineShardScores(final ScoreCombinationTechnique scoreCombination
...
+            topDocsPerSubQuery,
+            combinedNormalizedScoresByDocId,
+            sortedDocsIds,
+            getDocIdSortFieldsMap(compoundQueryTopDocs, combinedNormalizedScoresByDocId, sort),
+            sort != null
+        );
     }
 
-    private List<Integer> getSortedDocIds(final Map<Integer, Float> combinedNormalizedScoresByDocId) {
+    private boolean checkIfSortOrderByScore(Sort sort) {
```
VijayanB, line 116: nit: isOrderByScore

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -61,34 +96,177 @@ private void combineShardScores(final ScoreCombinationTechnique scoreCombination
...
+        if (sort == null) {
+            return false;
+        }
+
+        for (SortField sortField : sort.getSort()) {
+            if (sortField.getType().equals(SortField.Type.SCORE)) {
```
VijayanB, line 122: can you flip comparison like SortField.Type.SCORE.equals(sortField.getType())

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -61,34 +96,177 @@ private void combineShardScores(final ScoreCombinationTechnique scoreCombination
...
+     */
+    private List<TopFieldDocs> getTopFieldDocs(final Sort sort, final List<TopDocs> topDocsPerSubQuery) {
+        if (sort == null) {
+            return null;
+        }
+        List<TopFieldDocs> topFieldDocs = new ArrayList<>();
+        for (TopDocs topDocs : topDocsPerSubQuery) {
+            // Check for scoreDocs length.
+            // If scoreDocs length=0 then it means that no results are found for that particular subquery.
+            if (topDocs.scoreDocs.length != 0) {
```
VijayanB, line 144: can topDocs.scoreDocs be null?

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -61,34 +96,177 @@ private void combineShardScores(final ScoreCombinationTechnique scoreCombination
...
+        // If sort is null then no sort fields present therefore return null.
+        if (sort == null) {
+            return null;
+        }
         // we're merging docs with normalized and combined scores. we need to have only maxHits results
-        List<Integer> sortedDocsIds = new ArrayList<>(combinedNormalizedScoresByDocId.keySet());
-        sortedDocsIds.sort((a, b) -> Float.compare(combinedNormalizedScoresByDocId.get(b), combinedNormalizedScoresByDocId.get(a)));
-        return sortedDocsIds;
+        Map<Integer, Object[]> docIdSortFieldMap = new HashMap<>();
+        List<TopDocs> topFieldDocs = compoundTopDocs.getTopDocs();
```
VijayanB, line 169: nit: final

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -61,34 +96,177 @@ private void combineShardScores(final ScoreCombinationTechnique scoreCombination
...
+        return docIdSortFieldMap;
+    }
+
+    private Collection<Integer> getSortedDocIds(
+        final Map<Integer, Float> combinedNormalizedScoresByDocId,
+        final List<TopFieldDocs> topFieldDocs,
+        final Sort sort
+    ) {
+        // we're merging docs with normalized and combined scores. we need to have only maxHits results
+        if (Objects.isNull(sort)) {
```
VijayanB, line 194: nit: some place i see == null check is added and here we are using Objects api. can you make this as consistent? i don't have any preference, just want to see we are using one format to perform null check.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -61,34 +96,177 @@ private void combineShardScores(final ScoreCombinationTechnique scoreCombination
...
+            List<Integer> sortedDocsIds = new ArrayList<>(combinedNormalizedScoresByDocId.keySet());
+            sortedDocsIds.sort((a, b) -> Float.compare(combinedNormalizedScoresByDocId.get(b), combinedNormalizedScoresByDocId.get(a)));
+            return sortedDocsIds;
+        } else {
+            if (Objects.isNull(topFieldDocs)) {
+                throw new IllegalArgumentException("topFieldDocs cannot be null when sorting is enabled.");
+            }
+            int topN = 0;
```
VijayanB, line 202: what does topN means?

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -61,34 +96,177 @@ private void combineShardScores(final ScoreCombinationTechnique scoreCombination
...
+        final Map<Integer, Float> combinedNormalizedScoresByDocId,
+        final List<TopFieldDocs> topFieldDocs,
+        final Sort sort
+    ) {
+        // we're merging docs with normalized and combined scores. we need to have only maxHits results
+        if (Objects.isNull(sort)) {
+            List<Integer> sortedDocsIds = new ArrayList<>(combinedNormalizedScoresByDocId.keySet());
+            sortedDocsIds.sort((a, b) -> Float.compare(combinedNormalizedScoresByDocId.get(b), combinedNormalizedScoresByDocId.get(a)));
+            return sortedDocsIds;
+        } else {
```
VijayanB, line 198: nit: you can remove else

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -61,34 +96,177 @@ private void combineShardScores(final ScoreCombinationTechnique scoreCombination
...
+    private ScoreDoc getScoreDoc(
+        final boolean isSortEnabled,
+        final int docId,
+        final int shardId,
+        final Map<Integer, Float> combinedNormalizedScoresByDocId,
+        final Map<Integer, Object[]> docIdSortFieldMap
+    ) {
+        if (isSortEnabled && docIdSortFieldMap != null) {
+            return new FieldDoc(docId, combinedNormalizedScoresByDocId.get(docId), docIdSortFieldMap.get(docId), shardId);
+        } else {
```
VijayanB, line 265: same as above

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -61,34 +96,177 @@ private void combineShardScores(final ScoreCombinationTechnique scoreCombination
...
+            // < 1, 0.7, shardId, [70]>
+            // < 1, 0.3, shardId, [70]>
+            final TopDocs sortedTopDocs = TopDocs.merge(sort, 0, topN, topFieldDocs.toArray(new TopFieldDocs[0]), SORTING_TIE_BREAKER);
+
+            // Remove duplicates from the sorted top docs.
+            Set<Integer> uniqueDocIds = new LinkedHashSet<>();
+            for (ScoreDoc scoreDoc : sortedTopDocs.scoreDocs) {
+                uniqueDocIds.add(scoreDoc.doc);
+            }
+            return uniqueDocIds;
```
VijayanB, line 226: why not convert to List<Integer> to keep the interface return type less generic than Collection<Integer> ?

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchSortUtil.java
```diff
@@ -0,0 +1,119 @@
...
+        if (!checkIfSortEnabled(querySearchResults)) {
+            return null;
+        }
+        return createSort(getTopFieldDocs(queryTopDocs));
+    }
+
+    // Check if sort is enabled by checking docValueFormats Object
+    private static boolean checkIfSortEnabled(final List<QuerySearchResult> querySearchResults) {
+        if (querySearchResults == null || querySearchResults.isEmpty() || querySearchResults.get(0) == null) {
+            throw new IllegalArgumentException("shard result cannot be null in the normalization process.");
```
VijayanB, line 37: nit: search results

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchSortUtil.java
```diff
@@ -0,0 +1,119 @@
...
+                throw new IllegalArgumentException("CompoundTopDocs cannot be null in the normalization process");
+            }
+            if (checkIfTopFieldDocExists(compoundTopDocs.getTopDocs())) {
+                return compoundTopDocs.getTopDocs().toArray(new TopFieldDocs[0]);
+            }
+        }
+        return new TopFieldDocs[0];
+    }
+
+    private static boolean checkIfTopFieldDocExists(List<TopDocs> topDocs) {
```
VijayanB, line 57: nit: containsTopFieldDocs

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchSortUtil.java
```diff
@@ -0,0 +1,119 @@
...
+     * It is necessary to widen the SortField.Type to maximum byte size for merging sorted docs.
+     * Different indices might have different types. This will avoid user to do re-index of data
+     * in case of mapping field change for newly indexed data.
+     * This will support Int to Long and Float to Double.
+     * Earlier widening of type was taken care in IndexNumericFieldData, but since we now want to
+     * support sort optimization, we removed type widening there and taking care here during merging.
+     * More details here https://github.com/opensearch-project/OpenSearch/issues/6326
+     */
+    private static Sort createSort(TopFieldDocs[] topFieldDocs) {
+        if (topFieldDocs == null || topFieldDocs[0] == null) {
```
VijayanB, line 81: why are we only checking first value?

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchSortUtil.java
```diff
@@ -0,0 +1,119 @@
...
+            throw new IllegalArgumentException("topFieldDocs cannot be null when sorting is applied");
+        }
+        final SortField[] firstTopDocFields = topFieldDocs[0].fields;
+        final SortField[] newFields = new SortField[firstTopDocFields.length];
+
+        for (int i = 0; i < firstTopDocFields.length; i++) {
+            final SortField delegate = firstTopDocFields[i];
+            final SortField.Type type = delegate instanceof SortedNumericSortField
```
VijayanB, line 89: nit: sortFieldType

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
vibrantvarun: I would say lets keep this else because it gives more readability for someone who is new and trying to understand the code. It helps to explain clearly what happens when sorting is not applied. Removing else, would look aesthetically better, code readability would decrease. Considering the complexity of this whole feature, I would suggest to give priority to code readability.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
vibrantvarun: No, lets not directly return 1, someone who is trying to understand the code will not understand in which condition it returned 1 while debugging. The thing is here this comparator would be having duplicate result because of the sort fields values. Therefore if two documents from different shards with same docId and same sort field values then score will be same for them therefore just return 1 to act as tie breaker.

Also here I have explained this with example https://github.com/opensearch-project/neural-search/pull/802#discussion_r1655294422

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
vibrantvarun: nit: isSortOrderByScore

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
vibrantvarun: Nope. https://github.com/opensearch-project/neural-search/blob/25d2e82b62b52eae4d73f8f33aab704ff01c0390/src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java#L78

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
vibrantvarun: It is a parameter in TopDocs.merge method which is performed few lines below. The merge method as an argument named topN. It signifies how many documents to merge from the TopDocs on the basis of sort criteria. The reason why I kept is as topN so that no one gets confused while debugging the code.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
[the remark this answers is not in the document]
vibrantvarun: yes

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
vibrantvarun: I did the same before but @martin-gaievski asked me to go for generic like Collection. The reason behind that is Converting to list would be extra time consuming operation and as search is hotspot we kept Collection over here to let not impact performance. For more details check this comment out https://github.com/opensearch-project/neural-search/pull/802#discussion_r1657658609

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchSortUtil.java
vibrantvarun: Because, this check is not even needed because this condition would never happen. I have kept it here because later when I will add the additional unit tests then it doesn't fail.

Issue #110, opened 2023-01-27
[FEATURE] Nested field mapping support for text embedding processor


navneet1v:
### Is your feature request related to a problem?
This is related to customer created Github issue: https://github.com/opensearch-project/neural-search/issues/109

The following configuration using a nested source field, embeddings are not computed, which should be supported:

```
PUT /_ingest/pipeline/neural_pipeline_nested
{
  "description": "Neural Search Pipeline for message content",
  "processors": [
    {
      "text_embedding": {
        "model_id": "SXXx8YUBR2ZWhVQIkghB",
        "field_map": {
          "message.text": "message_embedding"
        }
      }
    }
  ]
}

PUT /neural-test-index-nested
{
    "settings": {
        "index.knn": true,
        "default_pipeline": "neural_pipeline_nested"
    },
    "mappings": {
        "properties": {
            "message_embedding": {
                "type": "knn_vector",
                "dimension": 384,
                "method": {
                    "name": "hnsw",
                    "engine": "lucene"
                }
            },
            "message.text": { 
                "type": "text"            
            },
            "color": {
                "type": "text"
            }
        }
    }
}

POST /_bulk
{"create":{"_index":"neural-test-index-nested","_id":"0"}}
{"message":{"text":"Text 1"},"color":"red"}
{"create":{"_index":"neural-test-index-nested","_id":"1"}}
{"message":{"text":"Text 2"}, "color": "black"}

GET /neural-test-index-nested/_search
```

### What solution would you like?
The fields map keys should support . operator to define the nested fields.

### What alternatives have you considered?
Customer can create a nested field mapping using:

```
PUT /neural-test-index-nested
{
    "description": "Neural Search Pipeline for message content",
    "processors": [
        {
            "text_embedding": {
                "model_id": "SXXx8YUBR2ZWhVQIkghB",
                "field_map": {
                    "message": {
                        "text": "message_embedding"
                    }
                }
            }
        }
    ]
}
```

asfoorial:
Is this going to also handle inner documents "nested" field types

Sanjana679:
I'm going to tackle this issue!

samuel-oci:
@navneet1v what is the expected behavior in case of nested field type as opposed to the above object field example?
How will the flattening to an array be handled? 
For context
Nested field: https://opensearch.org/docs/latest/field-types/supported-field-types/nested/
object field: https://opensearch.org/docs/latest/field-types/supported-field-types/object/

[...]

PR #811, opened 2024-07-03
Enable '.' for nested field in text embedding processor

martin-gaievski:
### Description
Adding support for complex structures in inference processor definition. Main purpose is to improve user experience for cases when object has complex hierarchical structure, so users can apply easier syntax in processor definition.

Example of such new format:

```
"a.b.c.d": "field"
```
or

```
"a.b" : {
   "c.d": "field"
}
```

in this case we will try to look for structure like this in the ingest document:

```
"a" : {
   "b": {
     "c" : {
       "d" : "field"
```

Today we do support only hierarchical type of definition in mapping. It must look exactly like it is in the document:

```
"a" : {
   "b": {
     "c" : {
       "d" : "field"
```  

**Note**: this change affects only source field (left part of the mapping, one that holds value that is a basis for embedding generation). Today logic for the destination field (right part in mapping, field that will store generated embeddings) will be unchanged. As per today's logic that destination field is expected at the same level with the final source field. Example: `"a.b.c: d"` in this case embeddings are inserted in following structure: 
```
"a" : {
   "b" : {
     "d" : [0.1, 0.2 ....]
```

### Issues Resolved
https://github.com/opensearch-project/neural-search/issues/110

### Check List
    - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
@@ -65,7 +65,6 @@ public abstract class InferenceProcessor extends AbstractProcessor {
 
     private final Environment environment;
     private final ClusterService clusterService;
-    List<Pair<String, Object>> mappingForNestedFields;
 
     public InferenceProcessor(
         String tag,
@@ -88,7 +87,6 @@ public abstract class InferenceProcessor extends AbstractProcessor {
         this.mlCommonsClientAccessor = clientAccessor;
         this.environment = environment;
         this.clusterService = clusterService;
-        mappingForNestedFields = fieldMap.entrySet().stream().map(this::processNestedKey).collect(Collectors.toList());
     }
 
     private void validateEmbeddingConfiguration(Map<String, Object> fieldMap) {
@@ -284,7 +282,8 @@ public abstract class InferenceProcessor extends AbstractProcessor {
     Map<String, Object> buildMapWithTargetKeys(IngestDocument ingestDocument) {
         Map<String, Object> sourceAndMetadataMap = ingestDocument.getSourceAndMetadata();
         Map<String, Object> mapWithProcessorKeys = new LinkedHashMap<>();
-        for (Pair<String, Object> processedNestedKey : mappingForNestedFields) {
+        for (Map.Entry<String, Object> fieldMapEntry : fieldMap.entrySet()) {
+            Pair<String, Object> processedNestedKey = processNestedKey(fieldMapEntry);
             String originalKey = processedNestedKey.getKey();
             Object targetKey = processedNestedKey.getValue();
 
--- a/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
@@ -6,6 +6,7 @@ package org.opensearch.neuralsearch.processor;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
@@ -21,6 +22,8 @@ import java.util.stream.IntStream;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
 import org.opensearch.common.collect.Tuple;
 import org.opensearch.core.common.util.CollectionUtils;
 import org.opensearch.cluster.service.ClusterService;
@@ -120,7 +123,7 @@ public abstract class InferenceProcessor extends AbstractProcessor {
     public void execute(IngestDocument ingestDocument, BiConsumer<IngestDocument, Exception> handler) {
         try {
             validateEmbeddingFieldsValue(ingestDocument);
-            Map<String, Object> processMap = buildMapWithTargetKeyAndOriginalValue(ingestDocument);
+            Map<String, Object> processMap = buildMapWithTargetKeys(ingestDocument);
             List<String> inferenceList = createInferenceList(processMap);
             if (inferenceList.size() == 0) {
                 handler.accept(ingestDocument, null);
@@ -228,7 +231,7 @@ public abstract class InferenceProcessor extends AbstractProcessor {
             List<String> inferenceList = null;
             try {
                 validateEmbeddingFieldsValue(ingestDocumentWrapper.getIngestDocument());
-                processMap = buildMapWithTargetKeyAndOriginalValue(ingestDocumentWrapper.getIngestDocument());
+                processMap = buildMapWithTargetKeys(ingestDocumentWrapper.getIngestDocument());
                 inferenceList = createInferenceList(processMap);
             } catch (Exception e) {
                 ingestDocumentWrapper.update(ingestDocumentWrapper.getIngestDocument(), e);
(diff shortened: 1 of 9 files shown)


src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -317,21 +325,45 @@ private void buildMapWithProcessorKeyAndOriginalValueForMapType(
...
             treeRes.put(key, sourceAndMetadataMap.get(parentKey));
         }
     }
 
+    /**
+     * Process the nested key, such as "a.b.c" to "a", "b.c"
+     * @param nestedFieldMapEntry
+     * @return A pair of the original key and the target key
+     */
+    private Pair<String, Object> processNestedKey(final Map.Entry<String, Object> nestedFieldMapEntry) {
```
heemin32, line 353: Do we have a unit test for this?
I don't see any test for this method in `InferenceProcessorTests` file.

Also, this might break backward compatibility for user who is using dot in their key already. Are we fine on that?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -293,20 +298,23 @@ Map<String, Object> buildMapWithTargetKeyAndOriginalValue(IngestDocument ingestD
         return mapWithProcessorKeys;
     }
 
-    private void buildMapWithProcessorKeyAndOriginalValueForMapType(
+    private void buildNestedMap(
```
heemin32, line 301: Would be nice to have unit test for this method

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -276,15 +279,17 @@ private void createInferenceListForMapTypeInput(Object sourceValue, List<String>
     }
 
     @VisibleForTesting
-    Map<String, Object> buildMapWithTargetKeyAndOriginalValue(IngestDocument ingestDocument) {
+    Map<String, Object> buildMapWithTargetKeys(IngestDocument ingestDocument) {
         Map<String, Object> sourceAndMetadataMap = ingestDocument.getSourceAndMetadata();
         Map<String, Object> mapWithProcessorKeys = new LinkedHashMap<>();
         for (Map.Entry<String, Object> fieldMapEntry : fieldMap.entrySet()) {
```
heemin32, line 285: I assume field map is what will be defined when user create processor?
Wouldn't it be better to parse and expand the key during creation time rather than ingestion time?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -276,15 +279,17 @@ private void createInferenceListForMapTypeInput(Object sourceValue, List<String>
...
-            String originalKey = fieldMapEntry.getKey();
-            Object targetKey = fieldMapEntry.getValue();
+            Pair<String, Object> processedNestedKey = processNestedKey(fieldMapEntry);
```
heemin32, line 286: Is the fieldMapEntry what user define during processor creation?
If that is the case, can we parse and expand the key during creation time?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
martin-gaievski: let me add the test. For the bwc we should be fine, it's more like an edge case if someone expecting "a.b.c" to be a single field and not the structure. It's one way or another and we're making this change based on requests from users and because this syntax supported in other places in Open search.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
martin-gaievski: yes, it should be possible to construct that map when we create a processor class. let me do the change

heemin32: Cool. That way, we might be able to reduce some latency during ingestion.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
heemin32: Not about bwc test. Just writing it as unit test friendly would make it easier to maintain.

martin-gaievski: Got it, added unit tests

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -62,6 +65,7 @@ public abstract class InferenceProcessor extends AbstractProcessor {
 
     private final Environment environment;
     private final ClusterService clusterService;
+    List<Pair<String, Object>> mappingForNestedFields;
```
heemin32, line 68: Can we make it as private? why should it be package scope?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -84,6 +88,7 @@ public InferenceProcessor(
         this.mlCommonsClientAccessor = clientAccessor;
         this.environment = environment;
         this.clusterService = clusterService;
+        mappingForNestedFields = fieldMap.entrySet().stream().map(this::processNestedKey).collect(Collectors.toList());
```
heemin32, line 91: I didn't mean to just move this in constructor. As we are calling processNestedKey inside execute method anyway, there is no much difference between this code and previous code. Actually, previous code is better...

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -317,21 +322,46 @@ private void buildMapWithProcessorKeyAndOriginalValueForMapType(
...
             treeRes.put(key, sourceAndMetadataMap.get(parentKey));
         }
     }
 
+    /**
+     * Process the nested key, such as "a.b.c" to "a", "b.c"
+     * @param nestedFieldMapEntry
+     * @return A pair of the original key and the target key
+     */
+    @VisibleForTesting
```
heemin32, line 350: I am fine for this PR but @VisibleForTest is not the best way for unit test because it makes it hard to write unit test for `execute` method as there is no easy way to mock the sub methods.

The ideal way should be something like
```
public InferenceProcessor (MyClass myclass) {
  this.myclass = myclass;
}

public void execute() {
  myclass.doSomething();
  myclass.doAnother();
  ...
}
```

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
martin-gaievski: got it, yeah there is no easy way to remove that call from recursion. reverting back to the previous state

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
martin-gaievski: not relevant as I'm reverting back to previous approach

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
martin-gaievski: agree, we can do such refactoring for tests specifically where we do complex logic that needs matrix of all permutations of params and/or input.

heemin32 (approved):

vibrantvarun (approved):
Looks good. Curious how nested field is processed currently without this change?

martin-gaievski:
> Looks good. Curious how nested field is processed currently without this change?

Today we treat field name separated with `.` as a single field name, if someone put `"a.b" : "c"` we will pickup the field only if doc has field `"a.b"`, not `"a": { "b"}`. There is a workaround, user can put mapping as a structured json, I put example in PR description. In this case UX isn't great for cases when objects have complex structure.

PR #868, opened 2024-08-20
Upgrade BWC version and fix neural search build

vibrantvarun:
### Description
This PR contains couple of things.
1. Fixing Neural Search build by encompassing latest KNN changes.
2. Increasing BWC version to 2.17.0-SNAPSHOT

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/neural-search/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #869, opened 2024-08-21
Fix neural search build & increment OS version to 2.17

vibrantvarun:
### Description
The PR is  backporting changes made in the following PR https://github.com/opensearch-project/neural-search/pull/868. 
Moreover, this PR also contains changes to increment OS version to 2.17 on branch 2.x. Additionally, it also updates BWC version to 2.17 in github workflows.

Also we have removed org.apache.logging.log4j.util.Strings from MLCommonsClientAccessor.java due to following campaign made in opensearch core. https://github.com/opensearch-project/OpenSearch/pull/15238

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
@@ -15,7 +15,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Collectors;
 
-import org.apache.logging.log4j.util.Strings;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.common.util.CollectionUtils;
 import org.opensearch.ml.client.MachineLearningNodeClient;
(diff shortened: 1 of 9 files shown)


navneet1v (approved):

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -311,4 +310,18 @@ private MLInput createMLMultimodalInput(final List<String> targetResponseFilters
         final MLInputDataset inputDataset = new TextDocsInputDataSet(inputText, modelResultFilter);
         return new MLInput(FunctionName.TEXT_EMBEDDING, null, inputDataset);
     }
+
+    private boolean checkIfStringIsNotBlank(String value) {
```
naveentatikonda, line 314: You can use [this function](https://github.com/opensearch-project/OpenSearch/blob/main/libs/core/src/main/java/org/opensearch/core/common/Strings.java#L131-L173) from opensearch core to achieve this functionality.

We need to replace `org.apache.logging.log4j.util.Strings` with `org.opensearch.core.common.Strings`

naveentatikonda (changes requested):

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
naveentatikonda: Also, you might need to raise a separate PR to fix it in main branch

naveentatikonda (approved):

Issue #429, opened 2023-10-10
[Tech debt] Improve test coverage - phase 1, meet minimal required level


martin-gaievski:
Improve test coverage for repo, all packages should be above 80%.

Currently report is showing following packages that are below 80%:

- https://github.com/opensearch-project/neural-search/tree/main/src/main/java/org/opensearch/neuralsearch/plugin - 71%
- https://github.com/opensearch-project/neural-search/tree/main/src/main/java/org/opensearch/neuralsearch/query - 77%
- https://github.com/opensearch-project/neural-search/tree/main/src/main/java/org/opensearch/neuralsearch/search - 77%

dbwiddis:
It would probably be helpful to [post the coverage report link](https://app.codecov.io/gh/opensearch-project/neural-search/tree/main/src%2Fmain%2Fjava%2Forg%2Fopensearch%2Fneuralsearch) in the issue description :)

PR #898, opened 2024-09-07
Add coverage for NeuralSearch class

dbwiddis:
### Description

Adds test coverage for the main plugin class file

### Related Issues

Part of #429 

### Check List

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,27 @@
-# Ignore Gradle project-specific cache directory
-.gradle
+# intellij files
+.idea/
+*.iml
+*.ipr
+*.iws
+*.log
+build-idea/
+out/
 
-# Ignore Gradle build output directory
-build
-.idea
+# eclipse files
+.classpath
+.project
+.settings
+
+# gradle stuff
+.gradle/
+build/
+bin/
+
+# vscode stuff
+.vscode/
+
+# osx stuff
 .DS_Store
+
+# git stuff
 .gitattributes
-
-
--- a/src/test/java/org/opensearch/neuralsearch/plugin/NeuralSearchTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/plugin/NeuralSearchTests.java
@@ -4,9 +4,11 @@
  */
 package org.opensearch.neuralsearch.plugin;
 
+import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -17,6 +19,7 @@ import org.mockito.MockitoAnnotations;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Setting;
 import org.opensearch.common.settings.Settings;
+import org.opensearch.common.util.concurrent.OpenSearchExecutors;
 import org.opensearch.env.Environment;
 import org.opensearch.indices.IndicesService;
 import org.opensearch.ingest.IngestService;
@@ -42,6 +45,7 @@ import org.opensearch.search.pipeline.SearchResponseProcessor;
 import org.opensearch.search.query.QueryPhaseSearcher;
 import org.opensearch.threadpool.ExecutorBuilder;
 import org.opensearch.threadpool.FixedExecutorBuilder;
+import org.opensearch.threadpool.ThreadPool;
 
 public class NeuralSearchTests extends OpenSearchQueryTestCase {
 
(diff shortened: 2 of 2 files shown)


vibrantvarun:
@dbwiddis can you check why gradle check is failing?

dbwiddis:
> @dbwiddis can you check why gradle check is failing?

Unrelated to my PR:
```
Tests with failures:
 - org.opensearch.neuralsearch.query.HybridQueryWeightTests.testSubQueries_whenMultipleEqualSubQueries_thenSuccessful
 - org.opensearch.neuralsearch.query.HybridQueryWeightTests.classMethod
 - org.opensearch.neuralsearch.search.query.HybridQueryPhaseSearcherTests.testQueryResult_whenMultipleTextSubQueriesWithSomeHits_thenHybridResultsAreSet
 - org.opensearch.neuralsearch.search.query.HybridQueryPhaseSearcherTests.classMethod
446 tests completed, 4 failed
```

I can keep force-pushing until it's green but generally it's easier for repo maintainers to retry just the failed tests.

vibrantvarun:
@dbwiddis can you add 
```
@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
```
HybridQueryWeightTests and HybridQueryPhaseSearcherTests classes and add it into your PR ?

dbwiddis:
> @dbwiddis can you check why gradle check is failing?

Looking into the log, it seems it's timing out and it's likely related to thread locking.  Multiple threads show
```
  2> 	Locked synchronizers:
  2> 	- java.util.concurrent.ThreadPoolExecutor$Worker@386f0da3
```
This makes me suspect some interaction with this PR which does mock a threadpool and an `ExcecutorService` to be returned by the threadpool:
```java
when(threadPool.executor(anyString())).thenReturn(executorService);
```
I expected that mock to be isolated only to this test, and of course when run in isolation the test passes.  However, looking at the plugin `createComponents`, I see this line:
https://github.com/opensearch-project/neural-search/blob/f58d989c68016b37e4c5e6354ab26c815a8ba7d5/src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java#L101C9-L101C22

That leads to this line:
https://github.com/opensearch-project/neural-search/blob/f58d989c68016b37e4c5e6354ab26c815a8ba7d5/src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutor.java#L60

That sets the value of a `static` variable, which is therefore no longer isolated to that test: it remains stuck with a mocked value for any test run after this one.
https://github.com/opensearch-project/neural-search/blob/f58d989c68016b37e4c5e6354ab26c815a8ba7d5/src/main/java/org/opensearch/neuralsearch/executors/HybridQueryExecutor.java#L31

I could replace the mock with a real thread pool and executor, somewhat like this pattern we use in Flow Framework and that I've carried over into ML Commons in a few places:
```java
public class ProcessNodeTests extends OpenSearchTestCase {

    private static TestThreadPool testThreadPool;

    @BeforeClass
    public static void setup() {
        testThreadPool = new TestThreadPool(
            ProcessNodeTests.class.getName(),
            new ScalingExecutorBuilder(
                PROVISION_WORKFLOW_THREAD_POOL,
                1,
                Math.max(1, OpenSearchExecutors.allocatedProcessors(Settings.EMPTY) - 1),
                TimeValue.timeValueMinutes(5),
                FLOW_FRAMEWORK_THREAD_POOL_PREFIX + PROVISION_WORKFLOW_THREAD_POOL
            )
        );
    }

    @AfterClass
    public static void cleanup() {
        ThreadPool.terminate(testThreadPool, 500, TimeUnit.MILLISECONDS);
    }
```
But this works because the static variable is confined to this class and used only within this test.  If I used this pattern in this test, it would still initialize the `TaskExecutor` in `HybridQueryExecutor` to this pool, and it would break tests when it was shut down.

I could leave it running for the entire duration of the test suite, but `OpenSearchTestCase` has leak detection mechanisms, which you've alluded to with the `@ThreadLeakScope(ThreadLeakScope.Scope.NONE)` annotation.

This still doesn't solve the underlying problem:  there is no way to ever execute `createComponents` without a real thread pool you want to use for all the tests, that you can shut down at the end of your testing.

Putting aside the question of whether a static object initialized by a different class is even a good idea at all (which I don't think it is), at this point I would prefer to just remove the perfunctory test of `createComponents()` whose entire `raison d'être` is to exercise the method and cover those lines to increase an abitrary coverage metric.

dbwiddis:
TLDR: I'll push a commit removing the coverage for `createComponents()` and leave it to a more well thought out design.

dbwiddis:
OpenSearch has a direct executor service that simulates the (Runnable::run) default behavior of the non-initialized taskExecutor, so I used that.

I still think there's a thread leak in the existing implementation, but that's not part of this PR.

vibrantvarun (approved):

martin-gaievski (approved):

Issue #926, opened 2024-10-07
[Meta] ReRankProcessor enhancement: ReRank by Field


brianf-aws:
### Is your feature request related to a problem?
Currently if a user has access to a field in their document or in a search response we would like to be able to re-rank by that metric. Currently the ReRank process supports [ML_OpenSearch](https://github.com/opensearch-project/neural-search/blob/9f4a49a7e45211821d96181ce2a6842af18ce7ea/src/main/java/org/opensearch/neuralsearch/processor/rerank/RerankType.java#L19) as a way to re rank. We would like to provide users with a way to perform a 2nd level re ranking, **We would like to make it a `ByFieldRerankProcessor`**

### What solution would you like?
The rescoring logic (i.e update the `_score` field to reflect a new score) is already provided in the [RescoringReRankProcessor](https://github.com/opensearch-project/neural-search/blob/9f4a49a7e45211821d96181ce2a6842af18ce7ea/src/main/java/org/opensearch/neuralsearch/processor/rerank/RescoringRerankProcessor.java#L80). All that would be required is to implement ByFieldRerankProcessor to use the scores provided by the document or a previous search response.

Ideally the interface would look like this (My implementation is in the Neural Search repo)
```
{
    "response_processors": [
        {
            "rerank": {
                "by_field": {
                    "target_field": "ml_score",
                     "remove_target_field": true, ## Default false
                     "keep_previous_score" : true ## Default false
                }
            }
        }
    ]
}
```
It was discussed that previous scores should be kept as an option as this may hinder the expectation of the user, thus we added a field called `keep_previous_score` to allow this. We also want to  give an option to be able delete the field (via `remove_target_field`) provided to perform the re ranking as this is redundant data.

### What alternatives have you considered?
Creating a separate response processor, in open search core, that replaced and sorted the response. This was initially proposed in OpenSearch Core https://github.com/opensearch-project/OpenSearch/issues/15631. But after offline discussion we decided that this functionality could be transferred to a processor that already does re ranking, in the Neural Search repo.

### Do you have any additional context?
This functionality was brought up as a necessity to enhance the ML Inference Processor in the ML-Commons codebase
- https://github.com/opensearch-project/ml-commons/issues/2878
- https://github.com/opensearch-project/OpenSearch/issues/15631

PR #932, opened 2024-10-14
ByFieldRerank Processor (ReRankProcessor enhancement)

brianf-aws:
### Description

```
{
    "response_processors": [
        {
            "rerank": {
                "by_field": {
                    "target_field": "ml_score",
                     "remove_target_field": true, ## Default false
                    "keep_previous_score" : true ## Default false
                }
            }
        }
    ]
}
```

The ByFieldRerank Processor applies a 2nd level re ranking on a search hit by specifying a targetField in the `_source` mapping of a search hit. It will update the `_score` field with what has been provided by `target_field`, When  `keep_previous_score` is asserted a new field is appended called `previous_score` which was the score prior to reranking. Currently we support both shallow target and also a nested field target such as `a.b.c` given that this mapping exists.

When deleting a target field as specified by `"remove_target_field": true` the processor will delete any empty maps as a result of this action for example, when the targetField is `ml_info.score_report`. It will transform
```
{
   "my_field": "value"
   "ml_info" : {
       "score_report" : 27
   }
}
```
into the following
```
{
   "my_field": "value"
}
```
It was chosen this way to give users the flexibility to clean up their searchHit, however the default behavior was disabled to let new users adapt to the feature.

### Related Issues
Resolves https://github.com/opensearch-project/neural-search/issues/926
Resolves https://github.com/opensearch-project/OpenSearch/issues/15631

### Changelog
- 10/17/2024 : onboarded a new parameter to `keep_previous_score` to be able to give possibility of adding the `previous_score` field to the search hit. Was introduced in commit https://github.com/opensearch-project/neural-search/pull/932/commits/66866a430241b3d8bd6306a97922f4685f9e1b5c
- 10/21/2024: Added a new functional interface `void validate(SearchHit hit) throws IllegalArgumentException;` in a new Utility class `processorSearchResponseUtil` so that future rerankers can take advantage of validating their data before reranking.

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ## [Unreleased 2.x](https://github.com/opensearch-project/neural-search/compare/2.17...2.x)
 ### Features
 ### Enhancements
+- Implement `ignore_missing` field in text chunking processors ([#907](https://github.com/opensearch-project/neural-search/pull/907))
+- Added rescorer in hybrid query ([#917](https://github.com/opensearch-project/neural-search/pull/917))
 ### Bug Fixes
 ### Infrastructure
 ### Documentation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Enhancements
 - Set neural-search plugin 3.0.0 baseline JDK version to JDK-2 ([#838](https://github.com/opensearch-project/neural-search/pull/838))
 ### Bug Fixes
+- Fix for nested field missing sub embedding field in text embedding processor ([#913](https://github.com/opensearch-project/neural-search/pull/913))
 ### Infrastructure
 ### Documentation
 ### Maintenance
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 2.x](https://github.com/opensearch-project/neural-search/compare/2.18...2.x)
 ### Features
-- Introduces ByFieldRerankProcessor for second level reranking on documents (This is done locally) ([#932](https://github.com/opensearch-project/neural-search/pull/932))
-
 ### Enhancements
 ### Bug Fixes
 ### Infrastructure
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,13 +14,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Maintenance
 ### Refactoring
 
-## [Unreleased 2.x](https://github.com/opensearch-project/neural-search/compare/2.17...2.x)
+## [Unreleased 2.x](https://github.com/opensearch-project/neural-search/compare/2.18...2.x)
 ### Features
 - Introduces ByFieldRerankProcessor for second level reranking on documents (This is done locally) ([#932](https://github.com/opensearch-project/neural-search/pull/932))
 
 ### Enhancements
-- Implement `ignore_missing` field in text chunking processors ([#907](https://github.com/opensearch-project/neural-search/pull/907))
-- Added rescorer in hybrid query ([#917](https://github.com/opensearch-project/neural-search/pull/917))
 ### Bug Fixes
 ### Infrastructure
 ### Documentation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 2.x](https://github.com/opensearch-project/neural-search/compare/2.17...2.x)
 ### Features
+- Introduces ByFieldRerankProcessor for second level reranking on documents (This is done locally) ([#932](https://github.com/opensearch-project/neural-search/pull/932))
+
 ### Enhancements
 - Implement `ignore_missing` field in text chunking processors ([#907](https://github.com/opensearch-project/neural-search/pull/907))
 - Added rescorer in hybrid query ([#917](https://github.com/opensearch-project/neural-search/pull/917))
--- a/release-notes/opensearch-neural-search.release-notes-2.18.0.0.md
+++ b/release-notes/opensearch-neural-search.release-notes-2.18.0.0.md
@@ -1,0 +1,9 @@
+
+## Version 2.18.0.0 Release Notes
+
+Compatible with OpenSearch 2.18.0
+
+### Enhancements
+- Implement `ignore_missing` field in text chunking processors ([#907](https://github.com/opensearch-project/neural-search/pull/907))
+- Added rescorer in hybrid query ([#917](https://github.com/opensearch-project/neural-search/pull/917))
+
--- a/release-notes/opensearch-neural-search.release-notes-2.18.0.0.md
+++ b/release-notes/opensearch-neural-search.release-notes-2.18.0.0.md
@@ -3,6 +3,9 @@
 
 Compatible with OpenSearch 2.18.0
 
+### Features
+- Introduces ByFieldRerankProcessor for second level reranking on documents (This is done locally) ([#932](https://github.com/opensearch-project/neural-search/pull/932))
+
 ### Enhancements
 - Implement `ignore_missing` field in text chunking processors ([#907](https://github.com/opensearch-project/neural-search/pull/907))
 - Added rescorer in hybrid query ([#917](https://github.com/opensearch-project/neural-search/pull/917))
--- a/release-notes/opensearch-neural-search.release-notes-2.18.0.0.md
+++ b/release-notes/opensearch-neural-search.release-notes-2.18.0.0.md
@@ -4,7 +4,7 @@
 Compatible with OpenSearch 2.18.0
 
 ### Features
-- Introduces ByFieldRerankProcessor for second level reranking on documents (This is done locally) ([#932](https://github.com/opensearch-project/neural-search/pull/932))
+- Introduces ByFieldRerankProcessor for second level reranking on documents ([#932](https://github.com/opensearch-project/neural-search/pull/932))
 
 ### Enhancements
 - Implement `ignore_missing` field in text chunking processors ([#907](https://github.com/opensearch-project/neural-search/pull/907))
--- a/src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
@@ -72,6 +72,7 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
                 );
                 return new MLOpenSearchRerankProcessor(description, tag, ignoreFailure, modelId, contextFetchers, clientAccessor);
             case BY_FIELD:
+                boolean DEFAULT_REMOVE_TARGET_FIELD = false;
                 String targetField = ConfigurationUtils.readStringProperty(
                     RERANK_PROCESSOR_TYPE,
                     tag,
@@ -83,7 +84,7 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
                     tag,
                     rerankerConfig,
                     ByFieldRerankProcessor.REMOVE_TARGET_FIELD,
-                    false
+                    DEFAULT_REMOVE_TARGET_FIELD
                 );
 
                 return new ByFieldRerankProcessor(description, tag, ignoreFailure, targetField, removeTargetField, contextFetchers);
--- a/src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
@@ -4,16 +4,12 @@
  */
 package org.opensearch.neuralsearch.processor.factory;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Set;
-import java.util.StringJoiner;
-
+import com.google.common.collect.Sets;
+import lombok.AllArgsConstructor;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.ingest.ConfigurationUtils;
 import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
+import org.opensearch.neuralsearch.processor.rerank.ByFieldRerankProcessor;
 import org.opensearch.neuralsearch.processor.rerank.MLOpenSearchRerankProcessor;
 import org.opensearch.neuralsearch.processor.rerank.RerankType;
 import org.opensearch.neuralsearch.processor.rerank.context.ContextSourceFetcher;
@@ -22,9 +18,12 @@ import org.opensearch.neuralsearch.processor.rerank.context.QueryContextSourceFe
 import org.opensearch.search.pipeline.Processor;
 import org.opensearch.search.pipeline.SearchResponseProcessor;
 
-import com.google.common.collect.Sets;
-
-import lombok.AllArgsConstructor;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+import java.util.StringJoiner;
 
 /**
  * Factory for rerank processors. Must:
@@ -51,15 +50,15 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
     ) {
         RerankType type = findRerankType(config);
         boolean includeQueryContextFetcher = ContextFetcherFactory.shouldIncludeQueryContextFetcher(type);
-        List<ContextSourceFetcher> contextFetchers = ContextFetcherFactory.createFetchers(
-            config,
-            includeQueryContextFetcher,
-            tag,
-            clusterService
-        );
+        List<ContextSourceFetcher> contextFetchers = List.of();
+        if (type != RerankType.BY_FIELD) {
+            contextFetchers = ContextFetcherFactory.createFetchers(config, includeQueryContextFetcher, tag, clusterService);
+        }
+
+        Map<String, Object> rerankerConfig = ConfigurationUtils.readMap(RERANK_PROCESSOR_TYPE, tag, config, type.getLabel());
+
         switch (type) {
             case ML_OPENSEARCH:
-                Map<String, Object> rerankerConfig = ConfigurationUtils.readMap(RERANK_PROCESSOR_TYPE, tag, config, type.getLabel());
                 String modelId = ConfigurationUtils.readStringProperty(
                     RERANK_PROCESSOR_TYPE,
                     tag,
@@ -67,6 +66,14 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
                     MLOpenSearchRerankProcessor.MODEL_ID_FIELD
                 );
                 return new MLOpenSearchRerankProcessor(description, tag, ignoreFailure, modelId, contextFetchers, clientAccessor);
+            case BY_FIELD:
+                String targetField = ConfigurationUtils.readStringProperty(
+                    RERANK_PROCESSOR_TYPE,
+                    tag,
+                    rerankerConfig,
+                    ByFieldRerankProcessor.TARGET_FIELD
+                );
+                return new ByFieldRerankProcessor(description, tag, ignoreFailure, targetField, contextFetchers);
             default:
                 throw new IllegalArgumentException(String.format(Locale.ROOT, "Cannot build reranker type %s", type.getLabel()));
         }
@@ -100,6 +107,7 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
 
         /**
          * Map rerank types to whether they should include the query context source fetcher
+         *
          * @param type the constructing RerankType
          * @return does this RerankType depend on the QueryContextSourceFetcher?
          */
@@ -109,7 +117,8 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
 
         /**
          * Create necessary queryContextFetchers for this processor
-         * @param config processor config object. Look for "context" field to find fetchers
+         *
+         * @param config                     processor config object. Look for "context" field to find fetchers
          * @param includeQueryContextFetcher should I include the queryContextFetcher?
          * @return list of contextFetchers for the processor to use
          */
--- a/src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
@@ -26,6 +26,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.StringJoiner;
 
+import static org.opensearch.neuralsearch.processor.rerank.ByFieldRerankProcessor.DEFAULT_KEEP_PREVIOUS_SCORE;
+import static org.opensearch.neuralsearch.processor.rerank.ByFieldRerankProcessor.DEFAULT_REMOVE_TARGET_FIELD;
 import static org.opensearch.neuralsearch.processor.rerank.RerankProcessor.processorRequiresContext;
 
 /**
@@ -72,9 +74,6 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
                 );
                 return new MLOpenSearchRerankProcessor(description, tag, ignoreFailure, modelId, contextFetchers, clientAccessor);
             case BY_FIELD:
-                boolean DEFAULT_REMOVE_TARGET_FIELD = false;
-                boolean DEFAULT_KEEP_PREVIOUS_SCORE = false;
-
                 String targetField = ConfigurationUtils.readStringProperty(
                     RERANK_PROCESSOR_TYPE,
                     tag,
--- a/src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
@@ -42,11 +42,6 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
     private final MLCommonsClientAccessor clientAccessor;
     private final ClusterService clusterService;
 
-    public RerankProcessorFactory(ClusterService clusterService) {
-        this.clusterService = clusterService;
-        this.clientAccessor = null;
-    }
-
     @Override
     public SearchResponseProcessor create(
         final Map<String, Processor.Factory<SearchResponseProcessor>> processorFactories,
@@ -59,6 +54,8 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
         RerankType type = findRerankType(config);
         boolean includeQueryContextFetcher = ContextFetcherFactory.shouldIncludeQueryContextFetcher(type);
 
+        // Currently the createFetchers method requires that you provide a context map, this branch makes sure we can ignore this on
+        // processors that don't need the context map
         List<ContextSourceFetcher> contextFetchers = processorRequiresContext(type)
             ? ContextFetcherFactory.createFetchers(config, includeQueryContextFetcher, tag, clusterService)
             : Collections.emptyList();
@@ -133,9 +130,8 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
 
         /**
          * Create necessary queryContextFetchers for this processor
-         *
-         * @param config                     processor config object. Look for "context" field to find fetchers
-         * @param includeQueryContextFetcher should I include the queryContextFetcher?
+         * @param config Processor config object. Look for "context" field to find fetchers
+         * @param includeQueryContextFetcher Should I include the queryContextFetcher?
          * @return list of contextFetchers for the processor to use
          */
         public static List<ContextSourceFetcher> createFetchers(
--- a/src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
@@ -19,11 +19,14 @@ import org.opensearch.search.pipeline.Processor;
 import org.opensearch.search.pipeline.SearchResponseProcessor;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.StringJoiner;
+
+import static org.opensearch.neuralsearch.processor.rerank.RerankProcessor.processorRequiresContext;
 
 /**
  * Factory for rerank processors. Must:
@@ -39,6 +42,11 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
     private final MLCommonsClientAccessor clientAccessor;
     private final ClusterService clusterService;
 
+    public RerankProcessorFactory(ClusterService clusterService) {
+        this.clusterService = clusterService;
+        this.clientAccessor = null;
+    }
+
     @Override
     public SearchResponseProcessor create(
         final Map<String, Processor.Factory<SearchResponseProcessor>> processorFactories,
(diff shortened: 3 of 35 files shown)


brianf-aws:
The functionality has been implemented all that is left is more UTs, hence I left the PR as a draft. I can't think of integration testing for it as it doesn't interact with a remote process.

martin-gaievski:
> The functionality has been implemented all that is left is more UTs, hence I left the PR as a draft. I can't think of integration testing for it as it doesn't interact with a remote process.

We can interact with local model, it should be fine for the sake of integ test, something similar to how it's done in existing test for rerank processor https://github.com/opensearch-project/neural-search/blob/main/src/test/java/org/opensearch/neuralsearch/processor/rerank/MLOpenSearchRerankProcessorIT.java

src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
```diff
@@ -51,22 +58,38 @@ public SearchResponseProcessor create(
...
         RerankType type = findRerankType(config);
         boolean includeQueryContextFetcher = ContextFetcherFactory.shouldIncludeQueryContextFetcher(type);
-        List<ContextSourceFetcher> contextFetchers = ContextFetcherFactory.createFetchers(
-            config,
-            includeQueryContextFetcher,
-            tag,
-            clusterService
-        );
+
+        List<ContextSourceFetcher> contextFetchers = processorRequiresContext(type)
```
vibrantvarun, line 62: Can you add a comment above explaining about why this condition is added.

src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
```diff
@@ -40,6 +42,11 @@ public class RerankProcessorFactory implements Processor.Factory<SearchResponseP
     private final MLCommonsClientAccessor clientAccessor;
     private final ClusterService clusterService;
 
+    public RerankProcessorFactory(ClusterService clusterService) {
```
vibrantvarun, line 45: There is already @AllArgsConstructor added at the top. You can use that while calling from test methods.
This constructor is not needed.

src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
```diff
@@ -51,22 +58,38 @@ public SearchResponseProcessor create(
...
                 );
                 return new MLOpenSearchRerankProcessor(description, tag, ignoreFailure, modelId, contextFetchers, clientAccessor);
+            case BY_FIELD:
+                String targetField = ConfigurationUtils.readStringProperty(
+                    RERANK_PROCESSOR_TYPE,
+                    tag,
+                    rerankerConfig,
+                    ByFieldRerankProcessor.TARGET_FIELD
+                );
+                boolean removeTargetField = ConfigurationUtils.readBooleanProperty(
```
vibrantvarun, line 84: Can this readBooleanProperty return null? If yes then use Boolean instead of boolean to cater the field value.

src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
```diff
@@ -109,7 +133,8 @@ public static boolean shouldIncludeQueryContextFetcher(RerankType type) {
 
         /**
          * Create necessary queryContextFetchers for this processor
-         * @param config processor config object. Look for "context" field to find fetchers
+         *
+         * @param config                     processor config object. Look for "context" field to find fetchers
```
vibrantvarun, line 137: Remove this extra space and revert to original form.

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,213 @@
...
+public class ByFieldRerankProcessor extends RescoringRerankProcessor {
+
+    public static final String TARGET_FIELD = "target_field";
+    public static final String REMOVE_TARGET_FIELD = "remove_target_field";
+
+    protected final String targetField;
+    protected final boolean removeTargetField;
+
+    /**
+     * Constructor. pass through to RerankProcessor constructor.
```
vibrantvarun, line 32:
```suggestion
     * Constructor to pass values to RerankProcessor constructor.
```

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,213 @@
...
+        super(RerankType.BY_FIELD, description, tag, ignoreFailure, contextSourceFetchers);
+        this.targetField = targetField;
+        this.removeTargetField = removeTargetField;
+    }
+
+    @Override
+    public void rescoreSearchResponse(SearchResponse response, Map<String, Object> rerankingContext, ActionListener<List<Float>> listener) {
+        SearchHit[] searchHits = response.getHits().getHits();
+
+        if (!searchHitsHaveValidForm(searchHits, listener)) {
```
vibrantvarun, line 58:
```suggestion
        if (!validateSearchHits(searchHits, listener)) {
```

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,213 @@
...
+            return;
+        }
+
+        List<Float> scores = new ArrayList<>(searchHits.length);
+
+        for (SearchHit hit : searchHits) {
+            Tuple<? extends MediaType, Map<String, Object>> typeAndSourceMap = getMapTuple(hit);
```
vibrantvarun, line 65: Can you come up with some logic method name? getMapTuple is complex and does help to understand the code.

Also typeAndSourceMap represent what type and source. The name should be self explanatory.

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,213 @@
...
+            Map<String, Object> sourceAsMap = typeAndSourceMap.v2();
+
+            Object val = getValueFromMap(sourceAsMap, targetField).get();
+            scores.add(((Number) val).floatValue());
+
+            sourceAsMap.put("previous_score", hit.getScore());
+            if (removeTargetField) {
```
vibrantvarun, line 72: Add comment on why this check is here. Also what does removetargetField do.

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,213 @@
...
+                removeTargetFieldFromMap(sourceAsMap);
```
vibrantvarun, line 73:
```suggestion
                removeTargetFieldFromSource(source);
```

vibrantvarun (commented):
Did the first round of review. Please address comments and let me know.

src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
brianf-aws: I can see what you mean as it is already being set to null in the definition of the constructor. My intention was to not pass in a clientAccessor as a null parameter.

src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
brianf-aws: Yeah so this last argument, has the default of false in case it can't find the mapping

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
brianf-aws: oh good point how does `getMapTuple(hit) ` -> `getTypeAndSourceMapTuple(hit)` sound?

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
brianf-aws: Right, this is the feature of the ByField Processor (It essentially gives the user to delete a field they have used to update  the `_score` field), I have updated the constructor to better explain was in a rush to push the draft pr 😅

src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
brianf-aws: Understood, I added this comment 
```
        // Currently the createFetchers method requires that you provide a context map, this branch makes sure we can ignore this on
        // processors that don't need the context map
```

src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
brianf-aws: Thanks, its back to how it was

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
brianf-aws: Good suggestion I've added that

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
brianf-aws: Ive implemented that will push the commit soon

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
brianf-aws: Nice catch, thanks!!

src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
brianf-aws: Ive implemented your suggestion after thinking through some more. Since this is a factory it wont be as confusing to users who would like to create the processor since they will just create the processor using the factory.

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,271 @@
...
+ *   <li>{@code target_field}: The field to be used for reranking (required)</li>
+ *   <li>{@code remove_target_field}: Whether to remove the target field from the final results (optional, default: false)</li>
+ * </ul>
+ * <p>
+ * Usage example:
+ * <pre>
+ * {
+ *   "rerank": {
+ *     "by_field": {
+ *       "target_field": "document.relevance_score",
```
ylwu-amzn, line 48: Really like such clear and detailed comments. 

Will the _score changed to the value from target_field after reranking?

brianf-aws: Yep it updates `_score` value with the target_field, It was suggested to keep the previous score so there is a onboarded `previous_score` field to keep this info

ylwu-amzn: Got it, can you explain this in the PR description and this java doc?

brianf-aws: Thanks for the feedback you can check commit [92e571a](https://github.com/opensearch-project/neural-search/pull/932/commits/92e571a70c3fa1ec494ad57ccd5836036dd02522) 😄

vibrantvarun:
Will do the second round of review tonight

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,280 @@
...
+     * be deleted. The consequence of this, is having to delete all subsequent empty maps , this is
+     * accounted for by the last check to see that the mapping should be removed.
+     * <hr>
+     * <b>This method assumes that the path to the mapping exists as checked by {@link #validateSearchHits(SearchHit[], ActionListener)}</b>
+     * As such no error cehcking is done in the methods implementing this functionality
+     * @param sourceAsMap the map of maps that contains the <code>targetField</code>
+     * @param keys The keys used to traverse the nested map
+     * @param currentKeyIndex A sentinel to get the current key to look at
+     */
+    private void exploreMapAndRemove(Map<String, Object> sourceAsMap, String[] keys, int currentKeyIndex) {
```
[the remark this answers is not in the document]
brianf-aws: Understood the new commit [ce9bd7d](https://github.com/opensearch-project/neural-search/pull/932/commits/ce9bd7d7e5db8945ecc166aa8857f4bcfe15ad7f) addresses this feedback by making the method use a stack so that it can emulate the recursive solution iteratively!

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
brianf-aws: In the new commit [ce9bd7d](https://github.com/opensearch-project/neural-search/pull/932/commits/ce9bd7d7e5db8945ecc166aa8857f4bcfe15ad7f) we delete this method altogether and just grab the source map from the search hit :)

vibrantvarun:
Add changelog in the PR.

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,303 @@
...
+        }
+
+        // Remove the last key this is guaranteed
+        Tuple<Map<String, Object>, String> currentParentMapWithChild = parentMapChildrenKeyTupleStack.pop();
+        Map<String, Object> parentMap = currentParentMapWithChild.v1();
+        String key = currentParentMapWithChild.v2();
+        parentMap.remove(key);
+
+        // Delete the empty maps inside out using the stack to mock a recursive solution
+        while (!parentMapChildrenKeyTupleStack.isEmpty()) {
```
vibrantvarun, line 201: I think this loop is not needed. JVM garbage collection will automatically delete this stack.

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,303 @@
...
+
+    /**
+     * This helper method is used to initialize the path to take to get to the targetField
+     * to remove as well as the empty maps as the result of the operation.
+     * <hr>
+     * <b>This method assumes that the path to the mapping exists as checked by {@link #validateSearchHits(SearchHit[], ActionListener)}</b>
+     * As such no error checking is done in the methods implementing this functionality
+     * @param sourceAsMap the map of maps that contains the <code>targetField</code>
+     */
+    private void removeTargetFieldFromSource(Map<String, Object> sourceAsMap) {
```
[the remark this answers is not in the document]
vibrantvarun: Move code in "deleteTargetFieldAndEmptyMaps" to this method itself. As logical name of whole operation is removing the target field from source.

martin-gaievski:
@vibrantvarun why we're flipping this request from feature to enhancement? To me that's a clear feature, we're adding functionality and extending interface with new parameters

vibrantvarun:
I thought it is the enhancement of rerank processor which already exist in the project. @martin-gaievski I am fine with either of the label.

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,303 @@
...
+     *     <li>Has a <code>_source</code> mapping</li>
+     *     <li>Has a valid mapping for <code>target_field</code></li>
+     *     <li>That value for the mapping is a valid number</li>
+     * </ul>
+     * When just one of the conditions fail the exception will be thrown to the listener.
+     * @param searchHits from the ByField ReRank Processor
+     * @param listener returns an error to the listener in case on of the conditions fail
+     * @return The status indicating that the SearchHits are in correct form to perform the Rerank
+     */
+    private boolean validateSearchHits(SearchHit[] searchHits, ActionListener<List<Float>> listener) {
```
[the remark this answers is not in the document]
brianf-aws: Good suggestion, thanks and added!

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,303 @@
...
+    /**
+     * Used to get the numeric mapping from the sourcemap using the <code>target_field</code>
+     * <hr>
+     * <b>This method assumes that the path to the mapping exists (and is numerical) as checked by {@link #validateSearchHits(SearchHit[], ActionListener)}</b>
+     * As such no error checking is done in the methods implementing this functionality
+     * @param sourceAsMap the map of maps that contains the <code>targetField</code>
+     * @param targetField the path to take to get the score to replace by
+     * @return The numerical score found using the <code>target_field</code>
+     */
+    private float getScoreFromSourceMap(Map<String, Object> sourceAsMap, String targetField) {
```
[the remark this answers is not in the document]
brianf-aws: Great Suggestion I added this functionality in commit [58152a3](https://github.com/opensearch-project/neural-search/pull/932/commits/58152a3959b243423fd5eb2476c1f88b2a095f74). This now adds a functional interface so future rerankers are able to validate their search response data before performing re rank. All UTs and IT pass

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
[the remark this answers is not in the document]
brianf-aws: Thank you this addressed in [58152a3](https://github.com/opensearch-project/neural-search/pull/932/commits/58152a3959b243423fd5eb2476c1f88b2a095f74). It is now ` removeTargetFieldFromSource(Map<String, Object> sourceAsMap, String targetField)`

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,303 @@
...
+     * </ol>
+     * <hr>
+     * <b>This method assumes that the path to the mapping exists as checked by {@link #validateSearchHits(SearchHit[], ActionListener)}</b>
+     * As such no error checking is done in the methods implementing this functionality
+     * @param sourceAsMap the map of maps that contains the <code>targetField</code>
+     * @param keys The keys used to traverse the nested map
+     * @implNote You can think of this algorithm as a recursive one the base case is deleting the targetField. The recursive case
+     * is going to the next map along with the respective key. Along the way if it finds a map is empty it will delete it
+     */
+    private void deleteTargetFieldAndEmptyMaps(Map<String, Object> sourceAsMap, String[] keys) {
```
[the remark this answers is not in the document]
brianf-aws: Thank you this addressed in [58152a3](https://github.com/opensearch-project/neural-search/pull/932/commits/58152a3959b243423fd5eb2476c1f88b2a095f74).

src/main/java/org/opensearch/neuralsearch/processor/util/processorSearchResponseUtil.java
```diff
@@ -0,0 +1,178 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.Stack;
+
+/**
+ * Utility class for evaluating SearchResponse data. This is useful when you want
+ * to see that the searchResponse is in correct form or if the data you want to extract/edit
+ * from the SearchResponse
+ */
+public class processorSearchResponseUtil {
```
vibrantvarun, line 21: Naming convention. Also these methods are generic and can be used in any processor, not necessarily search response processor.
```suggestion
public class ProcessorUtils {
```

vibrantvarun (commented):
Just one comment regarding naming of processor util class. Otherwise Looks good to me.

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,159 @@
...
+        listener.onResponse(scores);
+    }
+
+    public void byFieldSearchHitValidator(SearchHit hit) {
+        if (!hit.hasSource()) {
+            throw new IllegalArgumentException("There is no source field to be able to perform rerank on hit [" + hit.docId() + "]");
+        }
+
+        Map<String, Object> sourceMap = hit.getSourceAsMap();
+        if (!mappingExistsInSource(sourceMap, targetField)) {
```
mingshl, line 148: when one of the hit missing the field, it will throw exception. Have you test when one of the field is missing and ignore Failure, what is the response?

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
```diff
@@ -0,0 +1,182 @@
...
+     * <b>This method assumes that the path to the mapping exists as checked by {@link #validateRerankCriteria(SearchHit[], SearchHitValidator, ActionListener)}</b>
+     * As such no error checking is done in the methods implementing this functionality
+     * @param sourceAsMap the map of maps that contains the <code>targetField</code>
+     * @param targetField The path to take to remove the targetField
+     * @implNote You can think of this algorithm as a recursive one the base case is deleting the targetField. The recursive case
+     * is going to the next map along with the respective key. Along the way if it finds a map is empty it will delete it
+     */
+    public static void removeTargetFieldFromSource(Map<String, Object> sourceAsMap, String targetField) {
+        Stack<Tuple<Map<String, Object>, String>> parentMapChildrenKeyTupleStack = new Stack<>();
+        String[] keys = targetField.split("\\.");
```
mingshl, line 102: what if a key contains the special character "."?

brianf-aws: I didnt know that was possible the expected behavior would still split it and return a exception since it could not find it

src/main/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactory.java
```diff
@@ -72,6 +72,7 @@ public SearchResponseProcessor create(
                 );
                 return new MLOpenSearchRerankProcessor(description, tag, ignoreFailure, modelId, contextFetchers, clientAccessor);
             case BY_FIELD:
+                boolean DEFAULT_REMOVE_TARGET_FIELD = false;
```
martin-gaievski, line 75: it's a constant, please declare it on top of the class with other static variables.

src/test/java/org/opensearch/neuralsearch/processor/factory/RerankProcessorFactoryTests.java
```diff
@@ -218,5 +229,65 @@ public void testCrossEncoder_whenTooManyDocFields_thenFail() {
             () -> factory.create(Map.of(), TAG, DESC, false, config, pipelineContext)
         );
     }
+    // End of MLOpenSearchRerankProcessor Tests
+
+    // Start of ByFieldRerankProcessor Tests
```
martin-gaievski, line 232: nit: you can make this comment part of the test method name

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,159 @@
...
+        final List<ContextSourceFetcher> contextSourceFetchers
+    ) {
+        super(RerankType.BY_FIELD, description, tag, ignoreFailure, contextSourceFetchers);
+        this.targetField = targetField;
+        this.removeTargetField = removeTargetField;
+        this.keepPreviousScore = keepPreviousScore;
+    }
+
+    @Override
+    public void rescoreSearchResponse(SearchResponse response, Map<String, Object> rerankingContext, ActionListener<List<Float>> listener) {
```
martin-gaievski, line 104: please make all arguments `final`, that's what we do at least for public methods

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,159 @@
...
+            if (removeTargetField) {
+                removeTargetFieldFromSource(sourceAsMap, targetField);
+            }
+
+            try {
+                XContentBuilder builder = XContentBuilder.builder(XContentType.JSON.xContent());
+                BytesReference sourceMapAsBytes = BytesReference.bytes(builder.map(sourceAsMap));
+                hit.sourceRef(sourceMapAsBytes);
+            } catch (IOException e) {
+                listener.onFailure(new RuntimeException(e));
```
martin-gaievski, line 134: what will happen in this of this exception, does it go all the way up to the end user? Please log the exception to the log too, depending on the severity it can be up to the "error" log level.

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,159 @@
...
+                return;
+            }
+        }
+
+        listener.onResponse(scores);
+    }
+
+    public void byFieldSearchHitValidator(SearchHit hit) {
+        if (!hit.hasSource()) {
+            throw new IllegalArgumentException("There is no source field to be able to perform rerank on hit [" + hit.docId() + "]");
```
martin-gaievski, line 144: use String.format for this message, and also log to server logs.

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,159 @@
...
+            } catch (IOException e) {
+                listener.onFailure(new RuntimeException(e));
+                return;
+            }
+        }
+
+        listener.onResponse(scores);
+    }
+
+    public void byFieldSearchHitValidator(SearchHit hit) {
```
martin-gaievski, line 142: `final` for argument please

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,159 @@
...
+        if (!hit.hasSource()) {
+            throw new IllegalArgumentException("There is no source field to be able to perform rerank on hit [" + hit.docId() + "]");
+        }
+
+        Map<String, Object> sourceMap = hit.getSourceAsMap();
+        if (!mappingExistsInSource(sourceMap, targetField)) {
+            throw new IllegalArgumentException("The field to rerank [" + targetField + "] is not found at hit [" + hit.docId() + "]");
```
martin-gaievski, line 149: same as above, please use String.format. 
If `targetField` is defined by user please don't echo it in exception message, it's not secure. Put it to the log instead and keep message to the end-user simple

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        Optional<Object> val = getValueFromSource(sourceMap, targetField);
+
+        if (!(val.get() instanceof Number)) {
+            throw new IllegalArgumentException("The field mapping to rerank [" + targetField + ": " + val.get() + "] is a not Numerical");
```
martin-gaievski, line 155: String.format please

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        Map<String, Object> sourceMap = hit.getSourceAsMap();
+        if (!mappingExistsInSource(sourceMap, targetField)) {
+            throw new IllegalArgumentException("The field to rerank [" + targetField + "] is not found at hit [" + hit.docId() + "]");
+        }
+
+        Optional<Object> val = getValueFromSource(sourceMap, targetField);
+
+        if (!(val.get() instanceof Number)) {
```
martin-gaievski, line 154: how do we handle strings that have numerical value, like "2.0", that should be parsable to number?

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
mingshl: it's possible to have dot in field name, try this use case

```
PUT /sample_index/_doc/1
{
  "first.name": "Example",
  "price": 29.99,
  "description": "To be or not to be, that is the question"
}
GET  /sample_index/_doc/1
{
  "_index": "sample_index",
  "_id": "1",
  "_version": 1,
  "_seq_no": 0,
  "_primary_term": 1,
  "found": true,
  "_source": {
    "first.name": "Example",
    "price": 29.99,
    "description": "To be or not to be, that is the question"
  }
}
```

mingshl: if you are not using JsonPath, this is an example in the IngestDocument which has a `HasField` method https://github.com/opensearch-project/OpenSearch/blob/456ca97f077929ba51d53dc8578b9b770e3a2417/server/src/main/java/org/opensearch/ingest/IngestDocument.java#L233

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
brianf-aws: Currently it sees that as a string but I can implement that if it would be favorable

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
brianf-aws: Yeah this is one of those lines were I wasnt able to test If you look at other processors such as ml_inference or [sort_processor](https://github.com/opensearch-project/OpenSearch/blob/456ca97f077929ba51d53dc8578b9b770e3a2417/modules/search-pipeline-common/src/main/java/org/opensearch/search/pipeline/common/SortResponseProcessor.java#L159) they dont have a problem doing whats in the try block because we cant add a throws exception in the rescoreSearchResponse method. I believe this is because the exception for the bad SearchResponse would occur before it hits this method.

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
brianf-aws: Hey, thanks for the test case I just ran it and it still threw the exception. I believe this is because we are working with in the search response level If this were in a end to end workflow within search pipelines then the original search response would be returned as it states in [SearchPipeline](https://github.com/opensearch-project/OpenSearch/blob/c20458512ff4e68f359de32bfad91a6a7487f8a1/server/src/main/java/org/opensearch/search/pipeline/Pipeline.java#L238) class

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
brianf-aws: I think for now i will leave it how it is currently but I do see that as an enhancement simply because in the [doc pr](https://github.com/opensearch-project/documentation-website/pull/8593) I stated how the user should use the argument

src/main/java/org/opensearch/neuralsearch/processor/rerank/ByFieldRerankProcessor.java
martin-gaievski: what we do have now is fine, if it's not a requirement then we're doing right thing now, and we can add that parsing logic later

martin-gaievski (approved):
looks good to me, thank you

vibrantvarun (approved):
LGTM thanks.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 2.x](https://github.com/opensearch-project/neural-search/compare/2.18...2.x)
 ### Features
+- Introduces ByFieldRerankProcessor for second level reranking on documents (This is done locally) ([#932](https://github.com/opensearch-project/neural-search/pull/932))
```
vibrantvarun, line 19: @martin-gaievski should we add it in the changelog of main branch now as @brianf-aws already added in the release notes of 2.18?

vibrantvarun: Hey @brianf-aws don't add change in changelog.md as the release notes are already cut.

vibrantvarun: Just keep the change in release notes 2.18. file.

brianf-aws: Understood will remove from changelog

martin-gaievski: > @martin-gaievski should we add it in the changelog of main branch now as @brianf-aws already added in the release notes of 2.18?

we can skip changelog if we do have entry in the release notes

release-notes/opensearch-neural-search.release-notes-2.18.0.0.md
```diff
@@ -3,6 +3,9 @@
 
 Compatible with OpenSearch 2.18.0
 
+### Features
+- Introduces ByFieldRerankProcessor for second level reranking on documents ([#932](https://github.com/opensearch-project/neural-search/pull/932))
+
```
vibrantvarun, line 8: Remove this extra line 8. In the final formatting it will add 2 lines.

martin-gaievski: fixed it in main while merging PR https://github.com/opensearch-project/neural-search/pull/956

Issue #280, opened 2023-09-04
[FEATURE] Implement pagination for Hybrid Search


martin-gaievski:
### Is your feature request related to a problem?
Current implementation of Hybrid search doesn't have support for pagination, meaning all results are returned "at once". That is standard for many queries in OpenSearch and it's expected that Hybrid Search supports it, https://opensearch.org/docs/latest/search-plugins/searching-data/paginate/.

### What solution would you like?
It should be possible to define "from" and positions in as part of the Hybrid query, and results should have (from + size) number of records, starting from "from" position. Standard syntax should be fine in this case:

```
{
   "from": 20,
   "size": 10,
   "query": {
       "hybrid": [
           {},// First Query
           {} // Second Query
           ..... // Other Queries
       ] 
   }
}
```

### What alternatives have you considered?
It's possible to set higher size and then throw first X elements, but that is required extra processing logic on a client size and is not very optimal.

ankitas3:
When can we expect this in OpenSearch?

jackh-ncl:
@martin-gaievski @ankitas3 Just double checking, but `from` and `size` already seem to be working for me with Hybrid search while using Opensearch `2.11`. Has it been implemented since this issue was raised?

martin-gaievski:
@jackh-ncl it has not been implemented yet. With existing code it may work, but not for every scenario and when it works it does it not in an optimal way.

jackh-ncl:
Aha thank you for confirming, yeah I have now noticed that I seem to consistently end up with a total hits value of 5 * the size parameter with the query I'm running.

benmcginnis:
any chance we can get this in 2.14?

brandon-carag:
Is anyone from opensearch able to provide any broad details around when pagination will be made available?

qmauret:
+1 for this feature request

mkerimyilmaz:
@martin-gaievski Can you please clarify why it doesn't work optimally? 

> @jackh-ncl it has not been implemented yet. With existing code it may work, but not for every scenario and when it works it does it not in an optimal way.

brandon-carag:
Hi @vamshin and @martin-gaievski,

Just to briefly restate the case here, pagination is a critical feature for many users adopting hybrid search.  There has been significant community interest in this thread (at least 20 distinct users) that also point to strong user demand.  

[...]

PR #963, opened 2024-10-23
[Feature] Pagination in hybrid query

vibrantvarun:
### Description
This PR contains changes for enabling support for pagination in hybrid query.
The highlight of this PR are
1. Introduction of a new parameter "**pagination_depth**" to set a reference of hybrid query search results on which pagination can be applied.
2. Handling of **single shard scenario** where fetch phase can run before the normalization process.
3. Handling of **from** parameter conditions in Normalization processor.
4. Disabling scroll operation in hybrid query.

### Related Issues
https://github.com/opensearch-project/neural-search/issues/280

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
@@ -177,14 +177,23 @@ public class NormalizationProcessorWorkflow {
         final List<QuerySearchResult> querySearchResults = combineScoresDTO.getQuerySearchResults();
         final List<CompoundTopDocs> queryTopDocs = getCompoundTopDocs(combineScoresDTO, querySearchResults);
         final Sort sort = combineScoresDTO.getSort();
+        final int from = querySearchResults.get(0).from();
+        int totalScoreDocsCount = 0;
         for (int index = 0; index < querySearchResults.size(); index++) {
             QuerySearchResult querySearchResult = querySearchResults.get(index);
             CompoundTopDocs updatedTopDocs = queryTopDocs.get(index);
+            totalScoreDocsCount += updatedTopDocs.getScoreDocs().size();
             TopDocsAndMaxScore updatedTopDocsAndMaxScore = new TopDocsAndMaxScore(
                 buildTopDocs(updatedTopDocs, sort),
                 maxScoreForShard(updatedTopDocs, sort != null)
             );
             querySearchResult.topDocs(updatedTopDocsAndMaxScore, querySearchResult.sortValueFormats());
+        }
+
+        if (from > 0 && from > totalScoreDocsCount) {
+            throw new IllegalArgumentException(
+                String.format(Locale.ROOT, "Reached end of search result, increase pagination_depth value to see more results")
+            );
         }
     }
 
--- a/src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
@@ -19,8 +19,9 @@ import org.apache.lucene.search.TopDocs;
 import org.apache.lucene.search.Sort;
 import org.apache.lucene.search.TopFieldDocs;
 import org.apache.lucene.search.FieldDoc;
+import org.opensearch.action.search.SearchPhaseContext;
 import org.opensearch.common.lucene.search.TopDocsAndMaxScore;
-import org.opensearch.neuralsearch.processor.combination.CombineScoresDto;
+import org.opensearch.neuralsearch.processor.dto.CombineScoresDto;
 import org.opensearch.neuralsearch.processor.combination.ScoreCombinationTechnique;
 import org.opensearch.neuralsearch.processor.combination.ScoreCombiner;
 import org.opensearch.neuralsearch.processor.explain.CombinedExplanationDetails;
@@ -56,17 +57,16 @@ public class NormalizationProcessorWorkflow {
 
     /**
      * Start execution of this workflow
-     * @param querySearchResults input data with QuerySearchResult from multiple shards
-     * @param normalizationTechnique technique for score normalization
-     * @param combinationTechnique technique for score combination
+     * @param normalizationExecuteDto contains querySearchResults input data with QuerySearchResult
+     * from multiple shards, fetchSearchResultOptional, normalizationTechnique technique for score normalization
+     *  combinationTechnique technique for score combination, searchPhaseContext.
      */
     public void execute(
         final List<QuerySearchResult> querySearchResults,
         final Optional<FetchSearchResult> fetchSearchResultOptional,
         final ScoreNormalizationTechnique normalizationTechnique,
         final ScoreCombinationTechnique combinationTechnique,
-        final int fromValueForSingleShard,
-        final boolean isSingleShard
+        final int fromValueForSingleShard
     ) {
         NormalizationProcessorWorkflowExecuteRequest request = NormalizationProcessorWorkflowExecuteRequest.builder()
             .querySearchResults(querySearchResults)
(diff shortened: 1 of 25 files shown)


src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -461,6 +471,42 @@ private ReduceableSearchResult reduceSearchResults(final List<ReduceableSearchRe
...
+     * @param searchContext search context that contains pagination depth
+     * @return results size to collected
+     */
+    private static int getSubqueryResultsRetrievalSize(final SearchContext searchContext) {
+        int paginationDepth;
+        HybridQuery hybridQuery;
+        Query query = searchContext.query();
+        if (query instanceof BooleanQuery) {
+            BooleanQuery booleanQuery = (BooleanQuery) query;
+            hybridQuery = (HybridQuery) booleanQuery.clauses().get(0).getQuery();
```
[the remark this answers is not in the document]
vibrantvarun: I have added the comment on top why it is written like that. Basically in case of nested fields and alias filter, hybrid query gets wrapped under bool query.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -461,6 +471,42 @@ private ReduceableSearchResult reduceSearchResults(final List<ReduceableSearchRe
...
+    /**
+     * Get maximum subquery results count to be collected from each shard.
+     * @param searchContext search context that contains pagination depth
+     * @return results size to collected
+     */
+    private static int getSubqueryResultsRetrievalSize(final SearchContext searchContext) {
+        int paginationDepth;
+        HybridQuery hybridQuery;
+        Query query = searchContext.query();
+        if (query instanceof BooleanQuery) {
```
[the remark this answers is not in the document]
vibrantvarun: I have added the comment on top why it is written like that. Basically in case of nested fields and alias filter, hybrid query gets wrapped under bool query.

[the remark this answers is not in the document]
martin-gaievski: simple case with nested fields is taken care of in [QueryPhaseSearcher class](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java#L87-L103). This logic doesn't belong here, unless this is a special scenario that we missed in that extract query method I mentioned above.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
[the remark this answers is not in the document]
martin-gaievski: see my previous comment, this should be handled in [here](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java#L87-L103)

src/main/java/org/opensearch/neuralsearch/processor/dto/CombineScoresDto.java
```diff
@@ -29,4 +31,6 @@ public class CombineScoresDto {
     private List<QuerySearchResult> querySearchResults;
     @Nullable
     private Sort sort;
+    private int fromValueForSingleShard;
+    private boolean isFetchResultsPresent;
```
martin-gaievski, line 35: it doesn't look right to put this field here, it's not related to combination. please find alternative solution

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -212,14 +237,21 @@ private void updateOriginalFetchResults(
 
         QuerySearchResult querySearchResult = querySearchResults.get(0);
         TopDocs topDocs = querySearchResult.topDocs().topDocs;
+
+        // When normalization process will execute before the fetch phase, then from =0 is applicable.
+        // When normalization process runs after fetch phase, then search hits already fetched. Therefore, use the from value sent in the
+        // search request.
         // iterate over the normalized/combined scores, that solves (1) and (3)
-        SearchHit[] updatedSearchHitArray = Arrays.stream(topDocs.scoreDocs).map(scoreDoc -> {
+        SearchHit[] updatedSearchHitArray = new SearchHit[topDocs.scoreDocs.length - fromValueForSingleShard];
```
martin-gaievski, line 245: please pull `topDocs.scoreDocs.length - fromValueForSingleShard` expression to a variable and give it meaningful name

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
```diff
@@ -31,20 +32,25 @@
...
      * @param filterQueries list of filters that will be applied to each sub query. Each filter from the list is added as bool "filter" clause. If this is null sub queries will be executed as is
      */
-    public HybridQuery(final Collection<Query> subQueries, final List<Query> filterQueries) {
+    public HybridQuery(final Collection<Query> subQueries, final List<Query> filterQueries, Integer paginationDepth) {
         Objects.requireNonNull(subQueries, "collection of queries must not be null");
         if (subQueries.isEmpty()) {
             throw new IllegalArgumentException("collection of queries must not be empty");
         }
+        if (paginationDepth != null && paginationDepth == 0) {
+            throw new IllegalArgumentException("pagination depth must not be zero");
```
martin-gaievski, line 52: change to `pagination_depth` in error message to signify that this is a parameter

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -97,6 +109,9 @@ protected void doXContent(XContentBuilder builder, Params params) throws IOExcep
             queryBuilder.toXContent(builder, params);
         }
         builder.endArray();
+        if (isClusterOnOrAfterMinReqVersionForPaginationInHybridQuery()) {
+            builder.field(DEPTH_FIELD.getPreferredName(), paginationDepth == null ? DEFAULT_PAGINATION_DEPTH : paginationDepth);
```
martin-gaievski, line 113: when we're doing today's logic? from this statement we'll either get user provided value or default which is 10. Today's logic is equivalent to PD equals to 1, I'm not exactly sure, but definitely not 10.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -297,6 +326,15 @@ private Collection<Query> toQueries(Collection<QueryBuilder> queryBuilders, Quer
         return queries;
     }
 
+    private static void validatePaginationDepth(Integer paginationDepth) {
+        if (paginationDepth != null
+            && (paginationDepth < LOWER_BOUND_OF_PAGINATION_DEPTH || paginationDepth > UPPER_BOUND_OF_PAGINATION_DEPTH)) {
+            throw new IllegalArgumentException(
+                String.format(Locale.ROOT, "Pagination depth should lie in the range of 1-1000. Received: %s", paginationDepth)
```
martin-gaievski, line 333: please came up with better error message. The range is static here, but that's different in the logic where you have variables for lower and upper bounds, and we can drop the user-provided value as that's potentially unsafe, actual value should be obvious to user from the search request

src/test/java/org/opensearch/neuralsearch/query/HybridQueryIT.java
```diff
@@ -793,46 +796,221 @@ public void testConcurrentSearchWithMultipleSlices_whenMultipleShardsIndex_thenS
...
+
+        org.hamcrest.MatcherAssert.assertThat(
+            responseException.getMessage(),
+            allOf(containsString("Number of sub-queries exceeds maximum supported by [hybrid] query"))
+        );
+
+    }
+
+    @SneakyThrows
+    public void testPaginationOnSingleShard_whenConcurrentSearchEnabled_thenSuccessful() {
```
martin-gaievski, line 843: Looks like you're basically put all tests for pagination to integ test. Can this be moved to unit tests, so we have only minimal increase in new integ tests. If that's not possible please consider moving them to a new class, this one because way overloaded, about 1000 lines.

Issue #1010, opened 2024-12-12
[BUG] NeuralQueryBuilder doEquals() and doHashCode() don't consider all the parameters that can impact the query result


bzhangam:
### What is the bug?
The doEquals() and doHashCode() methods in NeuralQueryBuilder fail to account for all parameters that influence query results. This oversight can lead to scenarios where two queries are mistakenly identified as equal, even though their outcomes differ.

### How can one reproduce the bug?

Whenever two NeuralQueryBuilder instances have differing parameters that are not accounted for in the doEquals() and doHashCode() methods, these methods will incorrectly identify the instances as equal.

Below is the current implementation:

   
    @Override
    protected boolean doEquals(NeuralQueryBuilder obj) {
        if (this == obj) return true;
        if (obj == null || getClass() != obj.getClass()) return false;
        EqualsBuilder equalsBuilder = new EqualsBuilder();
        equalsBuilder.append(fieldName, obj.fieldName);
        equalsBuilder.append(queryText, obj.queryText);
        equalsBuilder.append(modelId, obj.modelId);
        equalsBuilder.append(k, obj.k);
        equalsBuilder.append(filter, obj.filter);
        return equalsBuilder.isEquals();
    }

    @Override
    protected int doHashCode() {
        return new HashCodeBuilder().append(fieldName).append(queryText).append(modelId).append(k).toHashCode();
    }

But it has [more parameters](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java#L106).

### What is the expected behavior?
doEquals() and doHashCode() methods should account for all the parameters that can impact the query result.

### What is your host/environment?
N/A

### Do you have any screenshots?
N/A

### Do you have any additional context?
If these methods are used for OpenSearch caching, it could lead to incorrect results being returned when the difference between a new query and an old query lies solely in the parameters that are not considered. Further investigation is required to understand how these two functions are utilized within OpenSearch.

bzhangam:
Looking into this issue.

martin-gaievski:
that's interesting finding, thanks fro reporting it. There is a `filter` field that is missing from the hashCode method. Do you know scenario (in core or plugin specific) where we see effect of this?

bzhangam:
[...]

PR #1045, opened 2024-12-27
Correct NeuralQueryBuilder doEquals() and doHashCode().

bzhangam:
### Description
NeuralQueryBuilder doEquals() and doHashCode() don't consider all the parameters so correct them to ensure those functions can work correctly if they are used for cache function.

Also fix the issue that we don't handle the queryImage properly. We should handle it similar to the queryText.

### Related Issues
Resolves #[1010](https://github.com/opensearch-project/neural-search/issues/1010)

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Address inconsistent scoring in hybrid query results ([#998](https://github.com/opensearch-project/neural-search/pull/998))
 - Fix bug where ingested document has list of nested objects ([#1040](https://github.com/opensearch-project/neural-search/pull/1040))
 - Fixed document source and score field mismatch in sorted hybrid queries ([#1043](https://github.com/opensearch-project/neural-search/pull/1043))
+- Update NeuralQueryBuilder doEquals() and doHashCode() to cater the missing parameters information ([#1045](https://github.com/opensearch-project/neural-search/pull/1045)).
 - Fix bug where embedding is missing when ingested document has "." in field name, and mismatches fieldMap config ([#1062](https://github.com/opensearch-project/neural-search/pull/1062))
 ### Infrastructure
 ### Documentation
--- a/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
@@ -18,6 +18,7 @@ import static org.opensearch.neuralsearch.processor.TextImageEmbeddingProcessor.
 import static org.opensearch.neuralsearch.processor.TextImageEmbeddingProcessor.INPUT_TEXT;
 
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
@@ -25,7 +26,6 @@ import java.util.function.Supplier;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.builder.EqualsBuilder;
-import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.apache.lucene.search.Query;
 import org.opensearch.common.SetOnce;
 import org.opensearch.core.ParseField;
@@ -76,8 +76,7 @@ public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder>
     @VisibleForTesting
     static final ParseField QUERY_TEXT_FIELD = new ParseField("query_text");
 
-    @VisibleForTesting
-    static final ParseField QUERY_IMAGE_FIELD = new ParseField("query_image");
+    public static final ParseField QUERY_IMAGE_FIELD = new ParseField("query_image");
 
     public static final ParseField MODEL_ID_FIELD = new ParseField("model_id");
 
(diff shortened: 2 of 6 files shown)


src/test/java/org/opensearch/neuralsearch/query/HybridQueryBuilderTests.java
```diff
@@ -469,8 +469,7 @@ public void testStreams_whenWrittingToStream_thenSuccessful() {
         NeuralQueryBuilder neuralQueryBuilder = new NeuralQueryBuilder().fieldName(VECTOR_FIELD_NAME)
             .queryText(QUERY_TEXT)
             .modelId(MODEL_ID)
-            .k(K)
-            .vectorSupplier(TEST_VECTOR_SUPPLIER);
```
bzhangam, line 473: Removed the vectorSupplier since our doWriteTo function cannot write a function to the StreamOutput. Besides this field is calculated during the runtime so we should be good to not have it in this test.

junqiu-lei:
Please check the failed CIs and update the changelog file.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -146,7 +147,8 @@ public NeuralQueryBuilder(StreamInput in) throws IOException {
     @Override
     protected void doWriteTo(StreamOutput out) throws IOException {
         out.writeString(this.fieldName);
-        out.writeString(this.queryText);
+        out.writeOptionalString(this.queryText);
```
bzhangam, line 150: Since either queryText and queryImage can be null we should use optional string here.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -381,15 +389,38 @@ protected boolean doEquals(NeuralQueryBuilder obj) {
...
         equalsBuilder.append(filter, obj.filter);
+        equalsBuilder.append(methodParameters, obj.methodParameters);
+        equalsBuilder.append(rescoreContext, obj.rescoreContext);
         return equalsBuilder.isEquals();
     }
 
     @Override
     protected int doHashCode() {
-        return new HashCodeBuilder().append(fieldName).append(queryText).append(modelId).append(k).toHashCode();
+        return new HashCodeBuilder().append(fieldName)
```
martin-gaievski, line 407: Please use `Objects.hash` method, it's widely used in other OS repos, you can check some example of usage here:  https://github.com/opensearch-project/OpenSearch/blob/845fbfa10407f3264e6aab8812eff4ef0ad8be24/server/src/main/java/org/opensearch/search/aggregations/metrics/TopHitsAggregationBuilder.java#L919-L936

martin-gaievski:
> Please check the failed CIs and update the changelog file.

Blocking issue is fixed, you can use CI after rebasing on main branch

martin-gaievski:
CI actions are failing for the changes in this PR, please check:
```
 -org.opensearch.neuralsearch.query.HybridQueryBuilderTests.testHashAndEquals_whenSameOrIdenticalObject_thenReturnEqual
 - org.opensearch.neuralsearch.query.NeuralQueryBuilderTests.testHashAndEquals
 ```

**Review by junqiu-lei — APPROVED** [review]

File: CHANGELOG.md [hunk 4]
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Set neural-search plugin 3.0.0 baseline JDK version to JDK-2 ([#838](https://github.com/opensearch-project/neural-search/pull/838))
 ### Bug Fixes
 - Fix for nested field missing sub embedding field in text embedding processor ([#913](https://github.com/opensearch-project/neural-search/pull/913))
+- Correct NeuralQueryBuilder doEquals() and doHashCode().
```
vibrantvarun, line 12:
```suggestion
- Update NeuralQueryBuilder doEquals() and doHashCode() to cater the missing parameters information ([#1045](https://github.com/opensearch-project/neural-search/pull/1045)).
```

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -236,7 +236,8 @@ public static NeuralQueryBuilder.Builder builder() {
     public NeuralQueryBuilder(StreamInput in) throws IOException {
         super(in);
         this.fieldName = in.readString();
-        this.queryText = in.readString();
+        this.queryText = in.readOptionalString();
```
heemin32, line 239: This is going to fail in bwc test. 
For example, old version of node will write as `out.writeString(this.queryText)`, and the new node will read as `this.queryText = in.readOptionalString()` which is incompatible.

bzhangam: Is it possible that we write a query into the stream in an old version of the node and then pass it to a new version of the node? If that's the case how can other field work? 
e.g. Old version of the node `out.writeString(this.modelId)` and then new version of the node will do `this.modelId = in.readOptionalString()`.

heemin32: That is possible during rolling upgrade. Because that, we add min version check of cluster to keep backward compatibility.

heemin32 (approved):

Issue #1025, opened 2024-12-17
[Code Improvement] Create a builder constructor for Neural Query Builder


bzhangam:
### Is your feature request related to a problem?
No

### What solution would you like?
Create a builder constructor for Neural Query Builder similar to [KNNQueryBuilder](https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java#L142).
1. Give us the flexibility that we don't need to pass all parameters when some optional ones are not needed when we want to construct a new Neural Query Builder.
2. Adding a new parameter will not break the existing tests which doesn't care about it.

### What alternatives have you considered?
N/A

### Do you have any additional context?
N/A

will-hwang:
assign this to me please

PR #1047, opened 2024-12-30
add support for builder constructor in neural query builder

will-hwang:
### Description
Adds support for builder constructor for Neural Query Builder

This PR adds support for builder constructor with `@Builder` lombok annotation in `NeuralQueryBuilder`

### Related Issues

#1025
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
@@ -23,7 +23,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.function.Supplier;
 
-import lombok.Builder;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.builder.EqualsBuilder;
 import org.apache.commons.lang.builder.HashCodeBuilder;
@@ -32,6 +31,7 @@ import org.opensearch.common.SetOnce;
 import org.opensearch.core.ParseField;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.common.ParsingException;
+import org.opensearch.core.common.Strings;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
 import org.opensearch.core.xcontent.XContentBuilder;
@@ -68,9 +68,8 @@ import lombok.extern.log4j.Log4j2;
 @Getter
 @Setter
 @Accessors(chain = true, fluent = true)
-@Builder(toBuilder = true)
-@NoArgsConstructor
-@AllArgsConstructor
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
 public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder> implements ModelInferenceQueryBuilder {
 
     public static final String NAME = "neural";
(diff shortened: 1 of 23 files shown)


src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -23,6 +23,7 @@
 import java.util.Objects;
 import java.util.function.Supplier;
 
+import lombok.Builder;
```
bzhangam, line 26: I think we want to validate the NeuralQueryBuilder before we create it like what we are doing in the KNNQueryBuilder. And to enforce this we will also need to either remove @NoArgsConstructor
@AllArgsConstructor or make them private.

Currently I think the builder is mainly used for our internal testing but I think it's good to enforce that only valid NeuralQueryBuilder can be created through our builder in case in future people accidentally use it create some invalid NeuralQueryBuilder in our business logic.

bzhangam: Besides I think currently we are using AllArgsConstructor to create NeuralQueryBuilder for some tests can we replace them with the builder you add? In this way if we need to add a new field we don't need to modify the tests that don't care about the new field.

junqiu-lei: +1 to @bzhangam's comment. You can have reference from kNN's query builder: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java#L142

junqiu-lei:
Please sign off your commits and add this PR to [CHANGELOG](https://github.com/opensearch-project/neural-search/blob/main/CHANGELOG.md).

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -108,6 +109,130 @@ public static void initialize(MLCommonsClientAccessor mlClient) {
...
+        }
+
+        public Builder rescoreContext(RescoreContext rescoreContext) {
+            this.rescoreContext = rescoreContext;
+            return this;
+        }
+
+        public NeuralQueryBuilder build() {
+            validate();
+            int k = this.k == null ? 0 : this.k;
```
bzhangam, line 205: I think we should use the DEFAULT_K here.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -108,6 +109,130 @@ public static void initialize(MLCommonsClientAccessor mlClient) {
...
+                minScore,
+                expandNested,
+                vectorSupplier,
+                filter,
+                methodParameters,
+                rescoreContext
+            ).boost(boost).queryName(queryName);
+        }
+
+        private void validate() {
```
bzhangam, line 222: The logic in this function is kind duplicated with what we have in fromXContent. Can we try to avoid duplication?

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
will-hwang: addressed in the newest commit

will-hwang:
adding comment for assignment

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
junqiu-lei: It also need verify the query type.https://github.com/opensearch-project/neural-search/blob/b98e6d02570fff45d6b1e50f635b5c5f55d5e7b4/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java#L525

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
will-hwang: done

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
will-hwang: refactored to remove duplication and added validation for KNN query type

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -108,6 +108,126 @@ public static void initialize(MLCommonsClientAccessor mlClient) {
     private Map<String, ?> methodParameters;
     private RescoreContext rescoreContext;
 
+    /**
+     * A custom builder class to enforce valid Neural Query Builder instance by validating the required fields are initialized
```
martin-gaievski, line 112: nit: please fix the wording in this comment

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -108,6 +108,126 @@ public static void initialize(MLCommonsClientAccessor mlClient) {
...
+
+        public Builder rescoreContext(RescoreContext rescoreContext) {
+            this.rescoreContext = rescoreContext;
+            return this;
+        }
+
+        public NeuralQueryBuilder build() {
+            validateQueryParameters(fieldName, queryText, queryImage);
+            int k = this.k == null ? DEFAULT_K : this.k;
+            boolean queryTypeIsProvided = validateKNNQueryType(k, maxDistance, minScore);
```
martin-gaievski, line 205: why we're adding this logic here, it wasn't part of the original constructor?

martin-gaievski (commented):
Why there are no changes in any of the callers? If we made both existing constructors private then I assume we should replace calls to that constructor(s) with the new builder class

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
will-hwang: As part of this change, we wanted to add validation before instantiation so ensure that only valid `NeuralSearchBuilder` can be instantiated. See previous comment here: https://github.com/opensearch-project/neural-search/pull/1047#discussion_r1899854693

Let me know your thoughts!

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
will-hwang: okay, will make wording better

will-hwang:
> Why there are no changes in any of the callers? If we made both existing constructors private then I assume we should replace calls to that constructor(s) with the new builder class

@martin-gaievski All the callers have been changed to the new builder class. See the changes in all 14 files

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
martin-gaievski: I'm rely on @junqiu-lei expertise here, I'm good if he's good

martin-gaievski:
> > Why there are no changes in any of the callers? If we made both existing constructors private then I assume we should replace calls to that constructor(s) with the new builder class
> 
> @martin-gaievski All the callers have been changed to the new builder class. See the changes in all 14 files

Those are all test classes; I mean classes related to application logic. I do see one reference that hasn't been changed (https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java#L112). Do you have an idea why that hasn't been changed after you made the constructor private?

will-hwang:
> > > Why there are no changes in any of the callers? If we made both existing constructors private then I assume we should replace calls to that constructor(s) with the new builder class
> > 
> > 
> > @martin-gaievski All the callers have been changed to the new builder class. See the changes in all 14 files
> 
> Those are all test classes; I mean classes related to application logic. I do see one reference that hasn't been changed (https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java#L112). Do you have an idea why that hasn't been changed after you made the constructor private?

As per `QuerySpec` specification:

```
public QuerySpec(     
    String name,
    org. opensearch. core. common. io. stream. Writeable. Reader<T> reader,
    org. opensearch. index. query. QueryParser<T> parser )
}
```
the parameter expects a Reader<T> type, not a NeuralSearchQuery instance. I believe this is why compiler allows `NeuralQueryBuilder::new`. I've changed the callers to the new builder class in other classes, including in flaky test suite.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
junqiu-lei: The CI is failure in radial integ test due to this change:
```
NeuralQueryIT > testQueryWithBoostAndImageQueryAndRadialQuery FAILED
    java.lang.IllegalArgumentException: Only one of k, max_distance, or min_score can be provided
        at __randomizedtesting.SeedInfo.seed([D0122DFF0BE48EA8:885F5A920D6E8D5B]:0)
        at org.opensearch.neuralsearch.query.NeuralQueryBuilder.validateKNNQueryType(NeuralQueryBuilder.java:540)
        at org.opensearch.neuralsearch.query.NeuralQueryBuilder$Builder.build(NeuralQueryBuilder.java:205)
        at org.opensearch.neuralsearch.query.NeuralQueryIT.testQueryWithBoostAndImageQueryAndRadialQuery(NeuralQueryIT.java:151)
```

heemin32:
> > > Why there are no changes in any of the callers? If we made both existing constructors private then I assume we should replace calls to that constructor(s) with the new builder class
> > 
> > 
> > @martin-gaievski All the callers have been changed to the new builder class. See the changes in all 14 files
> 
> Those are all test classes; I mean classes related to application logic. I do see one reference that hasn't been changed (https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java#L112). Do you have an idea why that hasn't been changed after you made the constructor private?

The `NeuralQueryBuilder::new` mapped to `public NeuralQueryBuilder(StreamInput in)`
https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java#L117

heemin32 (dismissed):

martin-gaievski (approved):
Looks good to me

heemin32 (approved):

PR #1048, opened 2024-12-31
Pagination in Hybrid query

vibrantvarun:
### Description
This PR contains changes for enabling support for pagination in hybrid query.
The highlight of this PR are
1. Introduction of a new parameter "**pagination_depth**" to set a reference of hybrid query search results on which pagination can be applied.
2. Handling of **single shard scenario** where fetch phase can run before the normalization process.
3. Handling of **from** parameter conditions in Normalization processor.
4. Disabling scroll operation in hybrid query.

### Related Issues
https://github.com/opensearch-project/neural-search/issues/280

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 2.x](https://github.com/opensearch-project/neural-search/compare/2.18...2.x)
 ### Features
+- Pagination in Hybrid query ([#963](https://github.com/opensearch-project/neural-search/pull/963))
 ### Enhancements
 - Explainability in hybrid query ([#970](https://github.com/opensearch-project/neural-search/pull/970))
 - Support new knn query parameter expand_nested ([#1013](https://github.com/opensearch-project/neural-search/pull/1013))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 2.x](https://github.com/opensearch-project/neural-search/compare/2.18...2.x)
 ### Features
-- Pagination in Hybrid query ([#963](https://github.com/opensearch-project/neural-search/pull/963))
+- Pagination in Hybrid query ([#1048](https://github.com/opensearch-project/neural-search/pull/1048))
 ### Enhancements
 - Explainability in hybrid query ([#970](https://github.com/opensearch-project/neural-search/pull/970))
 - Support new knn query parameter expand_nested ([#1013](https://github.com/opensearch-project/neural-search/pull/1013))
--- a/src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
@@ -19,6 +19,7 @@ import org.apache.lucene.search.TopDocs;
 import org.apache.lucene.search.Sort;
 import org.apache.lucene.search.TopFieldDocs;
 import org.apache.lucene.search.FieldDoc;
+import org.opensearch.action.search.SearchPhaseContext;
 import org.opensearch.common.lucene.search.TopDocsAndMaxScore;
 import org.opensearch.neuralsearch.processor.combination.CombineScoresDto;
 import org.opensearch.neuralsearch.processor.combination.ScoreCombinationTechnique;
@@ -64,7 +65,8 @@ public class NormalizationProcessorWorkflow {
         final List<QuerySearchResult> querySearchResults,
         final Optional<FetchSearchResult> fetchSearchResultOptional,
         final ScoreNormalizationTechnique normalizationTechnique,
-        final ScoreCombinationTechnique combinationTechnique
+        final ScoreCombinationTechnique combinationTechnique,
+        final SearchPhaseContext searchPhaseContext
     ) {
         NormalizationProcessorWorkflowExecuteRequest request = NormalizationProcessorWorkflowExecuteRequest.builder()
             .querySearchResults(querySearchResults)
@@ -72,17 +74,21 @@ public class NormalizationProcessorWorkflow {
             .normalizationTechnique(normalizationTechnique)
             .combinationTechnique(combinationTechnique)
             .explain(false)
+            .searchPhaseContext(searchPhaseContext)
             .build();
         execute(request);
     }
 
     public void execute(final NormalizationProcessorWorkflowExecuteRequest request) {
+        List<QuerySearchResult> querySearchResults = request.getQuerySearchResults();
+        Optional<FetchSearchResult> fetchSearchResultOptional = request.getFetchSearchResultOptional();
+
         // save original state
-        List<Integer> unprocessedDocIds = unprocessedDocIds(request.getQuerySearchResults());
+        List<Integer> unprocessedDocIds = unprocessedDocIds(querySearchResults);
 
         // pre-process data
         log.debug("Pre-process query results");
-        List<CompoundTopDocs> queryTopDocs = getQueryTopDocs(request.getQuerySearchResults());
+        List<CompoundTopDocs> queryTopDocs = getQueryTopDocs(querySearchResults);
 
         explain(request, queryTopDocs);
 
@@ -93,8 +99,9 @@ public class NormalizationProcessorWorkflow {
         CombineScoresDto combineScoresDTO = CombineScoresDto.builder()
             .queryTopDocs(queryTopDocs)
             .scoreCombinationTechnique(request.getCombinationTechnique())
-            .querySearchResults(request.getQuerySearchResults())
-            .sort(evaluateSortCriteria(request.getQuerySearchResults(), queryTopDocs))
+            .querySearchResults(querySearchResults)
+            .sort(evaluateSortCriteria(querySearchResults, queryTopDocs))
+            .fromValueForSingleShard(getFromValueIfSingleShard(request))
             .build();
 
         // combine
@@ -103,8 +110,26 @@ public class NormalizationProcessorWorkflow {
 
         // post-process data
         log.debug("Post-process query results after score normalization and combination");
-        updateOriginalQueryResults(combineScoresDTO);
-        updateOriginalFetchResults(request.getQuerySearchResults(), request.getFetchSearchResultOptional(), unprocessedDocIds);
+        updateOriginalQueryResults(combineScoresDTO, fetchSearchResultOptional.isPresent());
+        updateOriginalFetchResults(
+            querySearchResults,
+            fetchSearchResultOptional,
+            unprocessedDocIds,
+            combineScoresDTO.getFromValueForSingleShard()
+        );
+    }
+
+    /**
+     * Get value of from parameter when there is a single shard
+     * and fetch phase is already executed
+     * Ref https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/SearchService.java#L715
+     */
+    private int getFromValueIfSingleShard(final NormalizationProcessorWorkflowExecuteRequest request) {
+        final SearchPhaseContext searchPhaseContext = request.getSearchPhaseContext();
+        if (searchPhaseContext.getNumShards() > 1 || request.fetchSearchResultOptional.isEmpty()) {
+            return -1;
+        }
+        return searchPhaseContext.getRequest().source().from();
     }
 
     /**
@@ -173,18 +198,32 @@ public class NormalizationProcessorWorkflow {
         return queryTopDocs;
     }
 
-    private void updateOriginalQueryResults(final CombineScoresDto combineScoresDTO) {
+    private void updateOriginalQueryResults(final CombineScoresDto combineScoresDTO, final boolean isFetchPhaseExecuted) {
         final List<QuerySearchResult> querySearchResults = combineScoresDTO.getQuerySearchResults();
         final List<CompoundTopDocs> queryTopDocs = getCompoundTopDocs(combineScoresDTO, querySearchResults);
         final Sort sort = combineScoresDTO.getSort();
+        int totalScoreDocsCount = 0;
         for (int index = 0; index < querySearchResults.size(); index++) {
             QuerySearchResult querySearchResult = querySearchResults.get(index);
             CompoundTopDocs updatedTopDocs = queryTopDocs.get(index);
+            totalScoreDocsCount += updatedTopDocs.getScoreDocs().size();
             TopDocsAndMaxScore updatedTopDocsAndMaxScore = new TopDocsAndMaxScore(
                 buildTopDocs(updatedTopDocs, sort),
                 maxScoreForShard(updatedTopDocs, sort != null)
             );
+            // Fetch Phase had ran before the normalization phase, therefore update the from value in result of each shard.
+            // This will ensure the trimming of the search results.
+            if (isFetchPhaseExecuted) {
+                querySearchResult.from(combineScoresDTO.getFromValueForSingleShard());
+            }
             querySearchResult.topDocs(updatedTopDocsAndMaxScore, querySearchResult.sortValueFormats());
+        }
+
+        final int from = querySearchResults.get(0).from();
+        if (from > totalScoreDocsCount) {
+            throw new IllegalArgumentException(
+                String.format(Locale.ROOT, "Reached end of search result, increase pagination_depth value to see more results")
+            );
         }
     }
 
@@ -244,7 +283,8 @@ public class NormalizationProcessorWorkflow {
     private void updateOriginalFetchResults(
         final List<QuerySearchResult> querySearchResults,
         final Optional<FetchSearchResult> fetchSearchResultOptional,
-        final List<Integer> docIds
+        final List<Integer> docIds,
+        final int fromValueForSingleShard
     ) {
         if (fetchSearchResultOptional.isEmpty()) {
             return;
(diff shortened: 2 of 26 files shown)


vibrantvarun:
Integ test and BWC tests are failing due to https://github.com/opensearch-project/ml-commons/issues/3321

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -276,14 +316,21 @@ private void updateOriginalFetchResults(
...
+        for (int i = fromValueForSingleShard; i < topDocs.scoreDocs.length; i++) {
+            // Read topDocs after the desired from length
+            ScoreDoc scoreDoc = topDocs.scoreDocs[i];
             // get fetched hit content by doc_id
             SearchHit searchHit = docIdToSearchHit.get(scoreDoc.doc);
             // update score to normalized/combined value (3)
             searchHit.score(scoreDoc.score);
-            return searchHit;
-        }).toArray(SearchHit[]::new);
+            updatedSearchHitArray[i - fromValueForSingleShard] = searchHit;
```
martin-gaievski, line 332: we need to check if `i - fromValueForSingleShard` is a valid index, within [0, array.length- 1] boundaries

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
```diff
@@ -31,20 +32,25 @@
  * Implementation of Query interface for type "hybrid". It allows execution of multiple sub-queries and collect individual
  * scores for each sub-query.
  */
+@Getter
```
martin-gaievski, line 35: this Getter also applied to the `subQueries`, why do we need this in scope of pagination?

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -461,6 +477,34 @@ private ReduceableSearchResult reduceSearchResults(final List<ReduceableSearchRe
...
+     * @param searchContext search context that contains pagination depth
+     * @return results size to collected
+     */
+    private static int getSubqueryResultsRetrievalSize(final SearchContext searchContext) {
+        HybridQuery hybridQuery = getHybridQueryFromAbstractQuery(searchContext.query());
+        int paginationDepth = hybridQuery.getPaginationDepth();
+
+        // Switch to from+size retrieval size during standard hybrid query execution.
+        if (searchContext.from() == 0) {
+            return searchContext.from() + searchContext.size();
```
martin-gaievski, line 491: we can just return `searchContext.size()`

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -461,6 +477,34 @@ private ReduceableSearchResult reduceSearchResults(final List<ReduceableSearchRe
...
+        }
+        return paginationDepth;
+    }
+
+    private static HybridQuery getHybridQueryFromAbstractQuery(Query query) {
```
martin-gaievski, line 496: can you think of a better method name, it not clear what we're trying to do here

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
```diff
@@ -31,20 +32,25 @@
...
 public final class HybridQuery extends Query implements Iterable<Query> {
 
     private final List<Query> subQueries;
+    private int paginationDepth;
```
martin-gaievski, line 39: this parameter is low level for the Query level, can you create some more generic DTO for params like this, some sort of QueryContext, where pagination will be one of the parameters.We can put `explain` and `sort` there too, then pass only that context between the Query-QueryBuilder and even combination/normalization related classes

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
vibrantvarun: We already have getter for subqueries https://github.com/opensearch-project/neural-search/blob/f844a78c045116457dc75042dca1a6af410f2df3/src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java#L189.

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
vibrantvarun: I can create a HybridQueryContext dto object. Also I will add pagination depth in it. However, I will avoid touching explain and sort in this PR

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
vibrantvarun: Well, if you look at the code it is just putting values in the fresh array updateSearchHitArray. i starts from fromValueForSingleShard. So i-fromValueForSingleShard will be 0.  i can maximum go to array.length- 1 so it has already been taken care off. Do you still want me to put extra assert condition here?

martin-gaievski: Understood, my intention is that we have array index iterator that goes from 0 to the length -1, we can adjust the logic for the for loop temp variable, something like this:
 
for (int i = 0; i < Math.min(topDocs.scoreDocs.length - topDocs.scoreDocs.length, updatedSearchHitArray.length); i++) {
     ScoreDoc scoreDoc = topDocs.scoreDocs[i + fromValueForSingleShard];
     ....
     updatedSearchHitArray[i] = searchHit;

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
martin-gaievski: Makes sense, let add the class for context and with field for pagination, we can refactor other fields in later PR

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -483,17 +483,20 @@ private ReduceableSearchResult reduceSearchResults(final List<ReduceableSearchRe
...
+            return searchContext.size();
         }
         return paginationDepth;
     }
 
-    private static HybridQuery getHybridQueryFromAbstractQuery(Query query) {
+    /**
+     * Extract hybrid query from generic query object retrieved from search context
+     */
+    private static HybridQuery extractHybridQueryFromAbstractQuery(Query query) {
```
martin-gaievski, line 499:
```suggestion
/**
 * Unwraps a HybridQuery from either a direct query or a nested BooleanQuery
 */
private static HybridQuery unwrapHybridQuery(Query query) {
```

src/main/java/org/opensearch/neuralsearch/query/HybridQueryContext.java
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.query;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Getter;
+
+@AllArgsConstructor
```
martin-gaievski, line 11: We need to use either constructor or builder, but not both. You need to either remove constructor annotation or make it private. 
To enforce setting of required `pagination` param use NonNull annotation

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -297,6 +326,20 @@ private Collection<Query> toQueries(Collection<QueryBuilder> queryBuilders, Quer
         return queries;
     }
 
+    private static void validatePaginationDepth(final int paginationDepth, final QueryShardContext queryShardContext) {
+        if (paginationDepth < LOWER_BOUND_OF_PAGINATION_DEPTH) {
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "pagination_depth should be greater than 0"));
```
martin-gaievski, line 331: we should either hardcode '0' for the if condition check, or use  `LOWER_BOUND_OF_PAGINATION_DEPTH` constant in the exception message, this way it will be consistent

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -461,6 +477,37 @@ private ReduceableSearchResult reduceSearchResults(final List<ReduceableSearchRe
...
+
+    /**
+     * Unwraps a HybridQuery from either a direct query or a nested BooleanQuery
+     */
+    private static HybridQuery unwrapHybridQuery(Query query) {
+        HybridQuery hybridQuery;
+        // In case of nested fields and alias filter, hybrid query is wrapped under bool query and lies in the first clause.
+        if (query instanceof BooleanQuery) {
+            BooleanQuery booleanQuery = (BooleanQuery) query;
+            hybridQuery = (HybridQuery) booleanQuery.clauses().get(0).getQuery();
```
martin-gaievski, line 504: we need more checks here, I think what you're trying to do is similar to this method in QPS https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java#L87-L103. 
Can you extract some common code to the utility class, this one is a good candidate: https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/util/HybridQueryUtil.java

martin-gaievski (commented):
minor changes requested, overall code looks good to me

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
vibrantvarun: In the whole hybrid query work flow, the query phase searcher will execute first and will check these conditions thoroughly. In hybrid collector manager where I just want to get the hybrid query object, I think just checking for the instance of boolean query will work. Because I add those checks of QPS here also, then in the whole workflow it will be redundant and unecessorily impact performance. WDYT?

martin-gaievski: How we ended up with the HQ not being the top query (I assume that's the case because of the unwrap method) if the QPS should do such processing and unwrap everything at the first place? Is this some sort of a edge case?

martin-gaievski: Regarding the code organization question: I'm in favor of extracting common code, it should not add lot of overhead if we have 2-3 if or instance of checks, but we can eliminate potential problem of having different logic in different places for HQ.

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -78,16 +73,9 @@ private QueryPhaseSearcher getQueryPhaseSearcher(final SearchContext searchConte
             : defaultQueryPhaseSearcherWithEmptyCollectorContext;
     }
 
-    private static boolean isWrappedHybridQuery(final Query query) {
```
vibrantvarun, line 81: Duplicate method in HybridQueryUtil so removed it

src/main/java/org/opensearch/neuralsearch/util/HybridQueryUtil.java
```diff
@@ -56,12 +56,18 @@ public static boolean hasNestedFieldOrNestedDocs(final Query query, final Search
         return searchContext.mapperService().hasNested() && new NestedHelper(searchContext.mapperService()).mightMatchNestedDocs(query);
     }
 
-    private static boolean isWrappedHybridQuery(final Query query) {
+    public static boolean isWrappedHybridQuery(final Query query) {
```
martin-gaievski, line 59: please add a method level javadoc, now we need it once it became public

src/main/java/org/opensearch/neuralsearch/util/HybridQueryUtil.java
```diff
@@ -56,12 +56,18 @@ public static boolean hasNestedFieldOrNestedDocs(final Query query, final Search
...
         return query instanceof BooleanQuery
             && ((BooleanQuery) query).clauses().stream().anyMatch(clauseQuery -> clauseQuery.getQuery() instanceof HybridQuery);
     }
 
     public static boolean hasAliasFilter(final Query query, final SearchContext searchContext) {
         return Objects.nonNull(searchContext.aliasFilter());
     }
+
+    public static boolean isHybridQueryWrappedInBooleanQuery(final SearchContext searchContext, final Query query) {
```
martin-gaievski, line 68: please add javadoc for the method

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -297,6 +326,22 @@ private Collection<Query> toQueries(Collection<QueryBuilder> queryBuilders, Quer
...
+            throw new IllegalArgumentException(
+                String.format(Locale.ROOT, "pagination_depth should be greater than %s", LOWER_BOUND_OF_PAGINATION_DEPTH)
+            );
+        }
+        // compare pagination depth with OpenSearch setting index.max_result_window
+        // see https://opensearch.org/docs/latest/install-and-configure/configuring-opensearch/index-settings/
+        int maxResultWindowIndexSetting = queryShardContext.getIndexSettings().getMaxResultWindow();
+        if (paginationDepth > maxResultWindowIndexSetting) {
+            throw new IllegalArgumentException(
+                String.format(Locale.ROOT, "pagination_depth should be less than index.max_result_window setting")
```
martin-gaievski, line 340: message is not entirely correct, we're allowing values less or equal to index setting

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -337,7 +337,7 @@ private static void validatePaginationDepth(final int paginationDepth, final Que
         int maxResultWindowIndexSetting = queryShardContext.getIndexSettings().getMaxResultWindow();
         if (paginationDepth > maxResultWindowIndexSetting) {
             throw new IllegalArgumentException(
-                String.format(Locale.ROOT, "pagination_depth should be less than index.max_result_window setting")
+                String.format(Locale.ROOT, "pagination_depth should be less than or equal to index.max_result_window setting")
```
martin-gaievski, line 340:
```suggestion
                String.format(Locale.ROOT, "pagination_depth should be less than or equal to %s setting", IndexSettings.MAX_RESULT_WINDOW_SETTING.getKey())
```

martin-gaievski (approved):
great work Varun, approving

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
```diff
@@ -34,17 +34,21 @@
...
      * @param subQueries collection of queries that are executed individually and contribute to a final list of combined scores
      * @param filterQueries list of filters that will be applied to each sub query. Each filter from the list is added as bool "filter" clause. If this is null sub queries will be executed as is
      */
-    public HybridQuery(final Collection<Query> subQueries, final List<Query> filterQueries) {
+    public HybridQuery(final Collection<Query> subQueries, final List<Query> filterQueries, final HybridQueryContext hybridQueryContext) {
         Objects.requireNonNull(subQueries, "collection of queries must not be null");
         if (subQueries.isEmpty()) {
             throw new IllegalArgumentException("collection of queries must not be empty");
         }
+        if (hybridQueryContext.getPaginationDepth() == 0) {
```
ryanbogan, line 49: Should we call the `validatePaginationDepth` method in `HybridQueryBuilder` here since there are more restrictions than not being 0?

vibrantvarun: This is the validation to set the default value in case where user didn't send pagination depth in query and we add it a default value from the system.

vibrantvarun: We already have a validate pagination depth method in doToQuery method where on the basis of paginationdepth we evaluate whether to throw exception or move forward.

ryanbogan: Makes sense, approving

ryanbogan (approved):

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -263,7 +294,7 @@ protected boolean doEquals(HybridQueryBuilder obj) {
      */
     @Override
     protected int doHashCode() {
-        return Objects.hash(queries);
+        return Objects.hash(queries, paginationDepth);
```
VijayanB, line 297: shouldn't you check for cluster version here? similar to equals

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -294,6 +325,26 @@ private Collection<Query> toQueries(Collection<QueryBuilder> queryBuilders, Quer
         return queries;
     }
 
+    private static void validatePaginationDepth(final int paginationDepth, final QueryShardContext queryShardContext) {
+        if (paginationDepth <= LOWER_BOUND_OF_PAGINATION_DEPTH) {
```
VijayanB, line 329: i believe lower bound means it is valid value, i would recommend add + 1 to lower bound and change it to <

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 2.x](https://github.com/opensearch-project/neural-search/compare/2.18...2.x)
 ### Features
+- Pagination in Hybrid query ([#963](https://github.com/opensearch-project/neural-search/pull/963))
```
yuye-aws, line 19: Can you link the current one: https://github.com/opensearch-project/neural-search/pull/1048

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
vibrantvarun: Changing condition to `paginationDepth <= LOWER_BOUND_OF_PAGINATION_DEPTH`

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -254,6 +282,9 @@ protected boolean doEquals(HybridQueryBuilder obj) {
         }
         EqualsBuilder equalsBuilder = new EqualsBuilder();
         equalsBuilder.append(queries, obj.queries);
+        if (isClusterOnOrAfterMinReqVersionForPaginationInHybridQuery()) {
```
martin-gaievski, line 285: in what scenario we need to do this check, and what will happen if code does not have it?

vibrantvarun: I put it to maintain the consistency in the code.

vibrantvarun: Because this field is introduced in version 2.19, for code understanding purpose and keeping the consistency, I put this check

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
martin-gaievski: is this driven primarily by code consistency since we have the same check in the equals method, or is there an actual scenario where this could be a problem? I can't think of anything other than rolling upgrades, and I'm not even sure if it would be an issue in that case.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
martin-gaievski: then why we need it in hashCode method?

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
vibrantvarun: I think it would be good to put this check as it makes the code consistent in the entire file

martin-gaievski: consistency come for `equals` method after you have added it to `hashcode` method. I'm asking why it was added to `hashcode` in the first place.

VijayanB (approved):
LGTM

knoha:
Is this PR included in 2.19 version release?

heemin32:
> Is this PR included in 2.19 version release?

Yes

Issue #951, opened 2024-10-21
[Infrastructure] Add bwc tests for rescore functionality in hybrid query


vibrantvarun:
Add bwc tests for rescore functionality in hybrid query. PR [917](https://github.com/opensearch-project/neural-search/pull/917)

martin-gaievski:
I've added rescorer as part of the existing rolling upgrade test in initial PR https://github.com/opensearch-project/neural-search/pull/917/files#diff-7acc444fb409f783c5ff067c7d1e5a173eac25427c04765f8af9b6f26b7e18e7. Do you mean some additional/different test in scope of this issue @vibrantvarun?

vibrantvarun:
@martin-gaievski I think it is good. Because we already tested rescore complex functionality in integ test and unit test. For BWC one test would be fine. We can close this this issue.
cc: @minalsha

PR #924, opened 2024-10-03
[Backport 2.x] Added rescorer in hybrid query (#917)

martin-gaievski:
Manually backport https://github.com/opensearch-project/neural-search/commit/9f4a49a7e45211821d96181ce2a6842af18ce7ea from https://github.com/opensearch-project/neural-search/pull/917

PR #917, opened 2024-09-30
Added rescorer in hybrid query

martin-gaievski:
### Description
Enable rescore functionality in hybrid query. Currently `rescore` clause in ignored and plain results of hybrid query are returned. Following is example of query with rescore clause.

Logic will be similar to one for traditional queries: rescore query will be applied to search hits returned by sub-queries, based on match/no match scores and final order of documents will be adjusted. 

To implementing this we're using custom method to call the rescore. In QueryPhaseSearch we always return `rescore == false`, then we do actual check (rescore/not rescore) in the hybrid collector manager. We do the check for rescore clause at the shard level, before individual sub-query results are merged into a single object. That allows to minimize number of changes in existing logic for score normalization.

Following diagram shows flow that we introduced with the PR:

Rescore - flow diagram new

For comparison following diagram shows flow for today's implementation

Rescore - flow diagram existing

Example of rescore usage:

Query with rescore 

```
{
    "query": {
        "hybrid": {
            "queries": [
                {
                    "knn": {
                        "vector": {
                            "vector": [
                                4.2,
                                5.0,
                                8.5
                            ],
                            "k": 10
                        }
                    }
                },
                {
                    "range": {
                        "field1": {
                            "gte": 1,
                            "lte": 60
                        }
                    }
                }
            ]
        }
    },
    "rescore": [
        {
            "query": {
                "rescore_query": {
                    "range": {
                        "field1": {
                            "gte": 1,
                            "lte": 20
                        }
                    }
                },
                "score_mode": "total",
                "query_weight": 0.9,
                "rescore_query_weight": 1.2
            },
            "window_size": 10
        }
    ]
}
```

Response with implemented rescore from this PR, where documents with `field` in range of [1..20] are boosted

```
{
    "took": 4,
    "timed_out": false,
    "_shards": {
        "total": 1,
        "successful": 1,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "total": {
            "value": 3,
            "relation": "eq"
        },
        "max_score": 0.6695955,
        "hits": [
            {
                "_index": "index-test",
                "_id": "BhGESpIB67ctKZAvyp6_",
                "_score": 0.6695955,
                "_source": {
                    "field1": 2,
                    "vector": [
                        0.4,
                        0.5,
                        0.2
                    ],
                    "title": "basic"
                }
            },
            {
                "_index": "index-test",
                "_id": "BxGESpIB67ctKZAv057_",
                "_score": 0.6695361,
                "_source": {
[...]

PR #1066, opened 2025-01-07
add hybrid search with rescore IT

will-hwang:
### Description
Currently, `main` branch doesn't contain `HybridSearchWithRescoreIT` test cases in both` restart-upgrade` and `rolling-upgrade` test suites, which are present in 2.x branch. This discrepancy causes failures to changes made to 2.x (e.g: https://github.com/opensearch-project/neural-search/actions/runs/12644267233)

This PR adds the missing commits to main branch between #917 and manually created backport (#924)

### Related Issues
#1065 

### Check List

### DIFF
--- a/qa/restart-upgrade/build.gradle
+++ b/qa/restart-upgrade/build.gradle
@@ -102,6 +102,7 @@ task testAgainstOldCluster(type: StandaloneRestIntegTestTask) {
     if (versionsBelow2_15.any { ext.neural_search_bwc_version.startsWith(it) }){
         filter {
             excludeTestsMatching "org.opensearch.neuralsearch.bwc.NeuralSparseTwoPhaseProcessorIT.*"
+            excludeTestsMatching "org.opensearch.neuralsearch.bwc.HybridSearchWithRescoreIT.*"
         }
     }
 
@@ -166,6 +167,7 @@ task testAgainstNewCluster(type: StandaloneRestIntegTestTask) {
     if (versionsBelow2_15.any { ext.neural_search_bwc_version.startsWith(it) }){
         filter {
             excludeTestsMatching "org.opensearch.neuralsearch.bwc.NeuralSparseTwoPhaseProcessorIT.*"
+            excludeTestsMatching "org.opensearch.neuralsearch.bwc.HybridSearchWithRescoreIT.*"
         }
     }
 
--- a/qa/restart-upgrade/build.gradle
+++ b/qa/restart-upgrade/build.gradle
@@ -137,47 +137,6 @@ task testAgainstNewCluster(type: StandaloneRestIntegTestTask) {
     systemProperty 'tests.is_old_cluster', 'false'
     systemProperty 'tests.plugin_bwc_version', ext.neural_search_bwc_version
 
-    // Excluding MultiModalSearchIT, HybridSearchIT, NeuralSparseSearchIT, NeuralQueryEnricherProcessorIT tests from neural search version 2.9 and 2.10
-    // because these features were released in 2.11 version.
-    if (versionsBelow2_11.any { ext.neural_search_bwc_version.startsWith(it) }){
-        filter {
-            excludeTestsMatching "org.opensearch.neuralsearch.bwc.MultiModalSearchIT.*"
-            excludeTestsMatching "org.opensearch.neuralsearch.bwc.HybridSearchIT.*"
-            excludeTestsMatching "org.opensearch.neuralsearch.bwc.NeuralSparseSearchIT.*"
-            excludeTestsMatching "org.opensearch.neuralsearch.bwc.NeuralQueryEnricherProcessorIT.*"
-        }
-    }
-
-    // Excluding these tests because we introduce them in 2.13
-    if (versionsBelow2_13.any { ext.neural_search_bwc_version.startsWith(it) }){
-        filter {
-            excludeTestsMatching "org.opensearch.neuralsearch.bwc.NeuralQueryEnricherProcessorIT.testNeuralQueryEnricherProcessor_NeuralSparseSearch_E2EFlow"
-            excludeTestsMatching "org.opensearch.neuralsearch.bwc.TextChunkingProcessorIT.*"
-        }
-    }
-
-    // Excluding the k-NN radial search tests because we introduce this feature in 2.14
-    if (versionsBelow2_14.any { ext.neural_search_bwc_version.startsWith(it) }){
-        filter {
-            excludeTestsMatching "org.opensearch.neuralsearch.bwc.KnnRadialSearchIT.*"
-        }
-    }
-
-    // Excluding the NeuralSparseQuery two-phase search pipeline tests because we introduce this feature in 2.15
-    if (versionsBelow2_15.any { ext.neural_search_bwc_version.startsWith(it) }){
-        filter {
-            excludeTestsMatching "org.opensearch.neuralsearch.bwc.NeuralSparseTwoPhaseProcessorIT.*"
-            excludeTestsMatching "org.opensearch.neuralsearch.bwc.HybridSearchWithRescoreIT.*"
-        }
-    }
-
-    // Excluding the batch processor tests because we introduce this feature in 2.16
-    if (versionsBelow2_16.any { ext.neural_search_bwc_version.startsWith(it) }){
-        filter {
-            excludeTestsMatching "org.opensearch.neuralsearch.bwc.BatchIngestionIT.*"
-        }
-    }
-
     nonInputProperties.systemProperty('tests.rest.cluster', "${-> testClusters."${baseName}".allHttpSocketURI.join(",")}")
     nonInputProperties.systemProperty('tests.clustername', "${-> testClusters."${baseName}".getName()}")
     systemProperty 'tests.security.manager', 'false'
--- a/qa/restart-upgrade/build.gradle
+++ b/qa/restart-upgrade/build.gradle
@@ -102,7 +102,6 @@ task testAgainstOldCluster(type: StandaloneRestIntegTestTask) {
     if (versionsBelow2_15.any { ext.neural_search_bwc_version.startsWith(it) }){
         filter {
             excludeTestsMatching "org.opensearch.neuralsearch.bwc.NeuralSparseTwoPhaseProcessorIT.*"
-            excludeTestsMatching "org.opensearch.neuralsearch.bwc.HybridSearchWithRescoreIT.*"
         }
     }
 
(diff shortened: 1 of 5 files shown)


heemin32 (approved):
LGTM

martin-gaievski:
test for rescore for rolling upgrade is already part of he main https://github.com/opensearch-project/neural-search/blob/main/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/HybridSearchIT.java#L71-L72, added in https://github.com/opensearch-project/neural-search/pull/917. 
Main reason why we have a separate class for rescore test in 2.x is to be able to block is for lower versions, where rescore is not supported. Such blockage and filtering is easy to do if you have a separate test class.

What you need to do is to add rescore part to the restart BWC test, that's it.

qa/restart-upgrade/build.gradle
```diff
@@ -102,6 +102,7 @@ task testAgainstOldCluster(type: StandaloneRestIntegTestTask) {
     if (versionsBelow2_15.any { ext.neural_search_bwc_version.startsWith(it) }){
         filter {
             excludeTestsMatching "org.opensearch.neuralsearch.bwc.NeuralSparseTwoPhaseProcessorIT.*"
+            excludeTestsMatching "org.opensearch.neuralsearch.bwc.HybridSearchWithRescoreIT.*"
```
martin-gaievski, line 105: this check is not needed in main, latest 2.x which is 2.18 already supports rescore.

qa/restart-upgrade/build.gradle
```diff
@@ -166,6 +167,7 @@ task testAgainstNewCluster(type: StandaloneRestIntegTestTask) {
...
```
martin-gaievski, line 170: same as above

qa/rolling-upgrade/build.gradle
```diff
@@ -101,6 +101,7 @@ task testAgainstOldCluster(type: StandaloneRestIntegTestTask) {
     if (versionsBelow2_15.any { ext.neural_search_bwc_version.startsWith(it) }){
         filter {
             excludeTestsMatching "org.opensearch.neuralsearch.bwc.NeuralSparseTwoPhaseProcessorIT.*"
+            excludeTestsMatching "org.opensearch.neuralsearch.bwc.HybridSearchWithRescoreIT.*"
```
martin-gaievski, line 104: same as above, 2.18 supports rescore in hybrid not need to exclude test class

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/HybridSearchWithRescoreIT.java
```diff
@@ -0,0 +1,151 @@
...
+    private static final String QUERY = "Hi world";
+    private static final int NUM_DOCS_PER_ROUND = 1;
+    private static final String VECTOR_EMBEDDING_FIELD = "passage_embedding";
+    protected static final String RESCORE_QUERY = "hi";
+    private static String modelId = "";
+
+    /**
+     * Test normalization with hybrid query and rescore. This test is required as rescore will not be compatible with version lower than 2.15
+     */
+    public void testHybridQueryWithRescore_whenIndexWithMultipleShards_E2EFlow() throws Exception {
```
martin-gaievski, line 45: we do have rescore check as part of another test, https://github.com/opensearch-project/neural-search/blob/main/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/HybridSearchIT.java#L71-L72. This whole class is not needed, or if you want it to keep consistency please remove the rescore calls for the https://github.com/opensearch-project/neural-search/blob/main/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/HybridSearchIT.java

heemin32:
> test for rescore for rolling upgrade is already part of he main https://github.com/opensearch-project/neural-search/blob/main/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/HybridSearchIT.java#L71-L72, added in #917. Main reason why we have a separate class for rescore test in 2.x is to be able to block is for lower versions, where rescore is not supported. Such blockage and filtering is easy to do if you have a separate test class. At the same time it's not needed for main because in BWC we need to support only latest 2.x (2.18 as of today) and that version supports rescore.
> 
> What you need to do is to add rescore part to the restart BWC test, that's it.

1. It is not required in the `main` branch.  
2. The test is included in a separate class only in the `2.x` branch to simplify blocking for lower versions.  

Despite these reasons, I prefer to keep the `main` and `2.x` branches as similar as possible for easier maintenance, unless there are technical constraints preventing it.

martin-gaievski (commented):
please check my comments, I leaving up to you if want to have a separate class just for rescore as in 2.x, or keep it as part of the https://github.com/opensearch-project/neural-search/blob/main/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/HybridSearchIT.java, the wayt it's done in main. Just pick one of the approaches.
You need to drop unnecessary checks for version in both restart and rolling upgrade, this will be `main only` change

heemin32:
> You need to drop unnecessary checks for version in both restart and rolling upgrade, this will be `main only` change
I would like to have version check in main so that the code will looks same between `main` and `2.x` even if the check is not needed for `main`

qa/restart-upgrade/build.gradle
heemin32: I prefer to keep it so that the code looks same between `main` and `2.x`

martin-gaievski:
> > test for rescore for rolling upgrade is already part of he main https://github.com/opensearch-project/neural-search/blob/main/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/HybridSearchIT.java#L71-L72, added in #917. Main reason why we have a separate class for rescore test in 2.x is to be able to block is for lower versions, where rescore is not supported. Such blockage and filtering is easy to do if you have a separate test class. At the same time it's not needed for main because in BWC we need to support only latest 2.x (2.18 as of today) and that version supports rescore.
> > What you need to do is to add rescore part to the restart BWC test, that's it.
> 
> 1. The test is included in a separate class only in the `2.x` branch to simplify blocking for lower versions.
> 2. Blocking is not required in the `main` branch.
> 
> Despite these reasons, I prefer to keep the `main` and `2.x` branches as similar as possible for easier maintenance, unless there are technical constraints preventing it.

then in this PR we need to drop the part for rescore from the https://github.com/opensearch-project/neural-search/blob/main/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/HybridSearchIT.java, so classes in both branches are similar. 
We don't need version checks in main, it doesn't have any effect on the test execution because main suppose to be BWC  against 2.18 and above only

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/HybridSearchWithRescoreIT.java
heemin32: Let's remove the rescore test from `HybridSearchIT`. I prefer to keep the code consistent between the `main` and `2.x` branches whenever possible.

martin-gaievski:
> > We don't need version checks in main, it doesn't have any effect on the test execution because main suppose to be BWC against 2.18 and above only
> 
> Yes, but it would be ideal to keep the code consistent between `main` and `2.x`. Consistency simplifies maintenance and eliminates the need to figure out why the code differs between the two branches.

I don't agree, I think it actually makes it harder to understand and maintain, like having old code commented instead of just removing it after it's not longer needed. It's like the OPS activities after version release, checks that are no longer valid can be safely removed

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/HybridSearchWithRescoreIT.java
martin-gaievski: sure, we can do that so code is same

heemin32:
> > > We don't need version checks in main, it doesn't have any effect on the test execution because main suppose to be BWC against 2.18 and above only
> > 
> > 
> > Yes, but it would be ideal to keep the code consistent between `main` and `2.x`. Consistency simplifies maintenance and eliminates the need to figure out why the code differs between the two branches.
> 
> I don't agree, I think it actually makes it harder to understand and maintain, like having old code commented instead of just removing it after it's not longer needed. It's like the OPS activities after version release, checks that are no longer valid can be safely removed

If you see https://github.com/opensearch-project/neural-search/blob/main/qa/rolling-upgrade/build.gradle, we are still keeping version check in rolling upgrade in `main` branch instead of removing them for many tests.

martin-gaievski:
> > > > We don't need version checks in main, it doesn't have any effect on the test execution because main suppose to be BWC against 2.18 and above only
> > > 
> > > 
> > > Yes, but it would be ideal to keep the code consistent between `main` and `2.x`. Consistency simplifies maintenance and eliminates the need to figure out why the code differs between the two branches.
> > 
> > 
> > I don't agree, I think it actually makes it harder to understand and maintain, like having old code commented instead of just removing it after it's not longer needed. It's like the OPS activities after version release, checks that are no longer valid can be safely removed
> 
> If you see https://github.com/opensearch-project/neural-search/blob/main/qa/rolling-upgrade/build.gradle, we are still keeping version check in rolling upgrade in `main` branch instead of removing them for many tests.

and this isn't correct in my opinion, those should be cleared as part of the post-release activities. Typical sequence should be: add code and version check in main, do backport, release 2.x version, do cleanup. This cleanup doesn't happen for whatever reason. Why we should follow the broken pattern if we know it's not correct.

I still think we need to drop the version checks, I'm fine with the code in this PR.

will-hwang:
> > > > > We don't need version checks in main, it doesn't have any effect on the test execution because main suppose to be BWC against 2.18 and above only
> > > > 
> > > > 
> > > > Yes, but it would be ideal to keep the code consistent between `main` and `2.x`. Consistency simplifies maintenance and eliminates the need to figure out why the code differs between the two branches.
> > > 
> > > 
> > > I don't agree, I think it actually makes it harder to understand and maintain, like having old code commented instead of just removing it after it's not longer needed. It's like the OPS activities after version release, checks that are no longer valid can be safely removed
> > 
> > 
> > If you see https://github.com/opensearch-project/neural-search/blob/main/qa/rolling-upgrade/build.gradle, we are still keeping version check in rolling upgrade in `main` branch instead of removing them for many tests.
> 
> and this isn't correct in my opinion, those should be cleared as part of the post-release activities. Typical sequence should be: add code and version check in main, do backport, release 2.x version, do cleanup. This cleanup doesn't happen for whatever reason. Why we should follow the broken pattern if we know it's not correct.
> 
> I still think we need to drop the version checks, I'm fine with the java code for tests in this PR.

what is the practice followed in other teams? i think there should be a set guideline. For example, we should keep the version check for all 2.x upgrades, and clean them all at once once there's an update to 3.x. etc

martin-gaievski (approved):
Thank you for addressing all comments

martin-gaievski:
with this PR we can close https://github.com/opensearch-project/neural-search/issues/951

Issue #1052, opened 2025-01-03
[Proposal][RFC] Support analyzer-based neural sparse query & build BERT tokenizer as pre-defined tokenizer


zhichao-aws:
### What problems are you trying to solve?
Currently for neural sparse query, users need to register a sparse_encoding/sparse_tokenize model in advance and provide the model id in query body. For bi-encoder mode, we do need the ml-commons suite to manage the lifecycle of sparse encoding models. But for doc-only mode, we only use a tokenizer for query, and it would be somehow heavy to manage it with ml-commons suite. There will be several drawbacks:
- users need to configure the only_run_on_ml_node settings to enable the tokenizer on data nodes
- users need to register the model and manage the model groups, even to manage the model_id
- the tokenizer predict requests will be dispatched among cluster nodes, which brings extra traffic cost

### What are you proposing?
Build the analyzer-based neural sparse query. The sparse_tokenize model will be wrapped as a [Lucene Analyzer](https://lucene.apache.org/core/8_0_0/core/org/apache/lucene/analysis/Analyzer.html). Users bind the analyzer to index field, and the neural sparse query will call the analyzer to encode the query.

The pretrained `amazon/neural-sparse/opensearch-neural-sparse-tokenizer-v1` will be supported as pre-defined tokenizer. The token weight is encoded in payload attribute.

Besides used for neural sparse query, the analyzer can also be invoked like others. E.g. analyze API, chunking processor.

### What is the developer experience going to be?
Will alter the model_id verification logics at neural sparse query builder. And add the pre-defined bert analyzer.

#### Are there any security considerations? 
N/A

#### Are there any breaking changes to the API
We'll support a new query type for neural sparse query. I.e. users can bind the analyzer to index field, instead of providing the model id in query body.

### What is the user experience going to be?
create index
```
PUT /my-index
{
  "settings": {
    "default_pipeline": "nlp-ingest-pipeline-sparse"
  },
  "mappings": {
    "properties": {
      "passage_embedding": {
        "type": "rank_features"
      },
      "passage_text": {
        "type": "text"
      }
    }
  }
}
```

search
```
GET my-index/search
{
    "query":{
        "neural_sparse": {
            "passage_embedding": {
                "query_text": "hello world"，
                “analyzer": "bert-uncased"
            }
        }
    }
}
```

### What will it take to execute?
1. Will modify the neural sparse query logics. The model id is not required. And it will read analyzer from shard context, then use analyzer to encode the query text.
[...]

PR #1088, opened 2025-01-10
Support analyzer-based neural sparse query

zhichao-aws:
### Description
Implementation for #1052 

### Check List


src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -129,20 +137,50 @@ public NeuralSparseQueryBuilder(StreamInput in) throws IOException {
...
     /**
-     * Copy this QueryBuilder for two phase rescorer, set the copy one's twoPhasePruneRatio to -1.
+     * Copy this QueryBuilder for two phase rescorer.
      * @param pruneRatio the parameter of the NeuralSparseTwoPhaseProcessor, control how to split the queryTokens to two phase.
      * @return A copy NeuralSparseQueryBuilder for twoPhase, it will be added to the rescorer.
      */
     public NeuralSparseQueryBuilder getCopyNeuralSparseQueryBuilderForTwoPhase(float pruneRatio, PruneType pruneType) {
-        this.twoPhasePruneRatio(pruneRatio);
-        this.twoPhasePruneType(pruneType);
+        this.neuralSparseQueryTwoPhaseInfo = new NeuralSparseQueryTwoPhaseInfo(
```
will-hwang, line 170: instead of `PARENT`, `CHILD`, can we have more descriptive names?

`INITIAL`, `RECALCULATION` are more descriptive

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -129,20 +137,50 @@ public NeuralSparseQueryBuilder(StreamInput in) throws IOException {
...
+            pruneType
+        );
         NeuralSparseQueryBuilder copy = new NeuralSparseQueryBuilder().fieldName(this.fieldName)
             .queryName(this.queryName)
             .queryText(this.queryText)
             .modelId(this.modelId)
+            .analyzer(this.analyzer)
             .maxTokenScore(this.maxTokenScore)
-            .twoPhasePruneRatio(-1f * pruneRatio);
+            .neuralSparseQueryTwoPhaseInfo(
```
will-hwang, line 181: [nit] 

adding `.neuralSparseQueryTwoPhase` as an attribute doesn't give too much benefit in my opinion. We could directly have `twoPhaseStatus`, `twoPhasePruneRatio` and `twoPhasePruneType` as field attributes to `NeuralSparseQueryBuilder` as it was initially

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -363,14 +392,48 @@ private BiConsumer<Client, ActionListener<?>> getModelInferenceAsync(SetOnce<Map
...
+                while (stream.incrementToken()) {
+                    String token = term.toString();
+                    Float weight = Objects.isNull(payload.getPayload()) ? 1.0f : HFModelTokenizer.bytesToFloat(payload.getPayload().bytes);
+                    queryTokens.put(token, weight);
+                }
+                stream.end();
+            } catch (IOException e) {
+                throw new OpenSearchException("failed to analyze query text. ", e);
+            }
+            return switch (neuralSparseQueryTwoPhaseInfo.getStatus()) {
```
will-hwang, line 415: [nit] can we return after `stream.end() `at line 411?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryTwoPhaseInfo.java
```diff
@@ -0,0 +1,74 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Locale;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+@Getter
+@Setter
+public class NeuralSparseQueryTwoPhaseInfo implements Writeable {
```
will-hwang, line 23: add javadoc for this class

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: rename to `PHASE_ONE` and `PHASE_TWO`

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: ack

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryTwoPhaseInfo.java
zhichao-aws: ack

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: Using neuralSparseQueryTwoPhaseInfo can improve the code readiness and is more convenient for future extension

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -363,14 +396,51 @@ private BiConsumer<Client, ActionListener<?>> getModelInferenceAsync(SetOnce<Map
         ));
     }
 
+    protected Map<String, Float> getQueryTokens(QueryShardContext context) {
+        if (Objects.nonNull(queryTokensSupplier)) {
+            return queryTokensSupplier.get();
+        } else if (Objects.nonNull(analyzer)) {
+            Map<String, Float> queryTokens = new HashMap<>();
+            Analyzer luceneAnalyzer = context.convertToShardContext().getIndexAnalyzers().getAnalyzers().get(this.analyzer);
```
yuye-aws, line 404: 1. The convertToShardContext for QueryShardContext is simply returning itself. Can you explain why you are using this method?
2. When getting analyzer from a series of map, have you considered that case when the analyzer is not present? If so, please inform the user about the steps to register the analyzer beforehand.
3. The query and analyzer is shard level. When registering the analyzer, this line implicitly assumes that the analyzer is already present. When the analyzer is missing, is it possible to **reuse** the analyzer from index cluster or cluster level?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -363,14 +396,51 @@ private BiConsumer<Client, ActionListener<?>> getModelInferenceAsync(SetOnce<Map
...
+            try (TokenStream stream = luceneAnalyzer.tokenStream(fieldName, queryText)) {
+                stream.reset();
+                CharTermAttribute term = stream.addAttribute(CharTermAttribute.class);
+                PayloadAttribute payload = stream.addAttribute(PayloadAttribute.class);
+
+                while (stream.incrementToken()) {
+                    String token = term.toString();
+                    float weight = Objects.isNull(payload.getPayload()) ? 1.0f : bytesToFloat(payload.getPayload().bytes);
+                    if (weight > 0) {
+                        queryTokens.put(token, weight);
```
yuye-aws, line 414: I'm not sure which subclass of TokenStream you are using. Have you verified that there are no more duplicates?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -363,14 +396,51 @@ private BiConsumer<Client, ActionListener<?>> getModelInferenceAsync(SetOnce<Map
         ));
     }
 
+    protected Map<String, Float> getQueryTokens(QueryShardContext context) {
```
yuye-aws, line 399: I would suggest you refactor a little bit. You need a function to get query tokens from an analyzer. The queryTokensSupplier part can remain the same.

yuye-aws, line 399: Since you declare the function as protected and not private, are you indicating that the **NeuralSparseQueryBuilder** class will be inherited in the future?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -363,14 +396,51 @@ private BiConsumer<Client, ActionListener<?>> getModelInferenceAsync(SetOnce<Map
...
+                    case NeuralSparseQueryTwoPhaseInfo.TwoPhaseStatus.PHASE_ONE -> PruneUtils.splitSparseVector(
+                        neuralSparseQueryTwoPhaseInfo.getTwoPhasePruneType(),
+                        neuralSparseQueryTwoPhaseInfo.getTwoPhasePruneRatio(),
+                        queryTokens
+                    ).v1();
+                    default -> queryTokens;
+                };
+            } catch (IOException e) {
+                throw new OpenSearchException("failed to analyze query text. ", e);
+            }
```
yuye-aws, line 434: Getting token stream, parsing bytes into the float will also lead to error. Can you check other error types?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -363,14 +396,51 @@ private BiConsumer<Client, ActionListener<?>> getModelInferenceAsync(SetOnce<Map
...
+        }
+        throw new IllegalArgumentException("Query tokens cannot be null.");
```
yuye-aws, line 436: The error message needs to be updated. Both query tokens and analyzer is null.

zane-neo:
LGTM, will approve once some nit comments resolved.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: ack

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: It's BufferUnderflowException, extends RuntimeException.

Added error handling code and UT for it.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: 1. technically it can be inherited in other plugins, and we don't know whether users do it in their own codebase
2. using protected, we can test it easily in unit test

zhichao-aws: this function itself is to get query tokens from an analyzer + 2 lines of code to get from query token supplier. I think it's overdesign to further refactor it.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: do you mean a token can occur multiple times in token stream?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
yuye-aws: 1. Allowing Inheritance needs to be careful. You can remain the query builder class as the same for this PR.
2. This function can be declared as private and use @VisibleForTesting

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
yuye-aws: Yes

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: > The convertToShardContext for QueryShardContext is simply returning itself. Can you explain why you are using this method?
Thanks for the catch. I originally write it for QueryRewriteContext. Removed the convertToShardContext.

> When getting analyzer from a series of map, have you considered that case when the analyzer is not present? If so, please inform the user about the steps to register the analyzer beforehand.
Ack. Add logics to wrap the error message.

> The query and analyzer is shard level. When registering the analyzer, this line implicitly assumes that the analyzer is already present. When the analyzer is missing, is it possible to reuse the analyzer from index cluster or cluster level?
If an analyzer is cluster-level, it is included in all indexes. And a shard context contains all analyzers in the corresponding index. Cross-index analyzer-sharing is not supported now, it requires refactor in core and not in scope of this release.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: There can be such cases. But it won't impact the query tokens result. As we are using a map.

```
queryTokens.put(token, weight);
```

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: The visibleForTesting doesn't make private method visible in UT.

Let's use package-private.

yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
yuye-aws: Since there can be duplicates, are their weights going to be different? If so, it's not the best practice to simply override the previous result

zhichao-aws: The weights can't be different. For inf-free sparse model, the query token weight is just a look up table. And the same tokens should have identical weights.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -363,14 +397,56 @@ private BiConsumer<Client, ActionListener<?>> getModelInferenceAsync(SetOnce<Map
...
 
+    Map<String, Float> getQueryTokens(QueryShardContext context) {
+        if (Objects.nonNull(queryTokensSupplier)) {
+            return queryTokensSupplier.get();
+        } else if (Objects.nonNull(analyzer)) {
+            Map<String, Float> queryTokens = new HashMap<>();
+            Analyzer luceneAnalyzer = context.getIndexAnalyzers().getAnalyzers().get(this.analyzer);
+            if (Objects.isNull(luceneAnalyzer)) {
+                throw new IllegalArgumentException(String.format(Locale.ROOT, "Analyzer [%s] not found in shard context. ", this.analyzer));
+            }
```
yuye-aws, line 408: I'm afraid null pointer exception might be thrown if **context.getIndexAnalyzers()** is null. You can add a try catch for better error message

yuye-aws:
@zhichao-aws Must we merge this PR after: https://github.com/opensearch-project/ml-commons/pull/3719?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: I checked the implementation of MapperService.getIndexAnalyzers, and found this object is always set in OpenSearch code.

And in OpenSearch core, existing code also calls this method without null check https://github.com/opensearch-project/OpenSearch/blob/96481ccadb8e7fd40fac3dd09629bd00b2bf5c94/server/src/main/java/org/opensearch/index/search/MatchQuery.java#L216. So I think we don't need this.

zhichao-aws:
> @zhichao-aws Must we merge this PR after: [opensearch-project/ml-commons#3719](https://github.com/opensearch-project/ml-commons/pull/3719)?

We can merge this first. All UTs and ITs are built using standard analyzer. I'll create a follow-up PR to add bert-based and mbert-based IT after the ml-commons PR merged.

zane-neo (approved):

yuye-aws (approved):

PR #1094, opened 2025-01-14
Update pagination_depth datatype from int to Integer

vibrantvarun:
### Description
This PR changes pagination_depth datatype from int to Integer. We are doing as the current code is breaking the bwc on 2.x by sending pagination_depth in the older versions of Opensearch. 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
@@ -47,7 +47,7 @@ public final class HybridQuery extends Query implements Iterable<Query> {
             throw new IllegalArgumentException("collection of queries must not be empty");
         }
         Integer paginationDepth = hybridQueryContext.getPaginationDepth();
-        if (paginationDepth != null && paginationDepth == 0) {
+        if (Objects.nonNull(paginationDepth) && paginationDepth == 0) {
             throw new IllegalArgumentException("pagination_depth must not be zero");
         }
         if (Objects.isNull(filterQueries) || filterQueries.isEmpty()) {
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
@@ -46,7 +46,8 @@ public final class HybridQuery extends Query implements Iterable<Query> {
         if (subQueries.isEmpty()) {
             throw new IllegalArgumentException("collection of queries must not be empty");
         }
-        if (hybridQueryContext.getPaginationDepth() == 0) {
+        Integer paginationDepth = hybridQueryContext.getPaginationDepth();
+        if (paginationDepth != null && paginationDepth == 0) {
             throw new IllegalArgumentException("pagination_depth must not be zero");
         }
         if (Objects.isNull(filterQueries) || filterQueries.isEmpty()) {
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
@@ -55,7 +55,7 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
 
     private final List<QueryBuilder> queries = new ArrayList<>();
 
-    private int paginationDepth;
+    private Integer paginationDepth;
 
     static final int MAX_NUMBER_OF_SUB_QUERIES = 5;
     private final static int DEFAULT_PAGINATION_DEPTH = 10;
@@ -65,7 +65,7 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
         super(in);
         queries.addAll(readQueries(in));
         if (isClusterOnOrAfterMinReqVersionForPaginationInHybridQuery()) {
-            paginationDepth = in.readInt();
+            paginationDepth = in.readOptionalInt();
         }
     }
 
@@ -78,7 +78,7 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
     protected void doWriteTo(StreamOutput out) throws IOException {
         writeQueries(out, queries);
         if (isClusterOnOrAfterMinReqVersionForPaginationInHybridQuery()) {
-            out.writeInt(paginationDepth);
+            out.writeOptionalInt(paginationDepth);
         }
     }
 
@@ -109,8 +109,8 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
             queryBuilder.toXContent(builder, params);
         }
         builder.endArray();
-        if (isClusterOnOrAfterMinReqVersionForPaginationInHybridQuery()) {
-            builder.field(PAGINATION_DEPTH_FIELD.getPreferredName(), paginationDepth == 0 ? DEFAULT_PAGINATION_DEPTH : paginationDepth);
+        if (Objects.nonNull(paginationDepth)) {
+            builder.field(PAGINATION_DEPTH_FIELD.getPreferredName(), paginationDepth);
         }
         printBoostAndQueryName(builder);
         builder.endObject();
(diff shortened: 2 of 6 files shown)


vibrantvarun:
@martin-gaievski and @heemin32 Please review the PR. Thanks

heemin32 (approved):

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
```diff
@@ -46,7 +46,8 @@ public HybridQuery(final Collection<Query> subQueries, final List<Query> filterQ
         if (subQueries.isEmpty()) {
             throw new IllegalArgumentException("collection of queries must not be empty");
         }
-        if (hybridQueryContext.getPaginationDepth() == 0) {
+        Integer paginationDepth = hybridQueryContext.getPaginationDepth();
+        if (paginationDepth != null && paginationDepth == 0) {
```
martin-gaievski, line 50:
```suggestion
        if (Objects.nonNull(paginationDepth) && paginationDepth == 0) {
```
this way it's consistent with another similar check in this PR

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
```diff
@@ -46,7 +46,8 @@ public HybridQuery(final Collection<Query> subQueries, final List<Query> filterQ
...
             throw new IllegalArgumentException("pagination_depth must not be zero");
```
martin-gaievski, line 51: Do you want to log the actual value to understand if it's null or 0?
does this error message hold if the value is null? If not can you extend it and include the `null` scenario so user can understand and fix the problem.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryBuilderTests.java
```diff
@@ -426,7 +426,7 @@ public void testFromXContent_whenMultipleSubQueries_thenBuildSuccessfully() {
         assertEquals(2, queryTwoSubQueries.queries().size());
         assertTrue(queryTwoSubQueries.queries().get(0) instanceof NeuralQueryBuilder);
         assertTrue(queryTwoSubQueries.queries().get(1) instanceof TermQueryBuilder);
-        assertEquals(10, queryTwoSubQueries.paginationDepth());
+        assertEquals(10, queryTwoSubQueries.paginationDepth().intValue());
```
martin-gaievski, line 429: do we have a test scenario when pagination_depth is not set/is null?

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -109,8 +109,8 @@ protected void doXContent(XContentBuilder builder, Params params) throws IOExcep
             queryBuilder.toXContent(builder, params);
         }
         builder.endArray();
-        if (isClusterOnOrAfterMinReqVersionForPaginationInHybridQuery()) {
```
martin-gaievski, line 112: do we still need method `isClusterOnOrAfterMinReqVersionForPaginationInHybridQuery` in MinClusterVersionUtil? Looks like you're using isClusterOnOrAfterMinReqVersionForPaginationInHybridQuery == false and paginationDepth == null interchangeably

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
vibrantvarun: The value will never be null to the versions greater that or equal to 2.19.  In case if user didn't send pagination_depth in os >=2.19 then it will take the default value 10 [code ref](https://github.com/opensearch-project/neural-search/pull/1094/files#diff-e777f6753c5496193dda2794e4be74162f5d2cbb5c9f278df5cbb4bf5438a6e2R169).

src/test/java/org/opensearch/neuralsearch/query/HybridQueryBuilderTests.java
vibrantvarun: Well this test belongs to the exact same case. So when the value is not sent in the request then it will pick the default value and this is the exact thing what this assert condition validates. If you see this test case I haven't set the value of pagination_depth explicitly. So it is going null.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
vibrantvarun: So how this works is older versions of OS when there is no pagination_depth sent then we will not set that field in the builder object. But when the same method runs for version>=2.19 then in the fromXContent method we give default value 10 when we don't get pagination_depth by the user.

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
martin-gaievski: understood about the scenario. Please add the log the message part, it can be at the error level as we will fail with runtime exception in that case

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
martin-gaievski: in this case it's a combination of the version check and values being not null. Should you do OR of two? What I'm essentially trying to say - we are checking one condition in one place and another condition in another scenario, and it's not that great.

vibrantvarun: The thing is we do not use `isClusterOnOrAfterMinReqVersionForPaginationInHybridQuery` in doXContent of any other query builder. Every other place we use the same non null check. Recently expandNested param has been added by heemin in the neuralquerybuilder and it also follow the same non null check.

ryanbogan: Isn't there a way with StreamOutput and StreamInput to get the version that it's writing to or coming from?

ryanbogan: Something like this: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/indices/ModelMetadata.java#L83
https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/indices/ModelMetadata.java#L502

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -78,7 +78,7 @@ public HybridQueryBuilder(StreamInput in) throws IOException {
     protected void doWriteTo(StreamOutput out) throws IOException {
         writeQueries(out, queries);
         if (isClusterOnOrAfterMinReqVersionForPaginationInHybridQuery()) {
-            out.writeInt(paginationDepth);
+            out.writeOptionalInt(paginationDepth);
```
ryanbogan, line 81: I ran into a similar error as this with the StreamOutput on k-NN.  Using `out.getVersion()` to set what value to write might help avoid null checks elsewhere. StreamInput has a similar method to so you can set a default for the value.

Example: https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/indices/ModelMetadata.java#L83

vibrantvarun: I will address this in https://github.com/opensearch-project/neural-search/issues/1097

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
vibrantvarun: I will address this in https://github.com/opensearch-project/neural-search/issues/1097

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
vibrantvarun: I will address this in https://github.com/opensearch-project/neural-search/issues/1097

src/main/java/org/opensearch/neuralsearch/query/HybridQuery.java
vibrantvarun: Synced up offline and concluded that we do not need it

ryanbogan (approved):

vibrantvarun:
We need to merge this PR for now to unblock merging on 2.x
cc: @ryanbogan @martin-gaievski

martin-gaievski (approved):
looks good to me, thank you Varun

PR #1096, opened 2025-01-14
Use different package name between restart/rolling bwc tests

heemin32:
### Description
During bwc test debugging, intelliJ is pointing to wrong code part becase restart/rolling bwc tests are sharing same package name and class name. Separating the package name between the two so that intelliJ debugger can point to correct code part.

### Related Issues
N/A

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [X] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/neural-search/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1111, opened 2025-01-15
Revert "Use different package name between restart/rolling bwc tests …

heemin32:
…(#1096) (#1107)"

This reverts commit 173983409ba914a3e12d1423d45ef866adf723fa.


heemin32:
The error in additional tests are from OpenSearch core where it introduced new feature and the new OpenSearch artifact is not available in gradle repo but we are using the feature in our test code. The failure will go away once the new OpenSearch artifacts is available in gradle repo. https://github.com/opensearch-project/OpenSearch/commit/bf06726a4eff8584d25a81db66175f347d162278

No need to revert this PR.

Issue #480, opened 2023-10-30
Check if ml-commons plugin is installed in the Cluster


owaiskazi19:
### What is the bug?
Neural-search is using ml-commons APIs [here](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java#L143). There can be a case when the cluster doesn't have ml-commons installed and can return in failure of these APIs on neural-search side.

### How can one reproduce the bug?
1. Have a cluster running with just neural-search plugin installed.

### What is the expected behavior?
A check should be added in build.gradle [here](https://github.com/opensearch-project/neural-search/blob/main/build.gradle#L75) for ml-commons plugin.

### What is your host/environment?
Ubuntu

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._

### Do you have any additional context?
_Add any other context about the problem._

junqiu-lei:
In order to have ml-common dependency check for neural search plugin during OpenSearch cluster plugin installation, I tried to add `opensearch-ml` as extendedPlugin in neural search gralde.build [here](https://github.com/opensearch-project/neural-search/blob/main/build.gradle#L240).

After this change, it causes jar hell from org.opensearch.plugins.PluginsService.checkBundleJarHell:
```
| Caused by: java.lang.IllegalStateException: jar hell!
| class: com.google.common.annotations.Beta
| jar1: /local/home/junqiu/dev/neural-search/build/testclusters/integTest-0/distro/3.0.0-ARCHIVE/plugins/opensearch-ml/guava-32.1.3-jre.jar
| jar2: /local/home/junqiu/dev/neural-search/build/testclusters/integTest-0/distro/3.0.0-ARCHIVE/plugins/opensearch-knn/guava-32.1.3-jre.jar
```

k-NN and ML each ship Guava JAR, leading to jar-hell check failure. Because OpenSearch sees two copies of com.google.common.*, it fails with a jar-hell error. Unfortunately, I tried to relocate Guava in our own plugin, but it still leads to same jar hell from k-NN and ML.

Besides I don't find any method to disable jarhell check within plugin level. That might be the reason previously only k-NN plugin is added in the extendedPlugin list.

heemin32:
Here are a few alternatives to defining both k-NN and ml-common in the `extendedPlugins` parameter:

1. Eliminate the Guava dependency from either the k-NN or ml-common plugin to avoid the conflict.
[...]

PR #1119, opened 2025-01-17
Add dependent plugins validation during plugin load

junqiu-lei:
### Description
Add dependent plugins validation during neural search plugin load stage.

### Related Issues
Closes https://github.com/opensearch-project/neural-search/issues/480

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
+++ b/src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
@@ -14,6 +14,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.function.Supplier;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.opensearch.client.Client;
 import org.opensearch.cluster.metadata.IndexNameExpressionResolver;
 import org.opensearch.cluster.service.ClusterService;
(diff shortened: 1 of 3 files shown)


heemin32 (dismissed):

heemin32:
Thanks for the PR.

Could you confirm it will fail in both cases?
1. Attempting to install the neural plugin without having the ml-common plugin already installed.  
2. Installing both the ml-common and neural plugins, then removing the ml-common plugin and restarting the cluster.

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
```diff
@@ -86,6 +87,14 @@ public class NeuralSearch extends Plugin implements ActionPlugin, SearchPlugin,
     private final ScoreCombinationFactory scoreCombinationFactory = new ScoreCombinationFactory();
     public static final String EXPLANATION_RESPONSE_KEY = "explanation_response";
 
+    @VisibleForTesting
+    public static final Map<String, String> DEPENDENT_PLUGINS = Map.of(
+        "opensearch-knn",
+        "org.opensearch.knn.plugin.KNNPlugin",
+        "opensearch-ml-commons",
+        "org.opensearch.ml.client.MachineLearningClient"
```
yuye-aws, line 95: I do not have much context. Can you please explain why `org.opensearch.knn.plugin.KNNPlugin` and `org.opensearch.ml.client.MachineLearningClient` still needs check even if both `opensearch-knn` and `opensearch-ml-commons` are available

heemin32: `opensearch-knn` is just a plugin name we are using in a log and `org.opensearch.knn.plugin.KNNPlugin` is the actual class that we are validating its existence.

I think this code might look confusing. Can we change to something like this?
```
public enum DependentPlugin {
  private String klass;

  ML_COMMON("org.opensearch.ml.client.MachineLearningClient"),
  KNN("org.opensearch.knn.plugin.KNNPlugin");

  DependentPlugin(final String klass) {
    this.klass = klass;
  }

  public validateAvailability() {
    try {
               Class.forName(klass);
               log.info("Plugin [{}] is installed", this.name());
     } catch (ClassNotFoundException e) {
                throw new IllegalStateException("Neural Search plugin requires the " + this.name() + " plugin to be installed", e);
     }
  }
}
```

junqiu-lei: Updated to use enum DependentPlugin.

junqiu-lei:
> Thanks for the PR.
> 
> Could you confirm it will fail in both cases?
> 
> 1. Attempting to install the neural plugin without having the ml-common plugin already installed.
> 2. Installing both the ml-common and neural plugins, then removing the ml-common plugin and restarting the cluster.
> 
> Because we added knn and ml common code as dependency through gradle, this `Class.forName(org.opensearch.knn.plugin.KNNPlugin)` might fail to check the actual installation of the plugin in the OpenSearch cluster.

@heemin32 I found that at cluster install plugin stage, the validation from this PR can be passed even the ml-commons plugin is removed from OpenSearch bundle.

Going to close this PR, will raise another PR by adding ml-commons in the neural search extended plugin, which will require code changes in k-NN and ml-commons plugins to use some libs by compile only, this way can resolve all OpenSearch jar hell checks.

PR #1093, opened 2025-01-14
Fix race condition in TextImageEmbeddingProcessor integration tests

q-andy:
### Description
The integration tests for TextImageEmbeddingProcessor create and delete resources with the same name leading to issues when running integration tests in parallel, e.g. [failed Jenkins run](https://build.ci.opensearch.org/blue/organizations/jenkins/integ-test/detail/integ-test/9259/pipeline/130/)

```
REPRODUCE WITH: ./gradlew ':integTest' --tests "org.opensearch.neuralsearch.processor.TextImageEmbeddingProcessorIT.testEmbeddingProcessor_whenIngestingDocumentWithOrWithoutSourceMatchingMapping_thenSuccessful" -Dtests.seed=E062306E78BCA0FB -Dtests.security.manager=false -Dtests.locale=zh-SG -Dtests.timezone=America/Vancouver -Druntime.java=21

Suite: Test class org.opensearch.neuralsearch.processor.TextImageEmbeddingProcessorIT

  2> REPRODUCE WITH: ./gradlew ':integTest' --tests "org.opensearch.neuralsearch.processor.TextImageEmbeddingProcessorIT.testEmbeddingProcessor_whenIngestingDocumentWithOrWithoutSourceMatchingMapping_thenSuccessful" -Dtests.seed=E062306E78BCA0FB -Dtests.security.manager=false -Dtests.locale=zh-SG -Dtests.timezone=America/Vancouver -Druntime.java=21
  2> org.opensearch.client.ResponseException: method [DELETE], host [http://localhost:9200/], URI [/_ingest/pipeline/ingest-pipeline], status line [HTTP/1.1 404 Not Found]
    {"error":{"root_cause":[{"type":"resource_not_found_exception","reason":"pipeline [ingest-pipeline] is missing"}],"type":"resource_not_found_exception","reason":"pipeline [ingest-pipeline] is missing"},"status":404}
        at __randomizedtesting.SeedInfo.seed([E062306E78BCA0FB:CD13011798778F4D]:0)
        at app//org.opensearch.client.RestClient.convertResponse(RestClient.java:479)
        at app//org.opensearch.client.RestClient.performRequest(RestClient.java:371)
        at app//org.opensearch.client.RestClient.performRequest(RestClient.java:346)
        at app//org.opensearch.neuralsearch.BaseNeuralSearchIT.deletePipeline(BaseNeuralSearchIT.java:1440)
[...]

Issue #1142, opened 2025-01-24
[BUG] BWC Rolling upgrade tests fails for SparseEncoder Processor during Batch Ingestion


vibrantvarun:
### What is the bug?
BatchIngestionIT.testBatchIngestion_SparseEncodingProcessor_E2EFlow test is failing with the following error.
```
> Task :qa:rolling-upgrade:testAgainstOneThirdUpgradedCluster
REPRODUCE WITH: ./gradlew ':qa:rolling-upgrade:testAgainstOneThirdUpgradedCluster' --tests "org.opensearch.neuralsearch.bwc.rolling.BatchIngestionIT.testBatchIngestion_SparseEncodingProcessor_E2EFlow" -Dtests.seed=801B4B74838557A -Dtests.security.manager=false -Dtests.bwc.version=2.19.0-SNAPSHOT -Dtests.locale=sw-TZ -Dtests.timezone=America/Bahia_Banderas -Druntime.java=21
Suite: Test class org.opensearch.neuralsearch.bwc.rolling.BatchIngestionIT
  2> Jan 24, 2025 6:12:44 PM org.apache.lucene.internal.vectorization.VectorizationProvider lookup
  2> WARNING: Java vector incubator module is not readable. For optimal vector performance, pass '--add-modules jdk.incubator.vector' to enable Vector API.
  2> SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
  2> SLF4J: Defaulting to no-operation (NOP) logger implementation
  2> SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
  2> REPRODUCE WITH: ./gradlew ':qa:rolling-upgrade:testAgainstOneThirdUpgradedCluster' --tests "org.opensearch.neuralsearch.bwc.rolling.BatchIngestionIT.testBatchIngestion_SparseEncodingProcessor_E2EFlow" -Dtests.seed=801B4B74838557A -Dtests.security.manager=false -Dtests.bwc.version=2.19.0-SNAPSHOT -Dtests.locale=sw-TZ -Dtests.timezone=America/Bahia_Banderas -Druntime.java=21
  2> java.lang.AssertionError: expected:<10> but was:<8>
        at __randomizedtesting.SeedInfo.seed([801B4B74838557A:8487AE9ACE1A1513]:0)
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.failNotEquals(Assert.java:835)
[...]

PR #533, opened 2024-01-07
BWC tests for Multimodal Search, Hybrid Search and Neural Sparse Search

vibrantvarun:
### Description
The PR consists of BWC tests for Multimodal search, Hybrid search and Neural Sparse Search.

### Issues Resolved
[https://github.com/opensearch-project/neural-search/issues/202](https://github.com/opensearch-project/neural-search/issues/202)

### Check List
- [x] New functionality includes testing.
    - [x] All tests pass
- [x] New functionality has been documented.
    - [x] New functionality has javadoc added
- [x] Commits are signed as per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/k-NN/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1154, opened 2025-01-30
Handle exceptions gracefully when delete non-existent resources during integ test resource clean up

### LINKED ISSUE #1098: [Infrastructure] Wipe integration test resources and surface exceptions gracefully
Currently many of our integration tests for our processors (text embedding, text image embedding, sparse encoding) follow the following pattern to create and cleanup resources
```
public void processor_integrationTest() {
    try {
        // load model
        // create pipeline/index
        // do tests/asserts
    } finally {
        // cleanup pipeline, index, model
    }
}
```

However, this causes issues like #1091 where if the integration tests fail in the try block before creating the pipeline, it will enter the finally block and attempt to delete the nonexistent pipeline. This causes an exception in the cleanup and masks the true exception, such as #1093 where the root issue is a memory circuit breaker trigger which is unclear from the test report.

We should rework our integration test cleanup so we can:
- Log errors more clearly during resource cleanup to avoid confusion
- Resource cleanup after every test rather than a part of every test, so cleanup won't directly cause test failures

weijia-aws:
### Description
When integ test runs, the test will create model, index, pipeline, etc., then perform some assertions, finally clean up the resources. 

```
public void processor_integrationTest() {
    try {
        // load model
        // create pipeline/index
        // do tests/asserts
    } finally {
        // cleanup pipeline, index, model
    }
}
```

If somehow resource creation fail and receive an exception, the test will enter the finally block and try to delete the resources. Since these resources don't exist, exception will be thrown when try to delete them, see [issue 1091](https://github.com/opensearch-project/neural-search/issues/1091). I was able to reproduce the exception by following https://github.com/opensearch-project/neural-search/pull/1093#issuecomment-2590760941. So in this case, we should gracefully handle such NOT FOUND exceptions

### Related Issues
Resolves #1098 

Related issue: #1091 and #1093

### Check List

### DIFF
--- a/qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/HybridSearchIT.java
+++ b/qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/HybridSearchIT.java
@@ -55,28 +55,30 @@ public class HybridSearchIT extends AbstractRestartUpgradeRestTestCase {
     private void validateNormalizationProcessor(final String fileName, final String pipelineName, final String searchPipelineName)
         throws Exception {
         waitForClusterHealthGreen(NODES_BWC_CLUSTER);
-        super.ingestPipelineName = pipelineName;
-        super.searchPipelineName = searchPipelineName;
-
         if (isRunningAgainstOldCluster()) {
-            super.modelId = uploadTextEmbeddingModel();
-            loadModel(super.modelId);
-            createPipelineProcessor(super.modelId, pipelineName);
+            String modelId = uploadTextEmbeddingModel();
+            loadModel(modelId);
+            createPipelineProcessor(modelId, pipelineName);
             createIndexWithConfiguration(
-                super.indexName,
+                getIndexNameForTest(),
                 Files.readString(Path.of(classLoader.getResource(fileName).toURI())),
                 pipelineName
             );
-            addDocuments(super.indexName, true);
+            addDocuments(getIndexNameForTest(), true);
             createSearchPipeline(searchPipelineName);
         } else {
-            super.modelId = getModelId(getIngestionPipeline(pipelineName), TEXT_EMBEDDING_PROCESSOR);
-            loadModel(super.modelId);
-            addDocuments(super.indexName, false);
-            HybridQueryBuilder hybridQueryBuilder = getQueryBuilder(super.modelId, null, null, null);
-            validateTestIndex(super.indexName, searchPipelineName, hybridQueryBuilder);
-            hybridQueryBuilder = getQueryBuilder(super.modelId, Boolean.FALSE, Map.of("ef_search", 100), RescoreContext.getDefault());
-            validateTestIndex(super.indexName, searchPipelineName, hybridQueryBuilder);
+            String modelId = null;
+            try {
+                modelId = getModelId(getIngestionPipeline(pipelineName), TEXT_EMBEDDING_PROCESSOR);
+                loadModel(modelId);
+                addDocuments(getIndexNameForTest(), false);
+                HybridQueryBuilder hybridQueryBuilder = getQueryBuilder(modelId, null, null, null);
+                validateTestIndex(getIndexNameForTest(), searchPipelineName, hybridQueryBuilder);
+                hybridQueryBuilder = getQueryBuilder(modelId, Boolean.FALSE, Map.of("ef_search", 100), RescoreContext.getDefault());
+                validateTestIndex(getIndexNameForTest(), searchPipelineName, hybridQueryBuilder);
+            } finally {
+                wipeOfTestResources(getIndexNameForTest(), pipelineName, modelId, searchPipelineName);
+            }
         }
     }
 
--- a/qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/HybridSearchIT.java
+++ b/qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/HybridSearchIT.java
@@ -67,18 +67,13 @@ public class HybridSearchIT extends AbstractRestartUpgradeRestTestCase {
             addDocuments(getIndexNameForTest(), true);
             createSearchPipeline(searchPipelineName);
         } else {
-            String modelId = null;
-            try {
-                modelId = getModelId(getIngestionPipeline(pipelineName), TEXT_EMBEDDING_PROCESSOR);
-                loadModel(modelId);
-                addDocuments(getIndexNameForTest(), false);
-                HybridQueryBuilder hybridQueryBuilder = getQueryBuilder(modelId, null, null, null);
-                validateTestIndex(getIndexNameForTest(), searchPipelineName, hybridQueryBuilder);
-                hybridQueryBuilder = getQueryBuilder(modelId, Boolean.FALSE, Map.of("ef_search", 100), RescoreContext.getDefault());
-                validateTestIndex(getIndexNameForTest(), searchPipelineName, hybridQueryBuilder);
-            } finally {
-                wipeOfTestResources(getIndexNameForTest(), pipelineName, modelId, searchPipelineName);
-            }
+            String modelId = getModelId(getIngestionPipeline(pipelineName), TEXT_EMBEDDING_PROCESSOR);
+            loadModel(modelId);
+            addDocuments(getIndexNameForTest(), false);
+            HybridQueryBuilder hybridQueryBuilder = getQueryBuilder(modelId, null, null, null);
+            validateTestIndex(getIndexNameForTest(), searchPipelineName, hybridQueryBuilder);
+            hybridQueryBuilder = getQueryBuilder(modelId, Boolean.FALSE, Map.of("ef_search", 100), RescoreContext.getDefault());
+            validateTestIndex(getIndexNameForTest(), searchPipelineName, hybridQueryBuilder);
         }
     }
 
--- a/qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/HybridSearchIT.java
+++ b/qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/HybridSearchIT.java
@@ -55,25 +55,28 @@ public class HybridSearchIT extends AbstractRestartUpgradeRestTestCase {
     private void validateNormalizationProcessor(final String fileName, final String pipelineName, final String searchPipelineName)
         throws Exception {
         waitForClusterHealthGreen(NODES_BWC_CLUSTER);
+        super.ingestPipelineName = pipelineName;
+        super.searchPipelineName = searchPipelineName;
+
         if (isRunningAgainstOldCluster()) {
-            String modelId = uploadTextEmbeddingModel();
-            loadModel(modelId);
-            createPipelineProcessor(modelId, pipelineName);
+            super.modelId = uploadTextEmbeddingModel();
+            loadModel(super.modelId);
+            createPipelineProcessor(super.modelId, pipelineName);
             createIndexWithConfiguration(
-                getIndexNameForTest(),
+                super.indexName,
                 Files.readString(Path.of(classLoader.getResource(fileName).toURI())),
                 pipelineName
             );
-            addDocuments(getIndexNameForTest(), true);
+            addDocuments(super.indexName, true);
             createSearchPipeline(searchPipelineName);
         } else {
-            String modelId = getModelId(getIngestionPipeline(pipelineName), TEXT_EMBEDDING_PROCESSOR);
-            loadModel(modelId);
-            addDocuments(getIndexNameForTest(), false);
-            HybridQueryBuilder hybridQueryBuilder = getQueryBuilder(modelId, null, null, null);
-            validateTestIndex(getIndexNameForTest(), searchPipelineName, hybridQueryBuilder);
-            hybridQueryBuilder = getQueryBuilder(modelId, Boolean.FALSE, Map.of("ef_search", 100), RescoreContext.getDefault());
-            validateTestIndex(getIndexNameForTest(), searchPipelineName, hybridQueryBuilder);
+            super.modelId = getModelId(getIngestionPipeline(pipelineName), TEXT_EMBEDDING_PROCESSOR);
+            loadModel(super.modelId);
+            addDocuments(super.indexName, false);
+            HybridQueryBuilder hybridQueryBuilder = getQueryBuilder(super.modelId, null, null, null);
+            validateTestIndex(super.indexName, searchPipelineName, hybridQueryBuilder);
+            hybridQueryBuilder = getQueryBuilder(super.modelId, Boolean.FALSE, Map.of("ef_search", 100), RescoreContext.getDefault());
+            validateTestIndex(super.indexName, searchPipelineName, hybridQueryBuilder);
         }
     }
 
--- a/src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
+++ b/src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
@@ -134,9 +134,13 @@ public abstract class BaseNeuralSearchIT extends OpenSearchSecureRestTestCase {
         NeuralSearchClusterUtil.instance().initialize(clusterService);
     }
 
+    // Wipe of all the resources after execution of the tests.
     @After
-    public void cleanUp() throws IOException {
-        wipeOfTestResources();
+    public void cleanUp() {
+        deleteExistingIngestionPipelines();
+        deleteExistingSearchPipelines();
+        deleteExistingModels();
+        deleteExistingIndices();
     }
 
     protected ThreadPool setUpThreadPool() {
@@ -1431,7 +1435,8 @@ public abstract class BaseNeuralSearchIT extends OpenSearchSecureRestTestCase {
      * Retrieve all available index
      * @return index as a list of map object
      */
-    protected List<Map<String, Object>> retrieveIndices() throws IOException {
+    @SneakyThrows
+    protected List<Map<String, Object>> retrieveIndices() {
         Request request = new Request("GET", "/_cat/indices?format=json&expand_wildcards=all");
         Response response = client().performRequest(request);
         MediaType mediaType = MediaType.fromMediaType(response.getEntity().getContentType());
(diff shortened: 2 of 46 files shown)


heemin32:
Can we expand the scope of this PR to shift the responsibility for resource cleanup from individual tests to the framework? This way, each test wouldn’t need to handle resource cleanup individually.

vibrantvarun:
> Can we expand the scope of this PR to shift the responsibility for resource cleanup from individual tests to the framework? This way, each test wouldn’t need to handle resource cleanup individually.

I would suggest not to do that. As Integ tests are meant to be run in isolation mode. With resource cleanup at framework level, it will create issues where test will start sharing resources with each other and can cause critical issues to be skipped at integ test level because the resource might be created by some other method and used by some other method.

vibrantvarun (approved):

vibrantvarun:
LGTM

minalsha:
https://github.com/opensearch-project/neural-search/pull/1154#issuecomment-2627908252

+1 to Varun's comment here.

heemin32:
> @heemin32 I would suggest not to do that. As Integ tests are meant to be run in isolation mode. With resource cleanup at framework level, it will create issues where test will start sharing resources with each other and can cause critical issues to be skipped at integ test level because the resource might be created by some other method and used by some other method.

Have you seen any such issue in k-nn repo where the resource clean up is handling in framework level? The resource sharing issue can happen when clean up is happening in individual test level when developer missed to clean up the resource in a test.

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -1504,23 +1505,32 @@ protected void wipeOfTestResources(
...
+            }
+            if (indexName != null) {
+                deleteIndex(indexName);
+            }
+        } catch (ResponseException e) {
+            // It's possible that test fails during resources (index, model, pipeline, etc.) creation, when clean up
+            // these resources after test run, the delete methods will throw ResponseException with 404 Not Found code
+            // In this case, we just need to ignore this exception, for other exceptions, continue throwing
+            if (RestStatus.fromCode(e.getResponse().getStatusLine().getStatusCode()) != RestStatus.NOT_FOUND) {
+                throw e;
```
heemin32, line 1532: This exception will still obscure the original exception. Therefore, I believe resource cleanup should be managed in the `@After` method rather than in the `finally` block.

vibrantvarun: So are you saying we should shift this cleanUp at every integ class level?

vibrantvarun: When I developed the framework you suggested to do it in in current way https://github.com/opensearch-project/neural-search/pull/533#discussion_r1465646182

heemin32: cleanUp at every individual test level but inside base class so that each test class don't repeat the same code. https://github.com/opensearch-project/k-NN/blob/main/src/testFixtures/java/org/opensearch/knn/ODFERestTestCase.java#L152

vibrantvarun: So it will mean just removal of code under finally block and move it in @After. The cleanUp method will go to base class so it in a particular integ class it will look little better. But underlying work still remains the same

vibrantvarun: Is my understand correct?

heemin32: Yes. However, there is one more difference. When it fails inside @after method, it won't obscure the exception thrown during test I believe?

weijia-aws: > This exception will still obscure the original exception.

Yes, the original exception was overriden by the exception that thrown in the finally block, hence we were not able to see the original exception in the first place. 

I will make more code change to remove the finally block and add it to an @After method

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -1504,23 +1505,32 @@ protected void wipeOfTestResources(
...
-        }
-        if (modelId != null) {
-            try {
-                deleteModel(modelId);
-            } catch (AssertionError e) {
-                // sometimes we have flaky test that the model state doesn't change after call undeploy api
-                // for this case we can call undeploy api one more time
-                deleteModel(modelId);
+        try {
+            if (ingestPipeline != null) {
```
q-andy, line 1509: I agree with the discussion below to move cleanup to an @After block, but I want to point out that the root cause is that our integ tests call this method and pass in static constant Strings directly into ingestPipeline/searchPipeline/indexName regardless of whether or not the resources were created like here: https://github.com/opensearch-project/neural-search/blob/main/src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorIT.java#L76

In those conditions these blocks are never triggered. Not saying we need to change that right now, but pointing out with an @After block we still don't address the root cause here, just move the exception where we can't see it?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
q-andy: Even if we move to @after block the root issue is that we shouldn't be calling this method passing in resources we haven't created yet right? The method is still called incorrectly, we just hide the exception?

heemin32: > Even if we move to @after block the root issue is that we shouldn't be calling this method passing in resources we haven't created yet right? The method is still called incorrectly, we just hide the exception?

You are right. When we moved to @after block, we won't depend on static value of resource name but we will make an api call to list all available resources and clean them up.

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -1410,6 +1427,80 @@ protected void reindex(String fromIndexName, String toIndexName) throws Exceptio
...
+    }
+
+    /**
+     * Retrieve all existing pipelines or a specific pipeline
+     * @param pipelineType _ingest for retrieving ingest pipelines, _search for retrieving search pipelines
+     * @param pipelineName a specific pipeline to retrieve, if value is empty, it returns all pipelines
+     * @return get pipeline response as a map object
+     */
+    @SneakyThrows(ParseException.class)
+    protected Map<String, Object> retrievePipelines(final String pipelineType, final String pipelineName) throws IOException {
```
q-andy, line 1485: Great work, I'm curious if there's any increased latency for the integ tests by adding extra get requests per test? Is the runtime for all the integ tests impacted?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -1410,6 +1427,80 @@ protected void reindex(String fromIndexName, String toIndexName) throws Exceptio
...
+    }
+
+    /**
+     * Fetch all existing indices and clean up, note that some system generated indices are skipped
+     */
+    protected void deleteExistingIndices() throws IOException {
+        List<Map<String, Object>> indices = retrieveIndices();
+        for (Map<String, Object> index : indices) {
+            final String indexName = (String) index.get("index");
+            if (!shouldDeleteIndex(indexName)) {
```
weijia-aws, line 1465: I will rename this method to `shouldNotDeleteIndex`

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
heemin32: The latency should be minimal and it is generally okay to have some slowness for testing.

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -1498,29 +1651,40 @@ protected Float getFeatureFieldCompressedNumber(final Float originNumber) {
...
     // Wipe of all the resources after execution of the tests.
-    protected void wipeOfTestResources(
-        final String indexName,
-        final String ingestPipeline,
-        final String modelId,
-        final String searchPipeline
-    ) throws IOException {
-        if (ingestPipeline != null) {
-            deletePipeline(ingestPipeline);
+    protected void wipeOfTestResources() throws IOException {
```
heemin32, line 1654: Can we remove this method entirely so that people don't call this explicitly?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -1498,29 +1651,40 @@ protected Float getFeatureFieldCompressedNumber(final Float originNumber) {
...
+        try {
+            deleteExistingIngestionPipelines();
+        } catch (ResponseException e) {
+            // It's possible that test fails during resources (index, model, pipeline, etc.) creation, when clean up
+            // these resources after test run, the delete methods will throw ResponseException with 404 Not Found code
+            // In this case, we just need to ignore this exception, for other exceptions, continue throwing
+            if (RestStatus.NOT_FOUND != RestStatus.fromCode(e.getResponse().getStatusLine().getStatusCode())) {
```
heemin32, line 1661: Now we don't need to consume not found exception silently.

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -1410,6 +1427,80 @@ protected void reindex(String fromIndexName, String toIndexName) throws Exceptio
...
+
+    /**
+     * Retrieve all existing pipelines or a specific pipeline
+     * @param pipelineType _ingest for retrieving ingest pipelines, _search for retrieving search pipelines
+     * @param pipelineName a specific pipeline to retrieve, if value is empty, it returns all pipelines
+     * @return get pipeline response as a map object
+     */
+    @SneakyThrows(ParseException.class)
+    protected Map<String, Object> retrievePipelines(final String pipelineType, final String pipelineName) throws IOException {
+        Request request = new Request("GET", "/" + pipelineType + "/pipeline/" + (StringUtils.isEmpty(pipelineName) ? "" : pipelineName));
```
junqiu-lei, line 1486:
```suggestion
    String endpoint = String.format("/%s/pipeline/%s", pipelineType, 
        Optional.ofNullable(pipelineName).orElse(""));
    Request request = new Request("GET", endpoint);
```

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
weijia-aws: Sure

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
weijia-aws: So I ran the integ tests with and without this commit. Total runtime is:

Previous: 8m3.59s
Now: 8m2.34s 

The latency is really minimal

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
weijia-aws: We still need it, at least for pipelines. If there's no pipeline exists, and we make a get pipelines call, the API will throw us this exception. There's an issue opened as well in core: https://github.com/opensearch-project/OpenSearch/issues/15917

heemin32: However, do we make get call for non existing pipeline?

weijia-aws: There're some test cases that don't create ingest pipeline, such as this [one](https://github.com/opensearch-project/neural-search/blob/main/src/test/java/org/opensearch/neuralsearch/query/HybridQuerySortIT.java), there're some test cases that don't create search pipeline, such as [this](https://github.com/opensearch-project/neural-search/blob/main/src/test/java/org/opensearch/neuralsearch/query/NeuralQueryIT.java). Since we want to run this clean up method after every test, we need to make a get call for every resource

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
weijia-aws: Good suggestion, this is more readable for handling null values, but it's not able to handle strings with empty space.  However I think it is fine as this is only used in Integ tests. Will adopt the suggestion

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
heemin32: However, don't we call list pipeline API to get all available pipeline instead of getting individual pipline by its name?

weijia-aws: We do call the API to get all available pipelines, and it throws exception if none exist

weijia-aws: If [this issue](https://github.com/opensearch-project/OpenSearch/issues/15917) can be closed, then we don't have to handle this exception

heemin32: Can we handle that exception inside a method to get a list of pipeline instead of here?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -1498,29 +1651,40 @@ protected Float getFeatureFieldCompressedNumber(final Float originNumber) {
...
-                deleteModel(modelId);
-            } catch (AssertionError e) {
-                // sometimes we have flaky test that the model state doesn't change after call undeploy api
-                // for this case we can call undeploy api one more time
-                deleteModel(modelId);
+
+        try {
+            deleteExistingModels();
+        } catch (ResponseException e) {
+            if (RestStatus.NOT_FOUND != RestStatus.fromCode(e.getResponse().getStatusLine().getStatusCode())) {
```
heemin32, line 1677: I guess we don't need this condition check?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -1498,29 +1651,40 @@ protected Float getFeatureFieldCompressedNumber(final Float originNumber) {
...
+                throw e;
             }
         }
-        if (indexName != null) {
-            deleteIndex(indexName);
+
+        try {
+            deleteExistingIndices();
+        } catch (ResponseException e) {
+            if (RestStatus.NOT_FOUND != RestStatus.fromCode(e.getResponse().getStatusLine().getStatusCode())) {
```
heemin32, line 1685: I guess we don't need this condition check?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -1498,29 +1651,40 @@ protected Float getFeatureFieldCompressedNumber(final Float originNumber) {
...
+                throw e;
+            }
         }
-        if (searchPipeline != null) {
-            deleteSearchPipeline(searchPipeline);
+
+        try {
+            deleteExistingSearchPipelines();
+        } catch (ResponseException e) {
+            if (RestStatus.NOT_FOUND != RestStatus.fromCode(e.getResponse().getStatusLine().getStatusCode())) {
```
heemin32, line 1669: Same as above.

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
weijia-aws: Of course

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
weijia-aws: We don't need it if we move the exception handling logic to the get call

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
weijia-aws: No, will remove

heemin32 (commented):
LGTM. Thanks!

heemin32 (dismissed):

heemin32:
Test is failing. Please take a look.

weijia-aws:
> Test is failing. Please take a look.

qa tests are failing because: testAgainstNewCluster depends on resources that created in testAgainstOldCluster, however when testAgainstOldCluster finishes, all resources are cleaned up, resulting in testAgainstNewCluster tests fail with `java.lang.NullPointerException` or Resource not found exception. In order to fix this, we will need to re-create resources in testAgainstNewCluster tasks

weijia-aws:
> In order to fix this, we will need to re-create resources in testAgainstNewCluster tasks

We can also modify the cleanUp method to not delete resources if running against old cluster?

weijia-aws:
> We can also modify the cleanUp method to not delete resources if running against old cluster?

I think we should do this, as this is how the current behavior is, and requires less code changes comparing to re-create resources

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/HybridSearchIT.java
```diff
@@ -55,30 +55,28 @@ public void testNormalizationProcessor_whenIndexWithSingleShard_E2EFlow() throws
...
         throws Exception {
         waitForClusterHealthGreen(NODES_BWC_CLUSTER);
+        super.ingestPipelineName = pipelineName;
+        super.searchPipelineName = searchPipelineName;
+
         if (isRunningAgainstOldCluster()) {
-            String modelId = uploadTextEmbeddingModel();
-            loadModel(modelId);
-            createPipelineProcessor(modelId, pipelineName);
+            super.modelId = uploadTextEmbeddingModel();
```
heemin32, line 62: I suggest not making any changes to the BWC tests, as this approach may not work. For instance, there are two tests: `testNormalizationProcessor_whenIndexWithMultipleShards_E2EFlow` and `testNormalizationProcessor_whenIndexWithSingleShard_E2EFlow`. Each test uploads a model, but the model ID from the previous test is overwritten by the next, resulting in a failure to delete the all resources.

weijia-aws: I don't think the resources deletion logic is changed. I simply moved the cleanup logic from finally block to an After class method. In order to do that, we just need to set the resource names according. 

> but the model ID from the previous test is overwritten by the next

The model gets cleanup already when the test finishes, so it's fine and has to be overwritten

heemin32: > The model gets cleanup already when the test finishes, so it's fine and has to be overwritten

In old cluster, 
Test 1 create model A and set super.modelId. The test complete without deleting the model.
Test 2 create model B and override super.modelId. The test complete without deleting the model.

In new cluster,
Test 1 use model B because modelId is overwritten by Test 2. The test complete. If model B deletion happen here before test 2, the test 2 might fail because there is no model B.
Test 2 use model B. The test complete. Delete the model B.

Now, the model A stay without being deleted.

weijia-aws: In new cluster, we are getting model id by pipeline name then assign the value of `super.modelId`, since model A and model B are created in different pipelines, we shouldn't get the wrong model in the new cluster for these two test cases? The behavior is the same before this commit, see [code](https://github.com/opensearch-project/neural-search/blob/main/qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/HybridSearchIT.java#L72C27-L72C37)

weijia-aws: Each test is still isolated, it's just we need to update the resource name for each test case so that resources can be properly clean up. I agree this is error-prone as we need to maintain the value assignment logic for each test. But this approach align with the purpose of this issue.

How about we move the After method to each test class instead of the base class? will it make the logic clearer?

heemin32: Yes. It is isolated in class level. However, it is not isolated in method level. That being said, moving After method to each test class has no difference.

I understand that bwc also has similar issue where the exception during resource clean up will disguise the original exception. However, the nature of how bwc test works is completely different from integ test. We might need to think better option than this.

One option is creating another gradle task for resource clean up which will be executed after all tests are completed.

We can close this PR by focusing only in integ test but not bwc test. For bwc test, we can handle in other PR.

weijia-aws: Sounds good, glad we have another PR to fix

weijia-aws: Resolving the conversation as new commit is published

heemin32 (approved):
LGTM. Thanks!

heemin32:
The failure in bwc is a known issue.
https://github.com/opensearch-project/neural-search/issues/1142

PR #1171, opened 2025-02-05
refactor bwc test suite to re-use existing resources between tests

will-hwang:
### Description
Currently, models and pipelines are re-created and re-deployed per test, leading to redundant model loads and pipeline creations. This change avoids this redundancy by reusing the created resource between test cases.

### Related Issues

### Check List

### DIFF
--- a/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/SemanticSearchIT.java
+++ b/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/SemanticSearchIT.java
@@ -13,7 +13,7 @@ import static org.opensearch.neuralsearch.util.TestUtils.getModelId;
 import org.opensearch.neuralsearch.query.NeuralQueryBuilder;
 
 public class SemanticSearchIT extends AbstractRollingUpgradeTestCase {
-    private static final String PIPELINE_NAME = "nlp-pipeline-text-embedding";
+    private static final String PIPELINE_NAME = "nlp-pipeline";
     private static final String TEST_FIELD = "passage_text";
     private static final String TEXT = "Hello world";
     private static final String TEXT_MIXED = "Hello world mixed";
--- a/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/SemanticSearchIT.java
+++ b/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/SemanticSearchIT.java
@@ -13,7 +13,7 @@ import static org.opensearch.neuralsearch.util.TestUtils.getModelId;
 import org.opensearch.neuralsearch.query.NeuralQueryBuilder;
 
 public class SemanticSearchIT extends AbstractRollingUpgradeTestCase {
-    private static final String PIPELINE_NAME = "nlp-pipeline";
+    private static final String PIPELINE_NAME = "nlp-pipeline-text-embedding";
     private static final String TEST_FIELD = "passage_text";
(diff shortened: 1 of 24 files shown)


vibrantvarun:
We have faced earlier issues with this where during the test run due to shared names the pipeline either used to get deleted or wrong information has been shared amongst test.

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/SemanticSearchIT.java
```diff
@@ -28,9 +28,9 @@ public void testSemanticSearch_E2EFlow() throws Exception {
         waitForClusterHealthGreen(NODES_BWC_CLUSTER, 90);
         switch (getClusterType()) {
             case OLD:
-                modelId = uploadTextEmbeddingModel();
+                modelId = getOrUploadTextEmbeddingModel(getIngestionPipeline(PIPELINE_NAME), TEXT_EMBEDDING_PROCESSOR);
```
vibrantvarun, line 31: In case of OLD it will always upload. Then why it is changed to getOrUpload?

heemin32:
I suggest reusing only the model. Pipeline is cheap to create. We likely need just one model per type: `text`, `text_embedding`, and `sparse`. There should be a common method to retrieve these models based solely on the type, without requiring the caller to specify anything else. This will ensure the models can be shared correctly without errors.  

@vibrantvarun The reason for reusing the model is that during BWC tests, multiple tests deploy the same model simultaneously, leading to model deployment failures during node upgrades. This, in turn, causes flaky test failures.

vibrantvarun:
@heemin32 I think test resources created for any test should be just meant to run it. The root cause of the model deployment failure is not the redeployment of the model. It should be something else.

heemin32:
>  I think test resources created for any test should be just meant to run it.

Our goal isn’t to test model deployment—that should be covered by `ml-common`. For us, model deployment is merely a prerequisite for running our feature tests.  

Model deployment is resource-intensive, and if each test deploys its own model, the test cluster might not be large enough to handle them all at once. Sharing models across tests shouldn’t affect test coverage or validity in any way.  

The primary objective here is to avoid flaky test failures caused by model deployment issues. We also want to ensure that problems in `ml-common`(issue with model deployment) don’t impact our test velocity. To further minimize test flakiness related to model deployment, we could even consider using lightweight or mock models.

vibrantvarun:
@heemin32 in bwc we do test the model deployed in older version of node still exists in new version of node. Because of that test we are always able to find critical issues with ml-commons just before the release.

vibrantvarun:
FYI: ml-commons till data does not have BWC framework in place. We as neural-search catch issues in BWC tests.

vibrantvarun:
@heemin32 I am aligned with you on reducing test flakiness but we need to find the actual root cause. Here we are trying to reduce overhead on us by doing less model deployment at the cost of removing model deployment test in bwc.

heemin32:
> @heemin32 I am aligned with you on reducing test flakiness but we need to find the actual root cause. Here we are trying to reduce overhead on us by doing less model deployment at the cost of removing model deployment test in bwc. Model deployment is part of each feature we release.

The main objective of BWC tests is to ensure backward compatibility, not to conduct load testing. Currently, we are deploying the same text embedding model multiple times—once for each test. This approach doesn’t align with real customer scenarios, as they typically wouldn’t deploy the same model repeatedly. Reusing the model across tests will not compromise the purpose of the BWC tests but will help stabilize them, allowing us to focus on what truly matters.

martin-gaievski:
@will-hwang are you still working on this PR after neural got migrated to 3.0-alpha?

PR #1178, opened 2025-02-07
code refactoring on MLCommonsClientAccessor request

### LINKED ISSUE #790: [PROPOSAL] Refactor `MLCommonsClientAccessor` to make it open for extension
### What are you proposing?
Cf. https://github.com/opensearch-project/neural-search/pull/710#discussion_r1635185058

This is a requirement from @martin-gaievski.

The change in https://github.com/opensearch-project/neural-search/pull/710 surfaced this problem. If every addition of a new parameter leads to a new method needing to be written, this makes changes very expensive.

Improved.

None.

No. This is an internal class.

No change.

No.

-

Create a POJO class to encapsulate the parameters, making it easier to extend functionality.

No.

fen-qin:
### Description
This is a code refactoring on encapsulation of MLCommonsClientAccessor required parameters into a POJO object.

- create a new request java object
- update MLCommonsClientAccessor interface with new request

### Related Issues
Resolves #790 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
@@ -27,6 +27,9 @@ import org.opensearch.ml.common.output.model.ModelTensor;
 import org.opensearch.ml.common.output.model.ModelTensorOutput;
 import org.opensearch.ml.common.output.model.ModelTensors;
 import org.opensearch.neuralsearch.processor.InferenceRequest;
+import org.opensearch.neuralsearch.processor.MapInferenceRequest;
+import org.opensearch.neuralsearch.processor.SimilarityInferenceRequest;
+import org.opensearch.neuralsearch.processor.TextInferenceRequest;
 import org.opensearch.neuralsearch.util.RetryUtil;
 
 import lombok.NonNull;
@@ -39,37 +42,37 @@ import lombok.extern.log4j.Log4j2;
 @RequiredArgsConstructor
 @Log4j2
 public class MLCommonsClientAccessor {
-    private static final List<String> TARGET_RESPONSE_FILTERS = List.of("sentence_embedding");
     private final MachineLearningNodeClient mlClient;
 
     /**
      * Wrapper around {@link #inferenceSentences} that expected a single input text and produces a single floating
      * point vector as a response.
      *
-     * @param inferenceRequest {@link InferenceRequest}
+     * @param modelId {@link String}
+     * @param inputText {@link String}
      * @param listener {@link ActionListener} which will be called when prediction is completed or errored out
      */
-    public void inferenceSentence(@NonNull final InferenceRequest inferenceRequest, @NonNull final ActionListener<List<Float>> listener) {
-        if (inferenceRequest.getInputTexts().size() != 1) {
-            listener.onFailure(
-                new IllegalArgumentException(
-                    "Unexpected number of input texts. Expected 1 input text, but got [" + inferenceRequest.getInputTexts().size() + "]"
-                )
-            );
-            return;
-        }
-        inferenceSentences(inferenceRequest, ActionListener.wrap(response -> {
-            if (response.size() != 1) {
-                listener.onFailure(
-                    new IllegalStateException(
-                        "Unexpected number of vectors produced. Expected 1 vector to be returned, but got [" + response.size() + "]"
-                    )
-                );
-                return;
-            }
+    public void inferenceSentence(
+        @NonNull final String modelId,
+        @NonNull final String inputText,
+        @NonNull final ActionListener<List<Float>> listener
+    ) {
 
-            listener.onResponse(response.getFirst());
-        }, listener::onFailure));
+        inferenceSentences(
+            TextInferenceRequest.builder().modelId(modelId).inputTexts(List.of(inputText)).build(),
+            ActionListener.wrap(response -> {
+                if (response.size() != 1) {
+                    listener.onFailure(
+                        new IllegalStateException(
+                            "Unexpected number of vectors produced. Expected 1 vector to be returned, but got [" + response.size() + "]"
+                        )
+                    );
+                    return;
+                }
+
+                listener.onResponse(response.getFirst());
+            }, listener::onFailure)
+        );
     }
 
     /**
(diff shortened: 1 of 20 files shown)


src/test/java/org/opensearch/neuralsearch/processor/InferenceProcessorTests.java
```diff
@@ -63,7 +62,7 @@ public void test_batchExecute_emptyInput() {
         ArgumentCaptor<List<IngestDocumentWrapper>> captor = ArgumentCaptor.forClass(List.class);
         verify(resultHandler).accept(captor.capture());
         assertTrue(captor.getValue().isEmpty());
-        verify(clientAccessor, never()).inferenceSentences(anyString(), anyList(), any());
+        verify(clientAccessor, never()).inferenceSentences(any(), any());
```
will-hwang, line 65: nit
`isA(InferenceRequest.class)` and `isA(ActionListener.class)` could be used for all places instead of` any()`

src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
```diff
@@ -373,12 +353,7 @@ public void testInferenceSimilarity_whenValidInput_thenSuccess() {
             return null;
         }).when(client).predict(Mockito.eq(TestCommonConstants.MODEL_ID), Mockito.isA(MLInput.class), Mockito.isA(ActionListener.class));
 
-        accessor.inferenceSimilarity(
-            TestCommonConstants.MODEL_ID,
-            "is it sunny",
-            List.of("it is sunny today", "roses are red"),
-            singleSentenceResultListener
-        );
+        accessor.inferenceSimilarity(TestCommonConstants.INFERENCE_REQUEST, singleSentenceResultListener);
```
will-hwang, line 356: seems like INFERENCE_REQUEST does not contain "it is sunny today", "roses are red" in its request? we should make the example be the same

fen-qin: thanks, will, would like to update the test constants.
thought the draft pull request won't be visible to others, let me take some time to fix the unit tests and raise another revision

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -45,16 +46,19 @@ public class MLCommonsClientAccessor {
...
      * @param listener {@link ActionListener} which will be called when prediction is completed or errored out
      */
-    public void inferenceSentence(
-        @NonNull final String modelId,
-        @NonNull final String inputText,
-        @NonNull final ActionListener<List<Float>> listener
-    ) {
-        inferenceSentences(TARGET_RESPONSE_FILTERS, modelId, List.of(inputText), ActionListener.wrap(response -> {
+    public void inferenceSentence(@NonNull final InferenceRequest inferenceRequest, @NonNull final ActionListener<List<Float>> listener) {
+        if (inferenceRequest.getInputTexts().size() != 1) {
```
heemin32, line 53: I believe we shouldn’t enforce a single input text at runtime. Instead, it should be clear from the method signature that only one input text is allowed. Considering that we are transitioning to the `InferenceRequest` POJO object, this method doesn’t seem relevant to me. I suggest we remove it entirely.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -64,92 +68,77 @@ public void inferenceSentence(
...
+            inferenceRequest.getInputTexts(),
+            0,
+            listener
+        );
     }
 
     public void inferenceSentencesWithMapResult(
-        @NonNull final String modelId,
-        @NonNull final List<String> inputText,
+        @NonNull final InferenceRequest inferenceRequest,
```
heemin32, line 101: I feel like we might want to have different inference request object for each case with shared parent class.
Sharing single object make it confusing which parameter we have to set for each type of call.

For example, `queryText` is used only for `inferenceSimilarity` but not for `inferenceSentence`.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
fen-qin: please correct me if my understanding is wrong:
- **_inferenceSentence_** should take single input text from method signature
- but **_inferenceSentence_** method is not relevant, and can be removed entirely

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -64,92 +68,77 @@ public void inferenceSentence(
...
      */
     public void inferenceSentences(
-        @NonNull final List<String> targetResponseFilters,
-        @NonNull final String modelId,
-        @NonNull final List<String> inputText,
+        @NonNull final InferenceRequest inferenceRequest,
         @NonNull final ActionListener<List<List<Float>>> listener
     ) {
-        retryableInferenceSentencesWithVectorResult(targetResponseFilters, modelId, inputText, 0, listener);
+        retryableInferenceSentencesWithVectorResult(
```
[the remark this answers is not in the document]
fen-qin: - `retryableInferenceSentencesWithVectorResult` is a private function, I put all public functions to accept inferenceRequest
- but consider code maintenance, yes, that makes sense. will make `retryableInferenceSentencesWithVectorResult` to accept inferenceRequest as parameter

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -64,92 +68,77 @@ public void inferenceSentence(
...
+            (inferenceRequest.getTargetResponseFilters() == null || inferenceRequest.getTargetResponseFilters().isEmpty())
```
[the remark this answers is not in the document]
fen-qin: yes. will set `TARGET_RESPONSE_FILTERS` as defaulted value. 
there is a use case that targetResponseFilters will be customized from Builder. but still can set the defaulted value when no customization is needed.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
fen-qin: yes. this is a good idea.

can you walk me through the various inference request ?

- parent
```
- InferenceRequest 
  - modelId
  - inputText
  - targetResponseFilters
```

children
```
- MapInferenceRequest
  - inputObject
 
- SimilarityInferenceRequest
  - queryText
```

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
heemin32: In previous method implementation, we allow caller to pass single `inputText` from method signature which enforce them to pass single text. Now, we allow customer to pass `inferenceRequest` which hold a list of `inputText` and we enforce them to pass single text during runtime. Then, I don't see the need of this changed version of method.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
heemin32: That sounds right. I don't know the all different inference type. You need to check the code and see what are there.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
fen-qin: Got it. agree on this point. we should allow caller to pass single inputText and a model id as what it is in previous method implementation.

heemin32 (approved):
LGTM. Thanks!

junqiu-lei (approved):
LGTM, thanks!

Issue #1182, opened 2025-02-11
Semantic Highlighter Project Tasks Tracker


junqiu-lei:
This issue is created to track tasks and share updates related to Semantic Highlighter project at open source community. Origin Github issue at: https://github.com/opensearch-project/neural-search/issues/145

- [x] https://github.com/opensearch-project/neural-search/issues/1175
- [x] https://github.com/opensearch-project/neural-search/pull/1183
- [x] https://github.com/opensearch-project/ml-commons/pull/3600
- [x] https://github.com/opensearch-project/neural-search/pull/1193
- [x] https://github.com/opensearch-project/ml-commons/pull/3699
- [x] https://github.com/opensearch-project/neural-search/pull/1254
- [x] https://github.com/opensearch-project/neural-search/pull/1259
- [x] Benchmark https://github.com/opensearch-project/neural-search/issues/1182#issuecomment-2826488685
- [x] Publish Huggingface source model https://huggingface.co/opensearch-project/opensearch-semantic-highlighter-v1
- [x] https://github.com/opensearch-project/opensearch-py-ml/pull/486
- [x] https://github.com/opensearch-project/documentation-website/issues/9583

q-andy:
@junqiu-lei I think you linked the wrong issue as the RFC, it should be #1175 not #145

junqiu-lei:
> [@junqiu-lei](https://github.com/junqiu-lei) I think you linked the wrong issue as the RFC, it should be [#1175](https://github.com/opensearch-project/neural-search/issues/1175) not [#145](https://github.com/opensearch-project/neural-search/issues/145)

Thanks @q-andy, updated.

junqiu-lei:
# Benchmark Overview

This report details the highlight latency performance benchmarks conducted using the **OpenSearch Benchmark** tool. The tests are based on modified versions of the standard workloads available in the `opensearch-benchmark-workloads` repository:

*   **treccovid_semantic_search**: [https://github.com/opensearch-project/opensearch-benchmark-workloads/tree/main/treccovid_semantic_search](https://github.com/opensearch-project/opensearch-benchmark-workloads/tree/main/treccovid_semantic_search) [^1]
*   **neural_search**: [https://github.com/opensearch-project/opensearch-benchmark-workloads/tree/main/neural_search](https://github.com/opensearch-project/opensearch-benchmark-workloads/tree/main/neural_search) [^2]

[...]

Issue #1175, opened 2025-02-07
[RFC] OpenSearch Semantic Sentence Highlighting


junqiu-lei:
# OpenSearch Semantic Sentence Highlighting
## Introduction

This document outlines the design and implementation of semantic highlighting feature in OpenSearch. Semantic highlighting aims to enhance search result highlighting by leveraging machine learning models to identify semantically relevant text fragments, going beyond traditional lexical matching approaches.

## Problem Statement

Traditional highlighting in OpenSearch relies on lexical matching, which has several limitations. The current system cannot effectively capture semantically relevant content when there are no exact keyword matches. It struggles with identifying multiple relevant spans across a document and lacks the ability to provide context-aware highlighting based on query intent.

The need for semantic highlighting stems from several key requirements in modern search systems. Users expect search results that understand semantic meaning, not just keyword matches. This requires highlighting that can identify contextually relevant passages even when exact terms don't match. Additionally, long documents often contain multiple relevant sections that need to be highlighted to provide comprehensive search results. The system should also be able to understand and highlight content based on the user's search intent rather than just matching words.

Github issue for this feature request: https://github.com/opensearch-project/neural-search/issues/145

### Easy User Integration

* Simple configuration through existing highlight API
* No additional setup beyond model deployment
* Backward compatible with existing highlight syntax

### Highlighting Quality and Performance

* Semantically relevant highlights even without exact keyword matches
* Context-aware highlighting that captures complete thoughts/sentences
* Response times reasonable along with the model inference time

### Semantic Highlighting Integration

* Implement new "neural" highlighter type
* Support configuration through search request options
* Enable model specification in highlight requests
* Support fields type
    * Neural Query
        * Most critical for semantic search
        * Direct alignment with semantic highlighting
    * Match Query
        * Most common text query type
        * Basic natural language support

### Models
* Provide pre-trained model option while feature release
* Support remote model integration and provide blueprint docs
* Provide default model option implementation (post p0)

### Out of Scope
* Model training or fine-tuning
* Image or multi-modal highlighting
* Automatic model selection or optimization
* Semantic highlights feature stats API (will look into it after neural-search plugin publish stats api framework)
* Support GPU enabled sentence QA model

## Current State

The current highlighting system in OpenSearch:

1. Supports multiple highlighter types (unified, plain, fvh), reference to [OpenSearch highlighters document](https://opensearch.org/docs/latest/search-plugins/searching-data/highlight/)
[...]

PR #1183, opened 2025-02-11
Encapsulate KNNQueryBuilder creation within NeuralKNNQueryBuilder

junqiu-lei:
### Description

Centralizes KNNQueryBuilder creation in NeuralKNNQueryBuilder to:
1. Put KNNQueryBuilder in a single location for easier handle breaking change from k-NN plugin.
2. Provide abstraction for future extensions like for neural highlighter feature need the original query text as input, which is currently lost in kNNQuery object. More context can be found at "Neural Query Builder Wrapper (based on Highlighter Framework Approach)" in https://github.com/opensearch-project/neural-search/issues/1175

### Related Issues

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Documentation
 ### Maintenance
 ### Refactoring
+- Encapsulate KNNQueryBuilder creation within NeuralKNNQueryBuilder ([#1183](https://github.com/opensearch-project/neural-search/pull/1183))
 
 ## [Unreleased 2.x](https://github.com/opensearch-project/neural-search/compare/2.19...2.x)
 ### Features
--- a/src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
@@ -1,0 +1,167 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.query;
+
+import lombok.Getter;
+import org.apache.lucene.search.Query;
+import org.opensearch.core.common.io.stream.StreamOutput;
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.index.query.AbstractQueryBuilder;
+import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.index.query.QueryRewriteContext;
+import org.opensearch.index.query.QueryShardContext;
+import org.opensearch.knn.index.query.KNNQueryBuilder;
+import org.opensearch.knn.index.query.rescore.RescoreContext;
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * NeuralKNNQueryBuilder wraps KNNQueryBuilder to:
+ * 1. Isolate KNN plugin API changes to a single location
+ * 2. Allow extension with neural-search-specific information (e.g., query text)
+ */
+
+@Getter
+public class NeuralKNNQueryBuilder extends AbstractQueryBuilder<NeuralKNNQueryBuilder> {
+    private final KNNQueryBuilder knnQueryBuilder;
+
+    /**
+     * Creates a new builder instance.
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public String fieldName() {
+        return knnQueryBuilder.fieldName();
+    }
+
+    public int k() {
+        return knnQueryBuilder.getK();
+    }
+
+    /**
+     * Builder for NeuralKNNQueryBuilder.
+     */
+    public static class Builder {
+        private String fieldName;
+        private float[] vector;
+        private Integer k;
+        private QueryBuilder filter;
+        private Float maxDistance;
+        private Float minScore;
+        private Boolean expandNested;
+        private Map<String, ?> methodParameters;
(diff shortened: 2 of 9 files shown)


src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
```diff
@@ -0,0 +1,158 @@
...
+ * 2. Allow extension with neural-search-specific information (e.g., query text)
+ */
+public class NeuralKNNQueryBuilder extends AbstractQueryBuilder<NeuralKNNQueryBuilder> {
+    private final KNNQueryBuilder knnQueryBuilder;
+
+    /**
+     * Wraps KNN Lucene query to support neural search extensions.
+     * Delegates core operations to the underlying KNN query.
+     */
+    public static class NeuralKNNQuery extends Query {
```
heemin32, line 38: Can we have this in a separate file? Is there any reason why we want to have it inside builder?

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
```diff
@@ -0,0 +1,158 @@
...
+        public int hashCode() {
+            return Objects.hash(knnQuery);
+        }
+    }
+
+    /**
+     * Factory method to create NeuralKNNQueryBuilder with given parameters.
+     * Centralizes KNNQueryBuilder creation to handle breaking changes.
+     */
+    public static NeuralKNNQueryBuilder fromKNNBuilder(
```
heemin32, line 89: I think builder pattern might be better than having a method with large list of parameters. Also, the method name `fromKNNBuilder` is exposing internal implementation. We can simply call it `createNeuralKNNQueryBuilder()` but I think having builder pattern would be better than having this method.

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
```diff
@@ -0,0 +1,158 @@
...
+            .minScore(minScore)
+            .expandNested(expandNested)
+            .methodParameters(methodParameters)
+            .rescoreContext(rescoreContext)
+            .build();
+        return new NeuralKNNQueryBuilder(knnBuilder);
+    }
+
+    private NeuralKNNQueryBuilder(KNNQueryBuilder knnQueryBuilder) {
+        super();
```
heemin32, line 115: You don't need to call `super()`. It will be called automatically.

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
junqiu-lei: It's already a separate new file in NeuralKNNQueryBuilder

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
junqiu-lei: ack

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
heemin32: I mean, moving this class to `NeuralKNNQuery.java`

junqiu-lei: I see, will update to separate NeuralKNNQuery out to make it cleaner

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
junqiu-lei: Updated to use builder pattern in NeuralKNNQueryBuilder.

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
junqiu-lei: Removed.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Documentation
 ### Maintenance
 ### Refactoring
+- Isolate KNNQueryBuilder creation in NeuralKNNQueryBuilder ([#1183](https://github.com/opensearch-project/neural-search/pull/1183))
```
martin-gaievski, line 15: suggest slight change in description: "Encapsulate KNNQueryBuilder creation within NeuralKNNQueryBuilder", if you're ok with it please also change the PR header

junqiu-lei: Updated in the changeLog and PR description.

martin-gaievski (approved):
looks good to me, thanks

heemin32 (dismissed):

martin-gaievski:
if this change is for 3.0 we need to look into BWC 3.0 -> 2.19, that's our current strategy, support rolling upgrades to latest minor of previous major, we probably should continue doing same after 3.x will be released.

junqiu-lei:
> if this change is for 3.0 we need to look into BWC 3.0 -> 2.19, that's our current strategy, support rolling upgrades to latest minor of previous major, we probably should continue doing same after 3.x will be released. Of course this makes more sense if errors in BWC are hard to fix, proper deep dive is required

Makes sense, looking into the bwc tests failure.

junqiu-lei:
The 2.19->3.0.0 bwc test failed due to known issue https://github.com/opensearch-project/neural-search/issues/1142

heemin32 (approved):

Issue #1095, opened 2025-01-14
[Code maintenance] Fix the warning messages in the build logs


vibrantvarun:
Clear the warnings from the build logs.

```
/home/runner/work/neural-search/neural-search/build/generated/sources/delombok/java/main/org/opensearch/neuralsearch/executors/HybridQueryExecutorCollectorManager.java:10: warning: no @param for <C>
public interface HybridQueryExecutorCollectorManager<C extends HybridQueryExecutorCollector> {
       ^
/home/runner/work/neural-search/neural-search/build/generated/sources/delombok/java/main/org/opensearch/neuralsearch/executors/HybridQueryExecutorCollector.java:14: warning: no @param for <I>
public final class HybridQueryExecutorCollector<I, R> {
             ^
/home/runner/work/neural-search/neural-search/build/generated/sources/delombok/java/main/org/opensearch/neuralsearch/executors/HybridQueryExecutorCollector.java:14: warning: no @param for <R>
public final class HybridQueryExecutorCollector<I, R> {
             ^
/home/runner/work/neural-search/neural-search/build/generated/sources/delombok/java/main/org/opensearch/neuralsearch/processor/AbstractScoreHybridizationProcessor.java:41: warning: no description for @param
     * @param <Result>
       ^
/home/runner/work/neural-search/neural-search/build/generated/sources/delombok/java/main/org/opensearch/neuralsearch/processor/AbstractScoreHybridizationProcessor.java:20: warning: use of default constructor, which does not provide a comment
public abstract class AbstractScoreHybridizationProcessor implements SearchPhaseResultsProcessor {
                ^
/home/runner/work/neural-search/neural-search/build/generated/sources/delombok/java/main/org/opensearch/neuralsearch/processor/explain/ExplainableTechnique.java:22: warning: no @return
    default String describe() {
                   ^
/home/runner/work/neural-search/neural-search/build/generated/sources/delombok/java/main/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java:18: warning: no comment
    public static final String TECHNIQUE_NAME = "arithmetic_mean";
                               ^
/home/runner/work/neural-search/neural-search/build/generated/sources/delombok/java/main/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java:24: warning: no comment
[...]

PR #1193, opened 2025-02-19
Support semantic sentence highlighter

junqiu-lei:
## Overview
This PR introduces a semantic sentence highlighter framework that enhances OpenSearch's highlighting capabilities by using ML models to identify and highlight semantically relevant text spans. The highlighter works with both traditional keyword queries, neural and hybrid queries, providing more contextually relevant highlights.

## Key Changes
**Core Framework Implementation**
   - Added `SemanticHighlighter` and `SemanticHighlighterEngine` classes
   - Integrated with ML Commons for neural-based highlighting
   - Implemented text processing and highlight position validation

**Query Support**
   - Implemented query text extraction for various query types
     - Match, Term, Boolean, Query String, Neural, Hybrid
   - Enhanced highlight result handling for different query scenarios
   - Highlighter Comparison for Plain vs. Semantic details [here](https://gist.github.com/junqiu-lei/c0b708c39d28b2887341dd118f608fdc). 

**Engine Features**
   - Text processing with position validation
   - Support for multiple highlight spans

**Integration**
   - Integrated with OpenSearch's highlighter framework
   - Added ML Commons client integration
     - Related PR https://github.com/opensearch-project/ml-commons/pull/3600

## Query Example Usage
```json
{
  "query": {
    "neural": {
      "field": "text_embedding",
      "query_text": "AI image generation and text creation",
      "model_id": "text-embedding-model",
      "k": 3
    }
  },
  "highlight": {
    "fields": {
      "text": {
        "type": "semantic"
      }
    },
    "options": {
      "model_id": "sentence-highlighting-model"
    }
  }
}
```

<details>
<summary>View Response</summary>

```json
{
  "took": 157289,
  "timed_out": false,
  "_shards": {
    "total": 1,
    "successful": 1,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 3,
      "relation": "eq"
    },
    "max_score": 0.027053982,
    "hits": [
      {
        "_index": "my-nlp-index",
        "_id": "14",
        "_score": 0.027053982,
        "_source": {
          "text": "Generative AI models create original content including images, music, and text. Diffusion models generate high-quality visual content by gradually transforming noise into coherent images. Large language models can write essays, code, and creative fiction that mimics human-created work.",
          "id": "doc14"
        },
        "highlight": {
          "text": [
            "<em>Generative AI models create original content including images, music, and text.</em> <em>Diffusion models generate high-quality visual content by gradually transforming noise into coherent images.</em> <em>Large language models can write essays, code, and creative fiction that mimics human-created work.</em>"
          ]
        }
      },
      {
        "_index": "my-nlp-index",
        "_id": "7",
        "_score": 0.019589059,
        "_source": {
          "text": "Artificial intelligence systems can now generate realistic images and text. These generative models have applications in art, design, and content creation. They raise important questions about creativity and authenticity.",
          "id": "doc7"
        },
        "highlight": {
          "text": [
            "<em>Artificial intelligence systems can now generate realistic images and text.</em> These generative models have applications in art, design, and content creation. They raise important questions about creativity and authenticity."
          ]
        }
      },
      {
        "_index": "my-nlp-index",
        "_id": "2",
        "_score": 0.01394515,
        "_source": {
          "text": "Vector databases store embeddings that represent the meaning of text. They enable efficient similarity searches based on semantic relationships. This technology powers modern search engines and recommendation systems.",
          "id": "doc2"
        },
        "highlight": {
          "text": [
            "<em>Vector databases store embeddings that represent the meaning of text.</em> <em>They enable efficient similarity searches based on semantic relationships.</em> This technology powers modern search engines and recommendation systems."
          ]
        }
      }
    ]
  }
}
```
</details>

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
+++ b/src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.highlight;
+
+import lombok.extern.log4j.Log4j2;
+import org.opensearch.index.mapper.MappedFieldType;
+import org.opensearch.search.fetch.subphase.highlight.FieldHighlightContext;
+import org.opensearch.search.fetch.subphase.highlight.HighlightField;
+import org.opensearch.search.fetch.subphase.highlight.Highlighter;
+import org.opensearch.core.common.text.Text;
+
+/**
+ * Semantic highlighter that uses ML models to identify relevant text spans for highlighting
+ */
+@Log4j2
+public class SemanticHighlighter implements Highlighter {
+    public static final String NAME = "semantic";
+
+    private SemanticHighlighterEngine semanticHighlighterEngine;
+
+    public void initialize(SemanticHighlighterEngine semanticHighlighterEngine) {
+        if (this.semanticHighlighterEngine != null) {
+            throw new IllegalStateException(
+                "SemanticHighlighterEngine has already been initialized. Multiple initializations are not permitted."
+            );
+        }
+        this.semanticHighlighterEngine = semanticHighlighterEngine;
+    }
+
+    @Override
+    public boolean canHighlight(MappedFieldType fieldType) {
+        return true;
+    }
+
+    /**
+     * Highlights a field using semantic highlighting
+     *
+     * @param fieldContext The field context containing the query and field information
+     * @return The highlighted field or null if highlighting is not possible
+     */
+    @Override
+    public HighlightField highlight(FieldHighlightContext fieldContext) {
+        if (semanticHighlighterEngine == null) {
+            throw new IllegalStateException("SemanticHighlighter has not been initialized");
+        }
+
+        // Extract field text
+        String fieldText = semanticHighlighterEngine.getFieldText(fieldContext);
+
+        // Get model ID
+        String modelId = semanticHighlighterEngine.getModelId(fieldContext.field.fieldOptions().options());
+
+        // Try to extract query text
+        String originalQueryText = semanticHighlighterEngine.extractOriginalQuery(fieldContext.query, fieldContext.fieldName);
+
+        if (originalQueryText == null || originalQueryText.isEmpty()) {
+            log.warn("No query text found for field {}", fieldContext.fieldName);
+            return null;
+        }
+
+        // Get highlighted text - allow any exceptions from this call to propagate
+        String highlightedResponse = semanticHighlighterEngine.getHighlightedSentences(modelId, originalQueryText, fieldText);
+
+        if (highlightedResponse == null || highlightedResponse.isEmpty()) {
+            log.warn("No highlighted text found for field {}", fieldContext.fieldName);
+            return null;
+        }
+
+        // Create highlight field
+        Text[] fragments = new Text[] { new Text(highlightedResponse) };
+        return new HighlightField(fieldContext.fieldName, fragments);
+    }
+}
--- a/src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
+++ b/src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
@@ -1,0 +1,297 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.highlight;
+
+import lombok.extern.log4j.Log4j2;
+import org.apache.lucene.search.Query;
+import org.opensearch.OpenSearchException;
+import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
+import org.opensearch.neuralsearch.processor.highlight.SentenceHighlightingRequest;
+import org.opensearch.search.fetch.subphase.highlight.FieldHighlightContext;
+import org.opensearch.neuralsearch.highlight.extractor.QueryTextExtractorRegistry;
+import org.opensearch.action.support.PlainActionFuture;
+import lombok.NonNull;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Engine class for semantic highlighting operations
+ */
+@Log4j2
+public class SemanticHighlighterEngine {
+    private static final String MODEL_ID_FIELD = "model_id";
+    private static final String DEFAULT_PRE_TAG = "<em>";
+    private static final String DEFAULT_POST_TAG = "</em>";
+    private static final String MODEL_INFERENCE_RESULT_KEY = "highlights";
+    private static final String MODEL_INFERENCE_RESULT_START_KEY = "start";
+    private static final String MODEL_INFERENCE_RESULT_END_KEY = "end";
+
+    private final MLCommonsClientAccessor mlCommonsClient;
+    private final QueryTextExtractorRegistry queryTextExtractorRegistry;
+
+    public SemanticHighlighterEngine(
+        @NonNull MLCommonsClientAccessor mlCommonsClient,
+        @NonNull QueryTextExtractorRegistry queryTextExtractorRegistry
+    ) {
+        this.mlCommonsClient = mlCommonsClient;
+        this.queryTextExtractorRegistry = queryTextExtractorRegistry;
+    }
+
+    /**
+     * Gets the field text from the document
+     *
+     * @param fieldContext The field highlight context
+     * @return The field text
+     */
+    public String getFieldText(FieldHighlightContext fieldContext) {
+        if (fieldContext.hitContext == null || fieldContext.hitContext.sourceLookup() == null) {
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "Field %s is not found in the hit", fieldContext.fieldName));
+        }
+        Object fieldTextObject = fieldContext.hitContext.sourceLookup().extractValue(fieldContext.fieldName, null);
+        if (fieldTextObject == null) {
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "Field %s is not found in the document", fieldContext.fieldName));
+        }
+        if (fieldTextObject instanceof String == false) {
+            throw new IllegalArgumentException(
+                String.format(
+                    Locale.ROOT,
+                    "Field %s must be a string for highlighting, but was %s",
+                    fieldContext.fieldName,
+                    fieldTextObject.getClass().getSimpleName()
+                )
+            );
+        }
+        String fieldTextString = (String) fieldTextObject;
+        if (fieldTextString.isEmpty()) {
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "Field %s is empty", fieldContext.fieldName));
+        }
+        return fieldTextString;
+    }
+
+    /**
+     * Extracts the original query text from the search query object.
+     *
+     * @param query The query object from which to extract the original text
+     * @param fieldName The name of the field being highlighted
+     * @return The extracted original query text for highlighting
+     * @throws IllegalArgumentException if the extracted query text is empty
+     */
+    public String extractOriginalQuery(Query query, String fieldName) {
+        if (fieldName == null) {
+            log.warn("Field name is null, extraction may be less accurate");
+        }
+        return queryTextExtractorRegistry.extractQueryText(query, fieldName);
+    }
+
+    /**
+     * Gets the model ID from the options
+     *
+     * @param options The options map
+     * @return The model ID
+     */
+    public String getModelId(Map<String, Object> options) {
+        Object modelId = options.get(MODEL_ID_FIELD);
+        if (Objects.isNull(modelId) || (modelId instanceof String) == false) {
+            throw new IllegalArgumentException(
+                String.format(
+                    Locale.ROOT,
+                    "%s must be a non-null string, but was %s",
+                    MODEL_ID_FIELD,
+                    modelId == null ? "null" : modelId.getClass().getSimpleName()
+                )
+            );
+        }
+        return (String) modelId;
+    }
+
+    /**
+     * Gets highlighted text from the ML model
+     *
+     * @param modelId The ID of the model to use
+     * @param question The search query
+     * @param context The document text
+     * @return Formatted text with highlighting
+     */
+    public String getHighlightedSentences(String modelId, String question, String context) {
+        List<Map<String, Object>> results = fetchModelResults(modelId, question, context);
+        if (results == null || results.isEmpty()) {
+            return null;
+        }
+
+        return applyHighlighting(context, results.getFirst());
+    }
+
+    /**
+     * Fetches highlighting results from the ML model
+     *
+     * @param modelId The ID of the model to use
+     * @param question The search query
+     * @param context The document text
+     * @return The highlighting results
+     */
+    public List<Map<String, Object>> fetchModelResults(String modelId, String question, String context) {
+        PlainActionFuture<List<Map<String, Object>>> future = PlainActionFuture.newFuture();
+
+        SentenceHighlightingRequest request = SentenceHighlightingRequest.builder()
+            .modelId(modelId)
+            .question(question)
+            .context(context)
+            .build();
+
+        mlCommonsClient.inferenceSentenceHighlighting(request, future);
+
+        try {
+            return future.actionGet();
+        } catch (Exception e) {
+            log.error(
+                "Error during sentence highlighting inference - modelId: [{}], question: [{}], context: [{}]",
+                modelId,
+                question,
+                context,
+                e
+            );
+            throw new OpenSearchException(
+                String.format(Locale.ROOT, "Error during sentence highlighting inference from model [%s]", modelId),
+                e
+            );
+        }
+    }
+
+    /**
+     * Applies highlighting to the original context based on the ML model response
+     *
+     * @param context The original document text
+     * @param highlightResult The highlighting result from the ML model
+     * @return Formatted text with highlighting
+     * @throws IllegalArgumentException if highlight positions are invalid
+     */
+    public String applyHighlighting(String context, Map<String, Object> highlightResult) {
+        // Get the "highlights" list from the result
+        Object highlightsObj = highlightResult.get(MODEL_INFERENCE_RESULT_KEY);
+
+        if (!(highlightsObj instanceof List<?> highlightsList)) {
+            log.error(String.format(Locale.ROOT, "No valid highlights found in model inference result, highlightsObj: %s", highlightsObj));
+            return null;
+        }
+
+        if (highlightsList.isEmpty()) {
+            // No highlights found, return context as is
+            return context;
+        }
+
+        // Pre-allocate size * 2 since we store start and end positions as consecutive pairs
+        // Format: [start1, end1, start2, end2, start3, end3, ...]
+        ArrayList<Integer> validHighlights = new ArrayList<>(highlightsList.size() * 2);
+
+        for (Object item : highlightsList) {
+            Map<String, Number> map = getHighlightsPositionMap(item);
+
+            Number start = map.get(MODEL_INFERENCE_RESULT_START_KEY);
+            Number end = map.get(MODEL_INFERENCE_RESULT_END_KEY);
+
+            if (start == null || end == null) {
+                throw new OpenSearchException("Missing start or end position in highlight data");
+            }
+
+            // Validate positions and add them as a pair to maintain the start-end relationship
+            validateHighlightPositions(start.intValue(), end.intValue(), context.length());
+            validHighlights.add(start.intValue());  // Even indices (0,2,4,...) store start positions
+            validHighlights.add(end.intValue());    // Odd indices (1,3,5,...) store end positions
+        }
+
+        // Verify highlights are sorted by start position (ascending)
+        // We start from i=2 (second start position) and compare with previous start position (i-2)
+        // Using i+=2 to skip end positions and only compare start positions with each other
+        for (int i = 2; i < validHighlights.size(); i += 2) {
+            // Compare current start position with previous start position
+            if (validHighlights.get(i) < validHighlights.get(i - 2)) {
+                log.error(String.format(Locale.ROOT, "Highlights are not sorted: %s", validHighlights));
+                throw new OpenSearchException("Internal error while applying semantic highlight: received unsorted highlights from model");
+            }
+        }
+
+        return constructHighlightedText(context, validHighlights);
+    }
+
+    /**
+     * Validates highlight position values
+     *
+     * @param start The start position
+     * @param end The end position
+     * @param textLength The length of the text being highlighted
+     * @throws OpenSearchException if positions are invalid
+     */
+    private void validateHighlightPositions(int start, int end, int textLength) {
+        if (start < 0 || end > textLength || start >= end) {
+            throw new OpenSearchException(
+                String.format(
+                    Locale.ROOT,
+                    "Invalid highlight positions: start=%d, end=%d, textLength=%d. Positions must satisfy: 0 <= start < end <= textLength",
+                    start,
+                    end,
+                    textLength
+                )
+            );
+        }
+    }
+
+    /**
+     * Constructs highlighted text by iterating through the text once
+     *
+     * @param text The original text
+     * @param highlights The list of valid highlight positions in pairs [start1, end1, start2, end2, ...]
+     * @return The highlighted text
+     */
+    private String constructHighlightedText(String text, List<Integer> highlights) {
+        StringBuilder result = new StringBuilder();
+        int currentPos = 0;
+
+        // Iterate through highlight positions in pairs (start, end)
+        // i increments by 2 to move from one pair to the next
+        for (int i = 0; i < highlights.size(); i += 2) {
+            int start = highlights.get(i);     // Get start position from even index
+            int end = highlights.get(i + 1);   // Get end position from odd index
+
+            // Add text before the highlight if there is any
+            if (start > currentPos) {
+                result.append(text, currentPos, start);
+            }
+
+            // Add the highlighted text with highlight tags
+            result.append(DEFAULT_PRE_TAG);
+            result.append(text, start, end);
+            result.append(DEFAULT_POST_TAG);
+
+            // Update current position to end of this highlight
+            currentPos = end;
+        }
+
+        // Add any remaining text after the last highlight
+        if (currentPos < text.length()) {
+            result.append(text, currentPos, text.length());
+        }
+
+        return result.toString();
+    }
+
+    /**
+     * Extracts the highlight position map from a highlight item
+     *
+     * @param item The highlight item
+     * @return The highlight position map
+     * @throws OpenSearchException if the item cannot be cast to Map<String, Number>
+     */
+    private static Map<String, Number> getHighlightsPositionMap(Object item) {
+        try {
+            return (Map<String, Number>) item;
+        } catch (ClassCastException e) {
+            throw new OpenSearchException(String.format(Locale.ROOT, "Expect item to be map of string to number, but was: %s", item));
+        }
+    }
+}
--- a/src/main/java/org/opensearch/neuralsearch/highlight/extractor/BooleanQueryTextExtractor.java
+++ b/src/main/java/org/opensearch/neuralsearch/highlight/extractor/BooleanQueryTextExtractor.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.highlight.extractor;
+
+import org.apache.lucene.search.BooleanClause;
+import org.apache.lucene.search.BooleanQuery;
+import org.apache.lucene.search.Query;
+
+import lombok.extern.log4j.Log4j2;
+
+/**
+ * Extractor for boolean queries
+ */
+@Log4j2
+public class BooleanQueryTextExtractor implements QueryTextExtractor {
+
+    private final QueryTextExtractorRegistry registry;
+
+    public BooleanQueryTextExtractor(QueryTextExtractorRegistry registry) {
+        this.registry = registry;
+    }
+
+    @Override
+    public String extractQueryText(Query query, String fieldName) {
+        BooleanQuery booleanQuery = toQueryType(query, BooleanQuery.class);
+
+        StringBuilder sb = new StringBuilder();
+
+        for (BooleanClause clause : booleanQuery.clauses()) {
+            // Skip MUST_NOT clauses as they represent negative terms
+            if (clause.isProhibited()) {
+                continue;
+            }
+
+            try {
+                String clauseText = registry.extractQueryText(clause.query(), fieldName);
+                if (clauseText.isEmpty() == false) {
+                    if (sb.isEmpty() == false) {
+                        sb.append(" ");
+                    }
+                    sb.append(clauseText);
+                }
+            } catch (IllegalArgumentException e) {
+                log.warn("Failed to extract text from clause {}: {}", clause, e.getMessage(), e);
+            }
+        }
+
+        return sb.toString();
+    }
+}
--- a/src/main/java/org/opensearch/neuralsearch/highlight/extractor/QueryTextExtractor.java
+++ b/src/main/java/org/opensearch/neuralsearch/highlight/extractor/QueryTextExtractor.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.highlight.extractor;
+
+import org.apache.lucene.search.Query;
+
+import java.util.Locale;
+
+/**
+ * Interface for extracting query text from different query types
+ */
+public interface QueryTextExtractor {
+    /**
+     * Converts a query to the expected type, throwing an exception if the type doesn't match
+     *
+     * @param query The query to convert
+     * @param expectedType The expected query type
+     * @return The query cast to the expected type
+     * @throws IllegalArgumentException if the query is not of the expected type
+     */
+    default <T extends Query> T toQueryType(Query query, Class<T> expectedType) {
+        if (!expectedType.isInstance(query)) {
+            throw new IllegalArgumentException(
+                String.format(Locale.ROOT, "Expected %s but got %s", expectedType.getSimpleName(), query.getClass().getSimpleName())
+            );
+        }
+        return expectedType.cast(query);
+    }
+
+    /**
+     * Extracts text from a query for highlighting
+     *
+     * @param query The query to extract text from
+     * @param fieldName The name of the field being highlighted
+     * @return The extracted query text
+     */
+    String extractQueryText(Query query, String fieldName);
+}
--- a/src/main/java/org/opensearch/neuralsearch/highlight/extractor/QueryTextExtractorRegistry.java
+++ b/src/main/java/org/opensearch/neuralsearch/highlight/extractor/QueryTextExtractorRegistry.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.highlight.extractor;
+
+import org.apache.lucene.search.BooleanQuery;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.TermQuery;
+import org.opensearch.neuralsearch.query.NeuralKNNQuery;
+import org.opensearch.neuralsearch.query.HybridQuery;
+
+import lombok.extern.log4j.Log4j2;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Registry for query text extractors that manages the extraction process
+ */
+@Log4j2
+public class QueryTextExtractorRegistry {
+
+    private final Map<Class<? extends Query>, QueryTextExtractor> extractors = new HashMap<>();
+
+    /**
+     * Creates a new registry with default extractors
+     */
+    public QueryTextExtractorRegistry() {
+        initialize();
+    }
+
+    /**
+     * Initializes the registry with default extractors
+     */
+    private void initialize() {
+        register(NeuralKNNQuery.class, new NeuralQueryTextExtractor());
+        register(TermQuery.class, new TermQueryTextExtractor());
+        register(HybridQuery.class, new HybridQueryTextExtractor(this));
+
+        // BooleanQueryTextExtractor needs a reference to this registry
+        // so we need to register it after creating the registry instance
+        register(BooleanQuery.class, new BooleanQueryTextExtractor(this));
+    }
+
+    /**
+     * Registers an extractor for a specific query type
+     *
+     * @param queryClass The query class to register for
+     * @param extractor The extractor to use for this query type
+     */
+    public <T extends Query> void register(Class<T> queryClass, QueryTextExtractor extractor) {
+        extractors.put(queryClass, extractor);
+    }
+
+    /**
+     * Extracts text from a query using the appropriate extractor
+     *
+     * @param query The query to extract text from
+     * @param fieldName The name of the field being highlighted
+     * @return The extracted query text
+     */
+    public String extractQueryText(Query query, String fieldName) {
+        if (query == null) {
+            log.warn("Cannot extract text from null query");
+            return null;
+        }
+        Class<?> queryClass = query.getClass();
+        QueryTextExtractor extractor;
+
+        extractor = extractors.get(queryClass);
+
+        if (extractor == null) {
+            log.warn("No extractor found for query type: {}", queryClass.getName());
+            return null;
+        }
+
+        return extractor.extractQueryText(query, fieldName);
+    }
+}
--- a/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
@@ -9,6 +9,7 @@ import static org.opensearch.neuralsearch.processor.TextImageEmbeddingProcessor.
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -31,6 +32,8 @@ import org.opensearch.neuralsearch.processor.MapInferenceRequest;
 import org.opensearch.neuralsearch.processor.SimilarityInferenceRequest;
 import org.opensearch.neuralsearch.processor.TextInferenceRequest;
 import org.opensearch.neuralsearch.util.RetryUtil;
+import org.opensearch.ml.common.dataset.QuestionAnsweringInputDataSet;
+import org.opensearch.neuralsearch.processor.highlight.SentenceHighlightingRequest;
 
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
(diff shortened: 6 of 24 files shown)


q-andy (approved):

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,100 @@
...
+ * Neural highlighter that uses ML models to identify relevant text spans for highlighting
+ */
+@Log4j2
+public class NeuralHighlighter implements Highlighter {
+    public static final String NAME = "neural";
+    private static final String MODEL_ID_FIELD = "model_id";
+
+    private static MLCommonsClientAccessor mlCommonsClient;
+
+    public static void initialize(MLCommonsClientAccessor mlClient) {
```
heemin32, line 32: This pattern is not good for unit test. Can we receive mlClient through constructor?

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Neural highlighter that uses ML models to identify relevant text spans for highlighting
+ */
+@Log4j2
+public class NeuralHighlighter implements Highlighter {
+    public static final String NAME = "neural";
+    private static final String MODEL_ID_FIELD = "model_id";
+
+    private static MLCommonsClientAccessor mlCommonsClient;
```
heemin32, line 30: This static pattern is not ideal for unit test. Can we receive accessor thought constructor?

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public static void initialize(MLCommonsClientAccessor mlClient) {
+        NeuralHighlighter.mlCommonsClient = mlClient;
+    }
+
+    @Override
+    public boolean canHighlight(MappedFieldType fieldType) {
+        // TODO: Implement actual condition check in subsequent PR
```
heemin32, line 38: Could be better to restrict it to TEXT field type and expand it later in subsequent PR.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,100 @@
...
+            }
+
+            String searchQuery = extractOriginalQuery(fieldContext.query);
+            if (searchQuery.isEmpty()) {
+                return null;
+            }
+
+            Map<String, Object> options = fieldContext.field.fieldOptions().options();
+            String modelId = getModelId(options);
+            log.info("Using model ID: {}", modelId); // Will be replaced with actual model loading logic
```
heemin32, line 57: This log will be flooded and obscure other important logs. Let't move it to debug level.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,100 @@
...
+            log.info("Using ML client: {}", mlCommonsClient); // Will be replaced with actual model loading logic
+
+            // TODO: Implement actual highlighting logic in subsequent PR
```
heemin32, line 60: With all these TODO, it seems like this PR should not be merged to main now.
Let me stop here and review it when the code is fully completed.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    private String getFieldText(FieldHighlightContext fieldContext) {
+        Object value = fieldContext.hitContext.sourceLookup().extractValue(fieldContext.fieldName, null);
+        return value != null ? value.toString() : "";
+    }
+
+    private String formatHighlight(String text) {
+        // TODO: Implement user provided format options in subsequent PR
+        return "<em>" + text + "</em>";
```
martin-gaievski, line 87: use String format for safer concatenation

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,100 @@
...
+
+    private String extractOriginalQuery(Query query) {
+        if (query instanceof NeuralKNNQuery neuralQuery) {
+            String originalText = neuralQuery.getOriginalQueryText();
+            if (originalText != null) {
+                return originalText;
+            }
+        }
+
+        return query.toString().replaceAll("\\w+:", "").replaceAll("\\s+", " ").trim();
```
martin-gaievski, line 98: please add comment describing what you are trying to extract from string here

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,100 @@
...
+            Map<String, Object> options = fieldContext.field.fieldOptions().options();
+            String modelId = getModelId(options);
+            log.info("Using model ID: {}", modelId); // Will be replaced with actual model loading logic
+            log.info("Using ML client: {}", mlCommonsClient); // Will be replaced with actual model loading logic
+
+            // TODO: Implement actual highlighting logic in subsequent PR
+            // For now, return a basic highlight of the field text
+            Text[] fragments = new Text[] { new Text(formatHighlight(fieldText)) };
+            return new HighlightField(fieldContext.fieldName, fragments);
+        } catch (Exception e) {
```
martin-gaievski, line 64: is it possible to narrow cases when we can have exceptions? Having global level try/catch is not that great for performance on search path

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,100 @@
...
+    private String getModelId(Map<String, Object> options) {
+        Object modelId = options.get(MODEL_ID_FIELD);
+        if (Objects.isNull(modelId)) {
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "Missing required option: %s", MODEL_ID_FIELD));
+        }
+        return modelId.toString();
+    }
+
+    private String getFieldText(FieldHighlightContext fieldContext) {
+        Object value = fieldContext.hitContext.sourceLookup().extractValue(fieldContext.fieldName, null);
```
martin-gaievski, line 81: isn't this second argument a fallback in case requested key is missing? If yes can't we return empty string here and avoid the null check on a line below?

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
junqiu-lei: ack

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
junqiu-lei: sure

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
junqiu-lei: I was trying to add a temp placeholder here, since client api function was planed in another PR.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
junqiu-lei: Got it. My original purpose was trying to make the PR more clear by different component(highlighter framework, ml client, text processing logic...). I could also combine all code in one big PR.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
junqiu-lei: I see, let me try to adjust to narrow exception catch cases.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
vibrantvarun: Please add javadoc and a comment of all the important steps this method is trying to do.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,100 @@
...
+
+            String searchQuery = extractOriginalQuery(fieldContext.query);
+            if (searchQuery.isEmpty()) {
+                return null;
+            }
+
+            Map<String, Object> options = fieldContext.field.fieldOptions().options();
+            String modelId = getModelId(options);
+            log.info("Using model ID: {}", modelId); // Will be replaced with actual model loading logic
+            log.info("Using ML client: {}", mlCommonsClient); // Will be replaced with actual model loading logic
```
vibrantvarun, line 58:
```suggestion
            log.debug("Using ML client: {}", mlCommonsClient); // Will be replaced with actual model loading logic
```

vibrantvarun: Curious why do we need this log this?

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    private String formatHighlight(String text) {
+        // TODO: Implement user provided format options in subsequent PR
+        return "<em>" + text + "</em>";
+    }
+
+    private String extractOriginalQuery(Query query) {
+        if (query instanceof NeuralKNNQuery neuralQuery) {
+            String originalText = neuralQuery.getOriginalQueryText();
```
vibrantvarun, line 92:
```suggestion
            String originalQuery = neuralQuery.getOriginalQueryText();
```

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQuery.java
```diff
@@ -21,9 +21,11 @@
 @Getter
 public class NeuralKNNQuery extends Query {
     private final Query knnQuery;
+    private final String originalQueryText;
 
-    public NeuralKNNQuery(Query knnQuery) {
+    public NeuralKNNQuery(Query knnQuery, String originalQueryText) {
         this.knnQuery = knnQuery;
+        this.originalQueryText = originalQueryText;
```
vibrantvarun, line 28: Shall we add BWC test to validate that it does not break anything in bwc environment?

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
```diff
@@ -107,6 +109,11 @@ public Builder rescoreContext(RescoreContext rescoreContext) {
             return this;
         }
 
+        public Builder originalQueryText(String originalQueryText) {
```
vibrantvarun, line 112: Please add javadoc. We already have an issue for reducing the warning logs https://github.com/opensearch-project/neural-search/issues/1095

vibrantvarun: If possible please add the javadoc on all public methods of this class.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
junqiu-lei: @vibrantvarun sure

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
junqiu-lei: See the comment along with this line code

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQuery.java
junqiu-lei: Sure, will have a check on bwc side

junqiu-lei:
For reference in this PR, the PR for sentence highlighting model support in ml-commons:
https://github.com/opensearch-project/ml-commons/pull/3600

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
junqiu-lei: The code was outdated. I changed to use queryText with:
```
        String queryText = (query instanceof NeuralKNNQuery neuralQuery)
            ? neuralQuery.getOriginalQueryText()
            : query.toString().replaceAll("\\w+:", "").replaceAll("\\s+", " ").trim();
```

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,224 @@
...
+
+    private static MLCommonsClientAccessor mlCommonsClient;
+
+    public static void initialize(MLCommonsClientAccessor mlClient) {
+        NeuralHighlighter.mlCommonsClient = mlClient;
+    }
+
+    @Override
+    public boolean canHighlight(MappedFieldType fieldType) {
+        return SUPPORTED_FIELD_TYPE.equals(fieldType.typeName());
```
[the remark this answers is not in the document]
junqiu-lei: I believe the NeuralHighlighter should remain restricted to "text" fields only. Neural highlighting is designed for semantic understanding of text content and wouldn't be meaningful for non-text fields like numbers or dates. And the sentence highlighting ML models are trained on text data and expect text input.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,224 @@
...
+    private static final String SUPPORTED_FIELD_TYPE = "text";
+    private static final String MODEL_INFERENCE_RESULT_KEY = "highlights";
+    private static final String MODEL_INFERENCE_RESULT_START_KEY = "start";
+    private static final String MODEL_INFERENCE_RESULT_END_KEY = "end";
+    private static final int MODEL_INFERENCE_DEFAULT_TIMEOUT = 30;
+
+    private static MLCommonsClientAccessor mlCommonsClient;
+
+    public static void initialize(MLCommonsClientAccessor mlClient) {
+        NeuralHighlighter.mlCommonsClient = mlClient;
```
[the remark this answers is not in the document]
junqiu-lei: Updated

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,224 @@
...
+
+    @Override
+    public boolean canHighlight(MappedFieldType fieldType) {
+        return SUPPORTED_FIELD_TYPE.equals(fieldType.typeName());
+    }
+
+    @Override
+    public HighlightField highlight(FieldHighlightContext fieldContext) {
+        MappedFieldType fieldType = fieldContext.fieldType;
+        if (canHighlight(fieldType) == false) {
```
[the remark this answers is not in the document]
junqiu-lei: I removed this check inside NeuralHighlighter since it will be checked in core.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,224 @@
...
+                Map<String, Object> highlight = highlights.get(i);
+                if (highlight == null) {
+                    continue;
+                }
+
+                // Extract start and end positions
+                Object startObj = highlight.get(MODEL_INFERENCE_RESULT_START_KEY);
+                Object endObj = highlight.get(MODEL_INFERENCE_RESULT_END_KEY);
+
+                if (!(startObj instanceof Number) || !(endObj instanceof Number)) {
```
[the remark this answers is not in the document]
junqiu-lei: Added check for highlights and startObj and endObj ideally should be number. Updated.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,224 @@
...
+                    continue;
+                }
+
+                int start = ((Number) startObj).intValue();
+                int end = ((Number) endObj).intValue();
+
+                // Validate positions
+                if (start < 0 || end > highlightedText.length() || start >= end) {
+                    continue;
```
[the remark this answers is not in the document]
junqiu-lei: Updated to have a validation method for position

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,224 @@
...
+                }
+
+                // Insert highlighting tags
+                highlightedText.insert(end, DEFAULT_POST_TAG);
```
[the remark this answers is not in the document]
junqiu-lei: Updated to use O(n) way to construct highlighted text.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,224 @@
...
+        return hitValue;
+    }
+
+    private String extractOriginalQuery(Query query) {
+        String queryText = (query instanceof NeuralKNNQuery neuralQuery)
+            ? neuralQuery.getOriginalQueryText()
+            : query.toString().replaceAll("\\w+:", "").replaceAll("\\s+", " ").trim();
+
+        if (queryText.isEmpty()) {
+            throw new IllegalArgumentException("Original neural query text is empty");
```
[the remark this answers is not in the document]
junqiu-lei: updated.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,224 @@
...
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "Field %s is not found in the document", fieldContext.fieldName));
+        }
+        if (hitValue.isEmpty()) {
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "Field %s is empty", fieldContext.fieldName));
+        }
+        return hitValue;
+    }
+
+    private String extractOriginalQuery(Query query) {
+        String queryText = (query instanceof NeuralKNNQuery neuralQuery)
```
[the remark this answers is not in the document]
junqiu-lei: In core the existing highlighters use Lucene's query visitor pattern to extract terms from the query object, let me have a check to enhance the extraction method.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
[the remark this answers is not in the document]
heemin32: How about array, object, keyword, and semantic field(in the future)? https://opensearch.org/docs/latest/field-types/

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlightProcessor.java
```diff
@@ -0,0 +1,278 @@
...
+        mlCommonsClient.inferenceSentenceHighlighting(request, ActionListener.wrap(result -> {
+            resultRef.set(result);
+            latch.countDown();
+        }, exception -> {
+            exceptionRef.set(exception);
+            latch.countDown();
+        }));
+
+        try {
+            if (!latch.await(MODEL_INFERENCE_DEFAULT_TIMEOUT, TimeUnit.SECONDS)) {
```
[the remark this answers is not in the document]
junqiu-lei: There are remote model timeout set in ml-commons, [ref](https://github.com/search?q=repo:opensearch-project/ml-commons%20remote%20model%20timeout&type=code).

In this case, we can make the method wait indefinitely until the ML model returns a result or an error.
```
try {
    latch.await(); // Wait indefinitely
} catch (InterruptedException e) {
    Thread.currentThread().interrupt();
    throw new OpenSearchException("Interrupted while waiting for sentence highlighting inference from model", e);
}
```

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlightProcessor.java
```diff
@@ -0,0 +1,278 @@
...
+            }
+
+            // Process each item in the list, checking if it's a Map
+            List<Map<String, Object>> safeHighlights = new java.util.ArrayList<>();
+            for (Object item : highlightsList) {
+                if (item instanceof Map<?, ?> map) {
+                    // Create a type-safe map
+                    Map<String, Object> safeMap = new java.util.HashMap<>();
+                    for (Map.Entry<?, ?> entry : map.entrySet()) {
+                        if (entry.getKey() instanceof String key) {
```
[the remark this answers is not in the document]
junqiu-lei: It should be String type from ml-common model output anyway, updated.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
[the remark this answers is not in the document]
junqiu-lei: I introduced queryExtractor in https://github.com/opensearch-project/neural-search/pull/1193/commits/813226fdcb2b31ff357980502b4e166e2fc5f0e8#diff-376285b7459fe11bc1a187f7833d22d6f9e2ebaeadabec39f0c270fdd31c7795 which can handle different query types.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighter.java
```diff
@@ -0,0 +1,69 @@
...
+    public void initialize(MLCommonsClientAccessor mlClient) {
+        if (neuralHighlighterManager != null) {
+            throw new IllegalStateException("NeuralHighlighter has already been initialized. Multiple initializations are not permitted.");
+        }
+        this.neuralHighlighterManager = new NeuralHighlighterManager(mlClient);
+    }
+
+    @Override
+    public boolean canHighlight(MappedFieldType fieldType) {
+        return true;
```
junqiu-lei, line 33: Make it return to true, which is consistent to other core highlighters. If non-text field query passed in, the highlighter will return null, as a result no highlight part in query response, which is same behavior as other core highlighters.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlightProcessor.java
```diff
@@ -0,0 +1,278 @@
...
+    /**
+     * Extracts the original query text
+     *
+     * @param query The query object
+     * @return The original query text
+     */
+    public String extractOriginalQuery(Query query) {
+        String queryText = (query instanceof NeuralKNNQuery neuralQuery)
+            ? neuralQuery.getOriginalQueryText()
+            : query.toString().replaceAll("\\w+:", "").replaceAll("\\s+", " ").trim();
```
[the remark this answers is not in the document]
junqiu-lei: Introduced query text extractor to get the original query text.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighterManager.java
```diff
@@ -0,0 +1,323 @@
...
+            .context(context)
+            .build();
+
+        mlCommonsClient.inferenceSentenceHighlighting(request, ActionListener.wrap(result -> {
+            resultRef.set(result);
+            latch.countDown();
+        }, exception -> {
+            exceptionRef.set(exception);
+            latch.countDown();
+        }));
```
q-andy, line 141: I don't see this latch pattern anywhere else in OpenSearch. Could you use [PlainActionFuture](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/PlainActionFuture.java) to wrap the action listener with a future to make it synchronous instead of managing atomic references with callbacks?

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighterManager.java
```diff
@@ -0,0 +1,323 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Manager class for neural highlighting operations that handles the core highlighting logic
+ */
+@Log4j2
+public class NeuralHighlighterManager {
```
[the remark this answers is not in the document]
junqiu-lei: Updated to use `NeuralHighlighterEngine`.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighterManager.java
```diff
@@ -0,0 +1,323 @@
...
+    private static final String DEFAULT_POST_TAG = "</em>";
+    private static final String MODEL_INFERENCE_RESULT_KEY = "highlights";
+    private static final String MODEL_INFERENCE_RESULT_START_KEY = "start";
+    private static final String MODEL_INFERENCE_RESULT_END_KEY = "end";
+
+    private final MLCommonsClientAccessor mlCommonsClient;
+    private final QueryTextExtractorRegistry queryTextExtractorRegistry;
+
+    public NeuralHighlighterManager(MLCommonsClientAccessor mlCommonsClient) {
+        this.mlCommonsClient = Objects.requireNonNull(mlCommonsClient, "ML Commons client cannot be null");
```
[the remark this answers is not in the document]
junqiu-lei: Updated

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighterManager.java
junqiu-lei: Thanks @q-andy for suggestion, updated to use [PlainActionFuture](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/PlainActionFuture.java) here.

src/main/java/org/opensearch/neuralsearch/highlight/extractor/QueryTextExtractorRegistry.java
```diff
@@ -0,0 +1,88 @@
...
+     * Creates a new registry with default extractors
+     */
+    public QueryTextExtractorRegistry() {
+        initialize();
+    }
+
+    /**
+     * Initializes the registry with default extractors
+     */
+    public void initialize() {
```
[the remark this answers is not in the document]
junqiu-lei: Updated.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlightProcessor.java
```diff
@@ -0,0 +1,278 @@
...
+import java.util.Objects;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Processor class for neural highlighting operations that handles the core highlighting logic
+ */
+@Log4j2
+public class NeuralHighlightProcessor {
```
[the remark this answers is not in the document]
heemin32: NeuralHighlighter?

src/main/java/org/opensearch/neuralsearch/highlight/extractor/QueryTextExtractor.java
```diff
@@ -10,6 +10,21 @@
...
  */
 public interface QueryTextExtractor {
+    /**
+     * Validates if the query is of the expected type
+     *
+     * @param query The query to validate
+     * @param expectedType The expected query type
+     * @throws IllegalArgumentException if the query is not of the expected type
+     */
+    default void validateQueryType(Query query, Class<? extends Query> expectedType) {
```
heemin32, line 20: Better to return the type converted variable.
```
default <T extends Query> T toQueryType(Query query, Class<T> expectedType) {
    if (!expectedType.isInstance(query)) {
            String.format("Expected %s but got %s", expectedType.getSimpleName(), query.getClass().getSimpleName())
        );
    }
    return expectedType.cast(query);
}
```

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -262,4 +314,52 @@ private MLInput createMLMultimodalInput(final List<String> targetResponseFilters
...
+            },
+                e -> RetryUtil.handleRetryOrFailure(
+                    e,
+                    retryTime,
+                    () -> retryableInferenceSentenceHighlighting(inferenceRequest, retryTime + 1, listener),
+                    listener
+                )
+            ));
+        } catch (Exception e) {
+            log.error("Error preparing sentence highlighting inference", e);
```
heemin32, line 347: We might not need to write a log here because the exception will be handled by `listener.onFailure(e)` anyway.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighterEngine.java
```diff
@@ -0,0 +1,306 @@
...
+     * Applies highlighting to the original context based on the ML model response
+     *
+     * @param context The original document text
+     * @param highlightResults The highlighting results from the ML model
+     * @return Formatted text with highlighting
+     */
+    public String applyHighlighting(String context, List<Map<String, Object>> highlightResults) {
+        List<int[]> validHighlights = new ArrayList<>();
+
+        if (highlightResults != null && !highlightResults.isEmpty()) {
```
heemin32, line 153: Better to return early to reduce the level of depth.

if (highlightResults == null || highlightResults.isEmpty()) {
  return context;
}

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighterEngine.java
```diff
@@ -0,0 +1,306 @@
...
+            Map<String, Object> result = highlightResults.getFirst();
```
heemin32, line 154: If we are only handling the very first item from the list, the method should receive single item instead of list.
```
public String applyHighlighting(String context, Map<String, Object> highlightResults) {
```

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighterEngine.java
```diff
@@ -0,0 +1,306 @@
...
+            if (result != null) {
+                // Get the "highlights" list from the result
+                Object highlightsObj = result.get(MODEL_INFERENCE_RESULT_KEY);
+
+                if (highlightsObj instanceof List<?> highlightsList) {
```
heemin32, line 159: Shouldn't we throw an exception when it is not List type?

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighterEngine.java
```diff
@@ -0,0 +1,306 @@
...
+                    for (Object item : highlightsList) {
+                        if (item instanceof Map<?, ?> map) {
```
heemin32, line 161: Can't we just cast it to `Map<String, Number>` and throw an exception otherwise?

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighterEngine.java
```diff
@@ -0,0 +1,306 @@
...
+            }
+        }
+
+        // If no valid highlights, return the original text
+        if (validHighlights.isEmpty()) {
+            return context;
+        }
+
+        // Sort highlights by start position (ascending)
+        validHighlights.sort(Comparator.comparingInt(pos -> pos[0]));
```
heemin32, line 187: model does not guarantee to return sorted value? Shouldn't we enforce it just by throwing an error when value is not sorted while adding tags?

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighterEngine.java
```diff
@@ -0,0 +1,306 @@
...
+        return result.toString();
+    }
+
+    /**
+     * Merges overlapping highlights to ensure proper nesting
+     *
+     * @param highlights The list of highlights sorted by start position
+     * @return A list of merged highlights with no overlaps
+     */
+    private List<int[]> mergeOverlappingHighlights(List<int[]> highlights) {
```
heemin32, line 239: Shouldn't model generate non overlapping highlight?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -262,4 +314,52 @@ private MLInput createMLMultimodalInput(final List<String> targetResponseFilters
...
+        final SentenceHighlightingRequest inferenceRequest,
+        final int retryTime,
+        final ActionListener<List<Map<String, Object>>> listener
+    ) {
+        try {
+            MLInputDataset inputDataset = new QuestionAnsweringInputDataSet(inferenceRequest.getQuestion(), inferenceRequest.getContext());
+            MLInput mlInput = new MLInput(FunctionName.QUESTION_ANSWERING, null, inputDataset);
+
+            mlClient.predict(inferenceRequest.getModelId(), mlInput, ActionListener.wrap(mlOutput -> {
+                List<Map<String, Object>> result = processHighlightingOutput(mlOutput);
```
heemin32, line 336: You need to catch exception.
try {
  List<Map<String, Object>> result = processHighlightingOutput(mlOutput);
  listener.onRespone(result);
} catch (Exception e) {
  listener.onFailure(e);

}

src/main/java/org/opensearch/neuralsearch/highlight/extractor/QueryTextExtractor.java
```diff
@@ -0,0 +1,36 @@
...
+        if (expectedType.isInstance(query) == false) {
+            throw new IllegalArgumentException(
+                String.format("Expected %s but got %s", expectedType.getSimpleName(), query.getClass().getSimpleName())
```
heemin32, line 23: Missing `Locale.ROOT`

src/main/java/org/opensearch/neuralsearch/highlight/extractor/BooleanQueryTextExtractor.java
```diff
@@ -0,0 +1,53 @@
...
+            try {
+                String clauseText = registry.extractQueryText(clause.query(), fieldName);
+                if (clauseText.isEmpty() == false) {
+                    if (sb.isEmpty() == false) {
+                        sb.append(" ");
+                    }
+                    sb.append(clauseText);
+                }
+            } catch (IllegalArgumentException e) {
+                log.warn(String.format("Failed to extract text from clause %s: %s", clause, e.getMessage()), e);
```
heemin32, line 47:
```suggestion
                log.warn("Failed to extract text from clause {}: {}", clause, e.getMessage(), e);
```

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighterEngine.java
junqiu-lei: We'll have sorted response from our pre-trained model. But we cannot make sure customized model have sorted response. But I still agree the solution to make it simpler handle in client side(neural-search plugin) with exception while it has unsorted response.

heemin32: Yes. The main concern is a latency due to this sorting.

src/main/java/org/opensearch/neuralsearch/highlight/NeuralHighlighterEngine.java
junqiu-lei: Right, both our provided model and other models should handle this logic on server side, removed this part for latency conern.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
junqiu-lei: Ack and updated.

junqiu-lei:
The Build CI test failure is related to https://github.com/opensearch-project/k-NN/pull/2603.

```
NeuralSearchTests > testValidateKNNDependency FAILED
    java.lang.NoClassDefFoundError: org/opensearch/remoteindexbuild/model/RemoteIndexParameters
        at __randomizedtesting.SeedInfo.seed([C2644A9F68EECE85:3B0119099E47BDDD]:0)
        at org.opensearch.knn.index.engine.faiss.Faiss.<clinit>(Faiss.java:62)
        at org.opensearch.knn.index.engine.KNNEngine.<clinit>(KNNEngine.java:31)
        at org.opensearch.neuralsearch.NeuralSearchTests.testValidateKNNDependency(NeuralSearchTests.java:14)

        Caused by:
        java.lang.ClassNotFoundException: org.opensearch.remoteindexbuild.model.RemoteIndexParameters
            at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
            at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
            at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:528)
            ... 3 more
```

src/main/java/org/opensearch/neuralsearch/highlight/extractor/QueryTextExtractorRegistry.java
```diff
@@ -0,0 +1,90 @@
...
+     */
+    public String extractQueryText(Query query, String fieldName) {
+        // Find the most specific extractor for this query type
+        Class<?> queryClass = query.getClass();
+        QueryTextExtractor extractor;
+
+        // Look for an exact match first
+        extractor = extractors.get(queryClass);
+
+        // If no exact match, look for parent classes
```
heemin32, line 72: Wondering if there is such case where one query extend from another?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -252,6 +255,52 @@ private void retryableInferenceSentencesWithSingleVectorResult(
         ));
     }
 
+    /**
+     * Process the highlighting output from ML model response.
+     * Converts the model output into a list of maps containing highlighting information.
+     */
+    private List<Map<String, Object>> processHighlightingOutput(MLOutput mlOutput) {
```
heemin32, line 262: Better to accept only `ModelTensorOutput` if that is what we want it to be.

```suggestion
    private List<Map<String, Object>> processHighlightingOutput(ModelTensorOutput mlOutput) {
```

src/main/java/org/opensearch/neuralsearch/highlight/extractor/QueryTextExtractorRegistry.java
junqiu-lei: We can remove this part since we covered common query cases, and if there is a case of null extractor found, then no highlight section will be returned, which is same behavior like existing highlighter.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
junqiu-lei: ack

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
```diff
@@ -0,0 +1,278 @@
...
+
+    /**
+     * Extracts the highlight position map from a highlight item
+     *
+     * @param item The highlight item
+     * @return The highlight position map
+     * @throws IllegalArgumentException if the item is not a valid highlight position map
+     */
+    private static Map<String, Number> getHighlightsPositionMap(Object item) {
+        if (!(item instanceof Map<?, ?> rawMap)) {
```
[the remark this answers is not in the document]
junqiu-lei: Updated

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Semantic highlighter that uses ML models to identify relevant text spans for highlighting
+ */
+@Log4j2
+public class SemanticHighlighter implements Highlighter {
+    public static final String NAME = "semantic";
+
+    private SemanticHighlighterEngine semanticHighlighterEngine;
+
+    public void initialize(MLCommonsClientAccessor mlClient) {
```
heemin32, line 25: Better to receive `SemanticHighlighterEngine` directly for easier unit test.
```suggestion
    public void initialize(SemanticHighlighterEngine semanticHighlighterEngine) {
```

junqiu-lei: Updated.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
```diff
@@ -0,0 +1,76 @@
...
+        if (semanticHighlighterEngine != null) {
+            throw new IllegalStateException(
+                "SemanticHighlighterEngine has already been initialized. Multiple initializations are not permitted."
+            );
+        }
+        this.semanticHighlighterEngine = new SemanticHighlighterEngine(mlClient);
```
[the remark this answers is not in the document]
junqiu-lei: Updated.

src/test/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterIT.java
```diff
@@ -0,0 +1,377 @@
...
+            10,
+            TEST_TEXT_FIELD_NAME,
+            sentenceHighlightingModelId
+        );
+
+        verifyHighlightResults(searchResponse, "artificial intelligence");
+    }
+
+    /**
+     * Tests semantic highlighting with query string:
```
martin-gaievski, line 228: curious why you picked this query type, is this something users are used/planning to use often with the highlighter?

src/test/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterIT.java
```diff
@@ -0,0 +1,377 @@
...
+import org.opensearch.neuralsearch.BaseNeuralSearchIT;
+import org.opensearch.neuralsearch.query.NeuralQueryBuilder;
+import org.opensearch.index.query.QueryStringQueryBuilder;
+import org.opensearch.index.query.TermQueryBuilder;
+import org.opensearch.search.builder.SearchSourceBuilder;
+
+import com.google.common.primitives.Floats;
+import lombok.SneakyThrows;
+
+public class SemanticHighlighterIT extends BaseNeuralSearchIT {
```
martin-gaievski, line 28: is it possible to combine some of the tests into a single method to minimize number of time model deployments? You're testing happy path for different query types, and each separate method required the same model to be deployed and undeployed. So essentially you can use same model id and test multiple queries with that model. Ofter when integ test are failing in CI during this model deployment step.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
```diff
@@ -0,0 +1,288 @@
...
+
+    /**
+     * Gets the model ID from the options
+     *
+     * @param options The options map
+     * @return The model ID
+     */
+    public String getModelId(Map<String, Object> options) {
+        Object modelId = options.get(MODEL_ID_FIELD);
+        if (Objects.isNull(modelId)) {
```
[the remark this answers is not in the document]
junqiu-lei: Updated to have both null and string instance type check, added tests as well.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
```diff
@@ -0,0 +1,288 @@
...
+            .question(question)
+            .context(context)
+            .build();
+
+        mlCommonsClient.inferenceSentenceHighlighting(request, future);
+
+        try {
+            return future.actionGet();
+        } catch (Exception e) {
+            throw new OpenSearchException(
```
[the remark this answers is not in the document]
junqiu-lei: Added.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -252,6 +255,51 @@ private void retryableInferenceSentencesWithSingleVectorResult(
...
+                        @SuppressWarnings("unchecked")
+                        Map<String, Object> typedDataMap = (Map<String, Object>) dataMap;
+                        results.add(typedDataMap);
+                    }
+                }
+            }
+
+            // If no results were found, add an empty map to maintain consistent response format
+            if (results.isEmpty()) {
+                results.add(new HashMap<>());
```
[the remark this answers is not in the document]
junqiu-lei: updated

src/test/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterIT.java
junqiu-lei: No specific reason for this type, wanted to cover some query cases in the test.

src/test/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterIT.java
junqiu-lei: Makes sense, updated IT to use shared resources.

martin-gaievski: another point is concurrent execution of tests by test runner in CI, we had problems in the past with global  model_id that is shared between test methods

junqiu-lei: The cleanup() will can be called in SemanticHighlighterIT after tests as well.

> another point is concurrent execution of tests by test runner in CI, we had problems in the past with global model_id that is shared between test methods

I don't see current CI failed due to the concurrent execution.

junqiu-lei: Offline talked with @martin-gaievski, going to merge different query type test cases into one.

src/test/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterIT.java
```diff
@@ -0,0 +1,341 @@
...
+    private String sentenceHighlightingModelId;
+
+    @Before
+    public void setUp() throws Exception {
+        super.setUp();
+        updateClusterSettings();
+
+        // Set up models once for all tests
+        textEmbeddingModelId = prepareModel();
+        sentenceHighlightingModelId = prepareSentenceHighlightingModel();
```
martin-gaievski, line 49: these two variable can be inside the test method, better do it now to avoid problems later if someone need to add more tests and they can just reuse class level variables with model ids.

martin-gaievski (approved):
Overall looks good to me, thank you @junqiu-lei. Please address my last comment about moving model id variables inside the test method.

Issue #1189, opened 2025-02-18
[RFC] Lower bound for min-max normalization technique in Hybrid query


martin-gaievski:
### Introduction

This document describes details of design for Explainability in Hybrid Query. This feature has been requested through GitHub issues https://github.com/opensearch-project/neural-search/issues/150 and https://github.com/opensearch-project/neural-search/issues/299.

### Overview

Hybrid search combines multiple query types, like keyword and neural search, to improve search relevance. In version 2.11 team released hybrid query which is [part of the neural-search plugin](https://opensearch.org/docs/latest/search-plugins/hybrid-search/). Main responsibility of the hybrid query is to return combined scores of multiple queries. In common scenario those queries represent different search types, like lexical and  semantic based.

Hybrid query uses multiple techniques for preparing final list of matched document, main two types are score based normalization and rank base combination. For score base normalization, the most effective technique is [min-max normalization](https://en.wikipedia.org/wiki/Feature_scaling). In scope of this proposal we want to improve search relevance of min-max normalization by allowing setting of lower bound.

### Problem Statement

Min-max normalization technique is based on the usage of maximum and minimum scores from all matched documents with the following formula. 

```normalizedScore = (score - minScore) / (maxScore - minScore);```

In context of OpenSearch, finding the minimum score is based on assumption that may be not the most effective one. While handling search request, system retrieves limited amount of matching documents from each shard, this limit is defined by the query parameter size. Minimum score will be identified as minimum from all score from all collected documents. In case overall number of matching documents is much higher then number of retrieved documents, then the delta between real and retrieved minimum scores can be significant. This will negatively influence the final normalized score.

Following graphs illustrate described scenario, in shard 1 retrieved min score is 4.0, while actual lower bound is 0.0. Similarly for shard 2 retrieved and lower bound scores are 2.0 and 1.0. 

![Image](https://github.com/user-attachments/assets/9260dde4-0329-415b-ae94-6350cb00bb27)

#### Functional Requirements

We want to introduce a lower limit that reflects the actual minimum score for matching document.

Functional Requirements

* User defines the lower bound score for each sub-query. It should be possible to have mixed configuration, where some sub-queries use the lower bound and some still use today approach.
* Lower bound score applies to existing min-max normalization technique in hybrid query/normalization processor. For other techniques we should block this feature.
* All changes in interface should be minimal, backward compatible and aligned with existing conventions. New change should not replace the default behavior.
[...]

Issue #150, opened 2023-04-06
[FEATURE] Support for setting a default min/max score for the upcoming Normalization and Score Combination feature


SeyedAlirezaFatemi:
### Is your feature request related to a problem?
Related to [RFC](https://github.com/opensearch-project/neural-search/issues/126). The current problem with the RFC is that when we are combining scores from different queries (e.g. BM25 and kNN), we need the min and max score of each query part. However, when using approximate kNN, we cannot accurately calculate the min score unless we do an exact kNN search on the index which is not feasible. This leads to inconsistent score normalization, particularly when using pagination.

### What solution would you like?
As discussed in detail in the [RFC](https://github.com/opensearch-project/neural-search/issues/126), one solution is to rely on the statistics we get from the documents we see during the current query. However, in specific scenarios where the min score can be known, we can do better. For example, when using BM25 or Cosine similarity in kNN, the user can optionally define the min score in the query to be 0 and -1, respectively.

By allowing the user to optionally define a min/max score in the query for normalization, we can ensure consistent score normalization across different queries for specific scenarios, particularly when using pagination. This would improve the accuracy and reliability of the search results for users.

Here is an example where we have the issue of pagination inconsistency when we use the general solution:
Let's assume we have a query that consists of a text match query and a kNN query and we use this formula for score normalization:
`x_normalized = (x – min) / (max – min)`
and we set the page size to 10. Assume the top 10 kNN scores are between 1 and 0.9 and then the scores for the rest of the documents fall to 0. This changes the scores after normalization drastically if we go to the next page and we might get pagination inconsistency and get missing/double results.

navneet1v:
@SeyedAlirezaFatemi Thanks for creating the issue.

navneet1v:
@SeyedAlirezaFatemi thanks for providing this info. I will look into this. We are still in the development phase of the original scope.

heemin32:
@SeyedAlirezaFatemi, is the inconsistent pagination result the main reason for supporting this? Even with the customer-provided min/max score, the inconsistency in pagination will still occur. There's an ongoing project aimed at improving pagination consistency for hybrid search. It would be great if you could take a look at #933 and share your thoughts on whether this feature would still provide value.

martin-gaievski:
[...]

PR #1195, opened 2025-02-24
[Feature branch] Lower bounds for min-max normalization in hybrid query

martin-gaievski:
### Description

I'm adding `lower_bounds` parameter for min-max normalization technique of hybrid query.
User will be able to create search pipeline with `lower_bounds` using following request:

```
{
  "description": "Normalization processor for hybrid search",
  "phase_results_processors": [
    {
      "normalization-processor": {
        "normalization": {
          "technique": "min_max",
          "parameters": {
            "lower_bounds": [
              { 
                "mode": "apply",
                "min_score": 0.1
              }, 
                "mode": "clip",
                "min_score": 0.1
              }, 
                "mode": "ignore"
              }
            ]
          }
        },
        "combination": {
          "technique": "arithmetic_mean"
        }
      }
    }
  ]
}
```

I'm planning to merge this into feature branch in main repo until app sec team give a signoff. 

### Related Issues
https://github.com/opensearch-project/neural-search/issues/150
Implemented design described in RFC https://github.com/opensearch-project/neural-search/issues/1189

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
@@ -170,10 +170,10 @@ public class CompoundTopDocs {
             if (thisTopDoc == null) {
                 continue;
             }
-            if (!Objects.equals(thisTopDoc.totalHits, thatTopDoc.totalHits)) {
+            if (Objects.equals(thisTopDoc.totalHits, thatTopDoc.totalHits) == false) {
                 return false;
             }
-            if (!compareScoreDocs(thisTopDoc.scoreDocs, thatTopDoc.scoreDocs)) {
+            if (compareScoreDocs(thisTopDoc.scoreDocs, thatTopDoc.scoreDocs) == false) {
                 return false;
             }
         }
@@ -202,7 +202,7 @@ public class CompoundTopDocs {
             }
             if (firstDoc instanceof FieldDoc firstFieldDoc) {
                 FieldDoc secondFieldDoc = (FieldDoc) secondDoc;
-                if (!Arrays.equals(firstFieldDoc.fields, secondFieldDoc.fields)) {
+                if (Arrays.equals(firstFieldDoc.fields, secondFieldDoc.fields) == false) {
                     return false;
                 }
             }
--- a/src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
@@ -13,6 +13,7 @@ import static org.opensearch.neuralsearch.search.util.HybridSearchResultFormatUt
 import static org.opensearch.neuralsearch.search.util.HybridSearchResultFormatUtil.isHybridQueryStartStopElement;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 
@@ -149,5 +150,79 @@ public class CompoundTopDocs {
         }
         FieldDoc fieldDoc = (FieldDoc) scoreDoc;
         return new FieldDoc(fieldDoc.doc, fieldDoc.score, fieldDoc.fields, fieldDoc.shardIndex);
+    }
+
+    @Override
+    public boolean equals(Object other) {
+        if (this == other) return true;
+        if (other == null || getClass() != other.getClass()) return false;
+        CompoundTopDocs that = (CompoundTopDocs) other;
+
+        if (this.topDocs.size() != that.topDocs.size()) {
+            return false;
+        }
+        for (int i = 0; i < topDocs.size(); i++) {
+            TopDocs thisTopDoc = this.topDocs.get(i);
+            TopDocs thatTopDoc = that.topDocs.get(i);
+            if ((thisTopDoc == null) != (thatTopDoc == null)) {
+                return false;
+            }
+            if (thisTopDoc == null) {
+                continue;
+            }
+            if (!Objects.equals(thisTopDoc.totalHits, thatTopDoc.totalHits)) {
+                return false;
+            }
+            if (!compareScoreDocs(thisTopDoc.scoreDocs, thatTopDoc.scoreDocs)) {
+                return false;
+            }
+        }
+        return Objects.equals(totalHits, that.totalHits) && Objects.equals(searchShard, that.searchShard);
+    }
+
+    private boolean compareScoreDocs(ScoreDoc[] first, ScoreDoc[] second) {
+        if (first.length != second.length) {
+            return false;
+        }
+
+        for (int i = 0; i < first.length; i++) {
+            ScoreDoc firstDoc = first[i];
+            ScoreDoc secondDoc = second[i];
+            if ((firstDoc == null) != (secondDoc == null)) {
+                return false;
+            }
+            if (firstDoc == null) {
+                continue;
+            }
+            if (firstDoc.doc != secondDoc.doc || Float.compare(firstDoc.score, secondDoc.score) != 0) {
+                return false;
+            }
+            if (firstDoc instanceof FieldDoc != secondDoc instanceof FieldDoc) {
+                return false;
+            }
+            if (firstDoc instanceof FieldDoc firstFieldDoc) {
+                FieldDoc secondFieldDoc = (FieldDoc) secondDoc;
+                if (!Arrays.equals(firstFieldDoc.fields, secondFieldDoc.fields)) {
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = Objects.hash(totalHits, searchShard);
+        for (TopDocs topDoc : topDocs) {
+            result = 31 * result + topDoc.totalHits.hashCode();
+            for (ScoreDoc scoreDoc : topDoc.scoreDocs) {
+                result = 31 * result + Float.floatToIntBits(scoreDoc.score);
+                result = 31 * result + scoreDoc.doc;
+                if (scoreDoc instanceof FieldDoc fieldDoc && fieldDoc.fields != null) {
+                    result = 31 * result + Arrays.deepHashCode(fieldDoc.fields);
+                }
+            }
+        }
+        return result;
     }
 }
--- a/src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
@@ -10,6 +10,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TopDocs;
@@ -31,6 +32,14 @@ public class L2ScoreNormalizationTechnique implements ScoreNormalizationTechniqu
     @ToString.Include
     public static final String TECHNIQUE_NAME = "l2";
     private static final float MIN_SCORE = 0.0f;
+
+    public L2ScoreNormalizationTechnique() {
+        this(Map.of(), new ScoreNormalizationUtil());
+    }
+
+    public L2ScoreNormalizationTechnique(final Map<String, Object> params, final ScoreNormalizationUtil scoreNormalizationUtil) {
+        scoreNormalizationUtil.validateParameters(params, Set.of(), Map.of());
+    }
 
     /**
      * L2 normalization method.
--- a/src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
@@ -246,11 +246,10 @@ public class MinMaxScoreNormalizationTechnique implements ScoreNormalizationTech
             return Optional.empty();
         }
 
-        Object lowerBoundsObj = params.get(PARAM_NAME_LOWER_BOUNDS);
-        if (lowerBoundsObj instanceof List<?> == false) {
-            throw new IllegalArgumentException("lower_bounds must be a List");
-        }
-        List<?> lowerBoundsParams = (List<?>) lowerBoundsObj;
+        List<?> lowerBoundsParams = Optional.ofNullable(params.get(PARAM_NAME_LOWER_BOUNDS))
+            .filter(List.class::isInstance)
+            .map(List.class::cast)
+            .orElseThrow(() -> new IllegalArgumentException("lower_bounds must be a List"));
         // number of lower bounds must match the number of sub-queries in a hybrid query
         if (lowerBoundsParams.size() > MAX_NUMBER_OF_SUB_QUERIES) {
             throw new IllegalArgumentException(
@@ -324,7 +323,7 @@ public class MinMaxScoreNormalizationTechnique implements ScoreNormalizationTech
      * Result class to hold lower bound for each sub query
      */
     @Getter
-    public static class LowerBound {
+    static class LowerBound {
         static final float MIN_LOWER_BOUND_SCORE = -10_000f;
         static final float MAX_LOWER_BOUND_SCORE = 10_000f;
         static final float DEFAULT_LOWER_BOUND_SCORE = 0.0f;
--- a/src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
@@ -4,7 +4,6 @@
  */
 package org.opensearch.neuralsearch.processor.normalization;
 
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
@@ -236,18 +235,22 @@ public class MinMaxScoreNormalizationTechnique implements ScoreNormalizationTech
         return lowerBound.getMode().normalize(score, minScore, maxScore, lowerBound.getMinScore());
     }
 
+    /**
+     * Get lower bounds from input parameters
+     * @param params user provided input parameters for this technique
+     * @return optional list of lower bounds. Can be empty in case lower bounds are not provided
+     */
     private Optional<List<Pair<LowerBound.Mode, Float>>> getLowerBounds(final Map<String, Object> params) {
+        // validate that the input parameters are in correct format
         if (Objects.isNull(params) || !params.containsKey(PARAM_NAME_LOWER_BOUNDS)) {
             return Optional.empty();
         }
-
-        List<Pair<LowerBound.Mode, Float>> lowerBounds = new ArrayList<>();
 
         Object lowerBoundsObj = params.get(PARAM_NAME_LOWER_BOUNDS);
         if (!(lowerBoundsObj instanceof List<?> lowerBoundsParams)) {
             throw new IllegalArgumentException("lower_bounds must be a List");
         }
-
+        // number of lower bounds must match the number of sub-queries in a hybrid query
         if (lowerBoundsParams.size() > MAX_NUMBER_OF_SUB_QUERIES) {
             throw new IllegalArgumentException(
                 String.format(
@@ -258,42 +261,52 @@ public class MinMaxScoreNormalizationTechnique implements ScoreNormalizationTech
                 )
             );
         }
-
-        for (Object boundObj : lowerBoundsParams) {
-            if (!(boundObj instanceof Map)) {
-                throw new IllegalArgumentException("each lower bound must be a map");
-            }
-
-            @SuppressWarnings("unchecked")
-            Map<String, Object> lowerBound = (Map<String, Object>) boundObj;
-
-            try {
-                LowerBound.Mode mode = LowerBound.Mode.fromString(
-                    Objects.isNull(lowerBound.get(PARAM_NAME_LOWER_BOUND_MODE))
-                        ? ""
-                        : lowerBound.get(PARAM_NAME_LOWER_BOUND_MODE).toString()
-                );
-                float minScore;
-                if (Objects.isNull(lowerBound.get(PARAM_NAME_LOWER_BOUND_MIN_SCORE))) {
-                    minScore = LowerBound.DEFAULT_LOWER_BOUND_SCORE;
-                } else {
-                    minScore = Float.parseFloat(String.valueOf(lowerBound.get(PARAM_NAME_LOWER_BOUND_MIN_SCORE)));
-                }
-
-                Validate.isTrue(
-                    minScore >= LowerBound.MIN_LOWER_BOUND_SCORE && minScore <= LowerBound.MAX_LOWER_BOUND_SCORE,
-                    "min_score must be a valid finite number between %f and %f",
-                    LowerBound.MIN_LOWER_BOUND_SCORE,
-                    LowerBound.MAX_LOWER_BOUND_SCORE
-                );
-
-                lowerBounds.add(ImmutablePair.of(mode, minScore));
-            } catch (NumberFormatException e) {
-                throw new IllegalArgumentException("invalid format for min_score: must be a valid float value", e);
-            }
-        }
+        // parse each lower bound item and put all items in a list
+        List<Pair<LowerBound.Mode, Float>> lowerBounds = lowerBoundsParams.stream().map(this::parseLowerBound).collect(Collectors.toList());
 
         return Optional.of(lowerBounds);
+    }
+
+    @SuppressWarnings("unchecked")
+    /**
+     * Parse each lower bound item and return a pair of mode and min score
+     * @param boundObj lower bound item provided by the client
+     * @return a single pair of mode and min score
+     */
+    private Pair<LowerBound.Mode, Float> parseLowerBound(Object boundObj) {
+        if (!(boundObj instanceof Map)) {
+            throw new IllegalArgumentException("each lower bound must be a map");
+        }
+
+        Map<String, Object> lowerBound = (Map<String, Object>) boundObj;
+
+        String lowerBoundModeValue = Objects.toString(lowerBound.get(PARAM_NAME_LOWER_BOUND_MODE), "");
+        LowerBound.Mode mode = LowerBound.Mode.fromString(lowerBoundModeValue);
+        float minScore = extractAndValidateMinScore(lowerBound);
+
+        return ImmutablePair.of(mode, minScore);
+    }
+
+    private float extractAndValidateMinScore(Map<String, Object> lowerBound) {
+        Object minScoreObj = lowerBound.get(PARAM_NAME_LOWER_BOUND_MIN_SCORE);
+        if (minScoreObj == null) {
+            return LowerBound.DEFAULT_LOWER_BOUND_SCORE;
+        }
+        try {
+            float minScore = LowerBound.DEFAULT_LOWER_BOUND_SCORE;
+            if (Objects.nonNull(lowerBound.get(PARAM_NAME_LOWER_BOUND_MIN_SCORE))) {
+                minScore = Float.parseFloat(String.valueOf(lowerBound.get(PARAM_NAME_LOWER_BOUND_MIN_SCORE)));
+            }
+            Validate.isTrue(
+                minScore >= LowerBound.MIN_LOWER_BOUND_SCORE && minScore <= LowerBound.MAX_LOWER_BOUND_SCORE,
+                "min_score must be a valid finite number between %f and %f",
+                LowerBound.MIN_LOWER_BOUND_SCORE,
+                LowerBound.MAX_LOWER_BOUND_SCORE
+            );
+            return minScore;
+        } catch (NumberFormatException e) {
+            throw new IllegalArgumentException("invalid format for min_score: must be a valid float value", e);
+        }
     }
 
     /**
--- a/src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
@@ -215,38 +215,6 @@ public class MinMaxScoreNormalizationTechnique implements ScoreNormalizationTech
         return lowerBound.getMode().normalize(score, minScore, maxScore, lowerBound.getMinScore());
     }
 
-    private boolean shouldIgnoreLowerBound(LowerBound lowerBound) {
-        return !lowerBound.isEnabled() || lowerBound.getMode() == Mode.IGNORE;
-    }
-
-    private float normalizeWithoutLowerBound(float score, float minScore, float maxScore) {
-        float normalizedScore = (score - minScore) / (maxScore - minScore);
-        return normalizedScore == 0.0f ? MIN_SCORE : normalizedScore;
-    }
-
-    private float normalizeWithLowerBound(float score, float minScore, float maxScore, LowerBound lowerBound) {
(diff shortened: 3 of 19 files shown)


src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -57,7 +57,7 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
 
     private Integer paginationDepth;
 
-    static final int MAX_NUMBER_OF_SUB_QUERIES = 5;
+    public static final int MAX_NUMBER_OF_SUB_QUERIES = 5;
```
martin-gaievski, line 60: making this public to reaad and make validation in other places down the stream

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -167,22 +225,182 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
+                    Locale.ROOT,
+                    "lower_bounds size %d should be less than or equal to %d",
+                    lowerBoundsParams.size(),
+                    MAX_NUMBER_OF_SUB_QUERIES
+                )
+            );
+        }
+
+        for (Object boundObj : lowerBoundsParams) {
+            if (!(boundObj instanceof Map)) {
```
vibrantvarun, line 263:
```suggestion
            if ((boundObj instanceof Map)==false) {
```
improves readability

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -167,22 +225,182 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
+
+    private Optional<List<Pair<LowerBound.Mode, Float>>> getLowerBounds(final Map<String, Object> params) {
+        if (Objects.isNull(params) || !params.containsKey(PARAM_NAME_LOWER_BOUNDS)) {
+            return Optional.empty();
+        }
+
+        List<Pair<LowerBound.Mode, Float>> lowerBounds = new ArrayList<>();
+
+        Object lowerBoundsObj = params.get(PARAM_NAME_LOWER_BOUNDS);
+        if (!(lowerBoundsObj instanceof List<?> lowerBoundsParams)) {
```
vibrantvarun, line 247:
```suggestion
        if ((lowerBoundsObj instanceof List<?> lowerBoundsParams)==false) {
```

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -167,22 +225,182 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
 
-    private float normalizeSingleScore(final float score, final float minScore, final float maxScore) {
+    private float normalizeSingleScore(final float score, final float minScore, final float maxScore, final LowerBound lowerBound) {
         // edge case when there is only one score and min and max scores are same
         if (Floats.compare(maxScore, minScore) == 0 && Floats.compare(maxScore, score) == 0) {
             return SINGLE_RESULT_SCORE;
         }
-        float normalizedScore = (score - minScore) / (maxScore - minScore);
-        return normalizedScore == 0.0f ? MIN_SCORE : normalizedScore;
+        if (!lowerBound.isEnabled()) {
```
vibrantvarun, line 233:
```suggestion
        if (lowerBound.isEnabled()==false) {
```

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -167,22 +225,182 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
+            }
+
+            @SuppressWarnings("unchecked")
+            Map<String, Object> lowerBound = (Map<String, Object>) boundObj;
+
+            try {
+                LowerBound.Mode mode = LowerBound.Mode.fromString(
+                    Objects.isNull(lowerBound.get(PARAM_NAME_LOWER_BOUND_MODE))
+                        ? ""
+                        : lowerBound.get(PARAM_NAME_LOWER_BOUND_MODE).toString()
```
vibrantvarun, line 274:
```suggestion
                LowerBound.Mode mode = LowerBound.Mode.fromString(
                    Objects.nonNull(lowerBound.get(PARAM_NAME_LOWER_BOUND_MODE))
                        ? lowerBound.get(PARAM_NAME_LOWER_BOUND_MODE).toString():""
```

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -167,22 +225,182 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
         }
-        float normalizedScore = (score - minScore) / (maxScore - minScore);
-        return normalizedScore == 0.0f ? MIN_SCORE : normalizedScore;
+        if (!lowerBound.isEnabled()) {
+            return LowerBound.Mode.IGNORE.normalize(score, minScore, maxScore, lowerBound.getMinScore());
+        }
+        return lowerBound.getMode().normalize(score, minScore, maxScore, lowerBound.getMinScore());
+    }
+
+    private Optional<List<Pair<LowerBound.Mode, Float>>> getLowerBounds(final Map<String, Object> params) {
```
vibrantvarun, line 239: Please add comments on what is happening in this method to better understand.

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -34,8 +42,28 @@
 public class MinMaxScoreNormalizationTechnique implements ScoreNormalizationTechnique, ExplainableTechnique {
```
vibrantvarun, line 42: We have added static class LowerBound in the same class. Can we extract it to independent file? If in future we add similar functionality then we have to rework to extract Lower bound and put it in another file.

src/main/java/org/opensearch/neuralsearch/processor/normalization/ScoreNormalizationUtil.java
```diff
@@ -82,4 +82,78 @@ public static void setNormalizedScore(
...
+    ) {
+        if (Objects.isNull(actualParameters) || actualParameters.isEmpty()) {
+            return;
+        }
+        boolean hasUnknownParameters = false;
+        for (Map.Entry<String, Object> entry : actualParameters.entrySet()) {
+            String paramName = entry.getKey();
+            Object paramValue = entry.getValue();
+
+            if (!supportedParametersTopLevel.contains(paramName)) {
```
vibrantvarun, line 115:
```suggestion
            if (supportedParametersTopLevel.contains(paramName)==false) {
```

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
vibrantvarun: It is a nit. Just for better readablity

vibrantvarun (commented):
Completed 1st round of review

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: not sure I got your reasoning. I don't think it needs to be in a separate file, it's essentially a record class, and I don't see a need for it in other techniques in near future. We can do this later if needed, currently it's over engineering

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: I'm not in favor of this style, will keep it unchanged

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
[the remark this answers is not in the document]
martin-gaievski: yes, I think I can split into two mwethods

martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: same as above, I do like the minimalistic syntax

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: same as above, I do like the minimalistic syntax

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: not sure what's the benefit of this change, and format is enforced by spotless, will keep my version

src/main/java/org/opensearch/neuralsearch/processor/normalization/ScoreNormalizationUtil.java
martin-gaievski: same as above, I do like the minimalistic syntax

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
```diff
@@ -150,4 +151,78 @@ private ScoreDoc deepCopyScoreDoc(final ScoreDoc scoreDoc, final boolean isSortE
...
+            if ((firstDoc == null) != (secondDoc == null)) {
+                return false;
+            }
+            if (firstDoc == null) {
+                continue;
+            }
+            if (firstDoc.doc != secondDoc.doc || Float.compare(firstDoc.score, secondDoc.score) != 0) {
+                return false;
+            }
+            if (firstDoc instanceof FieldDoc != secondDoc instanceof FieldDoc) {
```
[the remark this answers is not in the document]
martin-gaievski: I not sure why that's needed

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
owaiskazi19: @martin-gaievski checking boolean expression with `==` flag is [OpenSearch dev guide lines](https://github.com/opensearch-project/OpenSearch/blob/main/DEVELOPER_GUIDE.md#java-language-formatting-guidelines). We should follow the same

martin-gaievski: ok, let's follow the guideline in core then, although I don't see benefits in that syntax. I want to tweak plugin rules so this is applied automatically, will work on this in a separate issue: https://github.com/opensearch-project/neural-search/issues/1203

owaiskazi19: The above change won't be handled by spotless though. Spotless is used for styling and formatting and the above one is a code change

martin-gaievski: to me this is styling, and this can be done via spotless custom rules

martin-gaievski: I've changed all negations to explicit comparisons with false, should be good for changes in this PR

vibrantvarun (approved):
Looks good to me.

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -167,22 +224,197 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
+        // validate that the input parameters are in correct format
+        if (Objects.isNull(params) || params.containsKey(PARAM_NAME_LOWER_BOUNDS) == false) {
+            return Optional.empty();
+        }
+
+        Object lowerBoundsObj = params.get(PARAM_NAME_LOWER_BOUNDS);
+        if (lowerBoundsObj instanceof List<?> == false) {
+            throw new IllegalArgumentException("lower_bounds must be a List");
+        }
+        List<?> lowerBoundsParams = (List<?>) lowerBoundsObj;
```
owaiskazi19, line 253: How about?
```suggestion
      List<?> lowerBoundsParams = Optional.ofNullable(params.get(PARAM_NAME_LOWER_BOUNDS))
        .filter(List.class::isInstance)
        .map(List.class::cast)
        .orElseThrow(() -> new IllegalArgumentException("lower_bounds must be a non-null List"));
```

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -167,22 +224,197 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
+        // number of lower bounds must match the number of sub-queries in a hybrid query
+        if (lowerBoundsParams.size() > MAX_NUMBER_OF_SUB_QUERIES) {
+            throw new IllegalArgumentException(
+                String.format(
+                    Locale.ROOT,
+                    "lower_bounds size %d should be less than or equal to %d",
```
owaiskazi19, line 259: For my understanding
Don't we have to support lower bound for every sub query? For the ones we don't we can have mode as `ignore`. That should mean the lower bound size should be equal to sub query size.

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -167,22 +224,197 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
+    private static class MinMaxScores {
         float[] minScoresPerSubquery;
         float[] maxScoresPerSubquery;
     }
+
+    /**
+     * Result class to hold lower bound for each sub query
+     */
+    @Getter
+    public static class LowerBound {
```
owaiskazi19, line 327: Since this is a public class anyway. We could move it to a separate class file

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -167,22 +224,197 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
+        static final float MIN_LOWER_BOUND_SCORE = -10_000f;
```
owaiskazi19, line 328: How are deciding this bound scores for min and max?

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -167,22 +224,197 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
+        /**
+         * Enum for normalization mode
+         */
+        protected enum Mode {
+            APPLY {
+                @Override
+                public float normalize(float score, float minScore, float maxScore, float lowerBoundScore) {
+                    // if we apply the lower bound this mean we use actual score in case it's less then the lower bound min score
+                    // same applied to case when actual max_score is less than lower bound min score
+                    if (maxScore < lowerBoundScore || score < lowerBoundScore) {
```
owaiskazi19, line 355: Why are we checking `maxScore` here? Shouldn't it be `minScore` that needs to be compared with `lowerBoundScore`

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -167,22 +224,197 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
+                    if (score < minScore) {
+                        return 0.0f;
+                    }
+                    if (maxScore < lowerBoundScore) {
+                        return (score - minScore) / (maxScore - minScore);
+                    }
+                    return (score - lowerBoundScore) / (maxScore - lowerBoundScore);
+                }
+            },
+            IGNORE {
```
owaiskazi19, line 375: Do we need a separate enum for the original logic for min max? When the mode is IGNORE we could rely on the original normalize method of MinMax class

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -167,22 +224,197 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
+        String lowerBoundModeValue = Objects.toString(lowerBound.get(PARAM_NAME_LOWER_BOUND_MODE), "");
+        LowerBound.Mode mode = LowerBound.Mode.fromString(lowerBoundModeValue);
+        float minScore = extractAndValidateMinScore(lowerBound);
+
+        return ImmutablePair.of(mode, minScore);
+    }
+
+    private float extractAndValidateMinScore(Map<String, Object> lowerBound) {
+        Object minScoreObj = lowerBound.get(PARAM_NAME_LOWER_BOUND_MIN_SCORE);
+        if (minScoreObj == null) {
```
owaiskazi19, line 293: A question here. If a user is defining lower bounds then shouldn't we mandate them to provide a min score otherwise is there any point to define lower bound?

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -167,22 +224,197 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
+            return LowerBound.DEFAULT_LOWER_BOUND_SCORE;
+        }
+        try {
+            float minScore = LowerBound.DEFAULT_LOWER_BOUND_SCORE;
+            if (Objects.nonNull(lowerBound.get(PARAM_NAME_LOWER_BOUND_MIN_SCORE))) {
+                minScore = Float.parseFloat(String.valueOf(lowerBound.get(PARAM_NAME_LOWER_BOUND_MIN_SCORE)));
```
owaiskazi19, line 299: Should we also check if the provided min score is float if not throw an error?

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -167,22 +224,197 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
+            },
+            CLIP {
+                @Override
+                public float normalize(float score, float minScore, float maxScore, float lowerBoundScore) {
+                    // apply clipping, return lower bound min score if score is less than min score. This effectively means 0 after
+                    // normalization
+                    if (score < minScore) {
+                        return 0.0f;
+                    }
+                    if (maxScore < lowerBoundScore) {
```
owaiskazi19, line 369: Same here

owaiskazi19 (commented):
Took a look except the test file and rest LGTM

src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
```diff
@@ -32,6 +33,14 @@ public class L2ScoreNormalizationTechnique implements ScoreNormalizationTechniqu
     public static final String TECHNIQUE_NAME = "l2";
     private static final float MIN_SCORE = 0.0f;
 
+    public L2ScoreNormalizationTechnique() {
+        this(Map.of(), new ScoreNormalizationUtil());
+    }
+
+    public L2ScoreNormalizationTechnique(final Map<String, Object> params, final ScoreNormalizationUtil scoreNormalizationUtil) {
+        scoreNormalizationUtil.validateParameters(params, Set.of(), Map.of());
+    }
```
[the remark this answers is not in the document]
martin-gaievski: we actually validate, `params` has the actual parameters provided by user, and other two parameters mean we don't support any of those in L2 method. So if user actually sets some parameters this validation will fail with "your parameters are not supported".

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -34,8 +41,28 @@
 public class MinMaxScoreNormalizationTechnique implements ScoreNormalizationTechnique, ExplainableTechnique {
     @ToString.Include
     public static final String TECHNIQUE_NAME = "min_max";
-    private static final float MIN_SCORE = 0.001f;
+    protected static final float MIN_SCORE = 0.001f;
     private static final float SINGLE_RESULT_SCORE = 1.0f;
+    private static final String PARAM_NAME_LOWER_BOUNDS = "lower_bounds";
+    private static final String PARAM_NAME_LOWER_BOUND_MODE = "mode";
+    private static final String PARAM_NAME_LOWER_BOUND_MIN_SCORE = "min_score";
```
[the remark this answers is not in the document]
martin-gaievski: nope, each class has it's own set of constants, they are not common between techniques/classes

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: ack will make this change

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: yes, lower bound is for every sub-query. Here we're cutting extremely large collection with lower bounds even before we start parsing them. For hybrid query we can have up to 5 sub-queries, and here we're checking if lower bounds list does not have more then 5 elements.

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: Public scope is too broad for this class, will change it to package private and keep it under MinMaxTechnique main class

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: those are random numbers, just to have some limits to avoid extreme edge cases. In theory score are unbounded for base scoring (TF-IDF, BM25), but typically range between 0 and around 20-30 for most practical cases. For semantic queries it depends, for non-normalized vectors and spaces like inner product theory score are unbounded, but if vectors are normalized they are within [-1...1]

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: we could for sure, but as you mentioned we're adding "ignore" as one of the new modes. What I can do is to refactor this logic into a common method that is used from both places: when not lower bound provided, and for "ignore" mode

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: I think that was one of the asks when we reviewed the design/RFC, we can use default of 0.0 that is giving reasonable results and simplify the interface for end user

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: this is checked in the wrapping try/catch, we do following exception in such case:
```
throw new IllegalArgumentException("invalid format for min_score: must be a valid float value", e);
```

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: this is an edge case I bumped into while working on implementation. Imagine most of your docs are not very relevant to the query and the lower bound min_score is relatively high - this is the scenario we're catching here. We effectively falling back to traditional min-max for the lack of better option.

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: please check my previous response, it's the case when user has the lower bound that is too high

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: we already going just that, it no lower bounds we do call Mode.IGNORE.normalize() https://github.com/martin-gaievski/neural-search/blob/feature/lower_bounds_for_minmax_normalization/src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java#L232-L233

src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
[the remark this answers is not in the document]
martin-gaievski: we have a test for exactly this scenario https://github.com/martin-gaievski/neural-search/blob/feature/lower_bounds_for_minmax_normalization/src/test/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechniqueTests.java#L324

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -167,22 +224,196 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
+                        return (score - minScore) / (maxScore - minScore);
+                    }
+                    return (score - lowerBoundScore) / (maxScore - lowerBoundScore);
+                }
+            },
+            IGNORE {
+                @Override
+                public float normalize(float score, float minScore, float maxScore, float lowerBoundScore) {
+                    // ignore lower bound logic and do raw min-max normalization using actual scores
+                    float normalizedScore = (score - minScore) / (maxScore - minScore);
```
junqiu-lei, line 378: How does the normalization logic handle negative scores? Are there edge cases when scores are negative, and have we added tests to cover these scenarios?

src/test/java/org/opensearch/neuralsearch/query/HybridQueryExplainIT.java
```diff
@@ -703,6 +668,154 @@ public void testExplain_whenRRFProcessor_thenSuccessful() {
         assertTrue((double) explanationsHit4.get("value") > 0.0f);
     }
 
+    @SneakyThrows
+    public void testExplain_whenMinMaxNormalizationWithLowerBounds_thenSuccessful() {
```
junqiu-lei, line 672: Shall we also add some edge cases in IT, like case: lower bound greater than max score?

junqiu-lei (approved):
Overall LGTM.

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: this is "ignore" mode that is the same logic we do have today for min-max. in this case there is no lower bound score involved, so none of these scores should be negative because:
- all scores are positive
- min and max calculated based on those positive scores

src/test/java/org/opensearch/neuralsearch/query/HybridQueryExplainIT.java
martin-gaievski: added test case to org.opensearch.neuralsearch.processor.NormalizationProcessorIT

owaiskazi19 (commented):
Looks good overall with few questions

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: [-5. 5] can be too tight, even for typical TF-IDF, BM25 cases it can go as high as 30

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: not sure what do you mean. I was referring to scenario when user sets just lower bound, in this case we use 0.0 score as min_score for that lower bound. Reason for it is a simpler request syntax and relatively good defaults for entry level users. User can be not aware of optimal min_score for lower bound, or can be fine with our defaults.

owaiskazi19: So in the scenario where user just define a lower bound but doesn't pass any min score then it will be directed to default value, right?
Wouldn't in such scenario rather than using the default min value we should have used the original min score we get from the sub query itself?

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: This should be common case, minScore should be lower then the lowerBound score in general. In this case we mainly case how actual score compares with lowerBound score, not minScore with lowerBound score. In case doc score is >= then lowerBound score we use lowerBound as min score:

```(score - lowerBoundScore) / (maxScore - lowerBoundScore);```

in case doc score < lowerBound score then we keep raw min-max formula:

```(score - minScore) / (maxScore - minScore)```

owaiskazi19 (approved):

PR #1204, opened 2025-02-27
Fix a bug to unflatten the doc with list of map with multiple entries correctly

bzhangam:
### Description
Fix a bug to unflatten the doc with list of map with multiple entries correctly.

Let's say we have a doc like:
```
{
   "test.products":[
      {
         "product_description":"product_description 1",
         "price":10
      },
      {
         "product_description":"product_description 2",
         "price":20
      }
   ]
}
```
If we define an ingest pipeline to create the embedding for the product_description we will try to unflatten the doc first to ensure the **.** in the field name will be handled properly. Before my fix the doc will be unflattened like:

```
{
   "test":{
      "products":[
         {
            "price":10
         },
         {
            "price":20
         }
      ]
   }
}
```
We lose the product_description because when we handle the map in a list we incorrectly do `targetList.set(targetList.size() - 1, tempMap);`. With my fix the doc will be unflattened correctly:

```
{
   "test":{
      "products":[
         {
            "product_description":"product_description 1",
            "price":10
         },
         {
            "product_description":"product_description 2",
            "price":20
         }
      ]
   }
}
```

### Related Issues
There is no related issue.

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Set neural-search plugin 3.0.0 baseline JDK version to JDK-21 ([#838](https://github.com/opensearch-project/neural-search/pull/838))
 - Support different embedding types in model's response ([#1007](https://github.com/opensearch-project/neural-search/pull/1007))
 ### Bug Fixes
+- Fix a bug to unflatten the doc with list of map with multiple entries correctly ([#1204](https://github.com/opensearch-project/neural-search/pull/1204)).
 ### Infrastructure
 - [3.0] Update neural-search for OpenSearch 3.0 compatibility ([#1141](https://github.com/opensearch-project/neural-search/pull/1141))
 ### Documentation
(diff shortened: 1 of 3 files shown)


will-hwang:
looks good to me. signoff your commit

will-hwang (approved):

heemin32:
Please fix the failed checks.

martin-gaievski:
few questions here:
- what does it mean for end-user if we loose that prefix part in the flattened document? 
- what the impact of this change, which exact processors can be affected by this change? To me it looks like all inference processors and that includes both single and batch modes, is this correct?

bzhangam:
> few questions here:
> 
> * what does it mean for end-user if we loose that prefix part in the flattened document?
> * what the impact of this change, which exact processors can be affected by this change? To me it looks like all inference processors and that includes both single and batch modes, is this correct?

Yeah all the inference processors are impacted by this bug. If we run into the bug the doc will lose some data and the missing data will not be stored to the source or index. The missing data is removed from the ingestDoc.

bzhangam:
> @bzhangam an alternative to consider is:
> 
> we could treat
> 
> ```
> {
> "product_description":"product_description 1",
> "price":10
> }
> ```
> 
> as a single item to be pushed to stack, and and modify unflattenSingleItem. That will allow the existing logic to work as is. I'll leave it up to you to decide, as this will require additional changes

I think if that one line ch

heemin32 (approved):
LGTM. Thanks!

junqiu-lei (approved):
Thanks for the fixing, LGTM.

Issue #1135, opened 2025-01-22
[RFC] Common Filter Support for Hybrid Query Sub-Queries


bzhangam:
<h2> Introduction</h2>

This document proposes a solution to support common filter for the sub-queries of the hybrid query in [Hybrid Search](https://opensearch.org/docs/latest/search-plugins/hybrid-search/).

<h2> Problem Statement</h2>

Currently hybrid query clause doesn't support a simple way to add a common filter to its sub-queries which leads to adding of duplicate filters in the sub-queries. This leads to user copying same query clause across sub-queries and a sub-optimal user experience.

<h2>Scope </h2>
Currently if we want to apply a filter to all the sub-queries of a Hybrid query we need to do it this way:

```
{
  "hybrid": {
    "queries": [
      {
        "neural": {
            "filter": <filter>,
            ...
        }
      },
      {
        "bool":{
          "must": [
            <sub-query2>
          ],
          "filter": [
            <filter>
          ]
        }
      },
      {
        "bool":{
          "must": [
            <sub-query3>
          ],
          "filter": [
            <filter>
          ]
        }
      }
    ]
  }
}
```

We are looking for a way to simplify the hybrid query so that we don’t need to add the duplicated common filter to each sub-query.

<h3>Out of the Scope</h3>

**Metrics.** Currently we haven’t set up the metric API for the neural plugin and we plan have a separate project for that so for this one we will not consider publishing the metrics for this new feature.

**Configurability.** We currently plan to implement a straightforward approach for applying a common filter to all sub-queries. If a sub-query already has its own filter, the common filter will be combined using AND logic. This aligns with the intuitive expectation that adding a common filter to a hybrid query should apply it to all inner queries. To move quickly and gather feedback, we will focus on this simple default behavior. However, in the future, we may explore adding configurability to support the following options:

* Allow sub-query to decide when to use the common filter. Either during the sub-query or post it.
* Allow sub-query to decide using AND, OR or IGNORE_IF_EXIST(ignore the common filter if the inner query already has its own filter) logic to combine the common filter with sub-query’s own filter.

[...]

Issue #282, opened 2023-09-04
[FEATURE] Add filtering to Hybrid query 


martin-gaievski:
### Is your feature request related to a problem?
Hybrid query does not have support for filtering clause at high level

### What solution would you like?
User should be able to define filter at high level, that filter is applied to every sub-query. If that sub-query doesn't support filtering it's ignored (alternatively query may fail, that is reasonable alternative). In case filtering has been defined for particular sub-query high level filter is ignored.

Example of desired syntax:

```
{
   "query": {
       "hybrid": {
          "queries": [
               {},// First Query
               {} // Second Query
               ..... // Other Queries
           ],
           "filter" : {}
       } 
   }
}
```

### What alternatives have you considered?
Individual sub-queries can have filtering clauses defined, in case it's required for multiple sub-queries that maybe inconvenient.

bzhangam:
I'll work on this item.

heemin32:
Regarding the filtering, I believe it should function as a boolean query where the filter is applied before combining the results from the subqueries, rather than being passed to each subquery individually.
In other words, the filter should act as post-filtering for the individual subqueries but as pre-filtering for the overall hybrid query.

https://opensearch.org/docs/latest/query-dsl/compound/bool/

For passing filter to sub queries, the implementation and maintenance effort would outweigh its benefits, especially since users can already define filters for each subquery if needed.

martin-gaievski:
we can do this type of filtering today with `post_filter` keyword. Is this something you had in mind @heemin32 https://opensearch.org/docs/latest/search-plugins/hybrid-search/#hybrid-search-with-post-filtering.

Main cons of post filtering are not reliable number of final results (like neural or knn where we set number of desired hits inside the query), and worse performance in case individual query has low level optimizations in area of filtering.

heemin32:
I’m aware of post-filtering, but it works slightly differently.  

In the case of filtering in a boolean query, the filter is applied after the subquery is executed but before the boolean operation is performed. For hybrid queries, we should follow a similar approach: applying the filter after the subquery is executed but before score normalization and combination.  

Post-filtering, on the other hand, occurs after score normalization and combination.  

The filter in boolean query can lead to unreliable final results as well, particularly when the subquery is neural or k-NN. In such cases, users must include the filter within the subquery itself when they use boolean query.

bzhangam:
Create an RFC for this feature: https://github.com/opensearch-project/neural-search/issues/1135

[...]

PR #1206, opened 2025-03-03
Add filter function for NeuralQueryBuilder and HybridQueryBuilder and…

chloewqg:
… modify fromXContent function in HybridQueryBuilder to support filter field.

### Description
Add filter function to NeuralQueryBuilder and HybridQueryBuilder, which allows to push down non null filter. One exception is that when HybridQueryBuilder has a nested HybridQueryBuilder, then calling the filter function will cause UnsupportedOperationException

### Related Issues
Resolves #1206 
Related: #282, #1135

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
@@ -8,6 +8,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.Locale;
 import java.util.Objects;
 import java.util.stream.Collectors;
@@ -51,6 +52,7 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
     public static final String NAME = "hybrid";
 
     private static final ParseField QUERIES_FIELD = new ParseField("queries");
+    private static final ParseField FILTER_FIELD = new ParseField("filter");
     private static final ParseField PAGINATION_DEPTH_FIELD = new ParseField("pagination_depth");
 
     private final List<QueryBuilder> queries = new ArrayList<>();
@@ -91,6 +93,26 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
             throw new IllegalArgumentException(String.format(Locale.ROOT, "inner %s query clause cannot be null", NAME));
         }
         queries.add(queryBuilder);
+        return this;
+    }
+
+    /**
+     * Function to support filter on HybridQueryBuilder filter.
+     * If the filter is null, then we do nothing and return.
+     * Otherwise, we push down the filter to queries list.
+     * @param filter the filter parameter
+     * @return HybridQueryBuilder itself
+     */
+    public QueryBuilder filter(QueryBuilder filter) {
+        if (validateFilterParams(filter) == false) {
+            return this;
+        }
+        ListIterator<QueryBuilder> iterator = queries.listIterator();
+        while (iterator.hasNext()) {
+            QueryBuilder query = iterator.next();
+            // set the query again because query.filter(filter) can return new query.
+            iterator.set(query.filter(filter));
+        }
         return this;
     }
 
@@ -155,6 +177,10 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
      *                       }
      *                    }
      *               ]
+     *               "filter":
+     *                  "term": {
+     *                      "text": "keyword"
+     *                  }
      *          }
      *     }
      * }
@@ -168,6 +194,7 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
 
         Integer paginationDepth = null;
         final List<QueryBuilder> queries = new ArrayList<>();
+        QueryBuilder filter = null;
         String queryName = null;
 
         String currentFieldName = null;
@@ -178,6 +205,8 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
             } else if (token == XContentParser.Token.START_OBJECT) {
                 if (QUERIES_FIELD.match(currentFieldName, parser.getDeprecationHandler())) {
                     queries.add(parseInnerQueryBuilder(parser));
+                } else if (FILTER_FIELD.match(currentFieldName, parser.getDeprecationHandler())) {
+                    filter = parseInnerQueryBuilder(parser);
                 } else {
                     log.error(String.format(Locale.ROOT, "[%s] query does not support [%s]", NAME, currentFieldName));
                     throw new ParsingException(
@@ -240,7 +269,11 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
             compoundQueryBuilder.paginationDepth(paginationDepth);
         }
         for (QueryBuilder query : queries) {
-            compoundQueryBuilder.add(query);
+            if (filter == null) {
+                compoundQueryBuilder.add(query);
+            } else {
+                compoundQueryBuilder.add(query.filter(filter));
+            }
         }
         return compoundQueryBuilder;
     }
--- a/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
@@ -309,6 +309,24 @@ public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder>
         RescoreParser.streamOutput(out, rescoreContext);
     }
 
+    /**
+     * Add a filter to Neural Query Builder
+     * @param filterToBeAdded filter to be added
+     * @return return itself with underlying filter combined with passed in filter
+     */
+    public QueryBuilder filter(QueryBuilder filterToBeAdded) {
+        if (validateFilterParams(filterToBeAdded) == false) {
+            return this;
+        }
+        if (filter == null) {
+            filter = filterToBeAdded;
+        } else {
+            filter = filter.filter(filterToBeAdded);
+        }
+        return this;
+
+    }
+
     @Override
     protected void doXContent(XContentBuilder xContentBuilder, Params params) throws IOException {
         xContentBuilder.startObject(NAME);
--- a/src/test/java/org/opensearch/neuralsearch/query/HybridQueryFilterIT.java
+++ b/src/test/java/org/opensearch/neuralsearch/query/HybridQueryFilterIT.java
@@ -1,0 +1,233 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.query;
+
+import com.google.common.primitives.Floats;
+import lombok.SneakyThrows;
+import org.junit.BeforeClass;
+import org.opensearch.index.query.MatchQueryBuilder;
+import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.index.query.QueryBuilders;
+import org.opensearch.index.query.RangeQueryBuilder;
+import org.opensearch.index.query.TermQueryBuilder;
+import org.opensearch.neuralsearch.BaseNeuralSearchIT;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import static org.opensearch.neuralsearch.util.TestUtils.TEST_DIMENSION;
+import static org.opensearch.neuralsearch.util.TestUtils.TEST_SPACE_TYPE;
+import static org.opensearch.neuralsearch.util.TestUtils.createRandomVector;
+
+public class HybridQueryFilterIT extends BaseNeuralSearchIT {
+    private static final String TEST_MULTI_DOC_INDEX_WITH_TEXT_AND_INT_MULTIPLE_SHARDS =
+        "test-hybrid-post-filter-multi-doc-index-multiple-shards";
+    private static final String TEST_MULTI_DOC_INDEX_WITH_TEXT_AND_INT_SINGLE_SHARD =
+        "test-hybrid-post-filter-multi-doc-index-single-shard";
+    private static final String SEARCH_PIPELINE = "phase-results-hybrid-post-filter-pipeline";
+    private static final String INTEGER_FIELD_1_STOCK = "stock";
+    private static final String TEXT_FIELD_1_NAME = "name";
+    private static final String KEYWORD_FIELD_2_CATEGORY = "category";
+    private static final String TEXT_FIELD_VALUE_1_DUNES = "Dunes part 1";
+    private static final String TEXT_FIELD_VALUE_2_DUNES = "Dunes part 2";
+    private static final String TEXT_FIELD_VALUE_3_MI_1 = "Mission Impossible 1";
+    private static final String TEXT_FIELD_VALUE_4_MI_2 = "Mission Impossible 2";
+    private static final String TEXT_FIELD_VALUE_5_TERMINAL = "The Terminal";
+    private static final String TEXT_FIELD_VALUE_6_AVENGERS = "Avengers";
+    private static final String TEST_QUERY_TEXT = "Hello world";
+    private static final int INTEGER_FIELD_STOCK_1_25 = 25;
+    private static final int INTEGER_FIELD_STOCK_2_22 = 22;
+    private static final int INTEGER_FIELD_STOCK_3_256 = 256;
+    private static final int INTEGER_FIELD_STOCK_4_25 = 25;
+    private static final int INTEGER_FIELD_STOCK_5_20 = 20;
+    private static final String KEYWORD_FIELD_CATEGORY_1_DRAMA = "Drama";
+    private static final String KEYWORD_FIELD_CATEGORY_2_ACTION = "Action";
+    private static final String KEYWORD_FIELD_CATEGORY_3_SCI_FI = "Sci-fi";
+    private static final int SHARDS_COUNT_IN_SINGLE_NODE_CLUSTER = 1;
+    private static final String TEST_KNN_VECTOR_FIELD_NAME_1 = "test-knn-vector-1";
+    private final float[] testVector1 = createRandomVector(TEST_DIMENSION);
+    private final float[] testVector2 = createRandomVector(TEST_DIMENSION);
+    private final float[] testVector3 = createRandomVector(TEST_DIMENSION);
+
+    @BeforeClass
+    @SneakyThrows
+    public static void setUpCluster() {
+        // we need new instance because we're calling non-static methods from static method.
+        // main purpose is to minimize network calls, initialization is only needed once
+        HybridQueryFilterIT instance = new HybridQueryFilterIT();
+        instance.initClient();
+        instance.updateClusterSettings();
+    }
+
+    @SneakyThrows
+    public void testFilterOnNeuralQueryFilterAndTermQueryFilter_thenSuccessful() {
+        updateClusterSettings(CONCURRENT_SEGMENT_SEARCH_ENABLED, false);
+        prepareResourcesBeforeTestExecution(SHARDS_COUNT_IN_SINGLE_NODE_CLUSTER);
+        testNeuralQueryBuilder(TEST_MULTI_DOC_INDEX_WITH_TEXT_AND_INT_SINGLE_SHARD);
+    }
+
+    @SneakyThrows
+    private void testNeuralQueryBuilder(String indexName) {
+        String modelId = null;
+        modelId = prepareModel();
+        NeuralQueryBuilder neuralQueryBuilderTextQuery = NeuralQueryBuilder.builder()
+            .fieldName(TEST_KNN_VECTOR_FIELD_NAME_1)
+            .queryText(TEST_QUERY_TEXT)
+            .modelId(modelId)
+            .k(1)
+            .build();
+
+        TermQueryBuilder termQueryBuilder = QueryBuilders.termQuery(TEXT_FIELD_1_NAME, TEST_QUERY_TEXT);
+
+        HybridQueryBuilder hybridQueryBuilder = new HybridQueryBuilder();
+        hybridQueryBuilder.add(neuralQueryBuilderTextQuery);
+        hybridQueryBuilder.add(termQueryBuilder);
+
+        hybridQueryBuilder.filter(new MatchQueryBuilder("_id", "1"));
+
+        Map<String, Object> searchResponseAsMapTextQuery = search(indexName, hybridQueryBuilder, 1);
+        assertEquals(1, getHitCount(searchResponseAsMapTextQuery));
+
+        Map<String, Object> firstInnerHitTextQuery = getFirstInnerHit(searchResponseAsMapTextQuery);
+        assertEquals("1", firstInnerHitTextQuery.get("_id"));
+    }
+
+    @SneakyThrows
+    void prepareResourcesBeforeTestExecution(int numShards) {
+        if (numShards == 1) {
+            initializeIndexIfNotExists(TEST_MULTI_DOC_INDEX_WITH_TEXT_AND_INT_SINGLE_SHARD, numShards);
+        } else {
+            initializeIndexIfNotExists(TEST_MULTI_DOC_INDEX_WITH_TEXT_AND_INT_MULTIPLE_SHARDS, numShards);
+        }
+        createSearchPipelineWithResultsPostProcessor(SEARCH_PIPELINE);
+    }
+
+    @SneakyThrows
+    private void initializeIndexIfNotExists(String indexName, int numShards) {
+        if (!indexExists(indexName)) {
+            prepareKnnIndex(
+                indexName,
+                Collections.singletonList(new KNNFieldConfig(TEST_KNN_VECTOR_FIELD_NAME_1, TEST_DIMENSION, TEST_SPACE_TYPE))
+            );
+
+            addKnnDoc(
+                indexName,
+                "1",
+                Collections.singletonList(TEST_KNN_VECTOR_FIELD_NAME_1),
+                Collections.singletonList(Floats.asList(testVector1).toArray()),
+                Collections.singletonList(TEXT_FIELD_1_NAME),
+                Collections.singletonList(TEXT_FIELD_VALUE_2_DUNES),
+                List.of(),
+                List.of(),
+                List.of(INTEGER_FIELD_1_STOCK),
+                List.of(INTEGER_FIELD_STOCK_1_25),
+                List.of(KEYWORD_FIELD_2_CATEGORY),
+                List.of(KEYWORD_FIELD_CATEGORY_1_DRAMA),
+                List.of(),
+                List.of()
+            );
+
+            addKnnDoc(
+                indexName,
+                "2",
+                Collections.singletonList(TEST_KNN_VECTOR_FIELD_NAME_1),
+                Collections.singletonList(Floats.asList(testVector2).toArray()),
+                Collections.singletonList(TEXT_FIELD_1_NAME),
+                Collections.singletonList(TEXT_FIELD_VALUE_1_DUNES),
+                List.of(),
+                List.of(),
+                List.of(INTEGER_FIELD_1_STOCK),
+                List.of(INTEGER_FIELD_STOCK_2_22),
+                List.of(KEYWORD_FIELD_2_CATEGORY),
+                List.of(KEYWORD_FIELD_CATEGORY_1_DRAMA),
+                List.of(),
+                List.of()
+            );
+
+            addKnnDoc(
+                indexName,
+                "3",
+                Collections.singletonList(TEST_KNN_VECTOR_FIELD_NAME_1),
+                Collections.singletonList(Floats.asList(testVector3).toArray()),
+                Collections.singletonList(TEXT_FIELD_1_NAME),
+                Collections.singletonList(TEXT_FIELD_VALUE_3_MI_1),
+                List.of(),
+                List.of(),
+                List.of(INTEGER_FIELD_1_STOCK),
+                List.of(INTEGER_FIELD_STOCK_3_256),
+                List.of(KEYWORD_FIELD_2_CATEGORY),
+                List.of(KEYWORD_FIELD_CATEGORY_2_ACTION),
+                List.of(),
+                List.of()
+            );
+
+            addKnnDoc(
+                indexName,
+                "4",
+                List.of(),
+                List.of(),
+                Collections.singletonList(TEXT_FIELD_1_NAME),
+                Collections.singletonList(TEXT_FIELD_VALUE_4_MI_2),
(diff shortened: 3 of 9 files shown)


src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -309,6 +309,23 @@ protected void doWriteTo(StreamOutput out) throws IOException {
         RescoreParser.streamOutput(out, rescoreContext);
     }
 
+    /**
+     * Add a filter to Neural Query Builder
+     * @param filterToBeAdded filter to be added
+     * @return return itself with underlying filter combined with passed in filter
+     */
+    public QueryBuilder filter(QueryBuilder filterToBeAdded) {
+        if (validateFilterParams(filterToBeAdded) == false) {
```
junqiu-lei, line 318: the method `validateFilterParams` is missed i think.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -94,6 +95,31 @@ public HybridQueryBuilder add(QueryBuilder queryBuilder) {
...
+     * and throw exception telling this is an unsupported operation. If the filter is null, then we do nothing and
+     * return.
+     * @param filter the filter parameter
+     * @return HybridQueryBuilder itself
+     */
+    public QueryBuilder filter(QueryBuilder filter) {
+        if (!validateFilterParams(filter)) {
+            return this;
+        }
+        HybridQueryBuilder compoundQueryBuilder = new HybridQueryBuilder();
```
heemin32, line 110: Why do we have to create a new query builder instead of returning `this`?

martin-gaievski: I cannot find validateFilterParams method. Besides that the naming is confusing, if you conclude validity of the object based on return value that needs to be in the name, something like `isValidFilterParams`. Also change negation to an explicit == false

martin-gaievski (commented):
there is a separate issue and the whole design/RFC for hybrid query. Is this a building block for that design or this is a conflicting issue? 
Ref: https://github.com/opensearch-project/neural-search/issues/282, https://github.com/opensearch-project/neural-search/issues/1135

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
chloewqg: Hi the validateFilterParams is in the core package, inside the AbstractQueryBuilder class. https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/AbstractQueryBuilder.java#L94

chloewqg: > Why do we have to create a new query builder instead of returning `this`?

Because I saw the the queries of HybridQueryBuilder is marked as final. https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java#L56

So I think as we need to reassign a list of queries to the queries variable inside the HybridQueryBuilder. So we can only create a new HybridQueryBuilder to solve this?

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -94,6 +95,31 @@ public HybridQueryBuilder add(QueryBuilder queryBuilder) {
         return this;
     }
 
+    /**
+     * Function to support filter function on HybridQueryBuilder filter. Currently pushing down a filter
```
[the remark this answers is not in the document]
chloewqg: Updated.

chloewqg:
> there is a separate issue and the whole design/RFC for hybrid query. Is this a building block for that design or this is a conflicting issue? Ref: #282, #1135

Hi Martin, this is a building block for that design. cc @bzhangam

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
martin-gaievski: > Thanks Martin for the comment! Actually the function name `validateFilterParams` was suggested by Daniel in the pull request of core package. [opensearch-project/OpenSearch#17409](https://github.com/opensearch-project/OpenSearch/pull/17409) , is it okay to leave as what it is for now?

if method is already there let's utilize it. Please use == false syntax instead of negation, I see you've used it in other places in your PR

martin-gaievski: > > Why do we have to create a new query builder instead of returning `this`?
> 
> Because I saw the the queries of HybridQueryBuilder is marked as final. https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java#L56
> 
> So I think as we need to reassign a list of queries to the queries variable inside the HybridQueryBuilder. So we can only create a new HybridQueryBuilder to solve this?

Can't we do this.add(query.filter(filter))? Final doesn't mean we cannot mutate the object, just cannot point ref to another object

martin-gaievski:
> > there is a separate issue and the whole design/RFC for hybrid query. Is this a building block for that design or this is a conflicting issue? Ref: #282, #1135
> 
> Hi Martin, this is a building block for that design. cc @bzhangam

Got it, makes sense then. Can you please mention same in the PR description and also update the link to the feature request issue under "Related issues"? If the filtering feature requires multiple PRs/building blocks I suggest you create a high-level meta issue and host smaller PRs under that meta issue.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
chloewqg: Wouldn't this double the filters sizes? Like if there are two filters, when we add(query.filter(filter)), wouldn't this create duplicate filters?

chloewqg: > > Thanks Martin for the comment! Actually the function name `validateFilterParams` was suggested by Daniel in the pull request of core package. [opensearch-project/OpenSearch#17409](https://github.com/opensearch-project/OpenSearch/pull/17409) , is it okay to leave as what it is for now?
> 
> if method is already there let's utilize it. Please use == false syntax instead of negation, I see you've used it in other places in your PR

Yep. Updated in the newer commit

chloewqg:
> > > there is a separate issue and the whole design/RFC for hybrid query. Is this a building block for that design or this is a conflicting issue? Ref: #282, #1135
> > 
> > 
> > Hi Martin, this is a building block for that design. cc @bzhangam
> 
> Got it, makes sense then. Can you please mention same in the PR description and also update the link to the feature request issue under "Related issues"? If the filtering feature requires multiple PRs/building blocks I suggest you create a high-level meta issue and host smaller PRs under that meta issue.

Gotcha. Thanks!

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
chloewqg: > Quote reply

I think the filter input check is common across all the query builders. Although current check is simple of whether it equals to null, adding a function here can allow complicated logic changes in the future

bzhangam: The reason we need to create a new HybridQueryBuilder here is because this function is a static function and we cannot set the value for the parameters in a static function unless we also make them static. I don't think we want to make all the parameters static so we have to create an instance in this static function.

src/test/java/org/opensearch/neuralsearch/query/NeuralQueryIT.java
```diff
@@ -176,6 +177,101 @@ public void testQueryWithBoostAndImageQueryAndRadialQuery() {
...
+     *                          "match_none" : {}
+     *                     ]
+     *                 }
+     *             }
+     *         }
+     *     }
+     * }
+     */
+    @SneakyThrows
+    public void testQueryWithBoostAndFilterApplied() {
```
bzhangam, line 244: Why we add this IT? I think the filter function is an existing feature for neural query?

src/test/java/org/opensearch/neuralsearch/query/HybridQueryPostFilterIT.java
```diff
@@ -167,6 +174,93 @@ private void testPostFilterRangeQuery(String indexName) {
...
+                                }
+                            }
+                        ]
+                    }
+                ]
+            }
+        }
+    }*/
+    @SneakyThrows
+    private void testRangeQueryAsFilter(String indexName) {
```
bzhangam, line 235: Why we add this IT in PostFilterIT? Filter is different from post filter. I think we should either add it to the HybridQueryIT or create a new HybridQueryFilterIT.

bzhangam: Do you plan to raise another PR to add tests for BWC? BWC tests are under the qa folder.

martin-gaievski: > Why we add this IT in PostFilterIT? Filter is different from post filter. I think we should either add it to the HybridQueryIT or create a new HybridQueryFilterIT.

+1, better add new class HybridQueryFilterIT

chloewqg: > Do you plan to raise another PR to add tests for BWC? BWC tests are under the qa folder.

I think the current bwc tests is blocking the merge. So I will do the bwc tests here in the same PR

chloewqg: > Why we add this IT in PostFilterIT? Filter is different from post filter. I think we should either add it to the HybridQueryIT or create a new HybridQueryFilterIT.

Sure. Will create a new HybridQueryFilterIT

src/test/java/org/opensearch/neuralsearch/query/NeuralQueryIT.java
chloewqg: Here we are testing whether adding another filter on NeuralQuery would be function as expected.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -94,6 +95,28 @@ public HybridQueryBuilder add(QueryBuilder queryBuilder) {
...
+     * and throw exception telling this is an unsupported operation. If the filter is null, then we do nothing and
+     * return.
+     * @param filter the filter parameter
+     * @return HybridQueryBuilder itself
+     */
+    public QueryBuilder filter(QueryBuilder filter) {
+        if (validateFilterParams(filter) == false) {
+            return this;
+        }
+        for (int i = 0; i < queries.size(); i++) {
```
heemin32, line 110: This is fine because the queries are array list. However, we should avoid the assumption that it will be ArrayList especially because we defined as `List`. 
```
private final List<QueryBuilder> queries = new ArrayList<>();
```

Therefore, following is better so that its performance does not degrade even if we switch to other list type.
```
ListIterator<QueryBuilder> iterator = queries.listIterator();
while(iterator.hasNext()) {
  QueryBuilder query = iterator.next();
  if (query instanceof HybridQueryBuilder) {
      throw new UnsupportedOperationException("Cannot push filter to nested hybridQueryBuilder");
  }
  iterator.set(query.filter(filter));
}
```

However, on a second thought, why do we need to set the query again? Also, we don't need to check if query is type of HybridQuery or not. It should be checked somewhere else. So we can simple do like
```
queries.stream().forEach(query -> query.filter(filter));
```

chloewqg: Gotcha! Thanks Heemin! Btw, I am little confused in the current bwc tests failure. Seems like the error message is 

```
Caused by: io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: 127.0.0.1/127.0.0.1:42235
»  Caused by: java.net.ConnectException: Connection refused
```

I am bit confused on what is wrong here with my filter function changes.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -94,6 +96,30 @@ public HybridQueryBuilder add(QueryBuilder queryBuilder) {
...
+     * @return HybridQueryBuilder itself
+     */
+    public QueryBuilder filter(QueryBuilder filter) {
+        if (validateFilterParams(filter) == false) {
+            return this;
+        }
+        ListIterator<QueryBuilder> iterator = queries.listIterator();
+        while(iterator.hasNext()) {
+            QueryBuilder query = iterator.next();
+            if (query instanceof HybridQueryBuilder) {
```
heemin32, line 114: I don't think we need this check here?

heemin32: That will be handled by https://github.com/opensearch-project/neural-search/pull/498/files#diff-64ab60781cf1e9f7963a415406f49283b26945beaa46b7042e7cc5044d8d95b1R74

I don't think this method should worry about it.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -94,6 +96,27 @@ public HybridQueryBuilder add(QueryBuilder queryBuilder) {
         return this;
     }
 
+    /**
+     * Function to support filter on HybridQueryBuilder filter. Currently pushing down a filter
+     * to HybridQueryBuilder is not supported by design. We would simply check if the filter is valid
+     * and throw exception telling this is an unsupported operation. If the filter is null, then we do nothing and
```
bzhangam, line 102: The description doesn't match the implementation. Have we changed the mind that we want to support the filter push down for a hybrid query? Basically it means we want to support a hybrid query nested in another hybrid query. I don't think we plan to support it in near future.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -51,6 +52,7 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
     public static final String NAME = "hybrid";
 
     private static final ParseField QUERIES_FIELD = new ParseField("queries");
+    private static final ParseField FILTER_FIELD = new ParseField("filter");
```
bzhangam, line 55: The new field should also be added to **HybridQueryBuilder(StreamInput in)** and **doXContent(XContentBuilder builder, Params params)** so that we will lose the filter info when we pass it across nodes. Even though the filter info is already included in the sub queries I think we still should follow the best practice to include it in those two functions.

Besides we also want to add it to the doEquals and doHashCode function in case they are used to compare two hybrid queries.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryPostFilterIT.java
bzhangam: I saw you created HybridQueryFilterIT. Do you plan to remove this test?

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
heemin32: I asked to remove the check of nested hybrid query here. I believe the blocking of the nested hybrid query should be handled elsewhere, but it doesn’t necessarily have to be in this specific location.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
heemin32: It seems like we are missing some test case here unless this filter setting is happening in coordination node.

src/test/java/org/opensearch/neuralsearch/query/NeuralQueryIT.java
bzhangam: I think IT is mock a real use case. If we want to test that we should test a hybrid query with a neural sub query. And both the hybrid query and the neural query have a filter.

In this test we simply test the filter function of the neural query builder which has already be done as a unit test.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFilterIT.java
```diff
@@ -0,0 +1,322 @@
...
+                                }
+                            }
+                        ]
+                    }
+                ]
+            }
+        }
+    }*/
+    @SneakyThrows
+    private void testRangeQueryAsFilter(String indexName) {
```
bzhangam, line 130: I think most common use case for hybrid query is using a neural query and a normal term query as sub queries. Can we build the IT for that use case?

One important thing we want to validate is that for neural query we should push the filter into it. It can give us different result compared to the case we use a bool query to combine a filter with a neural query. This happens because the filter in neural query will be applied before we return the top k result while the filter combined through a bool query will happen after returning the top k result. Would recommend to build a test case for that.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
heemin32: Still I think we should update the comment which does not match with the actual implementation.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFilterIT.java
chloewqg: Okay, sounds good

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
chloewqg: Sounds good

vibrantvarun (commented):
1. Please add bwc tests as this field is added new the query. We need to evaluate the impact in case of bwc environment.
2. Why we have not added this field in HybridQuery.java?

cc: @martin-gaievski @heemin32

heemin32:
> 2. Why we have not added this field in HybridQuery.java?
We just pass the filter down to sub queries. Therefore, there is no need for hybrid query to hold it as we are not going to filter after the sub queries return the result.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
chloewqg: Hi Bo, quick question while I am trying to follow your comments. 
```
The new field should also be added to HybridQueryBuilder(StreamInput in) and doXContent(XContentBuilder builder, Params params) so that we will lose the filter info.
```
What do you mean by we will lose the filter info? Could you please give an example?

```
add it to the doEquals and doHashCode function in case they are used to compare two hybrid queries.
```

The field I added here is not the filter itself but the filter parser field name. I believe the filter itself is pushed down to queries. So we are automatically comparing the queries and its filter in doEquals and doHashCode. What do you mean by adding it to the doEquals and doHashCode function?

bzhangam: For your first question when we pass the query across nodes we will convert it to a stream and later in another node build it based on the stream. If we don't clearly write the filter to the stream we will lose it.

For the second question we override the doEquals and doHashCode function so if we don't clearly check it those two functions will ignore it.

bzhangam: Synced offline. Since currently we always want to push down the filter to sub-queries and we always do this work in the fromXContent function there is no need to persist the filter in the HybridQuery. In this case there is not need to write it to the stream since it's already included in the sub-queries. Same for the doEqual and doHash functions.

martin-gaievski: I think this should be fine, one concern I do have is some edge case in context of cluster upgrade. Can you add a bwc test, it may be skipped for now for 2.x -> 3.x migration, but it will run in future for next 3.x, e.g. 3.0 -> 3.1. You can check how it's done in of the recent PR for stats API

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
martin-gaievski: Please state in the comment that filter is not stored in the top hybrid query but only pushed down to individual sub-queries

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -309,6 +309,23 @@ protected void doWriteTo(StreamOutput out) throws IOException {
...
+            return this;
+        }
+        if (filter == null) {
+            filter = filterToBeAdded;
+            return this;
+        }
+        filter = filter.filter(filterToBeAdded);
+        return this;
```
martin-gaievski, line 326:
```suggestion
        if (filter == null) {
            filter = filterToBeAdded;
        } else {
             filter = filter.filter(filterToBeAdded);
        }
        return this;
```

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
chloewqg: Yes. Will publish new revision for that.

src/test/java/org/opensearch/neuralsearch/query/NeuralQueryIT.java
```diff
@@ -176,6 +177,101 @@ public void testQueryWithBoostAndImageQueryAndRadialQuery() {
         );
     }
 
+    /**
+     * Tests basic query without filter:
```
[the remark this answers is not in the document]
chloewqg: Okay, I will remove the test here.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryPostFilterIT.java
```diff
@@ -74,6 +74,13 @@ public void testPostFilterOnIndexWithSingleShard_whenConcurrentSearchEnabled_the
         testPostFilterMatchAllAndMatchNoneQueries(TEST_MULTI_DOC_INDEX_WITH_TEXT_AND_INT_SINGLE_SHARD);
     }
 
+    @SneakyThrows
+    public void testFilterOnIndexWithSingleShard_whenConcurrentSearchEnabled_thenSuccessful() {
```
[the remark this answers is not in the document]
chloewqg: Yes will remove it.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -94,6 +96,27 @@ public HybridQueryBuilder add(QueryBuilder queryBuilder) {
...
+     */
+    public QueryBuilder filter(QueryBuilder filter) {
+        if (validateFilterParams(filter) == false) {
+            return this;
+        }
+        ListIterator<QueryBuilder> iterator = queries.listIterator();
+        while (iterator.hasNext()) {
+            QueryBuilder query = iterator.next();
+            // set the query again because query.filter(filter) can return new query.
+            iterator.set(query.filter(filter));
```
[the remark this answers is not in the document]
chloewqg: Sounds good

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -94,6 +96,29 @@ public HybridQueryBuilder add(QueryBuilder queryBuilder) {
...
+     */
+    public QueryBuilder filter(QueryBuilder filter) {
+        if (validateFilterParams(filter) == false) {
+            return this;
+        }
+        ListIterator<QueryBuilder> iterator = queries.listIterator();
+        while (iterator.hasNext()) {
+            final QueryBuilder query = iterator.next();
+            final QueryBuilder queryWithFilter = query.filter(filter);
+            if (!queryWithFilter.equals(query)) {
```
martin-gaievski, line 114: don't use `equals` method, you need to compare exact reference, like `queryWithFilter == query`. Also use == false instead of negation in if condition

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/HybridSearchWithRescoreIT.java
```diff
@@ -64,10 +64,15 @@ public void testHybridQueryWithRescore_whenIndexWithMultipleShards_E2EFlow() thr
...
-                HybridQueryBuilder hybridQueryBuilder = getQueryBuilder(modelId, null, null);
+                HybridQueryBuilder hybridQueryBuilder = getQueryBuilder(modelId, null, null, null);
                 QueryBuilder rescorer = QueryBuilders.matchQuery(TEST_FIELD, RESCORE_QUERY).boost(0.3f);
                 validateTestIndex(getIndexNameForTest(), hybridQueryBuilder, rescorer);
-                hybridQueryBuilder = getQueryBuilder(modelId, Map.of("ef_search", 100), RescoreContext.getDefault());
+                hybridQueryBuilder = getQueryBuilder(
+                    modelId,
+                    Map.of("ef_search", 100),
+                    RescoreContext.getDefault(),
+                    new MatchQueryBuilder("_id", "1")
```
martin-gaievski, line 74: why this change is needed, it doesn't look like a simple adoption of the new argument?

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFilterIT.java
```diff
@@ -0,0 +1,354 @@
...
+
+        Map<String, Object> searchResponseAsMapTextQuery = search(indexName, hybridQueryBuilder, 1);
+        assertEquals(1, getHitCount(searchResponseAsMapTextQuery));
+
+        Map<String, Object> firstInnerHitTextQuery = getFirstInnerHit(searchResponseAsMapTextQuery);
+        assertEquals("1", firstInnerHitTextQuery.get("_id"));
+    }
+
+    @SneakyThrows
+    public void testFilterOnIndexWithSingleShard_whenConcurrentSearchEnabled_thenSuccessful() {
```
martin-gaievski, line 106: I don't think we need this match of tests for CSS, one should be enough, this one you can delete

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/HybridSearchWithRescoreIT.java
chloewqg: Oh I checked previous PR and found they modify this for bwc tests. Shall I discard this change?

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFilterIT.java
```diff
@@ -0,0 +1,354 @@
...
+
+    @SneakyThrows
+    public void testFilterOnNeuralQueryFilterAndTermQueryFilter_thenSuccessful() {
+        updateClusterSettings(CONCURRENT_SEGMENT_SEARCH_ENABLED, true);
+        prepareResourcesBeforeTestExecution(SHARDS_COUNT_IN_SINGLE_NODE_CLUSTER);
+        testNeuralQueryBuilder(TEST_MULTI_DOC_INDEX_WITH_TEXT_AND_INT_SINGLE_SHARD);
+    }
+
+    @SneakyThrows
+    private void testNeuralQueryBuilder(String indexName) {
```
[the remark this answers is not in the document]
chloewqg: What do you mean by set the CSS setting explicitly?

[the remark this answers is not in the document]
chloewqg: Another question for [Build and Test Neural Search / Pre-commit Linux (23, ubuntu-latest) ] Failure. It is caused by `java.lang.NoClassDefFoundError: Could not initialize class org.opensearch.knn.index.engine.KNNEngine` Yesterday it works fine offline. What might be causing this issue?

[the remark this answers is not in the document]
martin-gaievski: > What do you mean by set the CSS setting explicitly?

if you want CSS to be disabled for the test, make sure to explicitly set it to `false`. It looks like you're enabling it in tests where you need it, and assuming that it will be reset to `false` before each next test. It's not the case if tests are sharing same index. In other words your test may be running against index where CSS is enabled when it's not required

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/HybridSearchWithRescoreIT.java
martin-gaievski: if by default filter can be `null` then pass it as `null`. With current logic you're testing multiple things, but the test it built just for rescore functionality

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFilterIT.java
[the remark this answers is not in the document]
martin-gaievski: > Another question for [Build and Test Neural Search / Pre-commit Linux (23, ubuntu-latest) ] Failure. It is caused by `java.lang.NoClassDefFoundError: Could not initialize class org.opensearch.knn.index.engine.KNNEngine` Yesterday it works fine offline. What might be causing this issue?

which run do you mean? I checked few failed ones at random, they are all flaky, model haven't loaded in time:

```
java.lang.AssertionError: failed to load the model, last task finished with status RUNNING
```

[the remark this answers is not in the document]
martin-gaievski: Another failure related to some recent changes in core, I think it's covered by this issue: https://github.com/opensearch-project/neural-search/issues/1232

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
heemin32: @martin-gaievski I don’t see any benefit from this change. The previous code was much better—using a `set` isn’t costly, even compared to an `equals` check. The earlier implementation was also simpler and more readable. Would it be possible to revert to the previous version?

By the way, in this code, we are calling `query.filter(filter);` two times.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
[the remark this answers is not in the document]
heemin32: @martin-gaievski, If you're okay with it, let's keep this as is. The difference between `equals` and `set` is negligible, so a simpler approach would be preferable.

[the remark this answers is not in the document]
martin-gaievski: ok, let's keep the `set()` method, I agree that original version of the code is simpler

martin-gaievski (approved):

Issue #1146, opened 2025-01-28
[FEATURE] Create stats api to capture hybrid query stats


vibrantvarun:
### Is your feature request related to a problem?

Create an api which maintains the stats of processors for hybrid query like normalization_processor, rrf_processor.

### What solution would you like?
Introducing stats APIs to display these metrics stored in memory.

### What alternatives have you considered?
Operating without access to these metrics.

### Do you have any additional context?
Cluster operators could use these metrics to analyze usage patterns and gain insights into how frequently specific features are being utilized.

martin-gaievski:
For hybridization processors we need fine granular stats: what exact technique, what are parameters for techniques. 
In addition to stats for processors I would like to see some metrics for the query itself: if they use pagination, push down filter, sorting etc.

Issue #1104, opened 2025-01-15
[FEATURE]Add Metrics for Neural Search Usage


heemin32:
### Is your feature request related to a problem?
I would like to track metrics related to the neural search feature, such as the number of neural query calls and the number of text_embedding processor calls.

### What solution would you like?
Introducing stats APIs to display these metrics stored in memory.

### What alternatives have you considered?
Operating without access to these metrics.

### Do you have any additional context?
Cluster operators could use these metrics to analyze usage patterns and gain insights into how frequently specific features are being utilized.

q-andy:
Hi, you can assign this issue to me

q-andy:
Hi, you can assign this to me instead

minalsha:
Hi @heemin32 , can we track features at the neural plugin level instead? This would help both hybrid and neural sparse queries as well.

q-andy:
@minalsha I'm looking at adding these at a plugin level through a /stats API similar to how [k-NN](https://opensearch.org/docs/latest/search-plugins/knn/api/) and [ml-commons](https://opensearch.org/docs/latest/ml-commons-plugin/api/stats/) stats APIs work, this should let us track features across neural, neural-sparse, and hybrid since they'll hook into the same interface. Our goal is also to make these metrics declarative and easy to implement and record on the developer side too, all through a single unified API.

For neural search use cases, we want to capture adoption for upcoming features on our roadmap like #803, #1138, and #1135. Also looking into what existing processor-level usage that would useful, like number of text chunking processor executions by algorithm or number of image vs text fields for the text image embedding processor.

@yuye-aws @zhichao-aws I'm wondering if there are any existing neural-sparse metrics you guys would be interested in capturing? Maybe something like sparse encoding processor executions by prune type? 

Tagging @vibrantvarun for hybrid as well

vibrantvarun:
https://github.com/opensearch-project/neural-search/issues/1146

cc: @minalsha @martin-gaievski

zhichao-aws:
@q-andy I think it would be good to track the sparse encoding processors usage and neural sparse query usage

minalsha:
@q-andy for hybrid, we want to track hybrid query usage as well as normalization processor and rrf usage(going to GA in 2.19).

martin-gaievski:
[...]

PR #1208, opened 2025-03-04
[Feature branch] Add Neural Stats API

### LINKED ISSUE #1196: [RFC] Neural Plugin Stats API
# [RFC] Neural Search Stats API

## Problem

The neural search plugin provides users a wide variety of features to power full text semantic search, such as neural search, hybrid search, and a variety other ML applications. 

To understand how the neural plugin is being used, cluster managers may want additional insight on things like which processors have the most executions or which hybrid search techniques are being employed most frequently. However, the only way to gather this information currently is through the OpenSearch core [node info API](https://opensearch.org/docs/latest/api-reference/nodes-apis/nodes-info/) which lacks granular, plugin-specific information.

## Proposal

In this RFC we propose **creating a /stats API for neural search** to provide the foundation to allow cluster managers to monitor adoption and operational info specific to the neural search plugin. This includes an optionally enabled backend to record and manage stat information, and an API to retrieve the information on demand. The purpose of this API is to provide high level information that cluster managers can use to observe usage trends over time.

**This will be an opt-in feature**. By default, stat collection will be disabled, and can be configured via cluster setting.

After the initial implementation is complete, our goal is to create documentation and drive new and existing features to onboard statistics they want to track onto this system. This may include:

**Existing features**

* Hybrid search (e.g. hybrid queries, normalization processor executions, breakdowns by normalization technique, combination technique) #1146 
* Neural search (e.g. neural queries, text embedding processor executions) #1104 
* Neural sparse search (e.g. neural sparse queries, sparse encoder processor executions, breakdowns by prune types, two phase processor executions)

**Upcoming features**

* #1175 Neural sentence highlighter
* #1138 Text embedding processor existing embedding optimization
* #1135 Push down filter for hybrid sub queries  
* #803 Neural field type for semantic search 

## Requirements

1. **Stats should track simple events and state inside the code**
    1. **Event stats:** Counter stats that increase with code events. Singled valued, numeric, node level counters.
        1. e.g. how many `RRFProcessor` executions have there been in the last 5 minutes?
        2. e.g. have there been `RRFProcessor` executions in the past 24 hours?
    2. **State stats:** Stats on existing cluster info that are determined on demand. Cluster level counters.
        1. e.g. How many `NormalizationProcessors` using `arithmetic_mean` does the cluster have in search pipelines?
2. **Recording stats should be declarative and easy to implement**
    1. Define a stat name/path, call a method to increment it, and it should available in the API automatically
3. **Stats should have minimal effect on performance**
    1. Recording and storing stats should have minimal memory and CPU footprint
4. **Stats should have single unified API for retrieval**
    1. A single API provides all stat info, eliminating the need to call multiple APIs and aggregate data. 
    2. Stats can have “stat metadata” with more information about stat values, such as value type or last active event
    3. Users should be able to filter nodes or stat names, or return the response flat JSON
5. **Cluster managers should have flexibility to configure stat collection for their clusters**
    1. For users who don’t require stats, there should be option to disable stats entirely to prevent additional resource consumption

## High level flow

* **Event-based stats**
    * Event stats are recorded in code at a node level (processor executions, documents ingested, etc)
    * When an API call is made, all node-level maps are fetched via transport action and returned in the response.
* **State stats**
    * State stats are defined by helper functions that populate state stat values
    * When an API call is made, the functions are invoked and the information is added to the response on demand

###  Stats API call fetches and returns stats

Image
## API Model

|API	|Method	|Status	|Mutating or Non-Mutating	|Functionality	|
|---	|---	|---	|---	|---	|
|/_plugins/_neural/stats	|GET	|New	|Non-Mutating	|Retrieves stat counters from nodes and returns them in response	|

**Path Parameters**

* `nodes`: specify node ids to retrieve stats from (default all)
* `stats`: specify stat names to retrieve (default all)

**Query Parameters**

* `include_metadata`: boolean, include recent_interval/stat_type/minutes_since_last_event (default false)
* `flatten`: boolean, flatten the JSON response (default false)

**Example calls**

```
GET /_plugins/_neural/stats
GET /_plugins/_neural/stats/include_metadata=false
GET /_plugins/_neural/<node_id>/stats/<stat_name>?include_metadata=true&flatten=true
```

**Cluster level setting to disable Stats API/Collection**

```
PUT /_cluster/settings
{
    "persistent" : {
        "plugins.neural_search.stats_enabled" : "false" // default false
    }
}
```

### **Get stats without metadata**

```
GET /_plugins/_neural/stats/
```

**Example Response:**

```
{
  "cluster_name": "integTest",
  "cluster_version": "3.0.0",
  "nodes": {
    "node_1": {
      "normalization_processor": {
        "normalization_processor_executions": 100,
        "normalization_technique_l2_executions": 80,
        "normalization_technique_z_score_executions": 20
      }
    }
  },
  "all_nodes": { ... },
  "state": {
    "normalization_processor": {
      "normalization_processors_in_pipelines_count": 3,
      "normalization_technique_z_score_in_pipelines_count": 1,
      "normalization_technique_l2_in_pipelines_count": 2
    }
  }
}
```

### **Get stats with metadata**

```
GET /_plugins/_neural/stats&include_metadata=true
```

**Example Response:**

```
{
    "cluster_name": "integTest",
    "cluster_version": "3.0.0",
    "stats_enabled": true,
    "nodes": {
        "node_1": {                
            "processors": {
                "normalization": {
                    "normalization_processor_executions": {
                        value: 100,
                        type: "event_counter",
                        trailing_interval_value: 9,
                        minutes_since_last_event: 5
                    },
                    "normalization_technique_l2_executions": {
                        value: 80,
                        type: "event_counter",
                        trailing_interval_value: 9,
                        minutes_since_last_event: 5
                    },
                    "normalization_technique_z_score_executions": {
                        value: 20,
                        type: "event_counter",
                        trailing_interval_value: 0,
                        minutes_since_last_event: 9999                    
                    }
                }
            }
        }
    },
    "all_nodes": {
       // ...
    }
    "state": {
        "processors": {
            "normalization": {
                "normalization_processors_in_pipelines_count": {
                    value: 3,
                    type: "state_counter"
                },
                "normalization_technique_z_score_in_pipelines_count": {
                    value: 1,
                    type: "state_counter"
                },
                "normalization_technique_l2_in_pipelines_count": {
                    value: 2,
                    type: "state_counter"
                }
            }
        }
    }
}
```

# Low Level Design

* **StatName** enums are enums used to uniquely identify stats. Each stat name maps directly to a single stat
    * **Stat Name** - uniquely identify the stat, will be used for filtering in user request
    * **Response Path** - the location in the response that the json will be located.
    * **StatType** is another Enum used by StatNames to define what kind of 
* **Stat** objects are objects that store information about a stat, including the stat value itself, + metadata

Image

```
// Declare enums
enum EventStatName {
   NORMALIZATION_PROCESSOR_EXECUTIONS(
        "normalization_processor_executions", // stat name
        "processors.normalization", // response path
        EventStatType.TIMESTAMPED_EVENT_COUNTER)
   ;
    
   // ...
}

enum StateStatName {
   NORMALIZATION_PROCESSORS_IN_PIPELINES(
        "normalization_processors_in_pipelines", 
        "processors.normalization",
        StateStatType.STATE_COUNTER)
   ;
    
   // ...
}
```

**Response**

```
{
    // Event stats are calculated on node level
    "nodes": {
        "data_node_1_asdjfalksdjfjasf": {
            "processors": {
                "normalization": {
                    "normalization_processor_executions": <value>
                }
            }
        }
    }
    // State stats are calculated on cluster level
    "state": {
        "processors": {
            "normalization": {
                "normalization_processors_in_pipelines": <value>
            }
        }
    }
}
```

## Event Stats

Image

### Stat Model

* **Value:** the cumulative count of the recorded event since nodes started
    * Storing the total allows for caller to manage delta calculations on their own
    * This is the format value used by other OpenSearch stat APIs
* **Type:** a type hint of the kind of value stored, used to help the Auto BR policy parse the response dynamically
* **Recent trailing time interval**: A trailing count of the event in a the past 5 minutes
    * This value will not be a perfect accurate count. The goal is to have a window into the overall trend of event statistics in a rolling time window, which can be used to observe broader usage patterns.
* **Last event time:** a timestamp of the last recorded event

## State Stats

Image

* Calculated on demand at call time, so less stat metadata data is needed

# Event Stat Flow

## Overall increment flow

Image

**Recording Stats**

1. Static method `increment()` call using StatName as a parameter
2. Finds Stat by StatName in the map
3. Calls the `eventStat.incremen``t()`
    1. Stat object handles updating count, updating time interval, and timestamp

**Getting Stats**

1. After API call,  transport action is sent to each node
2. On each node, `EventStatsManager.getStatsData()` is called to get the stat data based on the current store
3. This map is returned from each node

## Tracking recent time interval

High level diagram:

Image

**Implementation**
(see below for example)

1. **Option 1: Array with time buckets approach (preferred)**
    1. Hold a fixed length array with all buckets labelled by time (e.g. for 5 minutes, have a 6 element array, 5 past minutes + 1 current minute)
    2. Current bucket to increment is determined by system time in minutes % number of buckets
        1. At increment time, “current” bucket is accessed, if up to date, it’s incremented
        2. If it’s out of date, it will be reset with the new timestamp and overwritten
    3. To get values, iterate through all buckets, exclude expired buckets, and sum
    4. Pros:
        1. Constant time reads/writes
        2. Bucket rotation is determined at increment time, if there are no increments, there is no performance overhead for events
    5. Cons
        1. Difficult to extend with configurability, need to reinitialize fixed length arrays
2. **Option 2: Scheduled rotating queue approach**
    1. Current bucket to increment is maintained by atomic reference
    2. Hold a queue store for past buckets
    3. At scheduled time intervals the current bucket is automatically rotated into the queue and the last bucket is popped
    4. To get values, iterate through all buckets and sum
    5. Pros:
        1. Simpler mental model
        2. Easier to add configurability for interval size
    6. Cons:
        1. Need to manage scheduling system, leads to time desync issues
            1. Cannot guarentee perfect time aligned scheduled executions, which can lead to bucket time drift
            2. Performance overhead for buckets that do not need rotation, e.g. if there is a stat with no events, the scheduled executions would still run
        2. Need to manage atomic reference to current bucket

The following design will assume using option 1. Each event stat manages its own set of buckets.

### Update existing bucket happy case

Image
**Increment Call**
1. Increments the total value
2. Updates timestamp for time stamped stats
3. Determine which time bucket to access

```
// Get current time
long now = System.currentTimeMillis();

// Round down to minute
long currentBucketTime = now - (now % BUCKET_INTERVAL_MS); 

// Determine bucket index
int bucketIndex = (now // BUCKET_INTERVAL_MS) % BUCKETS_COUNT; 

// Get bucket 
Bucket bucket = buckets[bucketIndex];

// Check timestamp
if (bucket timestamp matches current timestamp) {
    // Increment bucket
    bucket.increment();
}
```

**Get recent interval call**

1. Get all buckets
2. Sum all buckets where bucket time follows conditions:
    1. Bucket time >= 5 minutes ago (rounded down)
    2. Bucket time < 1 minute ago (rounded down)

### Update new bucket

**Before Update**

Image

**After Update**

Image

Increment call is same as above, except
* Determine which time bucket to access
    * Same as above case, except:
    * **If bucket time is out of date, overwrite bucket time with current time and reset counter**
    * then Increment
* getRecentInterval() call is same as above

## Updating latest event timestamp

What format to return timestamp?

1. **Option 1: Relative timestamp (preferred)**
    1. e.g. “minutes since last event”
    2. Each stat on the node stores the most recent event timestamp locally
    3. When serialization call is made, the node uses its current system time to calculate a relative time (minutes_since_last_event) based on the most recent event
    4. Pros:
        1. Calculating node relative time minimizes impact of cluster time desync issues. Each node only compares to itself.
    5. Cons
        1. Timestamp will be relative to api call time rather than absolute
2. **Option 2: Unix timestamp**
    1. Each stat on the node stores the most recent event unix timestamp
    2. When serialization call is made, that timestamp is returned as is
    3. Pros:
        1. Simple to return a single timestamp, absolute time value has more flexibility to be used by caller
    4. Cons
        1. Nodes can have clock drift and face time sync issues, requires additional cluster level clock sync to ensure system time on a node is consistent across cluster

# State Stat Flow

Image

1. On coordinator node level , API request to get state stats
2. Create a new map to store state stat info
3. Call helper functions to get state stats
    1. Helper functions fetches info and update map with calculated values
4. ... more helper calls
5. ... more helper calls
6. ... more helper calls
7. Map is formatted to match user request
8. Return map with info to 

# Next steps

As mentioned above, our next steps after the framework is implemented is to onboarding existing and new feature use cases onto the system. 

In the future, as we add more complex stat use cases such as plugin health and operational monitoring, we also consider the following future enhancements

* Configurable time interval via cluster level setting
    * Custom time interval length (e.g. past 5 minutes, past 10 minutes, past 30 minutes)
    * Custom time bucket length (e.g. 1 minute granularity, 5 minute granularity)
    * Note: setting is changed at runtime, it would require resetting all event stats (reinitializing time bucket array) leading to data loss.
* State stat caching
    * If there any expensive on-demand calculations to compute state stats in the future, we can implement a caching system to reduce the impact of repeated calls 
* Detailed operational statistics
    * More detailed information for cluster health diagnostics, such as processor

q-andy:
### Description
Implementing Neural Stats API framework design proposed in #1196. This initial PR sets up the foundation for the framework to track event and state stats throughout the neural search plugin and exposed their values via API.

Image

* **Event-based stats**
    * Event stats are recorded in code at a node level (processor executions, documents ingested, etc)
    * When an API call is made, all node-level maps are fetched via transport action and returned in the response.
* **State stats**
    * State stats are defined by helper functions that populate state stat values
    * When an API call is made, the functions are invoked and the information is added to the response on demand

See RFC for more details.

Initial implementation includes 3 stats:
- Text embedding processor executions
- Text embedding processors in pipelines
- Cluster version

### Related Issues
Resolves #1196 #1104 
Related: #1146 

### Check List

### Example requests

```
GET /_plugins/_neural/stats
{
	"_nodes": {
		"total": 1,
		"successful": 1,
		"failed": 0
	},
	"cluster_name": "integTest",
	"cluster_version": "3.0.0",
	"processors": {
		"ingest": {
			"text_embedding_processors_in_pipelines": 0
		}
	},
	"all_nodes": {
		"processors": {
			"ingest": {
				"text_embedding_executions": 0
			}
		}
	},
	"nodes": {
		"r-hVPa7-Ra6FBbdpfVrHjg": {
			"processors": {
				"ingest": {
					"text_embedding_executions": 0
				}
			}
		}
	}
}
```

```
GET /_plugins/_neural/stats?include_metadata=true
{
	"_nodes": {
		"total": 1,
		"successful": 1,
		"failed": 0
	},
	"cluster_name": "integTest",
	"cluster_version": {
		"value": "3.0.0",
		"stat_type": "settable"
	},
	"processors": {
		"ingest": {
			"text_embedding_processors_in_pipelines": {
				"value": 0,
				"stat_type": "countable"
			}
		}
	},
	"all_nodes": {
		"processors": {
			"ingest": {
				"text_embedding_executions": {
					"value": 0,
					"stat_type": "timestamped_counter",
					"trailing_interval_value": 0,
					"minutes_since_last_event": 29018783
				}
			}
		}
	},
	"nodes": {
		"r-hVPa7-Ra6FBbdpfVrHjg": {
			"processors": {
				"ingest": {
					"text_embedding_executions": {
						"value": 0,
						"stat_type": "timestamped_counter",
						"trailing_interval_value": 0,
						"minutes_since_last_event": 29018783
					}
				}
			}
		}
	}
}
```

```
GET _plugins/_neural/stats/text_embedding_executions?include_metadata=true&flat_keys=true

{
	"_nodes": {
		"total": 1,
		"successful": 1,
		"failed": 0
	},
	"cluster_name": "integTest",
	"all_nodes.processors.ingest.text_embedding_executions": {
		"value": 0,
		"stat_type": "timestamped_counter",
		"trailing_interval_value": 0,
		"minutes_since_last_event": 29018784
	},
	"nodes": {
		"r-hVPa7-Ra6FBbdpfVrHjg": {
			"processors.ingest.text_embedding_executions": {
				"value": 0,
				"stat_type": "timestamped_counter",
				"trailing_interval_value": 0,
				"minutes_since_last_event": 29018784
			}
		}
	}
}
```


heemin32:
Should it be full flat?
```
GET _plugins/_neural/stats/text_embedding_executions?include_metadata=true&flat_keys=true

{
	"_nodes.total": 1,
        "_nodes.successful": 1,
        "_nodes.failed": 0
        "all_nodes.processors.ingest.text_embedding_executions.value": 0,
        ...
        "nodes.r-hVPa7-Ra6FBbdpfVrHjg.processors.ingest.text_embedding_executions.minutes_since_last_event": 29018784
}
```

q-andy:
It can go either way, but I think preserving some basic structure is helpful for navigating the high level categories. For example, cluster setting still divides the flat keys between `persistent`, `transient`, and `defaults` fields.

The way I see it, flatten is an option to simplify navigating the nested fields, not a hard rule to flatten everything, and keeping some info like metadata grouped together makes it easier to parse since than if it was completely flat.

Example cluster settings call with flatten enabled:

```
GET _cluster/settings?include_defaults=true&flat_settings=true

{
	"persistent": {
		"plugins.ml_commons.model_access_control_enabled": "true",
		"plugins.ml_commons.native_memory_threshold": "99",
		"plugins.ml_commons.only_run_on_ml_node": "false",
		"plugins.neural_search.stats_enabled": "true"
	},
	"transient": {},
	"defaults": {
		"_plugin_neural_search_hybrid_query_executor.queue_size": "1000",
		"_plugin_neural_search_hybrid_query_executor.size": "20",
		"action.auto_create_index": "true",
         . . .
       }
}
```
Compare to non-flat

```
GET _cluster/settings?include_defaults=true
{
	"persistent": {
		"plugins": {
			"neural_search": {
				"stats_enabled": "true"
			},
			"ml_commons": {
				"only_run_on_ml_node": "false",
				"model_access_control_enabled": "true",
				"native_memory_threshold": "99"
			}
		}
	},
	"transient": {},
	"defaults": {
		"task_resource_tracking": {
			"enabled": "true"
		},
	. . .
        },
        . . . 
}

**Comment by heemin32** [comment]
That make sense. We only need to flatten the key for easier search. Then, shouldn't it be like this?
```
"all_nodes" {
      "processors.ingest.text_embedding_executions": {
		"value": 0,
		"stat_type": "timestamped_counter",
		"trailing_interval_value": 0,
		"minutes_since_last_event": 29018784
	},
}
```

Also wondering if this could be better.
```
"all_nodes" {
      "processors.ingest.text_embedding_executions.value": 0,
      "processors.ingest.text_embedding_executions.stat_type": "timestamped_counter",
      "processors.ingest.text_embedding_executions.trailing_interval_value": 0,
      "processors.ingest.text_embedding_executions.minutes_since_last_event": 29018784
}
```

**Comment by heemin32** [comment]
There are two components: Stats API and Metric Management. For the Metric Management code, it might be better to have a separate package to make it more portable for use in other plugins.

**Comment by q-andy** [comment]
> Then, shouldn't it be like this? (all_nodes unflattened)

Yes, that makes more sense, good catch

> Also wondering if this could be better. (metadata flattened)

I considered that as well, what I'm thinking is the "stat metadata" object replaces the raw value so the flat key name is the same whether or not metadata is included or not. Like the format will always be at `"path.to.stat: <data>"`, sometimes the data is the raw value, sometimes its a the metadata, but the stat key name is same.

It also keeps things organized when you parse recursively. For example. if a cluster manager is periodically calling the stats API and parsing the response without hardcoding stat names, if we keep metadata non-flat, for every stat you can foreach on the categories with flat_keys and parse each stat and all its metadata one by one in a single pass. Whereas if it was completely flat metadata, running a foreach would need additional logic to split, sort, and group the metadata keys by the stat name again. Also you would probably never be interested in the value of `path.to.stat.stat_type` independently of the other stat metadata.

Just some thoughts. I don't see the usecase right now but I could still include an option to flatten the stat metadata as well, it's a pretty small change.

**Comment by q-andy** [comment]
Changed base to merge to feature branch instead of main

**Comment by vibrantvarun** [comment]
Please add the change in changelog.md

File: src/main/java/org/opensearch/neuralsearch/processor/util/RestActionUtils.java [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.processor.util;
+
+import org.opensearch.rest.RestRequest;
+
+import java.util.Optional;
+
+public class RestActionUtils {
```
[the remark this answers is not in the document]
vibrantvarun: Take reference from ProcessorUtils

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsHandler.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.Locale;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static org.opensearch.neuralsearch.plugin.NeuralSearch.NEURAL_BASE_URI;
+import static org.opensearch.neuralsearch.processor.util.RestActionUtils.splitCommaSeparatedParam;
+
+@Log4j2
+@AllArgsConstructor
```
vibrantvarun, line 36: Add javadoc

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsHandler.java
```diff
@@ -0,0 +1,157 @@
...
+    public static final String FLATTEN_PARAM = "flat_keys";
+    public static final String INCLUDE_METADATA_PARAM = "include_metadata";
+
+    private static final Set<String> EVENT_STAT_NAMES = EnumSet.allOf(EventStatName.class)
+        .stream()
+        .map(EventStatName::getNameString)
+        .map(String::toLowerCase)
+        .collect(Collectors.toSet());
+
+    private static final Set<String> STATE_STAT_NAMES = EnumSet.allOf(StateStatName.class)
```
vibrantvarun, line 48: Can you come up with more meaninful name? State_stat is not that reader friendly.

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsHandler.java
```diff
@@ -0,0 +1,157 @@
...
+    protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) {
+        if (settingsAccessor.getIsStatsEnabled() == false) {
+            // Process params, or else will automatically return a 400 instead of a 403
+            splitCommaSeparatedParam(request, "nodeId");
+            splitCommaSeparatedParam(request, "stat");
+            request.paramAsBoolean(FLATTEN_PARAM, false);
+            request.paramAsBoolean(INCLUDE_METADATA_PARAM, false);
+
+            return channel -> channel.sendResponse(new BytesRestResponse(RestStatus.FORBIDDEN, "Stats endpoint is disabled"));
+        }
```
vibrantvarun, line 81: Move this to a separate dedicated method as validation method.

QQ: why do you need to call split methods and updating request if you anyway sending forbidden request exception at line 80.

src/main/java/org/opensearch/neuralsearch/processor/util/RestActionUtils.java
vibrantvarun, line 11:
```suggestion
class RestActionUtils {
```

Make this class package-private

src/main/java/org/opensearch/neuralsearch/processor/util/RestActionUtils.java
```diff
@@ -0,0 +1,19 @@
...
+    public static Optional<String[]> splitCommaSeparatedParam(RestRequest request, String paramName) {
```
vibrantvarun, line 12: Why we need this class and validations seperately. Similar kind of methods are already being used by opensearch-core. Why not reuse them? Check RestSearchAction in OpenSearch core for the same.

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsHandler.java
```diff
@@ -0,0 +1,157 @@
...
+        );
+    }
+
+    /**
+     * Creates a NeuralStatsRequest from a RestRequest
+     *
+     * @param request Rest request
+     * @return NeuralStatsRequest
+     */
+    private NeuralStatsRequest getRequest(RestRequest request) {
```
vibrantvarun, line 99:
```suggestion
    private NeuralStatsRequest createNeuralStatsRequest(RestRequest request) {
```

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsHandler.java
```diff
@@ -0,0 +1,157 @@
...
+
+        NeuralStatsInput neuralStatsInput = createNeuralStatsInputFromRequestParams(request);
+
+        NeuralStatsRequest neuralStatsRequest = new NeuralStatsRequest(nodeIdsArr, neuralStatsInput);
+        neuralStatsRequest.timeout(request.param("timeout"));
+
+        return neuralStatsRequest;
+    }
+
+    NeuralStatsInput createNeuralStatsInputFromRequestParams(RestRequest request) {
```
vibrantvarun, line 115:
```suggestion
    private NeuralStatsInput createNeuralStatsInputFromRequestParams(RestRequest request) {
```

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsHandler.java
```diff
@@ -0,0 +1,157 @@
...
+                }
+            }
+        }
+
+        // If no stats are specified, add all stats to retrieve all by default
+        if (retrieveAllStats) {
+            neuralStatsInput.getEventStatNames().addAll(EnumSet.allOf(EventStatName.class));
+            neuralStatsInput.getStateStatNames().addAll(EnumSet.allOf(StateStatName.class));
+        }
+        return neuralStatsInput;
```
vibrantvarun, line 155:
```suggestion
  Optional<String[]> stats = splitCommaSeparatedParam(request, "stat");
    
    if (stats.isPresent()==false) {
        // No specific stats requested, add all stats by default
        addAllStats(neuralStatsInput);
        return;
    }

    // Process requested stats
    boolean anyStatAdded = processRequestedStats(stats.get(), neuralStatsInput);
    
    // If no valid stats were added, fall back to all stats
    if (anyStatAdded ==false) {
        addAllStats(neuralStatsInput);
    }
    
    return neuralStatsInput;
  }

private boolean processRequestedStats(String[] stats, NeuralStatsInput neuralStatsInput) {
    boolean statAdded = false;
    
    for (String stat : stats) {
        String normalizedStat = stat.toLowerCase(Locale.ROOT);
        
        if (EVENT_STAT_NAMES.contains(normalizedStat)) {
            neuralStatsInput.getEventStatNames().add(EventStatName.from(normalizedStat));
            statAdded = true;
        } else if (STATE_STAT_NAMES.contains(normalizedStat)) {
            neuralStatsInput.getStateStatNames().add(StateStatName.from(normalizedStat));
            statAdded = true;
        }
    }
    
    return statAdded;
}

private void addAllStats(NeuralStatsInput neuralStatsInput) {
    neuralStatsInput.getEventStatNames().addAll(EnumSet.allOf(EventStatName.class));
    neuralStatsInput.getStateStatNames().addAll(EnumSet.allOf(StateStatName.class));
}
```

vibrantvarun: Clean code.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
```diff
@@ -34,4 +34,11 @@ public final class NeuralSearchSettings {
         50,
         Setting.Property.NodeScope
     );
+
+    public static final Setting<Boolean> NEURAL_STATS_ENABLED = Setting.boolSetting(
```
vibrantvarun, line 38: Add javadoc

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettingsAccessor.java
```diff
@@ -0,0 +1,44 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.settings;
+
+import lombok.Getter;
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.neuralsearch.stats.events.EventStatsManager;
+
+public class NeuralSearchSettingsAccessor {
```
vibrantvarun, line 12: Add javadoc

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsHandler.java
```diff
@@ -0,0 +1,157 @@
...
+        .map(String::toLowerCase)
+        .collect(Collectors.toSet());
+
+    private static final Set<String> STATE_STAT_NAMES = EnumSet.allOf(StateStatName.class)
+        .stream()
+        .map(StateStatName::getNameString)
+        .map(String::toLowerCase)
+        .collect(Collectors.toSet());
+
+    private NeuralSearchSettingsAccessor settingsAccessor;
```
vibrantvarun, line 54: Move this declaration to top

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettingsAccessor.java
```diff
@@ -0,0 +1,44 @@
...
+    private static NeuralSearchSettingsAccessor INSTANCE;
+    private boolean initialized;
+
+    @Getter
+    private volatile Boolean isStatsEnabled;
```
vibrantvarun, line 17:
```suggestion
    private volatile boolean isStatsEnabled;
```

src/main/java/org/opensearch/neuralsearch/stats/NeuralStatsInput.java
```diff
@@ -0,0 +1,168 @@
...
+import org.opensearch.neuralsearch.stats.state.StateStatName;
+
+import java.io.IOException;
+import java.util.EnumSet;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Holds user requester input parameters for retrieving neural stats via rest handler API.
+ * It allows filtering statistics by node IDs, event statistic types, and state statistic types.
```
vibrantvarun, line 26:
```suggestion
 * Entity class that holds input parameters for retrieving neural stats.
 * It is responsible for filtering statistics by node IDs, event statistic types, and state statistic types.
```

src/main/java/org/opensearch/neuralsearch/stats/events/EventStatsManager.java
```diff
@@ -0,0 +1,107 @@
...
+import lombok.Getter;
+import org.opensearch.neuralsearch.settings.NeuralSearchSettingsAccessor;
+import org.opensearch.neuralsearch.stats.common.StatSnapshot;
+
+import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Manager class for event counter stats, used to track monotonically increasing counts plus some possible metadata
```
vibrantvarun, line 16:
```suggestion
 * Manager class for event counter stats, used to track monotonically increasing counts plus some possible metadata
```

Please rewrite to make it more reader friendly. It should provide high level information of this class.

vibrantvarun (commented):
Completed 1st round of review.

src/main/java/org/opensearch/neuralsearch/stats/NeuralStatsInput.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    /**
+     * Constructor for stream input
+     *
+     * @param input the StreamInput to read data from
+     * @throws IOException if there's an error reading from the stream
+     */
+    public NeuralStatsInput(StreamInput input) throws IOException {
+        nodeIds = input.readBoolean() ? new HashSet<>(input.readStringList()) : new HashSet<>();
```
heemin32, line 105: `readOptionalStringList();` might be better in terms of readability.
Wonder if we can make nodeIds as List so that we can avoid this list to set conversion.

src/main/java/org/opensearch/neuralsearch/stats/NeuralStatsInput.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    /**
+     * Helper method to write an EnumSet to a StreamOutput with a presence flag.
+     *
+     * @param out the StreamOutput to write to
+     * @param set the EnumSet to write
+     * @throws IOException If there's an error writing to the stream
+     */
+    private void writeOptionalEnumSet(StreamOutput out, EnumSet<?> set) throws IOException {
```
heemin32, line 134: Might be good to have corresponding `readOptionalEnumSet()`
Better if we can add this method inside core :)

src/main/java/org/opensearch/neuralsearch/stats/events/EventStatsManager.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     *  Instance level method to increment the counter for a specified event statistic.
+     *
+     * @param eventStatName The name of the event stat to increment
+     */
+    public void inc(EventStatName eventStatName) {
+        boolean enabled = settingsAccessor.getIsStatsEnabled();
+        if (enabled && stats.containsKey(eventStatName)) {
```
heemin32, line 72: This could become an issue when there are many `eventStatName` values. While `HashMap.get` is generally constant time, its performance may degrade as the number of items grows. Given that this method is called frequently, we may need to explore an alternative approach to access the instance directly for better efficiency.

heemin32 (commented):
Overall, I have two comments:  

1. The most frequently called method (currently the `increment` method) should always operate in O(1) time complexity, regardless of the number of metrics defined in the future. `Hash.get()` does not always guarantee O(1) performance when a large number of items are stored.  

2. We should avoid using the `public static instance()` pattern. This approach should only be considered as a last resort when no other alternatives exist, as it can make unit testing more difficult.

heemin32:
> > Then, shouldn't it be like this? (all_nodes unflattened)
> 
> Yes, that makes more sense, good catch
> 
> > Also wondering if this could be better. (metadata flattened)
> 
> I considered that as well, what I'm thinking is the "stat metadata" object replaces the raw value so the flat key name is the same whether or not metadata is included or not. Like the format will always be at `"path.to.stat: <data>"`, sometimes the data is the raw value, sometimes its a the metadata, but the stat key name is same.
> 
> It also keeps things organized when you parse recursively. For example. if a cluster manager is periodically calling the stats API and parsing the response without hardcoding stat names, if we keep metadata non-flat, for every stat you can foreach on the categories with flat_keys and parse each stat and all its metadata one by one in a single pass. Whereas if it was completely flat metadata, running a foreach would need additional logic to split, sort, and group the metadata keys by the stat name again. Also you would probably never be interested in the value of `path.to.stat.stat_type` independently of the other stat metadata.
> 
> Just some thoughts. I don't see the usecase right now but I could still include an option to flatten the stat metadata as well, it's a pretty small change.

Yes. One use case that I can think of is that, when operator want to grep stats containing processor, this whole flat option can be handy.
```
_plugin/neural/stats?flat=true | grep processor
```

src/main/java/org/opensearch/neuralsearch/stats/events/EventStatsManager.java
q-andy: Sure, I can have the EventStat be a member of the EventStatName enum and be instantiated in the enum constructor. That way we can access it directly without using a map at all.

If we have performance issues with hash map get, though, we do have other options, my understanding is the O(n) worst case happens when there are key object hash code collisions leading to putting objects in the same bucket and requiring a search on the bucket. There are ways we can mitigate this like implementing our hash code methods to ensure uniqueness, and Java has its own optimizations to address this like [storing collisions in a tree rather than linked list](https://openjdk.org/jeps/180).

I'll update again once I finish running performance benchmarks

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettingsAccessor.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Return instance of the cluster context, must be initialized first for proper usage
+     * @return instance of cluster context
+     */
+    public static synchronized NeuralSearchSettingsAccessor instance() {
```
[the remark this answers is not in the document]
q-andy: Sure, I can switch most of the singletons to injection. `PipelineServiceUtil`, `NeuralSearchSettingsAccessor`, and `StateStatsManager should be trivial. 

For the `EventStatsManager`, though, incrementing stats via static method call seems cleaner otherwise we'd have to add the instance to every single constructor of every object that wants to call it which is pretty invasive. I see OS core has an injector interface and uses @Inject for things like Transport Action, but it doesn't seem utilized in many other places, maybe for event stats its easier to keep the singleton instance

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsHandler.java
q-andy: Sure, I'll rename to `InfoStat`

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettingsAccessor.java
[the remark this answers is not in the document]
heemin32: Yes. For `EventStatsManager`, it might not suitable to inject.

src/main/java/org/opensearch/neuralsearch/stats/events/EventStatsManager.java
heemin32: Yes. Storing it on a tree is also not O(1). Also, even if the hashcode are distinct, I don't think they will all have their own bucket. The number of bucket can be smaller then the number of hashcode.

src/test/java/org/opensearch/neuralsearch/rest/RestNeuralStatsHandlerIT.java
```diff
@@ -0,0 +1,169 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@Log4j2
+public class RestNeuralStatsHandlerIT extends BaseNeuralSearchIT {
```
martin-gaievski, line 29: we also need a BWC test. It's fine if there is no 2.x version that supports stats, it's needed for future 3.x releases

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettingsAccessor.java
```diff
@@ -0,0 +1,44 @@
...
+        if (INSTANCE == null) {
+            INSTANCE = new NeuralSearchSettingsAccessor();
+        }
+        return INSTANCE;
+    }
+
+    public void initialize(ClusterService clusterService, Settings settings) {
+        if (initialized) return;
```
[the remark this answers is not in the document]
q-andy: ack, this method was removed

q-andy (commented):
Addressed review comments, I'll change the format of the response according to the discussion above (breaking down high level categories as unflat like `all_nodes`) and add examples in the next commit.

src/main/java/org/opensearch/neuralsearch/stats/NeuralStatsInput.java
heemin32: Until core is ready, we can have the method in this class temporary.

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
```diff
@@ -0,0 +1,195 @@
...
+     * @param stats map containing the stat to increment
+     * @param infoStatName the identifier for the stat to increment
+     */
+    private void increment(Map<InfoStatName, CountableInfoStatSnapshot> stats, InfoStatName infoStatName) {
+        incrementBy(stats, infoStatName, 1L);
+    }
+
+    private void incrementBy(Map<InfoStatName, CountableInfoStatSnapshot> stats, InfoStatName statName, Long amount) {
+        if (stats.containsKey(statName)) {
+            stats.get(statName).incrementBy(amount);
```
heemin32, line 148: Can we avoid `.get()` here as well?

q-andy:
Added BWC tests. Currently they don't run since there's no backwards versions, but included them to have basis for future 3.x versions.

Updated response formatting:

**Default**
```
GET {{ _.base_url }}/_plugins/_neural/stats
{
	"_nodes": {
		"total": 1,
		"successful": 1,
		"failed": 0
	},
	"cluster_name": "integTest",
	"info": {
		"cluster_version": "3.0.0",
		"processors": {
			"ingest": {
				"text_embedding_processors_in_pipelines": 0
			}
		}
	},
	"all_nodes": {
		"processors": {
			"ingest": {
				"text_embedding_executions": 0
			}
		}
	},
	"nodes": {
		"Qs3osnyfTz6AokNiPB2uRQ": {
			"processors": {
				"ingest": {
					"text_embedding_executions": 0
				}
			}
		}
	}
}
```
**Flatten**
```
GET {{ _.base_url }}/_plugins/_neural/stats?flat_stat_paths=true
{
	"_nodes": {
		"total": 1,
		"successful": 1,
		"failed": 0
	},
	"cluster_name": "integTest",
	"info": {
		"cluster_version": "3.0.0",
		"processors.ingest.text_embedding_processors_in_pipelines": 0
	},
	"all_nodes": {
		"processors.ingest.text_embedding_executions": 0
	},
	"nodes": {
		"Qs3osnyfTz6AokNiPB2uRQ": {
			"processors.ingest.text_embedding_executions": 0
		}
	}
}
```
**Flatten w/ metadata**
```
GET {{ _.base_url }}/_plugins/_neural/stats?flat_stat_paths=true&include_metadata=true
{
	"_nodes": {
		"total": 1,
		"successful": 1,
		"failed": 0
	},
	"cluster_name": "integTest",
	"info": {
		"cluster_version": {
			"value": "3.0.0",
			"stat_type": "settable_string"
		},
		"processors.ingest.text_embedding_processors_in_pipelines": {
			"value": 0,
			"stat_type": "info_countable"
		}
	},
	"all_nodes": {
		"processors.ingest.text_embedding_executions": {
			"value": 0,
			"stat_type": "timestamped_counter",
			"trailing_interval_value": 0,
			"minutes_since_last_event": 29028938
		}
	},
	"nodes": {
		"Qs3osnyfTz6AokNiPB2uRQ": {
			"processors.ingest.text_embedding_executions": {
				"value": 0,
				"stat_type": "timestamped_counter",
				"trailing_interval_value": 0,
				"minutes_since_last_event": 29028938
			}
		}
	}
}
```

qa/build.gradle
```diff
@@ -45,7 +45,9 @@ dependencies {
     testRuntimeOnly('com.jayway.jsonpath:json-path:2.9.0') {
         // OpenSearch core is using slf4j 1.7.36. Therefore, we cannot change the version here.
         exclude group: 'org.slf4j', module: 'slf4j-api'
+        exclude group: 'net.minidev', module: 'json-smart'
     }
+    testRuntimeOnly group: 'net.minidev', name:'json-smart', version: "${versions.json_smart}"
```
martin-gaievski, line 50: you need to rebase on latest main and remove this

martin-gaievski (approved):
Looks good to me, thanks Andy. please address my comment regarding the rebasing on latest main

qa/build.gradle
q-andy: My [fork branch](https://github.com/q-andy/neural-search-os/tree/neural-stats) is up to date with main with latest commits, I think this diff is from the [feature branch](https://github.com/opensearch-project/neural-search/tree/feature/neural-stats-api.) on this repo being out of date. I don't have write permissions to sync it, could you try syncing it and this should disappear?

vibrantvarun (approved):
Looks good

vibrantvarun:
@heemin32 if your all comments are resolved then can we merge this?

heemin32:
> @heemin32 if your all comments are resolved then can we merge this?

I think my comments are all resolved. G2G.

heemin32 (approved):

Issue #1138, opened 2025-01-23
[RFC] Optimizing Text Embedding Processor


will-hwang:
# Optimizing Text Embedding Processor

## Problem Statement

**Proposal:** https://github.com/opensearch-project/neural-search/issues/793

[Text Embedding Processor](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java) is a processor that is defined as an input to the ingest pipeline to create vector embeddings from text. In current state, text embedding processor makes model inference calls for every document ingestion or update. While this approach is necessary for generating embeddings during initial document ingestion, it's unnecessary to regenerate them during document update if the embedding-related fields remain unchanged. This inefficiency can lead to unnecessary cost increase for customers and computational overhead for model inference. This document discusses the design to achieving optimization of the Text Embedding Processor in Neural Search.

## Requirement

* Enable user to add a settingto configure Text Embedding Processor to call or skip model inferencing when appropriate
* Skip call to model inference when flag is enabled and the field for embeddings has not changed

## Out of Scope

* Skip the check for model ID change between initial ingestion of the document and the update document action due to complexity involved. 
* Exclude the feature in AOSS due to complexity involved with different indexing methodology

### Text Embedding Processor Configuration

Currently, Text Embedding Processor expects two fields: 

***model_id***: A model to be used for creating vector embeddings
***field_map***: specifies the name of the field from which to take the text (`text`) and the name of the field in which to record embeddings (`passage_embedding`)

```
{
  "description": "An NLP ingest pipeline",
  "processors": [
    {
      "text_embedding": {
        "model_id": "aVeif4oB5Vm0Tdw8zYO2",
        "field_map": {
          "text": "passage_embedding"
        }
      }
    }
  ]
}
```

**Reference:** [link](https://opensearch.org/docs/latest/search-plugins/neural-search-tutorial/#step-2a-create-an-ingest-pipeline-for-neural-search)

### Current Flows

In current flow, there is no difference between ingestion/update of the document. The embeddings are created every time a document is ingested and/or updated in the Text Embedding Processor. See below for different use case scenarios:

**Current Scenario 1: Single Document Update with embedding field**

![Image](https://github.com/user-attachments/assets/108a5126-f67d-43e4-943b-306b90d3b812)

Steps:

1. User ingests/updates Doc1
2. Text Embedding Processor invokes model inference for text via `MLCommonClientAccessor` 
[...]

PR #1217, opened 2025-03-07
implement batch document optimization for text embedding processor

will-hwang:
### Description
This PR includes changes for:

1. refactor of InferenceFilter to simplify logic for copying embeddings
2. implementation for batch document update optimization by overriding InferenceProcessor's `subBatchExecute` to include filter when `skip_existing` flag is on

### Proposed State [Batch Document Update]

proposed-batch-flow

Steps:

1. Process Maps are generated for each IngestDocument based on the Field Map defined. 
2. If skip_existing feature is set to true, filter the process map for each IngestDocument. 
    1. Existing documents are fetched via OpenSearch client’s Multi-Get Action to compare each of the existing inference text against its corresponding IngestDocument 
        1. if documents do not exist, or any exception is thrown, fallback to calling model inference
    2. Locate the embedding fields in each existing document
        1. Recursively traverse the process map to find embedding fields. Keeping track of the traversal path is required for look up in existing document. 
        2. Once embedding fields are found, attempt to copy embeddings from existing document to its corresponding ingest document. 
    3. If eligible, copy over the vector embeddings from existing document to its corresponding ingest document
        1. It is eligible for copy if inference text in ingest document and its corresponding existing document is the same, and embeddings for the inference text exist in its existing document. 
        2. Note, if in case of values in list, the fields in the same index are compared to determine text equality
    4. If eligible fields have been copied, remove the entry in process map 
3. Inference List is generated based on entries in Filtered Process Map. 
4. ML Commons InferenceSentence API is invoked with filtered inference list.
5. Embeddings for filtered inference list are generated in ML Commons. 
6. Embeddings for filtered inference list are mapped to target fields via entries defined in process map.
7. Embeddings for filtered inference list are populated to target fields in IngestDocument.

### Related Issues
HLD: https://github.com/opensearch-project/neural-search/issues/1138
### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
@@ -26,6 +26,8 @@ import lombok.Getter;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
+import org.opensearch.action.get.MultiGetItemResponse;
+import org.opensearch.action.get.MultiGetRequest;
 import org.opensearch.common.collect.Tuple;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.common.util.CollectionUtils;
@@ -54,6 +56,8 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
 
     public static final String MODEL_ID_FIELD = "model_id";
     public static final String FIELD_MAP_FIELD = "field_map";
+    public static final String INDEX_FIELD = "_index";
+    public static final String ID_FIELD = "_id";
     private static final BiFunction<Object, Object, Object> REMAPPING_FUNCTION = (v1, v2) -> {
         if (v1 instanceof Collection && v2 instanceof Collection) {
             ((Collection) v1).addAll((Collection) v2);
@@ -169,49 +173,91 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
      */
     abstract void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler, Consumer<Exception> onException);
 
+    /**
+     * This is the function which does actual inference work for subBatchExecute interface.
+     * @param ingestDocumentWrappers a list of IngestDocuments in a batch.
+     * @param handler a callback handler to handle inference results which is a list of objects.
+     */
     @Override
     public void subBatchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers, Consumer<List<IngestDocumentWrapper>> handler) {
-        if (CollectionUtils.isEmpty(ingestDocumentWrappers)) {
-            handler.accept(Collections.emptyList());
-            return;
-        }
+        try {
+            if (CollectionUtils.isEmpty(ingestDocumentWrappers)) {
+                handler.accept(Collections.emptyList());
+                return;
+            }
 
-        List<DataForInference> dataForInferences = getDataForInference(ingestDocumentWrappers);
-        List<String> inferenceList = constructInferenceTexts(dataForInferences);
-        if (inferenceList.isEmpty()) {
+            List<DataForInference> dataForInferences = getDataForInference(ingestDocumentWrappers);
+            List<String> inferenceList = constructInferenceTexts(dataForInferences);
+            if (inferenceList.isEmpty()) {
+                handler.accept(ingestDocumentWrappers);
+                return;
+            }
+            doSubBatchExecute(ingestDocumentWrappers, inferenceList, dataForInferences, handler);
+        } catch (Exception e) {
+            updateWithExceptions(ingestDocumentWrappers, e);
             handler.accept(ingestDocumentWrappers);
-            return;
         }
-        Tuple<List<String>, Map<Integer, Integer>> sortedResult = sortByLengthAndReturnOriginalOrder(inferenceList);
-        inferenceList = sortedResult.v1();
-        Map<Integer, Integer> originalOrder = sortedResult.v2();
-        doBatchExecute(inferenceList, results -> {
-            int startIndex = 0;
-            results = restoreToOriginalOrder(results, originalOrder);
-            for (DataForInference dataForInference : dataForInferences) {
-                if (dataForInference.getIngestDocumentWrapper().getException() != null
-                    || CollectionUtils.isEmpty(dataForInference.getInferenceList())) {
-                    continue;
+    }
+
+    /**
+     * This is a helper function for subBatchExecute, which invokes doBatchExecute for given inference list.
+     * @param ingestDocumentWrappers a list of IngestDocuments in a batch.
+     * @param inferenceList a list of String for inference.
+     * @param dataForInferences a list of data for inference, which includes ingestDocumentWrapper, processMap, inferenceList.
+     * @param handler a callback handler to handle inference results which is a list of objects.
+     */
+    protected void doSubBatchExecute(
+        List<IngestDocumentWrapper> ingestDocumentWrappers,
+        List<String> inferenceList,
+        List<DataForInference> dataForInferences,
+        Consumer<List<IngestDocumentWrapper>> handler
+    ) {
+        try {
+            Tuple<List<String>, Map<Integer, Integer>> sortedResult = sortByLengthAndReturnOriginalOrder(inferenceList);
+            inferenceList = sortedResult.v1();
+            Map<Integer, Integer> originalOrder = sortedResult.v2();
+            doBatchExecute(inferenceList, results -> {
+                try {
+                    int startIndex = 0;
+                    results = restoreToOriginalOrder(results, originalOrder);
+                    for (DataForInference dataForInference : dataForInferences) {
+                        if (dataForInference.getIngestDocumentWrapper().getException() != null
+                            || CollectionUtils.isEmpty(dataForInference.getInferenceList())) {
+                            continue;
+                        }
+                        List<?> inferenceResults = results.subList(startIndex, startIndex + dataForInference.getInferenceList().size());
+                        startIndex += dataForInference.getInferenceList().size();
+                        setVectorFieldsToDocument(
+                            dataForInference.getIngestDocumentWrapper().getIngestDocument(),
+                            dataForInference.getProcessMap(),
+                            inferenceResults
+                        );
+                    }
+                    handler.accept(ingestDocumentWrappers);
+                } catch (Exception e) {
+                    updateWithExceptions(ingestDocumentWrappers, e);
+                    handler.accept(ingestDocumentWrappers);
                 }
-                List<?> inferenceResults = results.subList(startIndex, startIndex + dataForInference.getInferenceList().size());
-                startIndex += dataForInference.getInferenceList().size();
-                setVectorFieldsToDocument(
-                    dataForInference.getIngestDocumentWrapper().getIngestDocument(),
-                    dataForInference.getProcessMap(),
-                    inferenceResults
-                );
-            }
-            handler.accept(ingestDocumentWrappers);
-        }, exception -> {
-            for (IngestDocumentWrapper ingestDocumentWrapper : ingestDocumentWrappers) {
-                // The IngestDocumentWrapper might already run into exception and not sent for inference. So here we only
-                // set exception to IngestDocumentWrapper which doesn't have exception before.
-                if (ingestDocumentWrapper.getException() == null) {
-                    ingestDocumentWrapper.update(ingestDocumentWrapper.getIngestDocument(), exception);
+            }, exception -> {
+                try {
+                    for (IngestDocumentWrapper ingestDocumentWrapper : ingestDocumentWrappers) {
+                        // The IngestDocumentWrapper might already run into exception and not sent for inference. So here we only
+                        // set exception to IngestDocumentWrapper which doesn't have exception before.
+                        if (ingestDocumentWrapper.getException() == null) {
+                            ingestDocumentWrapper.update(ingestDocumentWrapper.getIngestDocument(), exception);
+                        }
+                    }
+                    handler.accept(ingestDocumentWrappers);
+                } catch (Exception e) {
+                    updateWithExceptions(ingestDocumentWrappers, e);
+                    handler.accept(ingestDocumentWrappers);
                 }
-            }
+
+            });
+        } catch (Exception e) {
+            updateWithExceptions(ingestDocumentWrappers, e);
             handler.accept(ingestDocumentWrappers);
-        });
+        }
     }
 
     private Tuple<List<String>, Map<Integer, Integer>> sortByLengthAndReturnOriginalOrder(List<String> inferenceList) {
@@ -238,7 +284,7 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
         return sortedResults;
     }
 
-    private List<String> constructInferenceTexts(List<DataForInference> dataForInferences) {
+    protected List<String> constructInferenceTexts(List<DataForInference> dataForInferences) {
         List<String> inferenceTexts = new ArrayList<>();
         for (DataForInference dataForInference : dataForInferences) {
             if (dataForInference.getIngestDocumentWrapper().getException() != null
@@ -250,7 +296,7 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
         return inferenceTexts;
     }
 
-    private List<DataForInference> getDataForInference(List<IngestDocumentWrapper> ingestDocumentWrappers) {
+    protected List<DataForInference> getDataForInference(List<IngestDocumentWrapper> ingestDocumentWrappers) {
         List<DataForInference> dataForInferences = new ArrayList<>();
         for (IngestDocumentWrapper ingestDocumentWrapper : ingestDocumentWrappers) {
             Map<String, Object> processMap = null;
@@ -272,7 +318,7 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
 
     @Getter
     @AllArgsConstructor
-    private static class DataForInference {
+    protected static class DataForInference {
         private final IngestDocumentWrapper ingestDocumentWrapper;
         private final Map<String, Object> processMap;
         private final List<String> inferenceList;
@@ -415,6 +461,36 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
         nlpResult.forEach(ingestDocument::setFieldValue);
     }
 
+    /**
+     * This method creates a MultiGetRequest from a list of ingest documents to be fetched for comparison
+     * @param ingestDocumentWrappers, list of ingest documents
+     * */
+    protected MultiGetRequest buildMultiGetRequest(List<IngestDocumentWrapper> ingestDocumentWrappers) {
+        MultiGetRequest multiGetRequest = new MultiGetRequest();
+        for (IngestDocumentWrapper ingestDocumentWrapper : ingestDocumentWrappers) {
+            Object index = ingestDocumentWrapper.getIngestDocument().getSourceAndMetadata().get(INDEX_FIELD);
+            Object id = ingestDocumentWrapper.getIngestDocument().getSourceAndMetadata().get(ID_FIELD);
+            if (Objects.nonNull(index) && Objects.nonNull(id)) {
+                multiGetRequest.add(index.toString(), id.toString());
+            }
+        }
+        return multiGetRequest;
+    }
+
+    /**
+     * This method creates a map of documents from MultiGetItemResponse where the key is document ID and value is corresponding document
+     * @param multiGetItemResponses, array of responses from Multi Get Request
+     * */
+    protected Map<String, Map<String, Object>> createDocumentMap(MultiGetItemResponse[] multiGetItemResponses) {
+        Map<String, Map<String, Object>> existingDocuments = new HashMap<>();
+        for (MultiGetItemResponse item : multiGetItemResponses) {
+            String id = item.getId();
+            Map<String, Object> existingDocument = item.getResponse().getSourceAsMap();
+            existingDocuments.put(id, existingDocument);
+        }
+        return existingDocuments;
+    }
+
     @SuppressWarnings({ "unchecked" })
     @VisibleForTesting
     Map<String, Object> buildNLPResult(Map<String, Object> processorMap, List<?> results, Map<String, Object> sourceAndMetadataMap) {
@@ -504,6 +580,13 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
         }
     }
 
+    // This method updates each ingestDocument with exceptions
+    protected void updateWithExceptions(List<IngestDocumentWrapper> ingestDocumentWrappers, Exception e) {
+        for (IngestDocumentWrapper ingestDocumentWrapper : ingestDocumentWrappers) {
+            ingestDocumentWrapper.update(ingestDocumentWrapper.getIngestDocument(), e);
+        }
+    }
+
     private void processMapEntryValue(
         List<?> results,
         IndexWrapper indexWrapper,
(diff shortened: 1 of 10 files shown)


src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -214,6 +263,19 @@ public void subBatchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers,
...
     }
 
+    @Override
+    public void batchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers, Consumer<List<IngestDocumentWrapper>> handler) {
+        super.batchExecute(ingestDocumentWrappers, handler);
+    }
+
+    protected List<List<IngestDocumentWrapper>> cutBatches(List<IngestDocumentWrapper> ingestDocumentWrappers) {
+        List<List<IngestDocumentWrapper>> batches = new ArrayList();
+        for (int i = 0; i < ingestDocumentWrappers.size(); i += this.batchSize) {
```
heemin32, line 273: This document mention that `batch_size` parameter is deprecated. Should we keep supporting it for 3.0?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -214,6 +263,19 @@ public void subBatchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers,
...
+            batches.add(ingestDocumentWrappers.subList(i, Math.min(i + this.batchSize, ingestDocumentWrappers.size())));
```
heemin32, line 274: Using `subList` may not be efficient for certain data structures. For instance, with a `LinkedList`, if `batchSize` is 1, making `n` calls to `subList` would result in an O(n²) time complexity.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -214,6 +263,19 @@ public void subBatchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers,
         });
     }
 
+    @Override
+    public void batchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers, Consumer<List<IngestDocumentWrapper>> handler) {
+        super.batchExecute(ingestDocumentWrappers, handler);
+    }
+
+    protected List<List<IngestDocumentWrapper>> cutBatches(List<IngestDocumentWrapper> ingestDocumentWrappers) {
+        List<List<IngestDocumentWrapper>> batches = new ArrayList();
```
heemin32, line 272: If we know the size of this list upfront, better to specify it.
```suggestion
        List<List<IngestDocumentWrapper>> batches = new ArrayList((ingestDocumentWrappers.size() + this.batchSize) / this.batchSize - 1);
```

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -106,4 +111,44 @@ public void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler
...
+        // if skipExisting flag is turned on, inference texts in each document will be compared with existing documents fetched via
+        // MultiGet. TextEmbeddingInferenceFilter will be used to filter each inference texts
+        openSearchClient.execute(MultiGetAction.INSTANCE, buildMultiGetRequest(ingestDocumentWrappers), ActionListener.wrap(response -> {
+            MultiGetItemResponse[] multiGetItemResponses = response.getResponses();
+            if (multiGetItemResponses == null || multiGetItemResponses.length == 0) {
+                super.batchExecute(ingestDocumentWrappers, handler);
+                return;
+            }
+            Map<String, Map<String, Object>> existingDocuments = createDocumentMap(multiGetItemResponses);
+            if (this.batchSize >= ingestDocumentWrappers.size()) {
```
heemin32, line 135: I don't think we need this line of code. This case will be handled by the rest of the code anyway.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -106,4 +111,44 @@ public void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler
...
+                return;
+            }
+            List<List<IngestDocumentWrapper>> batches = cutBatches(ingestDocumentWrappers);
+            int size = ingestDocumentWrappers.size();
+            AtomicInteger counter = new AtomicInteger(size);
+            List<IngestDocumentWrapper> allResults = Collections.synchronizedList(new ArrayList());
+            for (List<IngestDocumentWrapper> batch : batches) {
+                this.subBatchExecuteWithFilter(batch, existingDocuments, textEmbeddingInferenceFilter, (batchResults) -> {
+                    allResults.addAll(batchResults);
+                    if (counter.addAndGet(-batchResults.size()) == 0) {
```
heemin32, line 146: Can we just compare the size of allResult without using the counter?
```suggestion
                    if (allResults.size == ingestDocumentWrappers.size()) {
```

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -106,4 +111,44 @@ public void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler
...
+                        handler.accept(allResults);
+                    }
+                    assert counter.get() >= 0 : "counter is negative";
```
heemin32, line 149: This assert won't help much. If counter is below zero, the bug should be caught by unit test or integ test.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -106,4 +111,44 @@ public void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler
...
+            }
+            List<List<IngestDocumentWrapper>> batches = cutBatches(ingestDocumentWrappers);
+            int size = ingestDocumentWrappers.size();
+            AtomicInteger counter = new AtomicInteger(size);
+            List<IngestDocumentWrapper> allResults = Collections.synchronizedList(new ArrayList());
+            for (List<IngestDocumentWrapper> batch : batches) {
+                this.subBatchExecuteWithFilter(batch, existingDocuments, textEmbeddingInferenceFilter, (batchResults) -> {
+                    allResults.addAll(batchResults);
+                    if (counter.addAndGet(-batchResults.size()) == 0) {
+                        handler.accept(allResults);
```
heemin32, line 147: My concern is the case when this handler.accept is not being called because some of thread does not call handler due to some error. Then, we should call exception somewhere here.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -106,4 +111,44 @@ public void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler
...
+                    }
+                    assert counter.get() >= 0 : "counter is negative";
+                });
+            }
+        }, e -> { handler.accept(null); }));
```
heemin32, line 152: If we call handler.accept(null), will the cx see the error?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
will-hwang: wanted to keep the changes minimal to the existing implementation of batchExecute in core here: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/ingest/AbstractBatchingProcessor.java#L48-L74

Do you think we should completely rewrite the function?

will-hwang: since there are conflicting explanation for `batch_size`, i think it's safe to leave it for now

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
will-hwang: this is used as a default behavior for batchExecute (https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/ingest/AbstractBatchingProcessor.java#L48-L74), which we extend in our inferenceProcessor. 
I think it's better to leave the existing logic as is, and make changes to what is needed only. let me know your thoughts.

heemin32: Got it. Can't we just override `subBatchExecute` instead of repeating the core code here?

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -106,4 +111,44 @@ public void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler
...
+            super.batchExecute(ingestDocumentWrappers, handler);
+            return;
+        }
+        if (ingestDocumentWrappers.isEmpty()) {
+            handler.accept(Collections.emptyList());
+            return;
+        }
+        // if skipExisting flag is turned on, inference texts in each document will be compared with existing documents fetched via
+        // MultiGet. TextEmbeddingInferenceFilter will be used to filter each inference texts
+        openSearchClient.execute(MultiGetAction.INSTANCE, buildMultiGetRequest(ingestDocumentWrappers), ActionListener.wrap(response -> {
```
heemin32, line 128: I believe `mget` should be performed with a `batch_size`, allowing customers to adjust it if the request fails due to requesting too many documents at once.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
will-hwang: if we just override `subBatchExecute`, we won't be able to use `multiGet` operation to fetch all documents at once, since subBatchExecute only contains a subset (mostly just one) of the documents.

heemin32: That is what we want. We should not get all documents at once. It might fail if the number of document is too big.

will-hwang: okay, with this change we would be calling multiple `multiGet` operation, since we are calling it at subBatch level. 

For example, If user ingests 6 documents, with a batch size of 2,

there will be 3 multiGet calls with each batch size of 2.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
will-hwang: understood, thanks

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
heemin32: Yes.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
will-hwang: no, will update the PR to include exception in each document

will-hwang: with the latest change, when `mget` throws an exception, handler will still accept the ingested document, with exception message populated in each document

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -106,4 +112,60 @@ public void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler
             ActionListener.wrap(handler::accept, onException)
         );
     }
+
+    @Override
+    public void subBatchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers, Consumer<List<IngestDocumentWrapper>> handler) {
```
vibrantvarun, line 117: Could you divide this method into three parts? 
1. Validations
2. MGet call
3. Checking for embeddings

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -182,6 +186,15 @@ public void subBatchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers,
             handler.accept(ingestDocumentWrappers);
             return;
         }
+        doSubBatchExecute(ingestDocumentWrappers, inferenceList, dataForInferences, handler);
+    }
+
+    protected void doSubBatchExecute(
```
vibrantvarun, line 192: Add javadoc to all the methods in this class.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
vibrantvarun: +1 to @heemin32

src/main/java/org/opensearch/neuralsearch/processor/optimization/TextEmbeddingInferenceFilter.java
```diff
@@ -47,7 +45,7 @@ public Object filterInferenceValue(
         Optional<Object> existingValueOptional = ProcessorUtils.getValueFromSource(existingSourceAndMetadataMap, textPath);
         Optional<Object> embeddingValueOptional = ProcessorUtils.getValueFromSource(existingSourceAndMetadataMap, embeddingKey);
         if (existingValueOptional.isPresent() && embeddingValueOptional.isPresent()) {
-            return copyEmbeddingForSingleValue(
```
vibrantvarun, line 50: QQ: why we change this method here?

will-hwang: this was a change from previous PR. Since we're comparing single value and list values as object values, we can have just one method for comparison and copying embeddings

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
will-hwang: the latest change includes this change as mget is performed under `subBatchExecute` method, which has a size of `batch_size`

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
will-hwang: i refactored the method and added comments to make it more readable. Left the validations at the beginning of the method for cleaner implementation

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -415,6 +440,36 @@ protected void setVectorFieldsToDocument(IngestDocument ingestDocument, Map<Stri
         nlpResult.forEach(ingestDocument::setFieldValue);
     }
 
+    /**
+     * This method creates a MultiGetRequest from a list of ingest documents to be fetched for comparison
+     * @param ingestDocumentWrappers, list of ingest documents
+     * */
+    protected MultiGetRequest buildMultiGetRequest(List<IngestDocumentWrapper> ingestDocumentWrappers) {
```
heemin32, line 447: Better to move these method out of this class then it will be much easier to write unit test.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -106,4 +112,74 @@ public void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler
...
+        if (CollectionUtils.isEmpty(ingestDocumentWrappers)) {
+            handler.accept(Collections.emptyList());
+            return;
+        }
+        List<DataForInference> dataForInferences = getDataForInference(ingestDocumentWrappers);
```
heemin32, line 122: What happen if this line throw exception? Shouldn't we catch it and set it in handler or will it be handled already?

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -106,4 +112,74 @@ public void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler
...
+        openSearchClient.execute(MultiGetAction.INSTANCE, buildMultiGetRequest(ingestDocumentWrappers), ActionListener.wrap(response -> {
+            MultiGetItemResponse[] multiGetItemResponses = response.getResponses();
+            if (multiGetItemResponses == null || multiGetItemResponses.length == 0) {
+                doSubBatchExecute(ingestDocumentWrappers, inferenceList, dataForInferences, handler);
+                return;
+            }
+            // create a map of documents with key: doc_id and value: doc
+            Map<String, Map<String, Object>> existingDocuments = createDocumentMap(multiGetItemResponses);
+            List<DataForInference> filteredDataForInference = filterDataForInference(dataForInferences, existingDocuments);
+            List<String> filteredInferenceList = constructInferenceTexts(filteredDataForInference);
```
heemin32, line 144: What happen if exception is thrown here. Will that be handled gracefully?

src/main/java/org/opensearch/neuralsearch/processor/optimization/InferenceFilter.java
```diff
@@ -194,13 +174,15 @@ protected List<Object> filterListValue(
...
+            Object processedList = copyEmbedding(
                 embeddingKey,
                 processList,
-                ProcessorUtils.unsafeCastToObjectList(existingListValue.getFirst()),
-                ProcessorUtils.unsafeCastToObjectList(embeddingListOptional.get()),
-                sourceAndMetadataMap
+                existingListValue.getFirst(),
+                embeddingListOptional.get(),
+                sourceAndMetadataMap,
+                -1
```
heemin32, line 183: This is a bit unclear—I’m not sure when to pass -1 versus another value. It would be helpful to have a separate method for clarity and include a detailed Javadoc explaining when to use each method.
```
// Use when xys 
public Object copyEmbeddingXYZ(
        String embeddingKey,
        Object processValue,
        Object existingValue,
        Object embeddingValue,
        Map<String, Object> sourceAndMetadataMap
    ) {
  return copyEmbedding(embeddingKey, processValue, existingValue, embeddingValue, sourceAndMetadataMap, -1);
}
```

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
will-hwang: okay, i can move these methods to [ProcessorUtil](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java) class. Let me know if you have better suggestions.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
will-hwang: it is already handled in the method: [here](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java#L264)

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
will-hwang: it checks for whether exception was thrown and only create inference text for valid data inferences [here](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java#L240-L249)

src/main/java/org/opensearch/neuralsearch/processor/optimization/InferenceFilter.java
will-hwang: okay, will create a separate method

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
heemin32: That's the issue—you need to check every method to ensure exceptions are handled properly, which isn't a future-proof approach. It would be better to handle the exception here as well, ensuring that any exceptions thrown by underlying methods are caught.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
heemin32: Same as above comment.

vibrantvarun (commented):
@will-hwang could you add tests in failure cases of this feature? Like when will the functionality fail and user will not see 200 status from api

vibrantvarun:
Overall I have similar comments like @heemin32 so will wait for once it gets addressed before doing final round of review

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
will-hwang: If we need this method separated from this class without using static, we would need to instantiate or inject a class into `TextEmbeddingProcessor`. Do we prefer this approach over defining the methods here?

heemin32: Yes, that is generally the preferred approach, but it's important to ensure that an instance is not created for every request. This approach will also significantly simplify writing unit tests. If you think that is too much then current approach is fine though.

will-hwang:
> @will-hwang could you add tests in failure cases of this feature? Like when will the functionality fail and user will not see 200 status from api

i've added a failure test case [here](https://github.com/opensearch-project/neural-search/blob/d69a21901810a79d2dbac694cf27065406437a91/src/test/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessorTests.java#L2415-L2425) where `mget` fails with runtime exception. In summary, the error handling is the same as our existing implementation, where if dependent service (opensearch client or mlcommonclient) fails with exception, ingestDocument is updated with exception, instead of failing ingestion.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
will-hwang: If it's fine, will leave the methods in InferenceProcessor since it already has similar helper methods for processing inferences, and will make child class extension simpler.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -169,49 +173,80 @@ void preprocessIngestDocument(IngestDocument ingestDocument) {
...
+        List<String> inferenceList,
+        List<DataForInference> dataForInferences,
+        Consumer<List<IngestDocumentWrapper>> handler
+    ) {
+        try {
+            Tuple<List<String>, Map<Integer, Integer>> sortedResult = sortByLengthAndReturnOriginalOrder(inferenceList);
+            inferenceList = sortedResult.v1();
+            Map<Integer, Integer> originalOrder = sortedResult.v2();
+            doBatchExecute(inferenceList, results -> {
+                int startIndex = 0;
```
heemin32, line 220: Missing try catch here.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -106,4 +112,85 @@ public void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler
...
+                doSubBatchExecute(ingestDocumentWrappers, inferenceList, dataForInferences, handler);
+                return;
+            }
+            // skipExisting flag is turned on, eligible inference texts in dataForInferences will be compared and filtered after embeddings
+            // are copied
+            openSearchClient.execute(
+                MultiGetAction.INSTANCE,
+                buildMultiGetRequest(ingestDocumentWrappers),
+                ActionListener.wrap(response -> {
+                    MultiGetItemResponse[] multiGetItemResponses = response.getResponses();
```
heemin32, line 144: missing try catch here.

heemin32 (approved):
LGTM

junqiu-lei (approved):
LGTM, thanks.

heemin32:
Failures are coming from changes in knn and it is resolved by https://github.com/opensearch-project/neural-search/pull/1233

Issue #803, opened 2024-06-25
[PROPOSAL] Neural Search field type


asfoorial:
Can we mimic this feature in OpenSearch https://www.elastic.co/search-labs/blog/semantic-search-simplified-semantic-text 

I know that a lot has been done recently in OpenSearch projects to make things headache free. I think a neural-search field type in OpenSearch would be an interesting addition. However, it should account for synonyms to avoid any fine-tuning headache.

navneet1v:
@asfoorial from Opensearch side we do this via a combination of ingestion processor and vector field. As there are multiple use-cases for semantic search including multi-model, this would be an interesting field to have.

But is there any specific reason you are looking for the field as compared to what is present currently. My main motive here is to know the advantages of a new field vs what is currently present in opensearch.

asfoorial:
The main reason is simplifying the process and keep the focus on the business. In fact elasticsearch had the same reason when they introduced the field. 

Another reason is alignment of new features across multiple OpenSearch projects. I have noticed over the past number of releases we get new features in ml-commons and kNN. But it takes a while until we see their benefits reflected in neural-search. If they become one component (neural-search field), then that would sort of guarantee that any new feature in ml-common or kNN must be reflected in the neural-search field type before their release.

navneet1v:
> If they become one component (neural-search field), then that would sort of guarantee that any new feature in ml-common or kNN must be reflected in the neural-search field type before their release.

@asfoorial thanks for providing the details. I want to know little bit more on what features added in ML/k-NN doesn't make into Neural. May be there is something missing.

But I really like the idea of having a field which can encapsulate the processor information.

navneet1v:
One place where having the field will be useful is nested fields. I see putting this information in the processor is very painful and not intutive.

navneet1v:
@minalsha please take a look into this and please add your thoughts

navneet1v:
@heemin32 any reason for closing this gh issue?

heemin32:
@navneet1v No. I think it is closed automatically when I added them in NeuralSearch RoadMap.

navneet1v:
One case where I feel this field type will be very useful is in cases of complex nested fields. Currently with TextEmbedding processor it is always feels like we are finding different cases where the processor is not working some GH issues:
1. https://github.com/opensearch-project/neural-search/issues/1042
2. https://github.com/opensearch-project/neural-search/issues/1024
3. #895 
4. #798 
5. #739 
6. #678 
[...]

PR #1225, opened 2025-03-11
Add semantic field mapper.

bzhangam:
### Description
Add semantic field mapper to support the semantic field type.

Also remove the incorrect FF NEURAL_SEARCH_HYBRID_SEARCH_DISABLED.

### Related Issues
Resolves part of the issue [#1226](https://github.com/opensearch-project/neural-search/issues/1226)

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -251,6 +251,7 @@ def knnJarDirectory = "$buildDir/dependencies/opensearch-knn"
 
 dependencies {
     api "org.opensearch:opensearch:${opensearch_version}"
+    implementation group: 'org.opensearch.plugin', name:'mapper-extras-client', version: "${opensearch_version}"
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-job-scheduler', version: "${opensearch_build}"
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-knn', version: "${opensearch_build}"
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-ml-plugin', version: "${opensearch_build}"
--- a/src/main/java/org/opensearch/neuralsearch/constants/SemanticFieldConstants.java
+++ b/src/main/java/org/opensearch/neuralsearch/constants/SemanticFieldConstants.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.constants;
+
+/**
+ * Constants for semantic field
+ */
+public class SemanticFieldConstants {
+    /**
+     * Name of the model id parameter. We use this key to define the id of the ML model that we will use for the
+     * semantic field.
+     */
+    public static final String MODEL_ID = "model_id";
+
+    /**
+     * Name of the search model id parameter. We use this key to define the id of the ML model that we will use to
+     * inference the query text during the search. If this parameter is not defined we will use the model_id instead.
+     */
+    public static final String SEARCH_MODEL_ID = "search_model_id";
+
+    /**
+     * Name of the raw field type parameter. We use this key to define the field type for the raw data. It will control
+     * how to store and query the raw data.
+     */
+    public static final String RAW_FIELD_TYPE = "raw_field_type";
+
+    /**
+     * Name of the raw field type parameter. We use this key to define a custom field name for the semantic info.
+     */
+    public static final String SEMANTIC_INFO_FIELD_NAME = "semantic_info_field_name";
+}
--- a/src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
+++ b/src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
@@ -1,0 +1,276 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.mapper;
+
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.Setter;
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.index.mapper.BinaryFieldMapper;
+import org.opensearch.index.mapper.FilterFieldType;
+import org.opensearch.index.mapper.KeywordFieldMapper;
+import org.opensearch.index.mapper.MappedFieldType;
+import org.opensearch.index.mapper.Mapper;
+import org.opensearch.index.mapper.MapperParsingException;
+import org.opensearch.index.mapper.MatchOnlyTextFieldMapper;
+import org.opensearch.index.mapper.ParametrizedFieldMapper;
+import org.opensearch.index.mapper.ParseContext;
+import org.opensearch.index.mapper.TextFieldMapper;
+import org.opensearch.index.mapper.TokenCountFieldMapper;
+import org.opensearch.index.mapper.WildcardFieldMapper;
+import org.opensearch.neuralsearch.constants.MappingConstants;
+import org.opensearch.neuralsearch.mapper.dto.SemanticParameters;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+
+import static org.opensearch.neuralsearch.constants.SemanticFieldConstants.MODEL_ID;
+import static org.opensearch.neuralsearch.constants.SemanticFieldConstants.RAW_FIELD_TYPE;
+import static org.opensearch.neuralsearch.constants.SemanticFieldConstants.SEARCH_MODEL_ID;
+import static org.opensearch.neuralsearch.constants.SemanticFieldConstants.SEMANTIC_INFO_FIELD_NAME;
+
+/**
+ * FieldMapper for the semantic field. It will hold a delegate field mapper to delegate the data parsing and query work
+ * based on the raw_field_type.
+ */
+public class SemanticFieldMapper extends ParametrizedFieldMapper {
+    public static final String CONTENT_TYPE = "semantic";
+    private final SemanticParameters semanticParameters;
+
+    @Setter
+    @Getter
+    private ParametrizedFieldMapper delegateFieldMapper;
+
+    protected SemanticFieldMapper(
+        String simpleName,
+        MappedFieldType mappedFieldType,
+        MultiFields multiFields,
+        CopyTo copyTo,
+        ParametrizedFieldMapper delegateFieldMapper,
+        SemanticParameters semanticParameters
+    ) {
+        super(simpleName, mappedFieldType, multiFields, copyTo);
+        this.delegateFieldMapper = delegateFieldMapper;
+        this.semanticParameters = semanticParameters;
+    }
+
+    @Override
+    public Builder getMergeBuilder() {
+        Builder semanticFieldMapperBuilder = (Builder) new Builder(simpleName()).init(this);
+        ParametrizedFieldMapper.Builder delegateBuilder = delegateFieldMapper.getMergeBuilder();
+        semanticFieldMapperBuilder.setDelegateBuilder(delegateBuilder);
+        return semanticFieldMapperBuilder;
+    }
+
+    @Override
+    public final ParametrizedFieldMapper merge(Mapper mergeWith) {
+        if (mergeWith instanceof SemanticFieldMapper) {
+            try {
+                delegateFieldMapper = delegateFieldMapper.merge(((SemanticFieldMapper) mergeWith).delegateFieldMapper);
+            } catch (IllegalArgumentException e) {
+                final String err = String.format(
+                    Locale.ROOT,
+                    "Failed to update the mapper %s because failed to update the delegate mapper for the raw_field_type %s due to %s",
+                    this.name(),
+                    this.semanticParameters.getRawFieldType(),
+                    e.getMessage()
+                );
+                throw new IllegalArgumentException(err, e);
+            }
+        }
+        return super.merge(mergeWith);
+    }
+
+    @Override
+    protected void parseCreateField(ParseContext context) throws IOException {
+        delegateFieldMapper.parse(context);
+    }
+
+    @Override
+    protected String contentType() {
+        return CONTENT_TYPE;
+    }
+
+    public static class Builder extends ParametrizedFieldMapper.Builder {
+        @Getter
+        protected final Parameter<String> modelId = Parameter.stringParam(
+            MODEL_ID,
+            true,
+            m -> ((SemanticFieldMapper) m).semanticParameters.getModelId(),
+            null
+        );
+        @Getter
+        protected final Parameter<String> searchModelId = Parameter.stringParam(
+            SEARCH_MODEL_ID,
+            true,
+            m -> ((SemanticFieldMapper) m).semanticParameters.getSearchModelId(),
+            null
+        );
+        @Getter
+        protected final Parameter<String> rawFieldType = Parameter.stringParam(
+            RAW_FIELD_TYPE,
+            false,
+            m -> ((SemanticFieldMapper) m).semanticParameters.getRawFieldType(),
+            TextFieldMapper.CONTENT_TYPE
+        );
+        @Getter
+        protected final Parameter<String> semanticInfoFieldName = Parameter.stringParam(
+            SEMANTIC_INFO_FIELD_NAME,
+            false,
+            m -> ((SemanticFieldMapper) m).semanticParameters.getSemanticInfoFieldName(),
+            null
+        );
+
+        @Setter
+        protected ParametrizedFieldMapper.Builder delegateBuilder;
+
+        protected Builder(String name) {
+            super(name);
+        }
+
+        @Override
+        protected List<Parameter<?>> getParameters() {
+            return List.of(modelId, searchModelId, rawFieldType, semanticInfoFieldName);
+        }
+
+        @Override
+        public SemanticFieldMapper build(BuilderContext context) {
+            final ParametrizedFieldMapper delegateMapper = delegateBuilder.build(context);
+
+            final SemanticParameters semanticParameters = this.getSemanticParameters();
+            final MappedFieldType semanticFieldType = new SemanticFieldType(delegateMapper.fieldType(), semanticParameters);
+
+            return new SemanticFieldMapper(
+                name,
+                semanticFieldType,
+                multiFieldsBuilder.build(this, context),
+                copyTo.build(),
+                delegateMapper,
+                semanticParameters
+            );
+        }
+
+        public SemanticParameters getSemanticParameters() {
+            return new SemanticParameters(
+                modelId.getValue(),
+                searchModelId.getValue(),
+                rawFieldType.getValue(),
+                semanticInfoFieldName.getValue()
+            );
+        }
+    }
+
+    public static class TypeParser implements Mapper.TypeParser {
+
+        private final static Set<String> SUPPORTED_RAW_FIELD_TYPE = Set.of(
+            TextFieldMapper.CONTENT_TYPE,
+            KeywordFieldMapper.CONTENT_TYPE,
+            MatchOnlyTextFieldMapper.CONTENT_TYPE,
+            WildcardFieldMapper.CONTENT_TYPE,
+            TokenCountFieldMapper.CONTENT_TYPE,
+            BinaryFieldMapper.CONTENT_TYPE
+        );
+
+        @Override
+        public Builder parse(String name, Map<String, Object> node, ParserContext parserContext) throws MapperParsingException {
+            final String rawFieldType = (String) node.getOrDefault(RAW_FIELD_TYPE, TextFieldMapper.CONTENT_TYPE);
+
+            validateRawFieldType(rawFieldType);
+
+            final ParametrizedFieldMapper.TypeParser typeParser = (ParametrizedFieldMapper.TypeParser) parserContext.typeParser(
+                rawFieldType
+            );
+            final Builder semanticFieldMapperBuilder = new Builder(name);
+
+            // semantic field mapper builder parse semantic fields
+            Map<String, Object> semanticConfig = extractSemanticConfig(node, semanticFieldMapperBuilder.getParameters(), rawFieldType);
+            semanticFieldMapperBuilder.parse(name, parserContext, semanticConfig);
+
+            // delegate field mapper builder parse remaining fields
+            ParametrizedFieldMapper.Builder delegateBuilder = typeParser.parse(name, node, parserContext);
+            semanticFieldMapperBuilder.setDelegateBuilder(delegateBuilder);
+
+            return semanticFieldMapperBuilder;
+        }
+
+        private void validateRawFieldType(final String rawFieldType) {
+            if (rawFieldType == null || !SUPPORTED_RAW_FIELD_TYPE.contains(rawFieldType)) {
+                final String err = String.format(
+                    Locale.ROOT,
+                    "raw_field_type %s is not supported. It should be one of [%s]",
+                    rawFieldType,
+                    String.join(", ", SUPPORTED_RAW_FIELD_TYPE)
+                );
+                throw new IllegalArgumentException(err);
+            }
+        }
+
+        /**
+         * In this function we will extract all the parameters defined in the semantic field mapper builder and parse it
+         * later. The remaining parameters will be processed by the type parser of the raw field type. Here we cannot
+         * pass the parameters defined by semantic field to the delegate type parser of the raw field type because it
+         * cannot recognize them.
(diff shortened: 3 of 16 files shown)


build.gradle
```diff
@@ -250,6 +250,7 @@ def knnJarDirectory = "$buildDir/dependencies/opensearch-knn"
 
 dependencies {
     api "org.opensearch:opensearch:${opensearch_version}"
+    implementation group: 'org.opensearch.plugin', name:'mapper-extras-client', version: "${opensearch_version}"
```
bzhangam, line 253: This is needed to access TokenCountFieldMapper to delegate work for token_count raw_field_type.

src/main/java/org/opensearch/neuralsearch/util/FeatureFlagUtil.java
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.util;
+
+import java.util.Map;
+
+public class FeatureFlagUtil {
```
bzhangam, line 9: We cannot use a feature flag like [NEURAL_SEARCH_HYBRID_SEARCH_DISABLED](https://github.com/opensearch-project/neural-search/blob/5f25d6c5e4f64e05b6c17a2eb2ec9e1f467f1974/src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java#L23). It's actually used in a wrong way [here](https://github.com/opensearch-project/neural-search/blob/5f25d6c5e4f64e05b6c17a2eb2ec9e1f467f1974/src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java#L145). I create an [issue](https://github.com/opensearch-project/neural-search/issues/1226) for it.

q-andy: What's the distinction between feature flags and cluster settings? One is configurable at runtime and the other isn't?

src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
```diff
@@ -0,0 +1,246 @@
...
+                if (config != null) {
+                    semanticConfig.put(parameter.name, config);
+                    node.remove(parameter.name);
+                }
+            }
+            semanticConfig.put(FieldConstants.TYPE, SemanticFieldMapper.CONTENT_TYPE);
+            node.put(FieldConstants.TYPE, rawFieldType);
+            return semanticConfig;
+        }
+    }
```
junqiu-lei, line 222: I noticed we're extracting semantic config parameters and removing them from the original node before passing it to the delegate field mapper. This approach could potentially cause validation issues in the delegate mapper, which might expect to see all original parameters during its validation phase.

To ensure more robust behavior, consider:
1. Using a copy of the node instead of modifying the original: `Map<String, Object> nodeCopy = new HashMap<>(node);`
2. Or creating a separate delegate node that preserves all original parameters while only changing the field type

This would prevent any unexpected side effects from parameter removal and ensure that both semantic and delegate mappers can perform complete validation with their expected parameter sets.

src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
```diff
@@ -0,0 +1,246 @@
...
+import static org.opensearch.neuralsearch.constants.SemanticFieldConstants.MODEL_ID;
+import static org.opensearch.neuralsearch.constants.SemanticFieldConstants.RAW_FIELD_TYPE;
+import static org.opensearch.neuralsearch.constants.SemanticFieldConstants.SEARCH_MODEL_ID;
+import static org.opensearch.neuralsearch.constants.SemanticFieldConstants.SEMANTIC_INFO_FIELD_NAME;
+
+/**
+ * FieldMapper for the semantic field. It will hold a delegate field mapper to delegate the data parsing and query work
+ * based on the raw_field_type.
+ */
+public class SemanticFieldMapper extends ParametrizedFieldMapper {
```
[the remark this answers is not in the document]
bzhangam: Yeah. I think it's a typo in core and change it will need all the plugin using it to change accordingly. So don't think we want to make it part of this PR. Probably we can create an issue in core for this.

src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
```diff
@@ -0,0 +1,246 @@
...
+    protected void parseCreateField(ParseContext context) throws IOException {
+        delegateFieldMapper.parse(context);
+    }
+
+    @Override
+    protected String contentType() {
+        return CONTENT_TYPE;
+    }
+
+    public static class Builder extends ParametrizedFieldMapper.Builder {
```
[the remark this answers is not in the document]
bzhangam: I think so. In this way we can access it like SemanticFieldMapper.Builder. This is a common pattern used across OpenSearch.

src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
```diff
@@ -0,0 +1,246 @@
...
+                semanticFieldType,
+                multiFieldsBuilder.build(this, context),
+                copyTo.build(),
+                delegateMapper,
+                this
+            );
+        }
+    }
+
+    public static class TypeParser implements Mapper.TypeParser {
```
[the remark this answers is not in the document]
bzhangam: Yes we can. But I think this is a common pattern used across OpenSearch.

src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
bzhangam: The delegate field mapper cannot process the semantic parameters so we have to extract them. The semantic field mapper should validate the semantic parameters and the delegate field mapper only need to process the non-semantic parameters. If we pass all original parameters to the delegate field mapper it will always fail since it cannot handle the semantic parameters.

src/main/java/org/opensearch/neuralsearch/mapper/semanticFieldTypes/SemanticFieldTypeParameters.java
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.mapper.semanticFieldTypes;
+
+import lombok.Getter;
+import org.opensearch.neuralsearch.mapper.SemanticFieldMapper;
+
+@Getter
+public class SemanticFieldTypeParameters {
```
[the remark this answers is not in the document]
bzhangam: I view this as a class to help us hold all the parameters. In this when we need to add new parameters we don't need to change it in all field types. We just need to update it here.

src/main/java/org/opensearch/neuralsearch/mapper/semanticFieldTypes/SemanticFieldTypeFactory.java
```diff
@@ -0,0 +1,52 @@
...
+        KeywordFieldMapper.CONTENT_TYPE,
+        WildcardFieldMapper.CONTENT_TYPE
+    );
+
+    public MappedFieldType createSemanticFieldType(
+        @NonNull final ParametrizedFieldMapper delegateMapper,
+        @NonNull final String rawFieldType,
+        SemanticFieldMapper.Builder builder
+    ) {
+        if (BinaryFieldMapper.CONTENT_TYPE.equals(rawFieldType)) {
```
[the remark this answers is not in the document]
bzhangam: I think with switch/case we cannot do STRING_FIELD_TYPES.contains(rawFieldType). So would like to keep this.

src/main/java/org/opensearch/neuralsearch/util/FeatureFlagUtil.java
bzhangam: You can view the settings in the feature flag is a private static variable which is only used to hold some feature flags while the cluster settings is shared across the cluster. And it can be any setting related to the cluster.

src/main/java/org/opensearch/neuralsearch/mapper/semanticFieldTypes/SemanticFieldTypeParameters.java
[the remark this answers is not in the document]
will-hwang: makes sense, if this class is used only for storing fields, maybe @Data annotation can be used instead? Also, I suggest using better name.

e.g SemanticFieldTypeAttributes, SemanticFieldTypeConfig, SemanticFieldTypeMetadata etc

src/main/java/org/opensearch/neuralsearch/constants/SemanticFieldConstants.java
```diff
@@ -0,0 +1,12 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.constants;
+
+public class SemanticFieldConstants {
+    public static final String MODEL_ID = "model_id";
+    public static final String SEARCH_MODEL_ID = "search_model_id";
+    public static final String RAW_FIELD_TYPE = "raw_field_type";
+    public static final String SEMANTIC_INFO_FIELD_NAME = "semantic_info_field_name";
+}
```
q-andy, line 12: Should we add javadocs for new public methods/fields? `./gradlew :javadoc` complains when public classes, fields, and methods are missing them

bzhangam: I think we can add.

src/main/java/org/opensearch/neuralsearch/mapper/semanticFieldTypes/SemanticFieldTypeParameters.java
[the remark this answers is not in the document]
bzhangam: I think it's better to use a custom constructor so that we can simply pass a builder to build this object.

For naming I don't see why parameters is worse than attributes/config/metadata. I use parameter because in field mapper we call them as parameters.

src/test/java/org/opensearch/neuralsearch/mapper/semanticFieldTypes/SemanticStringFieldTypeTests.java
```diff
@@ -0,0 +1,138 @@
...
+
+import java.io.IOException;
+
+import static org.opensearch.neuralsearch.mapper.SemanticFieldMapperTestUtil.buildSemanticFieldMapperWithTextAsRawFieldType;
+
+public class SemanticStringFieldTypeTests extends OpenSearchTestCase {
+    @Mock
+    Mapper.TypeParser.ParserContext parserContext;
+    @Mock
+    protected StringFieldType mockDelegateFieldType;
```
yizheliu-amazon, line 24: better to keep the naming and access level consistent. Maybe `private Mapper.TypeParser.ParserContext mockParserContext`? Can we make `mockDelegateFieldType ` private as well?

src/main/java/org/opensearch/neuralsearch/mapper/semanticFieldTypes/SemanticNumberFieldType.java
```diff
@@ -0,0 +1,47 @@
...
+            delegateFieldType.nullValue(),
+            delegateFieldType.meta()
+        );
+        this.delegateFieldType = delegateFieldType;
+        this.semanticFieldTypeParameters = new SemanticFieldTypeParameters(builder);
+    }
+
+    @Override
+    public String typeName() {
+        return SemanticFieldMapper.CONTENT_TYPE;
```
yizheliu-amazon, line 40: I can see other Semantic filed type classes share the same typeName value, aka `SemanticFieldMapper.CONTENT_TYPE`. 

Where is the `typeName ` being used? Is it okay to let 3 classes to share the same value?

src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
```diff
@@ -0,0 +1,246 @@
...
+
+        private final static List<String> SUPPORTED_RAW_FIELD_TYPE = List.of(
```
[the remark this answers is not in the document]
bzhangam: Yeah. We can use set.

src/main/java/org/opensearch/neuralsearch/mapper/semanticFieldTypes/SemanticFieldType.java
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.mapper.semanticFieldTypes;
```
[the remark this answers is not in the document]
bzhangam: yeah I can rename it

src/main/java/org/opensearch/neuralsearch/mapper/semanticFieldTypes/SemanticFieldTypeFactory.java
```diff
@@ -0,0 +1,52 @@
...
+import org.opensearch.index.mapper.TextFieldMapper;
+import org.opensearch.index.mapper.TokenCountFieldMapper;
+import org.opensearch.index.mapper.WildcardFieldMapper;
+import org.opensearch.neuralsearch.mapper.SemanticFieldMapper;
+
+import java.util.List;
+
+@NoArgsConstructor
+public class SemanticFieldTypeFactory {
+    private static final List<String> STRING_FIELD_TYPES = List.of(
```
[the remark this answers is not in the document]
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mapper/semanticFieldTypes/SemanticNumberFieldType.java
bzhangam: It's used in error message. I think it's ok you can view them as three different implementation for the semantic field type.

src/test/java/org/opensearch/neuralsearch/mapper/semanticFieldTypes/SemanticStringFieldTypeTests.java
bzhangam: yeah I think we can.

src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
junqiu-lei: Then this approach makes sense. Have you tested edge cases where parameter validation might be affected by this separation?

Perhaps adding a comment in the code explaining this design decision would help other maintainers understand why parameters are extracted rather than copied.

bzhangam: Ok I can add more comment to make it clear. The validation should not be affected since it works like two fields are defined. One is the semantic field with only semantic parameters and another is the raw field with non-semantic parameters. And each field's field mapper should handle it properly.

src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
```diff
@@ -0,0 +1,257 @@
...
+    @Getter
+    private ParametrizedFieldMapper delegateFieldMapper;
+
+    protected SemanticFieldMapper(
+        String simpleName,
+        MappedFieldType mappedFieldType,
+        MultiFields multiFields,
+        CopyTo copyTo,
+        ParametrizedFieldMapper delegateFieldMapper,
+        SemanticFieldMapper.Builder builder
```
heemin32, line 56: This code pattern seems uncommon to me. Would it be better to directly pass modelId, rawFieldType, semanticInfoFieldName, and searchModelId instead?

The SemanticFieldMapper doesn't have to know about builder.

src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
```diff
@@ -0,0 +1,257 @@
...
+            null
+        );
+
+        @Setter
+        protected ParametrizedFieldMapper.Builder delegateBuilder;
+        private final SemanticFieldTypeFactory semanticFieldTypeFactory;
+
+        protected Builder(String name) {
+            super(name);
+            semanticFieldTypeFactory = new SemanticFieldTypeFactory();
```
heemin32, line 134: It's better to inject it from outside, and even better if we can make it reusable.

```
protected Builder(String name, SemanticFieldTypeFactory semanticFieldTypeFactory) {
   ...
}
```

src/main/java/org/opensearch/neuralsearch/mapper/semanticfieldtypes/SemanticFieldTypeParameters.java
```diff
@@ -0,0 +1,26 @@
...
+ * An object to hold all the parameters we can have in a semantic field type.
+ */
+@Getter
+public class SemanticFieldTypeParameters {
+    private final String modelId;
+    private final String searchModelId;
+    private final String semanticInfoFieldName;
+    private final String rawFieldType;
+
+    public SemanticFieldTypeParameters(SemanticFieldMapper.Builder builder) {
```
heemin32, line 20: I think it is better to receive each variable directly instead of relying on Builder.
This create unnecessary dependency between SemanticFieldTypeParameters and SemanticFieldMapper.Builder.

src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
bzhangam: I saw something similar to this in [KnnVectorFieldMapper](https://github.com/opensearch-project/k-NN/blob/2e56790811ab56442ceb244e4a618ed49c6f4193/src/main/java/org/opensearch/knn/index/mapper/KNNVectorFieldMapper.java#L636). The main benefit of this is that when we need to add a new parameter we don't need to modify the signature of this constructor. And I also think it's cleaner to do it this way.

bzhangam: To decouple it from the builder I think I can only pass the semantic parameters.

heemin32: We better have DTO if we want to reduce the number of parameter passing to field mapper. Passing a builder instance to a constructor is not ideal. Knn code is not following best practice in many places so we should not reference their code patten as a guide imo.

src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mapper/semanticfieldtypes/SemanticFieldTypeParameters.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
bzhangam: Yeah. I plan to only pass the parameters as a DTO.

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
```diff
@@ -214,17 +225,6 @@ public Map<String, Processor.Factory> getProcessors(Processor.Parameters paramet
 
     @Override
     public Optional<QueryPhaseSearcher> getQueryPhaseSearcher() {
-        // we're using "is_disabled" flag as there are no proper implementation of FeatureFlags.isDisabled(). Both
```
bzhangam, line 217: This feature is already enabled so no need to use FF anymore.

src/test/java/org/opensearch/neuralsearch/query/OpenSearchQueryTestCase.java
```diff
@@ -232,11 +231,6 @@ public float getMaxScore(int upTo) {
         };
     }
 
-    @SuppressForbidden(reason = "manipulates system properties for testing")
-    protected static void initFeatureFlags() {
-        System.setProperty(NEURAL_SEARCH_HYBRID_SEARCH_DISABLED.getKey(), "true");
```
bzhangam, line 237: This manipulates is no longer supported by the FeatureFlag in core. Now it will only read the FF from the cluster settings.

src/main/java/org/opensearch/neuralsearch/mapper/dto/SemanticParameters.java
```diff
@@ -0,0 +1,25 @@
...
+@Getter
+public class SemanticParameters {
+    private final String modelId;
+    private final String searchModelId;
+    private final String rawFieldType;
+    private final String semanticInfoFieldName;
+
+    public SemanticParameters(String modelId, String searchModelId, String rawFieldType, String semanticInfoFieldName) {
+        this.modelId = modelId;
+        this.searchModelId = searchModelId;
```
junqiu-lei, line 21: Could you clarify how these are used differently in the implementation, and whether there are any validation checks to ensure compatible models are selected? Also, are there any default values if one is specified but not the other?

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
```diff
@@ -289,4 +288,12 @@ public List<SearchPlugin.SearchExtSpec<?>> getSearchExts() {
     public Map<String, Highlighter> getHighlighters() {
         return Collections.singletonMap(SemanticHighlighter.NAME, semanticHighlighter);
     }
+
+    @Override
+    public Map<String, Mapper.TypeParser> getMappers() {
+        if (FeatureFlagUtil.isEnabled(SEMANTIC_FIELD_ENABLED)) {
+            return Map.of(SemanticFieldMapper.CONTENT_TYPE, new SemanticFieldMapper.TypeParser());
```
junqiu-lei, line 295: What happens if a user defines both a semantic field type and a traditional pipeline for the same field? Is there any conflict detection?

src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
```diff
@@ -0,0 +1,277 @@
...
+        return semanticFieldMapperBuilder;
+    }
+
+    @Override
+    public final ParametrizedFieldMapper merge(Mapper mergeWith) {
+        if (mergeWith instanceof SemanticFieldMapper) {
+            try {
+                delegateFieldMapper = delegateFieldMapper.merge(((SemanticFieldMapper) mergeWith).delegateFieldMapper);
+            } catch (IllegalArgumentException e) {
+                String err = "Failed to update the mapper ["
```
junqiu-lei, line 76: use like 
```
String.format(Locale.ROOT, "inner %s query clause cannot be null", NAME)
```

src/main/java/org/opensearch/neuralsearch/mapper/dto/SemanticParameters.java
bzhangam: You can find the info in SemanticFieldConstants.

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
bzhangam: No. The traditional pipeline should work the same as today's UX. We will execute the system pipeline generated for the semantic field at the end.

src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
bzhangam: ok

junqiu-lei (approved):

src/main/java/org/opensearch/neuralsearch/constants/SemanticFieldConstants.java
```diff
@@ -0,0 +1,33 @@
...
+     * Name of the model id parameter. We use this key to define the id of the ML model that we will use for the
+     * semantic field.
+     */
+    public static final String MODEL_ID = "model_id";
+
+    /**
+     * Name of the search model id parameter. We use this key to define the id of the ML model that we will use to
+     * inference the query text during the search. If this parameter is not defined we will use the model_id instead.
+     */
+    public static final String SEARCH_MODEL_ID = "search_model_id";
```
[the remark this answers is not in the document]
bzhangam: It's needed to support the sparse model use case.

yizheliu-amazon (approved):

PR #1230, opened 2025-03-13
Remove validation on text and image field for text_image_embedding processor

### LINKED ISSUE #1221: [BUG] Remove validation on text/image field for text_image_embedding processor
### What is the bug?
When using `text_image_embedding` processor, user will specify target text field in `field_map`. Currently the processor will reuse the validation before ingesting new doc, in which will check the specified depth and non-empty field. But in `text_image_embedding` processor, it seems ok to remove the validation, as ingesting a doc with empty `text` field is expected.

1. Create a pipeline
```
PUT /_ingest/pipeline/nlp-ingest-pipeline
{
  "description": "A text/image embedding pipeline",
  "processors": [
    {
      "text_image_embedding": {
        "model_id": "bQ1J8ooBpBj3wT4HVUsb",
        "embedding": "vector_embedding",
        "field_map": {
          "text": "image_description",
          "image": "image_binary"
        }
      }
    }
  ]
}
```

2. Ingest a doc
```
POST nlp-index/_doc/1?pipeline=nlp-ingest-pipeline
{
  "text": ""
}
```

Ingestion will fail with following exception:
```
{
  "error": {
    "root_cause": [
      {
        "reason": "map type field [text] has empty string value, cannot process it"
      }
    ],
    "reason": "map type field [text] has empty string value, cannot process it"
  },
  "status": 400
}
```

The doc with empty `text` field should ingest successfully.

_Operating system, version._

weijia-aws:
### Description
This PR removes the text and image field validation for text_image_processor

### Related Issues
Resolves #1221

### Code Bug
Request:
```
POST nlp-index/_doc/1?pipeline=nlp-ingest-pipeline
{
  "text": ""
}
```
Response:
```
{
  "error": {
    "root_cause": [
      {
        "reason": "map type field [text] has empty string value, cannot process it"
      }
    ],
    "reason": "map type field [text] has empty string value, cannot process it"
  },
  "status": 400
}
```

#### Code Bug Deep Dive
When create a text_image_processor, we need to pass a `field_map` that at least contains `text` or `image` field, see doc [here](https://opensearch.org/docs/latest/ingest-pipelines/processors/text-image-embedding/). During document ingestion, the processor will read the actual value from `input_text_field` and `input_image_field`, NOT from `text` or `image`. 

However it's possible that some document contain the `text` or `image` field, as well as `input_text_field` and `input_image_field`. Since the processor reads the actual values from `input_text_field` and `input_image_field` and create embeddings against them,  these `text` or `image` fields should be treated as unmapped fields. We should allow users to ingest whatever value `text` or `image` field has.

Currently there's a [validation](https://github.com/opensearch-project/neural-search/blob/5f25d6c5e4f64e05b6c17a2eb2ec9e1f467f1974/src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java#L78) for creating the processor to ensure the `field_map` is valid.

During ingestion, there's another [validation](https://github.com/opensearch-project/neural-search/blob/5f25d6c5e4f64e05b6c17a2eb2ec9e1f467f1974/src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java#L176) to validate the embedding field values (from `input_text_field` and `input_image_field`). However this method is wrong, it checks the field values with `text` or `image` field, we should remove this method entirely. As for validating the actual embedding field values, it is already covered [here](https://github.com/opensearch-project/neural-search/blob/5f25d6c5e4f64e05b6c17a2eb2ec9e1f467f1974/src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java#L111C42-L111C76) 

### Check List
- [N] New functionality includes testing.
- [N/A] New functionality has been documented.
- [N] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [Y] Commits are signed per the DCO using `--signoff`.
- [N/A] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
@@ -17,7 +17,6 @@ import org.apache.commons.lang3.StringUtils;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.env.Environment;
-import org.opensearch.index.mapper.IndexFieldMapper;
 import org.opensearch.ingest.AbstractProcessor;
 import org.opensearch.ingest.IngestDocument;
 import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
@@ -25,7 +24,6 @@ import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
 import com.google.common.annotations.VisibleForTesting;
 
 import lombok.extern.log4j.Log4j2;
-import org.opensearch.neuralsearch.util.ProcessorDocumentUtils;
 
 /**
  * This processor is used for user input data text and image embedding processing, model_id can be used to indicate which model user use,
@@ -107,7 +105,6 @@ public class TextImageEmbeddingProcessor extends AbstractProcessor {
     @Override
     public void execute(final IngestDocument ingestDocument, final BiConsumer<IngestDocument, Exception> handler) {
         try {
-            validateEmbeddingFieldsValue(ingestDocument);
             Map<String, String> knnMap = buildMapWithKnnKeyAndOriginalValue(ingestDocument);
             Map<String, String> inferenceMap = createInferences(knnMap);
             if (inferenceMap.isEmpty()) {
@@ -171,20 +168,6 @@ public class TextImageEmbeddingProcessor extends AbstractProcessor {
         Map<String, Object> result = new LinkedHashMap<>();
         result.put(knnKey, modelTensorList);
         return result;
-    }
-
-    private void validateEmbeddingFieldsValue(final IngestDocument ingestDocument) {
-        Map<String, Object> sourceAndMetadataMap = ingestDocument.getSourceAndMetadata();
-        String indexName = sourceAndMetadataMap.get(IndexFieldMapper.NAME).toString();
-        ProcessorDocumentUtils.validateMapTypeValue(
-            FIELD_MAP_FIELD,
-            sourceAndMetadataMap,
-            fieldMap,
-            indexName,
-            clusterService,
-            environment,
-            false
-        );
     }
 
     @Override
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Lower bound for min-max normalization technique in hybrid query ([#1195](https://github.com/opensearch-project/neural-search/pull/1195))
 ### Enhancements
 ### Bug Fixes
+- Remove validations for unmapped fields (text and image) in TextImageEmbeddingProcessor ([#1230](https://github.com/opensearch-project/neural-search/pull/1230))
 ### Infrastructure
 ### Documentation
 ### Maintenance
--- a/src/test/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessorIT.java
+++ b/src/test/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessorIT.java
@@ -8,6 +8,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 
 import org.junit.Before;
+import org.opensearch.client.ResponseException;
 import org.opensearch.neuralsearch.BaseNeuralSearchIT;
 
 /**
@@ -21,17 +22,13 @@ public class TextImageEmbeddingProcessorIT extends BaseNeuralSearchIT {
     private static final String INGEST_DOCUMENT = "{\n"
         + "  \"title\": \"This is a good day\",\n"
         + "  \"description\": \"daily logging\",\n"
-        + "  \"passage_text\": \"A very nice day today\",\n"
+        + "  \"passage_text\": \"passage_text_value\",\n"
+        + "  \"text\": \"\",\n"
+        + "  \"image\": null,\n"
         + "  \"favorites\": {\n"
         + "    \"game\": \"overwatch\",\n"
         + "    \"movie\": null\n"
         + "  }\n"
-        + "}\n";
-
-    private static final String INGEST_DOCUMENT_UNMAPPED_FIELDS = "{\n"
-        + "  \"title\": \"This is a good day\",\n"
-        + "  \"description\": \"daily logging\",\n"
-        + "  \"some_random_field\": \"Today is a sunny weather\"\n"
         + "}\n";
 
     @Before
@@ -49,8 +46,19 @@ public class TextImageEmbeddingProcessorIT extends BaseNeuralSearchIT {
         ingestDocument(INDEX_NAME, INGEST_DOCUMENT);
         assertEquals(1, getDocCount(INDEX_NAME));
         // verify doc without mapping
-        ingestDocument(INDEX_NAME, INGEST_DOCUMENT_UNMAPPED_FIELDS);
+        String documentWithUnmappedFields;
+        documentWithUnmappedFields = INGEST_DOCUMENT.replace("passage_text", "random_field_1");
+        ingestDocument(INDEX_NAME, documentWithUnmappedFields);
         assertEquals(2, getDocCount(INDEX_NAME));
+    }
+
+    public void testEmbeddingProcessor_whenIngestingDocumentWithNullMappingValue_thenThrowException() throws Exception {
+        String modelId = uploadModel();
+        loadModel(modelId);
+        createPipelineProcessor(modelId, PIPELINE_NAME, ProcessorType.TEXT_IMAGE_EMBEDDING);
+        createIndexWithPipeline(INDEX_NAME, "IndexMappings.json", PIPELINE_NAME);
+
+        expectThrows(ResponseException.class, () -> ingestDocument(INDEX_NAME, INGEST_DOCUMENT.replace("\"passage_text_value\"", "null")));
     }
 
     private String uploadModel() throws Exception {
--- a/src/test/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessorTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessorTests.java
@@ -185,7 +185,9 @@ public class TextImageEmbeddingProcessorTests extends OpenSearchTestCase {
         sourceAndMetadata.put(IndexFieldMapper.NAME, "my_index");
         sourceAndMetadata.put("key1", "value1");
         sourceAndMetadata.put("my_text_field", "value2");
-        sourceAndMetadata.put("key3", "value3");
+        sourceAndMetadata.put("text", "");
+        sourceAndMetadata.put("image", null);
+        sourceAndMetadata.put("key5", Map.of("inner_field", "innerValue1"));
         sourceAndMetadata.put("image_field", "base64_of_image_1234567890");
         IngestDocument ingestDocument = new IngestDocument(sourceAndMetadata, new HashMap<>());
         TextImageEmbeddingProcessor processor = createInstance();


src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -173,20 +170,6 @@ Map<String, Object> buildTextEmbeddingResult(final String knnKey, List<Number> m
         return result;
     }
 
-    private void validateEmbeddingFieldsValue(final IngestDocument ingestDocument) {
```
will-hwang, line 176: we should not remove this method entirely, but pass in `true` for `allowEmpty` field [here](https://github.com/opensearch-project/neural-search/blob/5f25d6c5e4f64e05b6c17a2eb2ec9e1f467f1974/src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java#L186)

weijia-aws: This [method](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java#L55) simply validates the values in `field_map`, however this `field_map` contains:
```
{
  text: "input_text_field",
  image: "input_image_field",
}
```
It's correct to validate it during processor creation, but not document ingestion. We can accept whatever value `text` or `image` field has in a document, so there's no point to keep the validation method

will-hwang: we have the same validation for text chunking processor [here](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java#L178-L186)

I would propose to enable `allowEmpty` to `true` and avoid removing this method

weijia-aws: The text chunking processor is different, in the processor, the field map is:
```
"field_map": {
      "<input_field>": "<output_field>"
    }
```
where `input_field` is **The name of the field from which to obtain text for generating chunked passages.** [source](https://opensearch.org/docs/latest/ingest-pipelines/processors/text-chunking/), so it's correct to have that  validation method (this is also the only one validation for this processor).

While in the text_image_embedding processor, the field map is 
```
{
  text: "input_text_field",
  image: "input_image_field",
}
```
where `input_text_field` and `input_image_field` is the field to obtain values for embedding

we should validate the `input_text_field` and `input_image_field` instead of `text` and `image` (which is what that validation method does). Since we accept all values for `text` and `image` field, there's no validations needed.

bzhangam: I think we are using this function wrongly here for the TextImageEmbeddingProcessor. But I think we still want to validate:
1. The depth of the map should not exceed the limits.
2. The input_text_field and input_image_field should have string values.

The actual field map we should validate is
{
 "input_text_field": "vector_field",
 "input_image_field": "vector_field"
}

And we should unflatten it before the validation to handle the "." properly.

will-hwang: the validation is present for sparse encoding and text embedding as well [here](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java#L141). I understand that `input_text_field` and `input_image_field` are the fields used for embedding, but this method validates IngestDocument matches the field_map definition, which is a valid check. @zane-neo can you confirm if this method can be removed entirely?

heemin32: My understanding is same as @bzhangam. Not sure we support `.` in text/image embedding processor though.

weijia-aws: For 1, the depth of map will not exceed the limits, as it is always 1, see the map definition: 
```
{
  text: "input_text_field",
  image: "input_image_field",
}
```
Where `input_text_field` and `input_image_field` must have **String** type, so the depth is 1.


For 2, it's already validated somewhere else, see code [here](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java#L160)

weijia-aws: > not sure we support . in text/image embedding processor though.

Can you give some examples for this `.`?

weijia-aws: > do we have plans to support nested field map structure in text/image embedding processor?

I don't think we have, @heemin32 Do you know?

heemin32: > > not sure we support . in text/image embedding processor though.
> 
> Can you give some examples for this `.`?
```
// mapping
{
  text: "abc.def",
  image: "abc.efg",
}
```
```
// document
{
  "abc": {
    "def": "text",
    "efg": "sdfasdf"
  }
}
```

weijia-aws: I see, from the current processor implementation, the `.` operation is not supported

heemin32:
Please update changelog.

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
bzhangam: If "." is not supported then we should validate input_text_field and input_image_field should not contain it. And then we just need to validate if the doc has input_text_field and input_image_field then they should be string value.

I think simply pass the field map like below can utilize that function to validate if value exist it should be a string.
```
{
"input_text_field": "vector_field",
"input_image_field": "vector_field"
}
```

And in processor creation we should validate input_text_field and input_image_field should not contain ".". I think we don't have this validation for now. Probably we want to double check if we should support ".".

weijia-aws: > input_text_field and input_image_field should not contain it

Why do we want to deny it? they will just be treated as part of the name of the input field, the input field name can be anything including special characters.

> we just need to validate if the doc has input_text_field and input_image_field then they should be string value.

No, this is a valid use case where the doc doesn't have `input_text_field` and `input_image_field`, for the data type, they are already validated [here](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java#L160), outside of this validation method

> Probably we want to double check if we should support ".".
I verified it already on local cluster, we don't have logic to parse `.`,  if the field_map is:

```
{
  text: "abc.def",
  image: "abc.efg",
}
```

where `input_text_field` = `abc.def`, `input_image_field` = `abc.efg`. If we want to ingest a document like
```
// document
{
  "abc": {
    "def": "text",
    "efg": "sdfasdf"
  }
}
```
this `abc` object is treated as an unmapped field, no embedding is generated. However, if we ingest this document:
// document
{
  "abc.def": "text",
  "abc.efg": "image_binary",
}
```
embedding will be generated

zane-neo: Agree with @bzhangam , we still need this validation but this method needs modification:
1. In the creation, we're only validating the keys and value not null, but not validating the value's depth so we can't assume the depth is always 1, a malicious user could construct deeply nested configuration to exploit the system resources. So this is an issue and I suggest we fix this in this PR as well.
2. During the ingestion, it's true this [code](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java#L160) is validating but it only checks the type not the value, so user able to ingest empty string for both fields and that input doesn't make sense and we should block this request to model side.

The validation in TextImageEmbeddingProcessor indeed is different with other processors as for other processors the kv are mappings of original keys and target keys, but for TextImageEmbeddingProcessor, the kv are mappings of `reserved keywords` and original keys. So we should change the validation to validate the `target keys(which in fact is the original keys)` instead of the `reserved keywords`.

With above two changes, then I think the validation of TextImageEmbeddingProcessor is complete.

heemin32: @zane-neo The main goal of this PR is to address the reported bug. As long as it doesn’t remove any existing valid validation, we should keep the focus on the bug fix. For any additional validation, we can create a GitHub issue and address it in a separate PR.

weijia-aws: Hi @zane-neo,

For 1, during processor creation, we already validate the field_map must be a Map<String, String>, see code [here](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java#L49). So if there's nested configuration, this processor will fail to create at all

For 2, currently we do allow ingest document with mapped fields with empty strings. But why do we want to block them? We also allow ingest documents without any mapping fields

bzhangam: @weijia-aws 
For your first question:
I think we should make it clear that we don't support "." otherwise we silently ignore the field name with "." is not a good UX. We can create a RFC for it not have to address it in this PR.

For your second question:
My point is to validate if there is a value for the text or image it should be a string value rather than an object or array. I think we will fail to process the doc if we find an object or array on the path of the text and image. But it's better to keep the UX in sync that we do validate here like other processors.

weijia-aws: > My point is to validate if there is a value for the text or image it should be a string value rather than an object or array.

For unmapped field `text` and `image`, why do we want to restrict the data type? the value can be anything including string and any other object.

If you are talking about mapped fields `input_text_field` and `input_image_field`, we do need to validating the data type. But it's already handled [here](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java#L160) in `buildMapWithKnnKeyAndOriginalValue` method

martin-gaievski:
Can you address these two items:
- add test that will trigger that failure scenario and incorrectly fail with the code we do have today, without this PR
- add example of request response for this scenario to PR description, this will ease the understanding of your intension

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
weijia-aws: For now, I will merge as is. We can add the validation in future

weijia-aws:
> add test that will trigger that failure scenario and incorrectly fail with the code we do have today, without this PR

I don't quick follow. Why do we want to add the test, we already removed the validation in this PR. If you meant to validate the scenario through test, then it's already done, see the unit test here https://github.com/opensearch-project/neural-search/pull/1230/files#diff-c429eb1a1fb2495d60ee8f799c59893de5e77ae1e9273204113c57ab828f7a1eR188, if I keep the validation and run this test, the test will fail with the exact error message.

> add example of request response for this scenario to PR description, this will ease the understanding of your intension

Done

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
heemin32: @weijia-aws Could you create a github issue for those two cases above?

heemin32 (approved):

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
weijia-aws: Create one Github issue for the second case https://github.com/opensearch-project/neural-search/issues/1239, there's no issue with the first one

junqiu-lei (approved):

PR #1191, opened 2025-02-19
Optimized text embedding processor for single document update

will-hwang:
### Description
This PR implements optimized text embedding processor for single document update scenario. 

### Proposed State [Single Document Update]

proposed-flow

Steps:

1. Process Map is generated based on IngestDocument and Field Map defined.
2. **If ignore_existing feature is set to true, filter the process map.** 
    1. **Existing document is fetched via OpenSearch client’s Get Action to compare the existing inference text against IngestDocument**
        1. if document does not exist, or any exception is thrown, fallback to calling model inference
    2. **Locate the embedding fields in the existing document**
        1. Recursively traverse the process map to find embedding fields. Keeping track of the traversal path is required for look up in existing document. 
        2. Once embedding fields are found, attempt to copy embeddings from existing document to ingest document. 
    3. **If eligible, copy over the vector embeddings from existing document to ingest document** 
        1. It is eligible for copy if inference text in ingest document and existing document is the same, and embeddings for the inference text exist in the existing document. 
        2. Note, if in case of values in list, the fields in the same index are compared to determine text equality
    4. **If eligible fields have been copied, remove the entry in process map** 
3. Inference List is generated based on entries in Filtered Process Map. 
4. ML Commons InferenceSentence API is invoked with filtered inference list. 
5. Embeddings for filtered inference list are generated in ML Commons. 
6. Embeddings for filtered inference list are mapped to target fields via entries defined in process map.
7. Embeddings for filtered inference list are populated to target fields in IngestDocument

## User Scenarios

**Scenario 1: Document Update with no change to all fields**

```
Update Document
ID: 1
"category": [
    {
        "name": {
            "en": "this is 1st name"
        }
    },
    {
        "name": {
            "en": "this is 2nd name"
        }
    }
]
```

```
Existing Document
ID: 1
"category": [
    {
        "name": {
            "category_name_vector": [
                -0.1,
                -0.1,
                -0.1,
                ...
            ],
            "en": "this is 1st name"
        }
        
    },
    {
        "name": {
            "category_name_vector": [
                -0.2,
               -0.2,
               -0.2,
                ...
            ],
            "en": "this is 2nd name"
        }
    }
]
```

```
Filtered Process Map 
{
  "category":` `
        {
            "name": {
                "category_name_vector": [~~"this is 1st name", "this is 2nd name"~~]
            }
        }
}
```

```
Filtered Inference List
[]
```

```
Ingested Document
"category": [
    {
        "name": {
[...]

PR #1238, opened 2025-03-20
Implement Optimized embedding generation in text embedding processor

will-hwang:
### Description

This PR merges the previously approved PRs (#1191 , #1217 ) from feature branch onto main. 

See below for Benchmark results

## Setup

**Version**: OS 3.0 alpha
**Cluster**: 4 r6g.2xlarge nodes (1 coordinator node 3 data nodes)

## Benchmark Results 

**Dataset:** Trec-covid (https://public.ukp.informatik.tu-darmstadt.de/thakur/BEIR/datasets/trec-covid.zip)
**Ingest Pipeline**
```
{
	"description": "An NLP ingest pipeline",
	"processors": [
		{
			"text_embedding": {
				"model_id": "-l5Dr5UB17VvHPspuB6C",
				"field_map": {
				},
				"skip_existing": true/false
			}
		}
	]
}
```
**Index Mapping**
```
{
  "settings": {
    "index.knn": true,
    "number_of_shards": 3,
    "default_pipeline": "nlp-ingest-pipeline"
  },
  "mappings": {
    "properties": {
      "id": {
        "type": "text"
      },
      "passage_embedding": {
        "type": "knn_vector",
        "dimension": 768,
        "method": {
          "engine": "lucene",
          "space_type": "l2",
          "name": "hnsw",
          "parameters": {}
        }
      },
      "text": {
          "type": "text"
      }
    }
  }
}
```
**Ingest Latency**

The following table presents latency measurements of initial ingest operation (in milliseconds) with *skip_existing* feature enabled and disabled. The Percent difference columns show the relative performance impact between the two.

|Operation	|Doc Size	|Batch Size	|skip_existing_off latency (ms)	|skip_existing_on latency (ms)	|Percent Difference	|
|---	|---	|---	|---	|---	|---	|
|Single Ingest 	|1000	|1	|459896.33	|463579.12	|0.80%	|
|Single Ingest	|2000	|1	|932563.3	|938318.45	|0.62%	|
|Single Ingest 	|3000	|1	|1400710.43	|1401216.8	|0.04%	|
|Batch Ingest 	|171332	|200	|2247191.38	|2192883.54	|-2.42%	|
|Batch Ingest 	|171332	|500	|2065514.4	|2011408.73	|-2.62%	|

**Update Latency**

The following table presents the latency measurements of update operation after identical ingest operation (in milliseconds) with *skip_existing* feature enabled and disabled. The Percent difference columns show the relative performance impact between the two.

|Operation	|Doc Size	|Batch Size	|skip_existing_off latency (ms)	|skip_existing_on latency (ms)	|Percent Difference	|
|---	|---	|---	|---	|---	|---	|
|Single Update 	|1000	|1	|459896.33	|100957.91	|-78.05%	|
|Single Update 	|2000	|1	|932563.3	|201715.16	|-78.37%	|
|Single Update	|3000	|1	|1400710.43	|292020.02	|-79.15%	|
|Batch Update 	|171332	|200	|2247191.38	|352767.2	|-84.30%	|
|Batch Update 	|171332	|500	|2065514.4	|293164.93	|-85.81%	|

### Related Issues

### Check List


junqiu-lei (approved):
Thanks for sharing the benchmark results, LGTM.

heemin32:
Could you share the index mapping for knn field?

heemin32:
Can we test with shard number 3?
```
{
  "settings": {
    "index.knn": true,
    "number_of_shards": 3,
    "default_pipeline": "nlp-ingest-pipeline"
  },
  "mappings": {
    "properties": {
      "id": {
        "type": "text"
      },
      "passage_embedding": {
        "type": "knn_vector",
        "dimension": 768,
        "method": {
          "engine": "lucene",
          "space_type": "l2",
          "name": "hnsw",
          "parameters": {}
        }
      },
      "text": {
          "type": "text"
      }
    }
  }
}
```

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -168,49 +173,91 @@ void preprocessIngestDocument(IngestDocument ingestDocument) {
...
+     */
     @Override
     public void subBatchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers, Consumer<List<IngestDocumentWrapper>> handler) {
-        if (CollectionUtils.isEmpty(ingestDocumentWrappers)) {
-            handler.accept(Collections.emptyList());
-            return;
-        }
+        try {
+            if (CollectionUtils.isEmpty(ingestDocumentWrappers)) {
+                handler.accept(Collections.emptyList());
```
bzhangam, line 185: nit: Should we simply return ingestDocumentWrappers rather than construct a new empty list?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -168,49 +173,91 @@ void preprocessIngestDocument(IngestDocument ingestDocument) {
...
-            for (IngestDocumentWrapper ingestDocumentWrapper : ingestDocumentWrappers) {
-                // The IngestDocumentWrapper might already run into exception and not sent for inference. So here we only
-                // set exception to IngestDocumentWrapper which doesn't have exception before.
-                if (ingestDocumentWrapper.getException() == null) {
-                    ingestDocumentWrapper.update(ingestDocumentWrapper.getIngestDocument(), exception);
+            }, exception -> {
+                try {
+                    for (IngestDocumentWrapper ingestDocumentWrapper : ingestDocumentWrappers) {
+                        // The IngestDocumentWrapper might already run into exception and not sent for inference. So here we only
+                        // set exception to IngestDocumentWrapper which doesn't have exception before.
```
bzhangam, line 245: I think here we do updateWithExceptions(ingestDocumentWrappers, e) in multiple catch blocks. But we only consider the case that IngestDocumentWrapper might already run into exception and not sent for inference here which does not make sense. It's possible the other catch block can override the exception of IngestDocumentWrapper incorrectly. e.g. The IngestDocumentWrapper can have exception A and not sent for inference but now we can potentially override its exception related to inference.

I think we should either propagate the exception to one catch block and handle it with the consideration that IngestDocumentWrapper can already run into the exception. Or we should modify `updateWithExceptions` function to account for that.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -168,49 +173,91 @@ void preprocessIngestDocument(IngestDocument ingestDocument) {
...
+                        List<?> inferenceResults = results.subList(startIndex, startIndex + dataForInference.getInferenceList().size());
+                        startIndex += dataForInference.getInferenceList().size();
+                        setVectorFieldsToDocument(
+                            dataForInference.getIngestDocumentWrapper().getIngestDocument(),
+                            dataForInference.getProcessMap(),
+                            inferenceResults
+                        );
+                    }
+                    handler.accept(ingestDocumentWrappers);
+                } catch (Exception e) {
```
bzhangam, line 237: Wonder why we want to clearly try catch the exception here? I think if we don't catch the exception it will be caught by the exception handler automatically since it's set as the onFailure of a listener.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -168,49 +173,91 @@ void preprocessIngestDocument(IngestDocument ingestDocument) {
...
+                try {
+                    for (IngestDocumentWrapper ingestDocumentWrapper : ingestDocumentWrappers) {
+                        // The IngestDocumentWrapper might already run into exception and not sent for inference. So here we only
+                        // set exception to IngestDocumentWrapper which doesn't have exception before.
+                        if (ingestDocumentWrapper.getException() == null) {
+                            ingestDocumentWrapper.update(ingestDocumentWrapper.getIngestDocument(), exception);
+                        }
+                    }
+                    handler.accept(ingestDocumentWrappers);
+                } catch (Exception e) {
```
bzhangam, line 251: Why we need this try catch? If we can throw the exception when we update the exception in ingestDocumentWrappers then in the catch block we will throw exception again and will not invoke  `handler.accept`. If we can throw the exception when we do `handler.accept` shouldn't the exception handler of the handler to handle it rather than handle it here?

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -63,4 +112,86 @@ public void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler
...
+            // skip existing flag is turned off. Call doSubBatchExecute without filtering
+            if (skipExisting == false) {
+                doSubBatchExecute(ingestDocumentWrappers, inferenceList, dataForInferences, handler);
+                return;
+            }
+            // skipExisting flag is turned on, eligible inference texts in dataForInferences will be compared and filtered after embeddings
+            // are copied
+            openSearchClient.execute(
+                MultiGetAction.INSTANCE,
+                buildMultiGetRequest(ingestDocumentWrappers),
```
bzhangam, line 138: Do we need to get all doc? What if some docs don't have a value for inference?

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -63,4 +112,86 @@ public void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler
...
+                        // create a map of documents with key: doc_id and value: doc
+                        Map<String, Map<String, Object>> existingDocuments = createDocumentMap(multiGetItemResponses);
+                        List<DataForInference> filteredDataForInference = filterDataForInference(dataForInferences, existingDocuments);
+                        List<String> filteredInferenceList = constructInferenceTexts(filteredDataForInference);
+                        if (filteredInferenceList.isEmpty()) {
+                            handler.accept(ingestDocumentWrappers);
+                        } else {
+                            doSubBatchExecute(ingestDocumentWrappers, filteredInferenceList, filteredDataForInference, handler);
+                        }
+                    } catch (Exception e) {
```
bzhangam, line 155: Same as above do we really want to update the exception for all docs? Do we really need to have multiple try catch here? It seems like we are doing the same thing in the catch block. Can we pop up the exception and simply handle it at the top layer?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -168,49 +173,91 @@ void preprocessIngestDocument(IngestDocument ingestDocument) {
...
+     * @param handler a callback handler to handle inference results which is a list of objects.
+     */
+    protected void doSubBatchExecute(
+        List<IngestDocumentWrapper> ingestDocumentWrappers,
+        List<String> inferenceList,
+        List<DataForInference> dataForInferences,
+        Consumer<List<IngestDocumentWrapper>> handler
+    ) {
+        try {
+            Tuple<List<String>, Map<Integer, Integer>> sortedResult = sortByLengthAndReturnOriginalOrder(inferenceList);
```
bzhangam, line 216: Wonder why we need to sort the inference list here? Seems like we sort it and then restore it without doing anything special.

will-hwang: sort is done before inference call is made, and results is the retrieved embedding from the inference call. I think it's safe to leave as is for now

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
will-hwang: exception handler is catching the exception thrown only in doBatchExecute method [here](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java#L60). we do not catch exceptions if any of the lines between 221-236 throws exceptions.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
will-hwang: ActionListener.wrap(response -> {}) spawns a new thread, and needs to be handled separately

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -168,49 +173,91 @@ void preprocessIngestDocument(IngestDocument ingestDocument) {
...
+                    }
+                    handler.accept(ingestDocumentWrappers);
+                } catch (Exception e) {
+                    updateWithExceptions(ingestDocumentWrappers, e);
+                    handler.accept(ingestDocumentWrappers);
                 }
-            }
+
+            });
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
will-hwang: i'm okay with simplifying the try/catch, or even removing if we know for sure the underlying methods will not throw exceptions. But this was brought up as a possible concern in the previous PR, so decided to add them

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
will-hwang: this was added to ensure that if any of the lines while handling the exceptions in for loop throws exceptions, it will be handled properly. I understand that this may not be needed, but was brought up as a possible risk in previous pr

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
will-hwang: we need to get all docs first to check if docs have or don't have value for inference

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
heemin32: Good catch. `handler.accept()` should go to `finally` block.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
heemin32: When would this situation occur? `The IngestDocumentWrapper might have already encountered an exception and was not sent for inference.` It would be helpful if you could point out the code where the exception is set before inference.

I agree that there's an inconsistency in how we handle this—checking if an exception is already set only in this specific place but not elsewhere.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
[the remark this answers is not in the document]
heemin32: I think @bzhangam is talking about
```
try {
  sortByLengthAndReturnOriginalOrder()
} catch {
}
```
where we only wrap try-catch around `sortByLengthAndReturnOriginalOrder`.  but not removing it.


I think having try-catch wrapping the entire method might be simpler and safer. 
If the code is hard to read, we can try to introduce a method to return lambda.

```
protected void doSubBatchExecute(
        List<IngestDocumentWrapper> ingestDocumentWrappers,
        List<String> inferenceList,
        List<DataForInference> dataForInferences,
        Consumer<List<IngestDocumentWrapper>> handler
    ) {
   ...
   doBatchExecute(inferenceList, batchExecuteHandler());
   ...
}


private Consumer<InferenceResult> batchExecuteHandler() {
 ...
}
```

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
will-hwang: see [link](https://github.com/opensearch-project/neural-search/blob/fc5936c1c8b9d44eabd441c1393b846d0c707b01/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java#L310-L311)

will-hwang: maybe we can propagate the exception to top level and set the exceptions there?

heemin32: I don't think we can propagate the exception. One option is to remove this check and override the exception for all documents. When an exception occurs here, retaining the original exception may not make a significant difference.

will-hwang: another option is to keep the check for all exception cases, which will only store the first exception encountered for all documents.

will-hwang: this way, we'll know which ingest document failed with which exact exception

heemin32: that would work as well.

will-hwang:
> Can we test with shard number 3?
> 
> ```
> {
>   "settings": {
>     "index.knn": true,
>     "number_of_shards": 3,
>     "default_pipeline": "nlp-ingest-pipeline"
>   },
>   "mappings": {
>     "properties": {
>       "id": {
>         "type": "text"
>       },
>       "passage_embedding": {
>         "type": "knn_vector",
>         "dimension": 768,
>         "method": {
>           "engine": "lucene",
>           "space_type": "l2",
>           "name": "hnsw",
>           "parameters": {}
>         }
>       },
>       "text": {
>           "type": "text"
>       }
>     }
>   }
> }
> ```

yes, will update the table with shard number 3 before merge. So far, it's showing similar results

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
bzhangam: I understand the sort is done before the inference call but I don't know why we sort it. Because we simply send the sorted data to the inference call without any logic e.g. group or batch the data. If that's the case this seems like a redundant logic.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
bzhangam: I think in onResponse function of the listener if you don't explicitly catch the exception it will be handled by the onFailure function so no need to catch it here since we want to do the same thing as we are doing in the onFailure.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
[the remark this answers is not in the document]
bzhangam: Yeah. I'm more thinking can we make the try catch cleaner. I think when we do try catch we want to do some special logic to handle the exception. But here we do the same logic with multiple nested try catch which is kind complicated the code.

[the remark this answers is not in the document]
heemin32: Agree. I think keep passing handler to downstream method is not a good practice as that will make us to keep adding all those try-catch for every method. In general, we should pass handler only when we make async call.

@bzhangam If you have more specific recommendation on how we should refactor the code here, please provide one.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
bzhangam: I wonder do we really check if the doc already run into the exception and skip the embedding generation for it? If that's the case it makes sense to keep the actual exception fail the ingest of the doc which can be helpful for debugging.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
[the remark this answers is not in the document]
will-hwang: i've refactored the code to separate out the try catch to a separate method as heemin suggested. Do you have other suggestions on how to make try/catch cleaner?

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
bzhangam: If the doc in the request doesn't have a value for the neural field which means we should not generate embedding for it since there is no value there. In that I think we can skip pulling the existing doc since we know there is no neural field we need to inference.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -168,23 +176,67 @@ void preprocessIngestDocument(IngestDocument ingestDocument) {
...
+            inferenceList = sortedResult.v1();
+            Map<Integer, Integer> originalOrder = sortedResult.v2();
+            doBatchExecute(
+                inferenceList,
+                results -> batchExecuteHandler(results, ingestDocumentWrappers, dataForInferences, originalOrder, handler),
+                exception -> {
+                    updateWithExceptions(ingestDocumentWrappers, handler, exception);
+                }
+            );
+        } catch (Exception e) {
```
bzhangam, line 227: If we don't plan to do something special here I think we can rely on the caller to handle the exception.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -201,16 +253,9 @@ public void subBatchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers,
...
-            for (IngestDocumentWrapper ingestDocumentWrapper : ingestDocumentWrappers) {
-                // The IngestDocumentWrapper might already run into exception and not sent for inference. So here we only
-                // set exception to IngestDocumentWrapper which doesn't have exception before.
-                if (ingestDocumentWrapper.getException() == null) {
-                    ingestDocumentWrapper.update(ingestDocumentWrapper.getIngestDocument(), exception);
-                }
-            }
-            handler.accept(ingestDocumentWrappers);
-        });
+        } catch (Exception e) {
```
bzhangam, line 256: I think this is not needed since the exception handler will catch this exception. They are wrapped by an ActionListener wrap in doBatchExecute.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -503,6 +578,27 @@ private void processMapEntryValue(
...
+            for (IngestDocumentWrapper ingestDocumentWrapper : ingestDocumentWrappers) {
+                // The IngestDocumentWrapper might have already run into exception. So here we only
+                // set exception to IngestDocumentWrapper which doesn't have exception before.
+                if (ingestDocumentWrapper.getException() == null) {
+                    ingestDocumentWrapper.update(ingestDocumentWrapper.getIngestDocument(), e);
+                }
+            }
+            handler.accept(ingestDocumentWrappers);
+        } catch (Exception ex) {
+            handler.accept(null);
```
bzhangam, line 598: I think we should log the error here to help debug since we cannot add the exception to the doc in this case.

will-hwang: added

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
will-hwang: removed

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
will-hwang: removed to handle it at caller

bzhangam (approved):

heemin32 (approved):
LGTM. Thanks.

PR #1246, opened 2025-03-25
Implement Optimized embedding generation in sparse encoding processor

will-hwang:
### Description
This PR implements embedding optimization for sparse encoding processor. This PR contains similar changes as previously merged PR for optimizing text embedding processor. 

Text Embedding Processor Optimization PR: [link](https://github.com/opensearch-project/neural-search/pull/1238)
Related RFC: https://github.com/opensearch-project/neural-search/issues/1138

### Setup

**Version**: OS 3.0 alpha
**Cluster**: 4 r6g.2xlarge nodes (1 coordinator node 3 data nodes)

**Dataset:** Trec-covid (https://public.ukp.informatik.tu-darmstadt.de/thakur/BEIR/datasets/trec-covid.zip)
**Inference Model:** amazon/neural-sparse/opensearch-neural-sparse-encoding-doc-v2-distill

**Pipeline:** 

```
{
  "description": "A sparse encoding pipeline",
  "processors": [
    {
      "sparse_encoding": {
        "model_id": "nHXUzpUBA788_009ga9N",
        "prune_type": "max_ratio",
        "prune_ratio": 0.1,
        "field_map": {
        },
        "skip_existing": true/false
      }
    }
  ]
}
```

**Index:**

```
{
  "settings": {
    "index.knn": true,
    "number_of_shards": 3,
    "default_pipeline": "nlp-ingest-pipeline"
  },
  "mappings": {
    "properties": {
      "id": {
        "type": "text"
      },
      "passage_embedding": {
        "type": "rank_features"
      },
      "passage_text": {
        "type": "text"
      }
    }
  }
}
```

**Ingest Latency**

The following table presents latency measurements of initial ingest operation (in milliseconds) with *skip_existing* feature enabled and disabled. The Percent difference columns show the relative performance impact between the two.

|Operation	|Prune Type	|Prune Ratio	|Doc Size	|Batch Size	|skip_existing_off latency (ms)	|skip_existing_on latency (ms)	|Difference (ms)	|Percent Difference	|
|---	|---	|---	|---	|---	|---	|---	|---	|---	|
|Single Ingest 	|max_ratio	|0.1	|1000	|1	|623172.06	|655658.66	|32486.6	|5.21%	|
|Single Ingest	|max_ratio	|0.1	|2000	|1	|1241234.59	|1310683.17	|69448.58	|5.60%	|
|Single Ingest 	|max_ratio	|0.1	|3000	|1	|1942907.49	|1965918.53	|23011.04	|1.18%	|
|Batch Ingest 	|max_ratio	|0.1	|171332	|200	|3077040.12	|3101697.22	|24657.1	|0.80%	|
|Batch Ingest 	|max_ratio	|0.1	|171332	|500	|2889452.28	|2905052.15	|15599.87	|0.54%	|

**Update Latency**

The following table presents the latency measurements of update operation after identical ingest operation (in milliseconds) with *skip_existing* feature enabled and disabled. The Percent difference columns show the relative performance impact between the two.

|Operation	|Prune Type	|Prune Ratio	|Doc Size	|Batch Size	|skip_existing_off latency (ms)	|skip_existing_on latency (ms)	|Difference (ms)	|Percent Difference	|
|---	|---	|---	|---	|---	|---	|---	|---	|---	|
|Single Update 	|max_ratio	|0.1	|1000	|1	|623172.06	|111144.76	|-512027.3	|-82.16%	|
|Single Update 	|max_ratio	|0.1	|2000	|1	|1241234.59	|213015.48	|-1028219.11	|-82.84%	|
|Single Update	|max_ratio	|0.1	|3000	|1	|1942907.49	|306765.73	|-1636141.76	|-84.21%	|
|Batch Update 	|max_ratio	|0.1	|171332	|200	|3077040.12	|475197.34	|-2601842.78	|-84.56%	|
|Batch Update 	|max_ratio	|0.1	|171332	|500	|2889452.28	|437750.26	|-2451702.02	|-84.85%	|

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
@@ -48,6 +48,9 @@ import com.google.common.collect.ImmutableMap;
 import lombok.extern.log4j.Log4j2;
 import org.opensearch.neuralsearch.processor.optimization.InferenceFilter;
 import org.opensearch.neuralsearch.util.ProcessorDocumentUtils;
+import org.opensearch.neuralsearch.util.TokenWeightUtil;
+import org.opensearch.neuralsearch.util.prune.PruneType;
+import org.opensearch.neuralsearch.util.prune.PruneUtils;
 
 /**
  * The abstract class for text processing use cases. Users provide a field name map and a model id.
@@ -61,6 +64,8 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
     public static final String FIELD_MAP_FIELD = "field_map";
     public static final String INDEX_FIELD = "_index";
     public static final String ID_FIELD = "_id";
+    public static final String SKIP_EXISTING = "skip_existing";
+    public static final boolean DEFAULT_SKIP_EXISTING = false;
     private static final BiFunction<Object, Object, Object> REMAPPING_FUNCTION = (v1, v2) -> {
         if (v1 instanceof Collection && v2 instanceof Collection) {
             ((Collection) v1).addAll((Collection) v2);
@@ -672,7 +677,7 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
     }
 
     // This method validates and filters given inferenceList and processMap after response is successfully retrieved from get operation.
-    protected void getResponseHandler(
+    protected void reuseOrGenerateEmbedding(
         GetResponse response,
         IngestDocument ingestDocument,
         Map<String, Object> processMap,
@@ -682,11 +687,11 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
     ) {
         final Map<String, Object> existingDocument = response.getSourceAsMap();
         if (existingDocument == null || existingDocument.isEmpty()) {
-            makeInferenceCall(ingestDocument, processMap, inferenceList, handler);
+            generateAndSetInference(ingestDocument, processMap, inferenceList, handler);
             return;
         }
         // filter given ProcessMap by comparing existing document with ingestDocument
-        Map<String, Object> filteredProcessMap = inferenceFilter.filter(
+        Map<String, Object> filteredProcessMap = inferenceFilter.filterAndCopyExistingEmbeddings(
             existingDocument,
             ingestDocument.getSourceAndMetadata(),
             processMap
@@ -698,13 +703,13 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
         if (filteredInferenceList.isEmpty()) {
             handler.accept(ingestDocument, null);
         } else {
-            makeInferenceCall(ingestDocument, filteredProcessMap, filteredInferenceList, handler);
+            generateAndSetInference(ingestDocument, filteredProcessMap, filteredInferenceList, handler);
         }
     }
 
     // This method validates and filters given inferenceList and dataForInferences after response is successfully retrieved from multi-get
     // operation.
-    protected void multiGetResponseHandler(
+    protected void reuseOrGenerateEmbedding(
         MultiGetResponse response,
         List<IngestDocumentWrapper> ingestDocumentWrappers,
         List<String> inferenceList,
@@ -749,7 +754,11 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
             // filter dataForInference when existing document exists
             String docId = id.toString();
             Map<String, Object> existingDocument = existingDocuments.get(docId);
-            Map<String, Object> filteredProcessMap = inferenceFilter.filter(existingDocument, document, processMap);
+            Map<String, Object> filteredProcessMap = inferenceFilter.filterAndCopyExistingEmbeddings(
+                existingDocument,
+                document,
+                processMap
+            );
             List<String> filteredInferenceList = createInferenceList(filteredProcessMap);
             filteredDataForInference.add(new DataForInference(ingestDocumentWrapper, filteredProcessMap, filteredInferenceList));
         }
@@ -770,7 +779,7 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
      * @param handler SourceAndMetadataMap of ingestDocument Document
      *
      */
-    protected void makeInferenceCall(
+    protected void generateAndSetInference(
         IngestDocument ingestDocument,
         Map<String, Object> processMap,
         List<String> inferenceList,
@@ -780,6 +789,38 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
             TextInferenceRequest.builder().modelId(this.modelId).inputTexts(inferenceList).build(),
             ActionListener.wrap(vectors -> {
                 setVectorFieldsToDocument(ingestDocument, processMap, vectors);
+                handler.accept(ingestDocument, null);
+            }, e -> { handler.accept(null, e); })
+        );
+    }
+
+    /**
+     * This method invokes inference call that returns results as map through mlCommonsClientAccessor and populates retrieved embeddings to ingestDocument
+     *
+     * @param ingestDocument ingestDocument to populate embeddings to
+     * @param processMap map indicating the path in ingestDocument to populate embeddings
+     * @param inferenceList list of texts to be model inference
+     * @param pruneType    The type of prune strategy to use
+     * @param pruneRatio   The ratio or threshold for prune
+     * @param handler handler for accepting IngestionDocument
+     *
+     */
+    protected void generateAndSetMapInference(
+        IngestDocument ingestDocument,
+        Map<String, Object> processMap,
+        List<String> inferenceList,
+        PruneType pruneType,
+        float pruneRatio,
+        BiConsumer<IngestDocument, Exception> handler
+    ) {
+        mlCommonsClientAccessor.inferenceSentencesWithMapResult(
+            TextInferenceRequest.builder().modelId(this.modelId).inputTexts(inferenceList).build(),
+            ActionListener.wrap(resultMaps -> {
+                List<Map<String, Float>> sparseVectors = TokenWeightUtil.fetchListOfTokenWeightMap(resultMaps)
+                    .stream()
+                    .map(vector -> PruneUtils.pruneSparseVector(pruneType, pruneRatio, vector))
+                    .toList();
+                setVectorFieldsToDocument(ingestDocument, processMap, sparseVectors);
                 handler.accept(ingestDocument, null);
             }, e -> { handler.accept(null, e); })
         );
--- a/src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
@@ -6,20 +6,29 @@ package org.opensearch.neuralsearch.processor;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 
 import lombok.Getter;
+import org.opensearch.action.get.GetAction;
+import org.opensearch.action.get.GetRequest;
+import org.opensearch.action.get.MultiGetAction;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.core.action.ActionListener;
+import org.opensearch.core.common.util.CollectionUtils;
 import org.opensearch.env.Environment;
 import org.opensearch.ingest.IngestDocument;
+import org.opensearch.ingest.IngestDocumentWrapper;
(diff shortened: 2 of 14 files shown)


src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
```diff
@@ -85,4 +127,47 @@ public void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler
             }, onException)
         );
     }
+
+    @Override
+    public void subBatchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers, Consumer<List<IngestDocumentWrapper>> handler) {
```
zhichao-aws, line 132: All subclass for `InferenceProcessor` override the `subBatchExecute` now. Should we just modify the method in `InferenceProcessor`?

zhichao-aws:
In terms of the benchmarks on Latency, I'd prefer measure the absolute value of the `skip_existing` instead of  percent difference. Like it will take xxx-yyy ms for 1000 documents.

In current benchmark the ingest throughput is very low even without `skip_existing` (<2 doc/s). But with remote connector + inferentia/GPU, the ingest throughput can reach 400 even 800 docs per seconds. In this case, is it possible that we get larger update throughput when turn off the flag?

src/test/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessIT.java
```diff
@@ -83,4 +83,23 @@ public void testSparseEncodingProcessorWithReindex() throws Exception {
...
     }
+
+    public void testSparseEncodingProcessorWithSkipExisting() throws Exception {
+        String modelId = null;
+        modelId = prepareSparseEncodingModel();
+        createPipelineProcessor(modelId, PIPELINE_NAME, ProcessorType.SPARSE_ENCODING_WITH_SKIP_EXISTING);
+        createIndexWithPipeline(INDEX_NAME, "SparseEncodingIndexMappings.json", PIPELINE_NAME);
+        ingestDocument(INDEX_NAME, INGEST_DOCUMENT, "1");
+        updateDocument(INDEX_NAME, INGEST_DOCUMENT, "1");
+        assertEquals(1, getDocCount(INDEX_NAME));
```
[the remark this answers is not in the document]
will-hwang: the count of model inference call made for update is tested in unit tests

src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
will-hwang: yes, we can do that too, but that would force new processors that extend from `InferenceProcessor` to implement the skip_existing feature. I think it's better to give more flexibility for future use cases

will-hwang:
> In terms of the benchmarks on Latency, I'd prefer measure the absolute value of the `skip_existing` instead of percent difference. Like it will take xxx-yyy ms for 1000 documents.
> 
> In current benchmark the ingest throughput is very low even without `skip_existing` (<2 doc/s). But with remote connector + inferentia/GPU, the ingest throughput can reach 400 even 800 docs per seconds. In this case, is it possible that we get larger update throughput when turn off the flag?

sure, i can add a column for absolute value. To your second point, we are mostly concerned about `get`/`mget` calls we make in the feature do not seriously degrade performance. We don't necessarily have to improve the update throughput, since the main objective of the feature is to skip inference call when eligible.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -785,6 +788,38 @@ protected void makeInferenceCall(
...
+     *
+     * @param ingestDocument ingestDocument to populate embeddings to
+     * @param processMap map indicating the path in ingestDocument to populate embeddings
+     * @param inferenceList list of texts to be model inference
+     * @param pruneType    The type of prune strategy to use
+     * @param pruneRatio   The ratio or threshold for prune
+     * @param handler handler for accepting IngestionDocument
+     *
+     */
+    protected void makeInferenceCallWithMapResult(
```
bzhangam, line 802: nit: Should we name it like `generateAndSetInference` to make it more descriptive?

src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
```diff
@@ -41,35 +58,60 @@ public SparseEncodingProcessor(
...
-                    .map(vector -> PruneUtils.pruneSparseVector(pruneType, pruneRatio, vector))
-                    .toList();
-                setVectorFieldsToDocument(ingestDocument, ProcessMap, sparseVectors);
+        if (skipExisting == false) {
+            makeInferenceCallWithMapResult(ingestDocument, processMap, inferenceList, pruneType, pruneRatio, handler);
+            return;
+        }
+        // if skipExisting flag is turned on, eligible inference texts will be compared and filtered after embeddings are copied
+        String index = ingestDocument.getSourceAndMetadata().get(INDEX_FIELD).toString();
+        String id = ingestDocument.getSourceAndMetadata().get(ID_FIELD).toString();
```
bzhangam, line 91: When id is not provided in a create request we will auto generate an id [here](https://github.com/opensearch-project/OpenSearch/blob/e50baefb8f962e9a470338b81ab31f256baa05c3/server/src/main/java/org/opensearch/action/bulk/TransportBulkAction.java#L579) which is after the ingest. So we cannot assume we always have the id. The only case the id doesn't exist is create opType which we always need to generate the embedding so I think we can skip trying to find the existing doc for the doc without id.

src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
```diff
@@ -85,4 +127,47 @@ public void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler
...
+                doSubBatchExecute(ingestDocumentWrappers, inferenceList, dataForInferences, handler);
+                return;
+            }
+            // skipExisting flag is turned on, eligible inference texts in dataForInferences will be compared and filtered after embeddings
+            // are copied
+            openSearchClient.execute(
+                MultiGetAction.INSTANCE,
+                buildMultiGetRequest(dataForInferences),
+                ActionListener.wrap(
+                    response -> multiGetResponseHandler(
```
bzhangam, line 155: Can we rename `multiGetResponseHandler` to something like `reuseOrGenerateEmbedding` to show this function will try to reuse if possible otherwise generate embedding?

src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
will-hwang: added a check. thanks

src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
zhichao-aws: My point is the inference processor is the base class for all processors need ML inference. And for these processors, it makes sense that they inherit the logics to check existing to skip inference. And they can also choose to override the `subBatchExecute `

will-hwang: `skip_existing` feature is defined at a specific processor level, not at the `InferenceProcessor`. If the feature is enabled for one, but not for the other, it would make it difficult to have this check shared in `InferenceProcessor`. If there's a clean way to implement this, please share

src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
```diff
@@ -41,35 +54,64 @@ public SparseEncodingProcessor(
...
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());
+            if (filteredInferenceList.isEmpty()) {
                 handler.accept(ingestDocument, null);
-            }, e -> { handler.accept(null, e); })
-        );
+            } else {
+                generateAndSetMapInference(ingestDocument, filteredProcessMap, filteredInferenceList, pruneType, pruneRatio, handler);
+            }
+
```
q-andy, line 113: nit: empty line

src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
```diff
@@ -41,35 +54,64 @@ public SparseEncodingProcessor(
...
+            return;
+        }
+        // if skipExisting flag is turned on, eligible inference texts will be compared and filtered after embeddings are copied
+        Object index = ingestDocument.getSourceAndMetadata().get(INDEX_FIELD);
+        Object id = ingestDocument.getSourceAndMetadata().get(ID_FIELD);
+        if (Objects.isNull(index) || Objects.isNull(id)) {
+            generateAndSetMapInference(ingestDocument, processMap, inferenceList, pruneType, pruneRatio, handler);
+            return;
+        }
+        openSearchClient.execute(GetAction.INSTANCE, new GetRequest(index.toString(), id.toString()), ActionListener.wrap(response -> {
```
q-andy, line 92: nit: this line is hard to read, maybe break the parameters of the execute call into separate lines so the inner ActionListener wrap is on its own line? Like the original inferenceSentencesWithMapResult call

src/main/java/org/opensearch/neuralsearch/processor/factory/SparseEncodingProcessorFactory.java
```diff
@@ -4,13 +4,16 @@
...
+import static org.opensearch.ingest.ConfigurationUtils.readBooleanProperty;
 import static org.opensearch.ingest.ConfigurationUtils.readMap;
 import static org.opensearch.ingest.ConfigurationUtils.readStringProperty;
 import static org.opensearch.ingest.ConfigurationUtils.readOptionalStringProperty;
 import static org.opensearch.ingest.ConfigurationUtils.readDoubleProperty;
+import static org.opensearch.neuralsearch.processor.SparseEncodingProcessor.DEFAULT_SKIP_EXISTING;
 import static org.opensearch.neuralsearch.processor.TextEmbeddingProcessor.MODEL_ID_FIELD;
 import static org.opensearch.neuralsearch.processor.TextEmbeddingProcessor.FIELD_MAP_FIELD;
 import static org.opensearch.neuralsearch.processor.SparseEncodingProcessor.TYPE;
+import static org.opensearch.neuralsearch.processor.SparseEncodingProcessor.SKIP_EXISTING;
```
q-andy, line 16: nit: Should these imports be updated to get the constant from InferenceProcessor since the constants were moved? Or does it compile fine?

q-andy (approved):
LGTM, left some minor optional nits. Thanks!

src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
will-hwang: this is auto-formatted with `spotlessApply`

src/main/java/org/opensearch/neuralsearch/processor/factory/SparseEncodingProcessorFactory.java
will-hwang: the constants are inherited, so no imports need to be changed

src/test/java/org/opensearch/neuralsearch/processor/factory/SparseEncodingEmbeddingProcessorFactoryTests.java
```diff
@@ -179,4 +183,25 @@ public void testCreateProcessor_whenMissingPruneType_thenFail() {
         );
         assertEquals("prune_ratio field is not supported when prune_type is not provided", exception.getMessage());
     }
+
+    @SneakyThrows
+    public void testCreateProcessor_whenSkipExistingPassed_thenSuccessful() {
```
junqiu-lei, line 188: nit: can we add more test scenarios for skipping existing verification tests:
   - Test with completely unchanged content
   - Test with partially modified fields
   - Test with edge cases (empty fields, null values)

junqiu-lei (approved):
Overall LGTM, left some minor comment, thanks!

src/test/java/org/opensearch/neuralsearch/processor/factory/SparseEncodingEmbeddingProcessorFactoryTests.java
will-hwang: the new test cases do include
1. unchanged content
2. test with partially modified fields

edge cases are already handled in the existing test cases

will-hwang: added additional test case

Issue #476, opened 2023-10-26
[FEATURE] Multiple embeddings in one data ingestion request 


martin-gaievski:
### Is your feature request related to a problem?
Currently neural-search text_image_processor allows a single document field to be defined for each image and text mapping. A single field can be defined to stere embedding in OpenSearch. Example of processor definition:

```
{
    "description": "An example neural search pipeline",
    "processors": [
        {
            "text_image_embedding": {
                "model_id": "1234567890",
                "embedding": "vector_embedding",
                "field_map": {
                    "text": "caption",
                    "image": "field_with_image"
                }
            }
        }
    ]
}
```

### What solution would you like?
It should be possible to define multiple field pairs for image, text or image+text. It should be possible to define an OpenSearch field that stores embedding for a model. Request may look something like:

```
{
  "description": "An example neural search pipeline",
  "processors" : [
    {
      "text-image-embedding": {
        "model_id": "some_remote_model",
        "field_map": {
            "multimodal_embedding_1": {                                                
                 "text": "caption_1",
                 "image": "field_with_image_1"
            },
            "multimodal_embedding_2": {                                                
                 "text": "caption_2",
                 "image": "field_with_image_2"
            }
        }
    }
  ]
}
```

### What alternatives have you considered?
Today it's possible to define multiple embedding processors as part of a single pipeline, and each processor may have it's own definition of mapping and embedding field. 

```
{
    "description": "An example neural search pipeline",
    "processors": [
        {
            "text_image_embedding": {
                "model_id": "1234567890",
                "embedding": "vector_embedding_1",
                "field_map": {
                    "text": "caption_1",
                    "image": "field_with_image_1"
                }
            }
        },
        {
            "text_image_embedding": {
                "model_id": "1234567890",
                "embedding": "vector_embedding_2",
                "field_map": {
                    "text": "caption_2",
                    "image": "field_with_image_2"
                }
            }
        }
    ]
}
```

navneet1v:
@martin-gaievski It would be great if you can provide some details why the alternative approach which is provided is not feasible.

martin-gaievski:
Alternative with multiple processors is harder to maintain then a single processor entry. For instance, if one needs to update model id it should be done in X places vs a single place. Another concern is the performance, having a multiple processors adding some overhead that may be critical in systems with low SLA for search response.

PR #1249, opened 2025-03-28
Implement Optimized embedding generation in text and image embedding processor

will-hwang:
### Description
This PR implements embedding optimization for text/image embedding processor. 

Related PRs:

Text Embedding Processor Optimization PR: [link](https://github.com/opensearch-project/neural-search/pull/1238)
Sparse Encoding Processor Optimization PR: [link](https://github.com/opensearch-project/neural-search/pull/1246)
Related RFC: https://github.com/opensearch-project/neural-search/issues/1138

### Related Issues

### Check List


src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -107,17 +123,30 @@ public void execute(final IngestDocument ingestDocument, final BiConsumer<Ingest
...
-            } else {
-                mlCommonsClientAccessor.inferenceSentencesMap(
-                    MapInferenceRequest.builder().modelId(this.modelId).inputObjects(inferenceMap).build(),
-                    ActionListener.wrap(vectors -> {
-                        setVectorFieldsToDocument(ingestDocument, vectors);
-                        handler.accept(ingestDocument, null);
-                    }, e -> { handler.accept(null, e); })
-                );
+            if (skipExisting == false) {
+                if (inferenceMap.isEmpty()) {
```
bzhangam, line 127: If inferenceMap is empty we should end the work no matter skipExisting is true or false.

src/main/java/org/opensearch/neuralsearch/processor/optimization/TextImageEmbeddingInferenceFilter.java
```diff
@@ -0,0 +1,51 @@
...
+    public Map<String, String> filterAndCopyExistingEmbeddings(
+        IngestDocument ingestDocument,
+        Map<String, Object> existingDocument,
+        Map<String, String> knnMap,
+        String embeddingField
+    ) {
+        for (Map.Entry<String, String> entry : knnMap.entrySet()) {
+            String key = entry.getKey();
+            String value = entry.getValue();
+            if (existingDocument.containsKey(key) == false || existingDocument.get(key).equals(value) == false) {
```
bzhangam, line 40: I think we need to compare both the text and image here and this can be done by just checking one key?

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -107,17 +123,30 @@ public void execute(final IngestDocument ingestDocument, final BiConsumer<Ingest
...
-                handler.accept(ingestDocument, null);
-            } else {
-                mlCommonsClientAccessor.inferenceSentencesMap(
-                    MapInferenceRequest.builder().modelId(this.modelId).inputObjects(inferenceMap).build(),
-                    ActionListener.wrap(vectors -> {
-                        setVectorFieldsToDocument(ingestDocument, vectors);
-                        handler.accept(ingestDocument, null);
-                    }, e -> { handler.accept(null, e); })
-                );
+            if (skipExisting == false) {
```
weijia-aws, line 126:
```suggestion
            if (!skipExisting) {
```

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -107,17 +123,30 @@ public void execute(final IngestDocument ingestDocument, final BiConsumer<Ingest
...
+                } else {
+                    generateAndSetInference(ingestDocument, inferenceMap, handler);
+                }
+                return;
             }
+            // if skipExisting flag is turned on, eligible inference text and images will be compared and filtered after embeddings are
+            // copied
+            Object index = ingestDocument.getSourceAndMetadata().get(INDEX_FIELD);
+            Object id = ingestDocument.getSourceAndMetadata().get(ID_FIELD);
+            if (Objects.isNull(index) || Objects.isNull(id)) {
```
weijia-aws, line 138: Do we need to check if the value is empty or not?

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
will-hwang: thanks for pointing this out, but explicit check is recommended as per OS coding guideline here: https://github.com/opensearch-project/OpenSearch/blob/main/DEVELOPER_GUIDE.md#java-language-formatting-guidelines

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
will-hwang: okay, i can move up the check

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
will-hwang: yes, it was discussed in previous PR here: https://github.com/opensearch-project/neural-search/pull/1246#discussion_r2014990021

src/main/java/org/opensearch/neuralsearch/processor/optimization/TextImageEmbeddingInferenceFilter.java
will-hwang: knnMap contains two keys: one for text and one for value. For each entry, it will be compared with text and image values of the existing document

bzhangam: Synced offline that this code will work because currently we only allow user to define one image and one text field. So the knnMap only contains the text and image fields and both of them should be the same to reuse the existing embedding. We should add a comment to call it out.

Besides also thinking we may want to allow users to define multiple text and image fields in the processor. Probably we can create a RFC to see if there is a user need.

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
weijia-aws: I see, thanks for sharing

bzhangam (approved):

src/main/java/org/opensearch/neuralsearch/processor/optimization/TextImageEmbeddingInferenceFilter.java
heemin32: >Besides also thinking we may want to allow users to define multiple text and image fields in the processor. Probably we can create a RFC to see if there is a user need.
https://github.com/opensearch-project/neural-search/issues/476

weijia-aws (approved):

heemin32 (approved):

junqiu-lei (approved):

PR #1254, opened 2025-04-02
Support custom tags in semantic highlighter

junqiu-lei:
### Description
This PR adds support for custom highlighting tags in the semantic highlighter, allowing users to specify their own pre and post tags for highlighting text while default tags are`<em>` and `</em>` if these parameter are not provided. These parameters schema are already supported in OpenSearch highlighter framework.

### Example API
```
curl -X POST "localhost:9200/neural-search-index/_search?pretty" -H 'Content-Type: application/json' -d '{
  "_source": {
    "excludes": ["text_embedding"]
  },
  "query": {
    "neural": {
      "text_embedding": {
        "query_text": "AI systems that can learn",
        "model_id": "'$text_embedding_model_id'",
        "k": 5
      }
    }
  },
  "highlight": {
    "fields": {
      "text": {
        "type": "semantic"
      }
    },
    "options": {
      "model_id": "'$sentence_model_id'"
    },
    "pre_tags": [
      "<strong>"
    ],
    "post_tags": [
      "</strong>"
    ]
  }
}'
```

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
+++ b/src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
@@ -60,8 +60,18 @@ public class SemanticHighlighter implements Highlighter {
             return null;
         }
 
+        // The pre- and post- tags are provided by the user or defaulted to <em> and </em>
+        String[] preTags = fieldContext.field.fieldOptions().preTags();
+        String[] postTags = fieldContext.field.fieldOptions().postTags();
+
         // Get highlighted text - allow any exceptions from this call to propagate
-        String highlightedResponse = semanticHighlighterEngine.getHighlightedSentences(modelId, originalQueryText, fieldText);
+        String highlightedResponse = semanticHighlighterEngine.getHighlightedSentences(
+            modelId,
+            originalQueryText,
+            fieldText,
+            preTags[0],
+            postTags[0]
+        );
 
         if (highlightedResponse == null || highlightedResponse.isEmpty()) {
             log.warn("No highlighted text found for field {}", fieldContext.fieldName);
--- a/src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
+++ b/src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
@@ -13,6 +13,7 @@ import org.opensearch.search.fetch.subphase.highlight.FieldHighlightContext;
 import org.opensearch.neuralsearch.highlight.extractor.QueryTextExtractorRegistry;
 import org.opensearch.action.support.PlainActionFuture;
 import lombok.NonNull;
+import lombok.Builder;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -24,24 +25,18 @@ import java.util.Objects;
  * Engine class for semantic highlighting operations
  */
 @Log4j2
+@Builder
 public class SemanticHighlighterEngine {
     private static final String MODEL_ID_FIELD = "model_id";
-    private static final String DEFAULT_PRE_TAG = "<em>";
-    private static final String DEFAULT_POST_TAG = "</em>";
     private static final String MODEL_INFERENCE_RESULT_KEY = "highlights";
     private static final String MODEL_INFERENCE_RESULT_START_KEY = "start";
     private static final String MODEL_INFERENCE_RESULT_END_KEY = "end";
 
+    @NonNull
     private final MLCommonsClientAccessor mlCommonsClient;
-    private final QueryTextExtractorRegistry queryTextExtractorRegistry;
 
-    public SemanticHighlighterEngine(
-        @NonNull MLCommonsClientAccessor mlCommonsClient,
-        @NonNull QueryTextExtractorRegistry queryTextExtractorRegistry
-    ) {
-        this.mlCommonsClient = mlCommonsClient;
-        this.queryTextExtractorRegistry = queryTextExtractorRegistry;
-    }
+    @NonNull
+    private final QueryTextExtractorRegistry queryTextExtractorRegistry;
 
     /**
      * Gets the field text from the document
@@ -116,15 +111,17 @@ public class SemanticHighlighterEngine {
      * @param modelId The ID of the model to use
      * @param question The search query
      * @param context The document text
+     * @param preTag The pre tag to use for highlighting
+     * @param postTag The post tag to use for highlighting
      * @return Formatted text with highlighting
      */
-    public String getHighlightedSentences(String modelId, String question, String context) {
+    public String getHighlightedSentences(String modelId, String question, String context, String preTag, String postTag) {
         List<Map<String, Object>> results = fetchModelResults(modelId, question, context);
         if (results == null || results.isEmpty()) {
             return null;
         }
 
-        return applyHighlighting(context, results.getFirst());
+        return applyHighlighting(context, results.getFirst(), preTag, postTag);
     }
 
     /**
@@ -168,10 +165,12 @@ public class SemanticHighlighterEngine {
      *
      * @param context The original document text
      * @param highlightResult The highlighting result from the ML model
+     * @param preTag The pre tag to use for highlighting
+     * @param postTag The post tag to use for highlighting
      * @return Formatted text with highlighting
      * @throws IllegalArgumentException if highlight positions are invalid
      */
-    public String applyHighlighting(String context, Map<String, Object> highlightResult) {
+    public String applyHighlighting(String context, Map<String, Object> highlightResult, String preTag, String postTag) {
         // Get the "highlights" list from the result
         Object highlightsObj = highlightResult.get(MODEL_INFERENCE_RESULT_KEY);
 
@@ -216,7 +215,7 @@ public class SemanticHighlighterEngine {
             }
         }
 
-        return constructHighlightedText(context, validHighlights);
+        return constructHighlightedText(context, validHighlights, preTag, postTag);
     }
 
     /**
@@ -246,9 +245,11 @@ public class SemanticHighlighterEngine {
      *
      * @param text The original text
      * @param highlights The list of valid highlight positions in pairs [start1, end1, start2, end2, ...]
+     * @param preTag The pre tag to use for highlighting
+     * @param postTag The post tag to use for highlighting
      * @return The highlighted text
      */
-    private String constructHighlightedText(String text, List<Integer> highlights) {
+    private String constructHighlightedText(String text, List<Integer> highlights, String preTag, String postTag) {
         StringBuilder result = new StringBuilder();
         int currentPos = 0;
 
@@ -264,9 +265,9 @@ public class SemanticHighlighterEngine {
             }
 
             // Add the highlighted text with highlight tags
-            result.append(DEFAULT_PRE_TAG);
+            result.append(preTag);
             result.append(text, start, end);
-            result.append(DEFAULT_POST_TAG);
+            result.append(postTag);
 
             // Update current position to end of this highlight
             currentPos = end;
--- a/src/test/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngineTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngineTests.java
@@ -45,7 +45,10 @@ public class SemanticHighlighterEngineTests extends OpenSearchTestCase {
         super.setUp();
         mlCommonsClientAccessor = mock(MLCommonsClientAccessor.class);
         queryTextExtractorRegistry = new QueryTextExtractorRegistry();
-        highlighterEngine = new SemanticHighlighterEngine(mlCommonsClientAccessor, queryTextExtractorRegistry);
+        highlighterEngine = SemanticHighlighterEngine.builder()
+            .mlCommonsClient(mlCommonsClientAccessor)
+            .queryTextExtractorRegistry(queryTextExtractorRegistry)
+            .build();
 
         // Setup default mock behavior
         setupDefaultMockBehavior();
@@ -123,7 +126,7 @@ public class SemanticHighlighterEngineTests extends OpenSearchTestCase {
     }
 
     public void testGetHighlightedSentences() {
-        String result = highlighterEngine.getHighlightedSentences(MODEL_ID, TEST_QUERY, TEST_CONTENT);
+        String result = highlighterEngine.getHighlightedSentences(MODEL_ID, TEST_QUERY, TEST_CONTENT, "<em>", "</em>");
 
         assertNotNull("Should return highlighted text", result);
         assertTrue("Should contain highlighting tags", result.contains("<em>") && result.contains("</em>"));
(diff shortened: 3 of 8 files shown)


src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
```diff
@@ -24,23 +28,60 @@
...
+    /**
+     * Creates a new instance with updated highlight tags
+     *
+     * @param newPreTag The new pre tag to use, or null to keep the current one
+     * @param newPostTag The new post tag to use, or null to keep the current one
+     * @return A new SemanticHighlighterEngine instance with the updated tags
+     */
+    public SemanticHighlighterEngine withTags(String newPreTag, String newPostTag) {
+        // Either both tags must be set or both must be null
+        if ((newPreTag == null) != (newPostTag == null)) {
```
yizheliu-amazon, line 64: nit: instead of having multiple newPreTag == null or newPreTag != null check, you can have one boolean flag to store newPreTag != null result, and use the boolean flag in the following check. Same comment for newPostTag. 

Just want to make it more concise.

yizheliu-amazon (approved):

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
junqiu-lei: Updated

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
```diff
@@ -60,6 +60,12 @@ public HighlightField highlight(FieldHighlightContext fieldContext) {
             return null;
         }
 
+        String[] preTags = fieldContext.field.fieldOptions().preTags();
+        String[] postTags = fieldContext.field.fieldOptions().postTags();
+        if (preTags != null && preTags.length > 0 && postTags != null && postTags.length > 0) {
+            semanticHighlighterEngine = semanticHighlighterEngine.withTags(preTags[0], postTags[0]);
```
heemin32, line 66: Shouldn't we left a log for error case where only of of tags are provided?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
```diff
@@ -24,23 +28,62 @@
...
+        final boolean isPostTagNull = newPostTag == null;
+        if (isPreTagNull != isPostTagNull) {
+            throw new IllegalArgumentException("Both pre and post tags must be set together or both must be null");
+        }
 
-    private final MLCommonsClientAccessor mlCommonsClient;
-    private final QueryTextExtractorRegistry queryTextExtractorRegistry;
+        // Validate tag format if tags are provided
+        if (!isPreTagNull) {
+            if (!newPreTag.startsWith("<")) {
```
heemin32, line 72: Is this enough validation of its format? What if the pre tag is `<dfd`?
Is this logic same as what we have in opensearch core for other highlighter?

junqiu-lei: Actually the prefix validation isn't exist in OpenSearch core, core only has:
```
  if (hb.preTags() != null && hb.postTags() == null) {
      throw new ParsingException(p.getTokenLocation(), "pre_tags are set but post_tags are not set");
  }
```

https://github.com/opensearch-project/OpenSearch/blob/aca2e9de7daaadee77c33e21ff5215c3f1e8600f/server/src/main/java/org/opensearch/search/fetch/subphase/highlight/AbstractHighlighterBuilder.java#L697-L699

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
bzhangam: I think we should throw IllegalArgumentException if invalid custom tags are provided rather than silently ignore it.

Other than validate both pre and post tags should be provided and they should not be empty string is there a way to also validate the tag should be a valid tag that we can process? And also validate the length of the tags to ensure user cannot provide an extremely long string which may break our system?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
junqiu-lei: Since custom tags are provided by users, I now feel we might better not restrict on the format they provided.

Example of existing highlighter with just "pre" and "post" tags:

```
ccurl -X GET "localhost:9200/articles/_search?pretty" -H "Content-Type: application/json" -d'
{
  "query": {
    "match": {
      "content": "machine learning"
    }
  },
  "highlight": {
    "fields": {
      "content": {

      }
    },
    "pre_tags": [
      "pre"
    ],
    "post_tags": [
      "post"
    ]
  }
}'
{
  "took" : 3,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 2,
      "relation" : "eq"
    },
    "max_score" : 0.8263625,
    "hits" : [
      {
        "_index" : "articles",
        "_id" : "1",
        "_score" : 0.8263625,
        "_source" : {
          "content" : "Machine learning is transforming the way we analyze data. Advanced machine learning techniques help us discover patterns."
        },
        "highlight" : {
          "content" : [
            "preMachinepost prelearningpost is transforming the way we analyze data.",
            "Advanced premachinepost prelearningpost techniques help us discover patterns."
          ]
        }
      },
      {
        "_index" : "articles",
        "_id" : "4",
        "_score" : 0.6213684,
        "_source" : {
          "content" : "Machine learning models need proper training data to learn patterns effectively and make accurate predictions."
        },
        "highlight" : {
          "content" : [
            "preMachinepost prelearningpost models need proper training data to learn patterns effectively and make accurate predictions"
          ]
        }
      }
    ]
  }
}
```

heemin32: Is this parameter something we share with core? Can we rely on core code to validate this tag?

yizheliu-amazon: Given core does not have strict validation on value of pre/post tag, and only has validation on existence of pre/post tag, we can skip that as well. It is better to have consistent behavior as core has.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
junqiu-lei: Updated to not have validation here, since it always by default will pass in <em> and </em> from core highlighter.

Since it's customized tags controlled by user, we can't know what is invalid or not.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
junqiu-lei: > Is this parameter something we share with core? Can we rely on core code to validate this tag?

Yes, I updated PR to just rely on core code to validate tags.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
```diff
@@ -60,6 +60,11 @@ public HighlightField highlight(FieldHighlightContext fieldContext) {
             return null;
         }
 
+        String[] preTags = fieldContext.field.fieldOptions().preTags();
+        String[] postTags = fieldContext.field.fieldOptions().postTags();
+        // The pre- and post- tags are provided by the user or defaulted to <em> and </em>
+        semanticHighlighterEngine = semanticHighlighterEngine.withTags(preTags[0], postTags[0]);
```
heemin32, line 66: I think we should not modify the semanticHighlighterEngine which is not thread safe. Also, creating new semanticHighlighterEngine for every highlight request is not efficient. Instead we should just pass the tag to `semanticHighlighterEngine.getHighlightedSentences` to get correct response.

junqiu-lei: Updated to just use method instead of new instance.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
```diff
@@ -60,8 +60,18 @@ public HighlightField highlight(FieldHighlightContext fieldContext) {
...
+        String[] preTags = fieldContext.field.fieldOptions().preTags();
+        String[] postTags = fieldContext.field.fieldOptions().postTags();
+
         // Get highlighted text - allow any exceptions from this call to propagate
-        String highlightedResponse = semanticHighlighterEngine.getHighlightedSentences(modelId, originalQueryText, fieldText);
+        String highlightedResponse = semanticHighlighterEngine.getHighlightedSentences(
+            modelId,
+            originalQueryText,
+            fieldText,
+            preTags[0],
```
heemin32, line 72: I think tag can be a list of string and we need to append them all instead of just using the first string in the list?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
```diff
@@ -24,24 +28,21 @@
...
-    private static final String DEFAULT_PRE_TAG = "<em>";
-    private static final String DEFAULT_POST_TAG = "</em>";
     private static final String MODEL_INFERENCE_RESULT_KEY = "highlights";
     private static final String MODEL_INFERENCE_RESULT_START_KEY = "start";
     private static final String MODEL_INFERENCE_RESULT_END_KEY = "end";
 
-    private final MLCommonsClientAccessor mlCommonsClient;
-    private final QueryTextExtractorRegistry queryTextExtractorRegistry;
+    @NonNull
+    MLCommonsClientAccessor mlCommonsClient;
```
heemin32, line 42: Why remove private final?

junqiu-lei: the `@Value` annotation makes fields private and final implicitly.

heemin32 (approved):

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
vibrantvarun: I think keeping private final would be more convenient for debugging purposes.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
vibrantvarun: Can you declare this preTag[0] to a variable at line 67 and then pass that variable here. Also provide an comment about why you are using first element.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
vibrantvarun: Also this change is only limited to this class. Either we should adapt in the whole project or drop it.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
```diff
@@ -116,15 +117,17 @@ public String getModelId(Map<String, Object> options) {
      * @param modelId The ID of the model to use
      * @param question The search query
      * @param context The document text
+     * @param preTag The pre tag to use for highlighting
+     * @param postTag The post tag to use for highlighting
      * @return Formatted text with highlighting
      */
-    public String getHighlightedSentences(String modelId, String question, String context) {
+    public String getHighlightedSentences(String modelId, String question, String context, String preTag, String postTag) {
```
vibrantvarun, line 118: nit: Can we create a DTO object for the arguments that contains all the variables required for highlighting. Something around HighlighterDTO. Because later when this feature will be enabled for other queries like Hybrid we can reuse the DTO object.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
junqiu-lei: Sure added back.

src/test/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngineTests.java
```diff
@@ -45,7 +45,10 @@ public void setUp() throws Exception {
         super.setUp();
         mlCommonsClientAccessor = mock(MLCommonsClientAccessor.class);
         queryTextExtractorRegistry = new QueryTextExtractorRegistry();
-        highlighterEngine = new SemanticHighlighterEngine(mlCommonsClientAccessor, queryTextExtractorRegistry);
+        highlighterEngine = SemanticHighlighterEngine.builder()
```
vibrantvarun, line 48: QQ: Are we not adding any integ test of for this?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
junqiu-lei: As of now let's keep the current parameter-based approach since it's simpler. We can revisit using a DTO if it becomes more complex. Actually hybrid type query is supported in the semantic highlighter, it's not related to this part code.

vibrantvarun (approved):
LGTM.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterEngine.java
```diff
@@ -24,24 +25,18 @@
  * Engine class for semantic highlighting operations
  */
 @Log4j2
+@Builder
 public class SemanticHighlighterEngine {
     private static final String MODEL_ID_FIELD = "model_id";
-    private static final String DEFAULT_PRE_TAG = "<em>";
```
martin-gaievski, line 29: where defaults tags are defined? You're removing current default ones, are they coming from core?

junqiu-lei: Yes the default will anyway come from core if no customized tag provided by:
```
String[] preTags = fieldContext.field.fieldOptions().preTags();
String[] postTags = fieldContext.field.fieldOptions().postTags();
```

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
```diff
@@ -60,8 +60,18 @@ public HighlightField highlight(FieldHighlightContext fieldContext) {
             return null;
         }
 
+        // The pre- and post- tags are provided by the user or defaulted to <em> and </em>
+        String[] preTags = fieldContext.field.fieldOptions().preTags();
+        String[] postTags = fieldContext.field.fieldOptions().postTags();
```
martin-gaievski, line 65: I think we need to check that pre and post tags are actually the same opening and closing tag. E.g. <em>/</em> is valid, and <br>/</em> is probably not valid. Do we have such check?

junqiu-lei: There were similar conversation, we will keep use core for the validation part to keep existing behavior. https://github.com/opensearch-project/neural-search/pull/1254#discussion_r2025748835

PR #1257, opened 2025-04-04
Fix the edge case when the value of a fieldMap key in ingestDocument is empty String

### LINKED ISSUE #1239: Add validation for values in mapped fields in TextImageEmbeddingProcessor
Today we allow empty strings for mapped field values, but we should not do that, as ingest empty document doesn't add values.

For example, in the following field_map:
```
{
  text: "input_text_field",
  image: "input_image_field"
}
```
We should not allow ingest such document:
```
{
  "input_text_field": "",
  "input_image_field": ""
}

```
Currently there's no any validation for the above mapping field values, we should add one. Also check this [link](https://github.com/opensearch-project/neural-search/pull/1230#discussion_r2000322365) for more details

chloewqg:
### Description
Currently if we create an ingest pipeline like this: 

```
{
  "description": "A text/image embedding pipeline",
  "processors": [
    {
      "text_image_embedding": {
        "model_id": "NXTU_pUBDuT6AlC02kKK",
        "embedding": "vector_embedding",
        "field_map": {
          "text": "input_text_field",
          "image": "input_image_field"
        }
      }
    }
  ]
}
```

And we simulate the pipeline like 

```
{
  "docs": [
    {
      "_index": "nlp-index-2",
      "_id": "1",
      "_source":{
	 "input_text_field": ""
         "input_image_field": ""
      }
    }
  ]
}
```

The MultiModal BedRock model actually would throw status exception

```
{
	"docs": [
		{
			"error": {
				"root_cause": [
					{
						"type": "status_exception",
						"reason": "Error from remote service: {\"message\":\"Malformed input request: #/inputText: expected minLength: 1, actual: 0, #/inputImage: expected minLength: 1, actual: 0, #: required key [requests] not found, #: extraneous key [inputText] is not permitted, #: extraneous key [inputImage] is not permitted, please reformat your input and try again.\"}"
					}
				],
				"type": "status_exception",
				"reason": "Error from remote service: {\"message\":\"Malformed input request: #/inputText: expected minLength: 1, actual: 0, #/inputImage: expected minLength: 1, actual: 0, #: required key [requests] not found, #: extraneous key [inputText] is not permitted, #: extraneous key [inputImage] is not permitted, please reformat your input and try again.\"}"
			}
		}
	]
}
``` 

So we would like to throw exception for such unhandled empty string value edge case.

##### Input

```
{
  "docs": [
    {
      "_index": "nlp-index-2",
      "_id": "1",
      "_source":{
	"input_text_field": "",
        "input_image_field": "iVBORw0K..."
      }
    }
  ]
}
```

##### Output

```
{
	"docs": [
		{
			"error": {
				"root_cause": [
					{
						"reason": "Unsupported format of the field in the document, input_text_field value must be a non-empty string. Currently it is ''. Type is class java.lang.String"
					}
				],
				"reason": "Unsupported format of the field in the document, input_text_field value must be a non-empty string. Currently it is ''. Type is class java.lang.String"
			}
		}
	]
}
```

### Related Issues
Resolves #1239

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
@@ -183,8 +183,26 @@ public class TextImageEmbeddingProcessor extends AbstractProcessor {
             if (!sourceAndMetadataMap.containsKey(originalKey)) {
                 continue;
             }
-            if (!(sourceAndMetadataMap.get(originalKey) instanceof String)) {
-                throw new IllegalArgumentException("Unsupported format of the field in the document, value must be a string");
+
+            Object metaValue = sourceAndMetadataMap.get(originalKey);
+            if (Objects.isNull(metaValue)) {
+                throw new IllegalArgumentException(
+                    String.format(
+                        Locale.getDefault(),
+                        "Unsupported format of the field in the document, %s value must be a non-empty string. Currently it is null",
+                        originalKey
+                    )
+                );
+            } else if ((metaValue instanceof String) == false || Objects.isNull(metaValue) || StringUtils.isBlank((String) metaValue)) {
+                throw new IllegalArgumentException(
+                    String.format(
+                        Locale.getDefault(),
+                        "Unsupported format of the field in the document, %s value must be a non-empty string. Currently it is '%s'. Type is %s",
+                        originalKey,
+                        metaValue,
+                        metaValue.getClass()
+                    )
+                );
             }
             mapWithKnnKeys.put(originalKey, (String) sourceAndMetadataMap.get(originalKey));
         }
--- a/src/test/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessorTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessorTests.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.function.BiConsumer;
 import java.util.function.Supplier;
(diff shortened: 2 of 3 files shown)


src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -183,8 +183,8 @@ Map<String, String> buildMapWithKnnKeyAndOriginalValue(final IngestDocument inge
             if (!sourceAndMetadataMap.containsKey(originalKey)) {
                 continue;
             }
-            if (!(sourceAndMetadataMap.get(originalKey) instanceof String)) {
-                throw new IllegalArgumentException("Unsupported format of the field in the document, value must be a string");
+            if (!(sourceAndMetadataMap.get(originalKey) instanceof String) || ((String) sourceAndMetadataMap.get(originalKey)).isEmpty()) {
```
vibrantvarun, line 186:
```suggestion
            if ((sourceAndMetadataMap.get(originalKey) instanceof String) == false || ((String) sourceAndMetadataMap.get(originalKey)).isEmpty()) {
```

vibrantvarun, line 186:
```suggestion
((String) sourceAndMetadataMap.get(originalKey)).isEmpty()) {
```

It is not a safe type cast. In case of null string it will throw an exception.

src/test/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessorIT.java
```diff
@@ -61,6 +61,15 @@ public void testEmbeddingProcessor_whenIngestingDocumentWithNullMappingValue_the
         expectThrows(ResponseException.class, () -> ingestDocument(INDEX_NAME, INGEST_DOCUMENT.replace("\"passage_text_value\"", "null")));
     }
 
+    public void testEmbeddingProcessor_whenIngestingDocumentWithEmptyMappingValue_thenThrowException() throws Exception {
```
vibrantvarun, line 64: Is it super necessary to add this validation in the form of integ test? Can't it be just a unit test? Doing that will save some build resource.

vibrantvarun:
Also, please add the entry in changelog.

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -183,8 +183,8 @@ Map<String, String> buildMapWithKnnKeyAndOriginalValue(final IngestDocument inge
...
+                throw new IllegalArgumentException("Unsupported format of the field in the document, value must be a non-empty string");
```
martin-gaievski, line 187: better have the log entry for the actual value and its class, it can help in troubleshooting if this happens in prod

src/test/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessorIT.java
martin-gaievski: +1, this can be validated via unit test, integ test with model is a bit expensive to run

martin-gaievski (commented):
Please add the entry for this PR to the changelog

martin-gaievski:
what is the response from other models in the similar scenario?

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
vibrantvarun: Also please use String.format for the exception message.

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
q-andy: `null instanceof String` returns false even if the reference type fo the null is String (see https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html#jls-15.20.2). So the boolean short circuit should make this cast safe. But I also suggest `StringUtils.isNotBlank()` NPEs just to be safe.

heemin32:
Could you share the response after the change as well in the description?

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
```diff
@@ -183,8 +183,29 @@ Map<String, String> buildMapWithKnnKeyAndOriginalValue(final IngestDocument inge
             if (!sourceAndMetadataMap.containsKey(originalKey)) {
                 continue;
             }
-            if (!(sourceAndMetadataMap.get(originalKey) instanceof String)) {
-                throw new IllegalArgumentException("Unsupported format of the field in the document, value must be a string");
+
+            Object metaValue = sourceAndMetadataMap.get(originalKey);
+            if ((metaValue instanceof String) == false || Objects.isNull(metaValue) || StringUtils.isBlank((String) metaValue)) {
+                if (Objects.isNull(metaValue)) {
```
martin-gaievski, line 189: I think this logic can be in a single level if/else, something like:
```
if (Objects.isNull(metaValue)) {
    throw "Currently it is null"
} else if ((metaValue instanceof String) == false || StringUtils.isBlank((String) metaValue)) {
    throw "value must be a non-empty string. Currently it is '%s'. Type is %s"
}
```

src/test/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessorTests.java
```diff
@@ -337,11 +335,9 @@ public void testExecute_mapHasNonStringValue_throwIllegalArgumentException() {
     }
 
     public void testExecute_mapHasEmptyStringValue_throwIllegalArgumentException() {
-        Map<String, String> map1 = ImmutableMap.of("test1", "test2");
-        Map<String, String> map2 = ImmutableMap.of("test3", "   ");
         Map<String, Object> sourceAndMetadata = new HashMap<>();
-        sourceAndMetadata.put("key1", map1);
-        sourceAndMetadata.put("my_text_field", map2);
+        sourceAndMetadata.put("key1", "test2");
+        sourceAndMetadata.put("my_text_field", "");
```
martin-gaievski, line 340: please add check for the error message text

martin-gaievski: do we have a test case for scenario when value is "null"?

src/main/java/org/opensearch/neuralsearch/processor/TextImageEmbeddingProcessor.java
chloewqg: Updated the code structure. Thanks for suggestion!

heemin32 (approved):

Issue #1035, opened 2024-12-18
[Enhancement]Load Model Once and Reuse Across Multiple Tests in BWC Test


heemin32:
BWC tests are unstable with frequent random failures during model auto-loading in cluster upgrades, especially in rolling upgrade BWC tests. The issue arises because each test attempts to load an ML model into memory, increasing the likelihood of auto-loading failures as multiple models need to be redeployed. In reality, we only need three models: sparse embedding, text embedding, and text-image embedding. By loading only these three models and reusing them across tests, we can significantly reduce test failures caused by model auto-redeployment during cluster upgrades. This does not compromise the purpose of the BWC test since we are not validating how well the cluster upgrade handles multiple models in memory but rather focusing on testing the functionality of the neural plugin during cluster upgrade.

```
REPRODUCE WITH: gradlew ':qa:rolling-upgrade:testAgainstTwoThirdsUpgradedCluster' --tests "org.opensearch.neuralsearch.bwc.SemanticSearchIT.testSemanticSearch_E2EFlow" -Dtests.seed=3D4E586DC98C0331 -Dtests.security.manager=false -Dtests.bwc.version=2.12.0 -Dtests.locale=ja-JP -Dtests.timezone=Australia/South -Druntime.java=17
org.opensearch.neuralsearch.bwc.SemanticSearchIT > testSemanticSearch_E2EFlow FAILED
    org.opensearch.client.ResponseException: method [GET], host [http://[::1]:61211], URI [/_cluster/health?wait_for_nodes=3&timeout=60s&wait_for_status=green&cluster_manager_timeout=60s], status line [HTTP/1.1 408 Request Timeout]


Suite: Test class org.opensearch.neuralsearch.bwc.SemanticSearchIT
  2> REPRODUCE WITH: gradlew ':qa:rolling-upgrade:testAgainstTwoThirdsUpgradedCluster' --tests "org.opensearch.neuralsearch.bwc.SemanticSearchIT.testSemanticSearch_E2EFlow" -Dtests.seed=3D4E586DC98C0331 -Dtests.security.manager=false -Dtests.bwc.version=2.12.0 -Dtests.locale=ja-JP -Dtests.timezone=Australia/South -Druntime.java=17
[...]

PR #1276, opened 2025-04-17
Add semantic mapping transformer.

bzhangam:
### Description
Add semantic mapping transformer to transform the mapping for the semantic field.

### Related Issues

### Check List


chloewqg:
General Question, is this semantic field code change here related to what we discussed about the analyzer and tokenizer? https://github.com/opensearch-project/neural-search/pull/1088

src/main/java/org/opensearch/neuralsearch/constants/SemanticInfoFieldConstants.java
```diff
@@ -0,0 +1,118 @@
...
+     * @return Config of the semantic info fields.
+     */
+    public static Map<String, Object> getBaseSemanticInfoConfig(@NonNull final Map<String, Object> embeddingConfig) {
+        final Map<String, Object> chunksConfig = Map.of(
+            TYPE,
+            ObjectMapper.NESTED_CONTENT_TYPE,
+            PROPERTIES,
+            Map.of(CHUNKS_TEXT_FIELD_NAME, Map.of(TYPE, TextFieldMapper.CONTENT_TYPE), CHUNKS_EMBEDDING_FIELD_NAME, embeddingConfig)
+        );
+        return Map.of(PROPERTIES, Map.of(CHUNKS_FIELD_NAME, chunksConfig, MODEL_FIELD_NAME, modelConfig));
```
will-hwang, line 94: is there a reason why we initialize chunksConfig at a method level, but have the modelConfig as class constant? ideally, we should keep them consistent

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,425 @@
...
+        }
+    }
+
+    private List<String> validateModelId(@NonNull final String semanticFieldPath, @NonNull final Map<String, Object> semanticFieldConfig) {
+        final List<String> errors = new ArrayList<>();
+        if (!semanticFieldConfig.containsKey(MODEL_ID)) {
+            errors.add(String.format(Locale.ROOT, "%s is required for the semantic field at %s", MODEL_ID, semanticFieldPath));
+        } else {
+            Object modelId = semanticFieldConfig.get(MODEL_ID);
+            if (!(modelId instanceof String modelIdStr && !modelIdStr.isEmpty())) {
```
will-hwang, line 179: `if (modelId instanceof String == false || modelId.isEmpty())`

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,425 @@
...
+        }
+
+    }
+
+    private void validateSemanticFields(@NonNull final Map<String, Map<String, Object>> semanticFieldPathToConfigMap) {
+        final List<String> errors = new ArrayList<>();
+        for (Map.Entry<String, Map<String, Object>> entry : semanticFieldPathToConfigMap.entrySet()) {
+            final String semanticFieldPath = entry.getKey();
+            final Map<String, Object> semanticFieldConfig = entry.getValue();
+            errors.addAll(validateModelId(semanticFieldPath, semanticFieldConfig));
```
will-hwang, line 165: I'd prefer to throw exceptions directly in `validModelId`, `validateSemanticInfoFieldName` instead of compiling all errors. is there a good reason why we can't directly throw exceptions once encountered?

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,425 @@
...
+            errors.addAll(validateSemanticInfoFieldName(semanticFieldPath, semanticFieldConfig));
+        }
+        if (!errors.isEmpty()) {
+            throw new IllegalArgumentException(String.join("; ", errors));
```
will-hwang, line 169: [nit] It would help to have more descriptive error message beforehand like "Semantic Fields not valid with following errors"

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,425 @@
...
+        final AtomicInteger counter = new AtomicInteger(modelIdToFieldPathMap.size());
+        final AtomicBoolean hasError = new AtomicBoolean(false);
+        final List<String> errors = new ArrayList<>();
+        final Map<String, MLModel> modelIdToConfigMap = new HashMap<>(modelIdToFieldPathMap.size());
+
+        // We can have multiple semantic fields with different model ids, and we should get model config for each model
+        for (String modelId : modelIdToFieldPathMap.keySet()) {
+            mlClientAccessor.getModel(modelId, ActionListener.wrap(mlModel -> {
+                modelIdToConfigMap.put(modelId, mlModel);
+                if (counter.decrementAndGet() == 0) {
```
will-hwang, line 250: why wait til last iteration for this check? can we fail fast as soon as error is encountered?

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,425 @@
...
+            final MLModel mlModel = modelIdToConfigMap.get(modelId);
+            final List<String> fieldPathList = modelIdToFieldPathMap.get(modelId);
+            for (String fieldPath : fieldPathList) {
+                try {
+                    final Map<String, Object> semanticInfoConfig = createSemanticInfoField(mlModel);
+                    final Map<String, Object> fieldConfig = semanticFieldPathToConfigMap.get(fieldPath);
+                    setSemanticInfoField(mappings, fieldPath, fieldConfig.get(SEMANTIC_INFO_FIELD_NAME), semanticInfoConfig);
+                } catch (IllegalArgumentException e) {
+                    throw new IllegalArgumentException(getModifyMappingErrorMessage(fieldPath, e.getMessage()), e);
+                } catch (Exception e) {
```
will-hwang, line 289: will we ever reach this clause?

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,425 @@
...
+        );
+
+    }
+
+    @SuppressWarnings("unchecked")
+    private Map<String, Object> createEmbeddingConfigForTextEmbeddingModel(@NonNull final MLModel modelConfig) {
+        final Map<String, Object> embeddingConfig = getBaseTextEmbeddingConfig();
+        final String modelId = modelConfig.getModelId();
+
+        if (!(modelConfig.getModelConfig() instanceof TextEmbeddingModelConfig textEmbeddingModelConfig)) {
```
will-hwang, line 353:
```
if (modelConfig.getModelConfig() instanceof TextEmbeddingModelConfig == false) {
     TextEmbeddingModelConfig textEmbeddingModelConfig = modelConfig.getModelConfig();
}
```

seems more readable. Applicable to other similar places

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
```diff
@@ -0,0 +1,125 @@
...
+        @NonNull final Map<String, Map<String, Object>> semanticFieldPathToConfigMap
+    ) {
+        for (Map.Entry<String, Object> entry : currentMapping.entrySet()) {
+            final String fieldName = entry.getKey();
+            final Object fieldConfig = entry.getValue();
+
+            // Build the full path for the current field
+            final String fullPath = parentPath.isEmpty() ? fieldName : parentPath + "." + fieldName;
+
+            if (fieldConfig instanceof Map) {
```
will-hwang, line 43: `if (fieldConfig instanceof Map<String, Object>)`

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
```diff
@@ -0,0 +1,125 @@
...
+     */
+    public static Map<String, List<String>> extractModelIdToFieldPathMap(
+        @NonNull final Map<String, Map<String, Object>> semanticFieldPathToConfigMap
+    ) {
+        final Map<String, List<String>> modelIdToFieldPathMap = new HashMap<>();
+        for (Map.Entry<String, Map<String, Object>> entry : semanticFieldPathToConfigMap.entrySet()) {
+            final String fullPath = entry.getKey();
+            final Map<String, Object> fieldConfigMap = entry.getValue();
+            final String modelIdStr = getModelId(fieldConfigMap, fullPath);
+            if (modelIdToFieldPathMap.containsKey(modelIdStr)) {
```
will-hwang, line 82: [nit]
```
if (modelIdToFieldPathMap.containsKey(modelIdStr) == false) {
    modelIdToFieldPathMap.put(modelIdStr, new ArrayList<>());
}
modelIdToFieldPathMap.get(modelIdStr).add(fullPath);
```

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
```diff
@@ -0,0 +1,125 @@
...
+     * @return properties of the mapping
+     */
+    @SuppressWarnings("unchecked")
+    public static Map<String, Object> getProperties(Map<String, Object> mapping) {
+        if (mapping == null) {
+            return null;
+        }
+        // Actions like create index and legacy create/update index template will have the mapping properties under a
+        // _doc key. Other actions like update mapping and create/update index template will not have the _doc layer.
+        if (mapping.containsKey(DOC) && mapping.get(DOC) instanceof Map) {
```
will-hwang, line 112: `if (mapping.containsKey(DOC) && mapping.get(DOC) instanceof Map<String, Object>)`

src/main/java/org/opensearch/neuralsearch/constants/SemanticInfoFieldConstants.java
```diff
@@ -0,0 +1,118 @@
...
+    public static final String MODEL_ID_FIELD_NAME = "id";
+    public static final String MODEL_NAME_FIELD_NAME = "name";
+    public static final String MODEL_TYPE_FIELD_NAME = "type";
+
+    // A parameter to control if we want to make the field searchable.
+    public static final String INDEX_FIELD_NAME = "index";
+
+    // Default model info field config will be text, and we will not index it to save some effort since we don't
+    // expect users want to search against them.
+    private static final Map<String, Object> modelInfoFieldConfig = Map.of(
```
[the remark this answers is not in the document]
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/constants/SemanticInfoFieldConstants.java
```diff
@@ -0,0 +1,118 @@
...
+        TYPE,
+        TextFieldMapper.CONTENT_TYPE,
+        INDEX_FIELD_NAME,
+        Boolean.FALSE
+    );
+    // Default model info config which will have id, name and type. We will use it as the metadata to help us identify
+    // the embedding is generated by what model. This info can be useful when we want to reuse the embedding in the
+    // existing doc to help us ensure the existing embedding was generated by the same model as the current one.
+    private static final Map<String, Object> modelConfig = Map.of(
```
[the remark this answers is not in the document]
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/constants/SemanticInfoFieldConstants.java
bzhangam: Because chunksConfig contains the embeddingConfig which is not a constant. We need to decide it during runtime based on the model config.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
bzhangam: I think this is a better UX which can allow users to address all the errors at one time rather than address one then retry and run into another error.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
bzhangam: The API calls are executed in parallel so we have to wait for all done and then decide how to call back the listener. If we simply call back the listener then the listener can be invoked multiple times.

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
bzhangam: It's not a valid syntax. Cannot do a check like that.

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
bzhangam: same as above cannot check against Map<String, Object>)

bzhangam:
> General Question, is this semantic field code change here related to what we discussed about the analyzer and tokenizer? #1088

Yes. At the end we will also need to support the analyzer and tokenizer for the semantic field.

vibrantvarun:
@bzhangam Can I start reviewing this PR?

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,425 @@
...
+    private String getModifyMappingErrorMessage(@NonNull final String fieldPath, final String cause) {
+        return String.format(Locale.ROOT, "Failed to transform the mapping for the semantic field at %s due to %s", fieldPath, cause);
+    }
+
+    @VisibleForTesting
+    private Map<String, Object> createSemanticInfoField(final @NonNull MLModel modelConfig) {
+        final Map<String, Object> embeddingConfig = switch (modelConfig.getAlgorithm()) {
+            case FunctionName.TEXT_EMBEDDING -> createEmbeddingConfigForTextEmbeddingModel(modelConfig);
+            case FunctionName.SPARSE_ENCODING, FunctionName.SPARSE_TOKENIZE -> getBaseSparseEmbeddingConfig();
+            case FunctionName.REMOTE -> createEmbeddingConfigForRemoteModel(modelConfig);
```
q-andy, line 305: Rather than storing and editing all semantic info field information in a an ongoing map, what do you think about having an object like `SemanticInfoFieldData` or `SemanticInfoFieldConfig` that stores the info as private fields with direct getters and setters, that converts into map at the very end? I'm thinking keeping the information in a `Map<String, Object>` is kind of fragile, especially when the semantic info field is a deeply nested object. Having a dedicated object for it makes it clearer when and how we're mutating it and make things more organized, and consolidates type casts in one place. Maybe all the field validation logic could go in there as well so we can separate the mapping transformation logic from the validation logic.

This helps too if we plan to add more configurability to semantic field in the future (like additional text chunking options, faiss vs hnsw for knn), so its easier to identify where and how we read and write these metadata fields

src/main/java/org/opensearch/neuralsearch/constants/SemanticInfoFieldConstants.java
```diff
@@ -0,0 +1,118 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.constants;
+
+import org.opensearch.index.mapper.ObjectMapper;
+import org.opensearch.index.mapper.RankFeaturesFieldMapper;
+import org.opensearch.index.mapper.TextFieldMapper;
+import org.opensearch.knn.index.mapper.KNNVectorFieldMapper;
+import reactor.util.annotation.NonNull;
```
weijia-aws, line 11: Why don't we use lombok's NonNull? Currently we use lombok instead of reactor in the whole package

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,425 @@
...
+import org.opensearch.index.mapper.MapperService;
+import org.opensearch.index.mapper.MappingTransformer;
+
+import org.opensearch.ml.common.FunctionName;
+import org.opensearch.ml.common.MLModel;
+import org.opensearch.ml.common.model.TextEmbeddingModelConfig;
+import org.opensearch.neuralsearch.constants.SemanticFieldConstants;
+import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
+import org.opensearch.neuralsearch.util.SemanticMappingUtils;
+import reactor.util.annotation.NonNull;
```
weijia-aws, line 20: Same

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,425 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.mappingtransformer;
+
+import com.google.common.annotations.VisibleForTesting;
+import joptsimple.internal.Strings;
```
weijia-aws, line 8: Why do we need this package? Looks like the only usage is `Strings.EMPTY`, which can be replaced by `""`? https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/jdk.internal.opt/share/classes/jdk/internal/joptsimple/internal/Strings.java#L67

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,425 @@
...
+                            "%s cannot be an empty string for the semantic " + "field at %s",
+                            SEMANTIC_INFO_FIELD_NAME,
+                            semanticFieldPath
+                        )
+                    );
+                } else if (semanticInfoFieldNameStr.length() > FIELD_NAME_MAX_LENGTH) {
+                    errors.add(
+                        String.format(
+                            Locale.ROOT,
+                            "%s should not be longer than %d characters for " + "the semantic field at %s",
```
weijia-aws, line 209: Same as above

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,425 @@
...
+                            FIELD_NAME_MAX_LENGTH,
+                            semanticFieldPath
+                        )
+                    );
+                }
+            } else {
+                errors.add(
+                    String.format(
+                        Locale.ROOT,
+                        "%s should be a non-empty string for the semantic field" + " at %s",
```
weijia-aws, line 220: Same

bzhangam:
> @bzhangam Can I start reviewing this PR?

@vibrantvarun Yes. Your feedback is welcome.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,425 @@
...
+            errors.addAll(validateModelId(semanticFieldPath, semanticFieldConfig));
+            errors.addAll(validateSemanticInfoFieldName(semanticFieldPath, semanticFieldConfig));
+        }
+        if (!errors.isEmpty()) {
+            throw new IllegalArgumentException(String.join("; ", errors));
+        }
+    }
+
+    private List<String> validateModelId(@NonNull final String semanticFieldPath, @NonNull final Map<String, Object> semanticFieldConfig) {
+        final List<String> errors = new ArrayList<>();
```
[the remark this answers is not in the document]
bzhangam: yeah we can do that.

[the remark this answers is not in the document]
bzhangam: yeah we can do that.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,425 @@
...
+        final Map<String, List<String>> modelIdToFieldPathMap = SemanticMappingUtils.extractModelIdToFieldPathMap(
+            semanticFieldPathToConfigMap
+        );
+        if (modelIdToFieldPathMap.isEmpty()) {
+            listener.onResponse(null);
+        }
+        final AtomicInteger counter = new AtomicInteger(modelIdToFieldPathMap.size());
+        final AtomicBoolean hasError = new AtomicBoolean(false);
+        final List<String> errors = new ArrayList<>();
+        final Map<String, MLModel> modelIdToConfigMap = new HashMap<>(modelIdToFieldPathMap.size());
```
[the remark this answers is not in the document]
bzhangam: We don't need it anymore.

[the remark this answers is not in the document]
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
bzhangam: I think having a DTO can make the code more maintainable and I can add it.

src/main/java/org/opensearch/neuralsearch/constants/SemanticInfoFieldConstants.java
bzhangam: yeah we can use it

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
bzhangam: To make it more readable. But I'm ok with using `""` directly.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
bzhangam: removed

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
bzhangam: removed

weijia-aws (approved):

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
q-andy: We don't necessarily have to store the fields as nested objects in the DTO, just having them as flat fields on the object that convert to a nested map works as well. My concern is keeping the code readable, if we don't go with DTO, adding some javadocs to the private helpers would be helpful to know what they are validating and populating since we are making lots of deeply nested helper calls which are mutating the same

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,429 @@
...
+            case FunctionName.TEXT_EMBEDDING -> createEmbeddingConfigForTextEmbeddingModel(modelConfig);
+            case FunctionName.SPARSE_ENCODING, FunctionName.SPARSE_TOKENIZE -> getBaseSparseEmbeddingConfig();
+            case FunctionName.REMOTE -> createEmbeddingConfigForRemoteModel(modelConfig);
+            default -> throw new IllegalArgumentException(
+                String.format(
+                    Locale.ROOT,
+                    "The algorithm %s of the model %s is not supported in the semantic field. Supported algorithms: [%s].",
+                    modelConfig.getAlgorithm().name(),
+                    modelConfig.getModelId(),
+                    String.join(",", SUPPORTED_MODEL_ALGORITHMS)
```
yizheliu-amazon, line 316: nit: you may not need to use `String.join` since Set toString() already print elements with comma delimited.

src/test/resources/mapper/mappingWithNestedSemanticFields.json
```diff
@@ -0,0 +1,14 @@
+{
+  "products": {
+    "type": "nested",
+    "properties": {
+      "product_description": {
```
yizheliu-amazon, line 5: Will it work if field name has `dot`, like `product_description.en`?

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,429 @@
...
+                return String.format(
+                    Locale.ROOT,
+                    "%s should be a non-empty string for the semantic field at %s",
+                    SEMANTIC_INFO_FIELD_NAME,
+                    semanticFieldPath
+
+                );
+            }
+        }
+        return null;
```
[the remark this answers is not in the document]
vibrantvarun: Instead we should throw exception if semanticFieldConfig does not contain key.

vibrantvarun (commented):
Only was able to do the half review. Will do more later tonight.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,429 @@
...
+     */
+
+    @Override
+    public void transform(final Map<String, Object> mapping, final TransformContext context, @NonNull final ActionListener<Void> listener) {
+        try {
+            final Map<String, Object> properties = getProperties(mapping);
+            // If there is no property or its format is not valid we simply do nothing and rely on core to validate the
+            // mappings and handle the error.
+            if (properties == null) {
+                listener.onResponse(null);
```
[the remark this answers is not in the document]
junqiu-lei: Add return after this line? Same as other places.

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
```diff
@@ -296,4 +300,12 @@ public Map<String, Mapper.TypeParser> getMappers() {
         }
         return Collections.emptyMap();
     }
+
+    @Override
+    public List<MappingTransformer> getMappingTransformers() {
+        if (FeatureFlagUtil.isEnabled(SEMANTIC_FIELD_ENABLED)) {
```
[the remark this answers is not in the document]
bzhangam: I think I'll remove the FF since now the main branch is for 3.1

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,429 @@
...
+    private void fetchModelAndModifyMapping(
+        @NonNull final Map<String, Map<String, Object>> semanticFieldPathToConfigMap,
+        @NonNull final Map<String, Object> mappings,
+        @NonNull final ActionListener<Void> listener
+    ) {
+        final Map<String, List<String>> modelIdToFieldPathMap = SemanticMappingUtils.extractModelIdToFieldPathMap(
+            semanticFieldPathToConfigMap
+        );
+        if (modelIdToFieldPathMap.isEmpty()) {
+            listener.onResponse(null);
```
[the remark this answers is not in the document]
bzhangam: What empty we can use here? I think null is the only valid thing we can use for a ActionLister<Void>.

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/NeuralQueryEnricherProcessorIT.java
```diff
@@ -67,6 +67,7 @@ public void testNeuralQueryEnricherProcessor_NeuralSparseSearch_E2EFlow() throws
                 loadModel(sparseModelId);
                 sparseEncodingQueryBuilderWithModelId.modelId(sparseModelId);
 
+                waitForClusterHealthGreen(NODES_BWC_CLUSTER);
```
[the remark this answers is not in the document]
bzhangam: We need this because in rolling upgrade test case the node can be unstable. Even we have checked it at beginning the node still can turn into yellow and fail the test. Adding the wait here can help us make the test more stable. This is just a short term workaround and we plan to do some investigation to see how can we improve the tests.

src/main/java/org/opensearch/neuralsearch/constants/SemanticInfoFieldConstants.java
```diff
@@ -0,0 +1,118 @@
...
+import static org.opensearch.neuralsearch.constants.MappingConstants.TYPE;
+
+/**
+ * Constants for semantic info
+ */
+public class SemanticInfoFieldConstants {
+    public static final String KNN_VECTOR_DIMENSION_FIELD_NAME = "dimension";
+    public static final String KNN_VECTOR_METHOD_FIELD_NAME = "method";
+    public static final String KNN_VECTOR_METHOD_ENGINE_FIELD_NAME = "engine";
+    public static final String KNN_VECTOR_METHOD_DEFAULT_ENGINE = "faiss";
```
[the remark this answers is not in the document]
bzhangam: I think we can remove it and let knn plugin to populate the default value.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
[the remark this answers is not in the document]
vibrantvarun: Like whatever object you are expecting in success , make it empty object and send it. Refer in OpenSearch core they do lot of that stuff

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,429 @@
...
+import static org.opensearch.neuralsearch.constants.SemanticFieldConstants.SEMANTIC_INFO_FIELD_NAME;
+import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.KNN_VECTOR_DIMENSION_FIELD_NAME;
+import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.KNN_VECTOR_METHOD_FIELD_NAME;
+import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.KNN_VECTOR_METHOD_SPACE_TYPE_FIELD_NAME;
+import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.getBaseSemanticInfoConfig;
+import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.getBaseSparseEmbeddingConfig;
+import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.getBaseTextEmbeddingConfig;
+import static org.opensearch.neuralsearch.util.SemanticMappingUtils.getProperties;
+
+public class SemanticMappingTransformer implements MappingTransformer {
```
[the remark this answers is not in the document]
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,429 @@
...
+        FunctionName.SPARSE_ENCODING.name(),
+        FunctionName.SPARSE_TOKENIZE.name()
+    );
+    public final static Set<String> SUPPORTED_REMOTE_MODEL_TYPES = Set.of(
+        FunctionName.TEXT_EMBEDDING.name(),
+        FunctionName.SPARSE_ENCODING.name(),
+        FunctionName.SPARSE_TOKENIZE.name()
+    );
+
+    private final int FIELD_NAME_MAX_LENGTH = 512;
```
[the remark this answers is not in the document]
bzhangam: After thinking twice I decide to remove this validation since we are not validating the length of any field name.

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
```diff
@@ -0,0 +1,123 @@
...
+
+import static org.opensearch.neuralsearch.constants.MappingConstants.DOC;
+import static org.opensearch.neuralsearch.constants.MappingConstants.PROPERTIES;
+
+/**
+ * A util class to help process mapping with semantic field
+ */
+public class SemanticMappingUtils {
+    /**
+     * It will recursively traverse the mapping to collect the full path to field config map for semantic fields
```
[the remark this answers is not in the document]
bzhangam: I think we can very efficiently go through the mapping to collect the semantic fields so don't think we need to add the limitation. We can rely on the core to validate the mapping size if it's too large.

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
```diff
@@ -0,0 +1,123 @@
...
+            if (fieldConfig instanceof Map) {
+                final Map<String, Object> fieldConfigMap = (Map<String, Object>) fieldConfig;
+
+                if (isSemanticField(fieldConfigMap)) {
+                    semanticFieldPathToConfigMap.put(fullPath, fieldConfigMap);
+                }
+
+                // If it's an object field, recurse into the sub fields
+                if (fieldConfigMap.containsKey(MappingConstants.PROPERTIES)) {
+                    collectSemanticField(
```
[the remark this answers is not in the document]
bzhangam: I think this is an overkill here since in most of cases our index mapping is not extremely large and our logic is pretty simple that we just need to check the field type.

While parallelStream() can improve performance by using multiple threads, it isn't free — there’s overhead involved in managing those threads. This means that for small workloads, the cost of parallelism can actually outweigh the benefits.

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
```diff
@@ -0,0 +1,123 @@
...
+                        (Map<String, Object>) fieldConfigMap.get(MappingConstants.PROPERTIES),
+                        fullPath,
+                        semanticFieldPathToConfigMap
+                    );
+                }
+            }
+        }
+    }
+
+    private static boolean isSemanticField(Map<String, Object> fieldConfigMap) {
```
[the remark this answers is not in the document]
bzhangam: I think we can simply do
```
final Object type = fieldConfigMap.get(MappingConstants.TYPE);
return SemanticFieldMapper.CONTENT_TYPE.equals(type);
```

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
```diff
@@ -0,0 +1,123 @@
...
+     */
+    @SuppressWarnings("unchecked")
+    public static Map<String, Object> getProperties(Map<String, Object> mapping) {
+        if (mapping == null) {
+            return null;
+        }
+        // Actions like create index and legacy create/update index template will have the mapping properties under a
+        // _doc key. Other actions like update mapping and create/update index template will not have the _doc layer.
+        if (mapping.containsKey(DOC) && mapping.get(DOC) instanceof Map) {
+            Map<String, Object> doc = (Map<String, Object>) mapping.get(DOC);
```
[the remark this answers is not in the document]
bzhangam: yes.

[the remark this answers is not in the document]
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
```diff
@@ -0,0 +1,123 @@
...
+            if (doc.containsKey(PROPERTIES) && doc.get(PROPERTIES) instanceof Map) {
+                return (Map<String, Object>) doc.get(PROPERTIES);
+            } else {
+                return null;
+            }
+        } else if (mapping.containsKey(PROPERTIES) && mapping.get(PROPERTIES) instanceof Map) {
+            return (Map<String, Object>) mapping.get(PROPERTIES);
+        } else {
```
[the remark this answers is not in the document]
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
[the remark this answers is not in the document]
bzhangam: We only can use null for ActionListener(Void)

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,429 @@
...
+            final Map<String, Object> properties = getProperties(mapping);
+            // If there is no property or its format is not valid we simply do nothing and rely on core to validate the
+            // mappings and handle the error.
+            if (properties == null) {
+                listener.onResponse(null);
+                return;
+            }
+
+            final Map<String, Map<String, Object>> semanticFieldPathToConfigMap = new HashMap<>();
+            final String rootPath = "";
```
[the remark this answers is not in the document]
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,429 @@
...
+            SemanticMappingUtils.collectSemanticField(properties, rootPath, semanticFieldPathToConfigMap);
```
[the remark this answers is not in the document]
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,429 @@
...
+
+    }
+
+    private void validateSemanticFields(@NonNull final Map<String, Map<String, Object>> semanticFieldPathToConfigMap) {
+        final List<String> errors = new ArrayList<>();
+        for (Map.Entry<String, Map<String, Object>> entry : semanticFieldPathToConfigMap.entrySet()) {
+            final String semanticFieldPath = entry.getKey();
+            final Map<String, Object> semanticFieldConfig = entry.getValue();
+            final String validateModelIdError = validateModelId(semanticFieldPath, semanticFieldConfig);
+            final String validateSemanticInfoFieldNameError = validateSemanticInfoFieldName(semanticFieldPath, semanticFieldConfig);
```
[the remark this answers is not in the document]
bzhangam: We will rely on the field mapper to validate. Knn field mapper will validate dimension and space_type.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,429 @@
...
+            if (!(modelId instanceof String modelIdStr) || modelIdStr.isEmpty()) {
+                return String.format(
+                    Locale.ROOT,
+                    "%s should be a non-empty string for the semantic field at %s",
+                    MODEL_ID,
+                    semanticFieldPath
+                );
+            }
+        }
+        return null;
```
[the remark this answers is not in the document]
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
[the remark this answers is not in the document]
bzhangam: This is an optional field so if it's not there we don't need to validate it.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    private void fetchModelAndModifyMapping(
+        @NonNull final Map<String, Map<String, Object>> semanticFieldPathToConfigMap,
+        @NonNull final Map<String, Object> mappings,
+        @NonNull final ActionListener<Void> listener
+    ) {
+        final Map<String, List<String>> modelIdToFieldPathMap = SemanticMappingUtils.extractModelIdToFieldPathMap(
```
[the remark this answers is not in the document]
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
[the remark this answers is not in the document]
bzhangam: We don't expect to return any object that's why we use null here. We also have some existing example doing it this way. 
[example in core](https://github.com/opensearch-project/OpenSearch/blob/2cde1214cbc8bafac1193a03d98784a35f2e1e1d/server/src/main/java/org/opensearch/action/support/replication/TransportWriteAction.java#L417)
[example in neural](https://github.com/opensearch-project/neural-search/blob/c5625db32fdb66c2cb820b626a8cc702335769cb/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java#L518)

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,429 @@
...
+        final List<String> errors = new ArrayList<>();
+        final Map<String, MLModel> modelIdToConfigMap = new ConcurrentHashMap<>(modelIdToFieldPathMap.size());
+
+        // We can have multiple semantic fields with different model ids, and we should get model config for each model
+        for (String modelId : modelIdToFieldPathMap.keySet()) {
+            mlClientAccessor.getModel(modelId, ActionListener.wrap(mlModel -> {
+                modelIdToConfigMap.put(modelId, mlModel);
+                if (counter.decrementAndGet() == 0) {
+                    try {
+                        if (hasError.get()) {
```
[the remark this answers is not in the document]
bzhangam: Yes. Since we are using multiple threads.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
bzhangam: ok Let me and a builder to own the work to create the semantic info fields config.

src/test/resources/mapper/mappingWithNestedSemanticFields.json
bzhangam: Good catch. It will not work. I think I need to unflatten the mapping first.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,429 @@
...
+                    String.join(",", SUPPORTED_MODEL_ALGORITHMS)
+                )
+            );
+        };
+
+        return getBaseSemanticInfoConfig(embeddingConfig);
+    }
+
+    private Map<String, Object> createEmbeddingConfigForRemoteModel(@NonNull final MLModel modelConfig) {
+        final String modelType = modelConfig.getModelConfig().getModelType();
```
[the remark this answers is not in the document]
bzhangam: ok

junqiu-lei (approved):
LGTM.

vibrantvarun:
Codcov check is failing. Please increase the coverage.

vibrantvarun (approved):

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
```diff
@@ -0,0 +1,218 @@
...
+                String.format(
+                    Locale.ROOT,
+                    "Failed to parse the all_config of the model %s. Invalid all_config: %s",
+                    modelId,
+                    textEmbeddingModelConfig.getAllConfig()
+                )
+            );
+        }
+        final Object spaceTypeObject = allConfigMap.get(KNN_VECTOR_METHOD_SPACE_TYPE_FIELD_NAME);
+        if (!(spaceTypeObject instanceof String spaceTypeString)) {
```
vibrantvarun, line 176:
```suggestion
        if ((spaceTypeObject instanceof String spaceTypeString) == false) {
```

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -0,0 +1,377 @@
...
+        }
+    }
+
+    private String validateModelId(@NonNull final String semanticFieldPath, @NonNull final Map<String, Object> semanticFieldConfig) {
+        Object modelId = semanticFieldConfig.get(SemanticFieldConstants.MODEL_ID);
+        if (modelId == null) {
+            return String.format(Locale.ROOT, "%s is required for the semantic field at %s", MODEL_ID, semanticFieldPath);
+        }
+
+        if (!(modelId instanceof String modelIdStr) || modelIdStr.isEmpty()) {
```
vibrantvarun, line 177:
```suggestion
        if ((modelId instanceof String modelIdStr) == false || modelIdStr.isEmpty()) {
```

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
```diff
@@ -0,0 +1,139 @@
...
+                modelIdToFieldPathMap.put(modelIdStr, new ArrayList<>());
+            }
+            modelIdToFieldPathMap.get(modelIdStr).add(fullPath);
+        }
+        return modelIdToFieldPathMap;
+    }
+
+    private static String getModelId(@NonNull final Map<String, Object> fieldConfigMap, @NonNull final String fullPath) {
+        final Object modelId = fieldConfigMap.get(SemanticFieldConstants.MODEL_ID);
+        if (!(modelId instanceof String)) {
```
vibrantvarun, line 99:
```suggestion
        if ((modelId instanceof String)==false) {
```

vibrantvarun:
Please address nit.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
bzhangam: ok

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/NeuralQueryEnricherProcessorIT.java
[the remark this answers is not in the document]
heemin32: I don't think adding this line would make test be stable. I believe [this](https://github.com/opensearch-project/neural-search/issues/1035) might make the bwc test to be stable.

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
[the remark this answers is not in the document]
heemin32: There was a campaign to avoid recursion in OpenSearch. Even if mapping size is limited by core, the depth of recursion can be large. Either we need to change the code to be iterative or need a depth limitation like 1000. https://github.com/opensearch-project/OpenSearch/pull/14086

PR #1291, opened 2025-04-28
Return bad request for invalid stat parameters in stats API

q-andy:
### Description
Fixes bug in parsing stat names while calling the stats API. Currently it will log every stat specified when API is called. We change this to return a 400 bad request instead.

Technically this could be a breaking change, but only for users that are purposefully calling the API with invalid stat names, which doesn't have any valid use case.

Example 400:
```
GET /_plugins/_neural/stats/text_embedding
{
	"error": {
		"root_cause": [
			{
				"reason": "request [/_plugins/_neural/stats/text_embedding] contains unrecognized stat: [text_embedding] -> did you mean [text_embedding_executions]?"
			}
		],
		"reason": "request [/_plugins/_neural/stats/text_embedding] contains unrecognized stat: [text_embedding] -> did you mean [text_embedding_executions]?"
	},
	"status": 400
}
```
### Related Issues
n/a

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
+++ b/src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
@@ -8,6 +8,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import lombok.AllArgsConstructor;
 import lombok.extern.log4j.Log4j2;
+import org.opensearch.common.util.set.Sets;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.neuralsearch.settings.NeuralSearchSettingsAccessor;
 import org.opensearch.neuralsearch.stats.NeuralStatsInput;
@@ -23,6 +24,7 @@ import org.opensearch.transport.client.node.NodeClient;
 
 import java.util.Arrays;
 import java.util.EnumSet;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Optional;
@@ -38,6 +40,16 @@ import static org.opensearch.neuralsearch.plugin.NeuralSearch.NEURAL_BASE_URI;
 @Log4j2
 @AllArgsConstructor
 public class RestNeuralStatsAction extends BaseRestHandler {
+    /**
+     * Path parameter name for specified stats
+     */
+    public static final String STAT_PARAM = "stat";
+
+    /**
+     * Path parameter name for specified node ids
+     */
+    public static final String NODE_ID_PARAM = "nodeId";
+
     /**
      * Query parameter name to request flattened stat paths as keys
      */
@@ -79,7 +91,7 @@ public class RestNeuralStatsAction extends BaseRestHandler {
         new Route(RestRequest.Method.GET, NEURAL_BASE_URI + "/stats/{stat}")
     );
 
-    private static final Set<String> RESPONSE_PARAMS = ImmutableSet.of("nodeId", "stat", INCLUDE_METADATA_PARAM, FLATTEN_PARAM);
+    private static final Set<String> RESPONSE_PARAMS = ImmutableSet.of(NODE_ID_PARAM, STAT_PARAM, INCLUDE_METADATA_PARAM, FLATTEN_PARAM);
 
     /**
      * Validates a param string if its under the max length and matches simple string pattern
@@ -143,7 +155,7 @@ public class RestNeuralStatsAction extends BaseRestHandler {
         NeuralStatsInput neuralStatsInput = new NeuralStatsInput();
 
         // Parse specified nodes
-        Optional<String[]> nodeIds = splitCommaSeparatedParam(request, "nodeId");
+        Optional<String[]> nodeIds = splitCommaSeparatedParam(request, NODE_ID_PARAM);
         if (nodeIds.isPresent()) {
             // Ignore node ids that don't pattern match
             List<String> validFormatNodeIds = Arrays.stream(nodeIds.get()).filter(this::isValidNodeId).toList();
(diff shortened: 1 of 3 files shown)


weijia-aws (approved):

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
```diff
@@ -167,7 +168,7 @@ private NeuralStatsInput createNeuralStatsInputFromRequestParams(RestRequest req
         }
 
         // Process requested stats
-        boolean anyStatAdded = processRequestedStats(stats.get(), neuralStatsInput);
+        boolean anyStatAdded = processRequestedStats(request, stats.get(), neuralStatsInput);
```
yizheliu-amazon, line 171: I think you may simplify code a little bit by combining line 161 to 176  into `void processRequestedStats(RestRequest request, NeuralStatsInput neuralStatsInput)`. 

Basically, you can change `void processRequestedStats(RestRequest request, NeuralStatsInput neuralStatsInput)` to do below steps:
1. get `Optional<String[]> stats` from request
2. 
    - 2a: if stats is not present, add all stats to neuralStatsInput and return
    - 2b: if stats is present, continue
3. Validate stats and add stats to `neuralStatsInput` as same as current logic
    -3a: Throw exception if invalidStatNames is not empty
    -3b: Call if necessary addAllStats(neuralStatsInput);
6.

q-andy: Makes sense, I think we can skip the 3b `addAllStats` call as well. When we start validating, we will either be adding specific requested stats (meaning we do not want to add all) or we will be throwing an exception because there are invalid stats (meaning we do not want to add all).

src/test/java/org/opensearch/neuralsearch/rest/RestNeuralStatsActionIT.java
```diff
@@ -120,6 +120,12 @@ public void test_statsFiltering() throws Exception {
         assertNotNull(getNestedValue(stats, InfoStatName.TEXT_EMBEDDING_PROCESSORS.getFullPath()));
     }
 
+    public void test_statsFiltering_invalidThrows() throws Exception {
```
martin-gaievski, line 123: can you please change approach with testing from integ test to unit tests? We have way too many integ tests, and they are expensive. For such edge case and failure scenario unit test should do the job.

q-andy: Sure. Currently text embedding executions is the only event stat. As I add more stats I can refactor the integ tests to not use text embedding models as well so we can avoid model loading too.

src/test/java/org/opensearch/neuralsearch/rest/RestNeuralStatsActionTests.java
```diff
@@ -0,0 +1,130 @@
...
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class RestNeuralStatsActionTests extends InferenceProcessorTestCase {
+    private RestNeuralStatsAction restNeuralStatsAction;
```
martin-gaievski, line 45: why this is class level variable? if it's specific only for test method then it should be scoped at method level

src/test/java/org/opensearch/neuralsearch/rest/RestNeuralStatsActionTests.java
```diff
@@ -0,0 +1,130 @@
...
+
+    NodeClient client;
```
martin-gaievski, line 47: nit: this can be private

src/test/java/org/opensearch/neuralsearch/rest/RestNeuralStatsActionTests.java
q-andy: Sure I can remove, other tests have them at class level so they can factor out common initialization or cleanup in the setUp and tearDown methods. In this case, different some tests require different setup and there's no special cleanup so it's not necessary.

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
```diff
@@ -157,47 +158,46 @@ private NeuralStatsInput createNeuralStatsInputFromRequestParams(RestRequest req
...
-            return neuralStatsInput;
-        }
+        return neuralStatsInput;
+    }
 
-        // Process requested stats
-        boolean anyStatAdded = processRequestedStats(stats.get(), neuralStatsInput);
+    private void processStatsRequestParameters(RestRequest request, NeuralStatsInput neuralStatsInput) {
+        // Determine which stat names to retrieve based on user parameters
+        Optional<String[]> optionalStats = splitCommaSeparatedParam(request, "stat");
```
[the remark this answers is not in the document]
junqiu-lei: I see it being used in multiple places

junqiu-lei:
Could you please add the updated response example in PR description?

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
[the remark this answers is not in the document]
q-andy: sure, I factored out `nodeIds` as well

q-andy:
@junqiu-lei added the example response, since we use `unrecognized()` from core. it auto suggests actual stat names based on the invalid stat name which is a nice benefit.

`request [/_plugins/_neural/stats/text_embedding] contains unrecognized stat: [text_embedding] -> did you mean [text_embedding_executions]?`

junqiu-lei (approved):
LGTM.

martin-gaievski (approved):

Issue #1303, opened 2025-04-30
[BUG] Hybrid Search not working with Document Level Security


PaulRCampbell:
Hybrid Search queries fail when authenticating with an internal user that has Document Level Security enabled.

Steps to reproduce the behaviour:

_this is a rather contrived example that I have used for testing purposes_

create a role with a dls filter:

```
{"term": { "divisionId": "448385216"}}
```

authenticate with an internal user assigned to that role and then run a hybrid search query

```
GET my_test_index/_search?search_pipeline=hybrid-search-pipeline
{
  "query": {
    "hybrid": {
      "queries": [
        {
          "bool": {
            "filter": [
              {
                "term": {
                  "organisationId": 1
                }
              }
            ]
          }
        },
        {
          "neural": {
            "label_embedding.paraphrase-multilingual-MiniLM-L12-v2": {
              "query_text": "document",
              "model_id": "Cf9VfZYBPcPsA6UT332K",
              "k": "5",
              "filter": {
                "term": {
                  "organisationId": 1
                }
              }
            }
          }
        }
      ]
    }
  },
  "_source": ["label", "organisationId", "esType", "status", "divisionId"]
}

```
returns error

```
{
  "error": {
    "root_cause": [
      {
        "type": "class_cast_exception",
        "reason": "class org.apache.lucene.search.ConstantScoreQuery cannot be cast to class org.opensearch.neuralsearch.query.HybridQuery (org.apache.lucene.search.ConstantScoreQuery is in unnamed module of loader 'app'; org.opensearch.neuralsearch.query.HybridQuery is in unnamed module of loader java.net.FactoryURLClassLoader @57435801)"
      },
      {
        "type": "illegal_argument_exception",
        "reason": "hybrid query must be a top level query and cannot be wrapped into other queries"
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
        "index": "my_test_index",
        "node": "j0xoYCCsTh2Jjoe_SNqmdQ",
        "reason": {
          "type": "class_cast_exception",
[...]

PR #1305, opened 2025-05-02
add validation for invalid nested hybrid query

### LINKED ISSUE #1108: [BUG] Should not allow the hybrid query to be wrapped by another hybrid query.
### What is the bug?
The hybrid query should be the top query but we miss the validation for the case that a hybrid query can be wrapped by another hybrid query.

1. Follow this [doc](https://opensearch.org/docs/latest/search-plugins/neural-search-tutorial/) to set up the hybrid search.
2. Do query like below:
```{
   "_source":{
      "excludes":[
         "passage_embedding"
      ]
   },
   "query":{
      "hybrid":{
         "queries":[
            {
               "hybrid":{
                  "queries":[
                     {
                        "neural":{
                           "passage_embedding":{
                              "query_text":"wild west",
                              "model_id":"<your_model_id>",
                              "k":2
                           }
                        }
                     },
                     {
                        "match_phrase":{
                           "id":"1775029934.jpg"
                        }
                     }
                  ]
               }
            },
            {
               "match_phrase":{
                  "id":"1775029934.jpg"
               }
            }
         ]
      }
   }
}`

We should see an error saying the hybrid query should be the top query and cannot be nested in another hybrid query.

OS: macOS v15.2
OpenSearch: main branch code

N/A

N/A

### PR DESCRIPTION
**Pull request by will-hwang**
### Description
add validation for invalid nested hybrid queries by throwing parsing exception

Example: 
```
"query":{
      "hybrid":{
         "queries":[
            {
               "hybrid":{
                  "queries":[
                     {
                        "neural":{
                           "passage_embedding":{
                              "query_text":"wild west",
                              "model_id":"<your_model_id>",
                              "k":2
                           }
                        }
                     },
                     {
                        "match_phrase":{
                           "id":"1775029934.jpg"
                        }
                     }
                  ]
               }
            },
            {
               "match_phrase":{
                  "id":"1775029934.jpg"
               }
            }
         ]
      }
   }
```
Exception thrown:
`hybrid query cannot be nested in another hybrid query
`
### Related Issues
Resolves #1108 

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Bug Fixes
 - Fix score value as null for single shard when sorting is not done on score field ([#1277](https://github.com/opensearch-project/neural-search/pull/1277))
 - Return bad request for stats API calls with invalid stat names instead of ignoring them ([#1291](https://github.com/opensearch-project/neural-search/pull/1291))
-- add validation for invalid nested hybrid query ([#1305](https://github.com/opensearch-project/neural-search/pull/1305))
+- Add validation for invalid nested hybrid query ([#1305](https://github.com/opensearch-project/neural-search/pull/1305))
 ### Infrastructure
 
 ### Documentation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Bug Fixes
 - Fix score value as null for single shard when sorting is not done on score field ([#1277](https://github.com/opensearch-project/neural-search/pull/1277))
 - Return bad request for stats API calls with invalid stat names instead of ignoring them ([#1291](https://github.com/opensearch-project/neural-search/pull/1291))
-
+- add validation for invalid nested hybrid query ([#1305](https://github.com/opensearch-project/neural-search/pull/1305))
 ### Infrastructure
 
 ### Documentation
--- a/src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
+++ b/src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
@@ -61,6 +61,7 @@ public class HybridQueryPhaseSearcher extends QueryPhaseSearcherWrapper {
             return super.searchWith(searchContext, searcher, query, collectors, hasFilterCollector, hasTimeout);
         } else {
             Query hybridQuery = extractHybridQuery(searchContext, query);
+            validateHybridQuery((HybridQuery) hybridQuery);
             QueryPhaseSearcher queryPhaseSearcher = getQueryPhaseSearcher(searchContext);
             queryPhaseSearcher.searchWith(searchContext, searcher, hybridQuery, collectors, hasFilterCollector, hasTimeout);
             // we decide on rescore later in collector manager
@@ -151,6 +152,14 @@ public class HybridQueryPhaseSearcher extends QueryPhaseSearcherWrapper {
         if (query instanceof DisjunctionMaxQuery) {
             for (Query disjunct : (DisjunctionMaxQuery) query) {
                 validateNestedDisJunctionQuery(disjunct, level - 1);
+            }
+        }
+    }
+
+    private void validateHybridQuery(final HybridQuery query) {
+        for (Query innerQuery : query) {
+            if (innerQuery instanceof HybridQuery) {
+                throw new IllegalArgumentException("hybrid query cannot be nested in another hybrid query");
             }
         }
     }
--- a/src/test/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcherTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcherTests.java
@@ -519,6 +519,103 @@ public class HybridQueryPhaseSearcherTests extends OpenSearchQueryTestCase {
     }
 
     @SneakyThrows
+    public void testWrappedHybridQuery_whenHybridNestedInHybrid_thenFail() {
+        HybridQueryPhaseSearcher hybridQueryPhaseSearcher = new HybridQueryPhaseSearcher();
+        QueryShardContext mockQueryShardContext = mock(QueryShardContext.class);
+        when(mockQueryShardContext.index()).thenReturn(dummyIndex);
+        TextFieldMapper.TextFieldType fieldType = (TextFieldMapper.TextFieldType) createMapperService().fieldType(TEXT_FIELD_NAME);
+        when(mockQueryShardContext.fieldMapper(eq(TEXT_FIELD_NAME))).thenReturn(fieldType);
+        MapperService mapperService = mock(MapperService.class);
+        when(mapperService.hasNested()).thenReturn(false);
+
+        Directory directory = newDirectory();
+        IndexWriter w = new IndexWriter(directory, newIndexWriterConfig(new MockAnalyzer(random())));
+        FieldType ft = new FieldType(TextField.TYPE_NOT_STORED);
+        ft.setIndexOptions(random().nextBoolean() ? IndexOptions.DOCS : IndexOptions.DOCS_AND_FREQS);
+        ft.setOmitNorms(random().nextBoolean());
+        ft.freeze();
+        int docId1 = RandomizedTest.randomInt();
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java [hunk 1]
```diff
@@ -225,7 +225,14 @@ public static HybridQueryBuilder fromXContent(XContentParser parser) throws IOEx
                                 String.format(Locale.ROOT, "Number of sub-queries exceeds maximum supported by [%s] query", NAME)
                             );
                         }
-                        queries.add(parseInnerQueryBuilder(parser));
+                        QueryBuilder query = parseInnerQueryBuilder(parser);
```
bzhangam, line 228: I think this only handle the case when an hybrid query is nested in another hybrid query immediately. What if the hybrid query is nested in a bool query and the bool query is nested in another hybrid query?

@martin-gaievski Want to get your opinion on what we actually want to block. Do we actually want to ensure the hybrid query is the top level query or one query can only have only one hybrid query?

I think there are couple use cases that we do allow user to wrap a hybrid query in another query. e.g. A bool query can wrap a hybrid query. So not sure how we define the "top level query"? Besides I think today user also can use the hybrid query as a filter which may be a valid use case?

I think we may want to think through what use cases we want to block then we can consider is it makes more sense we should do this validation by checking the whole query structure rather than simply check it here.

martin-gaievski: as of now we cannot allow hybrid query to be wrapped into any other query. Main reason being current design, where score collection and normalization as split between query phase/shard level and search pipeline/normalization processor. This means that separate by-sub-query scores must be untouched after the query phase, so those scores are transported to pipeline processor. If there will be any top level query then hybrid score will be changed to fit format of that query, this means corrupted input for the normalization processor.

will-hwang: @martin-gaievski do you think preventing immediately nested hybrid is enough (in this PR), or should we recursively check for all inner queries?

will-hwang: https://docs.opensearch.org/docs/latest/query-dsl/compound/index/

@bzhangam 
In addition to this PR, we should make additional changes in OS core to prevent nested hybrid queries in other compound query types

will-hwang: After discussion, below are our options based on different use cases
Case 1: We do not want hybrid query as nested query in any/some query types
     1. pass in as parameter `querySource` from core to hybrid query builder add a check in hybrid query for validity
     2. add a check in core directly to prevent nested hybrid queries in any query type

Case 2: We want to check strictly for hybrid query nested inside another hybrid query 
     1. check inside hybrid query builder (addressed in this PR)


@martin-gaievski what are your thoughts?

will-hwang: seems like we do want the support for nested hybrid query inside bool based on the implementation here: https://github.com/opensearch-project/neural-search/pull/498/files#diff-64ab60781cf1e9f7963a415406f49283b26945beaa46b7042e7cc5044d8d95b1R81

will-hwang: as for hybrid query inside bool query, i believe these [lines](https://github.com/opensearch-project/neural-search/blob/0bb70530f8f2f31f8a26ca05241d64027f4461c0/src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java#L78-L90) should handle the case, but when run locally, the exception is not thrown.

will-hwang: sounds good to me

will-hwang: @owaiskazi19 As you mentioned [here](https://github.com/opensearch-project/neural-search/issues/1303#issuecomment-2868174144), hybrid query is lost due to DSL wrapping, so nested hybrid would not be caught in the validateQuery method above. 

we need to either fix the unwrapping or add a check directly in hybrid query builder.

owaiskazi19: yeah that's when DSL is enable but in this case that won't be an issue, right?

will-hwang: @owaiskazi19 not an issue, but it can't be inside the same `validateQuerymethod`. Created a separate validate method for nested hybrid. please take a look at the latest revision

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -155,6 +156,16 @@ private void validateNestedDisJunctionQuery(final Query query, final int level)
         }
     }
 
+    private void validateHybridQuery(final Query query) {
+        if (query instanceof HybridQuery) {
+            for (Query innerQuery : (HybridQuery) query) {
```
bzhangam, line 161: How can we loop a HybridQuery? You mean the inner queries of the hybrid query?

will-hwang: yes, this loop checks for inner queries in hybrid query

martin-gaievski:
can we have this logic as part of existing method that does similar check - [extractHybridQuery](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java#L78-L90)?

will-hwang:
> can we have this logic as part of existing method that does similar check - [extractHybridQuery](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java#L78-L90)?

@martin-gaievski we can, but wouldn't it be cleaner to check for nested hybrid after hybrid is extracted from the `extractHybridQuery` method?

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -155,6 +156,16 @@ private void validateNestedDisJunctionQuery(final Query query, final int level)
         }
     }
 
+    private void validateHybridQuery(final Query query) {
+        if (query instanceof HybridQuery) {
```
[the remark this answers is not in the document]
will-hwang: the check is already done [here](https://github.com/opensearch-project/neural-search/pull/1305/files#diff-64ab60781cf1e9f7963a415406f49283b26945beaa46b7042e7cc5044d8d95b1R59).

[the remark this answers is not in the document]
yizheliu-amazon: above link is not working for me. If this check is not necessary, can we skip the `if`  here?

yizheliu-amazon (approved):

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -155,6 +156,14 @@ private void validateNestedDisJunctionQuery(final Query query, final int level)
         }
     }
 
+    private void validateHybridQuery(final Query query) {
```
heemin32, line 159:
```suggestion
    private void validateHybridQuery(final HybridQuery query) {
```

heemin32 (approved):
LGTM. Thanks.

CHANGELOG.md
```diff
@@ -19,7 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Bug Fixes
 - Fix score value as null for single shard when sorting is not done on score field ([#1277](https://github.com/opensearch-project/neural-search/pull/1277))
 - Return bad request for stats API calls with invalid stat names instead of ignoring them ([#1291](https://github.com/opensearch-project/neural-search/pull/1291))
-
+- add validation for invalid nested hybrid query ([#1305](https://github.com/opensearch-project/neural-search/pull/1305))
```
owaiskazi19, line 22:
```suggestion
- Add validation for invalid nested hybrid query ([#1305](https://github.com/opensearch-project/neural-search/pull/1305))
```

owaiskazi19 (approved):
LGTM!

junqiu-lei (approved):

PR #1308, opened 2025-05-06
Add TextChunkingProcessor stats

q-andy:
### Description
Add stats for TextChunkingProcessor. Also adjust stat name unit tests to check for terminal paths.

Event stats:
- `text_chunking_executions`
- `text_chunking_delimiter_executions`
- `text_chunking_fixed_length_executions`

Info stats:
- `text_chunking_processors`
- `text_chunking_fixed_length_processors`
- `text_chunking_delimiter_processors`

```
{
	"_nodes": {
		"total": 1,
		"successful": 1,
		"failed": 0
	},
	"cluster_name": "integTest",
	"info": {
		"cluster_version": "3.1.0",
		"processors": {
			"ingest": {
				"text_chunking_delimiter_processors": 0,
				"text_chunking_fixed_length_processors": 0,
				"text_embedding_processors_in_pipelines": 0,
				"text_chunking_processors": 0
			}
		}
	},
	"all_nodes": {
		"processors": {
			"ingest": {
				"text_chunking_delimiter_executions": 0,
				"text_chunking_executions": 0,
				"text_embedding_executions": 0,
				"text_chunking_fixed_length_executions": 0
			}
		}
	},
	...
}
```

### Related Issues

### Check List
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md). <- Since we are backfilling stats for previous features, we will add them all to the spec and documentation near 3.1 release
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose). <- Since we are backfilling stats for previous features, we will add them all to the spec and documentation near 3.1 release

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
@@ -23,7 +23,10 @@ import org.opensearch.ingest.IngestDocument;
 import org.opensearch.neuralsearch.processor.chunker.Chunker;
 import org.opensearch.index.mapper.IndexFieldMapper;
 import org.opensearch.neuralsearch.processor.chunker.ChunkerFactory;
+import org.opensearch.neuralsearch.processor.chunker.DelimiterChunker;
 import org.opensearch.neuralsearch.processor.chunker.FixedTokenLengthChunker;
+import org.opensearch.neuralsearch.stats.events.EventStatName;
+import org.opensearch.neuralsearch.stats.events.EventStatsManager;
 import org.opensearch.neuralsearch.util.ProcessorDocumentUtils;
 
 import static org.opensearch.neuralsearch.processor.chunker.Chunker.MAX_CHUNK_LIMIT_FIELD;
@@ -192,6 +195,8 @@ public final class TextChunkingProcessor extends AbstractProcessor {
         runtimeParameters.put(MAX_CHUNK_LIMIT_FIELD, maxChunkLimit);
         runtimeParameters.put(CHUNK_STRING_COUNT_FIELD, chunkStringCount);
         chunkMapType(sourceAndMetadataMap, fieldMap, runtimeParameters);
+
+        recordChunkingExecutionStats(chunker.getAlgorithmName());
         return ingestDocument;
     }
 
@@ -315,5 +320,13 @@ public final class TextChunkingProcessor extends AbstractProcessor {
             result = chunkList((List<String>) value, runTimeParameters);
         }
         return result;
+    }
+
+    private void recordChunkingExecutionStats(String algorithmName) {
+        EventStatsManager.increment(EventStatName.TEXT_CHUNKING_PROCESSOR_EXECUTIONS);
+        switch (algorithmName) {
+            case DelimiterChunker.ALGORITHM_NAME -> EventStatsManager.increment(EventStatName.TEXT_CHUNKING_DELIMITER_EXECUTIONS);
+            case FixedTokenLengthChunker.ALGORITHM_NAME -> EventStatsManager.increment(EventStatName.TEXT_CHUNKING_FIXED_LENGTH_EXECUTIONS);
+        }
     }
 }
--- a/src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
@@ -195,7 +195,6 @@ public final class TextChunkingProcessor extends AbstractProcessor {
         runtimeParameters.put(MAX_CHUNK_LIMIT_FIELD, maxChunkLimit);
         runtimeParameters.put(CHUNK_STRING_COUNT_FIELD, chunkStringCount);
         chunkMapType(sourceAndMetadataMap, fieldMap, runtimeParameters);
-
         recordChunkingExecutionStats(chunker.getAlgorithmName());
         return ingestDocument;
     }
--- a/src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatName.java
+++ b/src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatName.java
@@ -20,7 +20,10 @@ import java.util.stream.Collectors;
 public enum InfoStatName implements StatName {
     // Cluster info
     CLUSTER_VERSION("cluster_version", "", InfoStatType.INFO_STRING),
-    TEXT_EMBEDDING_PROCESSORS("text_embedding_processors_in_pipelines", "processors.ingest", InfoStatType.INFO_COUNTER);
+    TEXT_EMBEDDING_PROCESSORS("text_embedding_processors_in_pipelines", "processors.ingest", InfoStatType.INFO_COUNTER),
+    TEXT_CHUNKING_PROCESSORS("text_chunking_processors", "processors.ingest", InfoStatType.INFO_COUNTER),
+    TEXT_CHUNKING_DELIMITER_PROCESSORS("text_chunking_delimiter_processors", "processors.ingest", InfoStatType.INFO_COUNTER),
+    TEXT_CHUNKING_FIXED_LENGTH_PROCESSORS("text_chunking_fixed_length_processors", "processors.ingest", InfoStatType.INFO_COUNTER);
 
     private final String nameString;
     private final String path;
--- a/src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
+++ b/src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
@@ -4,7 +4,10 @@
  */
 package org.opensearch.neuralsearch.stats.info;
 
+import org.opensearch.neuralsearch.processor.TextChunkingProcessor;
 import org.opensearch.neuralsearch.processor.TextEmbeddingProcessor;
+import org.opensearch.neuralsearch.processor.chunker.DelimiterChunker;
+import org.opensearch.neuralsearch.processor.chunker.FixedTokenLengthChunker;
 import org.opensearch.neuralsearch.settings.NeuralSearchSettingsAccessor;
 import org.opensearch.neuralsearch.stats.common.StatSnapshot;
 import org.opensearch.neuralsearch.util.NeuralSearchClusterUtil;
@@ -118,6 +121,7 @@ public class InfoStatsManager {
     private void addIngestProcessorStats(Map<InfoStatName, CountableInfoStatSnapshot> stats) {
         List<Map<String, Object>> pipelineConfigs = pipelineServiceUtil.getIngestPipelineConfigs();
 
+        // Iterate through all ingest processors and count their stats individually by calling helpers
         for (Map<String, Object> pipelineConfig : pipelineConfigs) {
             List<Map<String, Object>> ingestProcessors = asListOfMaps(pipelineConfig.get(PROCESSORS_KEY));
             for (Map<String, Object> ingestProcessor : ingestProcessors) {
@@ -128,9 +132,31 @@ public class InfoStatsManager {
                         case TextEmbeddingProcessor.TYPE:
                             increment(stats, InfoStatName.TEXT_EMBEDDING_PROCESSORS);
                             break;
+                        case TextChunkingProcessor.TYPE:
+                            countTextChunkingProcessorStats(stats, processorConfig);
+                            break;
                     }
                 }
             }
+        }
+    }
+
+    /**
+     * Counts text chunking processor stats based on processor config
+     * @param stats
+     * @param processorConfig
+     */
+    private void countTextChunkingProcessorStats(Map<InfoStatName, CountableInfoStatSnapshot> stats, Map<String, Object> processorConfig) {
+        increment(stats, InfoStatName.TEXT_CHUNKING_PROCESSORS);
+
+        Map<String, Object> algorithmMap = asMap(processorConfig.get(TextChunkingProcessor.ALGORITHM_FIELD));
+
+        Map.Entry<String, Object> algorithmEntry = algorithmMap.entrySet().iterator().next();
+        String algorithmKey = algorithmEntry.getKey();
+
+        switch (algorithmKey) {
+            case DelimiterChunker.ALGORITHM_NAME -> increment(stats, InfoStatName.TEXT_CHUNKING_DELIMITER_PROCESSORS);
+            case FixedTokenLengthChunker.ALGORITHM_NAME -> increment(stats, InfoStatName.TEXT_CHUNKING_FIXED_LENGTH_PROCESSORS);
         }
     }
 
--- a/src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
+++ b/src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
@@ -143,8 +143,8 @@ public class InfoStatsManager {
 
     /**
      * Counts text chunking processor stats based on processor config
-     * @param stats
-     * @param processorConfig
+     * @param stats map containing the stat to increment
+     * @param processorConfig map of the processor config, parsed to add stats
      */
     private void countTextChunkingProcessorStats(Map<InfoStatName, CountableInfoStatSnapshot> stats, Map<String, Object> processorConfig) {
         increment(stats, InfoStatName.TEXT_CHUNKING_PROCESSORS);
@@ -157,6 +157,8 @@ public class InfoStatsManager {
         switch (algorithmKey) {
             case DelimiterChunker.ALGORITHM_NAME -> increment(stats, InfoStatName.TEXT_CHUNKING_DELIMITER_PROCESSORS);
             case FixedTokenLengthChunker.ALGORITHM_NAME -> increment(stats, InfoStatName.TEXT_CHUNKING_FIXED_LENGTH_PROCESSORS);
+            // If no algorithm is specified, the default is fixed length
+            default -> increment(stats, InfoStatName.TEXT_CHUNKING_FIXED_LENGTH_PROCESSORS);
         }
     }
 
--- a/src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorIT.java
+++ b/src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorIT.java
@@ -17,9 +17,12 @@ import java.util.Objects;
 
 import org.opensearch.index.query.MatchAllQueryBuilder;
 import org.opensearch.neuralsearch.BaseNeuralSearchIT;
+import org.opensearch.neuralsearch.stats.events.EventStatName;
+import org.opensearch.neuralsearch.stats.info.InfoStatName;
 
 public class TextChunkingProcessorIT extends BaseNeuralSearchIT {
     private static final String INDEX_NAME = "text_chunking_test_index";
+    private static final String INDEX_NAME2 = "text_chunking_test_index_2nd";
 
     private static final String OUTPUT_FIELD = "body_chunk";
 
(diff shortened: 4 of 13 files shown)


src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
```diff
@@ -316,4 +321,12 @@ private List<String> chunkLeafType(final Object value, final Map<String, Object>
         }
         return result;
     }
+
+    private void recordChunkingExecutionStats() {
+        EventStatsManager.increment(EventStatName.TEXT_CHUNKING_PROCESSOR_EXECUTIONS);
+        switch (chunker.getAlgorithmName()) {
```
will-hwang, line 327: can we pass in chunker.getAlgorithmName as parameter to this method, and record the event directly?

something like this:

```
private void recordChunkingExecutionStats(String algorithmName) {
        EventStatsManager.increment(EventStateName.valueOf(algorithmName))
}
```
`recordChunkingExecutionStats(chunker.getAlgorithmName());`


that way, we don't have to add switch/case for additional algorithms in the future

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
```diff
@@ -128,12 +132,34 @@ private void addIngestProcessorStats(Map<InfoStatName, CountableInfoStatSnapshot
...
+     */
+    private void countTextChunkingProcessorStats(Map<InfoStatName, CountableInfoStatSnapshot> stats, Map<String, Object> processorConfig) {
+        increment(stats, InfoStatName.TEXT_CHUNKING_PROCESSORS);
+
+        Map<String, Object> algorithmMap = asMap(processorConfig.get(TextChunkingProcessor.ALGORITHM_FIELD));
+
+        Map.Entry<String, Object> algorithmEntry = algorithmMap.entrySet().iterator().next();
+        String algorithmKey = algorithmEntry.getKey();
+
+        switch (algorithmKey) {
```
will-hwang, line 157: same comment as above

src/test/java/org/opensearch/neuralsearch/stats/info/InfoStatNameTests.java
```diff
@@ -41,15 +40,17 @@ public void test_validNames() {
...
+        for (InfoStatName statName : INFO_STATS) {
             String path = statName.getPath().toLowerCase(Locale.ROOT);
-            assertFalse(String.format(Locale.ROOT, "Checking path uniqueness for %s", path), paths.contains(path));
             paths.add(path);
         }
 
         // Check possible path collisions
         // i.e. a full path is a terminal path that should not have any children
-        for (InfoStatName statName : STATE_STATS) {
+        for (InfoStatName statName : INFO_STATS) {
```
will-hwang, line 53: [nit] we could use stream

`INFO_STATS.forEach((stat) -> sortedPaths.add(stat...).distinct()`

src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
q-andy: Sure we can pass in the `chunker.getAlgorithmName()` as a parameter. But for recording the event directly, `EventStatName` doesn't have any kind of mapping for the algorithm name though so we can't call `valueOf` directly. In this case the stat name is `text_chunking_fixed_length_processors` and the algorithm name is `fixed_length` so we would have to move the switch block to the enum class to make it work anyways. In the general case, not all stats map cleanly to strings like this so I think its okay to keep it as a switch.

There's some additional manual work involved with adding new algorithms for stats like updating unit/integ tests, API spec, and documentation. Since it results in an API response change I think we don't wanna do it automatically necessarily.

src/test/java/org/opensearch/neuralsearch/stats/info/InfoStatNameTests.java
q-andy: I think I like foreach here a little better since it's more readable and matches the rest of the tests. Compare the stream, it's similar in length.
```
INFO_STATS.stream()
        .map(InfoStatName::getPath)
        .map(str -> return str.toLowerCase(Locale.ROOT))
        .forEach(paths::add);
```
vs
```
for (InfoStatName statName : INFO_STATS) {
    String path = statName.getPath().toLowerCase(Locale.ROOT);
    paths.add(path);
}
```

src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
will-hwang: i see, i think leaving as is should be okay then. For future, we should try to come up with a strategy to keep stats emission consistent. maybe having a class/map kind of structure could be of help here.

`EventStatsManager.increment(getEventStat(...))`

src/test/java/org/opensearch/neuralsearch/stats/info/InfoStatNameTests.java
will-hwang: for handling unique too? `distinct()` should handle duplicates

src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
q-andy: That could be useful, the drawback would be we'd just be moving the switch case into that class. I think the core concern you're pointing out is because we have to have this huge list of stats configured. But either way, we're having to declare a hardcoded set of stats somewhere, whether its an enum declaration or in map or wrapper class. We can revisit as we add more stats if the current approach gets unwieldy

src/test/java/org/opensearch/neuralsearch/stats/info/InfoStatNameTests.java
will-hwang: makes sense, thanks

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatName.java
```diff
@@ -20,7 +20,10 @@
 public enum InfoStatName implements StatName {
     // Cluster info
     CLUSTER_VERSION("cluster_version", "", InfoStatType.INFO_STRING),
-    TEXT_EMBEDDING_PROCESSORS("text_embedding_processors_in_pipelines", "processors.ingest", InfoStatType.INFO_COUNTER);
+    TEXT_EMBEDDING_PROCESSORS("text_embedding_processors_in_pipelines", "processors.ingest", InfoStatType.INFO_COUNTER),
```
will-hwang, line 23: should just be `text_embedding_processors`?

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
```diff
@@ -128,12 +132,34 @@ private void addIngestProcessorStats(Map<InfoStatName, CountableInfoStatSnapshot
                         case TextEmbeddingProcessor.TYPE:
                             increment(stats, InfoStatName.TEXT_EMBEDDING_PROCESSORS);
```
will-hwang, line 133: shouldn't we keep this consistent with TextChunkingProcessor.TYPE here too?

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatName.java
q-andy: Yeah ideally, but since that stat name was released in 3.0 I think changing it would technically be breaking. The alternative is that we change all of `text_chunking_<algorithm>_processors` to be `text_chunking_<algorithm>_processors_in_pipelines` instead but then the name is pretty verbose. For normalization processor stats I had names like `normalization_processors_combination_technique_harmonic_mean_in_pipelines` or `text_embedding_processor_skip_inference_enabled_in_pipelines`. I'm wondering if it's better to just accept the inconsistency early to slightly shorten the strings going forwards. What do you think?

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
q-andy: Do you mean incrementing the stat in-line vs having a helper? Sure, I can make a`countTestEmbeddingProcessorStats()` helper in the next PR to cover parsing the `skip_inference` flag. But I think for processors where there aren't additional config options, incrementing in line is ok.

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatName.java
will-hwang: i'm okay with either approach, does this mean we won't be able to change its name going forward?

junqiu-lei: I'm open to renaming to `text_embedding_processors`, but we need update and add note to the documentation for the backwards compatibility. https://docs.opensearch.org/docs/latest/vector-search/api/neural/#available-statistics. Not sure how widely this API is being used at open source community since 3.0.0 release.

q-andy: Makes sense, we'll also have to change the API spec since those run integ tests to verify response schema. I'll open a separate PR for renaming it to make it clearer to track. I opened https://github.com/opensearch-project/opensearch-api-specification/issues/890 to track all spec changes we'll need for 3.1

src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
```diff
@@ -192,6 +195,8 @@ public IngestDocument execute(final IngestDocument ingestDocument) {
         runtimeParameters.put(MAX_CHUNK_LIMIT_FIELD, maxChunkLimit);
         runtimeParameters.put(CHUNK_STRING_COUNT_FIELD, chunkStringCount);
         chunkMapType(sourceAndMetadataMap, fieldMap, runtimeParameters);
+
```
yuye-aws, line 198: nit: extra line

src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
```diff
@@ -316,4 +321,12 @@ private List<String> chunkLeafType(final Object value, final Map<String, Object>
         }
         return result;
     }
+
+    private void recordChunkingExecutionStats(String algorithmName) {
+        EventStatsManager.increment(EventStatName.TEXT_CHUNKING_PROCESSOR_EXECUTIONS);
+        switch (algorithmName) {
+            case DelimiterChunker.ALGORITHM_NAME -> EventStatsManager.increment(EventStatName.TEXT_CHUNKING_DELIMITER_EXECUTIONS);
+            case FixedTokenLengthChunker.ALGORITHM_NAME -> EventStatsManager.increment(EventStatName.TEXT_CHUNKING_FIXED_LENGTH_EXECUTIONS);
```
yuye-aws, line 329: You are using switch case, should you add break under each case?

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
```diff
@@ -128,12 +132,34 @@ private void addIngestProcessorStats(Map<InfoStatName, CountableInfoStatSnapshot
...
                 }
             }
         }
     }
 
+    /**
+     * Counts text chunking processor stats based on processor config
+     * @param stats
+     * @param processorConfig
+     */
```
yuye-aws, line 148: Can you complete the java doc?

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
```diff
@@ -128,12 +132,34 @@ private void addIngestProcessorStats(Map<InfoStatName, CountableInfoStatSnapshot
...
+        increment(stats, InfoStatName.TEXT_CHUNKING_PROCESSORS);
+
+        Map<String, Object> algorithmMap = asMap(processorConfig.get(TextChunkingProcessor.ALGORITHM_FIELD));
+
+        Map.Entry<String, Object> algorithmEntry = algorithmMap.entrySet().iterator().next();
+        String algorithmKey = algorithmEntry.getKey();
+
+        switch (algorithmKey) {
+            case DelimiterChunker.ALGORITHM_NAME -> increment(stats, InfoStatName.TEXT_CHUNKING_DELIMITER_PROCESSORS);
+            case FixedTokenLengthChunker.ALGORITHM_NAME -> increment(stats, InfoStatName.TEXT_CHUNKING_FIXED_LENGTH_PROCESSORS);
```
yuye-aws, line 159: Add break for switch case?

src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorIT.java
```diff
@@ -167,6 +170,44 @@ public void testTextChunkingProcessor_withFixedTokenLengthAlgorithmStandardToken
         assertEquals(1, getDocCount(INDEX_NAME));
     }
 
+    @SneakyThrows
+    public void testTextChunkingProcessor_processorStats() {
+        updateClusterSettings("plugins.neural_search.stats_enabled", true);
```
yuye-aws, line 175: Need 1 more unit test on stats_enabled false

src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorIT.java
```diff
@@ -167,6 +170,44 @@ public void testTextChunkingProcessor_withFixedTokenLengthAlgorithmStandardToken
...
+        createPipelineProcessor(DELIMITER_PIPELINE_NAME);
+        createTextChunkingIndex(INDEX_NAME2, DELIMITER_PIPELINE_NAME);
+
+        String document = getDocumentFromFilePath(TEST_DOCUMENT);
+        ingestDocument(INDEX_NAME, document);
+        ingestDocument(INDEX_NAME, document);
+
+        ingestDocument(INDEX_NAME2, document);
+        ingestDocument(INDEX_NAME2, document);
+        ingestDocument(INDEX_NAME2, document);
```
yuye-aws, line 191: Please compare the ingested document with expected result. After all, this is the IT for text chunking processor

src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorTests.java
```diff
@@ -92,6 +94,10 @@ public void setup() {
         when(clusterState.metadata()).thenReturn(metadata);
         when(clusterService.state()).thenReturn(clusterState);
         textChunkingProcessorFactory = new TextChunkingProcessorFactory(environment, clusterService, getAnalysisRegistry());
+
+        NeuralSearchSettingsAccessor settingsAccessor = mock(NeuralSearchSettingsAccessor.class);
+        when(settingsAccessor.isStatsEnabled()).thenReturn(true);
+        EventStatsManager.instance().initialize(settingsAccessor);
```
yuye-aws, line 100: Seems these mocks are not used. Are you going to unit tests later?

Besides, you need to consider settings accessor disabled

src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
q-andy: This uses switch expressions introduced in Java 12 so using `->` means we don't need `break` anymore: https://docs.oracle.com/en/java/javase/13/language/switch-expressions.html

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
q-andy: Sure, added

src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorIT.java
q-andy: This is already tested implicitly by the other tests, the rest of the ITs have `stats_enabled` false by default by adding this setting which is default false: https://github.com/opensearch-project/neural-search/blob/main/src/test/java/org/opensearch/neuralsearch/plugin/NeuralSearchTests.java#L95-L97. When stats are disabled, the response returns forbidden, tested here: https://github.com/opensearch-project/neural-search/blob/main/src/test/java/org/opensearch/neuralsearch/rest/RestNeuralStatsActionIT.java#L64-L70

src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorIT.java
q-andy: Sure, added

src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorTests.java
q-andy: Added unit tests for enabled case. Even without unit tests, these mocks are necessary since `EventStatsManager.increment()` checks the `settingsAccessor` passed in through `initialize`. It's difficult to mock static method calls otherwise with our current testing setup.

For disabled case, the increment operation is [treated as a noop](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/stats/events/EventStatsManager.java#L60-L62) and this case is covered tested in the static method tests in [EventStatsManagerTests](https://github.com/opensearch-project/neural-search/blob/283cc3af78c882c279d3a2663324c3442b9791af/src/test/java/org/opensearch/neuralsearch/stats/events/EventStatsManagerTests.java#L48-L58). When the setting is disabled, and the API is also disabled entirely (verified in other integ test). So disabled case should already be covered. I added them here anyways, but I don't think its necessary to duplicate that for every processor that increments stats since they share functionality when setting is disabled (all increment calls become noop).

src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorIT.java
```diff
@@ -167,8 +170,57 @@ public void testTextChunkingProcessor_withFixedTokenLengthAlgorithmStandardToken
...
+
+        expectedPassages = new ArrayList<>();
+        expectedPassages.add("This is an example document to be chunked.");
+        expectedPassages.add(" The document contains a single paragraph, two sentences and 24 tokens by standard tokenizer in OpenSearch.");
+        validateIndexIngestResultsWithMultipleDocs(INDEX_NAME2, OUTPUT_FIELD, expectedPassages, 3);
+
+        // Get stats requestfile:///Users/qinandy/opensearch/neural-search-os/build/reports/tests/integTest/index.html
```
yuye-aws, line 204: You are exposing your file path here

src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorIT.java
yuye-aws: Make sense

src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorTests.java
```diff
@@ -972,4 +982,70 @@ public void testExecute_withIgnoreMissing_thenSucceed() {
         IngestDocument document = processor.execute(ingestDocument);
         assertFalse(document.getSourceAndMetadata().containsKey(OUTPUT_FIELD));
     }
+
+    @SneakyThrows
+    public void testExecute_statsDisabled_happyCase() {
```
yuye-aws, line 987: nit: testExecute_statsDisabled_thenSucceed

yuye-aws (commented):
BWC tests for text chunking processor IT are needed

q-andy:
Could you explain what cases you want BWC to cover? Stats API will not work as intended during cluster restarts or rolling upgrades since the stat info is stored in memory and lost on restart. This is accepted by design, since the goal is to track long term trends by polling the API regularly.

Existing BWC tests for API are here: https://github.com/opensearch-project/neural-search/blob/main/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/RestNeuralStatsActionIT.java

yuye-aws:
> Could you explain what cases you want BWC to cover? Stats API will not work as intended during cluster restarts or rolling upgrades since the stat info is stored in memory and lost on restart. This is accepted by design, since the goal is to track long term trends by polling the API regularly.
> 
> Existing BWC tests for API are here: https://github.com/opensearch-project/neural-search/blob/main/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/RestNeuralStatsActionIT.java

@vibrantvarun Can you provide more context in the BWC tests?

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
```diff
@@ -128,12 +132,34 @@ private void addIngestProcessorStats(Map<InfoStatName, CountableInfoStatSnapshot
...
```
bzhangam, line 159: Same here. Handle unexpected algorithm properly.

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
```diff
@@ -128,12 +132,34 @@ private void addIngestProcessorStats(Map<InfoStatName, CountableInfoStatSnapshot
...
 
+    /**
+     * Counts text chunking processor stats based on processor config
+     * @param stats map containing the stat to increment
+     * @param processorConfig map of the processor config, parsed to add stats
+     */
+    private void countTextChunkingProcessorStats(Map<InfoStatName, CountableInfoStatSnapshot> stats, Map<String, Object> processorConfig) {
+        increment(stats, InfoStatName.TEXT_CHUNKING_PROCESSORS);
+
+        Map<String, Object> algorithmMap = asMap(processorConfig.get(TextChunkingProcessor.ALGORITHM_FIELD));
```
bzhangam, line 152: Algorithm can be empty and in that case we will use the `DEFAULT_ALGORITHM = FixedTokenLengthChunker.ALGORITHM_NAME` we should handle this case properly.

src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
```diff
@@ -316,4 +320,12 @@ private List<String> chunkLeafType(final Object value, final Map<String, Object>
         }
         return result;
     }
+
+    private void recordChunkingExecutionStats(String algorithmName) {
+        EventStatsManager.increment(EventStatName.TEXT_CHUNKING_PROCESSOR_EXECUTIONS);
```
bzhangam, line 325: I wonder how are we using the setting `NEURAL_STATS_ENABLED`? In your comments it says ```Enables or disables the Stats API and event stat collection.  If API is called when stats are disabled, the response will 403.  Event stat increment calls are also treated as no-ops.``` but I don't find where we check this setting to disable the stat related logic.

src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
```diff
@@ -316,4 +320,12 @@ private List<String> chunkLeafType(final Object value, final Map<String, Object>
...
+        switch (algorithmName) {
+            case DelimiterChunker.ALGORITHM_NAME -> EventStatsManager.increment(EventStatName.TEXT_CHUNKING_DELIMITER_EXECUTIONS);
+            case FixedTokenLengthChunker.ALGORITHM_NAME -> EventStatsManager.increment(EventStatName.TEXT_CHUNKING_FIXED_LENGTH_EXECUTIONS);
```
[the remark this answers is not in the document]
q-andy: What kind of default behavior are you thinking? I think it's okay that we don't enforce stat collection since it's a choice by the feature developer or plugin owners on what stats we choose to collect. Not collecting stats isn't breaking so I don't think it makes sense to throw an exception. And we can't do it automatically since creating a new stat requires creating a new enum which can't be done at runtime.

src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
q-andy: The idea behind `NEURAL_STATS_ENABLED` is to give cluster managers flexibility to record stats or not. Enabled or disabled, it depends on cluster manager use case, since cluster managers are the ones who are using the stats.

junqiu-lei (approved):

src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorTests.java
yuye-aws: Make sense. Resolving this thread now...

src/test/java/org/opensearch/neuralsearch/processor/TextChunkingProcessorIT.java
```diff
@@ -167,8 +170,57 @@ public void testTextChunkingProcessor_withFixedTokenLengthAlgorithmStandardToken
         assertEquals(1, getDocCount(INDEX_NAME));
     }
 
-    private void validateIndexIngestResults(String indexName, String fieldName, Object expected) {
-        assertEquals(1, getDocCount(indexName));
+    @SneakyThrows
+    public void testTextChunkingProcessor_processorStats() {
```
yuye-aws, line 174: nit: testTextChunkingProcessor_processorStats_successful

q-andy: sure

yuye-aws (approved):
LGTM

yuye-aws:
@q-andy Can you resolve the conflicts?

PR #1309, opened 2025-05-06
Add semantic ingest processor.

bzhangam:
### Description
Add the semantic ingest processor to handle the text chunking and embedding generation for the semantic fields.

This PR depends on a change in core and that [PR](https://github.com/opensearch-project/OpenSearch/pull/17817) is still in review so tests will fail until that PR is merged into core.

### Related Issues

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -288,6 +288,7 @@ dependencies {
     testFixturesImplementation fileTree(dir: knnJarDirectory, include: ["opensearch-knn-${opensearch_build}.jar", "remote-index-build-client-${opensearch_build}.jar"])
     testImplementation fileTree(dir: knnJarDirectory, include: ["opensearch-knn-${opensearch_build}.jar", "remote-index-build-client-${opensearch_build}.jar"])
     testImplementation "org.opensearch.plugin:parent-join-client:${opensearch_version}"
+    testImplementation 'org.assertj:assertj-core:3.24.2'
 }
 
 // In order to add the jar to the classpath, we need to unzip the
--- a/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
@@ -12,6 +12,11 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
 import org.opensearch.core.action.ActionListener;
@@ -314,6 +319,71 @@ public class MLCommonsClientAccessor {
 
     public void getModel(@NonNull final String modelId, @NonNull final ActionListener<MLModel> listener) {
         retryableGetModel(modelId, 0, listener);
+    }
+
+    /**
+     * Get model info for multiple model ids. It will send multiple getModel requests to get the model info in parallel.
+     * It will fail if any one of the get model request fail. Only return the success result if all model info is
+     * successfully retrieved.
+     * @param modelIds a set of model ids
+     * @param onSuccess onSuccess consumer
+     * @param onFailure onFailure consumer
+     */
+    public void getModels(
+        @NonNull final Set<String> modelIds,
+        @NonNull final Consumer<Map<String, MLModel>> onSuccess,
+        @NonNull final Consumer<Exception> onFailure
+    ) {
+        if (modelIds.isEmpty()) {
+            try {
+                onSuccess.accept(Collections.emptyMap());
+            } catch (Exception e) {
+                onFailure.accept(e);
+            }
+            return;
+        }
+
+        final Map<String, MLModel> modelMap = new ConcurrentHashMap<>();
+        final AtomicInteger counter = new AtomicInteger(modelIds.size());
+        final AtomicBoolean hasError = new AtomicBoolean(false);
+        final List<String> errors = Collections.synchronizedList(new ArrayList<>());
+
+        for (String modelId : modelIds) {
+            try {
+                getModel(modelId, ActionListener.wrap(model -> {
+                    modelMap.put(modelId, model);
+                    if (counter.decrementAndGet() == 0) {
+                        if (hasError.get()) {
+                            onFailure.accept(new RuntimeException(String.join(";", errors)));
+                        } else {
+                            try {
+                                onSuccess.accept(modelMap);
+                            } catch (Exception e) {
+                                onFailure.accept(e);
+                            }
+                        }
+                    }
+                }, e -> { handleGetModelException(hasError, errors, modelId, e, counter, onFailure); }));
+            } catch (Exception e) {
+                handleGetModelException(hasError, errors, modelId, e, counter, onFailure);
+            }
+        }
+
+    }
+
+    private void handleGetModelException(
+        AtomicBoolean hasError,
+        List<String> errors,
+        String modelId,
+        Exception e,
+        AtomicInteger counter,
+        @NonNull Consumer<Exception> onFailure
+    ) {
+        hasError.set(true);
+        errors.add("Failed to fetch model [" + modelId + "]: " + e.getMessage());
+        if (counter.decrementAndGet() == 0) {
+            onFailure.accept(new RuntimeException(String.join(";", errors)));
+        }
     }
 
     private void retryableGetModel(@NonNull final String modelId, final int retryTime, @NonNull final ActionListener<MLModel> listener) {
--- a/src/main/java/org/opensearch/neuralsearch/processor/dto/SemanticFieldInfo.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/dto/SemanticFieldInfo.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.processor.dto;
+
+import lombok.Data;
+
+import java.util.List;
+
+import static org.opensearch.neuralsearch.constants.MappingConstants.PATH_SEPARATOR;
+import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.CHUNKS_EMBEDDING_FIELD_NAME;
+import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.CHUNKS_FIELD_NAME;
+import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.MODEL_FIELD_NAME;
+
+/**
+ * SemanticFieldInfo is a data transfer object to help hold semantic field info
+ */
+@Data
+public class SemanticFieldInfo {
+    /**
+     * The raw string value of the semantic field
+     */
+    private String value;
+    /**
+     * The model id of the semantic field which will be used to generate the embedding
+     */
+    private String modelId;
+    /**
+     * The full path to the semantic field
+     */
+    private String fullPath;
+    /**
+     * The full path to the semantic info fields
+     */
+    private String semanticInfoFullPath;
+    /**
+     * The chunked strings of the original string value of the semantic field
+     */
+    private List<String> chunks;
+
+    /**
+     * @return full path to the chunks field of the semantic field
+     */
+    public String getFullPathForChunks() {
+        return new StringBuilder().append(semanticInfoFullPath).append(PATH_SEPARATOR).append(CHUNKS_FIELD_NAME).toString();
+    }
+
+    /**
+     * @param index index of the chunk the embedding is in
+     * @return full path to the embedding field of the semantic field
+     */
+    public String getFullPathForEmbedding(int index) {
+        return new StringBuilder().append(semanticInfoFullPath)
+            .append(PATH_SEPARATOR)
+            .append(CHUNKS_FIELD_NAME)
+            .append(PATH_SEPARATOR)
+            .append(index)
+            .append(PATH_SEPARATOR)
+            .append(CHUNKS_EMBEDDING_FIELD_NAME)
+            .toString();
+    }
+
+    /**
+     * @return full path to the model info fields
+     */
+    public String getFullPathForModelInfo() {
+        return new StringBuilder().append(semanticInfoFullPath).append(PATH_SEPARATOR).append(MODEL_FIELD_NAME).toString();
+    }
+}
--- a/src/main/java/org/opensearch/neuralsearch/processor/factory/SemanticFieldProcessorFactory.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/factory/SemanticFieldProcessorFactory.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.processor.factory;
+
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.env.Environment;
+import org.opensearch.index.analysis.AnalysisRegistry;
+import org.opensearch.ingest.AbstractBatchingSystemProcessor;
+import org.opensearch.neuralsearch.processor.chunker.Chunker;
+import org.opensearch.neuralsearch.processor.chunker.ChunkerFactory;
+import org.opensearch.neuralsearch.processor.chunker.FixedTokenLengthChunker;
+import org.opensearch.neuralsearch.util.SemanticMappingUtils;
+import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
+import org.opensearch.neuralsearch.processor.semantic.SemanticFieldProcessor;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.opensearch.plugins.IngestPlugin.SystemIngestPipelineConfigKeys.INDEX_MAPPINGS;
+import static org.opensearch.plugins.IngestPlugin.SystemIngestPipelineConfigKeys.INDEX_TEMPLATE_MAPPINGS;
+
+/**
+ *  Factory for semantic fields.
+ *
+ *  This factory is for internal usage and will be systematically invoked if we detect there are semantic fields
+ *  defined. Users should not be able to define this type of the processor in a regular ingest pipeline.
+ */
+public final class SemanticFieldProcessorFactory extends AbstractBatchingSystemProcessor.Factory {
+    public static final String PROCESSOR_FACTORY_TYPE = "system_ingest_processor_factory_semantic_field";
+    /**
+     * Ideally this should be configurable through the semantic fields. For P0 we will use a default value,
+     * and in P1 we will make it configurable.
+     * TODO: Make batch size configurable
+     */
+    private static final int DEFAULT_BATCH_SIZE = 10;
+
+    private final MLCommonsClientAccessor mlClientAccessor;
+
+    private final Environment environment;
+
+    private final ClusterService clusterService;
+    private final AnalysisRegistry analysisRegistry;
+    private final static int DEFAULT_TOKEN_LIMIT = 50;
+    private final static double DEFAULT_OVERLAP_RATE = 0.2;
+
+    public SemanticFieldProcessorFactory(
+        final MLCommonsClientAccessor mlClientAccessor,
+        final Environment environment,
+        final ClusterService clusterService,
+        final AnalysisRegistry analysisRegistry
+    ) {
+        super(PROCESSOR_FACTORY_TYPE);
+        this.mlClientAccessor = mlClientAccessor;
+        this.environment = environment;
+        this.clusterService = clusterService;
+        this.analysisRegistry = analysisRegistry;
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    protected AbstractBatchingSystemProcessor newProcessor(String tag, String description, Map<String, Object> config) {
+        final List<Map<String, Object>> mappings = new ArrayList<>();
+        final Object mappingFromIndex = config.get(INDEX_MAPPINGS);
+        final Object mappingFromTemplates = config.get(INDEX_TEMPLATE_MAPPINGS);
+        if (mappingFromTemplates instanceof List) {
+            mappings.addAll((List<Map<String, Object>>) mappingFromTemplates);
+        }
+        if (mappingFromIndex instanceof Map) {
+            mappings.add((Map<String, Object>) mappingFromIndex);
+        }
+
+        // If no config we are not able to create a processor so simply return a null to show no processor created
+        if (mappings.isEmpty()) {
+            return null;
+        }
+
+        if (description == null) {
+            description = "This is a system ingest processor for semantic fields. It will do text chunking and "
+                + "embedding generation for semantic fields.";
+        }
+
+        final Map<String, Map<String, Object>> semanticFieldPathToConfigMap = new HashMap<>();
+        for (final Map<String, Object> mapping : mappings) {
+            final Map<String, Object> properties = SemanticMappingUtils.getProperties(mapping);
+            // if there is no property in the mapping we simply skip it
+            if (properties.isEmpty()) {
+                continue;
+            }
+
+            SemanticMappingUtils.collectSemanticField(properties, semanticFieldPathToConfigMap);
+        }
+
+        // If no semantic field we don't need to create a processor so simply return null to show no processor created.
+        if (semanticFieldPathToConfigMap.isEmpty()) {
+            return null;
+        }
+
+        return new SemanticFieldProcessor(
+            tag,
+            description,
+            DEFAULT_BATCH_SIZE,
+            semanticFieldPathToConfigMap,
+            mlClientAccessor,
+            environment,
+            clusterService,
+            createChunker()
+        );
+    }
+
+    /**
+     * Create a default text chunker.
+     * TODO: Make it configurable
+     * @return A default fixed token length chunker
+     */
+    private Chunker createChunker() {
+        final Map<String, Object> chunkerParameters = new HashMap<>();
+        chunkerParameters.put(FixedTokenLengthChunker.TOKEN_LIMIT_FIELD, DEFAULT_TOKEN_LIMIT);
+        chunkerParameters.put(FixedTokenLengthChunker.OVERLAP_RATE_FIELD, DEFAULT_OVERLAP_RATE);
+        chunkerParameters.put(FixedTokenLengthChunker.ANALYSIS_REGISTRY_FIELD, analysisRegistry);
+        return ChunkerFactory.create(FixedTokenLengthChunker.ALGORITHM_NAME, chunkerParameters);
+    }
+}
--- a/src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
@@ -1,0 +1,555 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.processor.semantic;
+
+import lombok.NonNull;
+import lombok.extern.log4j.Log4j2;
+import org.apache.commons.lang3.tuple.Pair;
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.Nullable;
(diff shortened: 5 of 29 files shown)


build.gradle
```diff
@@ -288,6 +288,7 @@ dependencies {
     testFixturesImplementation fileTree(dir: knnJarDirectory, include: ["opensearch-knn-${opensearch_build}.jar", "remote-index-build-client-${opensearch_build}.jar"])
     testImplementation fileTree(dir: knnJarDirectory, include: ["opensearch-knn-${opensearch_build}.jar", "remote-index-build-client-${opensearch_build}.jar"])
     testImplementation "org.opensearch.plugin:parent-join-client:${opensearch_version}"
+    testImplementation 'org.assertj:assertj-core:3.24.2'
```
bzhangam, line 291: This is added for deep equal check for the ingested doc.

src/main/java/org/opensearch/neuralsearch/processor/factory/SemanticFieldProcessorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        // If no config we are not able to create a processor so simply return a null to show no processor created
+        if (mappings.isEmpty()) {
+            return null;
+        }
+
+        if (description == null) {
+            description = "This is a system ingest processor for semantic fields. It will do text chunking and "
+                + "embedding generation for semantic fields.";
```
q-andy, line 78: `String.format`

src/main/java/org/opensearch/neuralsearch/processor/factory/SemanticFieldProcessorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+    protected AbstractBatchingSystemProcessor newProcessor(String tag, String description, Map<String, Object> config) {
+        final List<Map<String, Object>> mappings = new ArrayList<>();
+        final Object mappingFromIndex = config.get(INDEX_MAPPINGS);
+        final Object mappingFromTemplates = config.get(INDEX_TEMPLATE_MAPPINGS);
+        if (mappingFromTemplates instanceof List) {
+            mappings.addAll((List<Map<String, Object>>) mappingFromTemplates);
+        }
+        if (mappingFromIndex instanceof Map) {
+            mappings.add((Map<String, Object>) mappingFromIndex);
+        }
```
q-andy, line 69: Since we try to execute and construct system ingest pipelines for every index request, is this additional object creation overhead for every single index request too? Should we move the validations higher up?

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -0,0 +1,547 @@
...
+import static org.opensearch.neuralsearch.processor.chunker.Chunker.MAX_CHUNK_LIMIT_FIELD;
+import static org.opensearch.neuralsearch.processor.util.ProcessorUtils.getMaxTokenCount;
+import static org.opensearch.neuralsearch.util.ProcessorDocumentUtils.unflattenIngestDoc;
+import static org.opensearch.neuralsearch.util.SemanticMLModelUtils.getModelType;
+import static org.opensearch.neuralsearch.util.SemanticMLModelUtils.isDenseModel;
+import static org.opensearch.neuralsearch.util.SemanticMappingUtils.getModelId;
+import static org.opensearch.neuralsearch.util.SemanticMappingUtils.getSemanticInfoFieldName;
+
+/**
+ * Processor to ingest the semantic fields.
```
q-andy, line 62: nit: javadoc should also describe what the processor does, e.g. in this case embedding generation and chunking

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -0,0 +1,547 @@
...
+    /**
+     * Create a default text chunker.
+     * TODO: Make it configurable
+     * @return A default fixed token length chunker
+     */
+    private Chunker createChunker() {
+        final Map<String, Object> chunkerParameters = new HashMap<>();
+        chunkerParameters.put(FixedTokenLengthChunker.TOKEN_LIMIT_FIELD, 50);
+        chunkerParameters.put(FixedTokenLengthChunker.OVERLAP_RATE_FIELD, 0.2);
+        chunkerParameters.put(FixedTokenLengthChunker.ANALYSIS_REGISTRY_FIELD, analysisRegistry);
```
q-andy, line 111: Have we checked these values against what customers usually use? The `DEFAULT_MAX_CHUNK_LIMIT` is 100 so with 50 token limit means the field has a soft max of 5000 tokens

src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
```diff
@@ -288,6 +289,17 @@ private static void flattenAndFlip(String path, Map<String, Object> map, Map<Str
         }
     }
 
+    public static void unflattenIngestDoc(IngestDocument document) {
+        if (document == null || document.getSourceAndMetadata() == null) {
+            return;
+        }
+        ;
```
q-andy, line 296: nit: semicolon?

q-andy (commented):
Overall looks good, some minor nits. Also see my question about the chunking max limit

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -0,0 +1,547 @@
...
+        return ChunkerFactory.create(FixedTokenLengthChunker.ALGORITHM_NAME, chunkerParameters);
+    }
+
+    /**
+     * Since we need to do async work in this processor we will not invoke this function to ingest the doc.
```
heemin32, line 116: Shouldn't we throw an exception instead if this is not a method to be invoked?

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -0,0 +1,547 @@
...
+        // actual changes) can trigger the processor to reload the latest model config.
+        //
+        // Since model config is already validated during index creation, we cache it here for better
+        // performance to avoid fetching it on every ingest request.
+        // TODO: Handle the model config update case more gracefully.
+        for (final String existingModelId : modelIdToModelMap.keySet()) {
+            modelIdsToGetModelInfo.remove(existingModelId);
+        }
+        if (modelIdsToGetModelInfo.isEmpty()) {
+            process(ingestDocument, semanticFieldInfoList, handler);
```
heemin32, line 173: Shouldn't we handle the case when process throw an exception?

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -0,0 +1,547 @@
...
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    private void generateAndSetEmbedding(
+        @NonNull final IngestDocument ingestDocument,
+        @NonNull final List<SemanticFieldInfo> semanticFieldInfoList,
+        @NonNull final BiConsumer<IngestDocument, Exception> handler
+    ) {
+        final Map<String, Set<String>> modelIdToRawDataMap = groupRawDataByModelId(semanticFieldInfoList);
```
heemin32, line 227: Shouldn't we handle the exception to be thrown from `groupRawDataByModelId`?

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -0,0 +1,547 @@
...
+
+        for (Map.Entry<IngestDocumentWrapper, List<SemanticFieldInfo>> entry : docToSemanticFieldInfoMap.entrySet()) {
+            final IngestDocumentWrapper ingestDocumentWrapper = entry.getKey();
+            final IngestDocument ingestDocument = entry.getKey().getIngestDocument();
+            final List<SemanticFieldInfo> semanticFieldInfoList = entry.getValue();
+            try {
+                setModelInfo(ingestDocument, semanticFieldInfoList);
+
+                chunk(ingestDocument, semanticFieldInfoList);
+            } catch (Exception e) {
```
heemin32, line 433: Can we add a log here?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -316,6 +321,57 @@ public void getModel(@NonNull final String modelId, @NonNull final ActionListene
...
+            return;
+        }
+
+        final Map<String, MLModel> modelMap = new ConcurrentHashMap<>();
+        final AtomicInteger counter = new AtomicInteger(modelIds.size());
+        final AtomicBoolean hasError = new AtomicBoolean(false);
+        final List<String> errors = Collections.synchronizedList(new ArrayList<>());
+
+        for (String modelId : modelIds) {
+            getModel(modelId, ActionListener.wrap(model -> {
```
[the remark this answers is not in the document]
junqiu-lei: +1.

The current implementation seems to call onFailure upon the first encountered error during the iteration. Considering the onFailure for the overall getModels call to aggregate errors if multiple model fetches fail.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -0,0 +1,547 @@
...
+/**
+ * Processor to ingest the semantic fields.
+ *
+ * This processor is for internal usage and will be systematically invoked if we detect there are semantic fields
+ * defined. Users should not be able to define this processor in a regular ingest pipeline.
+ */
+@Log4j2
+public class SemanticFieldProcessor extends AbstractBatchingSystemProcessor {
+
+    public static final String PROCESSOR_TYPE = "system_ingest_processor_semantic_field";
```
junqiu-lei, line 70: Do we need the duplicated static string at two places? 
```
public static final String PROCESSOR_FACTORY_TYPE = "system_ingest_processor_semantic_field";
```
https://github.com/opensearch-project/neural-search/pull/1309/files#diff-baf3af489186204954fc08235a9c81119d31a4c7d5e5b0503e200becc22f09e3R30

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -0,0 +1,547 @@
...
+        // dimension of a dense model changes and no longer matches the index mapping), errors will be
+        // thrown at runtime—for example, during document ingestion or query execution.
+        //
+        // Users must update the index mapping to reflect the new model config. Simply updating the
+        // model is not enough; a dummy mapping update that includes the semantic field (even without
+        // actual changes) can trigger the processor to reload the latest model config.
+        //
+        // Since model config is already validated during index creation, we cache it here for better
+        // performance to avoid fetching it on every ingest request.
+        // TODO: Handle the model config update case more gracefully.
```
junqiu-lei, line 168: We probably need add this note in documentation.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -0,0 +1,547 @@
...
+                    ingestDocumentWrapper.update(ingestDocument, e);
+                }
+            }
+        }
+
+        batchGenerateAndSetEmbedding(ingestDocumentWrappers, docToSemanticFieldInfoMap, handler);
+    }
+
+    @SuppressWarnings("unchecked")
+    private void generateEmbedding(
```
junqiu-lei, line 444: For generateEmbedding, if calls for different modelIds are independent, explore if failures can be isolated per model ID. The onSuccess callback takes Map<Pair<String, String>, Object> modelIdValueToEmbeddingMap. If inference for one model ID fails, could the processor continue with other model IDs and report partial success/failure? This would make the batch processing more resilient. The current aggregation of exceptions in subBatchExecute (via addExceptionToImpactedDocs) is good for document-level errors, but this is about error handling within the generateEmbedding step for multiple models.

src/main/java/org/opensearch/neuralsearch/processor/factory/SemanticFieldProcessorFactory.java
q-andy: nvm, this is mitigated by system ingest pipeline caching in core

src/main/java/org/opensearch/neuralsearch/processor/dto/SemanticFieldInfo.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * The chunked strings of the original string value of the semantic field
+     */
+    private List<String> chunks;
+
+    /**
+     * @return full path to the chunks field of the semantic field
+     */
+    public String getFullPathForChunks() {
+        return semanticInfoFullPath + PATH_SEPARATOR + CHUNKS_FIELD_NAME;
```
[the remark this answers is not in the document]
bzhangam: I don't feel String.format is cleaner here since we don't mix plain string and variable to build the string. Same for the below comments. Do you see any other benefit to use String.format?

```
String.format(Locale.ROOT, "%s%s%s",semanticInfoFullPath, PATH_SEPARATOR,CHUNKS_FIELD_NAME);
```

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
bzhangam: Not sure how can we get what value is common used by customers. It's a general recommended value when I search online. I think we will do a fast follow up to make the text chunking more configurable.

src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
bzhangam: typo. Will remove.

src/main/java/org/opensearch/neuralsearch/processor/dto/SemanticFieldInfo.java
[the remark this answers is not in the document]
bzhangam: yeah we can do that

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/processor/dto/SemanticFieldInfo.java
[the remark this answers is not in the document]
q-andy: That's true, this is a nit. In these cases since you're concatenating based on `PATH_SEPARATOR`, what about `String.join(PATH_SEPARATOR, List.of(...)`?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -316,6 +321,57 @@ public void getModel(@NonNull final String modelId, @NonNull final ActionListene
...
+     */
+    public void getModels(
+        @NonNull final Set<String> modelIds,
+        @NonNull final Consumer<Map<String, MLModel>> onSuccess,
+        @NonNull final Consumer<Exception> onFailure
+    ) {
+        if (modelIds.isEmpty()) {
+            try {
+                onSuccess.accept(Collections.emptyMap());
+            } catch (Exception e) {
```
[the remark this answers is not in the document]
bzhangam: I think onSuccess.accept(Collections.emptyMap()) will invoke the callback function and for sure that callback function can run into exceptions.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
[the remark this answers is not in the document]
bzhangam: We should only call `onFailure.accept(e)` when all the getModel requests are done otherwise we can invoke the call back function multiple times. e.g. If we have two model ids and one success and another fail. Then we can invoke the onSuccess and the onFailure.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
bzhangam: I think we are not actually using the type for semantic processor and semantic processor factory. I follow the existing codes to set two constants with the same value. But ideally I think we should keep them different in case in future we want semantic processor factory to support different semantic processors. So will update the type of the factory to make it more clear it is a factory.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -0,0 +1,547 @@
...
+ * This processor is for internal usage and will be systematically invoked if we detect there are semantic fields
+ * defined. Users should not be able to define this processor in a regular ingest pipeline.
+ */
+@Log4j2
+public class SemanticFieldProcessor extends AbstractBatchingSystemProcessor {
+
+    public static final String PROCESSOR_TYPE = "system_ingest_processor_semantic_field";
+
+    private final Map<String, Map<String, Object>> pathToFieldConfig;
+    private final Map<String, MLModel> modelIdToModelMap = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
bzhangam: Previous I was modifying them in multiple threads. But now seems like they are only used in single thread so I think I can change them to hashMap.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -0,0 +1,547 @@
...
+        @NonNull final Environment environment,
+        @NonNull final ClusterService clusterService
+    ) {
+        super(tag, description, batchSize);
+        this.pathToFieldConfig = pathToFieldConfig;
+        this.mlCommonsClientAccessor = mlClientAccessor;
+        this.environment = environment;
+        this.clusterService = clusterService;
+        this.analysisRegistry = analysisRegistry;
+        this.chunker = createChunker();
```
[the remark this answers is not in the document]
bzhangam: I think we can do that. But would like to wait later when we make the chunker configurable to see what's the best way to orgnize the chunker related codes. I'm still not sure how will we support the configuration and also not sure if we want to use different chunking configuration for different semantic fields if one index has multiple semantic fields.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -0,0 +1,547 @@
...
+    }
+
+    /**
+     * Create a default text chunker.
+     * TODO: Make it configurable
+     * @return A default fixed token length chunker
+     */
+    private Chunker createChunker() {
+        final Map<String, Object> chunkerParameters = new HashMap<>();
+        chunkerParameters.put(FixedTokenLengthChunker.TOKEN_LIMIT_FIELD, 50);
```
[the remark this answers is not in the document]
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
bzhangam: That's a good point. I'll throw an exception.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
bzhangam: Yeah. We will call this out as a limitation.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
bzhangam: It's handled in the root caller execute function. And I do miss the try catch for the batch case and I'll add it.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -0,0 +1,547 @@
...
+            }, e -> handler.accept(null, e));
+        }
+    }
+
+    private void process(
+        @NonNull final IngestDocument ingestDocument,
+        @NonNull final List<SemanticFieldInfo> semanticFieldInfoList,
+        @NonNull final BiConsumer<IngestDocument, Exception> handler
+    ) {
+        setModelInfo(ingestDocument, semanticFieldInfoList);
```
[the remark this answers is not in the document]
bzhangam: No. The unexpected exception will be handled by the caller.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -0,0 +1,547 @@
...
+    private void setInference(
+        @NonNull final IngestDocument ingestDocument,
+        @NonNull final List<SemanticFieldInfo> semanticFieldInfoList,
+        @NonNull final Map<Pair<String, String>, Object> modelIdValueToEmbeddingMap
+    ) {
+        for (final SemanticFieldInfo semanticFieldInfo : semanticFieldInfoList) {
+            final String modelId = semanticFieldInfo.getModelId();
+            final boolean isDenseModel = isDenseModel(modelIdToModelTypeMap.get(modelId));
+            final List<String> chunks = semanticFieldInfo.getChunks();
+            for (int i = 0; i < chunks.size(); i++) {
```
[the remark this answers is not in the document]
bzhangam: I think it's better to use i or index.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -0,0 +1,547 @@
...
+                final String chunk = chunks.get(i);
+                Object embedding = modelIdValueToEmbeddingMap.get(Pair.of(modelId, chunk));
+                // TODO: In future we should allow user to configure how we should prune the sparse embedding
+                // for each semantic field. Then we can pull the config from the semantic config and use it here.
+                if (!isDenseModel) {
+                    embedding = PruneUtils.pruneSparseVector(PruneType.MAX_RATIO, 0.1f, (Map<String, Float>) embedding);
```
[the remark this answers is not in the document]
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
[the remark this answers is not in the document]
heemin32: Can we make sure that getModel will always call the listener? Otherwise, counter.decrementAndGet() won't be 0.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -0,0 +1,547 @@
...
+    @Override
+    public void subBatchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers, Consumer<List<IngestDocumentWrapper>> handler) {
+        if (ingestDocumentWrappers == null || ingestDocumentWrappers.isEmpty()) {
+            handler.accept(ingestDocumentWrappers);
+            return;
+        }
+        final Map<IngestDocumentWrapper, List<SemanticFieldInfo>> docToSemanticFieldInfoMap = new HashMap<>();
+        for (final IngestDocumentWrapper ingestDocumentWrapper : ingestDocumentWrappers) {
+            final IngestDocument ingestDocument = ingestDocumentWrapper.getIngestDocument();
+            if (ingestDocument == null) {
```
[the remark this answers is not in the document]
bzhangam: Ideally I think it should not be null but don't want that edge case fail the request. If null doc is an invalid case I think it's better to rely on other logic to catch it. For semantic field I think it's simpler that we do a no-op.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -0,0 +1,547 @@
...
+                continue;
+            }
+            unflattenIngestDoc(ingestDocument);
+            final List<SemanticFieldInfo> semanticFieldInfoList = getSemanticFieldInfo(ingestDocument);
+            if (!semanticFieldInfoList.isEmpty()) {
```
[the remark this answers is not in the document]
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
bzhangam: sure

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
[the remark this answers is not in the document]
heemin32: It is not just about configurability. It is more about making it easy to write unit test by mocking the chunker.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
heemin32: This is tricky part; who is responsible to handle exception. I will leave it to you as there is no way to enforce one way or another: caller or receiver of handler

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/util/ProcessorDocumentUtils.java
```diff
@@ -288,6 +289,17 @@ private static void flattenAndFlip(String path, Map<String, Object> map, Map<Str
...
+        Map<String, Object> sourceAndMetadataMap = document.getSourceAndMetadata();
+        Map<String, Object> unflattened = ProcessorDocumentUtils.unflattenJson(sourceAndMetadataMap);
+        unflattened.forEach(document::setFieldValue);
+        sourceAndMetadataMap.keySet().removeIf(key -> key.contains("."));
```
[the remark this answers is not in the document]
bzhangam: I think `.unflattenJson` only add the unflattened fields to the map but not remove the existing ones. So need to do this to remove it.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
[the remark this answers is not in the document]
bzhangam: At the end the listener is passed to the ML client `this.client.execute(MLModelGetAction.INSTANCE, mlModelGetRequest, this.getMlGetModelResponseActionListener(listener));` and it should be invoked unless ml client introduce some bugs. I think we still can add a try catch here in case some unexpected issue happen.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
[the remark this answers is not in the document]
bzhangam: ic. Let me move it outside.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -316,6 +321,64 @@ public void getModel(@NonNull final String modelId, @NonNull final ActionListene
...
+                    errors.add("Failed to fetch model [" + modelId + "]: " + e.getMessage());
+                    if (counter.decrementAndGet() == 0) {
+                        onFailure.accept(new RuntimeException(String.join(";", errors)));
+                    }
+                }));
+            }
+        } catch (Exception e) {
+            // If we run into unexpected exception force set the counter as -1 to void invoking onFailure multiple times
+            // in different threads and ensure we will invoke onFailure in the current thread.
+            counter.set(-1);
```
heemin32, line 377: Or we can follow same pattern?
```
for(::) {
try{
} catch() {
  hasError.set(true);
                    errors.add("Failed to fetch model [" + modelId + "]: " + e.getMessage());
                    if (counter.decrementAndGet() == 0) {
                        onFailure.accept(new RuntimeException(String.join(";", errors)));
                    }
                }));
}
}
```

junqiu-lei (approved):

PR #1315, opened 2025-05-09
Implement the query logic for the semantic field.

bzhangam:
### Description
Implement the query logic for the semantic field.

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
+++ b/src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
@@ -21,22 +21,23 @@ import static org.opensearch.neuralsearch.query.NeuralQueryBuilder.QUERY_IMAGE_F
  */
 public final class MinClusterVersionUtil {
 
-    private static final Version MINIMAL_SUPPORTED_VERSION_DEFAULT_MODEL_ID = Version.V_2_11_0;
+    private static final Version MINIMAL_SUPPORTED_VERSION_DEFAULT_DENSE_MODEL_ID = Version.V_2_11_0;
     private static final Version MINIMAL_SUPPORTED_VERSION_RADIAL_SEARCH = Version.V_2_14_0;
     private static final Version MINIMAL_SUPPORTED_VERSION_QUERY_IMAGE_FIX = Version.V_2_19_0;
     private static final Version MINIMAL_SUPPORTED_VERSION_PAGINATION_IN_HYBRID_QUERY = Version.V_2_19_0;
     private static final Version MINIMAL_SUPPORTED_VERSION_NEURAL_ORIGINAL_QUERY_TEXT = Version.V_3_0_0;
+    public static final Version MINIMAL_SUPPORTED_VERSION_SEMANTIC_FIELD = Version.V_3_1_0;
 
-    // Note this minimal version will act as a override
+    // Note this minimal version will act as an override
     private static final Map<String, Version> MINIMAL_VERSION_NEURAL = ImmutableMap.<String, Version>builder()
-        .put(MODEL_ID_FIELD.getPreferredName(), MINIMAL_SUPPORTED_VERSION_DEFAULT_MODEL_ID)
+        .put(MODEL_ID_FIELD.getPreferredName(), MINIMAL_SUPPORTED_VERSION_DEFAULT_DENSE_MODEL_ID)
         .put(MAX_DISTANCE_FIELD.getPreferredName(), MINIMAL_SUPPORTED_VERSION_RADIAL_SEARCH)
         .put(MIN_SCORE_FIELD.getPreferredName(), MINIMAL_SUPPORTED_VERSION_RADIAL_SEARCH)
         .put(QUERY_IMAGE_FIELD.getPreferredName(), MINIMAL_SUPPORTED_VERSION_QUERY_IMAGE_FIX)
         .build();
 
-    public static boolean isClusterOnOrAfterMinReqVersionForDefaultModelIdSupport() {
-        return NeuralSearchClusterUtil.instance().getClusterMinVersion().onOrAfter(MINIMAL_SUPPORTED_VERSION_DEFAULT_MODEL_ID);
+    public static boolean isClusterOnOrAfterMinReqVersionForDefaultDenseModelIdSupport() {
+        return NeuralSearchClusterUtil.instance().getClusterMinVersion().onOrAfter(MINIMAL_SUPPORTED_VERSION_DEFAULT_DENSE_MODEL_ID);
     }
 
     public static boolean isClusterOnOrAfterMinReqVersionForRadialSearch() {
@@ -65,5 +66,14 @@ public final class MinClusterVersionUtil {
      */
     public static boolean isVersionOnOrAfterMinReqVersionForNeuralKNNQueryText(Version version) {
         return version.onOrAfter(MINIMAL_SUPPORTED_VERSION_NEURAL_ORIGINAL_QUERY_TEXT);
+    }
+
+    /**
+     * Checks if the cluster min version is on or after the minimum required version for semantic field type
+     *
+     * @return true if the version is on or after the minimum required version
+     */
+    public static boolean isClusterOnOrAfterMinReqVersionForSemanticFieldType() {
+        return NeuralSearchClusterUtil.instance().getClusterMinVersion().onOrAfter(MINIMAL_SUPPORTED_VERSION_SEMANTIC_FIELD);
     }
 }
--- a/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
@@ -10,23 +10,52 @@ import static org.opensearch.knn.index.query.KNNQueryBuilder.MAX_DISTANCE_FIELD;
 import static org.opensearch.knn.index.query.KNNQueryBuilder.METHOD_PARAMS_FIELD;
 import static org.opensearch.knn.index.query.KNNQueryBuilder.MIN_SCORE_FIELD;
 import static org.opensearch.knn.index.query.KNNQueryBuilder.RESCORE_FIELD;
+import static org.opensearch.neuralsearch.common.MinClusterVersionUtil.MINIMAL_SUPPORTED_VERSION_SEMANTIC_FIELD;
 import static org.opensearch.neuralsearch.common.MinClusterVersionUtil.isClusterOnOrAfterMinReqVersion;
-import static org.opensearch.neuralsearch.common.MinClusterVersionUtil.isClusterOnOrAfterMinReqVersionForDefaultModelIdSupport;
+import static org.opensearch.neuralsearch.common.MinClusterVersionUtil.isClusterOnOrAfterMinReqVersionForDefaultDenseModelIdSupport;
 import static org.opensearch.neuralsearch.common.MinClusterVersionUtil.isClusterOnOrAfterMinReqVersionForRadialSearch;
+import static org.opensearch.neuralsearch.common.MinClusterVersionUtil.isClusterOnOrAfterMinReqVersionForSemanticFieldType;
 import static org.opensearch.neuralsearch.common.VectorUtil.vectorAsListToArray;
+import static org.opensearch.neuralsearch.constants.MappingConstants.PATH_SEPARATOR;
+import static org.opensearch.neuralsearch.constants.SemanticFieldConstants.DEFAULT_SEMANTIC_INFO_FIELD_NAME_SUFFIX;
+import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.CHUNKS_EMBEDDING_FIELD_NAME;
+import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.CHUNKS_FIELD_NAME;
 import static org.opensearch.neuralsearch.processor.TextImageEmbeddingProcessor.INPUT_IMAGE;
 import static org.opensearch.neuralsearch.processor.TextImageEmbeddingProcessor.INPUT_TEXT;
+import static org.opensearch.neuralsearch.query.parser.NeuralQueryParser.modelIdToQueryTokensSupplierMapStreamInput;
+import static org.opensearch.neuralsearch.query.parser.NeuralQueryParser.modelIdToQueryTokensSupplierMapStreamOutput;
+import static org.opensearch.neuralsearch.query.parser.NeuralQueryParser.modelIdToVectorSupplierMapStreamInput;
+import static org.opensearch.neuralsearch.query.parser.NeuralQueryParser.modelIdToVectorSupplierMapStreamOutput;
+import static org.opensearch.neuralsearch.query.parser.NeuralQueryParser.queryTokensMapSupplierStreamInput;
+import static org.opensearch.neuralsearch.query.parser.NeuralQueryParser.queryTokensMapSupplierStreamOutput;
+import static org.opensearch.neuralsearch.query.parser.NeuralQueryParser.vectorSupplierStreamInput;
+import static org.opensearch.neuralsearch.query.parser.NeuralQueryParser.vectorSupplierStreamOutput;
+import static org.opensearch.neuralsearch.util.NeuralQueryValidationUtil.validateNeuralQueryForKnn;
+import static org.opensearch.neuralsearch.util.NeuralQueryValidationUtil.validateNeuralQueryForSemanticDense;
+import static org.opensearch.neuralsearch.util.NeuralQueryValidationUtil.validateNeuralQueryForSemanticSparse;
+import static org.opensearch.neuralsearch.util.NeuralQueryValidationUtil.validateTargetFieldConfig;
+import static org.opensearch.neuralsearch.util.SemanticMappingUtils.getIndexToTargetFieldConfigMapFromIndexMetadata;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 
+import lombok.NonNull;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.lucene.search.MatchNoDocsQuery;
 import org.apache.lucene.search.Query;
+import org.apache.lucene.search.join.ScoreMode;
+import org.opensearch.action.IndicesRequest;
+import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.common.SetOnce;
 import org.opensearch.core.ParseField;
 import org.opensearch.core.action.ActionListener;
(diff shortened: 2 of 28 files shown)


src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -472,8 +607,152 @@ private static void parseQueryParams(XContentParser parser, NeuralQueryBuilder n
...
+            final float[] vector = modelIdToVectorSupplierMap.get(searchModelId).get();
+            final NeuralKNNQueryBuilder neuralKNNQueryBuilder = createKNNQueryBuilder(vector);
+            return new NestedQueryBuilder(nestedQueryPath, neuralKNNQueryBuilder, ScoreMode.Max);
+        } else if (RankFeaturesFieldMapper.CONTENT_TYPE.equals(embeddingFieldType)) {
+            Supplier<Map<String, Float>> queryTokensSupplier = queryTokensMapSupplier;
+            // If the raw token is not provided then try to find the token generated by the ml model
+            if (queryTokensSupplier == null) {
+                if (modelIdToQueryTokensSupplierMap == null || modelIdToQueryTokensSupplierMap.get(searchModelId) == null) {
+                    throw new RuntimeException(
+                        getErrorMessageWithBaseErrorForSemantic(
```
heemin32, line 708: We don't expect this exception happen and if this exception happen, it mean there is a bug in the program?

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -387,81 +533,70 @@ protected void doXContent(XContentBuilder xContentBuilder, Params params) throws
      * @throws IOException can be thrown by parser
      */
     public static NeuralQueryBuilder fromXContent(XContentParser parser) throws IOException {
-        NeuralQueryBuilder neuralQueryBuilder = new NeuralQueryBuilder();
+        final Builder builder = new Builder();
```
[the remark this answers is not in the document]
bzhangam: We centralize the validation and set the default k logic into the build function of the builder to make the code cleaner.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -472,8 +607,152 @@ private static void parseQueryParams(XContentParser parser, NeuralQueryBuilder n
...
+        // coordinate rewrite, and we do not need to rewrite the neural query for this case. We will do rewrite later
+        // with queryShardContext.
+        return this;
+    }
+
+    private QueryBuilder rewriteQueryWithCoordinatorContext(@NonNull final QueryCoordinatorContext queryRewriteContext) {
+        final IndicesRequest searchRequest = queryRewriteContext.getSearchRequest();
+        final List<String> remoteIndices = getRemoteIndices(searchRequest);
+        // If we have remote target indices in the search request we assume the target field should be a knn field
+        // for backward compatibility. For semantic field we do not support cross cluster search for now.
```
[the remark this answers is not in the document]
bzhangam: We will actually try to query against a KNN field in the remote cluster but that target field is a semantic field so we will complain cannot run knn query against a semantic field.

src/main/java/org/opensearch/neuralsearch/query/parser/NeuralQueryParser.java
```diff
@@ -0,0 +1,173 @@
...
+     */
+    public static void queryTokensMapSupplierStreamOutput(
+        @NonNull final StreamOutput out,
+        final Supplier<Map<String, Float>> queryTokensMapSupplier
+    ) throws IOException {
+        if (queryTokensMapSupplier == null || queryTokensMapSupplier.get() == null) {
+            out.writeBoolean(false);
+        } else {
+            out.writeBoolean(true);
+            out.writeMap(queryTokensMapSupplier.get(), StreamOutput::writeString, StreamOutput::writeFloat);
```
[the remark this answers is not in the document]
bzhangam: no

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
bzhangam: correct.

src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
```diff
@@ -66,4 +66,23 @@ public static boolean isClusterOnOrAfterMinReqVersion(String key) {
...
         return version.onOrAfter(MINIMAL_SUPPORTED_VERSION_NEURAL_ORIGINAL_QUERY_TEXT);
     }
+
+    /**
+     * Checks if the cluster min version is on or after the minimum required version for semantic field type
+     *
+     * @return true if the version is on or after the minimum required version
+     */
+    public static boolean isClusterOnOrAfterMinReqVersionForSemanticFieldType() {
+        return NeuralSearchClusterUtil.instance().getClusterMinVersion().onOrAfter(Version.V_3_1_0);
```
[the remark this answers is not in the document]
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -308,6 +444,13 @@ protected void doWriteTo(StreamOutput out) throws IOException {
             MethodParametersParser.streamOutput(out, methodParameters, MinClusterVersionUtil::isClusterOnOrAfterMinReqVersion);
         }
         RescoreParser.streamOutput(out, rescoreContext);
+
+        if (isClusterOnOrAfterMinReqVersionForSemanticFieldType(out.getVersion())) {
```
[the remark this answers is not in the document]
heemin32: If that is the case, the cross cluster search might have broken multiple features as we have been checking min cluster version only in the past. I don't think the transport level request will be sent from one cluster to another?

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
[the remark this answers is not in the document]
heemin32: Can we throw more user friendly exception to let them know that semantic field query is not supported for cross cluster?

src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
```diff
@@ -66,4 +66,23 @@ public static boolean isClusterOnOrAfterMinReqVersion(String key) {
...
+    }
+
+    /**
+     * Checks if the cluster min version and stream version is on or after the minimum required version for semantic
+     * field type. In this function we check the stream version to handle the cross cluster search case.
+     *
+     * @return true if the version is on or after the minimum required version
+     */
+    public static boolean isClusterOnOrAfterMinReqVersionForSemanticFieldType(Version version) {
+        return version.onOrAfter(Version.V_3_1_0) && isClusterOnOrAfterMinReqVersionForSemanticFieldType();
```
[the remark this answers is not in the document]
bzhangam: I think for stream we just need to check the stream version and no need to check the min cluster version so I'll remove this function and check it directly.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -308,6 +445,13 @@ protected void doWriteTo(StreamOutput out) throws IOException {
             MethodParametersParser.streamOutput(out, methodParameters, MinClusterVersionUtil::isClusterOnOrAfterMinReqVersion);
         }
         RescoreParser.streamOutput(out, rescoreContext);
+
+        if (out.getVersion().onOrAfter(MINIMAL_SUPPORTED_VERSION_SEMANTIC_FIELD)) {
```
heemin32, line 449: I think out.getVersion() is opensearch version of current node. Therefore, better to use `isClusterOnOrAfterMinReqVersionForSemanticField()` instead.

bzhangam: It's the version of the target node. In OpenSearch core it's common that we just check the stream version.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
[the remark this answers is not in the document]
bzhangam: Yeah I think that's true. But probably it's common that users will keep all of their clusters on the same version so that they don't see the issue often.

For CCS we will try to do rewrite on the coordinate level first and then decide if we pass the request to the remote cluster on transport layer or shard layer(depends on the search configuration but in most of the cases I think we will do shard layer.) - related [code](https://github.com/opensearch-project/OpenSearch/blob/1c4ac4639f628371829d2b7c4f786525cd25fc14/server/src/main/java/org/opensearch/action/search/TransportSearchAction.java#L530) in core.

src/main/java/org/opensearch/neuralsearch/query/parser/NeuralQueryParser.java
```diff
@@ -0,0 +1,173 @@
...
+            return;
+        }
+        out.writeBoolean(true);
+
+        int size = modelIdToVectorSupplierMap.size();
+        out.writeVInt(size);
+
+        for (Map.Entry<String, Supplier<float[]>> entry : modelIdToVectorSupplierMap.entrySet()) {
+            out.writeString(entry.getKey());
+            out.writeFloatArray(entry.getValue().get());
```
junqiu-lei, line 92: Shall we need considering NPE here?

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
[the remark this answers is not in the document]
heemin32: Got it. Still, I think using cluster level version is better. For example,
When the data is `xyz` and we serialize `y` only when the version is 3.1 and above, the receiving side which is older version of 3.1 might not know about `y` and try to deserialize as `xz` which will fail.

src/main/java/org/opensearch/neuralsearch/query/parser/NeuralQueryParser.java
```diff
@@ -0,0 +1,173 @@
...
+            return;
+        }
+        out.writeBoolean(true);
+
+        int size = modelIdToQueryTokensMapSupplier.size();
+        out.writeVInt(size);
+
+        for (Map.Entry<String, Supplier<Map<String, Float>>> entry : modelIdToQueryTokensMapSupplier.entrySet()) {
+            out.writeString(entry.getKey());
+            out.writeMap(entry.getValue().get(), StreamOutput::writeString, StreamOutput::writeFloat);
```
junqiu-lei, line 170: Same, potential NPE.

src/main/java/org/opensearch/neuralsearch/util/NeuralQueryValidationUtil.java
```diff
@@ -0,0 +1,274 @@
...
+        return errors;
+    }
+
+    private static boolean isModelIdRequiredForNeuralQueryForKnn(final NeuralQueryBuildStage buildStage) {
+        // If we build the NeuralQueryBuilder targeting a knn field in the coordinate rewrite then it must have a model id.
+        // If we build the NeuralQueryBuilder targeting a knn field in the fromXContent function then it must have a
+        // model id if it cannot support the default model id otherwise the model id is optional.
+        return NeuralQueryBuildStage.REWRITE.equals(buildStage)
+            || ((buildStage == null || NeuralQueryBuildStage.FROM_X_CONTENT.equals(buildStage))
+                && isClusterOnOrAfterMinReqVersionForDefaultDenseModelIdSupport() == false);
```
junqiu-lei, line 56: The nested conditions are too complex for readability, maybe like:

```suggestion
    // During rewrite stage, model ID is always required
    if (NeuralQueryBuildStage.REWRITE.equals(buildStage)) {
        return true;
    }
    
    // During FROM_X_CONTENT or unspecified stage, model ID is required if default model ID isn't supported
    boolean isFreshBuild = (buildStage == null || NeuralQueryBuildStage.FROM_X_CONTENT.equals(buildStage));
    boolean supportsDefaultModel = isClusterOnOrAfterMinReqVersionForDefaultDenseModelIdSupport();
    
    return isFreshBuild && !supportsDefaultModel;
```

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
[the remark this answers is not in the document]
heemin32: If `out.getVersion()` is a version of destination node, I guess this should be fine..

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -199,26 +269,87 @@ public Builder rescoreContext(RescoreContext rescoreContext) {
...
+            this.modelIdToQueryTokensSupplierMap = modelIdToQueryTokensSupplierMap;
+            return this;
+        }
+
+        public Builder isSemanticField(Boolean isSemanticField) {
+            this.isSemanticField = isSemanticField;
+            return this;
+        }
+
+        public Builder builStage(NeuralQueryBuildStage buildStage) {
```
junqiu-lei, line 297: Typo: buildStage

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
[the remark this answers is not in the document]
bzhangam: Yeah. I think it's cleaner to just check the stream version since it's the version of the destination node.

heemin32 (approved):
LGTM. Thanks.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -472,8 +608,152 @@ private static void parseQueryParams(XContentParser parser, NeuralQueryBuilder n
...
+            if (modelIdToVectorSupplierMap == null
+                || modelIdToVectorSupplierMap.get(searchModelId) == null
+                || modelIdToVectorSupplierMap.get(searchModelId).get() == null) {
+                throw new RuntimeException(
+                    getErrorMessageWithBaseErrorForSemantic("Not able to find the dense embedding when try to rewrite it to the KNN query.")
+                );
+            }
+            final float[] vector = modelIdToVectorSupplierMap.get(searchModelId).get();
+            final NeuralKNNQueryBuilder neuralKNNQueryBuilder = createKNNQueryBuilder(vector);
+            return new NestedQueryBuilder(nestedQueryPath, neuralKNNQueryBuilder, ScoreMode.Max);
```
junqiu-lei, line 702: The semantic highlighter doesn't have nested query pattern used by semantic fields, we need it also registered and check how to adapt its support in https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/highlight/extractor/QueryTextExtractorRegistry.java#L36-L44 for the new semantic field type.

Also we need verify that the original query text into highlighter is properly preserved and passed through the nested query structure created for semantic fields.

src/main/java/org/opensearch/neuralsearch/util/NeuralQueryValidationUtil.java
```diff
@@ -0,0 +1,274 @@
...
+                        String.join(", ", indicesWithSemanticSparse)
+
+                    )
+                );
+            }
+        }
+
+        if (!validationErrors.isEmpty()) {
+            throw new IllegalArgumentException(
+                "Invalid neural query against the field " + fieldName + ". Errors: " + String.join("; ", validationErrors)
```
junqiu-lei, line 269: Use `String.format`

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -199,26 +269,87 @@ public Builder rescoreContext(RescoreContext rescoreContext) {
...
+                if (isSemanticField == null || isSemanticField == false) {
+                    // Currently if the target field is not a semantic field we will assume the target field should be a knn
+                    // field. In the future we can add the support for the case when the target field is a rank_features.
+                    errors = validateNeuralQueryForKnn(neuralQueryBuilder, buildStage);
+                } else if (KNNVectorFieldMapper.CONTENT_TYPE.equals(embeddingFieldType)) {
+                    errors = validateNeuralQueryForSemanticDense(neuralQueryBuilder);
+                } else if (RankFeaturesFieldMapper.CONTENT_TYPE.equals(embeddingFieldType)) {
+                    errors = validateNeuralQueryForSemanticSparse(neuralQueryBuilder);
+                } else {
+                    throw new IllegalArgumentException("Unsupported embedding field type: " + embeddingFieldType);
```
junqiu-lei, line 344: same

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -199,26 +269,87 @@ public Builder rescoreContext(RescoreContext rescoreContext) {
...
+                }
+            }
+
+            if (!errors.isEmpty()) {
+                throw new IllegalArgumentException("Failed to build the NeuralQueryBuilder: " + String.join("; ", errors));
```
junqiu-lei, line 349: same

src/main/java/org/opensearch/neuralsearch/query/parser/NeuralQueryParser.java
bzhangam: I think we should not run into NPE here but I can add a check in case there is bug.

src/main/java/org/opensearch/neuralsearch/util/NeuralQueryValidationUtil.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/util/NeuralQueryValidationUtil.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
bzhangam: I'll handle the case when we use a simple string field as the source field and the ML model is a dense model for the semantic field.

Want to call out that:
We cannot support the case when the source field is a nested field which can make the source field not a string but a list of items which our current highlighter set up cannot handle.

And we also cannot support the sparse model.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
bzhangam: ok

junqiu-lei (approved):

PR #1317, opened 2025-05-10
Support remote model type semantic sentence highlighting

junqiu-lei:
### Description
This PR adds remote model type support for semantic sentence highlighting. It uses guava cache strategy to let ml client choose right model type for inference request.

### Related Issues
Follow up with https://github.com/opensearch-project/neural-search/issues/1182

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
@@ -10,6 +10,7 @@ import static org.opensearch.neuralsearch.processor.TextImageEmbeddingProcessor.
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -27,6 +28,7 @@ import org.opensearch.ml.common.MLModel;
 import org.opensearch.ml.common.dataset.MLInputDataset;
 import org.opensearch.ml.common.dataset.TextDocsInputDataSet;
 import org.opensearch.ml.common.dataset.TextSimilarityInputDataSet;
+import org.opensearch.ml.common.dataset.remote.RemoteInferenceInputDataSet;
 import org.opensearch.ml.common.input.MLInput;
 import org.opensearch.ml.common.output.MLOutput;
 import org.opensearch.ml.common.output.model.ModelResultFilter;
@@ -45,6 +47,10 @@ import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.log4j.Log4j2;
 
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import java.util.concurrent.TimeUnit;
+
 /**
  * This class will act as an abstraction on the MLCommons client for accessing the ML Capabilities
  */
@@ -52,6 +58,10 @@ import lombok.extern.log4j.Log4j2;
 @Log4j2
 public class MLCommonsClientAccessor {
     private final MachineLearningNodeClient mlClient;
+
+    // Define constants for cache configuration
+    private static final long DEFAULT_MODEL_TYPE_CACHE_MAX_SIZE = 200L;
+    private static final long DEFAULT_MODEL_TYPE_CACHE_EXPIRE_AFTER_ACCESS_DAYS = 30L;
 
     /**
      * Wrapper around {@link #inferenceSentences} that expected a single input text and produces a single floating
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -398,38 +408,107 @@ private void retryableGetModel(@NonNull final String modelId, final int retryTim
...
+
+            if (algorithm != null) {
+                // Cache hit
+                processSentenceHighlightingWithKnownAlgorithm(inferenceRequest, algorithm, retryTime, listener);
+            } else {
+                // Cache miss or entry expired/evicted
+                // Fetch model metadata
+                getModel(modelId, ActionListener.wrap(mlModel -> {
+                    FunctionName fetchedAlgorithm = mlModel.getAlgorithm();
+                    if (fetchedAlgorithm != null) { // Avoid caching nulls unless specifically intended
```
bzhangam, line 449: In what case the algorithm can be null? Should we throw exception if we run into null?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -398,38 +408,107 @@ private void retryableGetModel(@NonNull final String modelId, final int retryTim
...
 
     /**
-     * Retryable method to perform sentence highlighting inference.
-     * This method will retry up to 3 times if a retryable exception occurs.
+     * Cache to store the {@link FunctionName} (algorithm) for a given model ID.
+     * This helps to avoid redundant calls to fetch model metadata (like its algorithm type)
+     * for sentence highlighting operations. The cache has a maximum size and entries expire
+     * after a defined period of inactivity.
+     */
+    private final Cache<String, FunctionName> modelTypeCache = CacheBuilder.newBuilder()
```
bzhangam, line 416: Should we consider to invalidate the cache when the model is deleted? I know it's a very edge case that:
1. We cache a local model with id x.
2. The we delete the model x.
3. We create a remote model and it happens its id is x. (Very low chance case).
4. Then we will use incorrect logic to process the remote model.

How can we mitigate this issue? If we run into it it's kind hard to debug and reproduce.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -398,38 +408,107 @@ private void retryableGetModel(@NonNull final String modelId, final int retryTim
...
+
+        if (algorithm == FunctionName.REMOTE) {
+            // For remote models, use RemoteInferenceInputDataSet
+            Map<String, String> parameters = new HashMap<>();
+            parameters.put(MLInput.QUESTION_FIELD, inferenceRequest.getQuestion());
+            parameters.put(MLInput.CONTEXT_FIELD, inferenceRequest.getContext());
+
+            MLInputDataset inputDataset = new RemoteInferenceInputDataSet(parameters);
+            mlInput = new MLInput(FunctionName.REMOTE, null, inputDataset);
+        } else {
```
bzhangam, line 488: What if a non QA model is provided? I think we should validate it to throw a clear exception for that case rather than blindly treat it as a QA model.

junqiu-lei:
Pending this PR update, as I'm looking for unified API call solution for both remote and local model.

junqiu-lei:
I've found a more elegant solution that allows us to use a unified API for both local and remote sentence highlighting models without needing additional code changes in the neural-search plugin.

## Solution Overview

The key insight is that ML-Commons already has a mechanism for determining whether a model is local or remote at runtime:

1. When neural-search calls ML-Commons with any request (including sentence highlighting), it sets the algorithm to `FunctionName.QUESTION_ANSWERING`

2. In ML-Commons' `TransportPredictionTaskAction.executePredict()`, the system resolves the true function name:
   ```java
   FunctionName functionName = modelCacheHelper
       .getOptionalFunctionName(modelId)
       .orElse(mlPredictionTaskRequest.getMlInput().getAlgorithm());
   ```
   - For local models, this keeps the algorithm as `QUESTION_ANSWERING`
   - For remote models, this changes it to `REMOTE`

3. ML-Commons then processes the request based on the resolved function name, automatically using either `QuestionAnsweringModel` or `RemoteModel`

## Remote Model Configuration

For remote sentence highlighting models, we just need a properly configured connector. The key is in the pre-processing function that correctly transforms the input:

```json
"pre_process_function": "// Extract question and context directly from params\nif (params.question != null && params.context != null) {\n    return '{\"parameters\":{\"question\":\"' + params.question + '\",\"context\":\"' + params.context + '\"}}'; \n} \nelse {\n    throw new IllegalArgumentException(\"Missing required parameters: question and context\");\n}"
```

This script extracts the question and context parameters and formats them for the remote endpoint, maintaining the same interface as local models.

## Benefits

1. **No Code Changes Needed**: Neural-search can use the same API call for both local and remote models
2. **Transparency**: The code handling the request doesn't need to know whether a model is local or remote
3. **Better Maintainability**: No need for model type caching or additional logic in neural-search
4. **Future Compatibility**: Works with any future model types without code changes

This approach leverages ML-Commons' existing architecture rather than implementing a parallel system in neural-search, resulting in cleaner, more maintainable code.

PR #1326, opened 2025-05-15
Add normalization processor and RRF processor stats

q-andy:
Draft PR to add stats for normalization processor and RRF processor.

Example response:
```
{
	"_nodes": {
		"total": 1,
		"successful": 1,
		"failed": 0
	},
	"cluster_name": "integTest",
	"info": {
		"cluster_version": "3.1.0",
		"processors": {
			"search": {
				"hybrid": {
					"comb_arithmetic_mean_processors": 0,
					"norm_l2_processors": 0,
					"norm_minmax_processors": 0,
					"comb_geometric_mean_processors": 0,
					"comb_technique_rrf_processors": 0,
					"rrf_processors": 0,
					"comb_harmonic_mean_processors": 0,
					"norm_zscore_processors": 0,
					"norm_rrf_processors": 0,
					"normalization_processors": 0
				}
			}
		}
	},
	"all_nodes": {
		"processors": {
			"search": {
				"hybrid": {
					"comb_harmonic_executions": 0,
					"rrf_processor_executions": 0,
					"norm_zscore_executions": 0,
					"norm_l2_executions": 0,
					"comb_rrf_executions": 0,
					"norm_rrf_executions": 0,
					"comb_arithmetic_executions": 0,
					"normalization_processor_explain_executions": 0,
					"normalization_processor_executions": 0,
					"comb_geometric_executions": 0,
					"norm_minmax_executions": 0,
					"rrf_processor_explain_executions": 0
				}
			}
		}
	}
}
```

### Description

### Related Issues

### Check List


src/main/java/org/opensearch/neuralsearch/processor/RRFProcessor.java
```diff
@@ -70,6 +72,7 @@ <Result extends SearchPhaseResult> void hybridizeScores(
         Optional<FetchSearchResult> fetchSearchResult = getFetchSearchResults(searchPhaseResult);
         boolean explain = Objects.nonNull(searchPhaseContext.getRequest().source().explain())
             && searchPhaseContext.getRequest().source().explain();
+        EventStatsManager.increment(EventStatName.RRF_PROCESSOR_EXECUTIONS);
```
martin-gaievski, line 75: for rank based normalization you also need to increment counter for combination technique calls, currently we do support "rrf" https://docs.opensearch.org/docs/latest/search-plugins/search-pipelines/score-ranker-processor/

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessor.java
```diff
@@ -135,4 +144,36 @@ private <Result extends SearchPhaseResult> Optional<FetchSearchResult> getFetchS
...
+    private void recordStats(ScoreNormalizationTechnique normalizationTechnique, ScoreCombinationTechnique combinationTechnique) {
+        EventStatsManager.increment(EventStatName.NORMALIZATION_PROCESSOR_EXECUTIONS);
+        recordNormalizationCombinationTechniqueStats(normalizationTechnique, combinationTechnique);
+    }
+
+    void recordNormalizationCombinationTechniqueStats(
+        ScoreNormalizationTechnique normalizationTechnique,
+        ScoreCombinationTechnique combinationTechnique
+    ) {
+        switch (normalizationTechnique.techniqueName()) {
```
martin-gaievski, line 157: nit: imho code is more clean if you use approach with map, where key is technique name and value is functional interface that does increment the corresponding counter, something like in factories for norm/comb techniques [ref](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombinationFactory.java)

src/main/java/org/opensearch/neuralsearch/stats/events/EventStatName.java
```diff
@@ -19,13 +19,28 @@
...
+        EventStatType.TIMESTAMPED_EVENT_COUNTER
+    ),
+    NORM_TECHNIQUE_L2_EXECUTIONS("norm_l2_executions", "processors.search.hybrid", EventStatType.TIMESTAMPED_EVENT_COUNTER),
+    NORM_TECHNIQUE_MINMAX_EXECUTIONS("norm_minmax_executions", "processors.search.hybrid", EventStatType.TIMESTAMPED_EVENT_COUNTER),
+    NORM_TECHNIQUE_NORM_ZSCORE_EXECUTIONS("norm_zscore_executions", "processors.search.hybrid", EventStatType.TIMESTAMPED_EVENT_COUNTER),
+    COMB_TECHNIQUE_ARITHMETIC_EXECUTIONS("comb_arithmetic_executions", "processors.search.hybrid", EventStatType.TIMESTAMPED_EVENT_COUNTER),
+    COMB_TECHNIQUE_GEOMETRIC_EXECUTIONS("comb_geometric_executions", "processors.search.hybrid", EventStatType.TIMESTAMPED_EVENT_COUNTER),
+    COMB_TECHNIQUE_HARMONIC_EXECUTIONS("comb_harmonic_executions", "processors.search.hybrid", EventStatType.TIMESTAMPED_EVENT_COUNTER),
+    // RRF processor stats
+    RRF_PROCESSOR_EXECUTIONS("rrf_processor_executions", "processors.search.hybrid", EventStatType.TIMESTAMPED_EVENT_COUNTER),;
```
martin-gaievski, line 43:
```suggestion
    RRF_PROCESSOR_EXECUTIONS("rank_based_normalization_processor_executions", "processors.search.hybrid", EventStatType.TIMESTAMPED_EVENT_COUNTER),;
```

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatName.java
```diff
@@ -21,9 +21,20 @@ public enum InfoStatName implements StatName {
...
+    // Normalization processor
+    NORMALIZATION_PROCESSORS("normalization_processors", "processors.search.hybrid", InfoStatType.INFO_COUNTER),
+    NORM_TECHNIQUE_L2_PROCESSORS("norm_l2_processors", "processors.search.hybrid", InfoStatType.INFO_COUNTER),
+    NORM_TECHNIQUE_MINMAX_PROCESSORS("norm_minmax_processors", "processors.search.hybrid", InfoStatType.INFO_COUNTER),
+    NORM_TECHNIQUE_ZSCORE_PROCESSORS("norm_zscore_processors", "processors.search.hybrid", InfoStatType.INFO_COUNTER),
+    COMB_TECHNIQUE_ARITHMETIC_PROCESSORS("comb_arithmetic_processors", "processors.search.hybrid", InfoStatType.INFO_COUNTER),
+    COMB_TECHNIQUE_GEOMETRIC_PROCESSORS("comb_geometric_processors", "processors.search.hybrid", InfoStatType.INFO_COUNTER),
+    COMB_TECHNIQUE_HARMONIC_PROCESSORS("comb_harmonic_processors", "processors.search.hybrid", InfoStatType.INFO_COUNTER),
+    // RRF processor
+    RRF_PROCESSORS("rrf_processors", "processors.search.hybrid", InfoStatType.INFO_COUNTER),;
```
martin-gaievski, line 37:
```suggestion
    RRF_PROCESSORS("rank_based_normalization_processors", "processors.search.hybrid", InfoStatType.INFO_COUNTER),;
```

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Enhancements
 - [Performance Improvement] Add custom bulk scorer for hybrid query (2-3x faster) ([#1289](https://github.com/opensearch-project/neural-search/pull/1289))
 - [Stats] Add stats for text chunking processor algorithms ([#1308](https://github.com/opensearch-project/neural-search/pull/1308))
+- [Stats] Add stats for normalization processor and RRF processor ([#1326](https://github.com/opensearch-project/neural-search/pull/1326))
```
martin-gaievski, line 17:
```suggestion
- [Stats] Add stats for score based and rank based normalization processors ([#1326](https://github.com/opensearch-project/neural-search/pull/1326))
```

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessor.java
```diff
@@ -135,4 +144,36 @@ private <Result extends SearchPhaseResult> Optional<FetchSearchResult> getFetchS
         Optional<Result> optionalFirstSearchPhaseResult = searchPhaseResults.getAtomicArray().asList().stream().findFirst();
         return optionalFirstSearchPhaseResult.map(SearchPhaseResult::fetchResult);
     }
+
+    private void recordStats(ScoreNormalizationTechnique normalizationTechnique, ScoreCombinationTechnique combinationTechnique) {
+        EventStatsManager.increment(EventStatName.NORMALIZATION_PROCESSOR_EXECUTIONS);
+        recordNormalizationCombinationTechniqueStats(normalizationTechnique, combinationTechnique);
+    }
+
+    void recordNormalizationCombinationTechniqueStats(
```
martin-gaievski, line 153: this method can be private

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessor.java
q-andy: Sure, see my recent changes. I added a helper to unwrap the nullable safely as well.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessor.java
```diff
@@ -40,6 +49,24 @@ public class NormalizationProcessor extends AbstractScoreHybridizationProcessor
     private final ScoreCombinationTechnique combinationTechnique;
     private final NormalizationProcessorWorkflow normalizationWorkflow;
 
+    private final Map<String, Runnable> normTechniqueIncrementers = Map.of(
```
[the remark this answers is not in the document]
vibrantvarun: +1. I am more inclined towards using map.

src/main/java/org/opensearch/neuralsearch/processor/RRFProcessor.java
```diff
@@ -143,4 +153,9 @@ <Result extends SearchPhaseResult> Optional<FetchSearchResult> getFetchSearchRes
         Optional<Result> optionalFirstSearchPhaseResult = searchPhaseResults.getAtomicArray().asList().stream().findFirst();
         return optionalFirstSearchPhaseResult.map(SearchPhaseResult::fetchResult);
     }
+
+    private void recordStats(ScoreCombinationTechnique combinationTechnique) {
+        EventStatsManager.increment(EventStatName.RRF_PROCESSOR_EXECUTIONS);
+        Optional.of(combTechniqueIncrementers.get(combinationTechnique.techniqueName())).ifPresent(Runnable::run);
```
vibrantvarun, line 159: RRF processor internally creates RRFNormalizationProcessor also. Check https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/factory/RRFProcessorFactory.java#L51

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -270,13 +273,26 @@ public static HybridQueryBuilder fromXContent(XContentParser parser) throws IOEx
...
+        boolean hasInnerHits = false;
         for (QueryBuilder query : queries) {
             if (filter == null) {
                 compoundQueryBuilder.add(query);
             } else {
                 compoundQueryBuilder.add(query.filter(filter));
             }
+
+            // Check if children have inner hits for stats
+            if (hasInnerHits == false) {
```
vibrantvarun, line 286: Where are we changing this parameter value?

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessor.java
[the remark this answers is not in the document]
q-andy: Sure, can refactor the others as maps.

src/main/java/org/opensearch/neuralsearch/processor/RRFProcessor.java
q-andy: I added this since @martin-gaievski requested, but my opinion is if currently RRF processor only works works with RRF combination technique and RRF normalization technique, then functionally the stats will be identical, and I don't think there's a point to including a breakdown, it will just be duplicated. Combination is configurable with a single option, but normalization technique isn't a configurable in processor config: https://docs.opensearch.org/docs/latest/search-plugins/search-pipelines/score-ranker-processor/ 

If in the future there are more normalizaiton/score techniques added we can add this granularity then.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -270,13 +273,26 @@ public static HybridQueryBuilder fromXContent(XContentParser parser) throws IOEx
...
+                Map<String, InnerHitContextBuilder> innerHits = new HashMap<>();
```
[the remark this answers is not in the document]
q-andy: The thought here is that we care about the stats at the level of the hybrid query. If at least one child query has inner hits then that means the hybrid query is an inner hits query. So we check each child, if at least one has inner hits, then we set hasInnerHits = true and we don't have to keep checking the rest.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
q-andy: It's set inside the loop, see my other comment for explanation. Basically my thought is we want to increment the stat once if at least one child query has inner hits, which means the hybrid query as a whole is an inner hits hybrid query.

src/main/java/org/opensearch/neuralsearch/processor/RRFProcessor.java
martin-gaievski: no in near future, my worry was more about consistent result format, because for RRF we have "rrf" as both normalization technique (hidden), and combination technique.

src/main/java/org/opensearch/neuralsearch/processor/RRFProcessor.java
martin-gaievski: Having RRF as processor and as technique gives more dimensions for reporting. For instance, with original version you can generate report with breakdown "by processor type". But report with breakdown on something like "by combination technique" will be harder because for RRF you raw data will have combination technique metric with empty value.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessor.java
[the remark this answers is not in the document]
martin-gaievski: this map can be static, same for the second map `combTechniqueIncrementers`

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
```diff
@@ -162,12 +199,66 @@ private void countTextChunkingProcessorStats(Map<InfoStatName, CountableInfoStat
...
+                for (Map.Entry<String, Object> entry : phaseResultsProcessor.entrySet()) {
+                    String processorType = entry.getKey();
+                    Map<String, Object> processorConfig = asMap(entry.getValue());
+                    switch (processorType) {
+                        case NormalizationProcessor.TYPE:
+                            countNormalizationProcessorStats(stats, processorConfig);
+                            break;
+                        case RRFProcessor.TYPE:
+                            countRRFProcessorStats(stats, processorConfig);
+                            break;
```
martin-gaievski, line 223: please add default case and throw one of runtime exceptions in case we reach it

src/main/java/org/opensearch/neuralsearch/processor/RRFProcessor.java
q-andy: I see your point, my concern is having duplicated information in the stats API makes it less readable and less performant. Especially as we add more features, we are concerned with possible response bloat, similar the issues core is facing with the size of `_nodes/info` and `_/nodes/stats` responses causing slowdowns on large clusters. And adding stats is one-way door in the sense that it's difficult to justify removing them since it's a breaking change. Ideally we save granularity for when it is needed and don't add more stats unless necessary 

From a report perspective, I'm thinking in terms of what kind of insight a breakdown would give you: If you are trying to determine which score combination techniques are seeing more usage, perhaps proportion of RRF to zscore or minmax isn't comparable since they're categorically different, e.g. used in different processors in different contexts. And if needed, the information is available implicitly from looking at RRF processor stats directly.

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
q-andy: Not sure if that makes sense here? Here we are iterating through all processors in the pipeline, which may contain processors from core or other plugins. If we encounter an `MLInferenceRequestProcessor` for example, we would not want to record stats for it, and  that would trigger the default case and throw an exception, breaking the API.

vibrantvarun:
LGTM

PR #1312, opened 2025-05-09
Feat: Add the `recursive_character` algorithm for Text chunking processor

YeonghyeonKO:
### Description
Add the recursive character chunker algorithm to TextChunking processors. It is designed for processing generic text, facilitating the division of larget text bodies into smaller, semantically meaningful chunks. With `delimiters` and `chunk_size` parameters, the algorithm attempts to split the text using these characters in sequential order until the resulting chunks satisfy the given size requirements.

The default list of delimiters used by the algorithm is:
- "\n\n": preserves paragraphs.
- "\n": preserves sentences when paragraphs are not feasible to keep intact.
- " ": keeps words together when sentences must be split.
- "": simply splits by character, ensuring the text is split even if no other delimiters are appropriate.

```
PUT _ingest/pipeline/text-chunking-ingest-pipeline
{
  "description": "A text chunking ingest pipeline using recursive character chunker",
  "processors": [
    {
      "text_chunking": {
        "algorithm": {
          "recursive_character": {
            "chunk_size": 50,
            "delimiters": ["\n\n", "\n", " ", ""]
          }
        },
        "field_map": {
          "passage_text": "passage_chunk"
        }
      }
    }
  ]
}
```

### Related Issues
Resolves #1261 ([PROPOSAL] RecursiveCharacterTextSplitter as a new algorithm of Text Chunking Processor)

### Check List
- [x] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/neural-search/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #1261, opened 2025-04-10
[PROPOSAL] RecursiveCharacterTextSplitter as a new algorithm of Text Chunking Processor


YeonghyeonKO:
### What are you proposing?
`RecursiveCharacterTextSplitter`
- ref: 
  - https://python.langchain.com/docs/how_to/recursive_text_splitter/
  - https://github.com/langchain-ai/langchain/blob/master/libs/text-splitters/langchain_text_splitters/character.py#L58

### What problems are you trying to solve?
I want to add a chunk processor to implement a recursive character text splitter, which is a popular chunking strategy. Text Chunking Processor in OpenSearch supports [fixed_token_length](https://opensearch.org/docs/latest/ingest-pipelines/processors/text-chunking/#fixed-token-length-algorithm) and [delimiter](https://opensearch.org/docs/latest/ingest-pipelines/processors/text-chunking/#delimiter-algorithm) for now, but the combination of these two cannot implement recursive character text splitter algorithm.

#### Are there any security considerations? 
No

#### Are there any breaking changes to the API
No

### What is the user experience going to be?
RecursiveCharacterTextSplitter will most likely help the developer who wants to split long texts in the format of  [Markdown](https://github.com/langchain-ai/langchain/blob/master/libs/text-splitters/langchain_text_splitters/character.py#L459), [Programming Language](https://github.com/langchain-ai/langchain/blob/master/libs/text-splitters/langchain_text_splitters/character.py#L196), or even [HTML](https://github.com/langchain-ai/langchain/blob/master/libs/text-splitters/langchain_text_splitters/character.py#L503).

#### Are there breaking changes to the User Experience?
No

YeonghyeonKO:
If the maintainers of [opensearch-project/neural-search](https://github.com/opensearch-project/neural-search) think this new algorithm valuable to be added, I will create a pull request.

heemin32:
@YeonghyeonKO Thanks for opening the issue. I believe this would be a nice feature to have. What are your thoughts, @yuye-aws and @zane-neo?

YeonghyeonKO:
Is there any idea or further progress about Text Chunking Processor including both this proposal and #1262 ?

q-andy:
[...]

PR #1330, opened 2025-05-17
Feat: Add `chunk_size` parameter to DelimiterChunker

YeonghyeonKO:
### Description
The existing implementation only splits input based on the given delimiter. However, in many cases, the segment between two delimiters can exceed a reasonable memory or display size (especially in UIs, summarization tasks, or embeddings). In such cases, enforcing a maximum `chunk_size` is necessary.

This PR enhances the `DelimiterChunker.chunk()` method to support:
- `chunk_size`-based splitting in addition to delimiter-based splitting.
- Smarter handling of content exceeding `chunk_size`.
- Better runtime chunk limit enforcement.
- Improved naming and inline documentation for clarity and maintainability.

The logic has been restructured to gracefully split large segments while preserving delimiters and avoiding OOM or excessive chunk generation when upper limits are reached.

<br>

### Related Issues
-  #1261
- https://github.com/opensearch-project/neural-search/pull/1312#issuecomment-2865631495

<br>

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/chunker/DelimiterChunker.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/chunker/DelimiterChunker.java
@@ -7,8 +7,10 @@ package org.opensearch.neuralsearch.processor.chunker;
 import java.util.Map;
 import java.util.List;
 import java.util.ArrayList;
+import java.util.Collections;
 
 import static org.opensearch.neuralsearch.processor.chunker.ChunkerParameterParser.parseInteger;
+import static org.opensearch.neuralsearch.processor.chunker.ChunkerParameterParser.parseIntegerWithDefault;
 import static org.opensearch.neuralsearch.processor.chunker.ChunkerParameterParser.parseStringWithDefault;
 
 /**
@@ -25,8 +27,17 @@ public final class DelimiterChunker implements Chunker {
     /** The default delimiter value used when none is specified. Uses two consecutive newline characters to split on paragraph boundaries. */
     public static final String DEFAULT_DELIMITER = "\n\n";
 
+    /** The parameter field name for specifying the chunk size. */
+    public static final String CHUNK_SIZE_FIELD = "chunk_size";
+
+    /** The default chunk size value. */
+    public static final int DEFAULT_CHUNK_SIZE = Integer.MAX_VALUE;
+
     /** The delimiter string used for text chunking. */
     private String delimiter;
+
+    /** The minimum chunk size. If the total text length is smaller than this, it will not be split. */
+    private int chunkSize;
 
     /**
      * Constructor that initializes the delimiter chunker with the specified parameters.
@@ -41,12 +52,14 @@ public final class DelimiterChunker implements Chunker {
      * Throw IllegalArgumentException if delimiter is not a string or an empty string.
      *
      * @param parameters a map with non-runtime parameters as the following:
-     * 1. delimiter A string as the paragraph split indicator
-     * 2. max_chunk_limit processor level max chunk limit
+     * 1. delimiter A string used as the split indicator (e.g., paragraph separator).
+     * 2. chunk_size An integer representing the maximum allowed length for each chunk;
+     *    if a segment split by delimiter exceeds this length, it will be further split.
      */
     @Override
     public void parseParameters(Map<String, Object> parameters) {
         this.delimiter = parseStringWithDefault(parameters, DELIMITER_FIELD, DEFAULT_DELIMITER);
+        this.chunkSize = parseIntegerWithDefault(parameters, CHUNK_SIZE_FIELD, DEFAULT_CHUNK_SIZE);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


YeonghyeonKO:
As discussed with @yuye-aws, I’ve submitted a pull request that implements the `chunk_size` parameter for the Delimiter Chunker.

Currently, only unit tests have been added. Once the code review is mostly complete, I’ll proceed with adding the integration test example that demonstrates usage of the chunk_size parameter.

src/main/java/org/opensearch/neuralsearch/processor/chunker/DelimiterChunker.java
```diff
@@ -41,12 +52,14 @@ public DelimiterChunker(final Map<String, Object> parameters) {
      * Throw IllegalArgumentException if delimiter is not a string or an empty string.
      *
      * @param parameters a map with non-runtime parameters as the following:
-     * 1. delimiter A string as the paragraph split indicator
-     * 2. max_chunk_limit processor level max chunk limit
```
YeonghyeonKO, line 45: The line
> 2. max_chunk_limit processor level max chunk limit

is already covered in the runtime parameter explanation below, so it can be removed.

src/main/java/org/opensearch/neuralsearch/processor/chunker/DelimiterChunker.java
```diff
@@ -62,23 +75,59 @@ public List<String> chunk(final String content, final Map<String, Object> runtim
...
-            nextDelimiterPosition = content.indexOf(delimiter, start);
-        }
 
-        // add the rest content into the chunk result
-        if (start < content.length()) {
-            chunkResult.add(content.substring(start));
+            int nextDelimiterPosition = content.indexOf(delimiter, start);
+            int end = (nextDelimiterPosition == -1) ? content.length() : nextDelimiterPosition + delimiter.length();
+
+            String segment = content.substring(start, end);
```
YeonghyeonKO, line 95: If the name of this variable is confused with the "segment" in Lucene indices, I'll rename it to something clearer like `fragment`, `partition`, or `slice`.

heemin32: `segment` might not be the best term here. Would `slice` or 'chunk' be clearer?

src/main/java/org/opensearch/neuralsearch/processor/chunker/DelimiterChunker.java
```diff
@@ -88,4 +137,24 @@ public List<String> chunk(final String content, final Map<String, Object> runtim
     public String getAlgorithmName() {
         return ALGORITHM_NAME;
     }
+
+    private List<String> splitByLength(final String text, final int maxLength) {
```
heemin32, line 141: Could you explain a little about this method?
if text is `abcde<tag>`, delimiter is `<tag>`, and maxLength is 3, this will return `abcde<tag>` which exceed maxLength.

src/main/java/org/opensearch/neuralsearch/processor/chunker/DelimiterChunker.java
```diff
@@ -88,4 +137,24 @@ public List<String> chunk(final String content, final Map<String, Object> runtim
...
+        List<String> result = new ArrayList<>();
+        int start = 0;
+
+        while (start < text.length()) {
+            int end = Math.min(start + maxLength, text.length());
+
+            // If a delimiter exists within the chunk and doesn't end at the boundary,
+            // adjust to include the full delimiter
+            if (!text.substring(start, end).endsWith(delimiter) && text.indexOf(delimiter, start) < end) {
```
heemin32, line 150: text: ab<tags>
delimiter: <tags>
maxLength: 3
Then, the result will be `ab<tags>` which is over maxLength?
If this behavior is correct, we might better change the method name to reflect what it actually does. It does not simply split by length.

src/main/java/org/opensearch/neuralsearch/processor/chunker/DelimiterChunker.java
```diff
@@ -25,9 +27,18 @@ public final class DelimiterChunker implements Chunker {
     /** The default delimiter value used when none is specified. Uses two consecutive newline characters to split on paragraph boundaries. */
     public static final String DEFAULT_DELIMITER = "\n\n";
 
+    /** The parameter field name for specifying the chunk size. */
+    public static final String CHUNK_SIZE_FIELD = "chunk_size";
+
+    /** The default chunk size value. */
+    public static final int DEFAULT_CHUNK_SIZE = 0;
```
heemin32, line 34: Could be better to set it as Integer.MAX_VALUE?

src/main/java/org/opensearch/neuralsearch/processor/chunker/DelimiterChunker.java
```diff
@@ -62,23 +75,59 @@ public List<String> chunk(final String content, final Map<String, Object> runtim
...
+
+                    // If adding this chunk exceeds the max chunk limit, merge all remaining parts into the last chunk
+                    if (Chunker.checkRunTimeMaxChunkLimit(chunkResult.size(), runtimeMaxChunkLimit, chunkStringCount)) {
+                        StringBuilder lastChunkBuilder = new StringBuilder();
+
+                        for (int j = i; j < splitChunks.size(); j++) {
+                            lastChunkBuilder.append(splitChunks.get(j));
+                        }
+
+                        int remainingContentStartIndex = start + segmentOffsetWithinContent;
```
heemin32, line 114: segmentOffsetWithinContent is updated when chunk is added to chunkResult in line 123. 
Shouldn't we simply do `chunkResult.add(content.substring(start + segmentOffsetWithinContent))` at line 110?

src/main/java/org/opensearch/neuralsearch/processor/chunker/DelimiterChunker.java
yuye-aws: Make sense

src/main/java/org/opensearch/neuralsearch/processor/chunker/DelimiterChunker.java
yuye-aws: I would prefer **chunk**

yuye-aws:
Hi @YeonghyeonKO ! Since we are considering cascading multiple chunking processor. Do you think it is a good idea to implement a new Chunker named FixedStringLengthChunker? It will check the max length of each chunk and perform chunking exactly as your PR. I am considering this idea because of the example:

["abcd.efgh"], max_chunk_limit = 2. Then the expected output should be ["abcd", "efgh"]. It does not matter on the "chunk_size".

yuye-aws:
@YeonghyeonKO You can also implement a few simple unit test cases for the corner case.

src/test/java/org/opensearch/neuralsearch/processor/chunker/DelimiterChunkerTests.java
```diff
@@ -109,4 +133,102 @@ public void testChunk_whenExceedMaxChunkLimit_withTwoStringsTobeChunked_thenLast
         List<String> expectedPassages = List.of("\n\na\n\n\n");
         assertEquals(expectedPassages, passages);
     }
+
+    public void testChunk_whenOnlyDelimitersPresent_withChunkSizeTooSmall_thenEachDelimiterIsOwnChunk() {
+        String content = "\n\n\n\n";
+        DelimiterChunker chunker = new DelimiterChunker(Map.of(DELIMITER_FIELD, "\n\n", CHUNK_SIZE_FIELD, 1));
```
yuye-aws, line 139: When instantiate a new DelimiterChunker, only non-runtime parameter, namely DELIMITER_FIELD gets parsed

src/test/java/org/opensearch/neuralsearch/processor/chunker/DelimiterChunkerTests.java
```diff
@@ -109,4 +133,102 @@ public void testChunk_whenExceedMaxChunkLimit_withTwoStringsTobeChunked_thenLast
...
+        List<String> chunkResult = chunker.chunk(content, runtimeParameters);
```
yuye-aws, line 140: You are still chunking with this parameter: private final Map<String, Object> runtimeParameters = Map.of(MAX_CHUNK_LIMIT_FIELD, 100, CHUNK_STRING_COUNT_FIELD, 1);

YeonghyeonKO:
@yuye-aws 
> ["abcd.efgh"], max_chunk_limit = 2. Then the expected output should be ["abcd", "efgh"]. It does not matter on the "chunk_size".
> From your current PR, if you set chunk_size to be 2. I am afraid the output would be ["ab", "cd.efgh"]. This would not be your expected output.

Actually, this result illustrates the logic I was referring to.

1) Starting with the input string `abcd.efgh`, the first 'chunk' (which I referred to as "segment" in this PR, but will now call "chunk") is `abcd.` . This chunk is determined by taking the substring from the beginning of the input up to and including the first delimiter (`.`). In this specific case, nextDelimiterPosition (the index of the delimiter `.`) is 4, and the end of this chunk is calculated as nextDelimiterPosition + 1 (=5).
2) Since the length of this first chunk (`abcd.`, which is 5 characters) is greater than chunkSize (which is 2), the splitByLength method then splits it further into `List.of("ab", "cd.")`.
3) At this point, because `Chunker.checkRunTimeMaxChunkLimit `evaluates to true, the remaining part of the original string (`efgh`) is merged into the last element of splitChunks (which is `cd.` .

src/main/java/org/opensearch/neuralsearch/processor/chunker/DelimiterChunker.java
YeonghyeonKO: I agree.

yuye-aws:
@YeonghyeonKO Since we have this PR now: https://github.com/opensearch-project/neural-search/pull/1342. Would you mind closing this PR?

PR #1332, opened 2025-05-19
add stats for text embedding processors with flags

will-hwang:
### Description
Enhance Stats for Text Embedding Processor, including stats for `skip_existing` option enabled

### Updated Response

```
{
	"_nodes": {
		"total": 1,
		"successful": 1,
		"failed": 0
	},
	"cluster_name": "integTest",
	"info": {
		"cluster_version": "3.1.0",
		"processors": {
			"ingest": {
				"text_chunking_delimiter_processors": 0,
				"text_chunking_fixed_length_processors": 0,
				"text_embedding_processors_in_pipelines": 1,
				"text_embedding_skip_existing_processors": 1,
				"text_chunking_processors": 0
			}
		}
	},
	"all_nodes": {
		"processors": {
			"ingest": {
				"text_chunking_executions": 0,
				"text_embedding_executions": 2,
				"text_embedding_skip_existing_executions": 2,
				"text_chunking_fixed_length_executions": 0,
				"text_chunking_delimiter_executions": 0
			}
		}
	},
	"nodes": {
		"rMyVPGp2SsWL-sLQ3HSjCQ": {
			"processors": {
				"ingest": {
					"text_chunking_executions": 0,
					"text_embedding_executions": 2,
					"text_embedding_skip_existing_executions": 2,
					"text_chunking_fixed_length_executions": 0,
					"text_chunking_delimiter_executions": 0
				}
			}
		}
	}
}
```

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
@@ -72,6 +72,7 @@ public final class TextEmbeddingProcessor extends InferenceProcessor {
             generateAndSetInference(ingestDocument, processMap, inferenceList, handler);
             return;
         }
+        EventStatsManager.increment(EventStatName.TEXT_EMBEDDING_PROCESSOR_SKIP_EXISTING_EXECUTIONS);
         // if skipExisting flag is turned on, eligible inference texts will be compared and filtered after embeddings are copied
         Object index = ingestDocument.getSourceAndMetadata().get(INDEX_FIELD);
         Object id = ingestDocument.getSourceAndMetadata().get(ID_FIELD);
@@ -109,6 +110,7 @@ public final class TextEmbeddingProcessor extends InferenceProcessor {
     @Override
     public void subBatchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers, Consumer<List<IngestDocumentWrapper>> handler) {
         try {
+            EventStatsManager.increment(EventStatName.TEXT_EMBEDDING_PROCESSOR_EXECUTIONS);
             if (CollectionUtils.isEmpty(ingestDocumentWrappers)) {
                 handler.accept(ingestDocumentWrappers);
                 return;
@@ -126,6 +128,7 @@ public final class TextEmbeddingProcessor extends InferenceProcessor {
             }
             // skipExisting flag is turned on, eligible inference texts in dataForInferences will be compared and filtered after embeddings
             // are copied
+            EventStatsManager.increment(EventStatName.TEXT_EMBEDDING_PROCESSOR_SKIP_EXISTING_EXECUTIONS);
             openSearchClient.execute(
                 MultiGetAction.INSTANCE,
                 buildMultiGetRequest(dataForInferences),
--- a/src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
+++ b/src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
@@ -17,6 +17,7 @@ import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.stream.Collectors;
 
 /**
@@ -130,7 +131,7 @@ public class InfoStatsManager {
                     Map<String, Object> processorConfig = asMap(entry.getValue());
                     switch (processorType) {
                         case TextEmbeddingProcessor.TYPE:
-                            increment(stats, InfoStatName.TEXT_EMBEDDING_PROCESSORS);
+                            countTextEmbeddingProcessorStats(stats, processorConfig);
                             break;
                         case TextChunkingProcessor.TYPE:
                             countTextChunkingProcessorStats(stats, processorConfig);
@@ -138,6 +139,19 @@ public class InfoStatsManager {
                     }
                 }
             }
+        }
+    }
+
+    /**
+     * Counts text embedding processor stats based on processor config
+     * @param stats map containing the stat to increment
+     * @param processorConfig map of the processor config, parsed to add stats
+     */
+    private void countTextEmbeddingProcessorStats(Map<InfoStatName, CountableInfoStatSnapshot> stats, Map<String, Object> processorConfig) {
+        increment(stats, InfoStatName.TEXT_EMBEDDING_PROCESSORS);
+        Object skipExisting = processorConfig.get(TextEmbeddingProcessor.SKIP_EXISTING);
+        if (Objects.nonNull(skipExisting) && skipExisting.equals(Boolean.TRUE)) {
+            increment(stats, InfoStatName.TEXT_EMBEDDING_SKIP_EXISTING_PROCESSORS);
         }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - [Stats] Add stats for text chunking processor algorithms ([#1308](https://github.com/opensearch-project/neural-search/pull/1308))
 - Support custom weights in RRF normalization processor ([#1322](https://github.com/opensearch-project/neural-search/pull/1322))
 - [Stats] Add stats tracking for semantic highlighting ([#1327](https://github.com/opensearch-project/neural-search/pull/1327))
-
+- [Stats] Add stats for text embedding processor with different settings ([#1332](https://github.com/opensearch-project/neural-search/pull/1332))
 ### Bug Fixes
 - Fix score value as null for single shard when sorting is not done on score field ([#1277](https://github.com/opensearch-project/neural-search/pull/1277))
 - Return bad request for stats API calls with invalid stat names instead of ignoring them ([#1291](https://github.com/opensearch-project/neural-search/pull/1291))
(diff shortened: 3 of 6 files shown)


heemin32:
Could you update the PR description with updated response of the stats api?

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
```diff
@@ -141,6 +142,19 @@ private void addIngestProcessorStats(Map<InfoStatName, CountableInfoStatSnapshot
...
+    /**
+     * Counts text embedding processor stats based on processor config
+     * @param stats map containing the stat to increment
+     * @param processorConfig map of the processor config, parsed to add stats
+     */
+    private void countTextEmbeddingProcessorStats(Map<InfoStatName, CountableInfoStatSnapshot> stats, Map<String, Object> processorConfig) {
+        increment(stats, InfoStatName.TEXT_EMBEDDING_PROCESSORS);
+        Object skipExisting = processorConfig.get(TextEmbeddingProcessor.SKIP_EXISTING);
+        if (Objects.nonNull(skipExisting) && skipExisting.equals(Boolean.TRUE)) {
+            increment(stats, InfoStatName.TEXT_EMBEDDING_SKIP_EXISTING_PROCESSORS);
```
q-andy, line 154: Can you use this caster helper here? Or does this not work because it's a boolean primative? https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java#L180-L193

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -72,6 +72,7 @@ public void doExecute(
             generateAndSetInference(ingestDocument, processMap, inferenceList, handler);
             return;
         }
+        EventStatsManager.increment(EventStatName.TEXT_EMBEDDING_PROCESSOR_SKIP_EXISTING_EXECUTIONS);
```
q-andy, line 75: Not this change but I think I noticed a bug in my initial PR, we should be incrementing the stat even when we run batch execute but currently we only increment during single execute so I think it might not be counted when we run pipelines in batch. Do you think you could include that change in this PR as well? If not I can try to catch it in the next one?

q-andy (commented):
lgtm, minor nits

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
will-hwang: it directly reads as boolean, not as map. I think it's cleaner this way

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
will-hwang: sure, i can add it to this pr

will-hwang:
> Currently we increment the stat whenever we have an execution that has the option enabled. Should we also have a stat to record when we have a "cache hit" and successfully skip an inference, or do you think that would be redundant?
> 
> cc: @heemin32

@q-andy It really comes down to what insight we want to gain from these stats. 

1. Do we want to know how many domains are using certain features? 
2. Do we also want to know how the features are performing with the features enabled?

I think we should start with 1 (which this PR addresses), and scope out how, and to what extent we should support 2.

heemin32:
For cache_hit stats, we can wait until there is an ask from users with the feature.

heemin32:
I'm not sure if the current stat name `text_embedding_skip_existing_processors` is appropriate, since it's not a standalone processor. Would it make more sense to track how many processors have the skip option enabled, regardless of type?

Also, do we really need a separate execution stat like `text_embedding_skip_existing_executions`? Is it providing distinct value?

junqiu-lei:
@will-hwang You might need rebase with main branch to pass the CI.

heemin32:
Regarding the stats APIs in the neural plugin — do we really need to track processor executions metrics? Wouldn’t the number of processors alone be sufficient to measure adoption?

My main concern is scalability. The stats APIs have limitations in that area, and once we add a metric, it's difficult to remove it later. That’s why I’d prefer to keep the metrics as minimal as possible.

will-hwang:
> Regarding the stats APIs in the neural plugin — do we really need to track processor executions metrics? Wouldn’t the number of processors alone be sufficient to measure adoption?
> 
> My main concern is scalability. The stats APIs have limitations in that area, and once we add a metric, it's difficult to remove it later. That’s why I’d prefer to keep the metrics as minimal as possible.

@heemin32 
if we are okay with simply tracking adoption, i think we can remove the eventStats metrics and only keep one for infoStats. The change will probably need to be made for other processors too then

martin-gaievski:
> Regarding the stats APIs in the neural plugin — do we really need to track processor executions metrics? Wouldn’t the number of processors alone be sufficient to measure adoption?
> 
> My main concern is scalability. The stats APIs have limitations in that area, and once we add a metric, it's difficult to remove it later. That’s why I’d prefer to keep the metrics as minimal as possible.

detailed metrics would be useful too, we can see which search configuration is most used one and invest there our efforts to improve relevance or other aspects of search. If number of metrics is critical for infra team then we can have only number of processors, that is P0.

heemin32:
> > Regarding the stats APIs in the neural plugin — do we really need to track processor executions metrics? Wouldn’t the number of processors alone be sufficient to measure adoption?
> > My main concern is scalability. The stats APIs have limitations in that area, and once we add a metric, it's difficult to remove it later. That’s why I’d prefer to keep the metrics as minimal as possible.
> 
> @heemin32 if we are okay with simply tracking adoption, i think we can remove the eventStats metrics and only keep one for infoStats. The change will probably need to be made for other processors too then. If we do make this change, for what cases should event execution metrics be emitted?

For event execution, we need to add metrics which cannot be retrieved from eventStats. For example, number of neural query execution of which data is not available from cluster info.

will-hwang:
> > > Regarding the stats APIs in the neural plugin — do we really need to track processor executions metrics? Wouldn’t the number of processors alone be sufficient to measure adoption?
> > > My main concern is scalability. The stats APIs have limitations in that area, and once we add a metric, it's difficult to remove it later. That’s why I’d prefer to keep the metrics as minimal as possible.
> > 
> > 
> > @heemin32 if we are okay with simply tracking adoption, i think we can remove the eventStats metrics and only keep one for infoStats. The change will probably need to be made for other processors too then. If we do make this change, for what cases should event execution metrics be emitted?
> 
> For event execution, we need to add metrics which cannot be retrieved from eventStats. For example, number of neural query execution of which data is not available from cluster info.

sounds reasonable to me. What do others think?
@q-andy @martin-gaievski

heemin32:
> > Regarding the stats APIs in the neural plugin — do we really need to track processor executions metrics? Wouldn’t the number of processors alone be sufficient to measure adoption?
> > My main concern is scalability. The stats APIs have limitations in that area, and once we add a metric, it's difficult to remove it later. That’s why I’d prefer to keep the metrics as minimal as possible.
> 
> detailed metrics would be useful too, we can see which search configuration is most used one and invest there our efforts to improve relevance or other aspects of search. If number of metrics is critical for infra team then we can have only number of processors, that is P0.

I agree that more data can be better, but we also have to consider that adding a metric isn't without cost. Also, there's a chance that a single user could generate a large number of calls to a specific processor, which might skew the data and not accurately reflect its true popularity.
Perhaps tracking the number of processors used would be a more reliable metric for measuring adoption and it alone might be sufficient?

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
```diff
@@ -109,6 +110,7 @@ public void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler
     @Override
     public void subBatchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers, Consumer<List<IngestDocumentWrapper>> handler) {
         try {
+            EventStatsManager.increment(EventStatName.TEXT_EMBEDDING_PROCESSOR_EXECUTIONS);
```
bzhangam, line 113: Should we count this by the num of the docs that we are processing? I feel we may want to know how many docs are processed by the processor.

Or we may want to use another event name for the batch processing use case otherwise it can be confusing if we want to rely on this event to tell how many docs we have processed.

q-andy:
> My main concern is scalability. The stats APIs have limitations in that area, and once we add a metric, it's difficult to remove it later. That’s why I’d prefer to keep the metrics as minimal as possible.

Had a chat with infra team, the primary concern is high memory consumption on large clusters seen when calling APIs like `_node/stats` and `_node/state` due to large response payloads. For production clusters they mitigate this by filtering by calling multiple times and filtering specific stats. I opened #1360 to give an option to mitigate the size of the payload, and opened an issue to add more filtering options in #1363.

Based on this, for 3.1 it should be okay to add granular stats, and caller side can filter them as needed if we run into scalability concerns.

src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
will-hwang: i think we're more concerned with the number of execution than how many docs being processed per execution in general for other processors as well

heemin32 (approved):

junqiu-lei (approved):

PR #1337, opened 2025-05-21
Enhance semantic field to allow to enable/disable chunking.

bzhangam:
### Description
Enhance semantic field to allow to enable/disable chunking.

By default, if the chunking is not defined we will use false and disable the chunking. We want to do this way because enabling the chunking will user nested object even for short text that doesn't need to be chunked. And the nested object and make the search slower.

Also refactor the code a little bit to pull the space type from the additional_config of the ML model config.

### Related Issues
N/A

### Check List


src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
```diff
@@ -314,7 +293,7 @@ private List<String> chunkLeafType(final Object value, final Map<String, Object>
             if (StringUtils.isBlank(String.valueOf(value))) {
                 return result;
             }
-            result = chunkString(value.toString(), runTimeParameters);
+            result = chunker.chunkString(value.toString(), runTimeParameters);
```
heemin32, line 296: Shouldn't we check if chunker is not null?
```
Objects.requireNonNull(chunker, "Chunker should not be null when chunk is requested");
```

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -272,27 +275,56 @@ private List<SemanticFieldInfo> getSemanticFieldInfo(IngestDocument ingestDocume
...
+    }
+
+    private void executeChunkers(@NonNull final SemanticFieldInfo semanticFieldInfo, int maxTokenCount) {
+        for (Chunker chunker : semanticFieldInfo.getChunkers()) {
+            final Map<String, Object> runtimeParameters = new HashMap<>();
+            final List<String> chunks = semanticFieldInfo.getChunks();
+            final boolean isFirstChunker = chunks == null;
+            runtimeParameters.put(FixedTokenLengthChunker.MAX_TOKEN_COUNT_FIELD, maxTokenCount);
+            runtimeParameters.put(CHUNK_STRING_COUNT_FIELD, isFirstChunker ? 1 : chunks.size());
+            // TODO: Should allow user to configure it for each chunker
```
heemin32, line 315: For all these TODO, might be better to create a github issue as a follow up item rather than recording it inside code?

src/test/resources/processor/semantic/ingested_doc1.json
```diff
@@ -1,12 +1,9 @@
 {
```
will-hwang, line 1: shouldn't we keep the ingested doc files for both chunking_enabled and disabled?

src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
bzhangam: I think it cannot be null. This validation is already done when we initiate the TextChunkingProcessor.

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
```diff
@@ -42,6 +46,29 @@ public interface Chunker {
      */
     List<String> chunk(String content, Map<String, Object> runtimeParameters);
 
+    /**
+     * Chunk a string and also update the runTimeParameters properly. At the end return the chunked results.
+     * @param content The string content to chunk
+     * @param runTimeParameters a map containing runtime parameters for chunking algorithms
+     * @return chunked passages
+     */
+    default List<String> chunkString(final String content, final Map<String, Object> runTimeParameters) {
```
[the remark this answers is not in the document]
bzhangam: True. It's better to make it an abstract class.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
bzhangam: Sure. I can create one.

src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
heemin32: It might be better for this method to receive chunker as a parameter to explicitly say that chunker should not be null to call this method?

bzhangam: I think we can do that but not necessary since this is just a private function in the TextChunkingProcessor.

src/test/resources/processor/semantic/ingested_doc1.json
bzhangam: I think this is what we are doing. We have two semantic fields in the doc. One with chunking disabled and another with chunking enabled.

will-hwang: looks good, the diff only showed the one

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
```diff
@@ -367,16 +377,43 @@ private static Map<String, Object> getSemanticEmbeddingFieldConfig(
...
+            embeddingFieldPath = semanticInfoFieldFullPath + PATH_SEPARATOR + EMBEDDING_FIELD_NAME;
         }
-        targetFieldConfigBuilder.chunksPath(embeddingFieldPath + PATH_SEPARATOR + CHUNKS_FIELD_NAME);
-        embeddingFieldPath += PATH_SEPARATOR + CHUNKS_FIELD_NAME + PATH_SEPARATOR + EMBEDDING_FIELD;
         targetFieldConfigBuilder.embeddingFieldPath(embeddingFieldPath);
         return SemanticMappingUtils.getFieldConfigByPath(mappings, embeddingFieldPath);
     }
 
+    /**
+     * Check if the chunking is enabled in the semantic field config. If the field is not defined then return false as
```
junqiu-lei, line 400: Please also mention the new default behavior on PR description as well as incoming documentation.

src/main/java/org/opensearch/neuralsearch/processor/dto/SemanticFieldInfo.java
```diff
@@ -27,44 +31,79 @@ public class SemanticFieldInfo {
...
-     * @return full path to the chunks field of the semantic field
+     * @return full path to the chunks field of the semantic field in a doc
      */
-    public String getFullPathForChunks() {
-        return new StringBuilder().append(semanticInfoFullPath).append(PATH_SEPARATOR).append(CHUNKS_FIELD_NAME).toString();
+    public String getFullPathForChunksInDoc() {
+        if (Boolean.TRUE.equals(chunkingEnabled)) {
+            return new StringBuilder().append(semanticInfoFullPathInDoc).append(PATH_SEPARATOR).append(CHUNKS_FIELD_NAME).toString();
+        }
+        throw new RuntimeException(
```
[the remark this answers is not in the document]
bzhangam: Sure we can do that.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -86,14 +86,14 @@ public SemanticFieldProcessor(
         @NonNull final MLCommonsClientAccessor mlClientAccessor,
         @NonNull final Environment environment,
         @NonNull final ClusterService clusterService,
-        @NonNull final Chunker chunker
+        @NonNull final Chunker defaultChunker
```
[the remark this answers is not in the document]
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
yuye-aws: > I think it cannot be null. This validation is already done when we initiate the TextChunkingProcessor.

Yes. Because there is a check in the chunker factory: https://github.com/opensearch-project/neural-search/blob/9f152879e3ceed23e15b9963763a7c6ad6adb45b/src/main/java/org/opensearch/neuralsearch/processor/chunker/ChunkerFactory.java#L41

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
```diff
@@ -40,7 +44,30 @@ public interface Chunker {
...
+        // update chunk_string_count for each string
+        int chunkStringCount = parseInteger(runTimeParameters, CHUNK_STRING_COUNT_FIELD);
+        runTimeParameters.put(CHUNK_STRING_COUNT_FIELD, chunkStringCount - 1);
+        // update runtime max_chunk_limit if not disabled
+        int runtimeMaxChunkLimit = parseInteger(runTimeParameters, MAX_CHUNK_LIMIT_FIELD);
+        if (runtimeMaxChunkLimit != DISABLED_MAX_CHUNK_LIMIT) {
+            runTimeParameters.put(MAX_CHUNK_LIMIT_FIELD, runtimeMaxChunkLimit - contentResult.size());
+        }
+        return contentResult;
+    }
```
yuye-aws, line 70: To be honest, I don't like this method being refactored into the Chunker class. When I am designing the Chunker interface, its sole purpose is to only perform chunking on the documents. 

Text chunking processor has the chunkString method because we are considering list and map objects. The chunkString method should be put together with methods like chunkLeafType and chunkList

yuye-aws: If this is the only reason to convert chunker from an interface to an abstract class. I prefer keeping chunker as an interface.

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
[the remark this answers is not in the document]
yuye-aws: Agree with abstract class. Both DelimiterChunker and FixedTokenLengthChunker have an is-a releationship with **chunker**.

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
bzhangam: The main purpose I move it to chunker is to make it easier for us to share the logic of updating the runTimeParameters. Since we already define the run paramater names in the Chunker class so I think it's ok for it also manage the runTimeParameters.

heemin32 (approved):
LGTM. Please fix the build failure before merge.

junqiu-lei (approved):
LGTM, please also publish related benchmark results on github while we're ready. Thanks.

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
yuye-aws: > make it easier for us to share the logic of updating the runTimeParameters

yuye-aws: For me, the four methods **chunkString** **chunkLeafType** **chunkList** and **chunkMapType** should at least share the same class. Different from the chunk method in the Chunker, these four methods are actually managing map type and list type, so I prefer putting them together in the chunking processor class.

yuye-aws: By single responsibility principle, the chunker abstract should only be aware of the chunking algorithm. These 4 methods handling nested or listed type is not related to the chunker class. I'll provide more context for you to better understand. You can take a look into the class **InferenceProcessor**, where it is handling nested type and listed type through functions like **buildMapWithTargetKeys**. So, I followed this class and keep these 4 methods in the processor class.

I think it would be better if we can move the chunkString class back to the processor.

bzhangam: Thanks for the detailed explanation, @yuye-aws.

I think what we can do is to create a util/helper class to own the logic which can be shared between the TextChunkingProcessor and the SemanticFieldProcessor. In this way we can keep the responsibility of the Chunker class unchanged but also make it easy to reuse the common logic. Does it sound good to you?

yuye-aws: Agreed. Can you also include the logic from the **InferenceProcessor**?

PR #1341, opened 2025-05-21
Add Search Analyzer Support at Query Time

chloewqg:
### Description
Add Search Analyzer Support at Query Time for semantic field

### Related Issues
Resolves #1341

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
@@ -50,6 +50,7 @@ import java.util.stream.Collectors;
 
 import lombok.NonNull;
 import org.apache.commons.lang.StringUtils;
+import org.opensearch.core.common.Strings;
 import org.apache.commons.lang.builder.EqualsBuilder;
 import org.apache.lucene.search.MatchNoDocsQuery;
 import org.apache.lucene.search.Query;
@@ -122,6 +123,7 @@ public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder>
     public static final ParseField QUERY_TEXT_FIELD = new ParseField("query_text");
 
     public static final ParseField MODEL_ID_FIELD = new ParseField("model_id");
+    public static final ParseField SEMANTIC_FIELD_SEARCH_ANALYZER_FIELD = new ParseField("semantic_field_search_analyzer");
 
     // fields only used for dense model
     public static final ParseField QUERY_IMAGE_FIELD = new ParseField("query_image");
@@ -171,6 +173,7 @@ public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder>
     private Supplier<Map<String, Float>> queryTokensMapSupplier;
     // fields to support the semantic field for sparse model
     private Map<String, Supplier<Map<String, Float>>> modelIdToQueryTokensSupplierMap;
+    private String searchAnalyzer;
 
     /**
      * A custom builder class to enforce valid Neural Query Builder instantiation
@@ -180,6 +183,7 @@ public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder>
         private String queryText;
         private String queryImage;
         private String modelId;
+        private String searchAnalyzer;
         private Integer k = null;
         private Float maxDistance = null;
         private Float minScore = null;
@@ -299,6 +303,11 @@ public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder>
             return this;
         }
 
+        public Builder searchAnalyzer(String searchAnalyzer) {
+            this.searchAnalyzer = searchAnalyzer;
+            return this;
+        }
+
         public NeuralQueryBuilder build() {
             requireValue(fieldName, "Field name must be provided for neural query");
 
@@ -318,7 +327,9 @@ public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder>
                 rescoreContext,
                 modelIdToVectorSupplierMap,
                 queryTokensMapSupplier,
-                modelIdToQueryTokensSupplierMap
+                modelIdToQueryTokensSupplierMap,
+                searchAnalyzer
+
             ).boost(boost).queryName(queryName);
 
             List<String> errors = new ArrayList<>();
@@ -409,6 +420,7 @@ public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder>
             this.queryTokensMapSupplier = queryTokensMapSupplierStreamInput(in);
             this.modelIdToVectorSupplierMap = modelIdToVectorSupplierMapStreamInput(in);
             this.modelIdToQueryTokensSupplierMap = modelIdToQueryTokensSupplierMapStreamInput(in);
+            this.searchAnalyzer = in.readOptionalString();
         }
     }
 
@@ -455,6 +467,7 @@ public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder>
             queryTokensMapSupplierStreamOutput(out, queryTokensMapSupplier);
             modelIdToVectorSupplierMapStreamOutput(out, modelIdToVectorSupplierMap);
             modelIdToQueryTokensSupplierMapStreamOutput(out, modelIdToQueryTokensSupplierMap);
+            out.writeOptionalString(this.searchAnalyzer);
         }
     }
 
@@ -583,6 +596,8 @@ public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder>
                     builder.minScore(parser.floatValue());
                 } else if (EXPAND_NESTED_FIELD.match(currentFieldName, parser.getDeprecationHandler())) {
                     builder.expandNested(parser.booleanValue());
+                } else if (SEMANTIC_FIELD_SEARCH_ANALYZER_FIELD.match(currentFieldName, parser.getDeprecationHandler())) {
+                    builder.searchAnalyzer(parser.text());
                 } else {
                     throw getUnsupportedFieldException(parser.getTokenLocation(), currentFieldName);
                 }
@@ -661,7 +676,10 @@ public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder>
         } else {
             // If it is not null it means we already start the async actions in the previous rewrite.
             // Current rewrite happens after all the async actions done so simply return this to end the rewrite.
-            if (modelIdToVectorSupplierMap != null || modelIdToQueryTokensSupplierMap != null || queryTokensMapSupplier != null) {
+            if (modelIdToVectorSupplierMap != null
+                || modelIdToQueryTokensSupplierMap != null
+                || queryTokensMapSupplier != null
+                || searchAnalyzer != null) {
                 // If we only have one target index it means we know the path to the target embedding feild so we can
                 // continue the rewrite
                 if (indexToTargetFieldConfigMap.size() == 1) {
@@ -705,8 +723,9 @@ public class NeuralQueryBuilder extends AbstractQueryBuilder<NeuralQueryBuilder>
             }
         } else if (RankFeaturesFieldMapper.CONTENT_TYPE.equals(embeddingFieldType)) {
             Supplier<Map<String, Float>> queryTokensSupplier = queryTokensMapSupplier;
-            // If the raw token is not provided then try to find the token generated by the ml model
-            if (queryTokensSupplier == null) {
+            // If the raw token is not provided or no search analyzer provided
+            // then try to find the token generated by the ml model
+            if (queryTokensSupplier == null && searchAnalyzer == null) {
                 if (modelIdToQueryTokensSupplierMap == null || modelIdToQueryTokensSupplierMap.get(searchModelId) == null) {
                     throw new RuntimeException(
                         getErrorMessageWithBaseErrorForSemantic(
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -299,13 +302,19 @@ public Builder buildStage(NeuralQueryBuildStage buildStage) {
...
+        }
+
         public NeuralQueryBuilder build() {
             requireValue(fieldName, "Field name must be provided for neural query");
 
             final NeuralQueryBuilder neuralQueryBuilder = new NeuralQueryBuilder(
                 fieldName,
                 queryText,
                 modelId,
+                searchAnalyzer,
```
bzhangam, line 317: And a validation that user should not define the model id and the analyzer at the same time.

src/main/java/org/opensearch/neuralsearch/util/NeuralQueryValidationUtil.java
```diff
@@ -142,8 +144,26 @@ public static List<String> validateNeuralQueryForSemanticSparse(@NonNull final N
             errors.add(String.format(Locale.ROOT, "%s field can not be empty", QUERY_TEXT_FIELD.getPreferredName()));
         }
 
-        if (StringUtils.EMPTY.equals(queryBuilder.modelId())) {
-            errors.add(String.format(Locale.ROOT, "%s field can not be empty", MODEL_ID_FIELD.getPreferredName()));
+        if (StringUtils.EMPTY.equals(queryBuilder.modelId()) && Strings.isNullOrEmpty(queryBuilder.searchAnalyzer())) {
```
bzhangam, line 147: For semantic sparse the model id and analyzer can be both null since we can pull the model id from the index mapping. The validation here is more to check if the model id is defined then it should not be an empty string. It's ok to be null.

For analyzer we should validation:
1. If the analyzer is defined then it should only be the supported analyzers.
2. If the analyzer is defined the the model id should not be defined at the same time.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -583,6 +592,8 @@ private static void parseQueryParams(XContentParser parser, Builder builder) thr
                     builder.minScore(parser.floatValue());
                 } else if (EXPAND_NESTED_FIELD.match(currentFieldName, parser.getDeprecationHandler())) {
                     builder.expandNested(parser.booleanValue());
+                } else if (SEARCH_ANALYZER_FIELD.match(currentFieldName, parser.getDeprecationHandler())) {
+                    builder.searchAnalyzer(parser.text());
```
bzhangam, line 596: The in and out stream function should also take care of it. And hashcode, equal functions.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -717,8 +728,11 @@ private QueryBuilder rewriteQueryForSemanticField(@NonNull final NeuralQueryTarg
                 queryTokensSupplier = modelIdToQueryTokensSupplierMap.get(searchModelId);
             }
 
-            final NeuralSparseQueryBuilder neuralSparseQueryBuilder = new NeuralSparseQueryBuilder().fieldName(embeddingFieldPath)
+            NeuralSparseQueryBuilder neuralSparseQueryBuilder = new NeuralSparseQueryBuilder().fieldName(embeddingFieldPath)
                 .queryTokensSupplier(queryTokensSupplier);
+            if (StringUtils.EMPTY.equals(this.searchAnalyzer) == false) {
```
bzhangam, line 733: We also need to modify rewriteQueryWithCoordinatorContext function to check if the analyzer exist then we don't need to try to inference the query text and we should rewriteQueryForSemanticField directly.

And here we should not try to pull the queryTokensSupplier from `modelIdToQueryTokensSupplierMap` since we will use analyzer to tokenize the query text.

src/main/java/org/opensearch/neuralsearch/util/NeuralQueryValidationUtil.java
```diff
@@ -142,10 +163,40 @@ public static List<String> validateNeuralQueryForSemanticSparse(@NonNull final N
             errors.add(String.format(Locale.ROOT, "%s field can not be empty", QUERY_TEXT_FIELD.getPreferredName()));
         }
 
+        final boolean hasModelIdDefined = queryBuilder.modelId() != null;
+        final boolean hasSemanticFieldSearchAnalyzerDefined = queryBuilder.searchAnalyzer() != null;
+
+        if (queryBuilder.queryTokensMapSupplier() != null && (hasModelIdDefined || hasSemanticFieldSearchAnalyzerDefined)) {
```
bzhangam, line 169: Can we do something similar to countKnnQueryTypes that only one of them should be provided?

src/main/java/org/opensearch/neuralsearch/util/NeuralQueryValidationUtil.java
```diff
@@ -142,10 +163,40 @@ public static List<String> validateNeuralQueryForSemanticSparse(@NonNull final N
...
+                )
+            );
+        }
+
         if (StringUtils.EMPTY.equals(queryBuilder.modelId())) {
             errors.add(String.format(Locale.ROOT, "%s field can not be empty", MODEL_ID_FIELD.getPreferredName()));
         }
 
+        if (StringUtils.EMPTY.equals(queryBuilder.searchAnalyzer())) {
+            errors.add(String.format(Locale.ROOT, "%s field can not be empty", SEMANTIC_FIELD_SEARCH_ANALYZER_FIELD.getPreferredName()));
```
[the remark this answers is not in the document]
chloewqg: why would we want to restrict to certain analyzers here? I remember standard could also be an option here.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -706,7 +723,7 @@ private QueryBuilder rewriteQueryForSemanticField(@NonNull final NeuralQueryTarg
         } else if (RankFeaturesFieldMapper.CONTENT_TYPE.equals(embeddingFieldType)) {
             Supplier<Map<String, Float>> queryTokensSupplier = queryTokensMapSupplier;
             // If the raw token is not provided then try to find the token generated by the ml model
-            if (queryTokensSupplier == null) {
+            if (queryTokensSupplier == null && Strings.isNullOrEmpty(this.searchAnalyzer)) {
```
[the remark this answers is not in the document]
chloewqg: Yeah the current validation should catch it already.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -149,6 +151,7 @@ public static void initialize(MLCommonsClientAccessor mlClient) {
     private String fieldName;
     private String queryText;
     private String modelId;
+    private String searchAnalyzer;
```
bzhangam, line 154: Can you also move this field to the below section with the comment ` // fields only used for sparse model`

chloewqg: Okay

src/main/java/org/opensearch/neuralsearch/util/NeuralQueryValidationUtil.java
```diff
@@ -44,6 +45,16 @@ public static List<String> validateNeuralQueryForKnn(
                 errors.add(String.format(Locale.ROOT, "%s must be provided.", MODEL_ID_FIELD.getPreferredName()));
             }
         }
+
+        if (queryBuilder.searchAnalyzer() != null) {
```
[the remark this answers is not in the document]
chloewqg: Updated it in latest revision.

src/main/java/org/opensearch/neuralsearch/util/NeuralQueryValidationUtil.java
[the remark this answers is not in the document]
bzhangam: @zhichao-aws Could you provide some insight on this should we validate we only support certain analyzers? My concern is that if users are using unexpected analyzers the may run into some unexpected behavior. If we silently drop the search quality with unexpected analyzers which is not good.

src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
```diff
@@ -138,6 +139,13 @@ public static class Builder extends ParametrizedFieldMapper.Builder {
             false
         );
 
+        private final Parameter<String> semanticFieldSearchAnalyzer = Parameter.stringParam(
+            SEARCH_ANALYZER,
+            false,
+            m -> ((SemanticFieldMapper) m).semanticParameters.getSemanticFieldSearchAnalyzer(),
+            ""
```
bzhangam, line 146: I don't think we can support an empty string analyzer. The default should be null.

bzhangam: I think we also should validate the search model id and the search analyzer should not be defined at the same time otherwise we don't know which one we should use.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -706,7 +724,7 @@ private QueryBuilder rewriteQueryForSemanticField(@NonNull final NeuralQueryTarg
         } else if (RankFeaturesFieldMapper.CONTENT_TYPE.equals(embeddingFieldType)) {
             Supplier<Map<String, Float>> queryTokensSupplier = queryTokensMapSupplier;
             // If the raw token is not provided then try to find the token generated by the ml model
```
bzhangam, line 726: Refactor the comment to reflect the updated logic.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -717,8 +735,12 @@ private QueryBuilder rewriteQueryForSemanticField(@NonNull final NeuralQueryTarg
                 queryTokensSupplier = modelIdToQueryTokensSupplierMap.get(searchModelId);
             }
 
-            final NeuralSparseQueryBuilder neuralSparseQueryBuilder = new NeuralSparseQueryBuilder().fieldName(embeddingFieldPath)
-                .queryTokensSupplier(queryTokensSupplier);
+            NeuralSparseQueryBuilder neuralSparseQueryBuilder = new NeuralSparseQueryBuilder().fieldName(embeddingFieldPath);
+            if (Strings.isNullOrEmpty(this.searchAnalyzer) == false) {
```
bzhangam, line 739: If this.searchAnalyzer is null which means searchAnalyzer is not passed through the query or neural enrich processor. Then we should try to find it from the semantic field config. Check how to get the model id: `getSearchModelId(targetFieldConfig)`

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -661,7 +676,10 @@ private QueryBuilder rewriteQueryWithCoordinatorContext(@NonNull final QueryCoor
         } else {
             // If it is not null it means we already start the async actions in the previous rewrite.
             // Current rewrite happens after all the async actions done so simply return this to end the rewrite.
-            if (modelIdToVectorSupplierMap != null || modelIdToQueryTokensSupplierMap != null || queryTokensMapSupplier != null) {
+            if (modelIdToVectorSupplierMap != null
+                || modelIdToQueryTokensSupplierMap != null
+                || queryTokensMapSupplier != null
+                || searchAnalyzer != null) {
```
bzhangam, line 682: What if the searchAnalyzer is defined in the semantic field?

chloewqg: This will be a separate PR. Let's merge this PR first.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
chloewqg: Yeah. I will do it in a separate PR

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
chloewqg: okay

src/main/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapper.java
chloewqg: Will do it in a separate PR

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
chloewqg: Updated

src/main/java/org/opensearch/neuralsearch/mapper/dto/SemanticParameters.java
```diff
@@ -18,4 +18,5 @@ public class SemanticParameters {
     private final String rawFieldType;
     private final String semanticInfoFieldName;
     private final Boolean chunkingEnabled;
+    private final String semanticFieldSearchAnalyzer;
```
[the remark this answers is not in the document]
chloewqg: Removed

src/main/java/org/opensearch/neuralsearch/util/NeuralQueryValidationUtil.java
```diff
@@ -142,10 +173,31 @@ public static List<String> validateNeuralQueryForSemanticSparse(@NonNull final N
             errors.add(String.format(Locale.ROOT, "%s field can not be empty", QUERY_TEXT_FIELD.getPreferredName()));
         }
 
+        int countValidInputsForSemanticSparseQuery = 0;
+        if (queryBuilder.modelId() != null) countValidInputsForSemanticSparseQuery++;
+        if (queryBuilder.searchAnalyzer() != null) countValidInputsForSemanticSparseQuery++;
+        if (queryBuilder.queryTokensMapSupplier() != null) countValidInputsForSemanticSparseQuery++;
```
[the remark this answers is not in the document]
chloewqg: Let's address it in next PR.

src/main/java/org/opensearch/neuralsearch/util/NeuralQueryValidationUtil.java
```diff
@@ -142,10 +173,31 @@ public static List<String> validateNeuralQueryForSemanticSparse(@NonNull final N
...
+
+        if (countValidInputsForSemanticSparseQuery > 1) {
```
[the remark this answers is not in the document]
chloewqg: I see. So only check model_id and search analyzer should not coexist instead of checking model id, search_analyzer and query tokens all together?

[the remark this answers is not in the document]
chloewqg: Let's address it in next PR then

src/main/java/org/opensearch/neuralsearch/util/NeuralQueryValidationUtil.java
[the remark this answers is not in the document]
bzhangam: I think we can skip this validation for now and add it if later we find it's needed.

heemin32 (approved):

junqiu-lei (approved):

PR #1345, opened 2025-05-22
[Feature/Collapse] Add collapse functionality to hybrid query

ryanbogan:
### Description
Adds collapse functionality to hybrid query

### Related Issues

### Check List


src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
```diff
@@ -99,14 +103,26 @@ public CompoundTopDocs(final QuerySearchResult querySearchResult) {
         // skipping first two elements, it's a start-stop element and delimiter for first series
         List<TopDocs> topDocsList = new ArrayList<>();
         List<ScoreDoc> scoreDocList = new ArrayList<>();
+        List<Object> collapseValueList = new ArrayList<>();
+        int collapseIndex = 2;
```
martin-gaievski: make 2 a constant

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -226,6 +252,201 @@ private void updateOriginalQueryResults(final CombineScoresDto combineScoresDTO,
...
+        Map<BytesRef, FieldDoc> collapseValueToTopScoreDocMap = new HashMap<>();
+        Map<BytesRef, Integer> collapseValueToShardMap = new HashMap<>();
+        String collapseField = "";
+
+        HybridQueryFieldDocComparator comparator = new HybridQueryFieldDocComparator(
+            ((CollapseTopFieldDocs) queryTopDocs.get(indexOfFirstNonEmpty).getTopDocs().getFirst()).fields,
+            Comparator.comparing((ScoreDoc scoreDoc) -> scoreDoc.score)
+        );
+
+        for (int i = 0; i < querySearchResults.size(); i++) {
```
martin-gaievski: can you rename i to signify what are we iterating on. I assume this is shardIndex

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -226,6 +252,201 @@ private void updateOriginalQueryResults(final CombineScoresDto combineScoresDTO,
...
+            CompoundTopDocs updatedTopDocs = queryTopDocs.get(i);
+            List<ScoreDoc> updatedScoreDocs = updatedTopDocs.getScoreDocs();
+            if (updatedScoreDocs.isEmpty()) {
+                continue;
+            }
+            collapseField = ((CollapseTopFieldDocs) updatedTopDocs.getTopDocs().getFirst()).field;
```
martin-gaievski: why getting first doc is enough in this scenario? can you add comment explaining this

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -226,6 +252,201 @@ private void updateOriginalQueryResults(final CombineScoresDto combineScoresDTO,
...
+
+            for (ScoreDoc scoreDoc : updatedScoreDocs) {
+                FieldDoc fieldDoc = (FieldDoc) scoreDoc;
+                BytesRef collapseValue = (BytesRef) fieldDoc.fields[fieldDoc.fields.length - 1];
```
martin-gaievski: add validation for fields array that it's not null and has non-zero length. In case any of those is true we probably need to log and continue to next iteration in the loop

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -226,6 +252,201 @@ private void updateOriginalQueryResults(final CombineScoresDto combineScoresDTO,
...
+                if (currentBestFieldDoc == null || comparator.compare(fieldDoc, currentBestFieldDoc) < 0) {
+                    collapseValueToTopScoreDocMap.put(BytesRef.deepCopyOf(collapseValue), fieldDoc);
+                    collapseValueToShardMap.put(BytesRef.deepCopyOf(collapseValue), i);
+                }
+            }
+        }
+
+        List<Map.Entry<BytesRef, FieldDoc>> entryList = new ArrayList<>(collapseValueToTopScoreDocMap.entrySet());
+
+        entryList.sort(Map.Entry.comparingByValue(comparator));
```
martin-gaievski: add comment on why we need this sorting

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -226,6 +252,201 @@ private void updateOriginalQueryResults(final CombineScoresDto combineScoresDTO,
...
+
+        for (int index = 0; index < querySearchResults.size(); index++) {
```
martin-gaievski: please pick more descriptive name for loop variable, maybe shardIndex?

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -226,6 +252,201 @@ private void updateOriginalQueryResults(final CombineScoresDto combineScoresDTO,
...
+                    collapseValues.add(collapseValue);
+                }
+            }
+
+            totalScoreDocsCount += newFieldDocs.size();
+
+            Object[] objectCollapseValues = new Object[collapseValues.size()];
+            for (int i = 0; i < objectCollapseValues.length; i++) {
+                objectCollapseValues[i] = collapseValues.get(i);
+            }
```
martin-gaievski: can we use simpler syntax:
```
Object[] objectCollapseValues = collapseValues.toArray();
```

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
martin-gaievski: can we do a pre-filtering for performance

The current approach checks each entry against each shard index, which is O(n×m) complexity. Consider reorganizing the data structure to avoid this nested iteration, potentially using:

Map<Integer, List<Map.Entry<BytesRef, FieldDoc>>> shardToEntriesMap = entryList.stream()
    .collect(Collectors.groupingBy(entry -> collapseValueToShardMap.get(entry.getKey())));

// then for each index:
List<Map.Entry<BytesRef, FieldDoc>> relevantEntries = shardToEntriesMap.getOrDefault(index, Collections.emptyList());

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -166,18 +167,37 @@ private Map<Integer, Object[]> getDocIdSortFieldsMap(
         Map<Integer, Object[]> docIdSortFieldMap = new HashMap<>();
         final List<TopDocs> topFieldDocs = compoundTopDocs.getTopDocs();
         final boolean isSortByScore = isSortOrderByScore(sort);
+        final boolean isCollapseEnabled = topFieldDocs.getFirst() instanceof CollapseTopFieldDocs;
         for (TopDocs topDocs : topFieldDocs) {
+            int i = 0;
```
martin-gaievski: pick more descriptive name

src/main/java/org/opensearch/neuralsearch/search/collector/CollapseDocSourceGroupSelector.java
```diff
@@ -0,0 +1,258 @@
...
+import org.apache.lucene.util.BytesRef;
+import org.opensearch.index.fielddata.AbstractNumericDocValues;
+import org.opensearch.index.fielddata.AbstractSortedDocValues;
+import org.opensearch.index.mapper.MappedFieldType;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Locale;
+
+abstract class CollapseDocSourceGroupSelector<T> extends GroupSelector<T> {
```
martin-gaievski, line 34: add class level javadoc

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -144,16 +155,42 @@ public Collector newCollector() {
     }
 
     private Collector getHybridQueryCollector() {
-        if (sortAndFormats == null) {
-            return new HybridTopScoreDocCollector(numHits, hitsThresholdChecker);
+        if (collapseContext != null) {
```
martin-gaievski: we need a factory for hybrid query collectors. Can you draft something simple - pack all needed class fields into DTO class and copy logic from this `getCollector` method to `createCollector` (or pick other name) method of that factory.

martin-gaievski (commented):
Finished first iteration of review

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
```diff
@@ -115,7 +132,12 @@ public CompoundTopDocs(final QuerySearchResult querySearchResult) {
                 scoreDocList.clear();
             } else {
                 scoreDocList.add(scoreDoc);
+                if (isCollapseEnabled) {
+                    CollapseTopFieldDocs collapseTopFieldDocs = (CollapseTopFieldDocs) topDocs;
+                    collapseValueList.add(collapseTopFieldDocs.collapseValues[collapseIndex]);
+                }
             }
+            collapseIndex++;
```
vibrantvarun: if the collapseIndex is used for iterating over the collapse values found in CollapseTopFieldDocs then why this counter is iterating with parent for loop?

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -226,6 +253,211 @@ private void updateOriginalQueryResults(final CombineScoresDto combineScoresDTO,
...
+        int indexOfFirstNonEmpty,
+        final boolean isFetchPhaseExecuted,
+        CombineScoresDto combineScoresDTO
+    ) {
+        int totalScoreDocsCount = 0;
+        Map<BytesRef, FieldDoc> collapseValueToTopScoreDocMap = new HashMap<>();
+        Map<BytesRef, Integer> collapseValueToShardMap = new HashMap<>();
+        String collapseField = "";
+
+        HybridQueryFieldDocComparator comparator = new HybridQueryFieldDocComparator(
```
vibrantvarun: This comparator is common in both the methods collapseKeyword and collapseNumeric. Pull this comparator out and declare it as static.

Also pull out all the things which are common put it in a separate method.

src/main/java/org/opensearch/neuralsearch/search/collector/CollapseDocSourceGroupSelector.java
```diff
@@ -0,0 +1,262 @@
...
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Locale;
+
+/**
+ * Group selector that selects based on a collapse value
+ * @param <T> Type of collapse value: Numeric or Keyword
+ */
+abstract class CollapseDocSourceGroupSelector<T> extends GroupSelector<T> {
```
vibrantvarun, line 34: Is this class is a copy of some class from lucene? If yes then add reference in java doc like https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/search/lucene/MultiLeafFieldComparator.java#L15

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -0,0 +1,316 @@
...
+    protected final String collapseField;
+    private int totalHitCount;
+    private float maxScore = 0.0f;
+    private Sort sort;
+    private final GroupSelector<T> groupSelector;
+    private final int[] reversed;
+    private final boolean needsScores;
+    private int docBase;
+    private final ConcurrentHashMap<T, FieldValueHitQueue<FieldValueHitQueue.Entry>[]> groupQueueMap;
+    private ConcurrentHashMap<T, int[]> collectedHitsPerSubQueryMap;
```
vibrantvarun: I see a lop of concurrentHashMaps declare below. Any specific reason of doing that. ConcurrentHashMaps are slower.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -202,20 +207,42 @@ private void updateOriginalQueryResults(final CombineScoresDto combineScoresDTO,
...
+        int indexOfFirstNonEmpty = 0;
+        for (CompoundTopDocs compoundTopDocs : queryTopDocs) {
+            if (!compoundTopDocs.getTopDocs().isEmpty() && compoundTopDocs.getTopDocs().getFirst() instanceof CollapseTopFieldDocs) {
+                isCollapseEnabled = true;
+                break;
+            }
+            indexOfFirstNonEmpty++;
+        }
+        if (isCollapseEnabled) {
+            totalScoreDocsCount = collapse(
```
vibrantvarun: Here we need totalScoreDocsCount But the method which returns it is named as collapse. Can you follow a principle of every method should do one specific type of operation and should be returning the information related to it?

Single Responsibility principle.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -202,20 +207,42 @@ private void updateOriginalQueryResults(final CombineScoresDto combineScoresDTO,
...
+                sort,
+                isFetchPhaseExecuted,
+                combineScoresDTO
             );
-            // Fetch Phase had ran before the normalization phase, therefore update the from value in result of each shard.
-            // This will ensure the trimming of the search results.
-            if (isFetchPhaseExecuted) {
-                querySearchResult.from(combineScoresDTO.getFromValueForSingleShard());
+        } else {
+            for (int shardIndex = 0; shardIndex < querySearchResults.size(); shardIndex++) {
```
vibrantvarun, line 232: Can this piece of code and the code used later in updateOriginalQuerySearchResultWithCollapseEnabled has similar structure. Can we combine them?

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -0,0 +1,316 @@
...
+            new CollapseDocSourceGroupSelector.Numeric(fieldType),
+            collapseField,
+            sort,
+            topNGroups,
+            hitsThresholdChecker
+        );
+    }
+
+    @Override
+    public List<CollapseTopFieldDocs> topDocs() throws IOException {
```
vibrantvarun, line 156: Add comments in the code. Please explain what this method is doing by adding High level comment on major operations.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -0,0 +1,316 @@
...
+        groupSelector.setNextReader(context);
+        return new HybridLeafCollector() {
+            ConcurrentHashMap<T, Boolean> initializeLeafComparatorsPerSegmentOnceMap;
+
+            {
+                initializeLeafComparatorsPerSegmentOnceMap = new ConcurrentHashMap<>();
+            }
+
+            @Override
+            public void collect(int doc) throws IOException {
```
vibrantvarun, line 327: Break down this method code by using Single Responsibility Principle. I am finding it hard time to understand it.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -0,0 +1,316 @@
...
+                            fieldValueLeafTrackers[i].doc = docBase + doc;
+                            fieldValueLeafTrackers[i] = compoundScores[i].updateTop();
+                            comparators[i].setBottom(fieldValueLeafTrackers[i].slot);
+
+                            comparatorsMap.put(groupSelector.copyValue(), comparators);
+                            fieldValueLeafTrackersMap.put(groupSelector.copyValue(), fieldValueLeafTrackers);
+                            groupQueueMap.put(groupSelector.copyValue(), compoundScores);
+                        }
+                    } else {
+                        int[] collectedHitsForCurrentSubQuery = collectedHitsPerSubQueryMap.get(groupValue);
```
vibrantvarun: There is lot happening below. I am not able to understand it completely. Can you take reference from HybridTopDocFieldCollector to see how we can create updateBottom like methods to segregate it properly.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -169,38 +170,54 @@ private Collector getHybridQueryCollector() {
...
-    private TopDocsAndMaxScore getTopDocsAndAndMaxScore(final HybridSearchCollector hybridSearchCollector, final boolean isSortEnabled) {
+    private TopDocsAndMaxScore getTopDocsAndAndMaxScore(
+        final HybridSearchCollector hybridSearchCollector,
+        final boolean isSortEnabled,
+        final boolean isCollapseEnabled
+    ) throws IOException {
         List topDocs = hybridSearchCollector.topDocs();
-        if (isSortEnabled) {
-            return getSortedTopDocsAndMaxScore(topDocs, hybridSearchCollector);
+        if (isSortEnabled || isCollapseEnabled) {
```
vibrantvarun, line 212: Can sort and collapse work together? Then what would happen in that situation?

ryanbogan: Yes, they work together and in that case the collapse code would be followed with the sort passed into it.  The only reason for checking both is the case where only one of them is enabled.  If collapse is enabled without sort, we still need to create FieldDocs instead of ScoreDocs because CollapseTopFieldDocs uses them.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
ryanbogan: I'll replace them with normal hash maps

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
ryanbogan: I refactored this part when collapse is enabled into a separate class in my last commit

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
vibrantvarun: Also use Map like private Map<T, int[]> collectedHitsPerSubQueryMap;

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
ryanbogan: Refactored this method

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
ryanbogan: Refactored this method

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
ryanbogan: Good catch.  This is leftover from a change in implementation I made with how the collapse values were combined into one list earlier in query phase.  This variable can be removed and just replaced with the index in the for loop.

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
```diff
@@ -42,6 +43,7 @@ public class CompoundTopDocs {
     private List<ScoreDoc> scoreDocs;
     @Getter
     private SearchShard searchShard;
+    private final int COLLAPSE_STARTING_INDEX = 2;
```
owaiskazi19: Not using this anywhere?

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
```diff
@@ -99,14 +104,25 @@ public CompoundTopDocs(final QuerySearchResult querySearchResult) {
...
+        List<Object> collapseValueList = new ArrayList<>();
         for (int index = 2; index < scoreDocs.length; index++) {
             // getting first element of score's series
             ScoreDoc scoreDoc = scoreDocs[index];
             if (isHybridQueryDelimiterElement(scoreDoc) || isHybridQueryStartStopElement(scoreDoc)) {
                 ScoreDoc[] subQueryScores = scoreDocList.toArray(new ScoreDoc[0]);
                 TotalHits totalHits = new TotalHits(subQueryScores.length, TotalHits.Relation.EQUAL_TO);
                 TopDocs subQueryTopDocs;
-                if (isSortEnabled) {
+                if (isCollapseEnabled) {
```
owaiskazi19, line 114:
```suggestion
                if (isCollapseEnabled && topDocs instanceof CollapseTopFieldDocs collapseDocs) {
```
then you can remove L116

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
```diff
@@ -115,6 +131,10 @@ public CompoundTopDocs(final QuerySearchResult querySearchResult) {
                 scoreDocList.clear();
             } else {
                 scoreDocList.add(scoreDoc);
+                if (isCollapseEnabled) {
```
owaiskazi19, line 133: Same can be done here as above

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -202,20 +205,46 @@ private void updateOriginalQueryResults(final CombineScoresDto combineScoresDTO,
...
-            QuerySearchResult querySearchResult = querySearchResults.get(index);
-            CompoundTopDocs updatedTopDocs = queryTopDocs.get(index);
-            totalScoreDocsCount += updatedTopDocs.getScoreDocs().size();
-            TopDocsAndMaxScore updatedTopDocsAndMaxScore = new TopDocsAndMaxScore(
-                buildTopDocs(updatedTopDocs, sort),
-                maxScoreForShard(updatedTopDocs, sort != null)
+
+        // Get index of first non-empty CompoundTopDocs to check if collapse is enabled
+        boolean isCollapseEnabled = false;
+        int indexOfFirstNonEmpty = 0;
```
owaiskazi19: We should keep this as `-1` if not found, otherwise it will always pick up index 0

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -202,20 +205,46 @@ private void updateOriginalQueryResults(final CombineScoresDto combineScoresDTO,
...
+        for (CompoundTopDocs compoundTopDocs : queryTopDocs) {
+            if (!compoundTopDocs.getTopDocs().isEmpty() && compoundTopDocs.getTopDocs().getFirst() instanceof CollapseTopFieldDocs) {
```
owaiskazi19: Get topDocs once and use it
```
  List<TopDocs> topDocsList = queryTopDocs.get(i).getTopDocs();
```

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -202,20 +205,46 @@ private void updateOriginalQueryResults(final CombineScoresDto combineScoresDTO,
...
+                isCollapseEnabled = true;
+                break;
+            }
+            indexOfFirstNonEmpty++;
```
owaiskazi19: Also, let's move the check to see if collapse is enabled to a separate method

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -202,20 +205,46 @@ private void updateOriginalQueryResults(final CombineScoresDto combineScoresDTO,
...
-            CompoundTopDocs updatedTopDocs = queryTopDocs.get(index);
-            totalScoreDocsCount += updatedTopDocs.getScoreDocs().size();
-            TopDocsAndMaxScore updatedTopDocsAndMaxScore = new TopDocsAndMaxScore(
-                buildTopDocs(updatedTopDocs, sort),
-                maxScoreForShard(updatedTopDocs, sort != null)
+
+        // Get index of first non-empty CompoundTopDocs to check if collapse is enabled
+        boolean isCollapseEnabled = false;
+        int indexOfFirstNonEmpty = 0;
+        for (CompoundTopDocs compoundTopDocs : queryTopDocs) {
```
owaiskazi19: Use index based for loop, avoid creating iterator objects

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
owaiskazi19:
```suggestion
        int  firstNonEmptyIndex= -1;
```

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -202,20 +205,46 @@ private void updateOriginalQueryResults(final CombineScoresDto combineScoresDTO,
...
+            if (!compoundTopDocs.getTopDocs().isEmpty() && compoundTopDocs.getTopDocs().getFirst() instanceof CollapseTopFieldDocs) {
+                isCollapseEnabled = true;
```
owaiskazi19, line 214: Add `firstNonEmptyIndex = i;` here

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
owaiskazi19: Remove this counter since, it's already happening with 

Add firstNonEmptyIndex = i;

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseDataCollector.java
```diff
@@ -0,0 +1,192 @@
...
+            ((CollapseTopFieldDocs) collapseDTO.getCollapseQueryTopDocs()
+                .get(collapseDTO.getIndexOfFirstNonEmpty())
+                .getTopDocs()
+                .getFirst()).fields,
+            Comparator.comparing((ScoreDoc scoreDoc) -> scoreDoc.score)
+        );
+
+        this.expectedType = determineExpectedType(collapseDTO);
+        if (this.expectedType == null) {
+            throw new IllegalArgumentException("Could not determine collapse value type from input data");
```
owaiskazi19, line 57: Need a test to check this

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseDataCollector.java
```diff
@@ -0,0 +1,192 @@
...
+                        Locale.ROOT,
+                        "Expected CollapseTopFieldDocs but got: %s",
+                        updatedCollapseTopDocs.getTopDocs().getFirst().getClass().getSimpleName()
+                    )
+                );
+            }
+
+            collapseField = ((CollapseTopFieldDocs) updatedCollapseTopDocs.getTopDocs().getFirst()).field;
+
+            for (ScoreDoc scoreDoc : updatedCollapseDocs) {
```
owaiskazi19: Use index based loop

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseDataCollector.java
owaiskazi19, line 57: Let's the log the error as well

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseProcessor.java
```diff
@@ -0,0 +1,42 @@
...
+import lombok.Getter;
+import org.apache.lucene.search.grouping.CollapseTopFieldDocs;
+import org.apache.lucene.util.BytesRef;
+
+/**
+ * Processes collapse operations on search results.
+ * Determines the appropriate collapse strategy based on the field type and executes the collapse.
+ */
+@Getter
+public class CollapseProcessor {
```
owaiskazi19: We should not name this as Processor. Let's call it as CollapseExecutor or CollapseExecutorService

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseProcessor.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Executes the collapse operation based on the provided collapse data.
+     * Determines whether the collapse is for a keyword or numeric field,
+     * creates the appropriate strategy, and performs the collapse.
+     *
+     * @param collapseDTO Data transfer object containing collapse configuration and results
+     */
+    public void executeCollapse(CollapseDTO collapseDTO) {
+        boolean isKeywordCollapse = isKeywordCollapseType(collapseDTO);
```
owaiskazi19:
```suggestion
        boolean isKeywordCollapse = isCollapseOnKeywordField(collapseDTO);
```

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseProcessor.java
```diff
@@ -0,0 +1,42 @@
...
+        CollapseStrategy collapseStrategy = isKeywordCollapse
+            ? CollapseStrategy.createKeywordStrategy()
+            : CollapseStrategy.createNumericStrategy();
+
+        collapseStrategy.executeCollapse(collapseDTO);
```
owaiskazi19: You can return the count `totalCollapsedDocsCount` directly from `collapseStrategy.executeCollapse`

owaiskazi19 (commented):
Great work @ryanbogan ! Finished 1st round of review

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
ryanbogan: Ah I just removed the use of that in last commit.

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
ryanbogan: This check already happens when isCollapseEnabled is set:
```
if (topDocs instanceof CollapseTopFieldDocs) {
            isCollapseEnabled = true;
        } else if (topDocs instanceof TopFieldDocs) {
            isSortEnabled = true;
        }
```

Therefore, if we add it to the if statement then the second half of the check would always be true.

src/main/java/org/opensearch/neuralsearch/processor/CompoundTopDocs.java
ryanbogan: Responded above

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
ryanbogan: Wouldn't it be the equal to the length of queryTopDocs if nothing is found?

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
ryanbogan: I kept it in this method because we are also finding the indexOfFirstNonEmpty in the same loop.  The alternative would be to return a Pair and then assign variables after that which I felt was too convoluted

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
ryanbogan: Changed it per suggestions below

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseProcessor.java
ryanbogan: @vibrantvarun pointed out that that wouldn't follow the Single Responsibility Principle, so I changed it to not return directly

owaiskazi19: It does not violate the Single Responsibility Principle — in fact, it improves adherence to it. 
The strategy's sole responsibility is to execute the collapse logic and return the result (e.g., the total number of collapsed docs). It does not mutate external state or take responsibility for tracking that count elsewhere.

With the new way CollapseStrategy is stateless and focused on behavior. The caller (CollapseProcessor) decides what to do with the result

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -247,7 +277,7 @@ private List<CompoundTopDocs> getCompoundTopDocs(CombineScoresDto combineScoresD
      * @param isSortEnabled if sort is enabled or disabled
      * @return  max score
      */
-    private float maxScoreForShard(CompoundTopDocs updatedTopDocs, boolean isSortEnabled) {
+    public static float maxScoreForShard(CompoundTopDocs updatedTopDocs, boolean isSortEnabled) {
```
martin-gaievski: why you need this as static? If you need utility method that needs to be pulled to another utility class.

src/main/java/org/opensearch/neuralsearch/search/collector/CollapseDocSourceGroupSelector.java
```diff
@@ -0,0 +1,264 @@
...
+                            public long longValue() throws IOException {
+                                return value;
+                            }
+
+                        };
+                    }
+                    break;
+
+                default:
+                    throw new IllegalStateException("unexpected doc values type " + type + "` for field `" + field + "`");
```
martin-gaievski: use String.format instead of direct string concatenation

martin-gaievski: same for few other similar calls in below code

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollectorFactory.java
```diff
@@ -0,0 +1,85 @@
...
+import org.opensearch.index.mapper.MappedFieldType;
+import org.opensearch.index.mapper.NumberFieldMapper;
+import org.opensearch.neuralsearch.search.HitsThresholdChecker;
+import org.opensearch.search.collapse.CollapseContext;
+import org.opensearch.search.internal.SearchContext;
+import org.opensearch.search.sort.SortAndFormats;
+
+import java.util.Locale;
+
+public class HybridCollectorFactory {
```
martin-gaievski, line 24: please add class level comments, some for public methods in this class

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -268,13 +285,15 @@ private List<HybridSearchCollector> getHybridSearchCollectors(final Collection<C
...
-            } else if (collector instanceof HybridTopScoreDocCollector || collector instanceof HybridTopFieldDocSortCollector) {
-                hybridSearchCollectors.add((HybridSearchCollector) collector);
-            } else if (collector instanceof FilteredCollector
-                && (((FilteredCollector) collector).getCollector() instanceof HybridTopScoreDocCollector
-                    || ((FilteredCollector) collector).getCollector() instanceof HybridTopFieldDocSortCollector)) {
-                        hybridSearchCollectors.add((HybridSearchCollector) ((FilteredCollector) collector).getCollector());
-                    }
+            } else if (collector instanceof HybridTopScoreDocCollector
+                || collector instanceof HybridTopFieldDocSortCollector
+                || collector instanceof HybridCollapsingTopDocsCollector) {
```
martin-gaievski: please extract condition inside this if into a separate method, this syntax became hard to read

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -268,13 +285,15 @@ private List<HybridSearchCollector> getHybridSearchCollectors(final Collection<C
...
+                    hybridSearchCollectors.add((HybridSearchCollector) collector);
+                } else if (collector instanceof FilteredCollector
+                    && (((FilteredCollector) collector).getCollector() instanceof HybridTopScoreDocCollector
+                        || ((FilteredCollector) collector).getCollector() instanceof HybridTopFieldDocSortCollector)) {
+                            hybridSearchCollectors.add((HybridSearchCollector) ((FilteredCollector) collector).getCollector());
```
martin-gaievski: same as previous comment, please pull condition into separate private method

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -0,0 +1,406 @@
...
+public class HybridCollapsingTopDocsCollector<T> implements HybridSearchCollector, Collector {
+    protected final String collapseField;
+    private int totalHitCount;
+    private float maxScore = 0.0f;
+    private Sort sort;
+    private final GroupSelector<T> groupSelector;
+    private final int[] reversed;
+    private final boolean needsScores = true;
+    private int docBase;
+    private final HashMap<T, FieldValueHitQueue<FieldValueHitQueue.Entry>[]> groupQueueMap;
```
vibrantvarun: Use Map interface here.
```suggestion
    private final Map<T, FieldValueHitQueue<FieldValueHitQueue.Entry>[]> groupQueueMap;
```

vibrantvarun: Similarly wherever possible.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -0,0 +1,406 @@
...
+                        }
+                        fieldDocs.add(new FieldDoc(queueEntry.doc, queueEntry.score, fields));
+                        collapseValues.add(groupValue instanceof BytesRef ? (T) BytesRef.deepCopyOf((BytesRef) groupValue) : groupValue);
+                    } else {
+                        // Break if queue is empty
+                        break;
+                    }
+                }
+                hitsOnCurrentSubQuery += collectedHitsPerSubQueryMap.get(groupValue)[i];
+            }
```
vibrantvarun: Could you add high level comment on what is happening here.

src/test/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollectorTests.java
```diff
@@ -0,0 +1,226 @@
...
+    private static final String TEXT_FIELD_NAME = "field";
+    private static final String INT_FIELD_NAME = "integerField";
+    private static final String COLLAPSE_FIELD_NAME = "collapseField";
+    private static final String TEST_QUERY_TEXT = "greeting";
+    private static final String TEST_QUERY_TEXT2 = "salute";
+    private static final int NUM_DOCS = 4;
+    private static final int TOP_N_GROUPS = 2;
+    private static final int TOTAL_HITS_UP_TO = 1000;
+
+    public void testKeywordCollapse_whenCollectAndTopDocs_thenSuccessful() throws IOException {
```
vibrantvarun, line 47: Try adding a test with 1000 documents ingestion and then test.

src/test/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollectorTests.java
```diff
@@ -0,0 +1,226 @@
...
+            assertTrue(uniqueGroups.contains("group1"));
+            assertTrue(uniqueGroups.contains("group2"));
+        }
+
+        reader.close();
+        writer.close();
+        directory.close();
+    }
+
+    public void testNumericCollapse_whenCollectAndTopDocs_thenSuccessful() throws IOException {
```
vibrantvarun, line 106: Write a test with and without concurrent segement search enabled.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -0,0 +1,406 @@
...
+     * @return A list of CollapseTopFieldDocs
+     * @throws IOException If an I/O error occurs
+     */
+    @Override
+    public List<CollapseTopFieldDocs> topDocs() throws IOException {
+        List<CollapseTopFieldDocs> topDocsList = new ArrayList<>();
+        if (collectedHitsPerSubQueryMap.isEmpty()) {
+            return topDocsList;
+        }
+        // Get num sub queries, there is probably a better way to do this
```
vibrantvarun: Is this comment added by mistake?

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -0,0 +1,406 @@
...
+            @Override
+            public void collect(int doc) throws IOException {
+                HybridSubQueryScorer compoundQueryScorer = getCompoundQueryScorer();
+                if (Objects.isNull(compoundQueryScorer)) {
+                    return;
+                }
+
+                groupSelector.advanceTo(doc);
+                T groupValue = groupSelector.currentValue();
+                assert groupValue != null;
```
vibrantvarun, line 336: Add comment here what does this do

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -0,0 +1,406 @@
...
+                updateHitCount();
+
+                for (int i = 0; i < subScoresByQuery.length; i++) {
+                    float score = subScoresByQuery[i];
+
+                    if (isQueueFull(groupValue, i)) {
+                        updateExistingEntry(groupValue, i, doc);
+                    } else {
+                        addNewEntry(groupValue, i, doc, score);
+                    }
```
vibrantvarun: Add a comment here about its functioning.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -0,0 +1,406 @@
...
+                fieldValueLeafTrackers[subQueryNumber] = compoundScores[subQueryNumber].add(bottomEntry);
+
+                updateMaps(comparators, fieldValueLeafTrackers, compoundScores);
+
+                if (slot == (numHits - 1)) {
+                    boolean[] queueFullArray = queueFullMap.get(groupValue);
+                    queueFullArray[subQueryNumber] = true;
+                    queueFullMap.put(groupSelector.copyValue(), queueFullArray);
+                }
+            }
```
vibrantvarun: Add comments. Lot of complex logic.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -0,0 +1,406 @@
...
+                if (reverseMulMap.get(groupValue) * comparators[index].compareBottom(doc) > 0) {
+                    FieldValueHitQueue.Entry[] fieldValueLeafTrackers = fieldValueLeafTrackersMap.get(groupValue);
+                    FieldValueHitQueue<FieldValueHitQueue.Entry>[] compoundScores = groupQueueMap.get(groupValue);
+
+                    comparators[index].copy(fieldValueLeafTrackers[index].slot, doc);
+                    fieldValueLeafTrackers[index].doc = docBase + doc;
+                    fieldValueLeafTrackers[index] = compoundScores[index].updateTop();
+                    comparators[index].setBottom(fieldValueLeafTrackers[index].slot);
+
+                    updateMaps(comparators, fieldValueLeafTrackers, compoundScores);
```
vibrantvarun: Add comments.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
ryanbogan: It needs to be static because otherwise we would have to pass in the NormalizationProcessorWorkflow object into the collapse logic, which is called from the NormalizationProcessorWorkflow object.  I'll create a new util class and move the method there.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
ryanbogan: Yes, removed

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseProcessor.java
ryanbogan: Makes sense, I changed it

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseResultUpdater.java
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Updates the collapse results based on the provided collapse data.
+     * This method creates new collapsed field documents, updates the top documents,
+     * and adjusts the search results accordingly.
+     *
+     * @param collapseDTO Data transfer object containing collapse configuration and results
+     */
+    public void updateCollapseResults(CollapseDTO collapseDTO) {
+        Object[] objectCollapseValues = collapseDTO.getRelevantCollapseEntries().stream().map(Map.Entry::getKey).toArray(Object[]::new);
```
owaiskazi19, line 58: Add the below check
```
 if (collapseDTO == null || collapseDTO.getRelevantCollapseEntries() == null) {
            return; 
        }
``` 
and handle it in the caller method

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseResultUpdater.java
```diff
@@ -0,0 +1,67 @@
...
+
+        ScoreDoc[] newCollapsedFieldDocs = collapseDTO.getRelevantCollapseEntries()
```
owaiskazi19: Add a comment for different functionalities here or move it to different method. I would suggest move to different methods for build and updating fields. I will leave the choice to you

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseResultUpdater.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Arrays;
+import java.util.Map;
+
+/**
+ * Updates collapse results based on the processed collapse data.
+ * This class is responsible for creating new collapsed field documents and updating the search results.
+ */
+@Getter
+public class CollapseResultUpdater {
+    private int processedCollapsedDocsCount = 0;
```
owaiskazi19: Nit: Can make this as atomic vari

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseResultUpdater.java
```diff
@@ -0,0 +1,67 @@
...
+     *
+     * @param collapseDTO Data transfer object containing collapse configuration and results
+     */
+    public void updateCollapseResults(CollapseDTO collapseDTO) {
+        Object[] objectCollapseValues = collapseDTO.getRelevantCollapseEntries().stream().map(Map.Entry::getKey).toArray(Object[]::new);
+
+        ScoreDoc[] newCollapsedFieldDocs = collapseDTO.getRelevantCollapseEntries()
+            .stream()
+            .map(Map.Entry::getValue)
+            .map(fieldDoc -> new FieldDoc(fieldDoc.doc, fieldDoc.score, Arrays.copyOfRange(fieldDoc.fields, 0, fieldDoc.fields.length - 1)))
```
owaiskazi19: Should we check the fieldDoc length before slicing here?

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseStrategy.java
```diff
@@ -0,0 +1,115 @@
...
+                    shardIndex
+                );
+
+                collapseResultUpdater.updateCollapseResults(collapseDTO);
+                this.totalCollapsedDocsCount += collapseResultUpdater.getProcessedCollapsedDocsCount();
+            }
+        }
+    }
+
+    private static class NumericCollapseStrategy extends CollapseStrategy {
```
owaiskazi19, line 68: The code is almost the same for both the strategies. Let's move it to a common method

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -166,18 +167,37 @@ private Map<Integer, Object[]> getDocIdSortFieldsMap(
         Map<Integer, Object[]> docIdSortFieldMap = new HashMap<>();
         final List<TopDocs> topFieldDocs = compoundTopDocs.getTopDocs();
         final boolean isSortByScore = isSortOrderByScore(sort);
+        final boolean isCollapseEnabled = topFieldDocs.getFirst() instanceof CollapseTopFieldDocs;
         for (TopDocs topDocs : topFieldDocs) {
+            int scoreDocIndex = 0;
             for (ScoreDoc scoreDoc : topDocs.scoreDocs) {
```
owaiskazi19: You can change this to index based loop and remove L172

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -0,0 +1,434 @@
...
+        String collapseField,
+        Sort groupSort,
+        int topNGroups,
+        HitsThresholdChecker hitsThresholdChecker
+    ) {
+        this.groupSelector = groupSelector;
+        this.collapseField = collapseField;
+        this.sort = groupSort;
+        if (topNGroups < 1) {
+            throw new IllegalArgumentException("topNGroups must be >= 1 (got " + topNGroups + ")");
```
owaiskazi19, line 74: Need a test for this

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollectorFactory.java
```diff
@@ -0,0 +1,101 @@
...
+                    fieldType,
+                    sortAndFormats == null ? new Sort(new SortField(null, SortField.Type.SCORE)) : sortAndFormats.sort,
+                    searchContext.size(),
+                    hitsThresholdChecker
+                );
+            } else {
+                throw new IllegalStateException(
+                    String.format(
+                        Locale.ROOT,
+                        "unknown type for collapse field %s, only keywords and numbers are accepted",
```
owaiskazi19, line 64: Need a test for this

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -169,38 +170,54 @@ private Collector getHybridQueryCollector() {
      * @return search results that can be reduced be the caller
      */
     @Override
-    public ReduceableSearchResult reduce(Collection<Collector> collectors) {
+    public ReduceableSearchResult reduce(Collection<Collector> collectors) throws IOException {
```
owaiskazi19, line 180: Is this exception getting swallowed or thrown from `CollectorManager`?

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -268,17 +285,27 @@ private List<HybridSearchCollector> getHybridSearchCollectors(final Collection<C
...
-                    }
+            } else if (isHybridFilteredCollector(collector)) {
+                hybridSearchCollectors.add((HybridSearchCollector) ((FilteredCollector) collector).getCollector());
+            }
         }
         return hybridSearchCollectors;
     }
 
+    private boolean isHybridNonFilteredCollector(Collector collector) {
+        return collector instanceof HybridTopScoreDocCollector
```
owaiskazi19: You can actually create set of collectors and check for instanceOf from the set

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseResultUpdater.java
ryanbogan: Added, I don't think we need to add anything to handle since if it returns there, the processed docs will be 0.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
ryanbogan: The interface definition says it is thrown: https://github.com/apache/lucene/blob/branch_10_2/lucene/core/src/java/org/apache/lucene/search/CollectorManager.java#L55

src/test/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollectorTests.java
ryanbogan: Should that be an integTest? I'm not familiar with how to enable concurrent segment search outside of the index or cluster settings

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseDTO.java
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Data Transfer Object (DTO) for managing collapse-related search data across shards.
+ * Contains both base collapse information and shard-specific details for search operations.
+ */
+@Getter
+public class CollapseDTO {
+    // Base collapse fields
+    private final List<CompoundTopDocs> collapseQueryTopDocs;
+    private final List<QuerySearchResult> collapseQuerySearchResults;
```
martin-gaievski, line 26: if you need "read-only" access for those lists, you need to wrap it into unmodifiable collection, otherwise they can be altered via getter. I assume that's needed because you marked them final and initialize in constructor

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseDTO.java
```diff
@@ -0,0 +1,103 @@
...
+     *
+     * @param collapseQueryTopDocs List of compound top documents from collapse query
+     * @param collapseQuerySearchResults List of search results from collapse query
+     * @param collapseSort Sort criteria for collapse operation
+     * @param indexOfFirstNonEmpty Index of the first non-empty result
+     * @param isFetchPhaseExecuted Flag indicating if  fetch phase is complete
+     * @param collapseCombineScoresDTO DTO containing score combination parameters
+     * @return A new CollapseDTO instance
+     */
+    public static CollapseDTO createInitialCollapseDTO(
```
martin-gaievski: isn't this essentially a call to constructor? If yes, why constructor is private, is this a factory pattern?

ryanbogan: Other DTO classes use a builder, I'll change it to that

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseDTO.java
ryanbogan: I'll do that for the queryTopDocs.  We want to modify the query search results, I must have accidentally copied the final for that variable.

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseDTO.java
ryanbogan: Actually, with the changes requested above I'll just make the constructor public, no reason not to

martin-gaievski:
@ryanbogan do you want to rebase on latest main to fix failing CI?

martin-gaievski:
@ryanbogan I rebased [feature/collapse_hybrid_query](https://github.com/opensearch-project/neural-search/tree/feature/collapse_hybrid_query) on main, now you need to rebase your branch on [feature/collapse_hybrid_query](https://github.com/opensearch-project/neural-search/tree/feature/collapse_hybrid_query)

ryanbogan:
Integration tests are passing on local.  Can someone retry?

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -203,6 +211,52 @@ public List<CollapseTopFieldDocs> topDocs() throws IOException {
...
+                Object valueB = comparator.value(entryB.slot);
+
+                int comparison = comparator.compareValues(valueA, valueB);
+                if (comparison != 0) {
+                    return reverseMul[i] * comparison < 0;
+                }
+            }
+
+            // If all comparisons are equal, use document ID as tie-breaker
+            return entryB.score < entryA.score;
```
martin-gaievski, line 268: is this correct that comment says "use doc id" but actual comparison is for entry.score?

ryanbogan: Yeah this needs to be changed

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -169,38 +177,54 @@ private Collector getHybridQueryCollector() {
...
             boolean isSortEnabled = docValueFormats != null;
-            TopDocsAndMaxScore topDocsAndMaxScore = getTopDocsAndAndMaxScore(collector, isSortEnabled);
-            results.add((QuerySearchResult result) -> reduceCollectorResults(result, topDocsAndMaxScore, docValueFormats));
+            boolean isCollapseEnabled = collector instanceof HybridCollapsingTopDocsCollector;
+            TopDocsAndMaxScore topDocsAndMaxScore = getTopDocsAndAndMaxScore(collector, isSortEnabled, isCollapseEnabled);
+            results.add(
+                (QuerySearchResult result) -> reduceCollectorResults(
+                    result,
+                    topDocsAndMaxScore,
+                    isCollapseEnabled && isSortEnabled == false ? new DocValueFormat[] { DocValueFormat.RAW } : docValueFormats
```
vibrantvarun, line 199:
```suggestion
                    isCollapseEnabled && isSortEnabled == false ? new DocValueFormat[] { DocValueFormat.RAW } : docValueFormats
```

Why we have put `&&` clause here? What if one of them is true and other is false?

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -381,7 +440,12 @@ private TotalHits getTotalHits(int trackTotalHitsUpTo, final List<?> topDocs, fi
         return new TotalHits(maxTotalHits, relation);
     }
 
-    private TopDocs getNewTopFieldDocs(final TotalHits totalHits, final List<TopFieldDocs> topFieldDocs, final SortField sortFields[]) {
+    private TopDocs getNewTopFieldDocs(
```
vibrantvarun, line 443: Can we create a issue to make a generic class which will decide which method to call?
1. getTopDocs
2. getTopFieldDocs
3. getCollapseTopFieldDocs

In that issue we can seperate this coillapse code to a dedicated method and the generic class would decide on the basis of which functionality to use.

cc: @martin-gaievski @owaiskazi19

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
ryanbogan: This check only passes when collapse is enabled but sorting is not.  Since collapse uses field docs, we need to make sure that we pass a non-empty DocValueFormat array to match the sort that we create (sort by score).  If sorting is enabled, then we can just use the formats that are already there.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
owaiskazi19: Always good to refactor. We can create an ENUM and pass to the class to decide on which method to return.

PR #1357, opened 2025-05-28
 Use stack to collect semantic fields to avoid stack overflow

bzhangam:
### Description
 Use stack to collect semantic fields to avoid stack overflow.

### Related Issues
N/A

### Check List


src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
```diff
@@ -37,53 +39,62 @@
...
+            final CollectSemanticFieldStackEntry entry = stack.pop();
+            final Map<String, Object> mapping = entry.mapping;
+            String parentPath = entry.path;
+            int currentDepth = entry.depth;
 
-            // Build the full path for the current field
-            final String fullPath = parentPath.isEmpty() ? fieldName : parentPath + PATH_SEPARATOR + fieldName;
+            if (currentDepth > MAX_DEPTH_OF_INDEX_MAPPING) {
+                throw new IllegalArgumentException(
+                    "Cannot transform the mapping for semantic fields because its depth exceeds the maximum allowed depth "
```
martin-gaievski, line 75: please format error message with String.format, take example from neural-search repo code

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
```diff
@@ -37,53 +39,62 @@
...
+                );
+            }
 
-            if (fieldConfig instanceof Map) {
-                final Map<String, Object> fieldConfigMap = (Map<String, Object>) fieldConfig;
+            for (Map.Entry<String, Object> fieldEntry : mapping.entrySet()) {
+                final String fieldName = fieldEntry.getKey();
+                final Object fieldConfig = fieldEntry.getValue();
+                // Build the full path for the current field
+                final String fullPath = parentPath.isEmpty() ? fieldName : parentPath + PATH_SEPARATOR + fieldName;
```
martin-gaievski, line 84: use String.join or similar existing method instead of direct string concat

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
```diff
@@ -37,53 +39,62 @@
...
 
-                if (isSemanticField(fieldConfigMap)) {
-                    semanticFieldPathToConfigMap.put(fullPath, fieldConfigMap);
-                }
+                if (fieldConfig instanceof Map) {
```
martin-gaievski, line 86: missing logic for "else" case, I assume we can throw illegalArgument exception in this case

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
```diff
@@ -37,53 +39,62 @@
...
      * @param semanticFieldPathToConfigMap path to field config map for semantic fields
      */
+    @SuppressWarnings("unchecked")
     public static void collectSemanticField(
         @NonNull final Map<String, Object> currentMapping,
         @NonNull final Map<String, Map<String, Object>> semanticFieldPathToConfigMap
     ) {
-        collectSemanticField(currentMapping, "", 0, semanticFieldPathToConfigMap);
-    }
+        final Deque<CollectSemanticFieldStackEntry> stack = new ArrayDeque<>();
```
martin-gaievski, line 64:
```suggestion
        final Deque<CollectSemanticFieldStackEntry> semanticFieldEntryStack = new ArrayDeque<>();
```

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/util/SemanticMappingUtils.java
```diff
@@ -37,52 +40,63 @@
...
      * @param semanticFieldPathToConfigMap path to field config map for semantic fields
      */
+    @SuppressWarnings("unchecked")
     public static void collectSemanticField(
         @NonNull final Map<String, Object> currentMapping,
         @NonNull final Map<String, Map<String, Object>> semanticFieldPathToConfigMap
     ) {
-        collectSemanticField(currentMapping, "", 0, semanticFieldPathToConfigMap);
-    }
+        final Deque<CollectSemanticFieldStackEntry> semanticFieldEntryStack = new ArrayDeque<>();
```
vibrantvarun, line 60: Why we used deque here? Why not simple Queue?

vibrantvarun: I didn't find any usecase of using both ends of queue in this implementation.

bzhangam: I think deque can be faster here since queue has locking and thread-safety mechanisms but we don't need it here.

martin-gaievski:
looks good, I think this can be in changelog if this is a bug you or someone else actually bumped into. Clarify the changelog this and I'll approve PR

PR #1358, opened 2025-05-28
Create a util class for chunking helper functions.

bzhangam:
### Description
Create a util class for chunking helper functions.

### Related Issues
N/A

### Check List


martin-gaievski (approved):
We need a unit test for this new utility class. Please create new GH issue to track this, not a blocker for this PR though.

bzhangam:
> Please create new GH issue to track this, not a blocker for this PR though.

They are kind verified through the tests of the chunking processor. But I agree we should have dedicated unit tests for the util class which I can add.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -317,12 +319,9 @@ private void executeChunkers(@NonNull final SemanticFieldInfo semanticFieldInfo,
...
             final List<String> chunkedText = new ArrayList<>();
             if (isFirstChunker) {
-                chunkedText.addAll(chunker.chunkString(semanticFieldInfo.getValue(), runtimeParameters));
+                chunkedText.addAll(chunkString(chunker, semanticFieldInfo.getValue(), runtimeParameters));
             } else {
-                for (String chunk : chunks) {
-                    final List<String> chunkedTextTemp = chunker.chunkString(chunk, runtimeParameters);
-                    chunkedText.addAll(chunkedTextTemp);
-                }
+                chunkedText.addAll(chunkList(chunker, chunks, runtimeParameters));
```
yuye-aws, line 324: Can you also consider nested type like map and list in semantic field? Then, we can have the same implementation as the text ingestion processor

src/main/java/org/opensearch/neuralsearch/processor/util/ChunkUtils.java
```diff
@@ -0,0 +1,71 @@
...
+
+import static org.opensearch.neuralsearch.processor.chunker.Chunker.CHUNK_STRING_COUNT_FIELD;
+import static org.opensearch.neuralsearch.processor.chunker.Chunker.DISABLED_MAX_CHUNK_LIMIT;
+import static org.opensearch.neuralsearch.processor.chunker.Chunker.MAX_CHUNK_LIMIT_FIELD;
+import static org.opensearch.neuralsearch.processor.chunker.ChunkerParameterParser.parseInteger;
+
+/**
+ * Utility class for text chunking. It holds reusable chunking related helper functions.
+ */
+public class ChunkUtils {
```
yuye-aws, line 23: I would recommend also adding the chunk leaf type, chunk map into this util class, so that the semantic field processor supports map type chunking

bzhangam: It's not a valid use case for the semantic field and also don't plan to support it in near future so I prefer not add them for now. If we do need them in future we can easily add them then.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
heemin32: @yuye-aws Thanks for the comment. The current semantic field doesn't require the additional functionality you've suggested at this time. We can certainly add it later if the need arises. So, I'll go ahead and merge this PR for now. If any of your concerns need to be addressed in the future, let's track them with a GitHub issue.

Same for your another comment below.

yuye-aws: Make sense

PR #1359, opened 2025-05-28
Validate model id and analyzer should not be provided at the same time for the neural sparse query

bzhangam:
### Description
Validate model id and analyzer should not be provided at the same time for the neural sparse query. We need this validation because if both analyzer and model id are provided we don't know which one we should use to tokenize the query text.

Also refactor the code addressing some backward compatible issues. When we introduce the analyzer we modified the `getCopyNeuralSparseQueryBuilderForTwoPhase` function to only set the `queryTokensSupplier` for the copy when the model id is available. If we simply check if the model id exist or not there can be an edge case that if user define a NeuralQueryEnrichProcessor to set the default model id after the two phase processor then we will incorrectly think there is no model id. And later during the rewrite we will try to inference the query text for both the original query and the copy which is against the purpose of using two phase processor.

### Related Issues
N/A

### Check List


src/main/java/org/opensearch/neuralsearch/query/visitor/NeuralSearchQueryVisitor.java
```diff
@@ -36,18 +38,22 @@ private Boolean isFieldDefaultModelIdApplied(String fieldName) {
...
+
             if (modelInferenceQueryBuilder.modelId() == null) {
-                if (isFieldDefaultModelIdApplied(modelInferenceQueryBuilder.fieldName())) {
-                    String fieldDefaultModelId = (String) neuralFieldMap.get(modelInferenceQueryBuilder.fieldName());
-                    modelInferenceQueryBuilder.modelId(fieldDefaultModelId);
+                if (isFieldMapApplied(fieldName, neuralFielIddMap)) {
+                    modelInferenceQueryBuilder.modelId(neuralFielIddMap.get(fieldName));
                 } else if (Objects.nonNull(modelId)) {
                     modelInferenceQueryBuilder.modelId(modelId);
-                } else {
```
bzhangam, line 47: I think we should not do the validation in the visitor. If no model id provided then we simply do nothing. We should let the query builder itself to validate if the model id is missing or not.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -184,20 +188,24 @@ public NeuralSparseQueryBuilder getCopyNeuralSparseQueryBuilderForTwoPhase(float
...
+
+        // If the raw tokes are provided through the query directly then split it directly
         if (Objects.nonNull(this.queryTokensSupplier)) {
             Map<String, Float> tokens = queryTokensSupplier.get();
             // Splitting tokens based on a threshold value: tokens greater than the threshold are stored in v1,
             // while those less than or equal to the threshold are stored in v2.
             Tuple<Map<String, Float>, Map<String, Float>> splitTokens = PruneUtils.splitSparseVector(pruneType, pruneRatio, tokens);
             this.queryTokensSupplier(() -> splitTokens.v1());
             copy.queryTokensSupplier(() -> splitTokens.v2());
-        } else if (shouldUseModelId()) {
```
bzhangam, line 199: If we simply check if the model id exist or not there can be an edge case that if user define a NeuralQueryEnrichProcessor to set the default model id after the two phase processor then we will incorrectly think there is no model id. So modify it to always set the queryTokensSupplier and when we get inference we will check the neuralSparseQueryTwoPhaseInfo.getStatus() to make it clear that we only need to split the tokens if we are processing the query on phase one.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -184,20 +188,24 @@ public NeuralSparseQueryBuilder getCopyNeuralSparseQueryBuilderForTwoPhase(float
...
-            this.twoPhaseSharedQueryToken = new HashMap<>();
```
bzhangam, line 200: Remove this so that later in doToQuery we can check if queryTokensSupplier.get() is not null to decide should we simply use it or we should try to find the analyzer to create the query tokens.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -375,6 +385,33 @@ protected QueryBuilder doRewrite(QueryRewriteContext queryRewriteContext) {
             .neuralSparseQueryTwoPhaseInfo(neuralSparseQueryTwoPhaseInfo);
     }
 
+    private boolean shouldUseAnalyzer() {
+        if (modelId != null && analyzer != null) {
+            throw new IllegalArgumentException(
+                String.format(
+                    Locale.ROOT,
+                    "We found both %s:%s and %s:%s are provided to the neural sparse query and we cannot decide which one should be used to tokenize the query_text. Please just provide one of them. You may provide them through the query and the %s search processor.",
```
martin-gaievski, line 393: this error message is a bit lengthy, can you use something shorter, like following one. that's fine if we have less details, it's not a documentation:

"Cannot use both [%s: %s] and [%s: %s] for neural sparse query tokenization. Specify only one tokenization method. These parameters can be set either in the query or through the %s search processor."

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -442,8 +485,11 @@ Map<String, Float> getQueryTokens(QueryShardContext context) {
             } catch (BufferUnderflowException e) {
                 throw new OpenSearchException("failed to parse query token weight from analyzer. ", e);
             }
+        } else {
+            throw new IllegalArgumentException(
+                "Query tokens cannot be null when try to convert neural sparse query builder to a lucene query."
```
martin-gaievski, line 490:
```suggestion
                "Cannot convert neural sparse query to Lucene query: query tokens must not be null."
```

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -189,15 +191,19 @@ public NeuralSparseQueryBuilder getCopyNeuralSparseQueryBuilderForTwoPhase(float
             .neuralSparseQueryTwoPhaseInfo(
                 new NeuralSparseQueryTwoPhaseInfo(NeuralSparseQueryTwoPhaseInfo.TwoPhaseStatus.PHASE_TWO, pruneRatio, pruneType)
             );
+
+        // If the raw tokes are provided through the query directly then split it directly
```
martin-gaievski, line 195: It a bit hard to get the idea of this comment, I put below my understanding, feel free to edit it
```suggestion
        // If raw tokens are provided directly in the query, split them without additional processing
```

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -189,15 +191,19 @@ public NeuralSparseQueryBuilder getCopyNeuralSparseQueryBuilderForTwoPhase(float
...
             Tuple<Map<String, Float>, Map<String, Float>> splitTokens = PruneUtils.splitSparseVector(pruneType, pruneRatio, tokens);
             this.queryTokensSupplier(() -> splitTokens.v1());
             copy.queryTokensSupplier(() -> splitTokens.v2());
-        } else if (shouldUseModelId()) {
-            this.twoPhaseSharedQueryToken = new HashMap<>();
+        } else {
+            // Otherwise delay the split work to when the query tokens are available.
+            // Here we leverage the twoPhaseSharedQueryToken to pass the tokens generated by ML model for the phase one
+            // query to the phase two query to save one inference call.
             copy.queryTokensSupplier(() -> this.twoPhaseSharedQueryToken);
```
martin-gaievski, line 207: do we use two phase token for performance reason? Any cons of using it?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: > there can be an edge case that if user define a NeuralQueryEnrichProcessor to set the default model id after the two phase processor then we will incorrectly think there is no model id

Can we cover this in integ test?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
bzhangam: Yes, it can improve search performance. The downside is that it might affect relevance and scoring, as it uses only a subset of tokens to retrieve the top documents. However, according to this [blog](https://opensearch.org/blog/introducing-a-neural-sparse-two-phase-algorithm/), the impact on relevance is minimal.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
bzhangam: ok

martin-gaievski:
Looks good to me, please add changelog entry and I'll approved

bzhangam:
> @bzhangam Could you either tag the issue this PR is trying to resolve or add a complete description of the purpose of this PR.
> 
> > Validate model id and analyzer should not be provided at the same time for the neural sparse query.
> 
> Please add in description What problem occurs when we provide modelId and analyzer together. How this PR solves that.
> 
> > Also refactor the code addressing some backward compatible issues.
> 
> Please mention the cases when the issue arises.

Ok.

vibrantvarun (approved):
Looks good to me.

PR #1361, opened 2025-06-02
add integration tests for semantic field

will-hwang:
### Description
Add integration test cases for semantic field.

Note: All of the added test cases use sparse encoding model, due to build compatibility issue with FAISS knn engine in build

### Related Issues

### Check List


qa/restart-upgrade/src/test/resources/processor/SemanticIndexMappings.json
```diff
@@ -0,0 +1,21 @@
+{
+  "settings": {
+    "index.knn": true
+  },
+  "mappings": {
+    "properties": {
+      "products":{
+        "type":"nested",
```
bzhangam: I think if you are using nested here you will need to build a nested query with the neural query when you validate the search. But now you simply use a neural query.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/SemanticFieldProcessorIT.java
```diff
@@ -0,0 +1,78 @@
...
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+
+import org.opensearch.neuralsearch.query.NeuralQueryBuilder;
+import org.opensearch.neuralsearch.util.TestUtils;
+
+import static org.opensearch.neuralsearch.util.TestUtils.NODES_BWC_CLUSTER;
+
+public class SemanticFieldProcessorIT extends AbstractRestartUpgradeRestTestCase {
```
bzhangam: Why we name it `SemanticFieldProcessorIT` I think here we only only test the ingest part but also test the index creation and search. Should we simply name it `SemanticFieldIT`?

will-hwang: sure, we can name it SemanticFieldIT

qa/restart-upgrade/src/test/resources/processor/SemanticIndexMappings.json
will-hwang: neural query with nested input seems to work too

products.product_description

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -358,6 +359,18 @@ protected void createIndexWithConfiguration(final String indexName, String index
         if (StringUtils.isNotBlank(pipelineName)) {
             indexConfiguration = String.format(LOCALE, indexConfiguration, pipelineName);
         }
+        createIndex(indexName, indexConfiguration);
+    }
+
+    protected void createSemanticIndexWithConfiguration(final String indexName, String indexConfiguration, final String modelId)
```
bzhangam, line 365: Should we create a `BaseNeuralSearchSemanticFieldIT` to extend the `BaseNeuralSearchIT` to hold the functions dedicated for semantic field related test cases to better organize the codes?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -1017,6 +1055,27 @@ protected void indexTheDocument(
...
+    protected void addSemanticDoc(
+        final String index,
+        final String docId,
+        final String semanticInfoField,
+        final List<String> fieldNames,
+        final List<Map<String, Float>> docs
+    ) {
+        Request request = new Request("POST", "/" + index + "/_doc/" + docId + "?refresh=true");
+        XContentBuilder builder = XContentFactory.jsonBuilder().startObject();
+        builder.startObject(semanticInfoField);
```
bzhangam, line 1097: Why we are manually add the doc with the raw embedding rather than add the doc with original data?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
will-hwang: seems like `BaseNeuralSearchIT` holds functions dedicated for all of the base operations. I don't think there's a need to create a different class that extends it

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
will-hwang: this is how existing sparse encoding and text embedding docs are being ingested. Wanted to keep it consistent when ingesting to semantic field index too

bzhangam: I think we should pass the original text to use semantic ingest processor to generate the embedding using the model. This is one important feature we should test.

qa/restart-upgrade/src/test/resources/processor/SemanticIndexMappings.json
bzhangam: I don't think it should work there must be something wrong. When I tested it on my end the search cannot find any matched doc. It's weird in your case it can work especially you add the doc with field `test_field_ semantic_info` but search against `products.product_description`. We should rewrite the query against `products.product_description_semantic_info.embedding`.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/SemanticFieldProcessorIT.java
```diff
@@ -0,0 +1,78 @@
...
+            }
+        }
+
+    }
+
+    private void validateTestIndex(final String modelId) {
+        NeuralQueryBuilder neuralQueryBuilder = NeuralQueryBuilder.builder()
+            .fieldName(INPUT_FIELD)
+            .queryText(TEST_QUERY_TEXT)
+            .modelId(modelId)
```
bzhangam: I think we don't need to provide the model id. This is also something we want to test.

qa/restart-upgrade/src/test/resources/processor/SemanticIndexMappings.json
bzhangam: I think it's working because we are skipping the BWC tests for now?

src/test/java/org/opensearch/neuralsearch/query/NeuralQueryIT.java
```diff
@@ -446,8 +456,46 @@ public void testFilterQuery() {
...
+     * }
+     */
+    @SneakyThrows
+    public void testNeuralQuery_InSemanticField_WithSparseModel() {
+        String modelId = prepareSparseEncodingModel();
+        initializeIndexIfNotExist(TEST_SEMANTIC_INDEX_SPARSE_NAME, modelId);
+        NeuralQueryBuilder neuralQueryBuilder = NeuralQueryBuilder.builder()
+            .fieldName(TEST_SEMANTIC_TEXT_FIELD)
+            .queryText(TEST_QUERY_TEXT)
+            .modelId(modelId)
```
bzhangam: No need to provide the model id.

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
will-hwang: sure, we can do that

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/SemanticFieldProcessorIT.java
will-hwang: okay

qa/restart-upgrade/src/test/resources/processor/SemanticIndexMappings.json
will-hwang: you're right. let me fix the query

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
will-hwang: embedding generation using the model is tested in `src/test/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessorIT.java` Seems like all query test cases use the mock data and actual embedding generation using the model is tested inside processor test cases. let meknow If you believe the added test cases in SemanticFieldProcessorIT are not enough

heemin32: If the ingest side is already covered by separate testing, it should be fine. In fact, that might be preferable, as it helps us isolate whether an issue originates from the ingest or query side.

bzhangam: If it's tested in SemanticFieldProcessorIT I think we should be good.

will-hwang: cool

Issue #1294, opened 2025-04-29
[FEATURE] Individual scores from queries in hybrid qery


silent-vim:
### Is your feature request related to a problem?
This is in regards to Hybrid query. As part of using Hybrid query and reranking we would like to know what is the individual scores of each. For example what is text match, knn match (title), knn match (chunks)

### What solution would you like?
Individual score tracking of each query before normalization

### What alternatives have you considered?
Script score, very poor performance. Tried to see if there is a response processor that can help add the score in the source map before normalization.

### Do you have any additional context?
This is in context of searching multiple sources. Due to normalization the relevance of each query is lost across sources. So if we are doing multisearch then it is hard to know the relative relevance across queries for different sources as the scores get normalized.

martin-gaievski:
@silent-vim thank you for submitting this feature request, I think it looks similar to another issue that has been submitted earlier https://github.com/opensearch-project/neural-search/issues/1180. 

Can you provide more details regarding desired solution:

- output format. I put one option as part of #1180, feel free to provide feedback if that works for you
```
"hits": {
        "total": {
            "value": 3,
            "relation": "eq"
        },
        "max_score": 1.8431856,
        "hits": [
            {
                "_shard": "[index-test][0]",
                "_node": "qnphyXiURve0A4xLRaRjSA",
                "_index": "index-test",
                "_id": "3",
                "_score": 1.8431856,
                "_source": {
                    "name": "Why would he go to all that effort for a free pack of ranch dressing?",
                    "category": "story",
                    "price": 10
                },
               "_hybridization": {
                    "subquery_scores: [  0.94318557, 0.9]
                }
            {
                "_shard": "[index-test][0]",
                "_node": "qnphyXiURve0A4xLRaRjSA",
                "_index": "index-test",
                "_id": "1",
                "_score": 1.0,
                "_source": {
                    "name": "A West Virginia university women 's basketball team , officials , and a small gathering of fans are in a West Virginia arena .",
                    "category": "novel",
                    "price": 20
                },
               "_hybridization": {
                    "subquery_scores: [  0.0, 1.0]
                }
            }
        ]
    }
```

[...]

Issue #1180, opened 2025-02-11
[FEATURE] Return individual scores as part of the hybrid query response


martin-gaievski:
### Is your feature request related to a problem?
Today we are using the [hybrid search feature](https://opensearch.org/docs/latest/search-plugins/hybrid-search/) where we're combining both BM25 and a semantic (knn/semantic) embedding based vector search using normalization processors (score or ranking based). OpenSearch would return the weighted averages for scores.

We need to return individual scores as part of the hybrid query, they can be either before or after normalization.

### What solution would you like?
Response for hybrid query needs to have both normalized/combined scores and individual scores for sub-query. Those scores can be before or after normalization.

For the sake of example let's pick following request with 2 sub-queries:

```
"query": {
        "hybrid": {
            "queries": [
                {
                    "range": {
                    }
                },
                {
                    "knn": {
                        "vector": {
                            "vector": [],
                            "k": 12
                        }
                    }
                }
            ]
      }
}
```

Possible format for hybrid query response:

```
"hits": {
        "total": {
            "value": 3,
            "relation": "eq"
        },
        "max_score": 1.8431856,
        "hits": [
            {
                "_shard": "[index-test][0]",
                "_node": "qnphyXiURve0A4xLRaRjSA",
                "_index": "index-test",
                "_id": "3",
                "_score": 1.8431856,
                "_source": {
                    "name": "Why would he go to all that effort for a free pack of ranch dressing?",
                    "category": "story",
                    "price": 10
                },
               "_hybridization": {
                    "subquery_scores: [  0.94318557, 0.9]
                }
            {
                "_shard": "[index-test][0]",
                "_node": "qnphyXiURve0A4xLRaRjSA",
                "_index": "index-test",
                "_id": "1",
                "_score": 1.0,
                "_source": {
                    "name": "A West Virginia university women 's basketball team , officials , and a small gathering of fans are in a West Virginia arena .",
                    "category": "novel",
                    "price": 20
                },
               "_hybridization": {
                    "subquery_scores: [  0.0, 1.0]
                }
            }
        ]
    }
```

### What alternatives have you considered?
New feature of `explain` https://github.com/opensearch-project/neural-search/issues/658 helps in some scenarios, but it has limitations related to added latency and not suited for use in production on main search path.

andrross:
[...]

PR #1369, opened 2025-06-03
Support sub query raw scores in Hybrid Search Response

owaiskazi19:
### Description

Support sub query raw scores in Hybrid Search Response through `getFetchSubPhases` extension point.
Also, added a small improvement for RRF processors while working on this.

Flag to enable sub query scores while creating the pipeline
```
{
  "phase_results_processors": [
    {
      "normalization-processor": {
        "normalization": {
          "technique": "z_score"
        },
        "sub-query-scores": true,
        "combination": {
          "parameters": {
            "weights": [
              0.3,
              0.4,
              0.3
            ]
          }
        }
      }
    }
  ]
}
```
Response
```
"hits": {
        "total": {
            "value": 2,
            "relation": "eq"
        },
        "max_score": 0.016393442,
        "hits": [
            {
                "_index": "my-nlp-index1",
                "_id": "3",
                "_score": 0.016393442,
                "_source": {
                    "passage_text": "zoo keeper here",
                    "id": "s3"
                },
                "fields": {
                    "hybridization_sub_query_scores": [
                        0.0,
                        0.0,
                        0.26156494
                    ]
                }
            },
            {
                "_index": "my-nlp-index1",
                "_id": "1",
                "_score": 0.016393442,
                "_source": {
                    "passage_text": "hello hi what",
                    "id": "s1"
                },
                "fields": {
                    "hybridization_sub_query_scores": [
                        0.13076457,
                        0.0,
                        0.0
                    ]
                }
            }
        ]
    }
}
```

### Related Issues
Resolves https://github.com/opensearch-project/neural-search/issues/1294, https://github.com/opensearch-project/neural-search/issues/1180 & https://github.com/opensearch-project/neural-search/issues/1419

### Check List

### DIFF
--- a/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/HybridSearchIT.java
+++ b/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/HybridSearchIT.java
@@ -57,7 +57,8 @@ public class HybridSearchIT extends AbstractRollingUpgradeTestCase {
                     SEARCH_PIPELINE_NAME,
                     DEFAULT_NORMALIZATION_METHOD,
                     DEFAULT_COMBINATION_METHOD,
-                    Map.of(PARAM_NAME_WEIGHTS, Arrays.toString(new float[] { 0.3f, 0.7f }))
+                    Map.of(PARAM_NAME_WEIGHTS, Arrays.toString(new float[] { 0.3f, 0.7f })),
+                    false
                 );
                 break;
             case MIXED:
--- a/src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
+++ b/src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
@@ -32,6 +32,7 @@ public final class MinClusterVersionUtil {
 
     // Constant for neural_knn_query version check
     public static final String NEURAL_KNN_QUERY = "neural_knn_query";
+    public static final Version MINIMAL_SUPPORTED_VERSION_SUB_QUERY_SCORES = Version.V_3_1_0;
 
     // Note this minimal version will act as an override
     private static final Map<String, Version> MINIMAL_VERSION_NEURAL = ImmutableMap.<String, Version>builder()
@@ -56,6 +57,10 @@ public final class MinClusterVersionUtil {
 
     public static boolean isClusterOnOrAfterMinReqVersionForStatCategoryFiltering() {
         return NeuralSearchClusterUtil.instance().getClusterMinVersion().onOrAfter(MINIMAL_SUPPORTED_VERSION_STATS_CATEGORY_FILTERING);
+    }
+
+    public static boolean isClusterOnOrAfterMinReqVersionForSubQuerySupport() {
+        return NeuralSearchClusterUtil.instance().getClusterMinVersion().onOrAfter(MINIMAL_SUPPORTED_VERSION_SUB_QUERY_SCORES);
     }
 
     public static boolean isClusterOnOrAfterMinReqVersion(String key) {
--- a/src/main/java/org/opensearch/neuralsearch/processor/HybridScoreRegistry.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/HybridScoreRegistry.java
@@ -6,8 +6,8 @@ package org.opensearch.neuralsearch.processor;
 
 import org.opensearch.action.search.SearchPhaseContext;
 
-import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.Map;
 
 /**
  * Registry to store hybrid score for each search phase context
--- a/src/main/java/org/opensearch/neuralsearch/processor/HybridScoreRegistry.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/HybridScoreRegistry.java
@@ -9,17 +9,34 @@ import org.opensearch.search.internal.SearchContext;
 import java.util.HashMap;
 import java.util.Map;
 
+/**
+ * Registry to store hybrid score for each search context
+ */
 public class HybridScoreRegistry {
     private static final Map<SearchContext, Map<Integer, float[]>> contextToHolder = new HashMap<>();
 
+    /**
+     * Store hybrid score for each search context
+     * @param context
+     * @param scoreMap
+     */
     public static void store(SearchContext context, Map<Integer, float[]> scoreMap) {
         contextToHolder.put(context, scoreMap);
     }
 
+    /**
+     * Get hybrid score for each search context
+     * @param context
+     * @return map of hybrid score
+     */
     public static Map<Integer, float[]> get(SearchContext context) {
         return contextToHolder.get(context);
     }
 
+    /**
+     * Remove hybrid score for each search context
+     * @param context
+     */
     public static void remove(SearchContext context) {
         contextToHolder.remove(context);
     }
--- a/src/main/java/org/opensearch/neuralsearch/processor/HybridScoreRegistry.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/HybridScoreRegistry.java
@@ -13,14 +13,14 @@ import java.util.Map;
  * Registry to store hybrid score for each search context
  */
 public class HybridScoreRegistry {
-    private static final Map<SearchPhaseContext, Map<Integer, float[]>> contextToHolder = new HashMap<>();
+    private static final Map<SearchPhaseContext, Map<String, float[]>> contextToHolder = new HashMap<>();
 
     /**
      * Store hybrid score for each search context
      * @param context
      * @param scoreMap
      */
-    public static void store(SearchPhaseContext context, Map<Integer, float[]> scoreMap) {
+    public static void store(SearchPhaseContext context, Map<String, float[]> scoreMap) {
         contextToHolder.put(context, scoreMap);
     }
 
@@ -29,7 +29,7 @@ public class HybridScoreRegistry {
      * @param context
      * @return map of hybrid score
      */
-    public static Map<Integer, float[]> get(SearchPhaseContext context) {
+    public static Map<String, float[]> get(SearchPhaseContext context) {
         return contextToHolder.get(context);
     }
 
--- a/src/main/java/org/opensearch/neuralsearch/processor/HybridScoreRegistry.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/HybridScoreRegistry.java
@@ -4,7 +4,7 @@
  */
 package org.opensearch.neuralsearch.processor;
 
-import org.opensearch.search.internal.SearchContext;
+import org.opensearch.action.search.SearchPhaseContext;
 
 import java.util.HashMap;
 import java.util.Map;
@@ -13,14 +13,14 @@ import java.util.Map;
  * Registry to store hybrid score for each search context
  */
 public class HybridScoreRegistry {
-    private static final Map<SearchContext, Map<Integer, float[]>> contextToHolder = new HashMap<>();
+    private static final Map<SearchPhaseContext, Map<Integer, float[]>> contextToHolder = new HashMap<>();
 
     /**
      * Store hybrid score for each search context
      * @param context
      * @param scoreMap
      */
-    public static void store(SearchContext context, Map<Integer, float[]> scoreMap) {
+    public static void store(SearchPhaseContext context, Map<Integer, float[]> scoreMap) {
         contextToHolder.put(context, scoreMap);
     }
 
@@ -29,7 +29,7 @@ public class HybridScoreRegistry {
      * @param context
      * @return map of hybrid score
      */
-    public static Map<Integer, float[]> get(SearchContext context) {
+    public static Map<Integer, float[]> get(SearchPhaseContext context) {
         return contextToHolder.get(context);
     }
 
@@ -37,7 +37,7 @@ public class HybridScoreRegistry {
      * Remove hybrid score for each search context
      * @param context
      */
-    public static void remove(SearchContext context) {
+    public static void remove(SearchPhaseContext context) {
         contextToHolder.remove(context);
     }
 }
--- a/src/main/java/org/opensearch/neuralsearch/processor/HybridScoreRegistry.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/HybridScoreRegistry.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.processor;
+
+import org.opensearch.search.internal.SearchContext;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class HybridScoreRegistry {
+    private static final Map<SearchContext, Map<Integer, float[]>> contextToHolder = new HashMap<>();
+
+    public static void store(SearchContext context, Map<Integer, float[]> scoreMap) {
+        contextToHolder.put(context, scoreMap);
+    }
+
+    public static Map<Integer, float[]> get(SearchContext context) {
+        return contextToHolder.get(context);
+    }
+
+    public static void remove(SearchContext context) {
+        contextToHolder.remove(context);
+    }
+}
--- a/src/main/java/org/opensearch/neuralsearch/processor/HybridScoreRegistry.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/HybridScoreRegistry.java
@@ -6,17 +6,17 @@ package org.opensearch.neuralsearch.processor;
 
 import org.opensearch.action.search.SearchPhaseContext;
 
-import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
- * Registry to store hybrid score for each search context
+ * Registry to store hybrid score for each search phase context
  */
 public class HybridScoreRegistry {
-    private static final Map<SearchPhaseContext, Map<String, float[]>> contextToHolder = new HashMap<>();
+    private static final ConcurrentHashMap<SearchPhaseContext, Map<String, float[]>> contextToHolder = new ConcurrentHashMap<>();
 
     /**
-     * Store hybrid score for each search context
+     * Store hybrid score for each search phase context
      * @param context
      * @param scoreMap
      */
@@ -25,7 +25,7 @@ public class HybridScoreRegistry {
     }
 
     /**
-     * Get hybrid score for each search context
+     * Get hybrid score for each search phase context
      * @param context
      * @return map of hybrid score
      */
@@ -34,7 +34,7 @@ public class HybridScoreRegistry {
     }
 
     /**
-     * Remove hybrid score for each search context
+     * Remove hybrid score for each search phase context
      * @param context
      */
     public static void remove(SearchPhaseContext context) {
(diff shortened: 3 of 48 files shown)


src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
```diff
@@ -99,36 +101,56 @@ public Map<DocIdAtSearchShard, ExplanationDetails> explain(List<CompoundTopDocs>
...
         TopDocs topDocs,
         SearchShard searchShard,
         int topDocsIndex,
-        TriConsumer<DocIdAtSearchShard, Float, Integer> scoreProcessor
+        TriConsumer<DocIdAtSearchShard, Float, Integer> scoreProcessor,
+        int topDocsSize,
+        Map<Integer, float[]> docIdToSubqueryScores
     ) {
-        for (ScoreDoc scoreDoc : topDocs.scoreDocs) {
-            float normalizedScore = calculateNormalizedScore(Arrays.asList(topDocs.scoreDocs).indexOf(scoreDoc));
```
owaiskazi19, line 129: This was causing O(N2), changed it to index based reference

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -54,6 +56,7 @@ public class NormalizationProcessorWorkflow {
 
     private final ScoreNormalizer scoreNormalizer;
     private final ScoreCombiner scoreCombiner;
+    private static final String NAME = "hybridization_sub_query_scores";
```
ryanbogan: Can we name this something more specific?

owaiskazi19: Isn't it specific `hybridization_sub_query_scores`? Letting us know about the sub query scores. Anything else you have in mind?

ryanbogan: I meant the constant name being `NAME`.  Maybe we could change to something like `SUB_QUERY_SCORES_NAME` or something similar

martin-gaievski:
can you sahre response for case when both inner hits are present and feature is enabled

martin-gaievski (approved):
Looks good to me, thanks Owais

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/HybridSearchIT.java
bzhangam, line 1: Can we add one IT with sub-query score on for rolling upgrade? Since we are doing version check for this new feature I think it's good to have an IT in rolling upgrade to ensure we can handle the nodes with different versions properly.

src/main/java/org/opensearch/neuralsearch/processor/HybridizationFetchSubPhase.java
```diff
@@ -0,0 +1,67 @@
...
+                        return;
+                    }
+                    int docId = hitContext.docId();
+                    int shardId = fetchContext.getQueryShardContext().getShardId();
+                    String key = shardId + "_" + docId;
+                    float[] subqueryScores = scoreMap.get(key);
+
+                    if (subqueryScores != null) {
+                        // Add it as a field rather than modifying _source
+                        List<Object> hybridScores = new ArrayList<>(subqueryScores.length);
```
bzhangam, line 63: nit: Why not use `List<Float>`? I think here you use the score as float `float[] subqueryScores = scoreMap.get(key);`.

src/main/java/org/opensearch/neuralsearch/processor/factory/RRFProcessorFactory.java
```diff
@@ -54,7 +57,9 @@ public SearchPhaseResultsProcessor create(
         ScoreCombinationTechnique scoreCombinationTechnique = scoreCombinationFactory.createCombination(
             RRFScoreCombinationTechnique.TECHNIQUE_NAME
         );
+
         Map<String, Object> combinationClause = readOptionalMap(RRFProcessor.TYPE, tag, config, COMBINATION_CLAUSE);
+        boolean subQueryScores = readBooleanProperty(RRFProcessor.TYPE, tag, config, SUB_QUERY_SCORES, DEFAULT_SUB_QUERY_SCORES);
```
bzhangam: Can we rename it to make it more clear this is a boolean to improve the code readability? e.g. `subQueryScoresEnabled`

src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
```diff
@@ -66,6 +76,13 @@ public void normalize(NormalizeScoresDTO normalizeScoresDTO) {
             for (int j = 0; j < topDocsPerSubQuery.size(); j++) {
                 TopDocs subQueryTopDoc = topDocsPerSubQuery.get(j);
                 for (ScoreDoc scoreDoc : subQueryTopDoc.scoreDocs) {
+                    // Initialize or update subquery scores array per doc
+                    if (normalizeScoresDTO.isSubQueryScores()) {
```
bzhangam: nit: This code block logic is shared by all the techniques can we make it a util function?

src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
```diff
@@ -58,6 +59,10 @@ public static boolean isClusterOnOrAfterMinReqVersionForStatCategoryFiltering()
         return NeuralSearchClusterUtil.instance().getClusterMinVersion().onOrAfter(MINIMAL_SUPPORTED_VERSION_STATS_CATEGORY_FILTERING);
     }
 
+    public static boolean isClusterOnOrAfterMinReqVersionForSubQuerySupport() {
```
vibrantvarun, line 62: We need one BWC test to validate this condition does not break anything.

src/main/java/org/opensearch/neuralsearch/processor/HybridScoreRegistry.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package org.opensearch.neuralsearch.processor;
+
+import org.opensearch.action.search.SearchPhaseContext;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Registry to store hybrid score for each search context
```
vibrantvarun: As per my understanding, there is only one searchcontext per search. Therefore, can we reframe this javadoc in a specific way like "Registry to store hybrid score for each document" ?

vibrantvarun: In the later part, I see we are using searchPhaseContext. SearchPhaseContext can be multiple like after the QP , then after the FP as well.

src/main/java/org/opensearch/neuralsearch/processor/HybridizationFetchSubPhase.java
```diff
@@ -0,0 +1,67 @@
...
+ * Fetch sub phase to add hybridization scores to the search response
+ */
+public class HybridizationFetchSubPhase implements FetchSubPhase {
+
+    private static final String SUB_QUERY_SCORES_NAME = "hybridization_sub_query_scores";
+
+    @Override
+    public FetchSubPhaseProcessor getProcessor(FetchContext fetchContext) throws IOException {
+        // Check if inner hits are present
+        boolean hasInnerHits = fetchContext.innerHits() != null && !fetchContext.innerHits().getInnerHits().isEmpty();
```
vibrantvarun, line 34: nit: Do we really need to check `!fetchContext.innerHits().getInnerHits().isEmpty()` as well?

vibrantvarun: I am assuming if fetchContext.innerHits is true then it should work?

src/main/java/org/opensearch/neuralsearch/processor/HybridizationFetchSubPhase.java
vibrantvarun: +1

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -334,6 +342,29 @@ private void updateOriginalFetchResults(
...
+            int shardIndex = querySearchResult.getShardIndex();
+            String key = shardIndex + "_" + scoreDoc.doc;
+            float[] subqueryScores = scoreMap != null ? scoreMap.get(key) : null;
+
+            Map<String, DocumentField> documentFields = searchHit.getDocumentFields();
+
+            // Check for all the conditions
+            boolean shouldAddHybridScores = subqueryScores != null
+                && documentFields.containsKey(SUB_QUERY_SCORES_NAME) == false
+                && isClusterOnOrAfterMinReqVersionForSubQuerySupport()
```
vibrantvarun: Shouldn't this whole check should be wrapped under 

`if (isClusterOnOrAfterMinReqVersionForSubQuerySupport() && hasInnerHits ==false )`

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -334,6 +342,29 @@ private void updateOriginalFetchResults(
...
+            boolean hasInnerHits = searchHit.getInnerHits() != null && !searchHit.getInnerHits().isEmpty();
+            int shardIndex = querySearchResult.getShardIndex();
+            String key = shardIndex + "_" + scoreDoc.doc;
+            float[] subqueryScores = scoreMap != null ? scoreMap.get(key) : null;
+
+            Map<String, DocumentField> documentFields = searchHit.getDocumentFields();
+
+            // Check for all the conditions
+            boolean shouldAddHybridScores = subqueryScores != null
+                && documentFields.containsKey(SUB_QUERY_SCORES_NAME) == false
```
vibrantvarun: Could you add a BWC test when out of 3 nodes 1 node is running on newer version and other 2 are running older version. Then request for hybridization scores.

src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
```diff
@@ -64,10 +65,18 @@ public void normalize(final NormalizeScoresDTO normalizeScoresDTO) {
             for (int j = 0; j < topDocsPerSubQuery.size(); j++) {
                 TopDocs subQueryTopDoc = topDocsPerSubQuery.get(j);
                 for (ScoreDoc scoreDoc : subQueryTopDoc.scoreDocs) {
+                    // Initialize or update subquery scores array per doc
+                    if (normalizeScoresDTO.isSubQueryScores()) {
```
vibrantvarun: This is exact same logic and code we are using in L2 and MinMax and Z score. Can we create util package in normalization package and add a util class over there?

src/main/java/org/opensearch/neuralsearch/processor/normalization/ScoreNormalizer.java
```diff
@@ -26,8 +35,22 @@ public class ScoreNormalizer {
     public void normalizeScores(final NormalizeScoresDTO normalizeScoresDTO) {
         final List<CompoundTopDocs> queryTopDocs = normalizeScoresDTO.getQueryTopDocs();
         final ScoreNormalizationTechnique scoreNormalizationTechnique = normalizeScoresDTO.getNormalizationTechnique();
+        final SearchPhaseContext searchPhaseContext = normalizeScoresDTO.getSearchPhaseContext();
         if (canQueryResultsBeNormalized(queryTopDocs)) {
-            scoreNormalizationTechnique.normalize(normalizeScoresDTO);
+
+            Map<String, float[]> hybridizationScores = scoreNormalizationTechnique.normalize(normalizeScoresDTO);
+
+            boolean isExplainEnabled = isExplainEnabled(searchPhaseContext);
```
vibrantvarun, line 43: nit: we need a util class that sets all these values isExplainEnabled, isSortEnabled, isCollapseEnabled etc at once before starting the processing. In the entire normalization processor workflow we calculated isSortEnabled etc at many places. This small checks when run multiple times slows down the search.

For this PR i am okay, I am anyway working on some refractoring. I will take care of it in that.

vibrantvarun (commented):
Great job @owaiskazi19 on creating the logic for capturing subquery scores.

src/main/java/org/opensearch/neuralsearch/processor/HybridScoreRegistry.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+public class HybridScoreRegistry {
+    private static final Map<SearchPhaseContext, Map<String, float[]>> contextToHolder = new HashMap<>();
```
martin-gaievski: improve the behavior of this map in concurrent scenarios, you can make it of type ConcurrentHashMap or similar

src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
martin-gaievski: +1 to this, we need to make this code shared, maybe use utility class as Varun has suggested

src/main/java/org/opensearch/neuralsearch/processor/HybridizationFetchSubPhase.java
```diff
@@ -0,0 +1,67 @@
...
+            public void setNextReader(LeafReaderContext leafReaderContext) throws IOException {
+                this.ctx = leafReaderContext;
+            }
+
+            @Override
+            public void process(HitContext hitContext) {
+                if (isClusterOnOrAfterMinReqVersionForSubQuerySupport() && hasInnerHits == false) {
+                    Map<String, float[]> scoreMap = HybridScoreRegistry.get(context);
+                    if (scoreMap == null) {
+                        return;
```
martin-gaievski, line 54: add debug or info level logs for debugging purposes

src/main/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizationTechnique.java
```diff
@@ -64,10 +65,18 @@ public void normalize(final NormalizeScoresDTO normalizeScoresDTO) {
...
+                        int shardIndex = compoundQueryTopDocs.getSearchShard().getShardId();
+                        String key = shardIndex + "_" + scoreDoc.doc;
```
martin-gaievski:
```suggestion
                        String key = String.format("%d_%d", shardId, docId);
```
can save you few cycles, may be handy as this is on a critical path

src/main/java/org/opensearch/neuralsearch/processor/HybridScoreRegistry.java
```diff
@@ -0,0 +1,43 @@
...
+     */
+    public static Map<String, float[]> get(SearchPhaseContext context) {
+        return contextToHolder.get(context);
+    }
+
+    /**
+     * Remove hybrid score for each search context
+     * @param context
+     */
+    public static void remove(SearchPhaseContext context) {
```
martin-gaievski, line 40: this may be a weak point in terms of memory management. This method may not be called in certain scenarios, if exceptions occur during search processing, the cleanup may never be triggered, leaving entries in the static registry indefinitely.
You have few options here:

1. use try/finally 
```
public void execute(final NormalizationProcessorWorkflowExecuteRequest request) {
    try {
        // ... existing logic
    } finally {
        // cleanup
        HybridScoreRegistry.remove(request.getSearchPhaseContext());
    }
}
```

2. use WeakHashMap for automatic cleanup 
```
private static final Map<SearchPhaseContext, Map<String, float[]>> contextToHolder = 
    Collections.synchronizedMap(new WeakHashMap<>());
```

3. add lifecycle listener
```
public class HybridScoreRegistry implements SearchPhaseContextListener {
    @Override
    public void onSearchPhaseCompleted(SearchPhaseContext context) {
        remove(context);
    }
}
```

I recommend option 1, it can be added to NormalizationProcessorWorkflow class, it's simple and reliable

src/main/java/org/opensearch/neuralsearch/processor/HybridizationFetchSubPhase.java
owaiskazi19: I initially used List<Float> for type safety, but [DocumentField](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/document/DocumentField.java#L73) expects a List<Object>, and Java generics are invariant — so List<Float> can't be directly passed without a cast. To avoid unsafe casting, I kept it as List<Object> while storing only Float values

src/main/java/org/opensearch/neuralsearch/processor/HybridizationFetchSubPhase.java
owaiskazi19: fetchContext.innerHits() returns an InnerHitsContext object — which may be non-null, but contain no inner hits. Then hasInnerHits will be true even if there are no actual inner hits to process

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
owaiskazi19: Isn't it the same thing checking condtion1 && condition2 and then condition3 && condition4 or all together like condtion1 && condition2 && condtion3 && condtion4 ?

src/main/java/org/opensearch/neuralsearch/processor/HybridScoreRegistry.java
owaiskazi19: This is a great suggestion. I can go with the first option in the ScoreNormalizer like the below
```java
public void normalizeScores(final NormalizeScoresDTO normalizeScoresDTO) {
        final List<CompoundTopDocs> queryTopDocs = normalizeScoresDTO.getQueryTopDocs();
        final ScoreNormalizationTechnique scoreNormalizationTechnique = normalizeScoresDTO.getNormalizationTechnique();
        final SearchPhaseContext searchPhaseContext = normalizeScoresDTO.getSearchPhaseContext();
        if (canQueryResultsBeNormalized(queryTopDocs)) {

            Map<String, float[]> hybridizationScores = scoreNormalizationTechnique.normalize(normalizeScoresDTO);

            boolean isExplainEnabled = isExplainEnabled(searchPhaseContext);
            boolean isSortEnabled = isSortEnabled(searchPhaseContext);

            if (isExplainEnabled == false && isSortEnabled == false) {
                try {
                    // Store in registry
                    setSearchPhaseContext(searchPhaseContext);
                    HybridScoreRegistry.store(searchPhaseContext, hybridizationScores);
                } finally {
                    // clean up later via context.addReleasable()
                    searchPhaseContext.addReleasable(() -> HybridScoreRegistry.remove(searchPhaseContext));
                }
            }
        }
    }
```

src/main/java/org/opensearch/neuralsearch/processor/normalization/ScoreNormalizer.java
owaiskazi19: I have already created util method for `isExplainEnabled` and `isSortEnabled`.

owaiskazi19:
https://github.com/opensearch-project/neural-search/issues/1416 is failing most of the times

heemin32:
I can help with retrying only failed tests.

PR #1373, opened 2025-06-05
Fix stats BWC tests by filtering stats based on version

### LINKED ISSUE #1368: [Test Failure]RestNeuralStatsActionIT.testNeuralStats_E2EFlow
```
REPRODUCE WITH: ./gradlew ':qa:restart-upgrade:testAgainstOldCluster' --tests "org.opensearch.neuralsearch.bwc.restart.RestNeuralStatsActionIT.testNeuralStats_E2EFlow" -Dtests.seed=8224F481785A1740 -Dtests.security.manager=false -Dtests.bwc.version=3.0.0 -Dtests.locale=nds -Dtests.timezone=Africa/Khartoum -Druntime.java=23


RestNeuralStatsActionIT > testNeuralStats_E2EFlow FAILED
    java.lang.AssertionError: expected:<1> but was:<5>
        at __randomizedtesting.SeedInfo.seed([8224F481785A1740:AAFA69B09F33323F]:0)
        at org.junit.Assert.assertEquals(Assert.java:120)
        at org.junit.Assert.assertEquals(Assert.java:146)
        at org.opensearch.neuralsearch.bwc.restart.RestNeuralStatsActionIT.testNeuralStats_E2EFlow(RestNeuralStatsActionIT.java:53)
  1> [2025-06-03T18:54:28,657][INFO ][o.o.n.b.r.RestNeuralStatsActionIT] [testNeuralStats_E2EFlow] before test
  1> [2025-06-03T18:54:28,661][INFO ][o.o.n.b.r.RestNeuralStatsActionIT] [testNeuralStats_E2EFlow] initializing REST clients against [http://[::1]:39895, http://127.0.0.1:44931, http://[::1]:33885, http://127.0.0.1:40203, http://[::1]:42561, http://127.0.0.1:46351]
  1> [2025-06-03T18:54:34,365][INFO ][o.o.n.b.r.RestNeuralStatsActionIT] [testNeuralStats_E2EFlow] There are still tasks running after this test that might break subsequent tests [cluster:admin/opensearch/ml/undeploy_model, indices:data/write/bulk, indices:data/write/bulk[s], indices:data/write/bulk[s][p], indices:data/write/bulk[s][r], internal:index/shard/recovery/start_recovery].
  1> [2025-06-03T18:54:34,367][INFO ][o.o.n.b.r.RestNeuralStatsActionIT] [testNeuralStats_E2EFlow] after test

Suite: Test class org.opensearch.neuralsearch.bwc.restart.RestNeuralStatsActionIT
  2> REPRODUCE WITH: ./gradlew ':qa:restart-upgrade:testAgainstOldCluster' --tests "org.opensearch.neuralsearch.bwc.restart.RestNeuralStatsActionIT.testNeuralStats_E2EFlow" -Dtests.seed=8224F481785A1740 -Dtests.security.manager=false -Dtests.bwc.version=3.0.0 -Dtests.locale=nds -Dtests.timezone=Africa/Khartoum -Druntime.java=23
  2> java.lang.AssertionError: expected:<1> but was:<5>
        at __randomizedtesting.SeedInfo.seed([8224F481785A1740:AAFA69B09F33323F]:0)
        at org.junit.Assert.assertEquals(Assert.java:120)
        at org.junit.Assert.assertEquals(Assert.java:146)
        at org.opensearch.neuralsearch.bwc.restart.RestNeuralStatsActionIT.testNeuralStats_E2EFlow(RestNeuralStatsActionIT.java:53)
  2> NOTE: leaving temporary files on disk at: /home/runner/work/neural-search/neural-search/qa/restart-upgrade/build/testrun/testAgainstOldCluster/temp/org.opensearch.neuralsearch.bwc.restart.RestNeuralStatsActionIT_8224F481785A1740-001
  2> NOTE: test params are: codec=Asserting(Lucene101): {}, docValues:{}, maxPointsInLeafNode=688, maxMBSortInHeap=6.8[423](https://github.com/opensearch-project/neural-search/actions/runs/15423132075/job/43403284592?pr=1365#step:4:424)366199232545, sim=Asserting(RandomSimilarity(queryNorm=true): {}), locale=nds, timezone=Africa/Khartoum
  2> NOTE: Linux 6.11.0-1014-azure amd64/Azul Systems, Inc. 23.0.2 (64-bit)/cpus=4,threads=2,free=371892712,total=536870912
  2> NOTE: All tests run in this JVM: [BatchIngestionIT, HybridSearchIT, HybridSearchWithRescoreIT, KnnRadialSearchIT, MultiModalSearchIT, NeuralQueryEnricherProcessorIT, NeuralSparseSearchIT, NeuralSparseTwoPhaseProcessorIT, RestNeuralStatsActionIT]
```

q-andy:
### Description
Fixes BWC tests for neural stats by filtering the input stats to the available version stats

Original issue is due to attempting to read enum ordinals on StreamInput from a higher version node than the receiving node without newer stats. To fix the enum reading requires changes in core: this workaround filters out stats if it detects that the cluster is mid upgrade.

Also fixes failing unit tests in hybrid query and neural query builder.

### Related Issues
Resolves #1368

### Check List


src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
```diff
@@ -195,12 +199,27 @@ private void processStatsRequestParameters(RestRequest request, NeuralStatsInput
...
+            Version minClusterVersion = neuralSearchClusterUtil.getClusterMinVersion();
+            if (minClusterVersion == Version.CURRENT) {
+                if (EVENT_STAT_NAMES.contains(normalizedStat)) {
+                    neuralStatsInput.getEventStatNames().add(EventStatName.from(normalizedStat));
+                } else if (STATE_STAT_NAMES.contains(normalizedStat)) {
+                    neuralStatsInput.getInfoStatNames().add(InfoStatName.from(normalizedStat));
+                } else {
+                    invalidStatNames.add(normalizedStat);
+                }
             } else {
```
heemin32: We could simply use `MinClusterVersionUtil.getInfoStatsAvailableInVersion(minClusterVersion);` even when `CURRENT_VERSION == minClusterVersion`?

q-andy: We could, but `MinClusterVersionUtil.getInfoStatsAvailableInVersion(minClusterVersion);` isn't a constant lookup, it does a set union operation on all the previous version stats. So there would be small overhead.

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
```diff
@@ -195,12 +199,27 @@ private void processStatsRequestParameters(RestRequest request, NeuralStatsInput
...
-                invalidStatNames.add(normalizedStat);
+                // If min cluster version does not match current, we are in rolling upgrade case
+                // If so, we want to only fetch stats that exist on the min version to prevent a serialization error
+                EnumSet<InfoStatName> availableInfoStats = MinClusterVersionUtil.getInfoStatsAvailableInVersion(minClusterVersion);
+                EnumSet<EventStatName> availableEventStats = MinClusterVersionUtil.getEventStatsAvailableInVersion(minClusterVersion);
+
+                if (InfoStatName.isValidName(normalizedStat) && availableInfoStats.contains(InfoStatName.from(normalizedStat))) {
```
heemin32: Shouldn't the `nomalizedStat` be validated beforehand? Why this can be invalid only here?

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
heemin32: Perhaps we could predefine the stat sets for each version to avoid performing union operations.
For example:

2.19.0 → aa_stat
3.0.0 → aa_stat, bb_stat
3.1.0 → aa_stat, bb_stat, cc_stat

This approach might make the code cleaner.

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
q-andy: We validate here since we need to determine whether the string is an EventStatName or InfoStatName, since we don't want to convert it into the wrong Enum. In the previous case we validate with the `EVENT_STAT_NAMES.contains(normalizedStat)` call, which essentially does the same thing. I'll change them both to use `isValid` to be consistent.

heemin32: Why do we need to validate once more here but not when current version is same as min-version?

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
q-andy: I considered that as well, the tradeoff is that then each set uses more memory for the duplicated stats, which will increase every version. My reasoning is the cluster upgrade scenario is the edge case so it's better to have the computation overhead in the edge case rather than bigger memory footprint that's not used the normal case

heemin32: Makes sense. One additional suggestion: consider creating a method that retrieves the set of stats for a given version, incorporating the optimization logic for the latest version as well. This way, the code here can be unified under a single logic path, regardless of version.

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
q-andy: for the additional`availableInfoStats.contains` call, we validate to make sure if user passes in a stat that isn't available in the min cluster version, we exclude it from being added.

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
q-andy: Good idea, will implement. That should resolve the other comment as well by combining the blocks.

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
```diff
@@ -190,17 +194,28 @@ private void processStatsRequestParameters(RestRequest request, NeuralStatsInput
...
-                neuralStatsInput.getInfoStatNames().add(InfoStatName.from(normalizedStat));
-            } else {
+            if (InfoStatName.isValidName(normalizedStat) == false && EventStatName.isValidName(normalizedStat) == false) {
                 invalidStatNames.add(normalizedStat);
+                continue;
+            }
+
+            Version minClusterVersion = neuralSearchClusterUtil.getClusterMinVersion();
+
+            // If min cluster version does not match current, we are in rolling upgrade case
```
heemin32: This comment is not relevant anymore?

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
```diff
@@ -214,8 +229,17 @@ private void processStatsRequestParameters(RestRequest request, NeuralStatsInput
     }
 
     private void addAllStats(NeuralStatsInput neuralStatsInput) {
-        neuralStatsInput.getEventStatNames().addAll(EnumSet.allOf(EventStatName.class));
-        neuralStatsInput.getInfoStatNames().addAll(EnumSet.allOf(InfoStatName.class));
+        Version minClusterVersion = neuralSearchClusterUtil.getClusterMinVersion();
+        if (minClusterVersion == Version.CURRENT) {
```
heemin32: We don't need this case I guess?

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
```diff
@@ -190,17 +194,28 @@ private void processStatsRequestParameters(RestRequest request, NeuralStatsInput
...
                 invalidStatNames.add(normalizedStat);
                 continue;
             }
 
-            if (EVENT_STAT_NAMES.contains(normalizedStat)) {
-                neuralStatsInput.getEventStatNames().add(EventStatName.from(normalizedStat));
-            } else if (STATE_STAT_NAMES.contains(normalizedStat)) {
-                neuralStatsInput.getInfoStatNames().add(InfoStatName.from(normalizedStat));
-            } else {
+            if (InfoStatName.isValidName(normalizedStat) == false && EventStatName.isValidName(normalizedStat) == false) {
```
heemin32: This is already covered by previous condition check I guess?

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
q-andy: I think its still relevant since it explains why we make the getAvailableStats call, I prefer to keep it since it explains why we ignore stats on lower version

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
q-andy: yup, removed

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
q-andy: removed

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
heemin32: Because there is no code checking if min cluster version match with current version or not, it is confusing.

How about "We fetch only the stats available in the minimum cluster version to avoid serialization errors in clusters with mixed node versions."

heemin32 (dismissed):

src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
```diff
@@ -36,6 +39,61 @@ public final class MinClusterVersionUtil {
         .put(QUERY_IMAGE_FIELD.getPreferredName(), MINIMAL_SUPPORTED_VERSION_QUERY_IMAGE_FIX)
         .build();
 
+    /**
+     * Info stats organized by version added
+     */
+    public static final Map<Version, EnumSet<InfoStatName>> infoStatsByVersion = ImmutableMap.<Version, EnumSet<InfoStatName>>builder()
```
will-hwang: will we have to add the new stats here every time we add one?

src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
```diff
@@ -36,6 +39,61 @@ public final class MinClusterVersionUtil {
...
+                InfoStatName.RRF_PROCESSORS,
+                InfoStatName.COMB_TECHNIQUE_RRF_PROCESSORS
+            )
+        )
+        .build();
+
+    /**
+     * Event stats organized by version added
+     */
+    public static final Map<Version, EnumSet<EventStatName>> eventStatsByVersion = ImmutableMap.<Version, EnumSet<EventStatName>>builder()
```
will-hwang: same here. Will we have to add the newly created stats here?

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
```diff
@@ -190,17 +194,23 @@ private void processStatsRequestParameters(RestRequest request, NeuralStatsInput
         for (String stat : stats) {
             // Validate parameter
             String normalizedStat = stat.toLowerCase(Locale.ROOT);
-            if (isValidParamString(normalizedStat) == false) {
+            if (isValidParamString(normalizedStat) == false
+                || (InfoStatName.isValidName(normalizedStat) == false && EventStatName.isValidName(normalizedStat) == false)) {
```
will-hwang: shouldn't this be 
```
isValidParamString(normalizedStat) == false
|| InfoStatName.isValidName(normalizedStat) == false 
|| EventStatName.isValidName(normalizedStat) == false
```

src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
q-andy: yes, the new unit tests should fail as well if the stat isn't added as well

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
will-hwang: nvm, this is correct

will-hwang: it could be clearer to understand if we have a separate method with description
```
// return true if passed in normalizedStats exists in either InfoStats or in EventStats, false otherwise
private boolean isValidStats(String normalizedStats) {
 return InfoStatName.isValidName(normalizedStat)  || EventStatName.isValidName(normalizedStat)
}
```

q-andy: sure, added

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
```diff
@@ -190,17 +194,22 @@ private void processStatsRequestParameters(RestRequest request, NeuralStatsInput
...
-            if (isValidParamString(normalizedStat) == false) {
+            if (isValidParamString(normalizedStat) == false || isValidEventOrInfoStatName(normalizedStat) == false) {
                 invalidStatNames.add(normalizedStat);
                 continue;
             }
 
-            if (EVENT_STAT_NAMES.contains(normalizedStat)) {
-                neuralStatsInput.getEventStatNames().add(EventStatName.from(normalizedStat));
-            } else if (STATE_STAT_NAMES.contains(normalizedStat)) {
+            Version minClusterVersion = neuralSearchClusterUtil.getClusterMinVersion();
```
vibrantvarun: Why are we getting this minClusterVersion here? Could you check the code of NeuralQueryBuilder how we use MinClusterUtil class to validate the version?

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
```diff
@@ -214,8 +223,9 @@ private void processStatsRequestParameters(RestRequest request, NeuralStatsInput
     }
 
     private void addAllStats(NeuralStatsInput neuralStatsInput) {
-        neuralStatsInput.getEventStatNames().addAll(EnumSet.allOf(EventStatName.class));
-        neuralStatsInput.getInfoStatNames().addAll(EnumSet.allOf(InfoStatName.class));
+        Version minClusterVersion = neuralSearchClusterUtil.getClusterMinVersion();
```
vibrantvarun: Same here.

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
```diff
@@ -183,7 +183,7 @@ public List<RestHandler> getRestHandlers(
         IndexNameExpressionResolver indexNameExpressionResolver,
         Supplier<DiscoveryNodes> nodesInCluster
     ) {
-        RestNeuralStatsAction restNeuralStatsAction = new RestNeuralStatsAction(settingsAccessor);
+        RestNeuralStatsAction restNeuralStatsAction = new RestNeuralStatsAction(settingsAccessor, NeuralSearchClusterUtil.instance());
```
vibrantvarun, line 186: I think you do not need to send NeuralSearchClusterUtil here. You can set a version in MinClusterUtil class and use it wherever needed.

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
q-andy: The way we store the min version for stats is slightly different than for other versioned fields. in NeuralQueryBuilder we check MinClusterUtil using a helper that does a map get on the field name. So the field name maps directly to the version. For stats, because we have multiple stats per version, we map the Version to the set of stats available in that version.

I think we can avoid using `getClusterMinVersion()` directly here, changed to refactor it behind MinClusterVersionUtil.

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
q-andy: changed

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
q-andy: good catch

src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
will-hwang: i don't think it's a good experience if we have to manually add the stats names here every time a new stat is added. Can we think of a way to import the added stats in `InfoStatName.java` and in `EventStatName.java` instead of manually adding it here?

q-andy: Agree that it's not ideal, but in order to make the BWC tests work we need to associate each stat with the version that it's added. If we wanted to import automatically we would need to know the stat version somehow in order to import it into the correct set. I'm not aware of anyway to do that automatically: I guess we could add the stat version to the stat name parameter, but that would be more work versus having a centralized set here in my opinion

will-hwang: can we modify the BWC test case itself so it properly checks for stats emission (which is the functionality we want to verify), but not necessarily the number of stats emitted? If stats is emitted correctly for one, i think it's safe to say it's working as expected.

heemin32: > I guess we could add the stat version to the stat name parameter, but that would be more work versus having a centralized set here in my opinion

I think this might be better and easier to maintain?
We can have version in the statNames itself and we can construct this data later.

src/main/java/org/opensearch/neuralsearch/rest/RestNeuralStatsAction.java
heemin32: `getClusterMinVersion` is better than having static method from `MinClusterVersionUtil` so that we can mock it easily in unit test.

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
heemin32: I am not a fan of static method which is not unit test friendly.

src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
heemin32: > can we modify the BWC test case itself so it properly checks for stats emission (which is the functionality we want to verify), but not necessarily the number of stats emitted? If stats is emitted correctly for one, i think it's safe to say it's working as expected.

This is not just for bwc test but to avoid api failure with a cluster with mixed version of nodes.

q-andy: >I think this might be better and easier to maintain?
We can have version in the statNames itself and we can construct this data later.

Refactored into parameter

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
vibrantvarun: This change particularly should be covered in BWC test. So it should be fine. Also, currently across the plugin we already have a way to check MinClusterVersion. My suggestion here is to follow the same. 

Tom if we come up with some other query Util class then folks would start adding BWC version in that class and we would need to send that instance as well in the constructor, which is not the right way. Whole MinClusterVersion functionality should stay at one place. It is good for maintainence as well. Open for suggestion from @heemin32

heemin32: Passing an instance to a constructor isn't wrong—in fact, it's a form of dependency injection, which is a recommended practice in object-oriented programming. On the other hand, relying on static methods is generally something we try to avoid unless it degrade the performance a lot.

heemin32: However, @q-andy your call. For this min cluster version, if there is no need to change min cluster version value for testing, I am fine with static method as well.

vibrantvarun (approved):

PR #1380, opened 2025-06-11
Max normalization technique

vibrantvarun:
This PR is a test PR. No need to merge or review.

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/combination/DedupCombinationTechnique.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/combination/DedupCombinationTechnique.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.neuralsearch.processor.combination;
+
+import lombok.ToString;
+
+@ToString(onlyExplicitlyIncluded = true)
+public class DedupCombinationTechnique implements ScoreCombinationTechnique{
+
+    @ToString.Include
+    public static final String TECHNIQUE_NAME = "dedup";
+
+    @Override
+    public float combine(float[] scores) {
+        if (scores.length<2 || scores.length>2){
+            throw new IllegalArgumentException("scores array length during dedup should be equal to 2");
+        }
+
+        if (scores[0]>scores[1]){
+            return scores[0];
+        }
+        return scores[1];
+    }
+}
--- a/src/main/java/org/opensearch/neuralsearch/processor/combination/DedupCombinationTechnique.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/combination/DedupCombinationTechnique.java
@@ -2,24 +2,25 @@
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
  */
-
 package org.opensearch.neuralsearch.processor.combination;
 
 import lombok.ToString;
+import lombok.extern.log4j.Log4j2;
 
 @ToString(onlyExplicitlyIncluded = true)
-public class DedupCombinationTechnique implements ScoreCombinationTechnique{
+@Log4j2
+public class DedupCombinationTechnique implements ScoreCombinationTechnique {
 
     @ToString.Include
     public static final String TECHNIQUE_NAME = "dedup";
 
     @Override
     public float combine(float[] scores) {
-        if (scores.length<2 || scores.length>2){
+        if (scores.length < 2 || scores.length > 2) {
             throw new IllegalArgumentException("scores array length during dedup should be equal to 2");
         }
-
-        if (scores[0]>scores[1]){
+        log.info("Score 1 and Score 2 are [{}] , [{}]", scores[0], scores[1]);
+        if (scores[0] > scores[1]) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/neuralsearch/processor/combination/DedupCombinationTechnique.java
```diff
@@ -0,0 +1,28 @@
...
+    @Override
+    public float combine(float[] scores) {
+        if (scores.length < 2 || scores.length > 2) {
+            throw new IllegalArgumentException("scores array length during dedup should be equal to 2");
+        }
+        log.info("Score 1 and Score 2 are [{}] , [{}]", scores[0], scores[1]);
+        if (scores[0] > scores[1]) {
+            return scores[0];
+        }
+        return scores[1];
```
martin-gaievski, line 25: you can do Math.max, but I guess this approach is also fine

src/main/java/org/opensearch/neuralsearch/processor/normalization/MaxNormalizationTechnique.java
```diff
@@ -0,0 +1,86 @@
...
+        log.info("max score of knn query [{}]", knnQueryMaxScore);
+
+        float multiplier;
+
+        if ((matchQueryMaxScore == 0.0f || matchQueryMaxScore == Float.MIN_VALUE)
+            || knnQueryMaxScore == 0.0f
+            || knnQueryMaxScore == Float.MIN_VALUE) {
+            multiplier = 1;
+        } else {
+            multiplier = matchQueryMaxScore / knnQueryMaxScore;
```
martin-gaievski: I don't know use case, but here you may loose some precision, make sure the error after doing math with floating point numbers is within acceptable threshold

src/main/java/org/opensearch/neuralsearch/processor/normalization/MaxNormalizationTechnique.java
```diff
@@ -0,0 +1,86 @@
...
+            .size();
+        if (numOfSubqueries < 2 || numOfSubqueries > 2) {
+            throw new IllegalArgumentException("Number of subqueries cannot be grater or lessor than 2");
+        }
+
+        float[] maxScoresPerSubquery = getMaxScores(queryTopDocs, numOfSubqueries);
+
+        float matchQueryMaxScore = maxScoresPerSubquery[0];
+        log.info("max score of match query [{}]", matchQueryMaxScore);
+        float knnQueryMaxScore = maxScoresPerSubquery[1];
```
martin-gaievski: I assume this is the requirement for them to send knn query always at the second position, but here we're relying on user's good intentions

vibrantvarun: So for the POC purposes, I have made it a compulsion that KNN query should be a second subquery for this normlization technique. This we can convey to the user for testing purposes to always provide knn query as 2nd subquery. Here the intention is to make this POC ready for running a few benchmarks.

saratvemulapalli: This is a fair trade-off for POC. Since testing is done by us to evaluate the technique.

src/main/java/org/opensearch/neuralsearch/processor/normalization/MaxNormalizationTechnique.java
```diff
@@ -0,0 +1,86 @@
...
+
+        if ((matchQueryMaxScore == 0.0f || matchQueryMaxScore == Float.MIN_VALUE)
+            || knnQueryMaxScore == 0.0f
+            || knnQueryMaxScore == Float.MIN_VALUE) {
+            multiplier = 1;
+        } else {
+            multiplier = matchQueryMaxScore / knnQueryMaxScore;
+        }
+
+        log.info("Multiplier value is [{}]", multiplier);
```
saratvemulapalli: Lets move all logging to debug since benchmarking would also track the performance (latency).

src/main/java/org/opensearch/neuralsearch/processor/normalization/MaxNormalizationTechnique.java
```diff
@@ -0,0 +1,81 @@
...
+        float matchQueryMaxScore = maxScoresPerSubquery[1];
+
+        float multiplier;
+
+        if ((matchQueryMaxScore == 0.0f || matchQueryMaxScore == Float.MIN_VALUE)
+            || knnQueryMaxScore == 0.0f
+            || knnQueryMaxScore == Float.MIN_VALUE) {
+            multiplier = 1;
+        } else {
+            multiplier = Math.round((matchQueryMaxScore / knnQueryMaxScore) * 1000f) / 1000f;
```
martin-gaievski: with such approach we'll be getting good results for 3 decimal places, just making sure that's fine with requirements.
Also if you want to be extra cautious you can add check for weird values like Infinity or NaN:
```
float ratio = matchQueryMaxScore / knnQueryMaxScore;
if (Float.isFinite(ratio)) {
    multiplier = roundToThreeDecimals(ratio);
} else {
    multiplier = 1.0f; // or another appropriate fallback
}
```

PR #1383, opened 2025-06-12
Fix ingest document wrapper constructor compilation errors in unit tests

q-andy:
### Description
Fixes compilation error from core 3.1

With https://github.com/opensearch-project/OpenSearch/pull/18277 in core, we introduce a new parameter to the constructor of IngestDocumentWrapper. AbstractBatchingProcessor interacts with IngestDocumentWrappers directly so we need to update the constructor usages to fix compilation error in #1382 1382

### Related Issues
n/a

### Check List

### DIFF
--- a/src/test/java/org/opensearch/neuralsearch/processor/InferenceProcessorTestCase.java
+++ b/src/test/java/org/opensearch/neuralsearch/processor/InferenceProcessorTestCase.java
@@ -33,7 +33,7 @@ public class InferenceProcessorTestCase extends OpenSearchTestCase {
             sourceAndMetadata.put("key1", value);
             sourceAndMetadata.put(IndexFieldMapper.NAME, "my_index");
             sourceAndMetadata.put("_id", String.valueOf(i));
-            wrapperList.add(new IngestDocumentWrapper(i, new IngestDocument(sourceAndMetadata, new HashMap<>()), null));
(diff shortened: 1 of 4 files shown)


martin-gaievski:
Looks good, let's wait for CI action to pass.
Btw, @q-andy what does the value 0 mean, is it a order in which objects will be retrieved?

q-andy:
`IngestDocumentWrapper` is a class used by core during bulk operations through bulk API. When users pass in a bulk request it can contain multiple index/update operations in order, the `slot` is used to map pipeline executions/document level processor executions to their corresponding operation in the user's bulk requests so errors will be surfaced correctly.

Update operations also use `IngestDocumentWrapper` and can have up to 2 separate ingest documents associated with a single operation (the regular doc and the upsert doc). So having a single slot isn't enough to disambiguate between different index requests anymore. Before this wasn't an issue because operations <-> ingest documents were one-to-one mapping, but with system ingest pipelines that's no longer the case. With the last PR to core we introduce a new `childSlot` to map multiple ingest documents in an update request back to the original slot in the user's requests instead, so we can surface exceptions correctly.

In this case 0 just means that the document being processed is either an index request, or the `doc` of an update request (so not an upsert). This isn't relevant at a processor level though.

vibrantvarun (approved):

PR #1385, opened 2025-06-12
Fix collapse with pagination bug

ryanbogan:
### Description
Fixes a bug where there were sometimes not enough groups to properly paginate.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollectorFactory.java
+++ b/src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollectorFactory.java
@@ -46,7 +46,7 @@ public class HybridCollectorFactory {
                     collapseContext.getFieldName(),
                     fieldType,
                     sortAndFormats == null ? new Sort(new SortField(null, SortField.Type.SCORE)) : sortAndFormats.sort,
-                    searchContext.size(),
+                    searchContext.size() + searchContext.from(),
                     hitsThresholdChecker
                 );
             } else if (fieldType instanceof NumberFieldMapper.NumberFieldType) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollectorFactory.java
```diff
@@ -46,15 +46,15 @@ public static Collector createCollector(HybridCollectorFactoryDTO hybridCollecto
                     collapseContext.getFieldName(),
                     fieldType,
                     sortAndFormats == null ? new Sort(new SortField(null, SortField.Type.SCORE)) : sortAndFormats.sort,
-                    searchContext.size(),
+                    searchContext.size() + searchContext.from(),
```
owaiskazi19: Should we only add this when pagination is enabled?

ryanbogan: from is set to 0 when pagination is not enabled, so we can always add it

owaiskazi19: Can we add a test for this change?

vibrantvarun: What happens in case of pagination_depth is provided?

martin-gaievski: > Can we add a test for this change?

+1, although this can affect only track with collapse, we need to make sure there is no regression. How the actual bug looks like for the end user?

martin-gaievski: answered offline, there will be a user facing exception

ryanbogan: Ah I should replace this with numHits

vibrantvarun: From my understanding it should be numHits

vibrantvarun (approved):

PR #1389, opened 2025-06-16
Optimize BWC Model Sharing Resources

chloewqg:
### Description
Optimize BWC Model Resources by sharing same model id across tests

### Check List

### DIFF
--- a/qa/restart-upgrade/build.gradle
+++ b/qa/restart-upgrade/build.gradle
@@ -68,11 +68,30 @@ def versionsBelow2_20 = versionsBelow2_19 + "2.19"
 def versionsBelow3_0 = versionsBelow2_20 + "2.20"
 def versionsBelow3_1 = versionsBelow3_0 + "3.0"
 
-// Task to run BWC tests against the old cluster
-task testAgainstOldCluster(type: StandaloneRestIntegTestTask) {
+task testSetup(type: StandaloneRestIntegTestTask) {
     if(!ext.bwcBundleTest){
         dependsOn "zipBwcPlugin"
     }
+    useCluster testClusters."${baseName}"
+    systemProperty 'tests.rest.bwcsuite_cluster', 'old_cluster'
+    systemProperty 'tests.is_old_cluster', 'true'
+    systemProperty 'tests.skip_delete_model_index', 'true'
+    systemProperty 'tests.plugin_bwc_version', ext.neural_search_bwc_version
+
+    filter {
+        excludeTestsMatching "org.opensearch.neuralsearch.bwc.restart.cleanup.*"
+        excludeTestsMatching "org.opensearch.neuralsearch.bwc.restart.test.*"
+    }
+
+    nonInputProperties.systemProperty('tests.rest.cluster', "${-> testClusters."${baseName}".allHttpSocketURI.join(",")}")
+    nonInputProperties.systemProperty('tests.clustername', "${-> testClusters."${baseName}".getName()}")
+    systemProperty 'tests.security.manager', 'false'
+}
+
+// Task to run BWC tests against the old cluster
+task testAgainstOldCluster(type: StandaloneRestIntegTestTask) {
+    dependsOn "testSetup"
+    dependsOn rootProject.tasks.assemble
     useCluster testClusters."${baseName}"
     systemProperty 'tests.rest.bwcsuite_cluster', 'old_cluster'
     systemProperty 'tests.is_old_cluster', 'true'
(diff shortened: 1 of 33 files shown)


qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/cleanup/ModelResourcesCleanUpIT.java
```diff
@@ -0,0 +1,23 @@
...
+package org.opensearch.neuralsearch.bwc.restart.cleanup;
+
+import org.opensearch.neuralsearch.bwc.restart.test.AbstractRestartUpgradeRestTestCase;
+
+import static org.opensearch.neuralsearch.util.TestUtils.NODES_BWC_CLUSTER;
+
+public class ModelResourcesCleanUpIT extends AbstractRestartUpgradeRestTestCase {
+
+    // Shared Model Resources CleanUp
+    public void testCleanUp_E2EFlow() throws Exception {
```
bzhangam, line 14: Why we want to have a dedicated test for resource clean up?

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/setup/ModelResourcesSetUpIT.java
```diff
@@ -0,0 +1,21 @@
...
+package org.opensearch.neuralsearch.bwc.restart.setup;
+
+import org.opensearch.neuralsearch.bwc.restart.test.AbstractRestartUpgradeRestTestCase;
+
+import static org.opensearch.neuralsearch.util.TestUtils.NODES_BWC_CLUSTER;
+
+public class ModelResourcesSetUpIT extends AbstractRestartUpgradeRestTestCase {
+
+    // Shared Model Resources SetUp
+    public void testSetUp_E2EFlow() throws Exception {
```
bzhangam, line 14: Same here why we need this test?

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/test/BatchIngestionIT.java
```diff
@@ -2,7 +2,7 @@
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
  */
-package org.opensearch.neuralsearch.bwc.restart;
+package org.opensearch.neuralsearch.bwc.restart.test;
```
bzhangam, line 5: I think we are also using models in this IT why don't we refactor it to share the model?

qa/restart-upgrade/build.gradle
```diff
@@ -79,61 +78,86 @@ task testAgainstOldCluster(type: StandaloneRestIntegTestTask) {
     systemProperty 'tests.skip_delete_model_index', 'true'
     systemProperty 'tests.plugin_bwc_version', ext.neural_search_bwc_version
 
+    filter {
+        excludeTestsMatching "org.opensearch.neuralsearch.bwc.restart.cleanup.*"
+        excludeTestsMatching "org.opensearch.neuralsearch.bwc.restart.test.*"
+    }
+
+    nonInputProperties.systemProperty('tests.rest.cluster', "${-> testClusters."${baseName}".allHttpSocketURI.join(",")}")
```
will-hwang, line 86: what do these lines do?

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/cleanup/ModelResourcesCleanUpIT.java
will-hwang: +1. it should be defined in build.gradle, not as a separate integ test

qa/restart-upgrade/build.gradle
```diff
@@ -79,61 +78,86 @@ task testAgainstOldCluster(type: StandaloneRestIntegTestTask) {
...
+
     // Excluding MultiModalSearchIT, HybridSearchIT, NeuralSparseSearchIT, NeuralQueryEnricherProcessorIT tests from neural search version 2.9 and 2.10
     // because these features were released in 2.11 version.
     if (versionsBelow2_11.any { ext.neural_search_bwc_version.startsWith(it) }){
         filter {
-            excludeTestsMatching "org.opensearch.neuralsearch.bwc.restart.MultiModalSearchIT.*"
-            excludeTestsMatching "org.opensearch.neuralsearch.bwc.restart.HybridSearchIT.*"
-            excludeTestsMatching "org.opensearch.neuralsearch.bwc.restart.NeuralSparseSearchIT.*"
-            excludeTestsMatching "org.opensearch.neuralsearch.bwc.restart.NeuralQueryEnricherProcessorIT.*"
+            excludeTestsMatching "org.opensearch.neuralsearch.bwc.restart.test.MultiModalSearchIT.*"
```
will-hwang, line 105: we shouldn't need separate test cases for clean/setup, and we should be able to get rid of the `/test` subfolder.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/test/BatchIngestionIT.java
chloewqg: This PR is working step by step. We need to first create stage that sets up the model first correctly and then refactor all the BWC tests here.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/setup/ModelResourcesSetUpIT.java
chloewqg: To create the model resources in one test and later be used by other models.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/cleanup/ModelResourcesCleanUpIT.java
chloewqg: So far I think model clean up in defined in each test by wipeOfTestResources. How do we define this build.gradle?

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/setup/ModelResourcesSetUpIT.java
heemin32: I think we should dynamically create model resource during test itself.

qa/restart-upgrade/build.gradle
chloewqg: Just go through the model resource set up stage and exclude other bwc tests. Line 86 is just cluster config setup

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/cleanup/ModelResourcesCleanUpIT.java
heemin32: Could you share an example of how we could clean up i build.gradle itself without integ test?

qa/restart-upgrade/build.gradle
chloewqg: Well, if we don't have a separate test cases to create dedicated and shared model resources. Would different bwc tests race against each other and create shared model resources and potentially later one would override earlier one?

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/setup/ModelResourcesSetUpIT.java
chloewqg: Dynamically create model resources mean each model will try to check if models have been creating by other tests and reuse it if the model has been created?

chloewqg: My concern is that this will create race condition, that say one model is creating model A in progress and not return yet. At the same time another test coming in and check that there is no model created yet and then is going to create another version of model A.

chloewqg: If we want to do it dynamically, we should use some lock mechanism to avoid modifying the same modelId parameter by two different bwc tests?

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/cleanup/ModelResourcesCleanUpIT.java
will-hwang: we should be able to utilize `doFirst` and `doLast` clauses and define setUp and cleanUp classes respectively. We can use the same classes for both rolling and restart

To run java class in build.gradle, we can probably do something like this? https://www.baeldung.com/gradle-run-java-main#run-with-application

qa/restart-upgrade/build.gradle
will-hwang: if possible, we should try to create resources as part of `build.gradle` in the `/qa` folder. That way we can ensure only one set of resources are created for both restart and rolling

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/setup/ModelResourcesSetUpIT.java
heemin32: We use model group with its name. Are we allowed to create more than two model group with same name? I think race condition should be handled by ml-common side as long as we use same identification?

chloewqg: I think the race condition here refers to creation of models with different model ids under same model group? Right now, each class can create a dedicated model id. And if we keep a static variable of model id that can be modified by every class, then this will cause race condition if I understand correctly.

heemin32: I believe, we can utilize lock so that only single model is created.

Issue #1392, opened 2025-06-18
[BUG] Unable to perform Neural query with min_score/max_distance on Opensearch 3.0


swamibalaji18:
### What is the bug?
I am currently using Opensearch 2.19 on GKE. I am planning to migrate to 3.0 as it offers great features. Currently i am using neural query with min_score on 2.19 and it works just fine. But same query is not working on 3.0. only k is working. min_score/max_distance are NOT working. I get "requires exactly one of k, distance or score to be set" error when i run query. 

### How can one reproduce the bug?
1. Install Opensearch 3.0.
2. Create test index:
`PUT /test-products
{
  "settings": {
    "index": {
      "knn": "true"
    }
  }, 
  "mappings": {
    "properties": {
      "description_dense_vector": {
        "type": "knn_vector",
        "dimension": 768,
        "method": {
          "engine": "faiss",
          "space_type": "cosinesimil",
          "name": "hnsw",
          "parameters": {
            "ef_search": 170,
            "ef_construction": 256,
            "m": 32
          }
        }
      },
      "product_id": { "type": "keyword" },
      "multifield_text": { "type": "text" }
    }
  }
}`

3. Add sample document:
`POST /test-products/_doc
{
  "description_dense_vector": [
            -0.24329314,
            0.80990225,
            -0.96670306,
            -0.003956463,
            0.33589908,
            0.19736764,
            -1.1747291,
            1.3404093,
            -0.42092466,
            -0.1316185,
            -0.62286276,
            1.9675168,
            0.3356531,
            0.5120529,
            -0.72481835,
            -0.240796,
            0.60869265,
            0.9932429,
            0.70794475,
            -0.1680499,
            0.037890457,
            -0.42113483,
            0.6781846,
            -0.53842556,
            -0.4520312,
            -0.123514496,
            -0.7172234,
            0.13718332,
            0.0016019717,
            0.93460137,
            0.85099447,
            0.24077156,
            -0.09392083,
            0.46281672,
            1.2512901,
            0.7672916,
            -0.41193196,
            -0.34979174,
            -0.18154137,
            -0.04843665,
            0.4791609,
            0.82509893,
            0.5490749,
            0.7886476,
            -1.2279992,
            -0.42511144,
            0.054081913,
            -0.49302652,
            0.7966083,
            -0.82614666,
            -0.51215583,
            -0.7470685,
            -0.84192395,
            -0.19130668,
            -0.78175324,
            -0.4794649,
            0.77635145,
[...]

PR #1320, opened 2025-05-12
Support multi node integ tests

owaiskazi19:
### Description
Support multi node integ tests.
Improvements done for CI in this PR:
1. [example 1](https://github.com/opensearch-project/neural-search/actions/runs/15864438668/job/44728552726#step:6:231), [example 2](https://github.com/opensearch-project/neural-search/actions/runs/15864438668/job/44728547797#step:4:106) integ test was running twice in precommit and check. Removed running integTest again in precommit
2. jacocotestReport tasks depends on test and integTest. Since gradle check already runs both, moved codeCov CI under gradle check.
3. Added multi node integration tests CI

### Related Issues
Resolves #https://github.com/opensearch-project/neural-search/issues/1307

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [ ] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/neural-search/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1393, opened 2025-06-19
Fix neural radial search serialization in multi-node clusters

junqiu-lei:
### Description
Fix neural radial search queries failing with "[knn] requires exactly one of k, distance or score to be set" error in multi-node clusters by k-NN plugin handles its own serialization/deserialization, avoiding version checking conflicts between the neural-search and k-NN plugins.

Current integ test CI won't be able cover this multi node env (PR for Multi node test CI https://github.com/opensearch-project/neural-search/pull/1320 in progress), manually do the integ test at multi node env: https://gist.github.com/junqiu-lei/22df5f0420d3e558f00e3390119a27ed

### Related Issues
Resolves https://github.com/opensearch-project/neural-search/issues/1392

### Check List


src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
```diff
@@ -274,17 +305,18 @@ public Builder originalQueryText(String originalQueryText) {
          * @return A new NeuralKNNQueryBuilder instance
          */
         public NeuralKNNQueryBuilder build() {
-            KNNQueryBuilder knnBuilder = KNNQueryBuilder.builder()
+            KNNQueryBuilder.Builder knnBuilderInstance = KNNQueryBuilder.builder()
```
heemin32: Why do we need this change? I don't think we need to make any change here.

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
```diff
@@ -308,7 +340,7 @@ private NeuralKNNQueryBuilder(KNNQueryBuilder knnQueryBuilder, String originalQu
      */
     @Override
     public void doWriteTo(StreamOutput out) throws IOException {
-        KNNQueryBuilderParser.streamOutput(out, knnQueryBuilder, IndexUtil::isClusterOnOrAfterMinRequiredVersion);
+        KNNQueryBuilderParser.streamOutput(out, knnQueryBuilder, MinClusterVersionUtil::isClusterOnOrAfterMinReqVersion);
```
heemin32: Should't we use knn util not neural util?

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
```diff
@@ -274,17 +305,18 @@ public Builder originalQueryText(String originalQueryText) {
          * @return A new NeuralKNNQueryBuilder instance
          */
         public NeuralKNNQueryBuilder build() {
-            KNNQueryBuilder knnBuilder = KNNQueryBuilder.builder()
+            KNNQueryBuilder.Builder knnBuilderInstance = KNNQueryBuilder.builder()
                 .fieldName(fieldName)
                 .vector(vector)
-                .k(k)
```
heemin32: If we have to make a change here, why not simply set `.k(k == 0 ? null : k)`

Also, do we still validate and throw an exception if user explicitly pass 0 for k?

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
```diff
@@ -323,8 +355,26 @@ public void doWriteTo(StreamOutput out) throws IOException {
...
     public NeuralKNNQueryBuilder(StreamInput in) throws IOException {
         super(in);
-        KNNQueryBuilder.Builder builder = KNNQueryBuilderParser.streamInput(in, IndexUtil::isClusterOnOrAfterMinRequiredVersion);
-        this.knnQueryBuilder = builder.build();
+        KNNQueryBuilder.Builder builder = KNNQueryBuilderParser.streamInput(in, MinClusterVersionUtil::isClusterOnOrAfterMinReqVersion);
+
+        // Radial search queries (with min_score/max_distance) have null k, which gets serialized as 0.
+        // We need to rebuild the query to properly handle this case.
+        // Performance impact is minimal as this only affects cross-node queries.
+        KNNQueryBuilder tempKnn = builder.build();
```
heemin32: The error will be thrown here. During build, it will validate the parameter and k is zero but not null already. 
We passed `KNNQueryBuilder` in `doWriteTo()` but reading as `KNNQueryBuilder.Builder` here. We should read it as `KNNQueryBuilder` directly.

heemin32:
We might need bwc rolling test later.

Also, wouldn't it be better to revert
https://github.com/opensearch-project/neural-search/pull/1183
https://github.com/opensearch-project/neural-search/pull/1272
And just add additional originalQueryText field?


I am not worrying about depending on knn internal method where we don't have a control. Also, do we have rolling upgrade test between 2.19 and 3.0 for this scenario? Not sure the serialization and deserialization is compatible between 2.19 and 3.0 as we changed from our custom serialization to knn serialization.

release-notes/opensearch-neural-search.release-notes-3.1.0.0.md
```diff
@@ -31,7 +31,8 @@ Compatible with OpenSearch 3.1.0
 - Return bad request for stats API calls with invalid stat names instead of ignoring them ([#1291](https://github.com/opensearch-project/neural-search/pull/1291))
 - Add validation for invalid nested hybrid query ([#1305](https://github.com/opensearch-project/neural-search/pull/1305))
 - Use stack to collect semantic fields to avoid stack overflow ([#1357](https://github.com/opensearch-project/neural-search/pull/1357))
-- Filter requested stats based on minimum cluster version to fix BWC tests for stats API ([#1373](https://github.com/opensearch-project/neural-search/pull/1373))
+- Filter requested stats based on minimum cluster version to fix BWC tests for stats API ([#1373](https://github.com/opensearch-project/neural-search/pull/1373)
+- Fix neural radial search serialization in multi-node clusters ([#1393](https://github.com/opensearch-project/neural-search/pull/1393)))
```
martin-gaievski, line 35: you probably need same entry in changelog

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
martin-gaievski: +1 to the latest quest, although it's a edge case, we probably threat such input as radial search, and may fail later

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
martin-gaievski: I think it's better rely on neural utils here, otherwise we have to keep an eye on changes in knn

heemin32: KNNQueryBuilderParser is already implemented inside knn. We don't know exactly which key they are using to check the supported version. As long as we don't pass any value in knn query for new parameter, there won't be any issue and therefore, it is safer to use knn util.

heemin32: Or, we should have our own serialization logic instead of depending on knn's logic.

release-notes/opensearch-neural-search.release-notes-3.1.0.0.md
heemin32: why do we want it to be in changelog if we add it directly to release note?

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
junqiu-lei: @heemin32 You're asking about the split between creating the Builder instance and calling build(). This change is necessary because:

  1. **Original code** would have chained everything including `.k(k)` directly:
     ```java
     KNNQueryBuilder knnBuilder = KNNQueryBuilder.builder()
         .fieldName(fieldName)
         .vector(vector)
         .k(k)  // This fails when k is null
         .build();

  2. Current code separates the builder creation from the build() call:
  KNNQueryBuilder.Builder knnBuilderInstance = KNNQueryBuilder.builder()
      .fieldName(fieldName)
      .vector(vector);
      // ... other fields except k, maxDistance, minScore

  // Conditionally set k only when it's not null
  applySearchParameters(knnBuilderInstance, k, maxDistance, minScore);

  KNNQueryBuilder knnBuilder = knnBuilderInstance.build();

  This separation is required because:
  - For radial search, k can be null (when using min_score or max_distance)
  - The .k(null) call throws a NullPointerException
  - We need to conditionally set these parameters based on whether they're null

Without this change, radial search queries (where k is null) would fail with NPE. This has been part of NeuralKNNQueryBuilder since it was introduced to support radial search.

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
junqiu-lei: updated to use knn util as suggested.

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
junqiu-lei: The issue is that the KNN builder validates k must be > 0 when set. If we try to set k=0, it throws a validation error. That's why we need the `applySearchParameters` helper method that conditionally sets k only when it's valid (> 0).

Users cannot explicitly pass k=0 - the validation in KNNQueryBuilder prevents this. The k=0 only appears during deserialization because: User creates a radial search query (k=null, with min_score/max_distance), during serialization, null k gets written as 0 to the stream, during deserialization, we read 0 and need to handle it specially.

src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
junqiu-lei: Right, the validation error would occur at `builder.build()`. However, if you look at the KNNQueryBuilderParser.streamInput implementation, it handles this case specially:

  ```
  // From KNNQueryBuilderParser.streamInput:
  builder.k(in.readInt()); // This reads the k value (0 for radial search)
```
The key insight is that the KNN parser's streamInput returns a Builder, not a built object. When k=0 is read from the stream, it's set on the builder but validation only happens when we call build().

That's exactly why we need to intercept and rebuild:
  1. We let the parser read all fields into the builder (including k=0)
  2. We build temporarily to extract all the values
  3. We rebuild with our applySearchParameters method that skips setting k when it's 0

Regarding reading as KNNQueryBuilder vs Builder - we are reading as Builder (that's what KNNQueryBuilderParser.streamInput returns). The issue is that the builder contains k=0 which will fail validation when built, hence our workaround.

This approach ensures radial search queries work correctly across nodes without changing the KNN plugin's serialization format.

File: release-notes/opensearch-neural-search.release-notes-3.1.0.0.md (hunk 5 above)
Comment (junqiu-lei) [reply release-notes/opensearch-neural-search.release-notes-3.1.0.0.md:35]: We might better just keep this change in 3.1.0 release note directly at this point?

Comment (martin-gaievski) [reply release-notes/opensearch-neural-search.release-notes-3.1.0.0.md:35]: for 3.1 branch is fine, for main we need it in changelog for 3.2 future release notes

Comment (junqiu-lei) [reply release-notes/opensearch-neural-search.release-notes-3.1.0.0.md:35]: we're target to merge it along with version 3.1, I don't think we need it in scope of 3.2 which is proceed from CHANGELOG.

Comment (martin-gaievski) [reply release-notes/opensearch-neural-search.release-notes-3.1.0.0.md:35]: got it, basically this content can be backported to 3.1 as is, with changelog it will need manual correction

**Review by martin-gaievski — APPROVED** [review]
Looks good, thank you @junqiu-lei

**Comment by junqiu-lei** [comment]
Pending merge with verifying latest changes in multi node env.

**Comment by heemin32** [comment]
We need to change here as well https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java#L838 so that if the version is 2.19, we should not return NeuralKnnQueryBuilder but we should return NeuralQueryBuilder. 
https://github.com/opensearch-project/neural-search/pull/1183/files

**Comment by junqiu-lei** [comment]
> We need to change here as well https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java#L838 so that if the version is 2.19, we should not return NeuralKnnQueryBuilder but we should return NeuralQueryBuilder. https://github.com/opensearch-project/neural-search/pull/1183/files
> 
> Also, we might need to update our current bwc rolling upgrade test to catch this issue. Please create a github issue and follow up if we cannot add it in this PR.

@heemin32 Added backward compatibility and unit test for createKNNQueryBuilder, added BWC rolling test for NeuralKNNQuery.

File: src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java [hunk 6]
```diff
@@ -835,19 +835,34 @@ private QueryBuilder rewriteQueryAgainstKnnField(QueryRewriteContext queryRewrit
...
+            NeuralKNNQueryBuilder.Builder builder = NeuralKNNQueryBuilder.builder()
+                .fieldName(fieldName)
+                .vector(vector)
+                .filter(filter())
+                .expandNested(expandNested())
+                .methodParameters(methodParameters())
+                .rescoreContext(rescoreContext())
+                .originalQueryText(queryText());
+
+            // Only set k if it's not a radial search (i.e., when minScore and maxDistance are null)
```
heemin32: Why do we need this logic here? I don't think this is necessary?

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -835,19 +835,34 @@ private QueryBuilder rewriteQueryAgainstKnnField(QueryRewriteContext queryRewrit
...
+
+            // Set radial search parameters if present
+            builder.maxDistance(maxDistance());
+            builder.minScore(minScore());
+
+            return builder.build();
+        } else {
+            // For versions before 3.0.0 (like 2.19.0), return NeuralQueryBuilder
+            // to maintain backward compatibility during rolling upgrades
+            return createNeuralQueryBuilder(KNNVectorFieldMapper.CONTENT_TYPE, () -> vector, false);
```
heemin32: We should not return NeuralQueryBuilder for 2.19.
It should return `KNNQueryBuilder`
```
return KNNQueryBuilder.builder()
                .fieldName(fieldName())
                .vector(vectorSupplier.get())
                .filter(filter())
                .maxDistance(maxDistance)
                .minScore(minScore)
                .expandNested(expandNested)
                .k(k)
                .methodParameters(methodParameters)
                .rescoreContext(rescoreContext)
                .build();
```

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/NeuralKNNQueryBWCIT.java
```diff
@@ -0,0 +1,267 @@
+/*
```
heemin32: Have we confirmed that this is failing without the change?

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
junqiu-lei: removed

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
junqiu-lei: updated

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/NeuralKNNQueryBWCIT.java
junqiu-lei: @heemin32 We actually no need this new bwc test, removed it. Instead we can re-use KnnRadialSearchIT bwc rolling test with add extra shards response validation.

And with: 
```
QueryBuilder createKNNQueryBuilder(String fieldName, float[] vector) {
    NeuralKNNQueryBuilder.Builder builder = NeuralKNNQueryBuilder.builder()
        .fieldName(fieldName)
        .vector(vector)
        .filter(filter())
        .expandNested(expandNested())
        .methodParameters(methodParameters())
        .rescoreContext(rescoreContext())
        .originalQueryText(queryText())
        .k(k())
        .maxDistance(maxDistance())
        .minScore(minScore());

    return builder.build();
}
```

we can hit the expected error:
```
> Task :qa:rolling-upgrade:testAgainstOneThirdUpgradedCluster

REPRODUCE WITH: ./gradlew ':qa:rolling-upgrade:testAgainstOneThirdUpgradedCluster' --tests 'org.opensearch.neuralsearch.bwc.rolling.KnnRadialSearchIT.testKnnRadialSearch_E2EFlow' -Dtests.seed=8DD4F3566687C760 -Dtests.security.manager=false -Dtests.bwc.version=2.19.0 -Dtests.locale=en-NG -Dtests.timezone=IET -Druntime.java=21

KnnRadialSearchIT > testKnnRadialSearch_E2EFlow FAILED
    java.lang.AssertionError: All shards should be successful for radial search with minScore. Total: 3, Successful: 2, Failed: 1 expected:<3> but was:<2>
        at __randomizedtesting.SeedInfo.seed([8DD4F3566687C760:5AB99F6AC828E751]:0)
        at org.junit.Assert.assertEquals(Assert.java:647)
        at org.opensearch.neuralsearch.bwc.rolling.KnnRadialSearchIT.validateSearchResponse(KnnRadialSearchIT.java:132)
        at org.opensearch.neuralsearch.bwc.rolling.KnnRadialSearchIT.validateIndexQueryOnUpgrade(KnnRadialSearchIT.java:103)
        at org.opensearch.neuralsearch.bwc.rolling.KnnRadialSearchIT.testKnnRadialSearch_E2EFlow(KnnRadialSearchIT.java:52)

```

qa/rolling-upgrade/build.gradle
```diff
@@ -133,6 +141,13 @@ task testAgainstOneThirdUpgradedCluster(type: StandaloneRestIntegTestTask) {
             excludeTestsMatching "org.opensearch.neuralsearch.bwc.rolling.SemanticFieldIT.*"
         }
     }
+
+    // Excluding KnnRadialSearchIT for 3.0.0 due to known radial search shard failure issue
```
heemin32, line 145:
```suggestion
    // Excluding KnnRadialSearchIT for 3.0.0 due to known radial search shard failure issue
    // https://github.com/opensearch-project/neural-search/issues/1392
```

qa/rolling-upgrade/build.gradle
```diff
@@ -171,6 +186,13 @@ task testAgainstTwoThirdsUpgradedCluster(type: StandaloneRestIntegTestTask) {
...
```
heemin32, line 191:
```suggestion
    // Excluding KnnRadialSearchIT for 3.0.0 due to known radial search shard failure issue
    // https://github.com/opensearch-project/neural-search/issues/1392
```

qa/rolling-upgrade/build.gradle
```diff
@@ -209,4 +231,11 @@ task testRollingUpgrade(type: StandaloneRestIntegTestTask) {
...
```
heemin32, line 237:
```suggestion
    // Excluding KnnRadialSearchIT for 3.0.0 due to known radial search shard failure issue
    // https://github.com/opensearch-project/neural-search/issues/1392
```

heemin32:
Let's backport this PR in 3.0 as well?

junqiu-lei:
> Let's backport this PR in 3.0 as well?

sure

PR #1410, opened 2025-06-25
Use shared model across bwc tests

heemin32:
### Description
Use shared model across bwc tests

### Related Issues
Resolves https://github.com/opensearch-project/neural-search/issues/1035

### Check List

### DIFF
--- a/qa/README.md
+++ b/qa/README.md
@@ -1,0 +1,17 @@
+# Neural Search QA
+
+## Model Deployment Guidelines
+
+### Important Notes for Model Handling
+
+1. **Non-ML Node Deployment**
+   - To deploy models on non-ML nodes, set `plugins.ml_commons.only_run_on_ml_node` to `false`
+   - **Warning**: Resetting this setting will undeploy all currently deployed models
+
+2. **Avoid Duplicate Deployments**
+   - Do not attempt to deploy already deployed models
+   - This can cause tasks to get stuck in `RUNNING` state (possible ml-commons bug)
+
+3. **Node Replacement Timing**
+   - Avoid deploying models immediately after node replacement
+   - This can interfere with the auto-deploy feature (possible ml-commons bug)
--- a/src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
+++ b/src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
@@ -4,40 +4,14 @@
  */
 package org.opensearch.neuralsearch;
 
+import com.carrotsearch.randomizedtesting.RandomizedTest;
+import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
+import com.google.common.collect.ImmutableList;
 import joptsimple.internal.Strings;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
 import lombok.NonNull;
-import org.junit.After;
-import org.opensearch.action.search.SearchResponse;
-import org.opensearch.client.ResponseException;
-import org.opensearch.cluster.metadata.IndexNameExpressionResolver;
-import org.opensearch.common.util.concurrent.ThreadContext;
-import org.opensearch.core.xcontent.DeprecationHandler;
-import org.opensearch.core.xcontent.MediaType;
-import org.opensearch.core.xcontent.MediaTypeRegistry;
-import org.opensearch.core.xcontent.NamedXContentRegistry;
-import org.opensearch.core.xcontent.XContentParser;
-import org.opensearch.ml.common.model.MLModelState;
-
-import static org.opensearch.knn.common.KNNConstants.MODEL_INDEX_NAME;
-import static org.opensearch.neuralsearch.common.VectorUtil.vectorAsListToArray;
-
-import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Set;
-import java.util.ArrayList;
-import java.util.function.Predicate;
-import java.util.function.Supplier;
-import java.util.stream.Collectors;
-
+import lombok.SneakyThrows;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hc.core5.http.ContentType;
 import org.apache.hc.core5.http.Header;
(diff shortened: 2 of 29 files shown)


src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -1773,6 +1780,31 @@ protected boolean isModelAlreadyDeployed(@NonNull final String modelId) throws I
         }
     }
 
+    protected boolean isModelReadyForInference(@NonNull final String modelId) throws IOException, ParseException {
+        Response getModelResponse = getModel(modelId);
+        if (RestStatus.NOT_FOUND.equals(RestStatus.fromCode(getModelResponse.getStatusLine().getStatusCode()))) {
+            return false;
+        } else {
+            MLModelState state = getModelState(getModelResponse);
+            return MLModelState.LOADED.equals(state) || MLModelState.DEPLOYED.equals(state);
```
will-hwang: should it just be `MLModelState.DEPLOYED.equals(state);`

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -2724,4 +2790,81 @@ protected void ingestBatchDocumentWithBulk(
...
+                logger.info("Model is already deployed. Skip loading.");
+                return;
+            }
+        }
+        loadModel(modelId);
+        waitForModelToBeReady(modelId);
+    }
+
+    static private synchronized void loadAndWaitForModelToBeReady(BaseNeuralSearchIT baseNeuralSearchIT, String modelId) throws Exception {
+        if (LOADED_MODE_IDS.contains(modelId)) {
```
will-hwang: `LOADED_MODEL_IDS`

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -2724,4 +2790,81 @@ protected void ingestBatchDocumentWithBulk(
         }
         assertEquals(failedIds.size(), failedDocCount);
     }
+
+    // ============== SHARED MODEL MANAGEMENT METHODS ==============
+    // These methods provide shared model functionality for upgrade tests
+
+    static private synchronized String getOrCreateSharedTextEmbeddingModel(BaseNeuralSearchIT baseNeuralSearchIT) throws Exception {
```
will-hwang: why do we need to pass in `BaseNeuralSearchIT baseNeuralSearchIT`, when we should just be able to call `doLoadAndWaitForModelToBeReady` directly?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -2273,6 +2306,43 @@ private void deleteExistingModels() {
         });
     }
 
+    /**
+     * Check if a model is being used by any ingest or search pipeline
+     * @param modelId the model ID to check
+     * @return true if the model is used by any pipeline, false otherwise
+     */
+    @SneakyThrows
+    protected boolean isModelUsedByAnyPipeline(final String modelId) {
```
will-hwang: unless we explicitly remove model ID from the pipeline, this will always return true once pipeline is created?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
heemin32: Older version of ml-common might use LOADED state?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
heemin32: yes

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
heemin32: doLoadAndWaitForModelToBeReady is non static method and this is static method.

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
heemin32: ack

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
will-hwang: how will we ever pass this condition then? https://github.com/opensearch-project/neural-search/pull/1410/files#diff-0eb119317f557769732b15b76927b9f79dbeb5c1e262cdfd515b35477ffbd1bcR114

will-hwang: nvm, we delete the pipelines beforehand

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
will-hwang: in that case, should we apply the same check here? [link](https://github.com/opensearch-project/neural-search/pull/1410/files#diff-0eb119317f557769732b15b76927b9f79dbeb5c1e262cdfd515b35477ffbd1bcR182)

src/testFixtures/java/org/opensearch/neuralsearch/BaseUpgradeTestCase.java
```diff
@@ -0,0 +1,215 @@
...
+    ) {
+        if (ingestPipeline != null) {
+            deleteIngestPipeline(ingestPipeline);
+        }
+        if (searchPipeline != null) {
+            deleteSearchPipeline(searchPipeline);
+        }
+        if (modelId != null) {
+            if (!isModelUsedByAnyPipeline(modelId)) {
+                deleteModel(modelId);
```
will-hwang: we should also remove the model id from `DEPLOYED_MODEL_IDS`

src/testFixtures/java/org/opensearch/neuralsearch/BaseUpgradeTestCase.java
```diff
@@ -0,0 +1,227 @@
...
+            if (ingestPipeline != null) {
+                deleteIngestPipeline(ingestPipeline);
+            }
+            if (searchPipeline != null) {
+                deleteSearchPipeline(searchPipeline);
+            }
+            if (modelId != null) {
+                if (!isModelUsedByAnyPipeline(modelId)) {
+                    deleteModel(modelId);
+                    DEPLOYED_MODEL_IDS.remove(modelId);
```
martin-gaievski, line 117: I think we need to remove model id unconditionally, meaning outside of if statement

will-hwang (approved):
LGTM

qa/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Model Deployment Guidelines
+
+### Important Notes for Model Handling
+
+1. **Non-ML Node Deployment**
+   - To deploy models on non-ML nodes, set `plugins.ml_commons.only_run_on_ml_node` to `false`
+   - **Warning**: Resetting this setting will undeploy all currently deployed models
+
+2. **Avoid Duplicate Deployments**
```
junqiu-lei, line 11: Discussion: Instead of adding note doc here, are we able to only allow the sharing model being deployed and used in code? So that new tests added by developers anyway will be limited to use the shared models rather than potential new model resources.

heemin32: They will most likely use existing methods, uploadXXXModel, defined in `BaseUpgradeTestCase` to load model which will use shared model. There is no simple way to block them to use their own model though.

src/testFixtures/java/org/opensearch/neuralsearch/BaseUpgradeTestCase.java
heemin32: I think model id should be removed only when they are deleted?

junqiu-lei (approved):

PR #1420, opened 2025-07-02
Support configuring the auto-generated knn_vector field through the semantic field.

### LINKED ISSUE #1356: [FEATURE] Semantic Field Enhancement - Allow to Configure knn_vector field
Currently the semantic field with a dense model will auto generate the knn_vector field to store the embedding. knn_vector field allow us to configure how to store the vector and how to do search but for now we will only use the default configuration to create the field in the index mapping. And we will pull the space type and embedding dimension from the model metadata.

We think we should give users more flexibility to configure the knn field.
Option 1:
We should allow users to configure the auto generated knn_vector field through the semantic field.
Check this [link](https://docs.opensearch.org/docs/latest/field-types/supported-field-types/knn-vector/) to see the configuration the knn_vector field can support.

N/A

bzhangam:
### Description
Support configuring the auto-generated knn_vector field through the semantic field.

### Related Issues
Resolves #1356 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
@@ -17,15 +17,18 @@ import org.opensearch.ml.common.model.TextEmbeddingModelConfig;
 import reactor.util.annotation.NonNull;
 
 import java.util.HashMap;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
 import static org.opensearch.neuralsearch.constants.MappingConstants.PROPERTIES;
 import static org.opensearch.neuralsearch.constants.MappingConstants.TYPE;
+import static org.opensearch.neuralsearch.constants.SemanticFieldConstants.DENSE_EMBEDDING_CONFIG;
 import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.EMBEDDING_FIELD_NAME;
 import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.CHUNKS_FIELD_NAME;
 import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.CHUNKS_TEXT_FIELD_NAME;
 import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.DEFAULT_MODEL_CONFIG;
+import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.KNN_VECTOR_DATA_TYPE_FIELD_NAME;
 import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.KNN_VECTOR_DIMENSION_FIELD_NAME;
 import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.KNN_VECTOR_METHOD_DEFAULT_NAME;
 import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.KNN_VECTOR_METHOD_FIELD_NAME;
@@ -47,6 +50,12 @@ public class SemanticInfoConfigBuilder {
     private Integer embeddingDimension;
     private Boolean chunkingEnabled;
     private String semanticFieldSearchAnalyzer;
+    private Map<String, Object> denseEmbeddingConfig;
+    private final static List<String> UNSUPPORTED_DENSE_EMBEDDING_CONFIG = List.of(
+        KNN_VECTOR_DIMENSION_FIELD_NAME,
+        KNN_VECTOR_DATA_TYPE_FIELD_NAME,
+        KNN_VECTOR_METHOD_SPACE_TYPE_FIELD_NAME
+    );
 
     public SemanticInfoConfigBuilder(@NonNull final NamedXContentRegistry xContentRegistry) {
         this.xContentRegistry = xContentRegistry;
@@ -74,29 +83,21 @@ public class SemanticInfoConfigBuilder {
      * @return Config of the semantic info fields.
      */
     public Map<String, Object> build() {
-        if (semanticFieldSearchAnalyzer != null && RankFeaturesFieldMapper.CONTENT_TYPE.equals(embeddingFieldType) == false) {
-            throw new IllegalArgumentException(
-                String.format(
-                    Locale.ROOT,
-                    "Cannot build the semantic info config because the embedding field type %s cannot build with semantic field search analyzer %s",
-                    embeddingFieldType,
-                    semanticFieldSearchAnalyzer
-                )
-            );
-        }
+        validate();
+
         final Map<String, Object> embeddingFieldConfig = switch (embeddingFieldType) {
             case KNNVectorFieldMapper.CONTENT_TYPE -> buildKnnFieldConfig();
             case RankFeaturesFieldMapper.CONTENT_TYPE -> buildRankFeaturesFieldConfig();
             default -> throw new IllegalArgumentException(
                 String.format(
                     Locale.ROOT,
-                    "Cannot build the semantic" + " info config because the embedding field type %s is not supported.",
+                    "Cannot build the semantic info config because the embedding field type %s is not supported.",
                     embeddingFieldType
                 )
             );
         };
 
-        if (chunkingEnabled) {
+        if (Boolean.TRUE.equals(chunkingEnabled)) {
             final Map<String, Object> chunksConfig = Map.of(
                 TYPE,
                 ObjectMapper.NESTED_CONTENT_TYPE,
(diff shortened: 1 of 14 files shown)


heemin32:
Could you add one integ test where we pass values to all available parameters?

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
```diff
@@ -73,29 +82,21 @@ public SemanticInfoConfigBuilder(@NonNull final NamedXContentRegistry xContentRe
...
                     Locale.ROOT,
-                    "Cannot build the semantic" + " info config because the embedding field type %s is not supported.",
+                    "Cannot build the semantic info config because the embedding field type %s is not supported.",
                     embeddingFieldType
                 )
             );
         };
 
-        if (chunkingEnabled) {
+        if (Boolean.TRUE.equals(chunkingEnabled)) {
```
will-hwang, line 100: i think `if (chunkingEnabled)` is simpler

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
```diff
@@ -108,17 +109,91 @@ public Map<String, Object> build() {
...
         config.put(KNN_VECTOR_DIMENSION_FIELD_NAME, this.embeddingDimension);
+
         final Map<String, Object> methodConfig = new HashMap<>();
-        methodConfig.put(KNN_VECTOR_METHOD_NAME_FIELD_NAME, this.knnMethodName);
+
+        if (denseEmbeddingConfig != null) {
+            validateDenseEmbeddingConfig(denseEmbeddingConfig);
+            config.putAll(denseEmbeddingConfig);
+
+            final Object methodConfigObj = denseEmbeddingConfig.get(KNN_VECTOR_METHOD_FIELD_NAME);
```
will-hwang, line 148: should we throw an exception if `methodconfigObj` is null?

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
```diff
@@ -108,17 +109,91 @@ public Map<String, Object> build() {
...
+        // Safe to override or add new method values
+        methodConfig.putIfAbsent(KNN_VECTOR_METHOD_NAME_FIELD_NAME, this.knnMethodName);
+        // Always use the one we found from the model config
         methodConfig.put(KNN_VECTOR_METHOD_SPACE_TYPE_FIELD_NAME, this.spaceType);
         config.put(KNN_VECTOR_METHOD_FIELD_NAME, methodConfig);
         return config;
     }
 
+    private void validateDenseEmbeddingConfig(@NonNull final Map<String, Object> denseEmbeddingConfig) {
+        UNSUPPORTED_DENSE_EMBEDDING_CONFIG.forEach(name -> {
```
will-hwang, line 174: instead of checking for unsupported config, should we keep the list of supported config, and check if the given config exists there

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
bzhangam: To handle the `null` better otherwise will run into an exception when we try to convert `null` to a boolean.

bzhangam:
> Could you add one integ test where we pass values to all available parameters?

Sure.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
bzhangam: It's a valid case to be null.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
bzhangam: That logic is kind complicated because knn can support different parameters combo so we just check the new validation logic here and the remaining will be handled by the knn field mapper.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
```diff
@@ -46,6 +49,12 @@ public class SemanticInfoConfigBuilder {
     private Integer embeddingDimension;
     private Boolean chunkingEnabled;
     private String semanticFieldSearchAnalyzer;
+    private Map<String, Object> denseEmbeddingConfig;
+    private List<String> UNSUPPORTED_DENSE_EMBEDDING_CONFIG = List.of(
```
junqiu-lei: should be declared as static final constant.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
```diff
@@ -108,17 +109,91 @@ public Map<String, Object> build() {
...
         final Map<String, Object> config = new HashMap<>();
         config.put(TYPE, KNNVectorFieldMapper.CONTENT_TYPE);
         config.put(KNN_VECTOR_DIMENSION_FIELD_NAME, this.embeddingDimension);
+
         final Map<String, Object> methodConfig = new HashMap<>();
-        methodConfig.put(KNN_VECTOR_METHOD_NAME_FIELD_NAME, this.knnMethodName);
+
+        if (denseEmbeddingConfig != null) {
+            validateDenseEmbeddingConfig(denseEmbeddingConfig);
+            config.putAll(denseEmbeddingConfig);
```
junqiu-lei, line 146: Is there any risk to override all config values here? Should we check for conflicting configuration items first, or use a more precise configuration merge strategy?

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
```diff
@@ -73,29 +82,21 @@ public SemanticInfoConfigBuilder(@NonNull final NamedXContentRegistry xContentRe
...
-            throw new IllegalArgumentException(
-                String.format(
-                    Locale.ROOT,
-                    "Cannot build the semantic info config because the embedding field type %s cannot build with semantic field search analyzer %s",
-                    embeddingFieldType,
-                    semanticFieldSearchAnalyzer
-                )
-            );
-        }
+        validate();
```
junqiu-lei, line 86: Shall we validate `dense_embedding_config` structure during parse phase rather than waiting until build phase? This would provide earlier error feedback to users when configuration is invalid.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
bzhangam: There is no risk. At the end this config will be passed to the knn_vector field and we rely on the knn vector field mapper to handle the conflict.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
bzhangam: Here we are not validating its structure, we simply validate it should not be used for the sparse embedding. Besides it's kind complicated to fully validate the data structure of the `dense_embedding_config` and we simply pass it through to the knn field mapper to handle the validation.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
bzhangam: yeah

src/test/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapperTests.java
```diff
@@ -51,13 +52,19 @@ public class SemanticFieldMapperTests extends OpenSearchTestCase {
     @Mock
     private ParseContext parseContext;
 
+    private final Map<String, Object> TEST_DENSE_EMBEDDING_CONFIG = Map.of(
+        "method",
+        Map.of("engine", "lucene", "parameters", Map.of("ef_construction", 128))
```
will-hwang, line 57: should we add one for FAISS after it can be supported? we should leave a comment here

junqiu-lei (approved):

src/test/java/org/opensearch/neuralsearch/mapper/SemanticFieldMapperTests.java
bzhangam: I think it should be fine. It's already covered by our unit test that by default we will use faiss.

PR #1427, opened 2025-07-07
Handle remote dense model properly during mapping transform

### LINKED ISSUE #1426: [BUG] Semantic Field Not Working with Remote Dense/Text_EMBEDDING Model
### What is the bug?
Semantic Field Not Working with Remote Dense/Text_EMBEDDING Model and throw error: 
```
"Failed to transform the mapping for the semantic field at text due to Model xxxxxxxxxxxxxxx is a remote text embedding model but model config is not a text embedding config"
```

1. Create a remote text embedding model.
2. Use it to create a semantic field in a index.

Then you will see the error.

The index creation should success and we should auto generate embedding field based on the remote text embedding model properly.

OpenSearch 3.1

N/A

We forget the refactor the validation after this [PR](https://github.com/opensearch-project/ml-commons/pull/3786/files#diff-2178c7be1c50d09b78c521c79aaaf4e7ccf62bc45856dbfd152662712236bdce). In [here](https://github.com/opensearch-project/neural-search/blob/ba374d537e47d499358c46a94e902d2df5ab4502/src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java#L148) we simply check the model config should be text embedding but now it also can be remote.

bzhangam:
### Description
Handle remote dense model properly during mapping transform.

### Related Issues
Closes #1426

### Check List
- [] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
@@ -12,6 +12,7 @@ import org.opensearch.knn.index.mapper.KNNVectorFieldMapper;
 import org.opensearch.ml.common.FunctionName;
 import org.opensearch.ml.common.MLModel;
 import org.opensearch.ml.common.model.MLModelConfig;
+import org.opensearch.ml.common.model.RemoteModelConfig;
 import org.opensearch.ml.common.model.TextEmbeddingModelConfig;
 import reactor.util.annotation.NonNull;
 
@@ -125,10 +126,16 @@ public class SemanticInfoConfigBuilder {
         return config;
     }
 
+    /**
+     * Extract the info from the ML model and set it in the SemanticInfoConfigBuilder for build later.
+     * @param mlModel ML model info
+     * @param modelId ID of the ML model
+     * @return this
+     */
     // Here we also require the model id because sometimes the MLModel does not have that info.
     public SemanticInfoConfigBuilder mlModel(@NonNull final MLModel mlModel, @NonNull final String modelId) {
         switch (mlModel.getAlgorithm()) {
-            case FunctionName.TEXT_EMBEDDING -> extractInfoForTextEmbeddingModel(mlModel, modelId);
+            case FunctionName.TEXT_EMBEDDING -> extractInfoForTextEmbeddingModel(mlModel, modelId, false);
             case FunctionName.SPARSE_ENCODING, FunctionName.SPARSE_TOKENIZE -> extractInfoForSparseModel();
             case FunctionName.REMOTE -> extractInfoForRemoteModel(mlModel, modelId);
             default -> throw new IllegalArgumentException(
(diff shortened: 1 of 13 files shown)


src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
```diff
@@ -145,43 +146,57 @@ public SemanticInfoConfigBuilder mlModel(@NonNull final MLModel mlModel, @NonNul
...
+                    String.format(
+                        Locale.ROOT,
+                        "Model %s is marked as remote text embedding model, but model_config is a %s rather than a RemoteModelConfig.",
+                        modelId,
+                        mlModel.getModelConfig().getClass().getName()
+                    )
+                );
+            }
+        } else {
+            if (mlModel.getModelConfig() instanceof TextEmbeddingModelConfig textEmbeddingModelConfig) {
```
martin-gaievski: there are multiple lines the are duplicates between this else section and previous if section, consider extracting common code into some utility method/class

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
```diff
@@ -145,43 +146,57 @@ public SemanticInfoConfigBuilder mlModel(@NonNull final MLModel mlModel, @NonNul
...
 
-        final TextEmbeddingModelConfig textEmbeddingModelConfig = (TextEmbeddingModelConfig) mlModel.getModelConfig();
-
-        if (textEmbeddingModelConfig.getEmbeddingDimension() == null) {
+        if (embeddingDimension == null) {
             throw new IllegalArgumentException(
                 String.format(
                     Locale.ROOT,
-                    "Model %s is a remote text embedding model but the embedding dimension is not defined in the model config.",
+                    "Model %s is a text embedding model, but the embedding dimension is not defined in the model config.",
```
martin-gaievski, line 173:
```suggestion
                    "Model %s has undefined embedding dimension in model config",
```

martin-gaievski (commented):
I don't see unit test for happy path with remote model, please add one.

Some minor comments regarding code organization, but overall looks good.

bzhangam:
> I don't see unit test for happy path with remote model, please add one.
> 
> Some minor comments regarding code organization, but overall looks good.

It's covered in testTransform_whenSemanticFieldsWithRemoteModels_thenTransformMappingsSuccessfully in SemanticMappingTransformerTests. But I can also add one in SemanticInfoConfigBuilderTests.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
bzhangam: I think they are not duplicates since we need to either convert the model config to RemoteModelConfig or TextEmbeddingModelConfig to access the embedding dimension. And the error messages in the exception are also different.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
bzhangam: Considering we also support sparse model I think it's better to keep the information that the model is a text embedding model. For sparse model it's ok to not define the embedding dimension.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
martin-gaievski: I mean the logic that we are doing is similar - check if config object is of a proper type then extract config, or throw otherwise. Exact type and error messages are parameters in such model. You can do something like this:

```
private void extractInfoForTextEmbeddingModel(@NonNull final MLModel mlModel, @NonNull final String modelId, final boolean isRemote) {
     final Function<MLModel, EmbeddingConfig> extractor = isRemote 
        ? this::extractRemoteConfig 
        : this::extractLocalConfig;
        
    final EmbeddingConfig config = extractor.apply(mlModel);
    validateAndSetEmbeddingConfig(config, modelId);
}

private EmbeddingConfig extractLocalConfig(MLModel mlModel) {
    return Optional.of(mlModel.getModelConfig())
        .filter(TextEmbeddingModelConfig.class::isInstance)
        .map(TextEmbeddingModelConfig.class::cast)
        .map(config -> new EmbeddingConfig(config.getEmbeddingDimension(), config.getAdditionalConfig()))
        .orElseThrow(() -> new IllegalArgumentException("Invalid local model config"));
}

private EmbeddingConfig extractRemoteConfig(MLModel mlModel) {
    return Optional.of(mlModel.getModelConfig())
        .filter(RemoteModelConfig.class::isInstance)
        .map(RemoteModelConfig.class::cast)
        .map(config -> new EmbeddingConfig(config.getEmbeddingDimension(), config.getAdditionalConfig()))
        .orElseThrow(() -> new IllegalArgumentException("Invalid remote model config"));
}

private void validateAndSetEmbeddingConfig(EmbeddingConfig config, String modelId) {
    this.embeddingDimension = Optional.ofNullable(config.embeddingDimension())
        .orElseThrow(() -> new IllegalArgumentException(
            String.format(Locale.ROOT,
                "Model %s is a text embedding model, but the embedding dimension is not defined in the model config.",
                modelId)
        ));
    
    this.spaceType = Optional.ofNullable(config.additionalConfig())
        .map(additionalConfig -> additionalConfig.get(KNN_VECTOR_METHOD_SPACE_TYPE_FIELD_NAME))
        .filter(String.class::isInstance)
        .map(String.class::cast)
        .orElseThrow(() -> createInvalidSpaceTypeException(modelId));
}
```

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
martin-gaievski: agree

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
junqiu-lei: +1 to split the method based on remote/local model. Could also easier to test individual scenarios.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
```diff
@@ -128,7 +129,7 @@ private Map<String, Object> buildRankFeaturesFieldConfig() {
     // Here we also require the model id because sometimes the MLModel does not have that info.
     public SemanticInfoConfigBuilder mlModel(@NonNull final MLModel mlModel, @NonNull final String modelId) {
```
junqiu-lei, line 136: Add JavaDoc comments

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
bzhangam: ok I think it's not about duplicates, it's more about split the methods to make it more readable. I can do that.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticInfoConfigBuilder.java
bzhangam: added

martin-gaievski (approved):
Looks good to me, thank you

Issue #1376, opened 2025-06-10
[Flaky test] testReduceAndRescore_whenRescorerThrowsException_thenFail  fails during gradle check


vibrantvarun:
### What is the issue?
`testReduceAndRescore_whenRescorerThrowsException_thenFail`  often fails while running gradle check. We need to investigate the fix.

### Stacktrace
https://github.com/opensearch-project/neural-search/actions/runs/15565340663/job/43827915424?pr=1345

### Logs
```
REPRODUCE WITH: gradlew ':test' --tests "org.opensearch.neuralsearch.search.query.HybridCollectorManagerTests.testReduceAndRescore_whenRescorerThrowsException_thenFail" -Dtests.seed=8B9D74D8ADA7B83D -Dtests.security.manager=false -Dtests.locale=pt-LU -Dtests.timezone=Asia/Urumqi -Druntime.java=21
> Task :test

HybridCollectorManagerTests > testReduceAndRescore_whenRescorerThrowsException_thenFail FAILED
    junit.framework.AssertionFailedError: Expected exception HybridSearchRescoreQueryException but no exception was thrown
        at __randomizedtesting.SeedInfo.seed([8B9D74D8ADA7B83D:71DB813DCFE2DE0F]:0)
        at org.apache.lucene.tests.util.LuceneTestCase.expectThrows(LuceneTestCase.java:2899)
        at org.apache.lucene.tests.util.LuceneTestCase.expectThrows(LuceneTestCase.java:2885)
        at org.opensearch.neuralsearch.search.query.HybridCollectorManagerTests.testReduceAndRescore_whenRescorerThrowsException_thenFail(HybridCollectorManagerTests.java:[110](https://github.com/opensearch-project/neural-search/actions/runs/15565340663/job/43827915424?pr=1345#step:4:111)5)
```

Issue #1199, opened 2025-02-25
[Flaky test]HybridCollectorManagerTests.testRescoreWithConcurrentSegmentSearch_whenMatchedDocsAndRescore_thenSuccessful 


martin-gaievski:
### What is the bug?
One of integ tests are failing if I execute following command:

```
./gradlew ':test' --tests "org.opensearch.neuralsearch.search.query.HybridCollectorManagerTests.testRescoreWithConcurrentSegmentSearch_whenMatchedDocsAndRescore_thenSuccessful" -Dtests.seed=F7CFD4E7CDCC6ADD
```

I do replicate the failure consistently in following environment

```
  Gradle Version        : 8.10.2
  OS Info               : Mac OS X 15.3.1 (aarch64)
  JDK Version           : 21 (Amazon Corretto JDK)
```

if I drop the test seed then error became flaky, most of times test will pass

heemin32:
https://github.com/opensearch-project/neural-search/pull/1430

PR #1430, opened 2025-07-08
Eliminate non-deterministic behavior in scoring for unit test

heemin32:
### Description
Two flaky test failures were identified. These tests are highly sensitive to document scoring for a given query. The issue stems from the IndexSearcher introducing randomness, resulting in inconsistent scores — sometimes returning zero unexpectedly, which causes the tests to fail.

To fix this, we removed the source of randomness to ensure deterministic scoring for a given query, regardless of seed value. The key change is the use of `new IndexSearcher(reader)`, which provides stable behavior.

Additionally, index options and omit norms were removed, as they are unrelated to the test’s purpose and do not contribute meaningful coverage.

### Related Issues
https://github.com/opensearch-project/neural-search/issues/1376
https://github.com/opensearch-project/neural-search/issues/1199

### Check List

### DIFF
--- a/src/test/java/org/opensearch/neuralsearch/search/query/HybridCollectorManagerTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/search/query/HybridCollectorManagerTests.java
@@ -924,10 +924,8 @@ public class HybridCollectorManagerTests extends OpenSearchQueryTestCase {
         when(searchContext.shouldUseConcurrentSearch()).thenReturn(true);
         // index segment 1
         Directory directory = newDirectory();
-        final IndexWriter w = new IndexWriter(directory, newIndexWriterConfig(new MockAnalyzer(random())));
+        final IndexWriter w = new IndexWriter(directory, newIndexWriterConfig());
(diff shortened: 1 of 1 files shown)


will-hwang (approved):

bzhangam:
I notice other tests are also using `newSearcher(reader) ` with randomness do you know why other tests are ok with it? Wonder should we unify it to just use `new IndexSearcher(reader)` if the randomness is not necessary? If we do need the randomness should we modify the verify logic to account for the 0 score case?

heemin32:
> I notice other tests are also using `newSearcher(reader) ` with randomness do you know why other tests are ok with it? Wonder should we unify it to just use `new IndexSearcher(reader)` if the randomness is not necessary? If we do need the randomness should we modify the verify logic to account for the 0 score case?

Have look into detail for other tests but I guess other tests might have more documents indexed and does not get impacted one of them got zero score in their scenario. We can change everything to use this code but I was waiting for this comment to make the change for the entire code base before doing it.

PR #1431, opened 2025-07-08
[Feature] Add upper bound parameter for min_max normalization technique

### LINKED ISSUE #1210: [FEATURE] Add upper_bound in min-max normalization
Currently, the min-max normalization technique uses the actual retrieved scores to find the minimum and maximum scores and then scales all scores using the interval [min score, ..., max score]. Team is adding a lower_bound feature that allows users to provide a minimum score value. For some scenarios, a similar feature is required for the maximum score.

In following scenario we are using knn/neural query, and we know that scores are always in a interval [0.75...1.0]. For some query resulting scores are [0.77, 0.77, 0.76, 0.75, 0.75], with min-max this will be normalized to [1.0, 1.0, 0.5, 0.0, 0.0], indicating that document with score 0.77 has score of 1.0 and is very relevant, which is not entirely correct. 
If we use an upper bound score of 1.0, then the scores in this example will be normalized to [0.08, 0.08, 0.04, 0.0, 0.0], where the same document has a score of 0.08, indicating low relevancy.

Proposed Solution:
The upper bound score should be configurable at the sub-query level, with the ability to skip it for some sub-queries. Configuration can be similar to the lower_bound score feature:

Modes:
- ignore: upper bound is ignored
- clip: score is clipped to the upper bound if it exceeds it
- apply: use actual score value if it exceeds upper bound score

```
{
  "description": "Normalization processor for hybrid search",
  "phase_results_processors": [
    {
      "normalization-processor": {
        "normalization": {
          "technique": "min_max",
          "parameters": {
            "upper_bounds": [
              { 
                "mode": "apply",
                "max_score": 1.0
              }, 
                "mode": "clip",
                "max_score": 5.0
              }, 
                "mode": "ignore"
              }
            ]
          }
        },
        "combination": {
          "technique": "arithmetic_mean"
          }
        }
      }
    }
  ]
}
```

ryanbogan:
### Description
Adds upper bound parameter for min_max normalization technique

### Related Issues
Resolves #1210

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
@@ -17,8 +17,6 @@ import java.util.stream.Collectors;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import org.apache.commons.lang3.Validate;
-import org.apache.commons.lang3.tuple.ImmutablePair;
-import org.apache.commons.lang3.tuple.Pair;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TopDocs;
 import org.opensearch.ml.repackage.com.google.common.annotations.VisibleForTesting;
@@ -65,8 +63,8 @@ public class MinMaxScoreNormalizationTechnique implements ScoreNormalizationTech
         Set.of(PARAM_NAME_BOUND_MODE, PARAM_NAME_UPPER_BOUND_MAX_SCORE)
     );
 
-    private final Optional<List<Pair<BoundMode, Float>>> lowerBoundsOptional;
-    private final Optional<List<Pair<BoundMode, Float>>> upperBoundsOptional;
+    private final Optional<List<Map<String, Object>>> lowerBoundsParamsOptional;
+    private final Optional<List<Map<String, Object>>> upperBoundsParamsOptional;
 
     public MinMaxScoreNormalizationTechnique() {
         this(Map.of(), new ScoreNormalizationUtil());
@@ -74,8 +72,8 @@ public class MinMaxScoreNormalizationTechnique implements ScoreNormalizationTech
 
     public MinMaxScoreNormalizationTechnique(final Map<String, Object> params, final ScoreNormalizationUtil scoreNormalizationUtil) {
         scoreNormalizationUtil.validateParameters(params, SUPPORTED_PARAMETERS, NESTED_PARAMETERS);
-        lowerBoundsOptional = getBounds(params, PARAM_NAME_LOWER_BOUNDS, this::parseLowerBound);
-        upperBoundsOptional = getBounds(params, PARAM_NAME_UPPER_BOUNDS, this::parseUpperBound);
+        lowerBoundsParamsOptional = getBoundsParams(params, PARAM_NAME_LOWER_BOUNDS);
+        upperBoundsParamsOptional = getBoundsParams(params, PARAM_NAME_UPPER_BOUNDS);
     }
 
     /**
@@ -122,25 +120,21 @@ public class MinMaxScoreNormalizationTechnique implements ScoreNormalizationTech
     }
 
     private boolean isBoundsAndSubQueriesCountMismatched(List<TopDocs> topDocsPerSubQuery) {
-        boolean lowerBoundsMismatch = lowerBoundsOptional.isPresent()
+        boolean lowerBoundsMismatch = lowerBoundsParamsOptional.isPresent()
             && !topDocsPerSubQuery.isEmpty()
-            && lowerBoundsOptional.get().size() != topDocsPerSubQuery.size();
-        boolean upperBoundsMismatch = upperBoundsOptional.isPresent()
+            && lowerBoundsParamsOptional.get().size() != topDocsPerSubQuery.size();
+        boolean upperBoundsMismatch = upperBoundsParamsOptional.isPresent()
             && !topDocsPerSubQuery.isEmpty()
-            && upperBoundsOptional.get().size() != topDocsPerSubQuery.size();
+            && upperBoundsParamsOptional.get().size() != topDocsPerSubQuery.size();
         return lowerBoundsMismatch || upperBoundsMismatch;
     }
 
     private LowerBound getLowerBound(int subQueryIndex) {
-        return lowerBoundsOptional.map(
-            pairs -> new LowerBound(true, pairs.get(subQueryIndex).getLeft(), pairs.get(subQueryIndex).getRight())
-        ).orElseGet(LowerBound::new);
+        return lowerBoundsParamsOptional.map(bounds -> bounds.get(subQueryIndex)).map(LowerBound::new).orElseGet(LowerBound::new);
     }
 
     private UpperBound getUpperBound(int subQueryIndex) {
-        return upperBoundsOptional.map(
-            pairs -> new UpperBound(true, pairs.get(subQueryIndex).getLeft(), pairs.get(subQueryIndex).getRight())
-        ).orElseGet(UpperBound::new);
+        return upperBoundsParamsOptional.map(bounds -> bounds.get(subQueryIndex)).map(UpperBound::new).orElseGet(UpperBound::new);
     }
 
     private MinMaxScores getMinMaxScoresResult(final List<CompoundTopDocs> queryTopDocs) {
@@ -161,17 +155,21 @@ public class MinMaxScoreNormalizationTechnique implements ScoreNormalizationTech
     public String describe() {
         StringBuilder description = new StringBuilder(TECHNIQUE_NAME);
 
-        lowerBoundsOptional.ifPresent(lb -> {
-            String lowerBounds = lb.stream()
-                .map(pair -> String.format(Locale.ROOT, "(%s, %s)", pair.getLeft(), pair.getRight()))
-                .collect(Collectors.joining(", ", "[", "]"));
+        lowerBoundsParamsOptional.ifPresent(lb -> {
+            String lowerBounds = lb.stream().map(boundMap -> {
+                BoundMode mode = BoundMode.fromString(Objects.toString(boundMap.get("mode"), ""));
+                String minScore = Objects.toString(boundMap.get("min_score"), String.valueOf(LowerBound.DEFAULT_LOWER_BOUND_SCORE));
+                return String.format(Locale.ROOT, "(%s, %s)", mode, minScore);
+            }).collect(Collectors.joining(", ", "[", "]"));
             description.append(String.format(Locale.ROOT, ", lower bounds %s", lowerBounds));
         });
 
-        upperBoundsOptional.ifPresent(ub -> {
-            String upperBounds = ub.stream()
-                .map(pair -> String.format(Locale.ROOT, "(%s, %s)", pair.getLeft(), pair.getRight()))
-                .collect(Collectors.joining(", ", "[", "]"));
+        upperBoundsParamsOptional.ifPresent(ub -> {
+            String upperBounds = ub.stream().map(boundMap -> {
+                BoundMode mode = BoundMode.fromString(Objects.toString(boundMap.get("mode"), ""));
+                String maxScore = Objects.toString(boundMap.get("max_score"), String.valueOf(UpperBound.DEFAULT_UPPER_BOUND_SCORE));
+                return String.format(Locale.ROOT, "(%s, %s)", mode, maxScore);
+            }).collect(Collectors.joining(", ", "[", "]"));
             description.append(String.format(Locale.ROOT, ", upper bounds %s", upperBounds));
         });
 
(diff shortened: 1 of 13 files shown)


ryanbogan:
This PR is currently in draft state as I still need to add unit tests for the combination of lower and upper bound score calculations and integration tests for multiple different scenarios.

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -220,47 +257,78 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
-     * @param params user provided input parameters for this technique
-     * @return optional list of lower bounds. Can be empty in case lower bounds are not provided
-     */
-    private Optional<List<Pair<LowerBound.Mode, Float>>> getLowerBounds(final Map<String, Object> params) {
+    private boolean isSingleScore(float score, float minScore, float maxScore) {
+        return Floats.compare(maxScore, minScore) == 0 && Floats.compare(maxScore, score) == 0;
+    }
+
+    private float calculateNormalizedScore(float score, float effectiveMinScore, float effectiveMaxScore) {
+        float normalizedScore = (score - effectiveMinScore) / (effectiveMaxScore - effectiveMinScore);
```
martin-gaievski, line 293: should we add protection from division by zero. In theory scenario when effectiveMaxScore == effectiveMinScore is possible. Simple check will do the job:
```
    if (Floats.compare(effectiveMaxScore, effectiveMinScore) == 0) {
        return SINGLE_RESULT_SCORE;
    }
```

src/main/java/org/opensearch/neuralsearch/processor/normalization/bounds/UpperBound.java
```diff
@@ -0,0 +1,64 @@
...
+     */
+    @Override
+    public float determineEffectiveScore(float score, float minScore, float maxScore) {
+        if (enabled == false) {
+            return maxScore;
+        }
+
+        return switch (mode) {
+            case APPLY -> (minScore < boundScore && score < boundScore) ? boundScore : maxScore;
+            case CLIP -> minScore > boundScore ? maxScore : boundScore;
```
martin-gaievski, line 60: If minScore > boundScore, this could create inconsistent normalization ranges

martin-gaievski (commented):
Is there a combined bounds test for scenarios with both upper and lower bounds applied simultaneously.

If it's not there please add something like this

```java
@Test
public void testCombinedBounds_whenBothUpperAndLowerBounds_thenSuccessful() {
    Map<String, Object> params = Map.of(
        "lower_bounds", List.of(Map.of("mode", "apply", "min_score", 0.2)),
        "upper_bounds", List.of(Map.of("mode", "apply", "max_score", 0.8))
    );
    // test and assert
}
```

src/main/java/org/opensearch/neuralsearch/processor/normalization/bounds/UpperBound.java
ryanbogan: How does that differ from this section in the LowerBound?

```
case CLIP -> maxScore < boundScore ? minScore : boundScore;
```

martin-gaievski: you're correct, both upper and lower bounds follow same approach. I checked scenario when that can lead to inconsistency, it's extreme edge case and system prioritizes mathematical validity. I'm ok with keeping your code

ryanbogan:
Looking into failing tests

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: @ryanbogan  where did we addressed this concern?

src/main/java/org/opensearch/neuralsearch/processor/normalization/bounds/ScoreBound.java
```diff
@@ -0,0 +1,42 @@
...
+    protected final float boundScore;
+
+    protected ScoreBound(boolean enabled, BoundMode mode, float boundScore) {
+        this.enabled = enabled;
+        this.mode = mode;
+        this.boundScore = boundScore;
+    }
+
+    /**
+     * Determines the effective score based on the upper bound constraints.
```
martin-gaievski: this shouldn't say "upper bound"

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
martin-gaievski: also can you add unit test specifically for scenario when effectiveMaxScore == effectiveMinScore

ryanbogan: https://github.com/opensearch-project/neural-search/pull/1431/files#diff-8f54d001a9ace0400eaa438e830c0a44f88072da24ecfdd40c9569faf31da0f0R275-R277

It's currently before this method is called.

I'll move it to the calculateNormalizedScore method though.

martin-gaievski: sounds good

martin-gaievski (approved):
Looks good, thank you @ryanbogan

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -220,47 +258,83 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
+    protected float calculateNormalizedScore(float score, float effectiveMinScore, float effectiveMaxScore) {
+        if (Floats.compare(effectiveMaxScore, effectiveMinScore) == 0) {
+            return SINGLE_RESULT_SCORE;
+        }
+
+        float normalizedScore = (score - effectiveMinScore) / (effectiveMaxScore - effectiveMinScore);
+        return normalizedScore == 0.0f ? MIN_SCORE : normalizedScore;
+    }
+
+    private Optional<List<Pair<BoundMode, Float>>> getLowerBounds(final Map<String, Object> params) {
```
owaiskazi19: We don't need this and the below getter. They can be called directly on L77 and 78. This approach would also reduce the almost similar names of the getters `getLowerBounds` and `getLowerBound`

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -220,47 +258,83 @@ private float[] getMinScores(final List<CompoundTopDocs> queryTopDocs, final int
...
+        return getBounds(params, PARAM_NAME_LOWER_BOUNDS, this::parseLowerBound);
+    }
+
+    private Optional<List<Pair<BoundMode, Float>>> getUpperBounds(final Map<String, Object> params) {
+        return getBounds(params, PARAM_NAME_UPPER_BOUNDS, this::parseUpperBound);
+    }
+
+    private <T> Optional<List<Pair<T, Float>>> getBounds(
```
owaiskazi19: Can we move this method to BoundMode.java

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -269,39 +343,60 @@ private Optional<List<Pair<LowerBound.Mode, Float>>> getLowerBounds(final Map<St
...
-        Object minScoreObj = lowerBound.get(PARAM_NAME_LOWER_BOUND_MIN_SCORE);
-        if (minScoreObj == null) {
-            return LowerBound.DEFAULT_LOWER_BOUND_SCORE;
+    /**
+     * Parse each upper bound item and return a pair of mode and max score
+     * @param boundObj upper bound item provided by the client
+     * @return a single pair of mode and max score
+     */
+    @SuppressWarnings("unchecked")
+    private Pair<BoundMode, Float> parseUpperBound(Object boundObj) {
```
owaiskazi19: This can be moved to UpperBound.java

src/test/java/org/opensearch/neuralsearch/processor/NormalizationProcessorIT.java
```diff
@@ -471,6 +473,301 @@ public void testMinMaxLowerBounds_whenLowerBoundsIsGreaterThenActualMinScore_the
         assertQueryResults(searchResponseAsMapPartialMatch, 4, false, Range.between(0.33f, 1.0f));
     }
 
+    @SneakyThrows
+    public void testMinMaxUpperBounds_whenMultipleShards_thenSuccessful() {
```
owaiskazi19, line 477: Do we have a test for single shard scenario?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -1862,6 +1863,29 @@ protected void createSearchPipeline(
                 }
                 stringBuilderForContentBody.append("]");
             }
+            if (normalizationParams.containsKey(PARAM_NAME_UPPER_BOUNDS)) {
```
owaiskazi19, line 1866: Do we have an IT with both upper bound and lower bound?

owaiskazi19 (approved):
Looks clean. Few comments

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
ryanbogan: https://github.com/opensearch-project/neural-search/pull/1431/files#diff-7f52eb53e630265669fc38cdb33b61a7093400cc8c1148adc0ec5aa999e86ec2R668-R770

src/test/java/org/opensearch/neuralsearch/processor/NormalizationProcessorIT.java
ryanbogan: Added

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -325,54 +317,37 @@ private <T> Optional<List<Pair<T, Float>>> getBounds(
...
-        }
-
-        Map<String, Object> lowerBound = (Map<String, Object>) boundObj;
-
-        String lowerBoundModeValue = Objects.toString(lowerBound.get(PARAM_NAME_BOUND_MODE), "");
-        BoundMode mode = BoundMode.fromString(lowerBoundModeValue);
-        float minScore = extractAndValidateScore(lowerBound, PARAM_NAME_LOWER_BOUND_MIN_SCORE, LowerBound.DEFAULT_LOWER_BOUND_SCORE);
+        String scoreParamName = paramName.equals(PARAM_NAME_LOWER_BOUNDS)
+            ? PARAM_NAME_LOWER_BOUND_MIN_SCORE
+            : PARAM_NAME_UPPER_BOUND_MAX_SCORE;
```
martin-gaievski: if we add new mode later this code will silently hide the error and will return  PARAM_NAME_UPPER_BOUND_MAX_SCORE. Replace this with exact match between score param and input param names, you can use Map and throw exception for case of unrecognized parameter

ryanbogan: Since both scoreParamName and defaultScore depend on this same parameter and would require different maps, I made a switch statement instead.

martin-gaievski: switch statement works too, please add a unit test for scenario with unsupported parameter

src/main/java/org/opensearch/neuralsearch/processor/normalization/MinMaxScoreNormalizationTechnique.java
```diff
@@ -134,12 +153,27 @@ public String techniqueName() {
...
+        });
+
+        upperBoundsParamsOptional.ifPresent(ub -> {
+            String upperBounds = ub.stream().map(boundMap -> {
+                BoundMode mode = BoundMode.fromString(Objects.toString(boundMap.get("mode"), ""));
+                String maxScore = Objects.toString(boundMap.get("max_score"), String.valueOf(UpperBound.DEFAULT_UPPER_BOUND_SCORE));
+                return String.format(Locale.ROOT, "(%s, %s)", mode, maxScore);
+            }).collect(Collectors.joining(", ", "[", "]"));
+            description.append(String.format(Locale.ROOT, ", upper bounds %s", upperBounds));
+        });
```
owaiskazi19: We can reduce the duplication here like
```java
private String buildBoundDescription(
    Optional<List<Map<String, Object>>> boundsOptional,
    String label,
    double defaultScore
) {
    return boundsOptional.map(bounds -> {
        String formatted = bounds.stream()
            .map(boundMap -> {
                BoundMode mode = BoundMode.fromString(Objects.toString(boundMap.get("mode"), ""));
                String score = Objects.toString(boundMap.get(label.equals("lower") ? "min_score" : "max_score"), String.valueOf(defaultScore));
                return String.format(Locale.ROOT, "(%s, %s)", mode, score);
            })
            .collect(Collectors.joining(", ", "[", "]"));
        return String.format(Locale.ROOT, ", %s bounds %s", label, formatted);
    }).orElse("");
}
```
and then call it with 
```java
description.append(buildBoundDescription(lowerBoundsParamsOptional, "lower", LowerBound.DEFAULT_LOWER_BOUND_SCORE));
description.append(buildBoundDescription(upperBoundsParamsOptional, "upper", UpperBound.DEFAULT_UPPER_BOUND_SCORE));

```

owaiskazi19 (commented):
Looks much cleaner with a minor comment

owaiskazi19 (approved):
Looks good

heemin32 (approved):

PR #1434, opened 2025-07-09
Allow configre prune strategy for sparse encoding of the semantic field.

### LINKED ISSUE #1351: [FEATURE] Semantic Field Enhancement - Allow to Configure Prune for Embedding Generation
In the initial launch, we don’t expose interfaces that let users configure how embeddings are generated for the semantic field. Instead, users can only specify the ID of the ML model to use, and we rely on the model's metadata to retrieve configuration details. However, there are certain settings that users should be able to configure more flexibly going forward.

Currently the embedding generation is done through a system generated ingest processor which users cannot configure it like a normal ingest processor. By default we set prune_type as max_ratio with 0.1 ratio based on this [doc](https://docs.opensearch.org/docs/latest/ingest-pipelines/processors/sparse-encoding/#:~:text=0.1%20demonstrates%20strong-,generalization,-on%20test%20datasets): Among all pruning options, specifying max_ratio as equal to 0.1 demonstrates strong generalization on test datasets. This approach reduces storage requirements by approximately 40% while incurring less than a 1% loss in search relevance. But we should allow users to configure it.

We should allow users to configure the system ingest processor through the semantic field to control the embedding generation behavior.

prune_type: The prune strategy for sparse vectors. Valid values are max_ratio, alpha_mass, top_k, abs_value, and none.

prune_ratio: The ratio for the pruning strategy. Required when prune_type is specified.

N/A

bzhangam:
### Description
Allow configre prune strategy for sparse encoding of the semantic field.

### Related Issues
Resolves #1351  

### Check List


src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -265,10 +265,13 @@ private void setInference(
...
                 Object embedding = modelIdValueToEmbeddingMap.get(Pair.of(modelId, chunk)).getLeft();
-                // TODO: In future we should allow user to configure how we should prune the sparse embedding
-                // for each semantic field. Then we can pull the config from the semantic config and use it here.
                 if (!isDenseModel) {
-                    embedding = PruneUtils.pruneSparseVector(PruneType.MAX_RATIO, pruneRatio, (Map<String, Float>) embedding);
+                    final SparseEncodingConfig sparseEncodingConfig = semanticFieldInfo.getSparseEncodingConfig();
+                    final PruneType pruneType = sparseEncodingConfig == null ? PruneType.MAX_RATIO : sparseEncodingConfig.getPruneType();
+                    if (PruneType.NONE.equals(pruneType) == false) {
+                        final Float pruneRatio = sparseEncodingConfig == null ? DEFAULT_PRUNE_RATIO : sparseEncodingConfig.getPruneRatio();
+                        embedding = PruneUtils.pruneSparseVector(pruneType, pruneRatio, (Map<String, Float>) embedding);
```
junqiu-lei, line 278: `applyPruningToSparseVector` would be more clearly show it's function.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -265,10 +265,13 @@ private void setInference(
                     throw exception;
                 }
                 Object embedding = modelIdValueToEmbeddingMap.get(Pair.of(modelId, chunk)).getLeft();
-                // TODO: In future we should allow user to configure how we should prune the sparse embedding
-                // for each semantic field. Then we can pull the config from the semantic config and use it here.
                 if (!isDenseModel) {
-                    embedding = PruneUtils.pruneSparseVector(PruneType.MAX_RATIO, pruneRatio, (Map<String, Float>) embedding);
+                    final SparseEncodingConfig sparseEncodingConfig = semanticFieldInfo.getSparseEncodingConfig();
+                    final PruneType pruneType = sparseEncodingConfig == null ? PruneType.MAX_RATIO : sparseEncodingConfig.getPruneType();
```
junqiu-lei: Looks like the apply of default pruneRatio is split across two different code paths:
1. here while `sparse_encoding_config` is not provided by the user
2. `sparse_encoding_config` is provided, but `prune_ratio` is omitted

Can we centralize this logic within the SparseEncodingConfig class?

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
bzhangam: I feel `pruneSparseVector` is already descriptive.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
bzhangam: Not sure how can we centralize it. We only can use SparseEncodingConfig if the config is defined by the user and when prune type is none we don't need to do pruning. It's not decided by if `prune_ratio` exists or not.

src/main/java/org/opensearch/neuralsearch/mapper/dto/SparseEncodingConfig.java
```diff
@@ -0,0 +1,201 @@
...
+    private Float pruneRatio;
+
+    /**
+     * Parse the json input when we define the sparse encoding config in the index mappings
+     * @param name parameter name
+     * @param ctx parse context
+     * @param value parameter value
+     * @return parsed SparseEncodingConfig
+     */
+    public static SparseEncodingConfig parse(@NonNull final String name, final Mapper.TypeParser.ParserContext ctx, final Object value) {
```
heemin32: I would rather make this method as a constructor. I think we should avoid static method as much as possible as it will make writing a unit test harder.

src/main/java/org/opensearch/neuralsearch/mapper/dto/SparseEncodingConfig.java
```diff
@@ -0,0 +1,201 @@
...
+        return SparseEncodingConfig.builder().pruneType(pruneType).pruneRatio(pruneRatio).build();
+    }
+
+    /**
+     * Parse the config of the semantic field to build the SparseEncodingConfig if it is defined. Only should be used
+     * to parse the valid semantic field config.
+     * @param fieldConfig semantic field config
+     * @return SparseEncodingConfig or null
+     */
+    public static SparseEncodingConfig parse(@NonNull final Map<String, Object> fieldConfig) {
```
heemin32: Same here. This can be a constructor.

src/main/java/org/opensearch/neuralsearch/mapper/dto/SparseEncodingConfig.java
```diff
@@ -0,0 +1,201 @@
...
+        final Map<String, Object> sparseEncodingConfig = (Map<String, Object>) fieldConfig.get(SPARSE_ENCODING_CONFIG);
+        final PruneType pruneType = readPruneType(sparseEncodingConfig, false);
+        final Float pruneRatio = readPruneRatio(sparseEncodingConfig, false);
+        if (pruneType == null && pruneRatio == null) {
+            return null;
+        }
+        return SparseEncodingConfig.builder().pruneType(pruneType).pruneRatio(pruneRatio).build();
+    }
+
+    private static Float readPruneRatio(@NonNull final Map<String, Object> config, final boolean shouldRemoveIt) {
```
heemin32: This method does two things: read and remove. Each method should do one thing at a time ideally. Could you make two separate methods?

src/main/java/org/opensearch/neuralsearch/mapper/dto/SparseEncodingConfig.java
```diff
@@ -0,0 +1,201 @@
...
+            } finally {
+                if (shouldRemoveIt) {
+                    config.remove(PRUNE_RATIO_FIELD);
+                }
+            }
+        }
+        return null;
+    }
+
+    private static PruneType readPruneType(@NonNull final Map<String, Object> config, final boolean shouldRemoveIt) {
```
heemin32: Same here. Could you make two separate methods?

src/main/java/org/opensearch/neuralsearch/mapper/dto/SparseEncodingConfig.java
```diff
@@ -0,0 +1,201 @@
...
+
+        // Case: pruneType is not NONE or null and pruneRatio is not null
+        if (PruneUtils.isValidPruneRatio(pruneType, pruneRatio) == false) {
+            throw new MapperParsingException(
+                String.format(
+                    Locale.ROOT,
+                    "Invalid %s and %s combo. Check %s for the valid combos.",
+                    PRUNE_RATIO_FIELD,
+                    PRUNE_TYPE_FIELD,
+                    "https://docs.opensearch.org/docs/latest/ingest-pipelines/processors/sparse-encoding/#pruning-sparse-vectors"
```
heemin32: I don't think it is good idea to have a link in the message because the link won't be stable. It can be broken or moved to other address in the future.

src/main/java/org/opensearch/neuralsearch/mapper/dto/SparseEncodingConfig.java
heemin32: Hoa about


```
// Read value.
private Float readPruneRatio(final Map<String, Object> config);
// Consume value. This remove value and return the removed value.
private Float consumePruneRatio(final Map<String, Object> config);
```

bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mapper/dto/SparseEncodingConfig.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mapper/dto/SparseEncodingConfig.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mapper/dto/SparseEncodingConfig.java
```diff
@@ -0,0 +1,204 @@
...
+            this.pruneType = pruneType;
+            return;
+        }
+
+        // Case: pruneType is not NONE or null and pruneRatio is not null
+        if (PruneUtils.isValidPruneRatio(pruneType, pruneRatio) == false) {
+            throw new MapperParsingException(
+                String.format(
+                    Locale.ROOT,
+                    "Invalid %s and %s combo. Check OpenSearch documentation sparse encoding processor for the valid combos.",
```
heemin32: This message is not very helpful. We should provide valid combination here instead asking user to look for other place.

src/main/java/org/opensearch/neuralsearch/mapper/dto/SparseEncodingConfig.java
```diff
@@ -0,0 +1,204 @@
...
+                )
+            );
+        }
+
+        this.pruneType = pruneType;
+        this.pruneRatio = pruneRatio;
+    }
+
+    /**
+     * Parse the config of the semantic field to build the SparseEncodingConfig if it is defined. Only should be used
```
heemin32: Update the comment.

src/main/java/org/opensearch/neuralsearch/mapper/dto/SparseEncodingConfig.java
```diff
@@ -0,0 +1,204 @@
...
+import static org.opensearch.neuralsearch.util.prune.PruneUtils.PRUNE_RATIO_FIELD;
+import static org.opensearch.neuralsearch.util.prune.PruneUtils.PRUNE_TYPE_FIELD;
+
+@Getter
+public class SparseEncodingConfig {
+    private PruneType pruneType;
+    private Float pruneRatio;
+
+    /**
+     * Parse the json input when we define the sparse encoding config in the index mappings
```
heemin32: Update the comment.

src/main/java/org/opensearch/neuralsearch/mapper/dto/SparseEncodingConfig.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mapper/dto/SparseEncodingConfig.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mapper/dto/SparseEncodingConfig.java
bzhangam: ok

heemin32 (approved):

junqiu-lei (approved):

PR #1446, opened 2025-07-14
Support configuring the chunking strategies through the semantic field.

### LINKED ISSUE #1354: [FEATURE] [FEATURE] Semantic Field Enhancement - Allow to Configure Chunking
In the initial launch we allow users to enable/disable the chunking for the semantic field but don't allow them to configure how to do the chunking. We simply use a default fixed length token chunker which is not flexible enough. So we should allow users to configure how to do chunking through the semantic field.

Allow users to configure how to do chunking through the semantic field.

N/A

bzhangam:
### Description
Support configuring the chunking strategies through the semantic field.

### Related Issues
Resolves #1354 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/mapper/dto/ChunkingConfig.java
+++ b/src/main/java/org/opensearch/neuralsearch/mapper/dto/ChunkingConfig.java
@@ -1,0 +1,173 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.mapper.dto;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.NonNull;
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.index.mapper.MapperParsingException;
+import org.opensearch.neuralsearch.processor.chunker.ChunkerValidatorFactory;
+import org.opensearch.neuralsearch.processor.chunker.Validator;
+
+import java.io.IOException;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static org.opensearch.neuralsearch.constants.SemanticFieldConstants.CHUNKING;
+import static org.opensearch.neuralsearch.processor.chunker.ChunkerFactory.CHUNKER_ALGORITHMS;
+
+@Builder
+@AllArgsConstructor
+@Getter
+public class ChunkingConfig {
+    /**
+     * algorithm parameter name
+     */
+    public static final String ALGORITHM_FIELD = "algorithm";
+    /**
+     * parameters parameter name
+     */
+    public static final String PARAMETERS_FIELD = "parameters";
+    private List<Map<String, Object>> configs;
+    private boolean enabled;
+
+    /**
+     * Construct ChunkingConfig with the json input defined in the chunking config of the semantic field in the index
+     * mappings.
+     * @param name parameter name
+     * @param value parameter value
+     */
+    public ChunkingConfig(@NonNull final String name, final Object value, @NonNull final ChunkerValidatorFactory chunkerValidatorFactory) {
+        if (value instanceof Boolean) {
+            this.enabled = (Boolean) value;
+            return;
+        }
+
+        if (value instanceof List<?>) {
+            List<Object> configList = (List<Object>) value;
+            if (configList.stream().allMatch(item -> item instanceof Map<?, ?>)) {
+                List<Map<String, Object>> parsedConfigs = configList.stream().map(item -> {
+                    validate((Map<String, Object>) item, chunkerValidatorFactory);
+                    return (Map<String, Object>) item;
+                }).collect(Collectors.toList());
+
+                this.enabled = true;
+                this.configs = parsedConfigs;
+                return;
+            }
+        }
+
+        throw new MapperParsingException(String.format(Locale.ROOT, "[%s] must be either a List of Maps or a Boolean", name));
+    }
+
+    /**
+     * Construct ChunkingConfig with valida semantic field config.
+     * @param fieldConfig semantic field config
+     */
+    public ChunkingConfig(@NonNull final Map<String, Object> fieldConfig) {
+        if (fieldConfig.containsKey(CHUNKING) == false) {
+            return;
+        }
+        final Object chunkingConfig = fieldConfig.get(CHUNKING);
+        if (chunkingConfig instanceof Boolean) {
+            this.enabled = (Boolean) chunkingConfig;
+            return;
+        }
+        if (chunkingConfig instanceof List<?>) {
+            List<Map<String, Object>> configList = (List<Map<String, Object>>) chunkingConfig;
+            this.configs = configList;
+            this.enabled = true;
+            return;
+        }
+        throw new IllegalStateException(
+            String.format(Locale.ROOT, "Invalid %s in the semantic field config. It is neither a Boolean nor a List of Maps.", CHUNKING)
+        );
+    }
+
+    private void validate(@NonNull final Map<String, Object> item, @NonNull final ChunkerValidatorFactory chunkerValidatorFactory) {
+        validateUnsupportedParameters(item);
+        final String algorithm = readAlgorithm(item);
+        if (item.get(PARAMETERS_FIELD) instanceof Map<?, ?>) {
+            Validator validator = chunkerValidatorFactory.getValidator(algorithm);
+            if (validator == null) {
+                throw new MapperParsingException(
+                    String.format(
+                        Locale.ROOT,
+                        "[%s] is not a valid chunking algorithm. A valid algorithm should be %s.",
+                        algorithm,
+                        CHUNKER_ALGORITHMS
+                    )
+                );
+            }
+            validator.validate((Map<String, Object>) item.get(PARAMETERS_FIELD));
+        } else if (item.get(PARAMETERS_FIELD) != null) {
+            throw new MapperParsingException(String.format(Locale.ROOT, "%s in %s should be a Map.", PARAMETERS_FIELD, CHUNKING));
+        }
+    }
+
+    private void validateUnsupportedParameters(@NonNull Map<String, Object> item) {
+        final Set<String> keys = new HashSet<>(item.keySet());
+        keys.remove(ALGORITHM_FIELD);
+        keys.remove(PARAMETERS_FIELD);
+        if (!keys.isEmpty()) {
+            throw new MapperParsingException(String.format(Locale.ROOT, "[%s] does not support parameters [%s]", CHUNKING, keys));
+        }
+    }
+
+    private String readAlgorithm(Map<String, Object> config) {
+        if (config.containsKey(ALGORITHM_FIELD)) {
+            if (config.get(ALGORITHM_FIELD) instanceof String algorithm && CHUNKER_ALGORITHMS.contains(algorithm)) {
+                return algorithm;
+            }
+            throw new MapperParsingException(String.format(Locale.ROOT, "[%s] must be %s", ALGORITHM_FIELD, CHUNKER_ALGORITHMS));
+        }
+        throw new MapperParsingException(
+            String.format(Locale.ROOT, "[%s] is required to configure the chunking strategy.", ALGORITHM_FIELD)
+        );
+    }
+
+    public void toXContent(@NonNull final XContentBuilder builder, String name) throws IOException {
+        if (configs == null) {
+            builder.field(name, enabled);
+        } else {
+            builder.startArray(name);
+            for (Map<String, Object> config : configs) {
+                builder.map(config);
+            }
+            builder.endArray();
+        }
+    }
+
+    @Override
+    public String toString() {
+        if (configs == null) {
+            return String.valueOf(enabled);
+        } else {
+            return configs.toString();
+        }
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) return true;
+        if (obj == null || getClass() != obj.getClass()) return false;
+
+        ChunkingConfig other = (ChunkingConfig) obj;
+        return new EqualsBuilder().append(enabled, other.enabled).append(configs, other.configs).isEquals();
+    }
+
+    @Override
+    public int hashCode() {
+        return (new HashCodeBuilder()).append(this.enabled).append(this.configs).toHashCode();
+    }
+}
--- a/src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
@@ -4,14 +4,19 @@
  */
 package org.opensearch.neuralsearch.processor.chunker;
 
+import lombok.Getter;
+
+import java.util.Locale;
 import java.util.Map;
 import java.util.List;
+
+import static org.opensearch.neuralsearch.processor.chunker.ChunkerParameterParser.parseIntegerWithDefault;
 
 /**
  * The interface for all chunking algorithms.
  * All algorithms need to parse parameters and chunk the content.
  */
-public abstract class Chunker {
+public abstract class Chunker implements Validator, Parser {
 
     /** Field name for specifying the maximum chunk limit in the configuration. */
     public static String MAX_CHUNK_LIMIT_FIELD = "max_chunk_limit";
@@ -25,13 +30,21 @@ public abstract class Chunker {
     /** Special value (-1) indicating that chunk limiting is disabled. */
     public static int DISABLED_MAX_CHUNK_LIMIT = -1;
 
+    @Getter
+    private int maxChunkLimit;
+
     /**
-     * Parse the parameters for chunking algorithm.
+     * Parse the common parameters for the chunking algorithm.
      * Throw IllegalArgumentException when parameters are invalid.
      *
      * @param parameters a map containing non-runtime parameters for chunking algorithms
      */
-    abstract void parseParameters(Map<String, Object> parameters);
+    @Override
+    public void parse(Map<String, Object> parameters) {
+        final int maxChunkLimit = parseIntegerWithDefault(parameters, MAX_CHUNK_LIMIT_FIELD, DEFAULT_MAX_CHUNK_LIMIT);
+        validateMaxChunkLimit(maxChunkLimit);
+        this.maxChunkLimit = maxChunkLimit;
+    }
 
     /**
      * Chunk the input string according to parameters and return chunked passages
@@ -56,4 +69,27 @@ public abstract class Chunker {
     }
 
     public abstract String getAlgorithmName();
+
+    /**
+     * Validate the common parameters for a chunker
+     * @param parameters parameters for a chunker
+     */
+    @Override
+    public void validate(Map<String, Object> parameters) {
+        final int maxChunkLimit = parseIntegerWithDefault(parameters, MAX_CHUNK_LIMIT_FIELD, DEFAULT_MAX_CHUNK_LIMIT);
+        validateMaxChunkLimit(maxChunkLimit);
+    }
+
+    void validateMaxChunkLimit(int maxChunkLimit) {
+        if (maxChunkLimit <= 0 && maxChunkLimit != DISABLED_MAX_CHUNK_LIMIT) {
+            throw new IllegalArgumentException(
+                String.format(
+                    Locale.ROOT,
+                    "Parameter [%s] must be positive or %s to disable this parameter",
+                    MAX_CHUNK_LIMIT_FIELD,
+                    DISABLED_MAX_CHUNK_LIMIT
+                )
+            );
+        }
+    }
 }
--- a/src/main/java/org/opensearch/neuralsearch/processor/chunker/Parser.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/chunker/Parser.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.processor.chunker;
+
+import java.util.Map;
+
+/**
+ * An interface to define the behavior of the chunker parser
+ */
+@FunctionalInterface
+public interface Parser {
+    /**
+     * Parse the parameters for the chunking algorithm.
+     * Throw IllegalArgumentException when parameters are invalid.
+     *
+     * @param parameters a map containing non-runtime parameters for chunking algorithms
+     */
+    void parse(Map<String, Object> parameters);
+}
--- a/src/main/java/org/opensearch/neuralsearch/processor/chunker/Validator.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/chunker/Validator.java
@@ -1,0 +1,19 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.processor.chunker;
+
+import java.util.Map;
+
+/**
+ * An interface to define the behavior of the chunker validator
+ */
+@FunctionalInterface
+public interface Validator {
+    /**
+     * Validate the parameters for a chunker
+     * @param parameters parameters for a chunker
+     */
+    void validate(Map<String, Object> parameters);
+}
--- a/src/main/java/org/opensearch/neuralsearch/processor/dto/SemanticFieldInfo.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/dto/SemanticFieldInfo.java
@@ -6,16 +6,26 @@ package org.opensearch.neuralsearch.processor.dto;
 
 import lombok.Builder;
 import lombok.Data;
+import lombok.NonNull;
+import org.opensearch.index.analysis.AnalysisRegistry;
+import org.opensearch.neuralsearch.mapper.dto.ChunkingConfig;
 import org.opensearch.neuralsearch.mapper.dto.SparseEncodingConfig;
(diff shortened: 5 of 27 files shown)


src/main/java/org/opensearch/neuralsearch/mapper/dto/ChunkingConfig.java
```diff
@@ -0,0 +1,157 @@
...
+    private boolean enabled;
+
+    /**
+     * Parse the json input when we define the chunking config in the index mappings
+     * @param name parameter name
+     * @param ctx parse context
+     * @param value parameter value
+     * @return ChunkingConfig
+     */
+    public static ChunkingConfig parse(@NonNull final String name, final Mapper.TypeParser.ParserContext ctx, final Object value) {
```
heemin32: Better to make it as a constructor.

src/main/java/org/opensearch/neuralsearch/mapper/dto/ChunkingConfig.java
```diff
@@ -0,0 +1,157 @@
...
+        );
+    }
+
+    /**
+     * Parse the config of the semantic field to build the ChunkingConfig if it is defined. Only should be used
+     * to parse the valid semantic field config.
+     * @param fieldConfig semantic field config
+     * @return ChunkingConfig or null
+     */
+    public static ChunkingConfig parse(@NonNull final Map<String, Object> fieldConfig) {
```
heemin32: Make it as constructor here as well.

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
```diff
@@ -25,13 +30,42 @@ public abstract class Chunker {
...
+    private int maxChunkLimit;
+
     /**
      * Parse the parameters for chunking algorithm.
      * Throw IllegalArgumentException when parameters are invalid.
      *
      * @param parameters a map containing non-runtime parameters for chunking algorithms
      */
-    abstract void parseParameters(Map<String, Object> parameters);
+    void parseParameters(Map<String, Object> parameters) {
```
heemin32: Shouldn't we make is as constructor?
Setting a value in maxChunkLimit in this method doesn't looks obvious my looking at the method name.

src/main/java/org/opensearch/neuralsearch/processor/chunker/ChunkerFactory.java
```diff
@@ -27,6 +28,15 @@ private ChunkerFactory() {} // no instance of this factory class
         FixedCharLengthChunker::new
     );
 
+    private static final Map<String, Consumer<Map<String, Object>>> CHUNKERS_VALIDATORS = ImmutableMap.of(
```
heemin32: I would rather use polymorphism here. 
1. Enforce every chunker to implement validate method by adding abstract method in parent class.
2. Factory doesn't have to call validation explicitly. I think the validation can happen during instance initiation.

src/main/java/org/opensearch/neuralsearch/mapper/dto/ChunkingConfig.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/processor/chunker/ChunkerFactory.java
bzhangam: We want to do validation without instance initiation because when we validate the index mapping config in the field mapper we don't need to create the actual chunker. Besides to create the actual fixed token length chunker we need `analysisRegistry` which is not available in the field mapper.

Since we want to use the validate functions without creating an instance we need to make it static method and we cannot make an abstract static method.

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
bzhangam: I think it's ok to keep it this way. This is for parsing the common parameters for chunkers.

src/main/java/org/opensearch/neuralsearch/mapper/dto/ChunkingConfig.java
```diff
@@ -0,0 +1,161 @@
...
+    public static final String ALGORITHM_FIELD = "algorithm";
+    /**
+     * parameters parameter name
+     */
+    public static final String PARAMETERS_FIELD = "parameters";
+    private List<Map<String, Object>> configs;
+    private boolean enabled;
+
+    /**
+     * Parse the json input when we define the chunking config in the index mappings
```
heemin32: Please update the comment

src/main/java/org/opensearch/neuralsearch/mapper/dto/ChunkingConfig.java
```diff
@@ -0,0 +1,161 @@
...
+            String.format(Locale.ROOT, "[%s] is required to configure the chunking strategy.", ALGORITHM_FIELD)
+        );
+    }
+
+    /**
+     * Parse the config of the semantic field to build the ChunkingConfig if it is defined. Only should be used
+     * to parse the valid semantic field config.
+     * @param fieldConfig semantic field config
+     */
+    public ChunkingConfig(@NonNull final Map<String, Object> fieldConfig) {
```
heemin32: Maybe move to top as this is constructor now?

src/main/java/org/opensearch/neuralsearch/mapper/dto/ChunkingConfig.java
```diff
@@ -0,0 +1,161 @@
...
+            }
+            throw new MapperParsingException(String.format(Locale.ROOT, "[%s] must be [%s]", ALGORITHM_FIELD, CHUNKER_ALGORITHMS));
+        }
+        throw new MapperParsingException(
+            String.format(Locale.ROOT, "[%s] is required to configure the chunking strategy.", ALGORITHM_FIELD)
+        );
+    }
+
+    /**
+     * Parse the config of the semantic field to build the ChunkingConfig if it is defined. Only should be used
```
heemin32: Update the comment

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
heemin32: Then, I think, the method name should be `void setMaxChunkLimit(Map<String, Object> parameters)`


Also, validation should happen before setting it to variable.
```
int macChunkLimit = parseIntegerWithDefault(parameters, MAX_CHUNK_LIMIT_FIELD, DEFAULT_MAX_CHUNK_LIMIT);
validateMaxChunkLimit(maxChunkLimit);
this.maxChunkLimit = maxChunkLimit;
```

src/main/java/org/opensearch/neuralsearch/processor/chunker/ChunkerFactory.java
heemin32: My main concern is the tight coupling between ChunkerFactory and each individual chunker. Additionally, the validateParameters method is only validated at runtime, which is suboptimal.

The approach I believe would work best includes:
1. When a user adds a new chunker, it should automatically be included in the validation mechanism.

2. Users should be required to implement the `void validateParameters(Map<String, Object> parameters)` method—ideally enforced through an interface or an abstract base class.

In my opinion, the validator logic would be better placed inside ChunkingConfig rather than here. Creating chunking instances shouldn’t pose any issues. Alternatively, we could introduce a new class such as `XChunkerValidator.java` to handle this responsibility as well.

What do you think?

src/main/java/org/opensearch/neuralsearch/mapper/dto/ChunkingConfig.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mapper/dto/ChunkingConfig.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/mapper/dto/ChunkingConfig.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
bzhangam: I think I can have a new method called setMaxChunkLimit to wrap the logic but don't think we should rename the parseParameters itself since we don't know if in future we can have other common parameters. And each chunker also implement this method to parse their own parameters.

src/main/java/org/opensearch/neuralsearch/processor/chunker/ChunkerFactory.java
bzhangam: I think tight coupling between ChunkerFactory and each individual chunker is how factory design pattern works and we cannot decouple them. And I think we have the validation invoked during the chunker creation.

Considering each chunker has its own validation logic I think it makes sense to let each chunker own its validation logic. If we move the logic to the ChunkingConfig then there will be duplicates since in each chunker we still need the validation logic. It may be ok to introduce a util class for the validation logic but in it we still need build the logic for each chunker differently which I don't feel is better than keeping the logic in each chunker.

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
heemin32: If that is the case, we should
```
abstract void doParseParameters(Map<String, Object> parameters);

public void parseParameters(Map<String, Object> parameters) {
  int maxChunkLimit = parseIntegerWithDefault(parameters, MAX_CHUNK_LIMIT_FIELD, DEFAULT_MAX_CHUNK_LIMIT);
  validateMaxChunkLimit(maxChunkLimit);
  this.maxChunkLimit = maxChunkLimit;
  doParseParameters(parameters);
}
```
Otherwise, the child class might forget to set maxChunkLimit by accident?

src/main/java/org/opensearch/neuralsearch/processor/chunker/ChunkerFactory.java
heemin32: Factory class depending on chunker is fine. However, chunker depending on factory is not necessary. With current code, chunker have to implement the validation method with specific signature defined in factory.

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
bzhangam: This is a good idea.

heemin32 (approved):

src/main/java/org/opensearch/neuralsearch/processor/dto/SemanticFieldInfo.java
```diff
@@ -106,4 +116,26 @@ public String getFullPathForEmbeddingInDoc(int index) {
...
+    public void setChunkingConfig(final ChunkingConfig chunkingConfig, @NonNull final AnalysisRegistry analysisRegistry) {
+        if (chunkingConfig == null) {
+            chunkingEnabled = false;
+            return;
+        }
+        chunkingEnabled = chunkingConfig.isEnabled();
+        if (chunkingConfig.getConfigs() != null && chunkingConfig.getConfigs().isEmpty() == false) {
+            chunkers = new ArrayList<>();
+            for (Map<String, Object> config : chunkingConfig.getConfigs()) {
+                final String algorithm = (String) config.get(ALGORITHM_FIELD);
```
junqiu-lei, line 132: Could cast fail if algorithm is null?

src/main/java/org/opensearch/neuralsearch/processor/dto/SemanticFieldInfo.java
```diff
@@ -106,4 +116,26 @@ public String getFullPathForEmbeddingInDoc(int index) {
...
+                final Object paramObject = config.get(PARAMETERS_FIELD);
+                final Map<String, Object> parameters = paramObject == null
+                    ? new HashMap<>()
```
junqiu-lei, line 135:
```suggestion
                    ? Collections.emptyMap()
```

src/main/java/org/opensearch/neuralsearch/mapper/dto/ChunkingConfig.java
```diff
@@ -0,0 +1,173 @@
...
+                    )
+                );
+            }
+            validator.validateParameters((Map<String, Object>) item.get(PARAMETERS_FIELD));
+        } else if (item.get(PARAMETERS_FIELD) != null) {
+            throw new MapperParsingException(String.format(Locale.ROOT, "%s in %s should be a Map.", PARAMETERS_FIELD, CHUNKING));
+        }
+    }
+
+    private void validateNoUnsupportedParameters(@NonNull Map<String, Object> item) {
```
yuye-aws: NoUnsupportedParameters?

src/main/java/org/opensearch/neuralsearch/processor/chunker/Validator.java
```diff
@@ -0,0 +1,18 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.processor.chunker;
+
+import java.util.Map;
+
+/**
+ * An interface to define the behavior of the chunker validator
+ */
+public interface Validator {
```
yuye-aws, line 13: It's good to have a validator interface. Can it be declared as functional interface?

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
```diff
@@ -4,14 +4,19 @@
...
 import java.util.List;
 
+import static org.opensearch.neuralsearch.processor.chunker.ChunkerParameterParser.parseIntegerWithDefault;
+
 /**
  * The interface for all chunking algorithms.
  * All algorithms need to parse parameters and chunk the content.
  */
-public abstract class Chunker {
+public abstract class Chunker implements Validator {
```
yuye-aws: 1. Can you also make it implement a Parser? Or maybe also implement a third interface: Chunker. For the abstract class, you can name it as chunking algorithm
2. Since it is already an abstract, can you use composition with 2 member variables: a parser and a validator. This can avoid duplicate function name like doParseParameters and parseParameters. In doParseParameters, you can call this.parser.parseParameters

src/main/java/org/opensearch/neuralsearch/mapper/dto/ChunkingConfig.java
bzhangam: What do you mean?

src/main/java/org/opensearch/neuralsearch/processor/chunker/Validator.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/processor/dto/SemanticFieldInfo.java
bzhangam: We need to modify it later so cannot use Collections.emptyMap() which is an immutable map.

src/main/java/org/opensearch/neuralsearch/processor/dto/SemanticFieldInfo.java
bzhangam: It should not happen since we will validate the algorithm cannot be null when we construct the ChunkingConfig.

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
bzhangam: To use 2 member variables we need to initiate them in the construct function where we need to do something like
```
public FixedTokenLengthChunker() {
        super(
            FixedTokenLengthChunker::parseFixedParams,
            FixedTokenLengthChunker::validateFixedParams
        );
    }
```

But we want to avoid adding more static methods to make our codes more maintainable. Then we may need to do something like 

```
 public FixedTokenLengthChunker() {
        super(
            // method reference to instance method
            new FixedTokenLengthChunker()::parseInternal,
            new FixedTokenLengthChunker()::validateInternal
        );
    }
```
I think probably it's better to use two abstract method to enforce and reference the implementation in each chunker.
```
protected abstract Parser getParser();
protected abstract Validator getValidator();
```

src/main/java/org/opensearch/neuralsearch/mapper/dto/ChunkingConfig.java
yuye-aws: The function name is confusing. It's actually validating whether parameters are supported. Can you rename to **validateUnsupportedParameters**?

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
```diff
@@ -25,13 +30,25 @@ public abstract class Chunker {
...
+
     /**
-     * Parse the parameters for chunking algorithm.
+     * Parse the common parameters and specific parameters for the chunking algorithm.
      * Throw IllegalArgumentException when parameters are invalid.
      *
      * @param parameters a map containing non-runtime parameters for chunking algorithms
      */
-    abstract void parseParameters(Map<String, Object> parameters);
+    public void parseParameters(Map<String, Object> parameters) {
```
yuye-aws: Declare override?

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
```diff
@@ -56,4 +73,27 @@ static boolean checkRunTimeMaxChunkLimit(int chunkResultSize, int runtimeMaxChun
     }
 
     public abstract String getAlgorithmName();
+
+    /**
+     * Validate the common and the specific parameters for a chunker
+     * @param parameters parameters for a chunker
+     */
+    public void validateParameters(Map<String, Object> parameters) {
```
yuye-aws: Declare override?

src/main/java/org/opensearch/neuralsearch/processor/chunker/Parser.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+@FunctionalInterface
+public interface Parser {
+    /**
+     * Parse the parameters for the chunking algorithm.
+     * Throw IllegalArgumentException when parameters are invalid.
+     *
+     * @param parameters a map containing non-runtime parameters for chunking algorithms
+     */
+    void parseParameters(Map<String, Object> parameters);
```
yuye-aws: nit: parse

src/main/java/org/opensearch/neuralsearch/processor/chunker/Validator.java
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * An interface to define the behavior of the chunker validator
+ */
+@FunctionalInterface
+public interface Validator {
+    /**
+     * Validate the parameters for a chunker
+     * @param parameters parameters for a chunker
+     */
+    void validateParameters(Map<String, Object> parameters);
```
yuye-aws: nit: validate

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
yuye-aws: Using **getParser** and **getValidator** makes sense to me. Just a small of point:
The function names are confusing: **parseParameters** and **doParseParameters**. In final classes, you can change the **doParseParameters** into **parseParameters**, which overrides the abstract class method and calls super.parseParameters() from beginning.

src/main/java/org/opensearch/neuralsearch/mapper/dto/ChunkingConfig.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
bzhangam: ok

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
bzhangam: To do so I think we don't need getParser and getValidator anymore. I'm ok with override but the downside is that we cannot enforce each chunker to override the parse and validate function though we should always do that.

yuye-aws: > To do so I think we don't need getParser and getValidator anymore.

Totally agree.

> I'm ok with override but the downside is that we cannot enforce each chunker to override the parse and validate function though we should always do that.

We can simply add some comments in the chunker abstract class that the child class must call super.parse and super.validate

yuye-aws: I am resolving this conversation now because the code change has been updated

src/main/java/org/opensearch/neuralsearch/mapper/dto/ChunkingConfig.java
```diff
@@ -0,0 +1,173 @@
...
+        if (fieldConfig.containsKey(CHUNKING) == false) {
+            return;
+        }
+        final Object chunkingConfig = fieldConfig.get(CHUNKING);
+        if (chunkingConfig instanceof Boolean) {
+            this.enabled = (Boolean) chunkingConfig;
+            return;
+        }
+        if (chunkingConfig instanceof List<?>) {
+            List<Map<String, Object>> configList = (List<Map<String, Object>>) chunkingConfig;
```
yuye-aws, line 87: Could there be non-map elements in the list

bzhangam: No. The field config of the semantic field is validated when we create it in the index mapping to ensure this.

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
```diff
@@ -56,4 +69,27 @@ static boolean checkRunTimeMaxChunkLimit(int chunkResultSize, int runtimeMaxChun
...
+     * Validate the common parameters for a chunker
+     * @param parameters parameters for a chunker
+     */
+    @Override
+    public void validate(Map<String, Object> parameters) {
+        final int maxChunkLimit = parseIntegerWithDefault(parameters, MAX_CHUNK_LIMIT_FIELD, DEFAULT_MAX_CHUNK_LIMIT);
+        validateMaxChunkLimit(maxChunkLimit);
+    }
+
+    void validateMaxChunkLimit(int maxChunkLimit) {
```
yuye-aws, line 83: nit: Can you declare throw IllegalArgumentException

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
```diff
@@ -25,13 +30,21 @@ public abstract class Chunker {
...
     /**
-     * Parse the parameters for chunking algorithm.
+     * Parse the common parameters for the chunking algorithm.
      * Throw IllegalArgumentException when parameters are invalid.
      *
      * @param parameters a map containing non-runtime parameters for chunking algorithms
      */
-    abstract void parseParameters(Map<String, Object> parameters);
+    @Override
+    public void parse(Map<String, Object> parameters) {
```
yuye-aws, line 43: nit: Can you declare throw IllegalArgumentException

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
```diff
@@ -56,4 +69,27 @@ static boolean checkRunTimeMaxChunkLimit(int chunkResultSize, int runtimeMaxChun
     }
 
     public abstract String getAlgorithmName();
+
+    /**
+     * Validate the common parameters for a chunker
+     * @param parameters parameters for a chunker
+     */
+    @Override
+    public void validate(Map<String, Object> parameters) {
```
yuye-aws, line 78: nit: Can you declare throw IllegalArgumentException

src/main/java/org/opensearch/neuralsearch/processor/chunker/Chunker.java
```diff
@@ -56,4 +69,27 @@ static boolean checkRunTimeMaxChunkLimit(int chunkResultSize, int runtimeMaxChun
     }
 
     public abstract String getAlgorithmName();
+
+    /**
+     * Validate the common parameters for a chunker
+     * @param parameters parameters for a chunker
+     */
```
yuye-aws, line 76: nit: Append 1 lineThrow IllegalArgumentException when parameters are invalid.

src/main/java/org/opensearch/neuralsearch/processor/chunker/Validator.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.Map;
+
+/**
+ * An interface to define the behavior of the chunker validator
+ */
+@FunctionalInterface
+public interface Validator {
+    /**
+     * Validate the parameters for a chunker
+     * @param parameters parameters for a chunker
```
yuye-aws, line 16: nit 1: You can specify the parameter as run-time and non-runtime
nit 2: Throw IllegalArgumentException when parameters are invalid.

src/main/java/org/opensearch/neuralsearch/processor/chunker/Parser.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+@FunctionalInterface
+public interface Parser {
+    /**
+     * Parse the parameters for the chunking algorithm.
+     * Throw IllegalArgumentException when parameters are invalid.
+     *
+     * @param parameters a map containing non-runtime parameters for chunking algorithms
+     */
+    void parse(Map<String, Object> parameters);
```
yuye-aws, line 20: nit: declare throw IllegalArgumentException in method signature

yuye-aws:
Create a simple PR to fix these nit-picking comments: https://github.com/opensearch-project/neural-search/pull/1477

PR #1448, opened 2025-07-15
refactor bwc tests to find model ID in index for semantic field

will-hwang:
### Description
This PR addresses model ID missing in semanticFieldIT by fetching model ID in semantic index. For Semantic Field, model ID is stored in index, and not in pipeline. We had previously created a manual pipeline in SemanticField IT for model ID to be shared across different tests, but it should be fetched from index

### Check List

### DIFF
--- a/qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/SemanticFieldIT.java
+++ b/qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/SemanticFieldIT.java
@@ -12,10 +12,11 @@ import org.opensearch.neuralsearch.query.NeuralQueryBuilder;
 import org.opensearch.neuralsearch.util.TestUtils;
 
 import static org.opensearch.neuralsearch.util.TestUtils.NODES_BWC_CLUSTER;
+import static org.opensearch.neuralsearch.util.TestUtils.getModelId;
 
 public class SemanticFieldIT extends AbstractRestartUpgradeRestTestCase {
-    private static final String PIPELINE_NAME = "nlp-pipeline-semantic-field";
     private static final String TEST_SEMANTIC_TEXT_FIELD = "test_field";
+    private static final String TEST_SEMANTIC_TEXT_FIELD_PATH = "mappings.properties.test_field";
     private static final String SEMANTIC_INFO_FIELD = "semantic_info";
     private static final String SEMANTIC_EMBEDDING_FIELD = "embedding";
     private static final String TEST_QUERY_TEXT = "Hello world";
@@ -43,11 +44,11 @@ public class SemanticFieldIT extends AbstractRestartUpgradeRestTestCase {
                 List.of(SEMANTIC_EMBEDDING_FIELD),
                 List.of(testRankFeaturesDoc)
             );
-            // This is just to block the deletion of the model by other tests
-            createPipelineForSparseEncodingProcessor(modelId, PIPELINE_NAME);
         } else {
             try {
-                loadAndWaitForModelToBeReady(modelId);
+                loadAndWaitForModelToBeReady(
+                    getModelId(getIndexMapping(getIndexNameForTest()), getIndexNameForTest(), TEST_SEMANTIC_TEXT_FIELD_PATH)
+                );
                 addSemanticDoc(
                     getIndexNameForTest(),
                     "1",
@@ -57,7 +58,7 @@ public class SemanticFieldIT extends AbstractRestartUpgradeRestTestCase {
                 );
                 validateTestIndex();
             } finally {
-                wipeOfTestResources(getIndexNameForTest(), PIPELINE_NAME, modelId, null);
+                wipeOfTestResources(getIndexNameForTest(), null, modelId, null);
             }
         }
 
@@ -71,7 +72,7 @@ public class SemanticFieldIT extends AbstractRestartUpgradeRestTestCase {
             .build();
 
         Map<String, Object> searchResponseAsMap = search(getIndexNameForTest(), neuralQueryBuilder, 1);
-        assertEquals(2, getHitCount(searchResponseAsMap));
+        assertEquals(1, getHitCount(searchResponseAsMap));
         Map<String, Object> firstInnerHit = getFirstInnerHit(searchResponseAsMap);
         assertEquals("0", firstInnerHit.get("_id"));
     }
--- a/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/SemanticFieldIT.java
+++ b/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/SemanticFieldIT.java
@@ -11,10 +11,11 @@ import org.opensearch.neuralsearch.query.NeuralQueryBuilder;
 import org.opensearch.neuralsearch.util.TestUtils;
 
 import static org.opensearch.neuralsearch.util.TestUtils.NODES_BWC_CLUSTER;
+import static org.opensearch.neuralsearch.util.TestUtils.getModelId;
 
 public class SemanticFieldIT extends AbstractRollingUpgradeTestCase {
-    private static final String PIPELINE_NAME = "nlp-pipeline-semantic-field";
     private static final String TEST_SEMANTIC_TEXT_FIELD = "test_field";
+    private static final String TEST_SEMANTIC_TEXT_FIELD_PATH = "mappings.properties.test_field";
     private static final String SEMANTIC_INFO_FIELD = "semantic_info";
     private static final String SEMANTIC_EMBEDDING_FIELD = "embedding";
     private static final String TEST_QUERY_TEXT = "Hello world";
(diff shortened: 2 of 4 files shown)


qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/SemanticFieldIT.java
```diff
@@ -57,7 +57,7 @@ public void testSemanticFieldProcessor_E2EFlow() throws Exception {
                 );
                 validateTestIndex();
             } finally {
-                wipeOfTestResources(getIndexNameForTest(), PIPELINE_NAME, modelId, null);
+                wipeOfTestResources(getIndexNameForTest(), null, modelId, null);
```
owaiskazi19, line 61: Why null and not `PIPELINE_NAME` here?

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/SemanticFieldIT.java
```diff
@@ -71,7 +71,6 @@ private void validateTestIndex() {
             .build();
 
         Map<String, Object> searchResponseAsMap = search(getIndexNameForTest(), neuralQueryBuilder, 1);
-        assertEquals(2, getHitCount(searchResponseAsMap));
```
owaiskazi19, line 74: Why are we removing this hit check?

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/SemanticFieldIT.java
```diff
@@ -78,7 +77,7 @@ public void testSemanticFieldProcessor_E2EFlow() throws Exception {
                     );
                     validateTestIndex(totalDocsCountUpgraded);
                 } finally {
-                    wipeOfTestResources(getIndexNameForTest(), PIPELINE_NAME, modelId, null);
+                    wipeOfTestResources(getIndexNameForTest(), null, modelId, null);
```
owaiskazi19, line 82: Same as above

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/SemanticFieldIT.java
```diff
@@ -43,11 +43,11 @@ public void testSemanticFieldProcessor_E2EFlow() throws Exception {
                 List.of(SEMANTIC_EMBEDDING_FIELD),
                 List.of(testRankFeaturesDoc)
             );
-            // This is just to block the deletion of the model by other tests
-            createPipelineForSparseEncodingProcessor(modelId, PIPELINE_NAME);
```
owaiskazi19, line 47: Did we create the pipeline for semantic field before when the pipeline is actually system generated?

will-hwang: yes, this was to enable model sharing across different BWC tests

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/SemanticFieldIT.java
will-hwang: we don't need to delete pipeline since we don't create one anymore

src/testFixtures/java/org/opensearch/neuralsearch/BaseUpgradeTestCase.java
```diff
@@ -96,6 +113,30 @@ protected boolean isPipelineMapContainingModel(final Map<String, Object> pipelin
...
+    @SneakyThrows
+    protected boolean isIndexContainingModel(final Map<String, Object> indexMap, final String modelId) {
+        if (indexMap == null || indexMap.isEmpty()) {
+            return false;
+        }
+
+        for (Object indexObj : indexMap.values()) {
+            if (indexObj instanceof Map) {
+                Map<String, Object> index = (Map<String, Object>) indexObj;
+                String indexStr = index.toString();
```
bzhangam: Since anyway we rely on `index.toString()` to convert the map to string and check I think there is no need to check each value of the `indexMap `. Can we simply do:
```
String indexStr = indexMap.toString();
```

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/SemanticFieldIT.java
will-hwang: this was a test case copied from integ test. since we're ingesting more documents on restart/rolling, this check is not needed. This keeps it consistent with other BWC test cases

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/SemanticFieldIT.java
will-hwang: replied above

src/testFixtures/java/org/opensearch/neuralsearch/BaseUpgradeTestCase.java
will-hwang: yeah, but this is to keep it consistent with check in pipeline: https://github.com/will-hwang/neural-search/blob/2bb58d607073e090969793a9c62c4407faffc078/src/testFixtures/java/org/opensearch/neuralsearch/BaseUpgradeTestCase.java#L99-L113

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/SemanticFieldIT.java
owaiskazi19: Yeah added this comment before [this](https://github.com/opensearch-project/neural-search/pull/1448#discussion_r2208925988) one. Can ignore

src/testFixtures/java/org/opensearch/neuralsearch/BaseUpgradeTestCase.java
owaiskazi19: Can we improve the check in `isPipelineMapContainingModel` as well then?

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/SemanticFieldIT.java
owaiskazi19: Isn't it adding one doc now with addSemanticDoc, so we can we check like
```
 assertEquals(1, getHitCount(searchResponseAsMap));
```

src/testFixtures/java/org/opensearch/neuralsearch/BaseUpgradeTestCase.java
bzhangam: Actually I think we have the string response [here](https://github.com/opensearch-project/neural-search/blob/7a873cb50e00e2b06255fd3a81b6b03eb693807d/src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java#L2181) so can add a new `retrieveIndices()` and `retrievePipelines()` functions to return the string response? In this way we can 1. Simply do a string contain check here. 2. Save the time to convert the string to object and then convert it back to a string.

src/testFixtures/java/org/opensearch/neuralsearch/util/TestUtils.java
```diff
@@ -388,6 +388,16 @@ public static String getModelId(Map<String, Object> pipeline, String processor)
         return modelId;
     }
 
+    public static String getModelId(Map<String, Object> indexMapping, String indexName, String semanticField) {
```
bzhangam: Can we make this method more generic that we can take the path to the semantic field in case in future we have a use case that the semantic field is not at the top level? Besides can you also add some comment to describe the function of this public static method?

bzhangam:
> Why are these changes necessary? As long as the same model ID is used consistently between the semantic field and the pipeline, the source of the model ID doesn't matter. However, if we don't create a pipeline, the shared model may be deleted by other test cases during cleanup. This is because the cleanup process assumes that a model still in use by a pipeline should not be removed. Without a pipeline referencing the model, the test could fail due to the model being unintentionally deleted.
> 
> I think we should take this change. [d7173ec](https://github.com/opensearch-project/neural-search/commit/d7173ec3dcae3b4359ffeb8f35069662298c885b)

I agree that manually creating a pipeline to "hold" the model can work as a workaround, but conceptually it’s not the correct behavior. When a model is used by a semantic field, that usage should be considered a valid dependency — just like being referenced in a pipeline.

In the long term, instead of relying on pipelines to protect the model from deletion, we should enhance the cleanup logic to check whether the model is in use by any semantic field. That would make the model lifecycle management more accurate and aligned with actual usage patterns.

So while I’m okay with this change as a short-term fix, I think it's better to have this long term fix.

src/testFixtures/java/org/opensearch/neuralsearch/BaseUpgradeTestCase.java
will-hwang: can we just do .string() on existing retrieveIndices and retrievePipelines? .toString() is a cheap opereration, and adding additional methods to get response as string is somewhat redundant

bzhangam: I'm ok with it since this is just for testing.

bzhangam (approved):

PR #1414, opened 2025-06-27
Fix the HybridQueryDocIdStream to properly handle upTo value

heemin32:
### Description
The HybridQueryDocIdStream class contained a bug in how it handled the upTo parameter. This issue led to a flaky test failure in HybridQueryTests.testWithRandomDocuments_whenMultipleTermSubQueriesWithMatch_thenReturnSuccessfully.

The failure occurred because the mayHaveRemaining method returned true even after all documents had been consumed from the hybridBulkScorer. As a result, OpenSearch Core attempted to call the forEach method again, which reconsumed documents from the beginning—triggering an assertion failure.

This PR addresses the issue by properly clearing document ID which was already consumed, ensuring that documents aren't reprocessed. It also fixes the mayHaveRemaining method to return false if there is no more document to process.

### Related Issues
Resolves #1344 

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [X] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/neural-search/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1433, opened 2025-07-09
Eliminate non-deterministic behavior in scoring for the entire unit test

heemin32:
### Description
The tests in the neural plugin rely on the document scoring for a given text. If this scoring is non-deterministic, it increases the likelihood of flaky test failures. To address this, this PR replace `newSearcher(reader)` with `new IndexSearcher(reader)`, ensuring more consistent scoring behavior. Additionally, the PR removed some code that had no impact on code coverage.

### Related Issues
N/A

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [X] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/neural-search/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #1344, opened 2025-05-22
[Flaky Test] HybridQueryTests.testWithRandomDocuments_whenMultipleTermSubQueriesWithMatch_thenReturnSuccessfully


ryanbogan:
https://github.com/opensearch-project/neural-search/pull/1336 introduced a flaky test org.opensearch.neuralsearch.query.HybridQueryTests.testWithRandomDocuments_whenMultipleTermSubQueriesWithMatch_thenReturnSuccessfully.

The test needs be investigated and fixed so it is not flaky.

heemin32:
```
REPRODUCE WITH: gradlew ':test' --tests "org.opensearch.neuralsearch.query.HybridQueryTests.testWithRandomDocuments_whenMultipleTermSubQueriesWithMatch_thenReturnSuccessfully" -Dtests.seed=D6D9F52CDEBD6AE7 -Dtests.security.manager=false -Dtests.locale=pt-MO -Dtests.timezone=Australia/Darwin -Druntime.java=23
HybridQueryTests > testWithRandomDocuments_whenMultipleTermSubQueriesWithMatch_thenReturnSuccessfully FAILED
    java.lang.AssertionError
        at __randomizedtesting.SeedInfo.seed([D6D9F52CDEBD6AE7:9F6ADF20695CBC7C]:0)
        at org.apache.lucene.tests.search.AssertingLeafCollector$AssertingDocIdStream.forEach(AssertingLeafCollector.java:141)
        at org.apache.lucene.tests.search.AssertingLeafCollector$AssertingDocIdStream.forEach(AssertingLeafCollector.java:132)
        at org.apache.lucene.tests.search.AssertingLeafCollector$AssertingDocIdStream.forEach(AssertingLeafCollector.java:132)
        at org.apache.lucene.search.LeafCollector.collect(LeafCollector.java:106)
        at org.apache.lucene.tests.search.AssertingLeafCollector.collect(AssertingLeafCollector.java:50)
        at org.apache.lucene.tests.search.AssertingLeafCollector.collect(AssertingLeafCollector.java:50)
        at org.apache.lucene.tests.search.AssertingLeafCollector.collect(AssertingLeafCollector.java:50)
        at org.opensearch.neuralsearch.query.HybridBulkScorer.scoreWindowIntoBitSetWithSubqueryScorers(HybridBulkScorer.java:150)
[...]

Issue #1278, opened 2025-04-21
[Feature] Framework to check the impact of a new commit in the repo on query performance (Neural Query, Hybrid Query, Neural Sparse Query)


vibrantvarun:
### Is your feature request related to a problem?
With the fast evolving search space and new commits going in the repository, there is a need of creating a framework/script to validate the impact of every new commit on the query performance. It will help the team in following ways:
1. Saves from performance regression at an early stage
2. helps in debugging the faulty code 
3. Saves the efforts of doing performance benchmarking.

### What solution would you like?
Solution 1
A script that can be ran has GH workflow on every PR.

Solution 2
Nightly benchmarks. The current of doing it is we need to raise a PR by in opensearch-benchmark workloads and check the query performance. We need a better way of doing it as the raising a PR on benchmarks and then setting it up for nightlies is a time consuming process.

vibrantvarun:
> Question about Solution 1: Would it be possible to utilize the opensearch-benchmark framework within GitHub actions? Or would it be more similar to an integration test action that spins up a cluster and then runs a series of API calls on the cluster to measure benchmarks?

We need to do some deep-dive on how exactly it can be done. I mentioned just at high level what is the need here.

weijia-aws:
> there is a need of creating a framework/script to validate the impact of every new commit on the query performance and recall

Recall metric is only available for vector search, not neural

vibrantvarun:
> > there is a need of creating a framework/script to validate the impact of every new commit on the query performance and recall
> 
> Recall metric is only available for vector search, not neural

Then we need make it available. Because neural is essentally a knn query at low level.

vibrantvarun:
Running before can be capture with the last merged PR right. So it would be good if only after a commit is pushed we should run it.

will-hwang:
In my opinion, not all change requires benchmarking. Running benchmarking on every commit would be unnecessary. We should have a tag for changes that require benchmarking, and only trigger benchmarking for the tagged PRs

vibrantvarun:
Agree with @will-hwang

owaiskazi19:
We can follow the same framework like core [here](https://github.com/opensearch-project/OpenSearch/pull/17970#issuecomment-2810193359) to run the benchmarks only on the performance related commit PR.

vibrantvarun:
That's nice suggestion @owaiskazi19

junqiu-lei:
[...]

PR #1457, opened 2025-07-18
Fixed how bitset works for custom bulk scorer in hybrid query

martin-gaievski:
### Description
Relevancy of final search results dropped significantly, see below examples for [Quora](https://quoradata.quora.com/First-Quora-Dataset-Release-Question-Pairs) dataset, where NDCG and other metrics dropped 2x:

```
version 3.1
2025-07-17 17:05:42 - NDCG_mean: 0.8277
2025-07-17 17:05:42 - NDCG_median: 0.8418

current main
2025-07-17 17:16:01 - NDCG_mean: 0.3698
2025-07-17 17:16:01 - NDCG_median: 0.3704
```

How HybridBulkScorer Works

1. The `HybridBulkScorer` processes documents in windows of 4096 documents (WINDOW_SIZE = 1 << 12)
2. For each window:
   - It populates a `matching` bitset with documents that match any sub-query
   - It fills `windowScores` with scores for matching documents
   - It calls `collector.collect(hybridQueryDocIdStream)` to process the window
   - It then **clears the matching bitset** for the next window.
 
PR #1414 introduced a local copy of the matching bitset in `HybridQueryDocIdStream`:

```java
private FixedBitSet getLocalMatchingBitSet() {
    if (localMatchingBitSet == null) {
        localMatchingBitSet = hybridBulkScorer.getMatching().clone();
    }
    return localMatchingBitSet;
}
```

This clones the bitset only once, but the `HybridBulkScorer` reuses the same `HybridQueryDocIdStream` instance across multiple windows. This can be illustrated with following example:

1. **Window 1** (docs 0-4095):
   - HybridBulkScorer populates matching bitset with docs in window 1
   - HybridQueryDocIdStream clones this bitset (first and only time)
   - Documents are processed and cleared from the local copy
   - HybridBulkScorer clears its matching bitset

2. **Window 2** (docs 4096-8191):
   - HybridBulkScorer populates matching bitset with docs in window 2
   - HybridQueryDocIdStream still uses the stale clone from window 1
   - Since all bits were cleared in window 1, no documents from window 2+ are processed
   - **All documents in windows 2, 3, 4... are missed**

While this condition was logically not correct (adding base to upTo), it accidentally allowed more documents to be processed because:
- when `base = 0`, it worked correctly
- when `base > 0`, the condition `doc < upTo + base` was often true even for docs that should have been excluded
- this masked the issue and maintained reasonable relevance scores

Proper fix needs to:
1. Remove the local caching of the matching bitset
2. Always use the current state of the HybridBulkScorer's matching bitset
3. Properly handle the `upTo` parameter without early returns

### Related Issues
Spotted during internal testing

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridQueryDocIdStream.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridQueryDocIdStream.java
@@ -25,11 +25,6 @@ public class HybridQueryDocIdStream extends DocIdStream {
     private int base;
 
     @Override
-    public boolean mayHaveRemaining() {
-        return getLocalMatchingBitSet().cardinality() > 0;
-    }
-
-    @Override
     public int count(int upTo) {
         // Use a counter to track how many documents are processed
         final int[] count = { 0 };
@@ -43,6 +38,12 @@ public class HybridQueryDocIdStream extends DocIdStream {
     }
 
     @Override
+    public boolean mayHaveRemaining() {
+        return false;
+    }
+
+    // This class does not respect the upTo value; it consumes all matching documents.
+    @Override
     public void forEach(int upTo, CheckedIntConsumer<IOException> consumer) throws IOException {
         // Always get the current matching bitset from the bulk scorer
         FixedBitSet matchingBitSet = hybridBulkScorer.getMatching();
@@ -55,21 +56,17 @@ public class HybridQueryDocIdStream extends DocIdStream {
                 final int docIndexInWindow = (idx << BLOCK_SHIFT) | numberOfTrailingZeros;
                 final int docId = base | docIndexInWindow;
 
-                // Only process documents < upTo
-                if (docId < upTo) {
-                    float[][] windowScores = hybridBulkScorer.getWindowScores();
-                    for (int subQueryIndex = 0; subQueryIndex < windowScores.length; subQueryIndex++) {
-                        if (Objects.isNull(windowScores[subQueryIndex])) {
-                            continue;
-                        }
-                        float scoreOfDocIdForSubQuery = windowScores[subQueryIndex][docIndexInWindow];
-                        hybridBulkScorer.getHybridSubQueryScorer().getSubQueryScores()[subQueryIndex] = scoreOfDocIdForSubQuery;
+                float[][] windowScores = hybridBulkScorer.getWindowScores();
+                for (int subQueryIndex = 0; subQueryIndex < windowScores.length; subQueryIndex++) {
+                    if (Objects.isNull(windowScores[subQueryIndex])) {
+                        continue;
                     }
-                    consumer.accept(docId);
-                    hybridBulkScorer.getHybridSubQueryScorer().resetScores();
+                    float scoreOfDocIdForSubQuery = windowScores[subQueryIndex][docIndexInWindow];
+                    hybridBulkScorer.getHybridSubQueryScorer().getSubQueryScores()[subQueryIndex] = scoreOfDocIdForSubQuery;
                 }
+                consumer.accept(docId);
+                hybridBulkScorer.getHybridSubQueryScorer().resetScores();
 
-                // Don't clear bits from the shared bitset!
                 bits ^= 1L << numberOfTrailingZeros;
             }
         }
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridQueryDocIdStream.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridQueryDocIdStream.java
@@ -44,41 +44,32 @@ public class HybridQueryDocIdStream extends DocIdStream {
 
     @Override
     public void forEach(int upTo, CheckedIntConsumer<IOException> consumer) throws IOException {
-        // bitset that represents matching documents, bit is set (1) if doc id is a match
-        FixedBitSet matchingBitSet = getLocalMatchingBitSet();
+        // Always get the current matching bitset from the bulk scorer
+        FixedBitSet matchingBitSet = hybridBulkScorer.getMatching();
         long[] bitArray = matchingBitSet.getBits();
-        // iterate through each block of 64 documents (since each long contains 64 bits)
+
         for (int idx = 0; idx < bitArray.length; idx++) {
             long bits = bitArray[idx];
             while (bits != 0L) {
-                // find position of the rightmost set bit (1)
                 int numberOfTrailingZeros = Long.numberOfTrailingZeros(bits);
-                // calculate actual document ID within the window
-                // idx << 6 is equivalent to idx * 64 (block offset)
-                // numberOfTrailingZeros gives position within the block
                 final int docIndexInWindow = (idx << BLOCK_SHIFT) | numberOfTrailingZeros;
                 final int docId = base | docIndexInWindow;
 
-                // Only process documents up to the specified limit
-                if (docId >= upTo) {
-                    return;
+                // Only process documents < upTo
+                if (docId < upTo) {
+                    float[][] windowScores = hybridBulkScorer.getWindowScores();
+                    for (int subQueryIndex = 0; subQueryIndex < windowScores.length; subQueryIndex++) {
+                        if (Objects.isNull(windowScores[subQueryIndex])) {
+                            continue;
+                        }
+                        float scoreOfDocIdForSubQuery = windowScores[subQueryIndex][docIndexInWindow];
+                        hybridBulkScorer.getHybridSubQueryScorer().getSubQueryScores()[subQueryIndex] = scoreOfDocIdForSubQuery;
+                    }
+                    consumer.accept(docId);
+                    hybridBulkScorer.getHybridSubQueryScorer().resetScores();
                 }
 
-                float[][] windowScores = hybridBulkScorer.getWindowScores();
-                for (int subQueryIndex = 0; subQueryIndex < windowScores.length; subQueryIndex++) {
-                    if (Objects.isNull(windowScores[subQueryIndex])) {
-                        continue;
-                    }
-                    float scoreOfDocIdForSubQuery = windowScores[subQueryIndex][docIndexInWindow];
-                    hybridBulkScorer.getHybridSubQueryScorer().getSubQueryScores()[subQueryIndex] = scoreOfDocIdForSubQuery;
-                }
-                // process the document with its base offset
-                consumer.accept(docId);
-                // reset scores after processing of one doc, this is required because scorer object is re-used
-                hybridBulkScorer.getHybridSubQueryScorer().resetScores();
-                // clear bit from the local bitset copy to indicate that it has been consumed
-                matchingBitSet.clear(docIndexInWindow);
-                // reset bit for this doc id to indicate that it has been consumed
+                // Don't clear bits from the shared bitset!
                 bits ^= 1L << numberOfTrailingZeros;
             }
         }
--- a/src/test/java/org/opensearch/neuralsearch/query/HybridQueryDocIdStreamTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/query/HybridQueryDocIdStreamTests.java
@@ -122,7 +122,7 @@ public class HybridQueryDocIdStreamTests extends OpenSearchTestCase {
     }
 
     @SneakyThrows
-    public void testForEach_whenSubsequentCalls_thenUpToParameterRespected() {
+    public void testForEach_whenSubsequentCalls_thenUpToParameterIgnored() {
         // setup
         FixedBitSet matchingDocs = new FixedBitSet(NUM_DOCS);
         matchingDocs.set(DOC_ID_1); // docId = 1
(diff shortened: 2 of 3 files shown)


heemin32:
The current implementation is not correct either regarding adherence to upTo value.
If the method called twice, for the second call, the document consumed previously should not be consumed again. Otherwise, it will throw an exception and that was the reason of flaky test failure before.

For the testing, maybe we could run both boolean query and hybrid query(make it equivalent with boolean query) with 100,000 documents(larger than window size), then, it could catch this kind of issue?

martin-gaievski:
> The current implementation does not correctly adhere to the upTo value. If the method is called twice, documents consumed during the first call should not be processed again in the second. Failing to do so can lead to exceptions, which previously caused flaky test failures.

That depends on definition on correctness, we need to focus on scenario that is realistic for production OpenSeach cluster. For sure test can be constructed in a way that method is called twice, but is it possible for OpenSearch? 
What exact flaky test do you mean?  You have addressed quite a lot of them, is this the one: https://github.com/opensearch-project/neural-search/issues/1344, and it has been addressed in https://github.com/opensearch-project/neural-search/pull/1433, or it's something else? If you assumption is correct that test became flaky again. 

> For testing, we could consider running both a Boolean query and an equivalent Hybrid query on a dataset of 100,000 documents (larger than the window size) and check if the result is same between the two. This might help uncover issues like this.

Such comparison can cover certain issues like dropped recall. But this will not cover problems with NDCG, where position of the document matters.  
Ingesting 100K of docs can be challenging for integ test. For search-relevance I've build test with 1,200 docs (those a big docs from ESCI dataset, but still only a bit more then 1K) and whole CI became slow. I think realistically we can ingest maybe up to 10K of relatively simple docs with 2-3 fields and values with length of 100-200.

owaiskazi19:
If we are can integrate relevancy with the framework mentioned [here](https://github.com/opensearch-project/neural-search/issues/1278) which core already support as of now can solve our issue rather having a big integ test with 10k+ docs. We can run the framework only on the commit where we need the relevancy difference.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryDocIdStream.java
```diff
@@ -44,41 +44,32 @@ public int count(int upTo) {
...
-                if (docId >= upTo) {
-                    return;
-                }
-
-                float[][] windowScores = hybridBulkScorer.getWindowScores();
-                for (int subQueryIndex = 0; subQueryIndex < windowScores.length; subQueryIndex++) {
-                    if (Objects.isNull(windowScores[subQueryIndex])) {
-                        continue;
+                // Only process documents < upTo
+                if (docId < upTo) {
```
owaiskazi19: Should we make it inclusive?
```suggestion
                if (docId <= upTo) {
```

src/main/java/org/opensearch/neuralsearch/query/HybridQueryDocIdStream.java
```diff
@@ -44,41 +44,32 @@ public int count(int upTo) {
...
                 }
-                // process the document with its base offset
-                consumer.accept(docId);
-                // reset scores after processing of one doc, this is required because scorer object is re-used
-                hybridBulkScorer.getHybridSubQueryScorer().resetScores();
-                // clear bit from the local bitset copy to indicate that it has been consumed
-                matchingBitSet.clear(docIndexInWindow);
-                // reset bit for this doc id to indicate that it has been consumed
+
+                // Don't clear bits from the shared bitset!
```
owaiskazi19: Since clearing the bits didn't let us process the next set of documents, should we have a local and global copy of bitset? Local will be the current window and global would be for all rather than working on just one

owaiskazi19 (commented):
Thanks for deep diving on this @martin-gaievski!

src/main/java/org/opensearch/neuralsearch/query/HybridQueryDocIdStream.java
martin-gaievski: this should be exclusive, that's the Lucene contract we need to follow https://github.com/apache/lucene/blob/branch_10_2/lucene/core/src/java/org/apache/lucene/search/DocIdStream.java

src/main/java/org/opensearch/neuralsearch/query/HybridQueryDocIdStream.java
martin-gaievski: I think having a global copy will have few drawbacks:
- extra resource consumption, currently resources are freed after each window has been processed, with global state we have to maintain it for duration of the query
- concurrency issues like race conditions and synchronization overhead, if we start synchronize on bitset state we loose the benefit of parallelism and thus the gain in response times
I do like a mental model of a streaming video, you don't load entire clip into memory, but you process it frame by frame, where each frame is independent.

owaiskazi19 (approved):

heemin32:
> That depends on definition on correctness, we need to focus on scenario that is realistic for production OpenSeach cluster. For sure test can be constructed in a way that method is called twice, but is it possible for OpenSearch?
What exact flaky test do you mean? You have addressed quite a lot of them, is this the one: https://github.com/opensearch-project/neural-search/issues/1344, and it has been addressed in https://github.com/opensearch-project/neural-search/pull/1433, or it's something else? If you assumption is correct that test became flaky again.

This test failure will appear again. https://github.com/opensearch-project/neural-search/issues/1344
If we don't assume that the method won't be called twice, we don't even need this check
`if (docId < upTo) {`

> Such comparison can cover certain issues like dropped recall. But this will not cover problems with NDCG, where position of the document matters.
Ingesting 100K of docs can be challenging for integ test. For search-relevance I've build test with 1,200 docs (those a big docs from ESCI dataset, but still only a bit more then 1K) and whole CI became slow. I think realistically we can ingest maybe up to 10K of relatively simple docs with 2-3 fields and values with length of 100-200.

It was typo. I was going to talk about 10K because window size is 4096. Very basic mistake like this should be able to caught by comparing result between boolean and hybrid.

martin-gaievski:
I've added new integ test that does that type of check, it's passing with this PR, and is failing with the current implementation from main which is expected. Thanks for idea @heemin32, this is indeed much simpler then full scale test for relevance metrics.

```
 2> REPRODUCE WITH: ./gradlew ':integTest' --tests 'org.opensearch.neuralsearch.query.HybridQueryWindowBoundaryIT.testHybridQuery_whenResultsSpanMultipleWindows_thenNoDocumentsDropped' -Dtests.seed=62F16C6350F3F9F8 -Dtests.security.manager=false -Dtests.locale=es-Latn-ES -Dtests.timezone=Asia/Bahrain -Druntime.java=21
  2> java.lang.AssertionError: expected:<10000> but was:<7981>
        at __randomizedtesting.SeedInfo.seed([62F16C6350F3F9F8:8EDBFE3036725A70]:0)
        at org.junit.Assert.assertEquals(Assert.java:647)
        at org.junit.Assert.assertEquals(Assert.java:633)
        at org.opensearch.neuralsearch.query.HybridQueryWindowBoundaryIT.testHybridQuery_whenResultsSpanMultipleWindows_thenNoDocumentsDropped(HybridQueryWindowBoundaryIT.java:73)
```

With current implementation in main branch behavior is not deterministic, here are results from 10 consecutive runs of this test:

```
| Run # | Documents Collected | 
|-------|---------------------|
| 1     | 7,901              | 
| 2     | 7,900              | 
| 3     | 8,118              | 
| 4     | 7,753              | 
| 5     | 7,737              | 
| 6     | 7,496              | 
| 7     | 7,443              |
| 8     | 7,750              |
| 9     | 7,721              | 
| 10    | 7,817              |
```  

I think we can let the code live in the branch to see if we have any new occurrence of flaky tests from https://github.com/opensearch-project/neural-search/issues/1344. So far I haven't seen any, we can reiterate on fix if we met one.

heemin32:
Could you make two changes and see if any of the test fails? I think current implementation does not adhere to upTo value anyway so we should just do like

```
// This class does not respect the upTo value; it consumes all matching documents.
// Therefore, always return false for `mayHaveRemaining`.
@Override
public boolean mayHaveRemaining() {
    return false;
}

// This class does not respect the upTo value; it consumes all matching documents.
@Override
public void forEach(int upTo, CheckedIntConsumer<IOException> consumer) throws IOException {
    // Always get the current matching bitset from the bulk scorer
    FixedBitSet matchingBitSet = hybridBulkScorer.getMatching();
    long[] bitArray = matchingBitSet.getBits();

    for (int idx = 0; idx < bitArray.length; idx++) {
        long bits = bitArray[idx];
        while (bits != 0L) {
            int numberOfTrailingZeros = Long.numberOfTrailingZeros(bits);
            final int docIndexInWindow = (idx << BLOCK_SHIFT) | numberOfTrailingZeros;
            final int docId = base | docIndexInWindow;

            float[][] windowScores = hybridBulkScorer.getWindowScores();
            for (int subQueryIndex = 0; subQueryIndex < windowScores.length; subQueryIndex++) {
                if (Objects.isNull(windowScores[subQueryIndex])) {
                    continue;
                }
                float scoreOfDocIdForSubQuery = windowScores[subQueryIndex][docIndexInWindow];
                hybridBulkScorer.getHybridSubQueryScorer().getSubQueryScores()[subQueryIndex] = scoreOfDocIdForSubQuery;
            }
            consumer.accept(docId);
            hybridBulkScorer.getHybridSubQueryScorer().resetScores();

            bits ^= 1L << numberOfTrailingZeros;
        }
    }
}

```

martin-gaievski:
After making this change (ignoring upTo) all tests are passing except for [one unit test](https://github.com/opensearch-project/neural-search/blob/main/src/test/java/org/opensearch/neuralsearch/query/HybridQueryDocIdStreamTests.java#L125-L156) that specifically checking if upTo value is respected , which is expected after we made this new change. If I change test to following state it's passing:

```
    public void testForEach_whenSubsequentCalls_thenAllDocsProcessedProperly() {
        // setup
        FixedBitSet matchingDocs = new FixedBitSet(NUM_DOCS);
        matchingDocs.set(DOC_ID_1); // docId = 1
        matchingDocs.set(DOC_ID_2); // docId = 2
        matchingDocs.set(DOC_ID_3); // docId = 3

        HybridBulkScorer mockScorer = createMockScorerWithDocs(matchingDocs);
        HybridQueryDocIdStream stream = new HybridQueryDocIdStream(mockScorer);
        List<Integer> processedDocs = new ArrayList<>();

        // Note: HybridQueryDocIdStream does not respect the upTo parameter and processes all matching documents

        // first call with upTo = 2 (will process all matching docs despite upTo value)
        stream.forEach(2, docId -> processedDocs.add(docId));

        // verify first call results - all 3 docs are processed (upTo is ignored)
        assertEquals(3, processedDocs.size());
        assertTrue(processedDocs.contains(DOC_ID_1));
        assertTrue(processedDocs.contains(DOC_ID_2));
        assertTrue(processedDocs.contains(DOC_ID_3));

        // clear processed docs list
        processedDocs.clear();

        // second call with upTo = 4 (will process all matching docs)
        stream.forEach(4, docId -> processedDocs.add(docId));

        // verify second call results - all docs are processed
        assertEquals(3, processedDocs.size());
        assertTrue(processedDocs.contains(DOC_ID_1));
        assertTrue(processedDocs.contains(DOC_ID_2));
        assertTrue(processedDocs.contains(DOC_ID_3));

        // clear processed docs list
        processedDocs.clear();

        // third call with upTo = 1 (will still process all documents since upTo is ignored)
        stream.forEach(1, docId -> processedDocs.add(docId));
        assertEquals(3, processedDocs.size());
        assertTrue(processedDocs.contains(DOC_ID_1));
        assertTrue(processedDocs.contains(DOC_ID_2));
        assertTrue(processedDocs.contains(DOC_ID_3));
    }
```

martin-gaievski:
I've committed change where upTo check is no longer present, also modifies corresponding unit test so it's not expecting that.

Checked relevance score for Quora dataset, it's still correct with this latest change

```
2025-07-21 12:16:51 - NDCG@1: 0.7551
2025-07-21 12:16:51 - NDCG@3: 0.8014
2025-07-21 12:16:51 - NDCG@5: 0.8208
2025-07-21 12:16:51 - NDCG@10: 0.8365
2025-07-21 12:16:51 - NDCG@25: 0.8475
2025-07-21 12:16:51 - NDCG@50: 0.8521
2025-07-21 12:16:51 - NDCG@100: 0.8544
2025-07-21 12:16:51 - NDCG@200: 0.8555
2025-07-21 12:16:51 - NDCG_mean: 0.8279
2025-07-21 12:16:51 - NDCG_median: 0.8420
2025-07-21 12:16:51 -

2025-07-21 12:16:51 - MAP@1: 0.6565
2025-07-21 12:16:51 - MAP@3: 0.7603
2025-07-21 12:16:51 - MAP@5: 0.7792
2025-07-21 12:16:51 - MAP@10: 0.7904
2025-07-21 12:16:51 - MAP@25: 0.7958
2025-07-21 12:16:51 - MAP@50: 0.7973
2025-07-21 12:16:51 - MAP@100: 0.7977
2025-07-21 12:16:51 - MAP@200: 0.7979
2025-07-21 12:16:51 - MAP_mean: 0.7719
2025-07-21 12:16:51 - MAP_median: 0.7931
2025-07-21 12:16:51 -

2025-07-21 12:16:51 - Recall@1: 0.6565
2025-07-21 12:16:51 - Recall@3: 0.8263
2025-07-21 12:16:51 - Recall@5: 0.8796
2025-07-21 12:16:51 - Recall@10: 0.9256
2025-07-21 12:16:51 - Recall@25: 0.9622
2025-07-21 12:16:51 - Recall@50: 0.9796
2025-07-21 12:16:51 - Recall@100: 0.9895
2025-07-21 12:16:51 - Recall@200: 0.9943
2025-07-21 12:16:51 - Recall_mean: 0.9017
2025-07-21 12:16:51 - Recall_median: 0.9439
2025-07-21 12:16:51 -

2025-07-21 12:16:51 - P@1: 0.7551
2025-07-21 12:16:51 - P@3: 0.3479
2025-07-21 12:16:51 - P@5: 0.2304
2025-07-21 12:16:51 - P@10: 0.1265
2025-07-21 12:16:51 - P@25: 0.0551
2025-07-21 12:16:51 - P@50: 0.0289
2025-07-21 12:16:51 - P@100: 0.0149
2025-07-21 12:16:51 - P@200: 0.0076
2025-07-21 12:16:51 - P_mean: 0.1958
2025-07-21 12:16:51 - P_median: 0.0908
```

heemin32 (approved):
LGTM. Thanks!

junqiu-lei (approved):
LGTM, thanks Martin.

Issue #1416, opened 2025-06-27
[Flaky test]rolling.NeuralSparseSearchIT.testSparseEncodingProcessor_E2EFlow


heemin32:
```
REPRODUCE WITH: ./gradlew ':qa:rolling-upgrade:testAgainstTwoThirdsUpgradedCluster' --tests "org.opensearch.neuralsearch.bwc.rolling.NeuralSparseSearchIT.testSparseEncodingProcessor_E2EFlow" -Dtests.seed=E42B39786EA922E1 -Dtests.security.manager=false -Dtests.bwc.version=3.0.0 -Dtests.locale=ff-Adlm-NE -Dtests.timezone=BET -Druntime.java=23


Suite: Test class org.opensearch.neuralsearch.bwc.rolling.NeuralSparseSearchIT
  2> REPRODUCE WITH: ./gradlew ':qa:rolling-upgrade:testAgainstTwoThirdsUpgradedCluster' --tests "org.opensearch.neuralsearch.bwc.rolling.NeuralSparseSearchIT.testSparseEncodingProcessor_E2EFlow" -Dtests.seed=E42B39786EA922E1 -Dtests.security.manager=false -Dtests.bwc.version=3.0.0 -Dtests.locale=ff-Adlm-NE -Dtests.timezone=BET -Druntime.java=23
  2> org.junit.ComparisonFailure: expected:<[0]> but was:<[1]>
        at __randomizedtesting.SeedInfo.seed([E42B39786EA922E1:2617D9E173359055]:0)
        at org.junit.Assert.assertEquals(Assert.java:117)
        at org.junit.Assert.assertEquals(Assert.java:146)
        at org.opensearch.neuralsearch.bwc.rolling.NeuralSparseSearchIT.validateTestIndexOnUpgrade(NeuralSparseSearchIT.java:115)
        at org.opensearch.neuralsearch.bwc.rolling.NeuralSparseSearchIT.testSparseEncodingProcessor_E2EFlow(NeuralSparseSearchIT.java:77)
  2> NOTE: leaving temporary files on disk at: /home/runner/work/neural-search/neural-search/qa/rolling-upgrade/build/testrun/testAgainstTwoThirdsUpgradedCluster/temp/org.opensearch.neuralsearch.bwc.rolling.NeuralSparseSearchIT_E42B39786EA922E1-001
[...]

PR #1465, opened 2025-07-18
Fix flaky testSparseEncodingProcessor_E2EFlow and yellow cluster issue

owaiskazi19:
### Description
Trying fix for `testSparseEncodingProcessor_E2EFlow` flkay test.
The test unconditionally expects document ID "0" to be the top hit. During the MIXED phase, after adding document "1", this document appears to rank higher than document "0"

### Related Issues
Resolves https://github.com/opensearch-project/neural-search/issues/1416 and https://github.com/opensearch-project/neural-search/issues/1468
### Check List

### DIFF
--- a/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/BatchIngestionIT.java
+++ b/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/BatchIngestionIT.java
@@ -22,7 +22,7 @@ public class BatchIngestionIT extends AbstractRollingUpgradeTestCase {
     private static final String EMBEDDING_FIELD_NAME = "passage_embedding";
 
     public void testBatchIngestion_SparseEncodingProcessor_E2EFlow() throws Exception {
-        waitForClusterHealthGreen(NODES_BWC_CLUSTER, 120);
+        waitForClusterHealthGreenOrYellow(NODES_BWC_CLUSTER);
         String indexName = getIndexNameForTest();
         String sparseModelId = null;
         switch (getClusterType()) {
--- a/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/BatchIngestionIT.java
+++ b/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/BatchIngestionIT.java
@@ -22,7 +22,7 @@ public class BatchIngestionIT extends AbstractRollingUpgradeTestCase {
     private static final String EMBEDDING_FIELD_NAME = "passage_embedding";
 
     public void testBatchIngestion_SparseEncodingProcessor_E2EFlow() throws Exception {
-        waitForClusterHealthGreen(NODES_BWC_CLUSTER, 90);
+        waitForClusterHealthGreen(NODES_BWC_CLUSTER, 120);
         String indexName = getIndexNameForTest();
         String sparseModelId = null;
         switch (getClusterType()) {
--- a/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/NeuralSparseSearchIT.java
+++ b/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/NeuralSparseSearchIT.java
@@ -39,7 +39,7 @@ public class NeuralSparseSearchIT extends AbstractRollingUpgradeTestCase {
     // Create Sparse Encoding Processor, Ingestion Pipeline and add document
     // Validate process , pipeline and document count in rolling-upgrade scenario
     public void testSparseEncodingProcessor_E2EFlow() throws Exception {
-        waitForClusterHealthGreen(NODES_BWC_CLUSTER, 120);
+        waitForClusterHealthGreenOrYellow(NODES_BWC_CLUSTER);
         switch (getClusterType()) {
             case OLD:
                 modelId = uploadSparseEncodingModel();
--- a/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/NeuralSparseSearchIT.java
+++ b/qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/NeuralSparseSearchIT.java
@@ -10,6 +10,7 @@ import java.util.List;
 import java.util.Map;
 import org.opensearch.index.query.BoolQueryBuilder;
 import org.opensearch.index.query.MatchQueryBuilder;
+import org.opensearch.index.query.TermQueryBuilder;
 import org.opensearch.neuralsearch.util.TestUtils;
 import static org.opensearch.neuralsearch.util.TestUtils.NODES_BWC_CLUSTER;
 import static org.opensearch.neuralsearch.util.TestUtils.SPARSE_ENCODING_PROCESSOR;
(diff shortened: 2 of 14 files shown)


owaiskazi19:
@heemin32 @martin-gaievski should we accept the cluster state as yellow for rolling upgrade as one shard sometimes get stays unassigned?

heemin32:
I triggered retry. 
```
The test unconditionally expects document ID "0" to be the top hit. During the MIXED phase, after adding document "1", this document appears to rank higher than document "0"
```
Do we know why it is passing sometimes and not passing other times?

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/NeuralSparseSearchIT.java
```diff
@@ -103,12 +104,33 @@ public void testSparseEncodingProcessor_E2EFlow() throws Exception {
...
+            }
+            assertEquals(1, totalHits);
+        }
         BoolQueryBuilder boolQueryBuilder = new BoolQueryBuilder();
         NeuralSparseQueryBuilder sparseEncodingQueryBuilder = new NeuralSparseQueryBuilder().fieldName(TEST_SPARSE_ENCODING_FIELD)
             .queryText(TEXT)
             .modelId(modelId);
         MatchQueryBuilder matchQueryBuilder = new MatchQueryBuilder(TEST_TEXT_FIELD, TEXT);
-        boolQueryBuilder.should(sparseEncodingQueryBuilder).should(matchQueryBuilder);
+        boolQueryBuilder.should(sparseEncodingQueryBuilder).should(matchQueryBuilder).filter(new TermQueryBuilder("_id", "0"));
```
heemin32: I think filtering on id of 0 might not correct approach here?
We might want to check why the search result is not consistent between run?

owaiskazi19: Doc 1 is preceding doc 0 in the mixed cluster state causing doc 0 to be not the first doc

heemin32: Then, shouldn't we switch the assertion that the first doc is doc 1?

owaiskazi19: Including the check for total hits actually solved it. Now it checks for the total hits and then proceed with checking on the doc value which turns to be 0

martin-gaievski:
> Should we keep the cluster state as yellow for bwc test to avoid such flakiness?

It's not clear why this yellow cluster state happens, I don't think it's related to this change itself. I'm fine with relaxing validation and accepting yellow state as valid if we do proper deep dive later to figure out what's causing this

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/NeuralSparseSearchIT.java
```diff
@@ -103,6 +104,27 @@ public void testSparseEncodingProcessor_E2EFlow() throws Exception {
     private void validateTestIndexOnUpgrade(final int numberOfDocs, final String modelId) throws Exception {
         int docCount = getDocCount(getIndexNameForTest());
         assertEquals(numberOfDocs, docCount);
+        for (int i = 0; i < numberOfDocs; i++) {
```
heemin32, line 107: How this change fix the flaky failures on the test?

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/BatchIngestionIT.java
```diff
@@ -22,7 +22,7 @@ public class BatchIngestionIT extends AbstractRollingUpgradeTestCase {
     private static final String EMBEDDING_FIELD_NAME = "passage_embedding";
 
     public void testBatchIngestion_SparseEncodingProcessor_E2EFlow() throws Exception {
-        waitForClusterHealthGreen(NODES_BWC_CLUSTER, 90);
+        waitForClusterHealthGreen(NODES_BWC_CLUSTER, 120);
```
heemin32: I think this increased value only delay the time to failure of the test but not the chance of success.

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/NeuralSparseSearchIT.java
owaiskazi19: We are checking for total hits before checking the innerhits. If the hits are not present that would fail the assertEquals check on L126 itself.

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/BatchIngestionIT.java
owaiskazi19: yes, observed that in few test runs. We can remove these time in https://github.com/opensearch-project/neural-search/issues/1468 and accept the yellow state for all the rolling tests

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/NeuralSparseSearchIT.java
heemin32: Got it. So this is like adding more check to find the root cause of flaky failure but does not resolve the issue https://github.com/opensearch-project/neural-search/issues/1416 entirely?

owaiskazi19: yes but it is also providing a little time for the response to get stabilized with first checking total hits and then inner. From the couple of the test run didn't came across the failure for the test due to assertion but we can keep the issue open if you'd like to see if we see the flakiness again

qa/rolling-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/rolling/BatchIngestionIT.java
heemin32: If this value change does not help much, can we revert it back to 90?

owaiskazi19: sure, will take care of that in https://github.com/opensearch-project/neural-search/issues/1468. will raise it soon

owaiskazi19: Added in this PR to check for yellow state

heemin32:
@owaiskazi19 Really appreciate your effort in working to resolve the issue. That said, I’m still a bit hesitant to merge this change for the following reasons:

1. I’m not yet clear on how the added code helps prevent the flaky failure.

2. I’m unsure whether it's acceptable to treat a yellow cluster state as normal. With this change, we might lose the chance to find the root cause of unassigned shard.

Would you mind giving me a bit more time to think this through? Thanks again!

heemin32:
We can close this PR. The fix is in https://github.com/opensearch-project/neural-search/pull/1475

PR #1471, opened 2025-07-22
Add setting for number of documents stored by HybridCollapsingTopDocsCollector

### LINKED ISSUE #1381: [Hybrid] Add setting for number of documents stored by HybridCollapsingTopDocsCollector
In HybridCollapsingTopDocs collector, a priority queue is created for each group and subquery combination: https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java#L489 

The size of each queue is set to the size parameter provided in the query, which is the same as when using hybrid query without collapse.  However, we are storing a lot more documents than in hybrid search without collapse because there are multiple queues instead of a single queue.

We should add a setting that allows users to control the number of documents that are stored in each queue.  Lowering this setting would prioritize latency over recall, while raising the setting would do the opposite.  This setting would allow users the ability to customize for their individual use case.

ryanbogan:
### Description
Adds a setting that allows users to control the number of documents that are stored in each queue within HybridCollapsingTopDocsCollector. Lowering this setting prioritizes latency over recall, while raising the setting does the opposite. 

### Related Issues
Resolves #1381 

### Check List


ryanbogan:
Open to feedback on default value and upper limit for the new setting

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
```diff
@@ -47,4 +47,13 @@ public final class NeuralSearchSettings {
         Setting.Property.IndexScope,
         Setting.Property.Dynamic
     );
+
+    public static final Setting<Integer> HYBRID_COLLAPSE_DOCS_PER_GROUP = Setting.intSetting(
+        "index.neural_search.hybrid_collapse_docs_per_group",
+        10,
```
martin-gaievski: I believe that is default value for this setting, is this based on some experimental data? does it give good balance between accuracy and performance?

martin-gaievski (commented):
please add test for checking user provided value for new setting, like what is behavior if they put some invalid string or float value.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
```diff
@@ -47,4 +47,13 @@ public final class NeuralSearchSettings {
         Setting.Property.IndexScope,
         Setting.Property.Dynamic
     );
+
+    public static final Setting<Integer> HYBRID_COLLAPSE_DOCS_PER_GROUP = Setting.intSetting(
+        "index.neural_search.hybrid_collapse_docs_per_group",
```
martin-gaievski: This setting name is lacking visibility that it's per sub-query. Can we add that context, maybe something like `hybrid_collapse_docs_per_group_per_subquery` or a bit less informative `hybrid_collapse_max_docs_per_group`

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
ryanbogan: I'm changing this to default to the current behavior, which is using the size of the query.

owaiskazi19:
@ryanbogan rather than making it configurable for users and making them think about it, can we experiment and define an optimal and fixed number of documents to be stored in queue?

owaiskazi19 (approved):
Fine with keeping settings for now but we should document this somewhere in the collapse docs

ryanbogan:
> Fine with keeping settings for now but we should document this somewhere in the collapse docs

Agreed, I'm going to raise a doc PR to update for both this setting and the addition of collapse with inner hits

vibrantvarun (commented):
Looks good to me. Please resolve changelog conflicts.

vibrantvarun (approved):

Issue #1468, opened 2025-07-21
[BUG] Flaky BWC tests because of yellow cluster


owaiskazi19:
### What is the bug?
Lots of bwc rolling upgrade tests fails intermittently because of the state as yellow while running the tests.
```
     org.opensearch.client.ResponseException: method [GET], host [http://[::1]:46225], URI [/_cluster/health?wait_for_nodes=3&timeout=60s&wait_for_status=green&cluster_manager_timeout=60s], status line [HTTP/1.1 408 Request Timeout]
    {"cluster_name":"neuralSearchBwcCluster-rolling","status":"yellow","timed_out":true,"number_of_nodes":3,"number_of_data_nodes":3,"discovered_master":true,"discovered_cluster_manager":true,"active_primary_shards":40,"active_shards":80,"relocating_shards":0,"initializing_shards":0,"unassigned_shards":1,"delayed_unassigned_shards":0,"number_of_pending_tasks":0,"number_of_in_flight_fetch":0,"task_max_waiting_in_queue_millis":0,"active_shards_percent_as_number":98.76543209876543}
        at __randomizedtesting.SeedInfo.seed([33911F30822AFEE2:930032FDAA3577F9]:0)
        at app//org.opensearch.client.RestClient.convertResponse(RestClient.java:501)
        at app//org.opensearch.client.RestClient.performRequest(RestClient.java:384)
        at app//org.opensearch.client.RestClient.performRequest(RestClient.java:359)
        at app//org.opensearch.neuralsearch.BaseNeuralSearchIT.waitForClusterHealthGreen(BaseN
```

### How can one reproduce the bug?
1. Run rolling upgrade bwc tests on java 21 or 24 and see the error

### What is the expected behavior?
Tests should pass

### What is your host/environment?
_Operating system, version._

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._

### Do you have any additional context?
_Add any other context about the problem._

heemin32:
The cluster turns yellow because the node was killed before replica initialization finished. Then, the primary shard gets assigned to the new node with a higher version, while the replica shard cannot be allocated to the other nodes due to their lower versions. We should add a delay after the last cluster operation before ending the test, allowing enough time for replication to complete before the node is terminated.

```
{
  "cluster_name": "neuralSearchBwcCluster-rolling",
  "status": "yellow",
  "timed_out": true,
  "number_of_nodes": 3,
  "number_of_data_nodes": 3,
[...]

PR #1473, opened 2025-07-23
Add log to debug bwc failures

heemin32:
### Description
Testing

### Related Issues

### Check List


qa/restart-upgrade/build.gradle
```diff
@@ -144,6 +144,10 @@ task testAgainstNewCluster(type: StandaloneRestIntegTestTask) {
     dependsOn "testAgainstOldCluster"
     dependsOn rootProject.tasks.assemble
     useCluster testClusters."${baseName}"
+    doFirst {
+        println "Waiting for cluster to stabilize after previous test, including data synchronization on replica shards"
+        Thread.sleep(10000) // 10 seconds delay
+    }
```
yuye-aws: Waiting 10 seconds may not necessarily resolve the stabilization case. Can you share more context on what job are not stabilized? We can then check detailed cluster info like wait for cluster health

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -1992,8 +1995,19 @@ protected void waitForClusterHealthGreen(final String numOfNodes, final int time
...
+        } catch (ResponseException e) {
+            // Perform additional API calls to log the cause of the yellow cluster state
+            Request explain = new Request("GET", "/_cluster/allocation/explain");
+            logger.info(EntityUtils.toString(client().performRequest(explain).getEntity()));
+            Request shards = new Request("GET", "/_cat/shards?v");
+            logger.info(EntityUtils.toString(client().performRequest(shards).getEntity()));
+            Request health = new Request("GET", "/_cat/health?v");
+            logger.info(EntityUtils.toString(client().performRequest(health).getEntity()));
+            throw e;
+        }
```
yuye-aws, line 2010: Adding log helps to debug. Have you considered what should we do when cluster health is not green after waiting for 60 seconds? For example, if shards is not green, then shall we wait for longer?

qa/restart-upgrade/build.gradle
heemin32: Please see https://github.com/opensearch-project/neural-search/issues/1468

By the way, seems like yellow state still happening. Might need to revisit the approach.

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
heemin32: Waiting longer doesn't work. See https://github.com/opensearch-project/neural-search/issues/1468 for more in detail

qa/restart-upgrade/build.gradle
heemin32: Seem like doFirst block is executed in reverse order. Raised new revision.

qa/restart-upgrade/build.gradle
```diff
@@ -144,12 +144,15 @@ task testAgainstNewCluster(type: StandaloneRestIntegTestTask) {
     dependsOn "testAgainstOldCluster"
     dependsOn rootProject.tasks.assemble
     useCluster testClusters."${baseName}"
-    if(ext.bwcBundleTest){
-        doFirst {
+
+    doFirst {
+        // This is added to prevent the cluster from getting stuck in yellow state
+        println "Waiting for cluster to stabilize after previous test, including data synchronization on replica shards"
+        Thread.sleep(10000) // 10 seconds delay
```
martin-gaievski, line 151: hardcoded value is good first step, we can try to change it to something more sophisticated later

heemin32: Good comment. Another option is to wait for the cluster to be green before upgrading the node to the next version in this gradle file, or we could wait for the cluster to be green after every test before completing it, since node upgrades only occur once the previous test has finished.

zhichao-aws (approved):
LGTM. Thanks.

qa/restart-upgrade/build.gradle
owaiskazi19: @heemin32 This would be a better option to wait for cluster to become green as a next phase after we added the hardcoded delay as the first step

heemin32: Yes. Let's see whether yellow cluster issues occur again in the future. For now, this approach should work well since we only need a 1-2 second delay at most, and there's minimal downside to using a fixed delay. Added a follow up github issue. https://github.com/opensearch-project/neural-search/issues/1478

PR #1480, opened 2025-07-25
Support configuring reusing existing embedding for the semantic field.

### LINKED ISSUE #1350: [FEATURE] Semantic Field Enhancement - Allow to Reuse Existing Embedding

Currently the embedding generation for the semantic field is done through a system generated ingest processor which users cannot configure it like a normal ingest processor. We will always generate the embedding no matter if there is existing embedding in the existing doc. We should allow users to configure if they want to reuse the existing embedding to save some predict API call.

We should allow users to configure the system ingest processor through the semantic field to control the embedding generation behavior.

skip_existing: When true, the embedding generation does not make inference calls for fields that already contain embeddings, leaving existing embeddings unchanged. Default is false.

N/A

bzhangam:
### Description
Support configuring reusing existing embedding for the semantic field.

Configurable reusing existing embedding

* Support a configurable parameter that controls should we reuse the exist embedding if the value of the semantic field is not changed and the exist doc has the embedding for reuse.
* Default should be false and CX should explicitly turn it on if they want to use it.

### Related Issues
Resolves #1350 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/constants/SemanticFieldConstants.java
+++ b/src/main/java/org/opensearch/neuralsearch/constants/SemanticFieldConstants.java
@@ -59,4 +59,11 @@ public class SemanticFieldConstants {
      * {@link org.opensearch.neuralsearch.mapper.dto.SparseEncodingConfig}
      */
     public static final String SPARSE_ENCODING_CONFIG = "sparse_encoding_config";
+
+    /**
+     * Name of the field to configure if we should skip the embedding generation for the semantic field with existing
+     * reusable embedding. We think the embedding is reusable if it exists and the semantic field value and the
+     * model is not changed.
+     */
+    public static final String SKIP_EXISTING_EMBEDDING = "skip_existing_embedding";
 }
--- a/src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
@@ -8,6 +8,9 @@ import com.google.common.annotations.VisibleForTesting;
 import lombok.NonNull;
 import lombok.extern.log4j.Log4j2;
 import org.apache.commons.lang3.tuple.Pair;
+import org.opensearch.action.get.MultiGetAction;
+import org.opensearch.action.get.MultiGetItemResponse;
+import org.opensearch.action.get.MultiGetRequest;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.Nullable;
 import org.opensearch.core.action.ActionListener;
@@ -29,6 +32,7 @@ import org.opensearch.neuralsearch.stats.events.EventStatsManager;
 import org.opensearch.neuralsearch.util.TokenWeightUtil;
 import org.opensearch.neuralsearch.util.prune.PruneType;
 import org.opensearch.neuralsearch.util.prune.PruneUtils;
+import org.opensearch.transport.client.OpenSearchClient;
 
 import java.util.ArrayList;
 import java.util.Collection;
@@ -38,6 +42,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -45,6 +50,8 @@ import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
+import static org.opensearch.neuralsearch.constants.DocFieldNames.ID_FIELD;
+import static org.opensearch.neuralsearch.constants.DocFieldNames.INDEX_FIELD;
 import static org.opensearch.neuralsearch.constants.MappingConstants.PATH_SEPARATOR;
 import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.CHUNKS_TEXT_FIELD_NAME;
 import static org.opensearch.neuralsearch.constants.SemanticInfoFieldConstants.MODEL_ID_FIELD_NAME;
@@ -55,12 +62,14 @@ import static org.opensearch.neuralsearch.processor.chunker.Chunker.MAX_CHUNK_LI
 import static org.opensearch.neuralsearch.processor.util.ChunkUtils.chunkList;
 import static org.opensearch.neuralsearch.processor.util.ChunkUtils.chunkString;
 import static org.opensearch.neuralsearch.processor.util.ProcessorUtils.getMaxTokenCount;
+import static org.opensearch.neuralsearch.processor.util.ProcessorUtils.getValueFromSourceByFullPath;
 import static org.opensearch.neuralsearch.util.ProcessorDocumentUtils.unflattenIngestDoc;
 import static org.opensearch.neuralsearch.util.SemanticMLModelUtils.getModelType;
 import static org.opensearch.neuralsearch.util.SemanticMLModelUtils.isDenseModel;
 import static org.opensearch.neuralsearch.util.SemanticMappingUtils.getModelId;
 import static org.opensearch.neuralsearch.util.SemanticMappingUtils.getSemanticInfoFieldFullPath;
 import static org.opensearch.neuralsearch.util.SemanticMappingUtils.isChunkingEnabled;
+import static org.opensearch.neuralsearch.util.SemanticMappingUtils.isSkipExistingEmbeddingEnabled;
 
 /**
  * Processor to ingest the semantic fields. It will do text chunking and embedding generation for the semantic field.
@@ -86,6 +95,8 @@ public class SemanticFieldProcessor extends AbstractBatchingSystemProcessor {
 
     private final static float DEFAULT_PRUNE_RATIO = 0.1f;
 
+    private final OpenSearchClient openSearchClient;
+
     public SemanticFieldProcessor(
         @Nullable final String tag,
         @Nullable final String description,
@@ -95,7 +106,8 @@ public class SemanticFieldProcessor extends AbstractBatchingSystemProcessor {
         @NonNull final Environment environment,
         @NonNull final ClusterService clusterService,
         @NonNull final Chunker defaultTextChunker,
-        @NonNull final AnalysisRegistry analysisRegistry
+        @NonNull final AnalysisRegistry analysisRegistry,
+        @NonNull final OpenSearchClient openSearchClient
     ) {
         super(tag, description, batchSize);
         this.pathToFieldConfig = pathToFieldConfig;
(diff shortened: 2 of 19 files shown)


src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -180,16 +193,121 @@ private void process(
...
+            String.format(
+                Locale.ROOT,
+                "Failed to get existing docs to check embedding reusability for the semantic field. Error: %s",
+                exception.getMessage()
+            ),
+            exception
+        );
+    }
+
+    private List<SemanticFieldInfo> applyReusableEmbeddingsAndFilterUnprocessedFields(
```
junqiu-lei, line 239: How about to refactor and share function with embedding processor for the reusable embeddings logic?

src/main/java/org/opensearch/neuralsearch/constants/SemanticFieldConstants.java
```diff
@@ -59,4 +59,11 @@ public class SemanticFieldConstants {
      * {@link org.opensearch.neuralsearch.mapper.dto.SparseEncodingConfig}
      */
     public static final String SPARSE_ENCODING_CONFIG = "sparse_encoding_config";
+
+    /**
+     * Name of the field to configure if we should reuse the exist embedding for the semantic field if the original data
+     * and the model is not changed.
+     * {@link org.opensearch.neuralsearch.mapper.dto.SparseEncodingConfig}
+     */
+    public static final String REUSE_EXIST_EMBEDDING = "reuse_exist_embedding";
```
will-hwang: the feature is behind the name `skip_existing`(https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java#L45) we should have the same name here

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
```diff
@@ -180,16 +193,121 @@ private void process(
...
-        setModelInfo(ingestDocument, semanticFieldInfoList);
-
         boolean isChunked = chunk(ingestDocument, semanticFieldInfoList);
         if (isChunked) {
             EventStatsManager.increment(EventStatName.SEMANTIC_FIELD_PROCESSOR_CHUNKING_EXECUTIONS);
         }
 
+        final Set<String> docIdsToCheckReuse = getDocIdsToCheckReuse(List.of(semanticFieldInfoList));
+        final Object index = ingestDocument.getSourceAndMetadata().get(INDEX_FIELD);
+        if (shouldCheckExistDoc(docIdsToCheckReuse, index)) {
```
will-hwang, line 203: instead of having a separate check for `shouldCheckExistDoc`, we can have `getDocIdsToCheckReuse` return empty set if there is none

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
```diff
@@ -209,6 +209,45 @@ && unsafeCastToObjectList(current).getFirst() instanceof String) {
         return Optional.ofNullable(current);
     }
 
+    /**
+     * Returns the data found in the doc source based on the full path.
+     * @param sourceAsMap The source of a doc.
+     * @param fullPath The full path in the doc to the value.
+     * @return The data at the full path in the doc.
+     */
+    public static Object getValueFromSourceByFullPath(final Map<String, Object> sourceAsMap, final String fullPath) {
```
will-hwang, line 244: can we re-use the method here (https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java#L184), instead of defining a new method?

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
bzhangam: The logic is different since the embedding processor relies on the field map to figure out which field we should generate the embedding for and where to store the embedding. But here we rely on the semantic field to get that info.

Besides embedding processor is separated from the chunking but here we handle the chunking at the same time.

src/main/java/org/opensearch/neuralsearch/constants/SemanticFieldConstants.java
bzhangam: I think there is a difference. It makes sense that we name it as `skip_existing` in a embedding processor since we know we are talking about the embedding. But this new parameter is in the semantic field itself where we don't know what we want to skip so it's better to make it more descriptive to say if we want to reuse the existing embedding for the semantic field.

src/main/java/org/opensearch/neuralsearch/processor/semantic/SemanticFieldProcessor.java
bzhangam: We already did that. This check is check the list is empty and also check if the index is a valid string or not.

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
bzhangam: In semantic field processor we create the semantic info for each semantic field in the doc and to process each field we need to get the value by the full path in the doc rather than use the path defined the index mapping. The existing function will pull the nested object as an array but I think if we have multiple nested layers there will be an issue. So I'd like to keep using the full path to process each field which is more robust.

src/main/java/org/opensearch/neuralsearch/constants/SemanticFieldConstants.java
will-hwang: i still think having two different names for same feature should be avoided. i think having the same feature name with documentation makes more sense even in the semantic field

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
will-hwang: can we append `semantic_info` field as a prefix to the existing method?

src/main/java/org/opensearch/neuralsearch/constants/SemanticFieldConstants.java
junqiu-lei: +1 to @will-hwang. The functions are same from customer perspective, ideally we should have same name style and default configuration. Or say `skip_existing_embedding` in the semantic field.

bzhangam: I'm ok with using `skip_existing_embedding ` but don't think we should simply call it `skip_existing ` since we don't have the context in the semantic field config that we are talking about the embedding. This is different from the parameter we have in the embedding processor. @junqiu-lei @will-hwang How do you think?

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
bzhangam: Sorry I don't get what do you mean. If we want to differentiate them I think we can make the rename the existing method as `getValueFromSourceByPathInMapping` to make it clear and name this one as `getValueFromSourceByPathInDoc`?

src/main/java/org/opensearch/neuralsearch/constants/SemanticFieldConstants.java
heemin32: Can we make this the default behavior and avoid exposing the parameter to users? The only downside is a slight slowdown during ingestion due to reading the document, but since the inference call itself is already significantly more expensive, the added latency from this feature should be negligible. Is there any scenario where a customer would benefit from not enabling it?

will-hwang: no, but this feature is default to `false` in embedding processors, i think default should be false in semantic field too

heemin32: The semantic field and the embedding processor don't need to behave the same way—they serve different purposes. Whether the semantic field uses the embedding processor is an internal implementation detail that customers don't need to be concerned with.

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
will-hwang: the existing method also takes in doc, and should be compatible with this use case too. We should still be able to pass in full path to semantic field to get their values

src/main/java/org/opensearch/neuralsearch/constants/SemanticFieldConstants.java
bzhangam: If we do it that way I think we are changing the behavior silently, is it ok for a minor version upgrade? Besides what if user wants to always update the embedding if they are keep improving their remote model without changing the model id in OpenSearch?

If we are ok with the behavior change I think we can enable it as the default behavior. But probably it's better to keep it updatable for users.

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
heemin32: @bzhangam Could you show an example where existing method cannot handle but this method can?

src/main/java/org/opensearch/neuralsearch/constants/SemanticFieldConstants.java
heemin32: Fair enough. I initially thought that switching the remote model without fully regenerating the embeddings wasn't the right approach, but I can see how some users might choose to do that.

As for the naming, I think it would be better to use something that aligns more closely with the context of semantic fields, rather than reusing the name from the text embedding ingest processor.

src/main/java/org/opensearch/neuralsearch/constants/SemanticFieldConstants.java
```diff
@@ -59,4 +59,11 @@ public class SemanticFieldConstants {
      * {@link org.opensearch.neuralsearch.mapper.dto.SparseEncodingConfig}
      */
     public static final String SPARSE_ENCODING_CONFIG = "sparse_encoding_config";
+
+    /**
+     * Name of the field to configure if we should skip the embedding generation for the semantic field with existing
+     * reusable embedding. We think the embedding is reusable if it exists and the semantic field value and the
+     * model is not changed.
+     */
+    public static final String SKIP_EXIST_EMBEDDING = "skip_exist_embedding";
```
will-hwang: i would prefer `skip_existing_embedding`

bzhangam: ok. was trying to keep it short

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
bzhangam: This method can get value for path like `products.0.product_description` but the existing one only can handle `products.product_description`. It simply put all the data in a nested object into an array. Not able to pull each value based on its index.

junqiu-lei (approved):

Issue #1479, opened 2025-07-24
[RFC]  Design for Agentic Search


owaiskazi19:
## Problem Statement

Users often struggle to find the exact information they need when searching through large amounts of documents. They have to guess the right keywords or learn special search rules to get good results. When using OpenSearch,  to build up proper queries,  query understanding requires a long of background information for users and it’s an important key to retrieve relevant information. To solve this problem, we want to build a smart search helper, Agentic Search,  that can understand questions asked in natural language, turn them into proper search queries for OpenSearch, find the most relevant documents, evaluate the documents and give back clear, simple answers along with the documents. This will make it much easier for users to find what they're looking for just by asking questions naturally, like they would ask another person.

Currently OpenSearch offers 3 types of searches: lexical, semantic and hybrid search. To get the best of both world of keyword matching and semantic understanding of the query, Hybrid search was introduced. It combines the best of both lexical search and semantic search: sparse vectors or BM25 handle the precise keyword matching, while dense vectors power the semantic understanding. This dual approach was brought in to help users find exactly what they need, even when their search terms aren't perfect. Users now want the queries to be handled by OpenSearch itself rather they learning about searches available. This can be achieve by introducing intelligent or agentic search.

## Collaborators
@mingshl @owaiskazi19 

## **Solution Overview: Agentic Search**

Agentic Search is an LLM-enhanced search orchestration framework designed to act as an intelligent search assistant layered over OpenSearch. It enables users to interact with search systems in plain natural language—without needing to understand how search engines work.

### **Approach 1: Using Plan and Execute Agent**

This approach can be powered by ml-commons agent framework capable of planning, executing, and reflecting on complex search tasks.
Built on the foundation of the [**Plan-Execute-Reflect Agent**](https://docs.opensearch.org/docs/latest/tutorials/gen-ai/agents/build-plan-execute-reflect-agent/), the agent interprets user intent, decomposes queries into a sequence of actions, and dynamically selects tools to gather and synthesize information.

### Plan and Execute Agent

* **Planning** – The planner LLM generates an initial step-by-step plan using the available tools.
* **Execution** – Each step is executed sequentially using the conversational agent and the available tools.
* **Re-evaluation** – After executing each step, the planner LLM re-evaluates the plan using intermediate results. The LLM can adjust the plan dynamically to skip, add, or change steps based on new context.

This loop ensures that the agent can handle complex, evolving information needs with precision and flexibility.

### **Tools for Agentic Search**

[...]

PR #1483, opened 2025-07-28
[Experimental] Added new query clause agentic search

owaiskazi19:
### Description
Added new query clause agentic search and tests
```
GET /<index_name>/_search
{
  "query": {
    "agentic_search": {
      "query_text": "Find me a red car",
      "agent_id": "asn123456789".
      "query_fields": ["title", "description", "location"] //optional
    }
  }
}
```

### Related Issues
https://github.com/opensearch-project/neural-search/issues/1479

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -264,6 +264,7 @@ dependencies {
     testFixturesImplementation "org.opensearch.test:framework:${opensearch_version}"
     implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.14.0'
     implementation group: 'org.apache.commons', name: 'commons-math3', version: '3.6.1'
+    implementation group: 'com.google.code.gson', name: 'gson', version: '2.10.1'
     // ml-common excluded reflection for runtime so we need to add it by ourselves.
     // https://github.com/opensearch-project/ml-commons/commit/464bfe34c66d7a729a00dd457f03587ea4e504d9
     // TODO: Remove following three lines of dependencies if ml-common include them in their jar
@@ -271,7 +272,6 @@ dependencies {
     runtimeOnly group: 'org.javassist', name: 'javassist', version: '3.29.2-GA'
     runtimeOnly group: 'org.opensearch', name: 'common-utils', version: "${opensearch_build}"
     runtimeOnly group: 'org.apache.commons', name: 'commons-text', version: '1.10.0'
-    runtimeOnly group: 'com.google.code.gson', name: 'gson', version: '2.10.1'
     runtimeOnly group: 'org.json', name: 'json', version: '20231013'
     // json-path 2.9.0 depends on slf4j 2.0.11, which conflicts with the version used by OpenSearch core.
     // Excluding slf4j here since json-path is only used for testing, and logging failures in this context are acceptable.
--- a/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
@@ -247,8 +247,7 @@ public class MLCommonsClientAccessor {
             );
         }
         List<ModelTensor> tensorList = tensorOutputList.get(0).getMlModelTensors();
-        String result = tensorList.get(0).getResult();
-        return result;
+        return tensorList.get(0).getResult();
     }
 
     private <T extends Number> List<T> buildSingleVectorFromResponse(final MLOutput mlOutput) {
@@ -341,7 +340,7 @@ public class MLCommonsClientAccessor {
      * It will fail if any one of the get model request fail. Only return the success result if all model info is
      * successfully retrieved.
      *
-     * @param modelIds  a set of model ids
+     * @param modelIds a set of model ids
      * @param onSuccess onSuccess consumer
      * @param onFailure onFailure consumer
      */
@@ -451,7 +450,7 @@ public class MLCommonsClientAccessor {
      * This method will highlight relevant sentences in the context based on the question.
      *
      * @param inferenceRequest the request containing the question and context for highlighting
-     * @param listener         the listener to be called with the highlighting results
+     * @param listener the listener to be called with the highlighting results
      */
     public void inferenceSentenceHighlighting(
         @NonNull final SentenceHighlightingRequest inferenceRequest,
@@ -488,7 +487,6 @@ public class MLCommonsClientAccessor {
                 // Extract DSL query from inference results following the structure:
                 MLOutput mlOutput = (MLOutput) response.getOutput();
                 final String inferenceResults = buildQueryResultFromResponseOfOutput(mlOutput);
-                ;
 
                 listener.onResponse(inferenceResults);
             } catch (Exception e) {
--- a/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
@@ -27,7 +27,9 @@ import org.opensearch.ml.common.MLModel;
 import org.opensearch.ml.common.dataset.MLInputDataset;
 import org.opensearch.ml.common.dataset.TextDocsInputDataSet;
 import org.opensearch.ml.common.dataset.TextSimilarityInputDataSet;
+import org.opensearch.ml.common.dataset.remote.RemoteInferenceInputDataSet;
 import org.opensearch.ml.common.input.MLInput;
+import org.opensearch.ml.common.input.execute.agent.AgentMLInput;
 import org.opensearch.ml.common.output.MLOutput;
 import org.opensearch.ml.common.output.model.ModelResultFilter;
 import org.opensearch.ml.common.output.model.ModelTensor;
@@ -234,6 +236,19 @@ public class MLCommonsClientAccessor {
             }
         }
         return resultMaps;
+    }
+
+    private String buildQueryResultFromResponseOfOutput(MLOutput mlOutput) {
+        final ModelTensorOutput modelTensorOutput = (ModelTensorOutput) mlOutput;
+        final List<ModelTensors> tensorOutputList = modelTensorOutput.getMlModelOutputs();
+        if (CollectionUtils.isEmpty(tensorOutputList) || CollectionUtils.isEmpty(tensorOutputList.get(0).getMlModelTensors())) {
+            throw new IllegalStateException(
+                "Empty model result produced. Expected at least [1] tensor output and [1] model tensor, but got [0]"
+            );
+        }
+        List<ModelTensor> tensorList = tensorOutputList.get(0).getMlModelTensors();
+        String result = tensorList.get(0).getResult();
+        return result;
     }
 
     private <T extends Number> List<T> buildSingleVectorFromResponse(final MLOutput mlOutput) {
@@ -452,8 +467,7 @@ public class MLCommonsClientAccessor {
      * @param parameters the parameters to pass to the agent
      * @param listener   the listener to be called with the DSL query result
      */
-    //TODO add execute Agent
-    /*public void executeAgent(
+    public void executeAgent(
         @NonNull final String agentId,
         @NonNull final Map<String, String> parameters,
         @NonNull final ActionListener<String> listener
@@ -467,10 +481,19 @@ public class MLCommonsClientAccessor {
         final int retryTime,
         final ActionListener<String> listener
     ) {
-        mlClient.executeAgent(agentId, "POST", parameters, ActionListener.wrap(response -> {
-            // Extract DSL query from inference results
-            String dslQuery = response.getInferenceResults().get(0).get("output").get(0).get("result");
-            listener.onResponse(dslQuery);
+        RemoteInferenceInputDataSet dataset = RemoteInferenceInputDataSet.builder().parameters(parameters).build();
+        AgentMLInput agentMLInput = new AgentMLInput(agentId, null, FunctionName.AGENT, dataset);
+        mlClient.execute(FunctionName.AGENT, agentMLInput, ActionListener.wrap(response -> {
+            try {
+                // Extract DSL query from inference results following the structure:
+                MLOutput mlOutput = (MLOutput) response.getOutput();
+                final String inferenceResults = buildQueryResultFromResponseOfOutput(mlOutput);
+                ;
+
+                listener.onResponse(inferenceResults);
+            } catch (Exception e) {
+                listener.onFailure(new IllegalStateException("Failed to extract result from agent response", e));
+            }
         },
             e -> RetryUtil.handleRetryOrFailure(
                 e,
@@ -479,5 +502,5 @@ public class MLCommonsClientAccessor {
                 listener
             )
         ));
-    }*/
+    }
 }
--- a/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
@@ -57,9 +57,9 @@ public class MLCommonsClientAccessor {
      * Wrapper around {@link #inferenceSentences} that expected a single input text and produces a single floating
      * point vector as a response.
      *
-     * @param modelId {@link String}
+     * @param modelId   {@link String}
      * @param inputText {@link String}
-     * @param listener {@link ActionListener} which will be called when prediction is completed or errored out
+     * @param listener  {@link ActionListener} which will be called when prediction is completed or errored out
      */
     public void inferenceSentence(
         @NonNull final String modelId,
@@ -92,7 +92,7 @@ public class MLCommonsClientAccessor {
      * need to run only TextEmbedding tasks only.
      *
      * @param inferenceRequest {@link InferenceRequest}
-     * @param listener {@link ActionListener} which will be called when prediction is completed or errored out.
+     * @param listener         {@link ActionListener} which will be called when prediction is completed or errored out.
      */
     public void inferenceSentences(
         @NonNull final TextInferenceRequest inferenceRequest,
@@ -114,7 +114,7 @@ public class MLCommonsClientAccessor {
      * using the actionListener which will have a list of floats in the order of inputText.
      *
      * @param inferenceRequest {@link InferenceRequest}
-     * @param listener {@link ActionListener} which will be called when prediction is completed or errored out.
+     * @param listener         {@link ActionListener} which will be called when prediction is completed or errored out.
      */
     public void inferenceSentencesMap(@NonNull MapInferenceRequest inferenceRequest, @NonNull final ActionListener<List<Number>> listener) {
         retryableInferenceSentencesWithSingleVectorResult(inferenceRequest, 0, listener);
@@ -126,7 +126,7 @@ public class MLCommonsClientAccessor {
      * the similarity scores of the texts w.r.t. the query text, in the order of the input texts.
      *
      * @param inferenceRequest {@link InferenceRequest}
-     * @param listener {@link ActionListener} receives the result of the inference
+     * @param listener         {@link ActionListener} receives the result of the inference
      */
     public void inferenceSimilarity(
         @NonNull SimilarityInferenceRequest inferenceRequest,
@@ -325,7 +325,8 @@ public class MLCommonsClientAccessor {
      * Get model info for multiple model ids. It will send multiple getModel requests to get the model info in parallel.
      * It will fail if any one of the get model request fail. Only return the success result if all model info is
      * successfully retrieved.
-     * @param modelIds a set of model ids
+     *
+     * @param modelIds  a set of model ids
      * @param onSuccess onSuccess consumer
      * @param onFailure onFailure consumer
      */
@@ -435,7 +436,7 @@ public class MLCommonsClientAccessor {
      * This method will highlight relevant sentences in the context based on the question.
      *
      * @param inferenceRequest the request containing the question and context for highlighting
-     * @param listener the listener to be called with the highlighting results
+     * @param listener         the listener to be called with the highlighting results
      */
     public void inferenceSentenceHighlighting(
         @NonNull final SentenceHighlightingRequest inferenceRequest,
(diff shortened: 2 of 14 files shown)


src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,227 @@
...
+ * AgenticSearchQueryBuilder is responsible for agentic_search query type. It executes a root agent to extract the DSL query and parse it to innerQueryBuilder for parsing.
+ */
+@Accessors(chain = true, fluent = true)
+@NoArgsConstructor
+@AllArgsConstructor
+@Getter(AccessLevel.PACKAGE)
+@Setter(AccessLevel.PACKAGE)
+public final class AgenticSearchQueryBuilder extends AbstractQueryBuilder<AgenticSearchQueryBuilder> implements WithFieldName {
+
+    public static final String NAME = "agentic_search";
```
martin-gaievski: I think "search" part in the query name is redundant, maybe simply "agentic". Although I don't know how much marketing is in the name :)

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,227 @@
...
+    @Override
+    protected void doXContent(XContentBuilder xContentBuilder, Params params) throws IOException {
+        xContentBuilder.startObject(NAME);
+        if (Objects.nonNull(QUERY_TEXT_FIELD)) {
+            xContentBuilder.field(QUERY_TEXT_FIELD.getPreferredName(), queryText);
+        }
+        if (Objects.nonNull(AGENT_ID)) {
+            xContentBuilder.field(AGENT_ID.getPreferredName(), agentId);
+        }
+        if (Objects.nonNull(queryFields) && !queryFields.isEmpty()) {
```
martin-gaievski, line 106: in case queryFields.isEmpty() what's the value that we use? From the quick glance it can be null

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,227 @@
...
+        }
+
+//        ML_CLIENT.executeAgent(agentId, parameters, ActionListener.wrap(future::complete, future::completeExceptionally));
+
+        try {
+            String dslQueryString = future.get();
+            // String dslQueryString = "{" + "\"match\": {" + "\"passage_text\": {" + "\"query\": \"science\"" + "}" + "}" + "}";
+            XContentParser parser = XContentType.JSON.xContent()
+                .createParser(queryRewriteContext.getXContentRegistry(), LoggingDeprecationHandler.INSTANCE, dslQueryString);
+            return parseInnerQueryBuilder(parser);
```
martin-gaievski: add logs so we know exact query that agent returned

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,227 @@
...
+        } catch (Exception e) {
+            throw new IOException("Failed to execute agentic search", e);
```
martin-gaievski, line 216: add logs with agent input

src/main/java/org/opensearch/neuralsearch/util/NeuralSearchClusterUtil.java
```diff
@@ -65,4 +67,20 @@ public List<IndexMetadata> getIndexMetadataList(@NonNull final IndicesRequest se
...
+            if (indices != null && indices.length > 0) {
+                IndexMetadata indexMetadata = clusterService.state().metadata().index(indices[0]);
+                if (indexMetadata != null) {
+                    return Objects.requireNonNull(indexMetadata.mapping()).source().toString();
+                }
+            }
+        } catch (Exception e) {
+            log.warn("Failed to extract index mapping", e);
+        }
+        return "{}";
```
martin-gaievski: is there a case when empty response is expected? I can think of something like index has not yet created, but we can fail in this case. I'm doubting if we need to return empty string, I think it's better to throw exception and fail fast

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,227 @@
...
+
+    @Override
+    protected QueryBuilder doRewrite(QueryRewriteContext queryRewriteContext) throws IOException {
+        if (ML_CLIENT == null) {
+            throw new IllegalStateException("ML client not initialized");
+        }
+
+        // the queryRewrite is expected at QueryCoordinator level
+        if (!(queryRewriteContext instanceof QueryCoordinatorContext)) {
+            throw new IllegalStateException(
```
martin-gaievski, line 178: why we're not simply return earlier in this case? What's the user side behavior in such scenario, can they do anything from their end to fix this?

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,227 @@
...
+        }
+        if (agenticSearchQueryBuilder.agentId == null || agenticSearchQueryBuilder.agentId.trim().isEmpty()) {
+            throw new ParsingException(parser.getTokenLocation(), "[" + AGENT_ID.getPreferredName() + "] is required");
+        }
+
+        return agenticSearchQueryBuilder;
+    }
+
+    @Override
+    protected QueryBuilder doRewrite(QueryRewriteContext queryRewriteContext) throws IOException {
```
martin-gaievski, line 166: Is this idempotent call or we are going to call agent every time? Rewrite can be called multiple times and that's expected, typical exit criteria is if output is same as input (basic idea is - we rewrite query to the basic possible form)

src/test/java/org/opensearch/neuralsearch/query/AgenticSearchQueryIT.java
```diff
@@ -0,0 +1,184 @@
...
+import java.util.Map;
+
+@AwaitsFix(bugUrl = "Ignoring until we find a way to fetch access, secret key for remote model")
+public class AgenticSearchQueryIT extends BaseNeuralSearchIT {
+
+    private static final String TEST_INDEX = "test-agentic-index";
+    private static String TEST_AGENT_ID;
+    private static String TEST_MODEL_ID;
+    private static final String TEST_QUERY_TEXT = "Find documents about machine learning";
+    private static final String AWS_ACCESS_KEY_ID = System.getenv("AWS_ACCESS_KEY_ID");
```
owaiskazi19, line 25: Need to discuss on fetching these credentials to use for remote model creating connector

build.gradle
```diff
@@ -264,14 +264,14 @@ dependencies {
     testFixturesImplementation "org.opensearch.test:framework:${opensearch_version}"
     implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.14.0'
     implementation group: 'org.apache.commons', name: 'commons-math3', version: '3.6.1'
+    implementation group: 'com.google.code.gson', name: 'gson', version: '2.10.1'
```
vibrantvarun, line 267: Can we not hardcode the dependency version? Instead we can use opensearch_version? So that in future we need not have to do anything for maintainence.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -236,6 +238,19 @@ private <T extends Number> List<List<T>> buildVectorFromResponse(MLOutput mlOutp
...
+        final ModelTensorOutput modelTensorOutput = (ModelTensorOutput) mlOutput;
+        final List<ModelTensors> tensorOutputList = modelTensorOutput.getMlModelOutputs();
+        if (CollectionUtils.isEmpty(tensorOutputList) || CollectionUtils.isEmpty(tensorOutputList.get(0).getMlModelTensors())) {
+            throw new IllegalStateException(
+                "Empty model result produced. Expected at least [1] tensor output and [1] model tensor, but got [0]"
+            );
+        }
+        List<ModelTensor> tensorList = tensorOutputList.get(0).getMlModelTensors();
+        String result = tensorList.get(0).getResult();
+        return result;
```
vibrantvarun:
```suggestion
       return tensorList.get(0).getResult().toString();
```

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -325,7 +340,8 @@ public void getModel(@NonNull final String modelId, @NonNull final ActionListene
      * Get model info for multiple model ids. It will send multiple getModel requests to get the model info in parallel.
      * It will fail if any one of the get model request fail. Only return the success result if all model info is
      * successfully retrieved.
-     * @param modelIds a set of model ids
+     *
+     * @param modelIds  a set of model ids
```
vibrantvarun:
```suggestion
     * @param modelIds a set of model ids
```

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -435,12 +451,56 @@ private void retryableInferenceSentenceHighlighting(
      * This method will highlight relevant sentences in the context based on the question.
      *
      * @param inferenceRequest the request containing the question and context for highlighting
-     * @param listener the listener to be called with the highlighting results
+     * @param listener         the listener to be called with the highlighting results
```
vibrantvarun:
```suggestion
     * @param listener to be called with the highlighting results
```

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -171,24 +181,31 @@ protected QueryBuilder doRewrite(QueryRewriteContext queryRewriteContext) throws
...
             parameters.put("query_fields", String.join(",", queryFields));
         }
 
-//        ML_CLIENT.executeAgent(agentId, parameters, ActionListener.wrap(future::complete, future::completeExceptionally));
+        ML_CLIENT.executeAgent(agentId, parameters, ActionListener.wrap(future::complete, future::completeExceptionally));
 
         try {
-            String dslQueryString = future.get();
-            // String dslQueryString = "{" + "\"match\": {" + "\"passage_text\": {" + "\"query\": \"science\"" + "}" + "}" + "}";
+            String agentResponse = future.get();
```
mingshl, line 201: when the prediction return no result, the string is "null" or "", we should return `multi-match` as default. That would give a much better user experience. Because if prediction result is null, in this case, it would throw out parsing exception. 

for example, the prediction result is null or empty. we should return a multi-match query
`
{
  "query": {
    "multi_match" : {
      "query":    "help me find a shoes", // the question from query_text
      "fields": [ "title", "*_name" ] // the fields from query_fields
    }
  }
}`

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -435,12 +451,56 @@ private void retryableInferenceSentenceHighlighting(
...
+        final ActionListener<String> listener
+    ) {
+        RemoteInferenceInputDataSet dataset = RemoteInferenceInputDataSet.builder().parameters(parameters).build();
+        AgentMLInput agentMLInput = new AgentMLInput(agentId, null, FunctionName.AGENT, dataset);
+        mlClient.execute(FunctionName.AGENT, agentMLInput, ActionListener.wrap(response -> {
+            try {
+                // Extract DSL query from inference results following the structure:
+                MLOutput mlOutput = (MLOutput) response.getOutput();
+                final String inferenceResults = buildQueryResultFromResponseOfOutput(mlOutput);
+                ;
```
vibrantvarun:
```suggestion
                
```

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
owaiskazi19: Shouldn't this be handled on the queryplannertool side if model doesn't return a query? On the query builder side, we should just focus on executing the query. Also, do we have a case when a LLM doesn't return a query?

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -171,24 +181,31 @@ protected QueryBuilder doRewrite(QueryRewriteContext queryRewriteContext) throws
...
+
```
mingshl, line 199: some basic validation should happen here, for example, when llm return the string not in starting with start object, then we can throw out meaningful exception, indicating the model prediction didn't return a query string, try indicating them to execute the agent execute to examine the prediction result. This is the case we should consider too.

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
vibrantvarun: I agree with @martin-gaievski here. For every query in OpenSearch we just use the first word. Like neural, hybrid, bool etc.

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,266 @@
...
+    public static final ParseField QUERY_FIELDS = new ParseField("query_fields");
+    private static final Gson gson = new Gson();
+
+    private String queryText;
+    private String agentId;
+    private List<String> queryFields;
+    private QueryBuilder rewrittenQuery;
+
+    // client to invoke ml-common APIs
+    private static MLCommonsClientAccessor ML_CLIENT;
```
vibrantvarun, line 72: QQ: Why ML_CLIENT is static?

vibrantvarun: It should be a instance variable as per my understanding.

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -224,4 +241,26 @@ public String getWriteableName() {
...
+        while ((token = parser.nextToken()) != XContentParser.Token.END_OBJECT) {
+            if (token == XContentParser.Token.FIELD_NAME && "query".equals(parser.currentName())) {
+                parser.nextToken(); // Move to the query value
+
+                // Use XContentBuilder to extract the inner query as JSON string
+                XContentBuilder builder = XContentBuilder.builder(XContentType.JSON.xContent());
+                builder.copyCurrentStructure(parser);
+                return builder.toString();
+            } else {
+                parser.skipChildren();
```
mingshl: wonder why skipChildren here? will this cause pitfall?

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
mingshl: +1

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
mingshl: I can make query planning tool to return multi_match query when model prediction is null or empty. 

But on the query side, I still think some validations should happen on the query side to double check the string getting from the agent response is a proper query string.

owaiskazi19: Sure, will add more validation on the query side

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,266 @@
...
+            return rewrittenQuery;
+        }
+
+        if (ML_CLIENT == null) {
+            throw new IllegalStateException("ML client not initialized");
+        }
+
+        // the queryRewrite is expected at QueryCoordinator level
+        if (!(queryRewriteContext instanceof QueryCoordinatorContext)) {
+            throw new IllegalStateException(
```
mingshl, line 175: this exception is properly never thrown if this query builder is actually working

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
mingshl: this exception is never thrown if this query builder is actually working. this is the same way like the template query builder works https://github.com/opensearch-project/OpenSearch/blob/5fd0a36a8f3dff8affb56d2bd687c3c3b2262d49/server/src/main/java/org/opensearch/index/query/TemplateQueryBuilder.java#L122 Ideally it will never reach here

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,266 @@
...
+                "Agentic query must be rewritten at the coordinator node. Rewriting at shard level is not supported."
+            );
+        }
+
+        QueryCoordinatorContext coordinatorContext = (QueryCoordinatorContext) queryRewriteContext;
+
+        // Get index mapping from cluster state
+        String indexMapping = NeuralSearchClusterUtil.instance().getIndexMapping(coordinatorContext);
+        String indexMappingJson = gson.toJson(indexMapping);
```
mingshl, line 187: Maybe not in this release, but later, we should give a default mapping_length_limit, to avoid very complicated indexMapping to blow out the agentExecute request.

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,266 @@
...
+                // Use XContentBuilder to extract the inner query as JSON string
+                XContentBuilder builder = XContentBuilder.builder(XContentType.JSON.xContent());
+                builder.copyCurrentStructure(parser);
+                return builder.toString();
+            } else {
+                parser.skipChildren();
+            }
+        }
+
+        throw new IOException("No 'query' field found in agent response");
```
mingshl: should log this error including the agent id?

src/test/java/org/opensearch/neuralsearch/query/AgenticSearchQueryIT.java
```diff
@@ -0,0 +1,184 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+        super.setUp();
+        updateClusterSettings();
+        if (TEST_AGENT_ID == null) {
+            try {
+                String connectorId = createTestConnector();
```
mingshl, line 33: when the AWS_ACCESS_KEY_ID is null, please skip the IT, because when it runs in local clusters, it cannot read the variables from github, this file will have AWS_ACCESS_KEY_ID is null

owaiskazi19: This is not done yet, that's why ignored IT until we have aws credentials

mingshl: right just point out the difference. because running on github and local cluster would get different variables value assigned.

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
owaiskazi19: Thanks for the feedback. Changed it to `agentic`

src/main/java/org/opensearch/neuralsearch/util/NeuralSearchClusterUtil.java
owaiskazi19: Thrown an exception

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
owaiskazi19: Since it is not a mandatory field, we won't add it to params then

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
owaiskazi19: Added an idempotent check to return the rewritten query

build.gradle
owaiskazi19: This is a dependency version and not opensearch dependency, so passing opensearch_version like 3.2.0 won't be right here.

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
owaiskazi19: We are initializing it in creacomponents [here](https://github.com/opensearch-project/neural-search/pull/1483/files#diff-0b5f74541932d2881f876b8452af44942e80abd0c316662a1d5e2c44810e74caR152) and that's how other [query builders has too ](https://github.com/search?q=repo%3Aopensearch-project%2Fneural-search%20static%20MLCommonsClientAccessor%20ML_CLIENT&type=code)

build.gradle
vibrantvarun: Yeah I mean to say there will be gson version as well. Version.properties.

owaiskazi19: We don't have gson in [core](https://github.com/opensearch-project/OpenSearch/blob/main/gradle/libs.versions.toml)

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
owaiskazi19: Added validation if the generated query is not wrapped inside a "query" field

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
owaiskazi19: Thrown an exception if the first field is not "query"

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
owaiskazi19: This needs to be discussed more, we should rely on tools of agent framework for index mapping rather from query side

owaiskazi19:
Since a new feature branch `feature/agentic-search` is created. Will close this and raise a new PR

Issue #1329, opened 2025-05-17
[FEATURE] Deploy Remote Model for Integrayion tests


owaiskazi19:
### Is your feature request related to a problem?
Currently, as part of integration, bwc etc tests we register a local model [here](https://github.com/opensearch-project/neural-search/blob/main/src/test/resources/processor/UploadModelRequestBody.json). This issue talks about supporting integration tests for remote model as those are the ones used in production environment.

### What solution would you like?
_A clear and concise description of what you want to happen._

### What alternatives have you considered?
No change

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

will-hwang:
which model should be used for integ test? For integ test, I think testing it against local model should be sufficient for our test cases

owaiskazi19:
@will-hwang ml-commons uses [remote model](https://github.com/opensearch-project/ml-commons/blob/main/plugin/src/test/java/org/opensearch/ml/rest/RestBedRockInferenceIT.java) for integration tests as well. Agreed, it is enough to test against a local model but on a production system we deploy a remote model so for our integ tests to behave the same on the CI, suggested to use remote model instead.

heemin32:
Should we do entire of our test using remote model or only the basic funtionality?

owaiskazi19:
> Should we do entire of our test using remote model or only the basic funtionality?

We cam use the remote model for basic functionality

andrross:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain) [2](https://github.com/andrross)

PR #1484, opened 2025-07-30
[Experimental] Added new query clause agentic search

owaiskazi19:
### Description
Added new query clause agentic search and tests

```
GET /<index_name>/_search?search_pipeline=agentic_pipeline
{
  "query": {
    "agentic": {
      "query_text": "Find me a red car",
      "query_fields": ["title", "description", "location"] //optional
    }
  }
}
```

```
PUT _search/pipeline/agentic_pipeline
{
     "request_processors": [
        {
            "agentic_query_translator": {
                "agent_id": "w8XuZJgBPU-lkq5jPnOF"
            }
        }
     ]
}
```
Settings to enable agentic search
```
{
  "transient": {
    "plugins.neural_search.agentic_search_enabled": "true"
  }
}
```

### Related Issues
https://github.com/opensearch-project/neural-search/issues/1479

### Check List


src/test/java/org/opensearch/neuralsearch/query/AgenticSearchQueryIT.java
```diff
@@ -0,0 +1,184 @@
...
+import java.util.Map;
+
+@AwaitsFix(bugUrl = "Ignoring until we find a way to fetch access, secret key for remote model")
+public class AgenticSearchQueryIT extends BaseNeuralSearchIT {
+
+    private static final String TEST_INDEX = "test-agentic-index";
+    private static String TEST_AGENT_ID;
+    private static String TEST_MODEL_ID;
+    private static final String TEST_QUERY_TEXT = "Find documents about machine learning";
+    private static final String AWS_ACCESS_KEY_ID = System.getenv("AWS_ACCESS_KEY_ID");
```
owaiskazi19: Setting up env would need https://github.com/opensearch-project/neural-search/issues/1329

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+        }
+
+        if (ML_CLIENT == null) {
+            throw new IllegalStateException("ML client not initialized");
+        }
+
+        // the queryRewrite is expected at QueryCoordinator level
+        if (!(queryRewriteContext instanceof QueryCoordinatorContext)) {
+            throw new IllegalStateException(
+                "Agentic query must be rewritten at the coordinator node. Rewriting at shard level is not supported."
```
heemin32: Just curious. How is it decided if rewrite is happening in coordinator level or shard level?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -236,6 +238,18 @@ private <T extends Number> List<List<T>> buildVectorFromResponse(MLOutput mlOutp
         return resultMaps;
     }
 
+    private String buildQueryResultFromResponseOfOutput(MLOutput mlOutput) {
+        final ModelTensorOutput modelTensorOutput = (ModelTensorOutput) mlOutput;
```
martin-gaievski, line 245: check for type before casting it to ModelTensorOutput

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -443,4 +458,47 @@ public void inferenceSentenceHighlighting(
...
+        @NonNull final Map<String, String> parameters,
+        @NonNull final ActionListener<String> listener
+    ) {
+        retryableExecuteAgent(agentId, parameters, 0, listener);
+    }
+
+    private void retryableExecuteAgent(
+        final String agentId,
+        final Map<String, String> parameters,
+        final int retryTime,
```
martin-gaievski, line 495: what's the default times we're retrying the call before we fail?

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+    public static final ParseField QUERY_FIELDS = new ParseField("query_fields");
+    private static final Gson gson = new Gson();
+
+    private String queryText;
+    private String agentId;
+    private List<String> queryFields;
+    private QueryBuilder rewrittenQuery;
+
+    // client to invoke ml-common APIs
+    private static MLCommonsClientAccessor ML_CLIENT;
```
martin-gaievski: why ml client is static?

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+                if (QUERY_TEXT_FIELD.match(currentFieldName, parser.getDeprecationHandler())) {
+                    agenticSearchQueryBuilder.queryText = parser.text();
+                } else if (AGENT_ID.match(currentFieldName, parser.getDeprecationHandler())) {
+                    agenticSearchQueryBuilder.agentId = parser.text();
+                } else {
+                    throw new ParsingException(parser.getTokenLocation(), "Unknown field [" + currentFieldName + "]");
+                }
+            } else if (token == XContentParser.Token.START_ARRAY) {
+                if (QUERY_FIELDS.match(currentFieldName, parser.getDeprecationHandler())) {
+                    List<String> fieldsList = new ArrayList<>();
```
martin-gaievski, line 131: we need to have a limit for number of fields we support, something big enough for potential use cases, maybe 25?

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
owaiskazi19: Since we are making the execute agent call to get the dsl from agent, performing the initial query rewriting on coordinator level would be optimized way and then all shards would execute the same request. Otherwise, if we rewrite on shard level then there will be shard specifc rewrites. 
OS already provides shard and coordinator context. With the check of `QueryCoordinatorContext` it ensures the ML calls happen on coordinator level.

src/main/java/org/opensearch/neuralsearch/util/NeuralSearchClusterUtil.java
```diff
@@ -65,4 +67,21 @@ public List<IndexMetadata> getIndexMetadataList(@NonNull final IndicesRequest se
             .map(concreteIndex -> clusterService.state().metadata().index(concreteIndex))
             .collect(Collectors.toList());
     }
+
+    public String getIndexMapping(QueryCoordinatorContext coordinatorContext) {
+        try {
+            String[] indices = coordinatorContext.getSearchRequest().indices();
+
+            if (indices != null && indices.length > 0) {
+                IndexMetadata indexMetadata = clusterService.state().metadata().index(indices[0]);
```
martin-gaievski: what is specifically this first index has a problem with mapping. Either aggregate mappings from all indexes, or document this limitation clearly.

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+@Getter(AccessLevel.PACKAGE)
+@Setter(AccessLevel.PACKAGE)
+@Log4j2
+public final class AgenticSearchQueryBuilder extends AbstractQueryBuilder<AgenticSearchQueryBuilder> implements WithFieldName {
+
+    public static final String NAME = "agentic";
+    public static final ParseField QUERY_TEXT_FIELD = new ParseField("query_text");
+    public static final ParseField AGENT_ID = new ParseField("agent_id");
+    public static final ParseField QUERY_FIELDS = new ParseField("query_fields");
+    private static final Gson gson = new Gson();
```
martin-gaievski: have you checked if there is already something in OpenSearch that does the same. I mean if we can minimize number of external dependencies that's good

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+            throw new IllegalStateException(
+                "Agentic query must be rewritten at the coordinator node. Rewriting at shard level is not supported."
+            );
+        }
+
+        QueryCoordinatorContext coordinatorContext = (QueryCoordinatorContext) queryRewriteContext;
+
+        // Get index mapping from cluster state
+        String indexMapping = NeuralSearchClusterUtil.instance().getIndexMapping(coordinatorContext);
+        String indexMappingJson = gson.toJson(indexMapping);
```
martin-gaievski: this can be a good candidate for caching, index mapping are pretty static objects

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
heemin32: Got it. So the OS already performs the rewrite at the coordinator level first. If, for some reason, this agentic query is nested inside another query and ends up being rewritten at the shard level instead, it will fail. This failure could be triggered by user input.

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -26,71 +24,47 @@
...
         this.queryText = in.readString();
-        this.agentId = in.readString();
         this.queryFields = in.readOptionalStringList();
     }
 
+    public String getQueryText() {
+        return queryText;
+    }
+
+    public List<String> getQueryFields() {
```
mingshl, line 69: if moving the rewrite functionality to the search request processors, basically this new query type just a new name that doesn't have any functionality. Should we still keep this query type then? 

We can run other query types with an agent search request processor to  achieve the same purpose. This new query type didn’t conduct any functionalities at all just wait for the search request processor to rewrite.

owaiskazi19: I will try to explain this. The reason we need a new query type is because it supports taking natural language query and query fields from the user. If not a new query clause, can you suggest which one can we use? `match`, `neural`, `hybrid` . Passing a natural language query to any of these queries would conflict with their own use case. Having a separate pipeline with an agent id can help users to run any request with the same agent id.
Also, it's not just a name, it takes a natural language queries with fields to query on. We can't have such query in search pipeline

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,165 @@
...
+        if (agenticSearchQueryBuilder.queryText == null || agenticSearchQueryBuilder.queryText.trim().isEmpty()) {
+            throw new ParsingException(parser.getTokenLocation(), "[" + QUERY_TEXT_FIELD.getPreferredName() + "] is required");
+        }
+
+        return agenticSearchQueryBuilder;
+    }
+
+    @Override
+    protected QueryBuilder doRewrite(QueryRewriteContext queryRewriteContext) throws IOException {
+        // No rewriting needed
```
mingshl, line 155: since no need to rewrite, we should throw exception here, so the query builder can save the effort travel through multiple doRewrite and reach to the doToQuery and finally throw exception. We knew the rewrite is handled by Processor then we should stop any doRewrite at this querybuilder

owaiskazi19: doRewrite is called during query parsing and validation, returning this would do no change in the flow. We need exception thrown in `doToQuery`  because that's where lucene query is generated, if `doToQuery` is run that means the search pipeline didn't run. So throwing an exception there would makes sens

src/main/java/org/opensearch/neuralsearch/processor/QueryRewriterProcessor.java
```diff
@@ -0,0 +1,157 @@
...
+import org.opensearch.core.action.ActionListener;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.opensearch.ingest.ConfigurationUtils.readStringProperty;
+
+@Log4j2
+public class QueryRewriterProcessor extends AbstractProcessor implements SearchRequestProcessor {
```
mingshl: Most of the search request processors are doing rewrite query, so this name "QueryRewriteProcessor" is not distinguished from other search request processors. Since its main functionality is to call agent to rewrite query, I would name it agent search request processor. I always want to have a agent search request processor. So this is a good chance to make it happened.

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
martin-gaievski: we do have queries like this, those are compound queries, like for instance, bool query. they are like containers for other nested/sub queries. Plus it's easy to parse request source if it's a new query clause. 
In long perspective, I think this isn't an actual query we we use this abstraction due to lack of a better choice. This needs a new abstraction, something like `question` instead of `query`

src/main/java/org/opensearch/neuralsearch/processor/QueryRewriterProcessor.java
owaiskazi19: > call agent to rewrite query

again won't justify the name `AgentProcessor`, since that would mean invoking any agent but this processor requires an agent with `QueryPlanningTool`. Looking for your inputs on this @martin-gaievski @heemin32

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
mingshl: if using search pipeline to invoke rewrite, maybe we can consider other options to make the agentic search clause in neural query clause, or multi-match query clause with an agentic search extension. 

I am thinking more about this. just an idea, 

maybe using multi-match query with agentic search extension would be a good idea. 

thinking normally you run multi match query like this 

```
{
  "query": {
    "multi_match" : {
      "query":    "this is a test",
      "fields": [ "subject", "message" ]
    }
  }
}
```

then if you want to use agent in rewriting your query, then you can attach a query extension and with a search pipeline to rewrite your query. 

```
{
  "query": {
    "multi_match": {
      "query": "this is a test",
      "fields": [
        "subject",
        "message"
      ]
    }
  },
  "ext": {
    "agentic_rewrite": {
      "agent_id": "agent123"
    }
  }
}
```

or if using neural query it would be 

```
"neural": {
  "<field>": {
    "query_text": "<query_text>",
    "agent_id": "agent_id",
  }
}
```

owaiskazi19: Agreed martin. In the long run there can be a case, where we are adding more agentic functionality to this query so we might need it as a query as well

src/main/java/org/opensearch/neuralsearch/processor/QueryRewriterProcessor.java
mingshl: naming is hard. any better brainstorm option in naming this processor? @ylwu-amzn

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
owaiskazi19: If we include agnetic extension with any query, the hits from the query and the extension needs to be merged. Also, the main purpose of agentic search is to help user write queries. We should keep agentic query as the top most query for the same purpose. The approach of multi match or other query would again add the work of creating the query.
Adding it with neural query like
```
"neural": {
  "<field>": {
    "query_text": "<query_text>",
    "agent_id": "agent_id",
  }
}
```
conflicts with the neural query definition

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -236,6 +238,18 @@ private <T extends Number> List<List<T>> buildVectorFromResponse(MLOutput mlOutp
...
+    private String buildQueryResultFromResponseOfOutput(MLOutput mlOutput) {
+        final ModelTensorOutput modelTensorOutput = (ModelTensorOutput) mlOutput;
+        final List<ModelTensors> tensorOutputList = modelTensorOutput.getMlModelOutputs();
+        if (CollectionUtils.isEmpty(tensorOutputList) || CollectionUtils.isEmpty(tensorOutputList.get(0).getMlModelTensors())) {
+            throw new IllegalStateException(
+                "Empty model result produced. Expected at least [1] tensor output and [1] model tensor, but got [0]"
+            );
+        }
+        List<ModelTensor> tensorList = tensorOutputList.get(0).getMlModelTensors();
+        return tensorList.get(0).getResult();
```
q-andy: Any reason to only use the first element? For other inference calls we iterate and add tensors from the whole list because the format of the model tensors (collected in the first element versus from all elements) depends on how we make the model call. 

E.g. for inference map results used for sparse encoding processors, a local model call will have all inference results in the first ModelTensors object, but for a batch remote call, the results are distributed across multiple ModelTensors, requiring us to iterate through all. See also https://github.com/opensearch-project/neural-search/blob/fa5b809b20d5b2b4d1c5ff7f6a4557845302fc0f/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java#L230-L235 for reference

src/main/java/org/opensearch/neuralsearch/processor/QueryRewriterProcessor.java
ylwu-amzn: Agree naming is hard.  Some options: `ai_query_translator` , `agentic_query_translator`, `query_translator`

owaiskazi19: +1 to `agentic_query_translator` or `agentic_query_rewriter`

q-andy (commented):
We should take an action item to add stats to this feature after it leaves experimental, similar to other query types and processor executions: https://docs.opensearch.org/latest/vector-search/api/neural/

src/main/java/org/opensearch/neuralsearch/util/NeuralSearchClusterUtil.java
bzhangam: Why we require indices but only use the first one to pull the index mapping?

src/main/java/org/opensearch/neuralsearch/processor/QueryRewriterProcessor.java
```diff
@@ -0,0 +1,157 @@
...
+        ActionListener<SearchRequest> requestListener
+    ) {
+        SearchSourceBuilder sourceBuilder = request.source();
+        if (sourceBuilder == null || sourceBuilder.query() == null) {
+            requestListener.onResponse(request);
+            return;
+        }
+
+        QueryBuilder query = sourceBuilder.query();
+        if (!(query instanceof AgenticSearchQueryBuilder)) {
```
bzhangam: We expect the agentic query to be the top level query? If that's the case we should do a validation here to catch the incorrect use otherwise users may think this rewriter is not working.

Besides do we plan to support it in bool query and hybrid query in future?

src/main/java/org/opensearch/neuralsearch/processor/QueryRewriterProcessor.java
```diff
@@ -0,0 +1,157 @@
...
+        SearchRequest request,
+        ActionListener<SearchRequest> requestListener
+    ) {
+        Map<String, String> parameters = new HashMap<>();
+        parameters.put("query_text", agenticQuery.getQueryText());
+
+        // Get index mapping from the search request
+        if (request.indices() != null && request.indices().length > 0) {
+            try {
+                String indexMapping = NeuralSearchClusterUtil.instance().getIndexMapping(request.indices());
```
bzhangam: We don't plan to support searching against multiple indices? Here we assume we will only use the first index.

src/main/java/org/opensearch/neuralsearch/processor/QueryRewriterProcessor.java
```diff
@@ -0,0 +1,157 @@
...
+                }
+
+                requestListener.onResponse(request);
+            } catch (IOException e) {
+                log.error("Failed to parse agent response", e);
+                requestListener.onFailure(e);
+            }
+        }, e -> {
+            log.error("Failed to execute agent", e);
+            requestListener.onFailure(new IOException("Failed to execute agentic search", e));
```
bzhangam: Why we assume it's IOException? What if the exception is from the call to ml client? Would RuntimeException be better?

src/main/java/org/opensearch/neuralsearch/processor/QueryRewriterProcessor.java
heemin32: Agree that QueryRewriterProcessor is too generic. Maybe, AgenticQueryBuilderProcessor?

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
bzhangam: I think it's not common that we only rely on a search request processor to rewrite a type of a query. Should we do similar thing like the neural query that we allow users to provide the agent id through the query itself and do rewrite logic here? And we can make a processor like the `NeuralQueryEnricherProcessor` to inject the `agent id` to the agentic query if we want to reuse the id for different queries. In this way we can decouple the agent query and the rewrite processor to give users more flexibility.

src/main/java/org/opensearch/neuralsearch/processor/QueryRewriterProcessor.java
heemin32: ~Anyway, this will be hidden from users and only be used by agentic query clause internally right?~

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
bzhangam: I also think using a new query type is better for long term. We may want to add more agentic specific parameters in future which doesn't make sense to be added to the existing query type. Reusing the existing query type can limit the future extendability and also introduce unnecessary complexity to the existing query type.

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
bzhangam: Synced offline that since the query generated by the LLM can include something like the aggregation which cannot be fully handled by a query builder so we have to modify the whole original query request in the search request processor. I think we should add a comment here to specify where will we do the rewrite and why we need to do it that way.

src/main/java/org/opensearch/neuralsearch/processor/QueryRewriterProcessor.java
```diff
@@ -0,0 +1,157 @@
...
+
+        mlClient.executeAgent(agentId, parameters, ActionListener.wrap(agentResponse -> {
+            try {
+                log.info("Generated Query: [{}]", agentResponse);
+
+                // Parse the agent response to get the new search source
+                BytesReference bytes = new BytesArray(agentResponse);
+                try (XContentParser parser = XContentType.JSON.xContent().createParser(xContentRegistry, null, bytes.streamInput())) {
+                    SearchSourceBuilder newSourceBuilder = SearchSourceBuilder.fromXContent(parser);
+                    request.source(newSourceBuilder);
```
bzhangam: We generate the query only based on the info in the query part but here we use the response to override the whole request source so it's possible we can lose some query info. e.g. The original query can have a highlighter but after this override it's gone. We should either block user using agentic query with other functions or we should merge the query generated by the LLM and the original query.

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
bzhangam: This agentic query is actually not just about the query field since LLM can generate the search request with other fields like `sort` and `aggs`. But the query builder is designed to just handle the `query` field and this also force us to use a search request processor to rewrite the whole search request.

Ideally I think we should add a new API for this function but to keep it as part of the search API can have more people to try it out. And plugin also cannot easily add a new parameter to the search API. So probably it's better to consider using the `ext` field to take input and trigger the rewrite logic since it's a reserved namespace for plugin-specific parameters.

Potentially we can do something like the below and block using the query field and the agentic at the same time.
```
{
  "ext": {
    "agentic": {
      "question": "this is a test",
      "fields": [
        "subject",
        "message"
      ],
      "agent_id": "agent123"
    }
  }
}
```

Then we can still use a search request processor to rewrite it by checking if the request has agentic in the ext. Or we even can explore if we can systematically generate search request processor to rewrite the search request to save the effort for users to set up the search pipeline.

owaiskazi19: We considered this approach, but agentic search fundamentally represents a natural language query from the user that gets transformed into structured search logic. The ext field is designed for extensions that enhance existing functionality of the search query, whereas agentic search is the primary query mechanism. Using a query builder better reflects this semantic distinction - the user is providing their actual query in natural language, not adding an extension to modify some other query.
Still open to feedback here to support `agentic` as query builder or `ext`. @martin-gaievski @heemin32 @navneet1v @ylwu-amzn @sean-zheng-amazon

bzhangam: Yeah semantically the natural language query is still a query but in the search API it actually works more like a search request since it can contain the sort and aggregation info. I agree the ext may also not be the best option. Since user doesn’t care about how our query builder works using a new query type actually can be a better UX. And if we can avoid manually setting up the search pipeline it can be better.

owaiskazi19: Yes, in the long run, we can actually create system search pipeline with this processor. One major thing to consider over there is we have to create the pipeline only for the first request and check for the later ones to not create others to avoid the race condition.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
owaiskazi19: 0

src/main/java/org/opensearch/neuralsearch/util/NeuralSearchClusterUtil.java
owaiskazi19: Added support for multiple indices

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
owaiskazi19: yes, core doesn't have gson

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
owaiskazi19: Added caching

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
owaiskazi19: We are expecting a string from the model which is mentioned in the default prompt on the agent side. To handle it, I have added iterations over tensor list rather fetching the first element

src/main/java/org/opensearch/neuralsearch/processor/QueryRewriterProcessor.java
owaiskazi19: Added the support

src/main/java/org/opensearch/neuralsearch/processor/QueryRewriterProcessor.java
owaiskazi19: The doToQuery of the builder handles agentic query as top level query and the processor, only processes top-level agentic query. And if processor is not executed because the top level query is not agentic then the flow goes to doToQuery which will throw an exception. I have added the validation and test for the same

src/main/java/org/opensearch/neuralsearch/processor/QueryRewriterProcessor.java
owaiskazi19: This is a good callout and yes we don't want to support any other search features because those will be generated by the LLM queries like
```
{"size":0.0,"query":{"term":{"species.keyword":"setosa"}},"aggs":{"setosa_count":{"value_count":{"field":"species.keyword"}}}}
```
 I have added validation to block such features and added tests

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
owaiskazi19: Added the limit of 25 and a test

src/main/java/org/opensearch/neuralsearch/processor/QueryRewriterProcessor.java
owaiskazi19: Named the processor `agentic_query_translator`

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -236,6 +238,33 @@ private <T extends Number> List<List<T>> buildVectorFromResponse(MLOutput mlOutp
...
+            throw new IllegalStateException("Expected ModelTensorOutput but got: " + mlOutput.getClass().getSimpleName());
+        }
+        final ModelTensorOutput modelTensorOutput = (ModelTensorOutput) mlOutput;
+
+        final List<ModelTensors> tensorOutputList = modelTensorOutput.getMlModelOutputs();
+        if (CollectionUtils.isEmpty(tensorOutputList)) {
+            throw new IllegalStateException("Empty model result produced. Expected at least [1] tensor output, but got [0]");
+        }
+
+        // Iterate through all ModelTensors to find the DSL result
```
bzhangam, line 252: Here I think we simply try to find the first non-empty result and then use it. Is it possible that the first non-empty result is not the DSL result we want?

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
```diff
@@ -0,0 +1,188 @@
...
+        ActionListener<SearchRequest> requestListener
+    ) {
+        Map<String, String> parameters = new HashMap<>();
+        parameters.put("query_text", agenticQuery.getQueryText());
+
+        // Get index mapping from the search request
+        if (request.indices() != null && request.indices().length > 0) {
+            try {
+                String cacheKey = String.join(",", request.indices());
+                String cachedMappingJson = indexMappingCache.get(cacheKey);
```
bzhangam: If we want to use cache here we need to consider:
1. When to invalidate the cache to avoid using the stale index mapping info?
2. How to limit the size of the cache to avoid using too much memory?

Probably it's better to not use the cache here for now since the operation is simply read the index mapping from the state which is done locally.

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,168 @@
...
+        return this;
+    }
+
+    @Override
+    protected Query doToQuery(QueryShardContext context) throws IOException {
+        throw new IllegalStateException("Agentic search query must be used as top-level query, not nested inside other queries");
```
bzhangam: We should also call out it should be used with the translator processor otherwise even it's used as the top-level query we still can run into this exception.

src/main/java/org/opensearch/neuralsearch/util/NeuralSearchClusterUtil.java
```diff
@@ -65,4 +66,23 @@ public List<IndexMetadata> getIndexMetadataList(@NonNull final IndicesRequest se
...
             .collect(Collectors.toList());
     }
+
+    public List<String> getIndexMapping(String[] indices) {
+        try {
+            if (indices != null && indices.length > 0) {
+                return Arrays.stream(indices).map(indexName -> {
+                    IndexMetadata indexMetadata = clusterService.state().metadata().index(indexName);
+                    if (indexMetadata != null) {
+                        String mapping = Objects.requireNonNull(indexMetadata.mapping()).source().toString();
```
bzhangam: Is it possible that indexMetadata is not null but the mapping is null? e.g. An index just with some settings. Probably it's better to also check the mapping is not null?

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -0,0 +1,168 @@
...
+    @Override
+    protected void doWriteTo(StreamOutput out) throws IOException {
+        out.writeString(this.queryText);
+        out.writeOptionalStringCollection(this.queryFields);
+    }
+
+    @Override
+    protected void doXContent(XContentBuilder xContentBuilder, Params params) throws IOException {
+        xContentBuilder.startObject(NAME);
+        if (Objects.nonNull(QUERY_TEXT_FIELD)) {
```
martin-gaievski, line 94: this must be the `queryText` I suppose

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
martin-gaievski: +1, such cache can grow and cause resource utilization issues, if the bounded/eviction strategy is missing or weak. We need something like limit of max elements or TTL

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
```diff
@@ -0,0 +1,188 @@
...
+
+                if (cachedMappingJson == null) {
+                    List<String> indexMappings = NeuralSearchClusterUtil.instance().getIndexMapping(request.indices());
```
martin-gaievski: this place is a potential race condition point

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
```diff
@@ -0,0 +1,188 @@
...
+        if (agenticQuery.getQueryFields() != null && !agenticQuery.getQueryFields().isEmpty()) {
+            parameters.put("query_fields", String.join(",", agenticQuery.getQueryFields()));
+        }
+
+        mlClient.executeAgent(agentId, parameters, ActionListener.wrap(agentResponse -> {
+            try {
+                log.info("Generated Query: [{}]", agentResponse);
+
+                // Parse the agent response to get the new search source
+                BytesReference bytes = new BytesArray(agentResponse);
```
martin-gaievski, line 126: my understanding we're fully relying on the agent for correctness of the returned query. This can be risky, also how about the query complexity and size, should we allow execution of any query?
It's ok if this is experimental, but at least add a new issue and TODO so we can return to this later. Main areas should be: query validation, clarification if query is expensive, maybe CB or similar mechanism to prevent system crushes

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -236,6 +238,33 @@ private <T extends Number> List<List<T>> buildVectorFromResponse(MLOutput mlOutp
         return resultMaps;
     }
 
+    private String buildQueryResultFromResponseOfOutput(MLOutput mlOutput) {
```
martin-gaievski, line 241: this whole method works on assumptions regarding the format of the output from ml client. Do we have a clear API format/contract between ml agent and its clients? If not we need to create one to avoid regressions in future

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
owaiskazi19: The race condition risk can be low to moderate because:
ClusterState is immutable after we get the snapshot
Index deletion is rare during active search operations
Error handling exists - Exceptions are caught and logged

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
owaiskazi19: This is query_text `public static final ParseField QUERY_TEXT_FIELD = new ParseField("query_text")`

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
owaiskazi19: Removed caching here

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
owaiskazi19: This is [output format](https://github.com/opensearch-project/ml-commons/pull/4046/files#diff-760b7ef67a60f4ff3946ce84552b96e12eeb95aa2c0a202d5ded9867118fac4dR34-R43) of the prompt given to the LLM

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
owaiskazi19: [Output format](https://github.com/opensearch-project/ml-commons/pull/4046/files#diff-760b7ef67a60f4ff3946ce84552b96e12eeb95aa2c0a202d5ded9867118fac4dR34-R43) of the prompt

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
owaiskazi19: Added TODO

bzhangam (approved):

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
martin-gaievski: what I'm trying to say - there is no strict enforcement for ml-commons part to follow the format, like classes that will fail in compile time etc. If they make a change we'll know it in runtime only. I guess that's fine for experimental, but should be an action item in next releases

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
martin-gaievski: while those points about immutability of cluster state and error handling are valid, I still think we need to address the race condition. my point was not about data corruption but about the efficiency:
- multiple threads doing duplicate work (get mapping, json serialization)
- CPU and memory usage
- potential latency increase under high concurrency

solution can be pretty simple:

```
String cachedMappingJson = indexMappingCache.computeIfAbsent(cacheKey, key -> {
    // single computation guaranteed
    List<String> indexMappings = NeuralSearchClusterUtil.instance()
            .getIndexMapping(request.indices());
   ....
});
```

owaiskazi19:
@martin-gaievski @bzhangam merged feature flag change in https://github.com/opensearch-project/neural-search/pull/1484/commits/246078b75d71394efa81430cb11e8af6f184a4e2

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
owaiskazi19: I have removed the caching as asked [here](https://github.com/opensearch-project/neural-search/pull/1484#discussion_r2251967500). Should I add it back with the above solution?

martin-gaievski: Oh I see, I was looking into outdated code. Nope, if the whole construct got removed let's skip my comment

martin-gaievski (approved):
Looks good to me, thanks for addressing all comments

martin-gaievski:
CI action for "Security enabled integ test" is failing, I think that's unrelated. Created separate issue for it: https://github.com/opensearch-project/neural-search/issues/1496

Issue #1489, opened 2025-07-31
[Flaky Test] HybridQueryAggregationsIT testNestedAggs_whenMultipleShardsAndConcurrentSearchEnabled_thenSuccessful flaky for Multi-Node IT


ryanbogan:
Flaky error:
```
java.lang.AssertionError: expected:<0> but was:<26>
	at __randomizedtesting.SeedInfo.seed([5F1228FA0C6ACE76:CA18953153DE4846]:0)
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.junit.Assert.assertEquals(Assert.java:633)
	at org.opensearch.neuralsearch.BaseNeuralSearchIT.bulkIngest(BaseNeuralSearchIT.java:1292)
	at org.opensearch.neuralsearch.query.HybridQueryAggregationsIT.prepareResourcesForNestegAggregationsScenario(HybridQueryAggregationsIT.java:499)
	at org.opensearch.neuralsearch.query.HybridQueryAggregationsIT.testNestedAggs_whenMultipleShardsAndConcurrentSearchEnabled_thenSuccessful(HybridQueryAggregationsIT.java:475)
```

PR #1503, opened 2025-08-07
Fix flaky test in HybridQueryAggregationsIT

### LINKED ISSUE #1488: [Flaky Test] HybridQueryAggregationsIT testNestedAggs_whenMultipleShardsAndConcurrentSearchDisabled_thenSuccessful flaky for Multi-Node IT
Flaky error:
```
java.lang.AssertionError: expected:<42> but was:<84>
	at __randomizedtesting.SeedInfo.seed([E6CDBE73F2980B06:FA5B74F3F47A04AB]:0)
	at org.junit.Assert.assertEquals(Assert.java:120)
	at org.junit.Assert.assertEquals(Assert.java:146)
	at org.opensearch.neuralsearch.query.HybridQueryAggregationsIT.assertNestedAggregations(HybridQueryAggregationsIT.java:616)
	at org.opensearch.neuralsearch.query.HybridQueryAggregationsIT.testNestedAggs_whenMultipleShardsAndConcurrentSearchDisabled_thenSuccessful(HybridQueryAggregationsIT.java:469)
```

chishui:
### Description
The test fails because when preparing documents using `bulkIngest` function, we may get `primary shard is not active` issue. The fix adds a `waitForClusterHealthGreen` check before running `bulkIngest`. Also update `waitForClusterHealthGreen` to wait for active shard as well.

Verified through:
```
./gradlew integTest -PnumNodes=3  -Dtests.iters=10 --tests "org.opensearch.neuralsearch.query.HybridQueryAggregationsIT.testNestedAggs_whenMultipleShardsAndConcurrentSearchDisabled_thenSuccessful"
```

### Related Issues
Resolves #1488 #1489 

### Check List

### DIFF
--- a/src/test/java/org/opensearch/neuralsearch/query/HybridQueryAggregationsIT.java
+++ b/src/test/java/org/opensearch/neuralsearch/query/HybridQueryAggregationsIT.java
@@ -497,7 +497,7 @@ public class HybridQueryAggregationsIT extends BaseNeuralSearchIT {
                 .replace("\"{indexname}\"", "\"" + index + "\"");
 
             // wait until cluster is healthy before running ingest to reduce flakiness
-            waitForClusterHealthGreen(numOfNode, 100);
+            waitForClusterHealthGreen(numOfNodes);
             bulkIngest(ingestBulkPayload, null);
         }
         createSearchPipelineWithResultsPostProcessor(SEARCH_PIPELINE);
--- a/src/test/java/org/opensearch/neuralsearch/query/HybridQueryAggregationsIT.java
+++ b/src/test/java/org/opensearch/neuralsearch/query/HybridQueryAggregationsIT.java
@@ -497,7 +497,7 @@ public class HybridQueryAggregationsIT extends BaseNeuralSearchIT {
                 .replace("\"{indexname}\"", "\"" + index + "\"");
 
             // wait until cluster is healthy before running ingest to reduce flakiness
-            waitForClusterHealthGreen(numOfNode, 60);
+            waitForClusterHealthGreen(numOfNode, 100);
             bulkIngest(ingestBulkPayload, null);
         }
         createSearchPipelineWithResultsPostProcessor(SEARCH_PIPELINE);
--- a/src/test/java/org/opensearch/neuralsearch/query/HybridQueryAggregationsIT.java
+++ b/src/test/java/org/opensearch/neuralsearch/query/HybridQueryAggregationsIT.java
@@ -497,7 +497,7 @@ public class HybridQueryAggregationsIT extends BaseNeuralSearchIT {
                 .replace("\"{indexname}\"", "\"" + index + "\"");
 
             // wait until cluster is healthy before running ingest to reduce flakiness
-            waitForClusterHealthGreen(numOfNode, 60, "all");
+            waitForClusterHealthGreen(numOfNode, 60);
             bulkIngest(ingestBulkPayload, null);
         }
         createSearchPipelineWithResultsPostProcessor(SEARCH_PIPELINE);
--- a/src/test/java/org/opensearch/neuralsearch/query/HybridQueryAggregationsIT.java
+++ b/src/test/java/org/opensearch/neuralsearch/query/HybridQueryAggregationsIT.java
@@ -496,6 +496,8 @@ public class HybridQueryAggregationsIT extends BaseNeuralSearchIT {
             String ingestBulkPayload = Files.readString(Path.of(classLoader.getResource("processor/ingest_bulk.json").toURI()))
                 .replace("\"{indexname}\"", "\"" + index + "\"");
 
+            // wait until cluster is healthy before running ingest to reduce flakiness
+            waitForClusterHealthGreen(numOfNode, 60, "all");
             bulkIngest(ingestBulkPayload, null);
         }
         createSearchPipelineWithResultsPostProcessor(SEARCH_PIPELINE);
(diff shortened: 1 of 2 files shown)


src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -2025,11 +2026,18 @@ protected String registerModelGroup(final String modelGroupRegisterRequestBody)
...
+        waitForClusterHealthGreen(numOfNodes, timeoutInSeconds, "0");
+    }
+
+    // Method that waits till the health of nodes in the cluster goes green
+    protected void waitForClusterHealthGreen(final String numOfNodes, final int timeoutInSeconds, final String activeShards)
+        throws IOException {
         Request waitForGreen = new Request("GET", "/_cluster/health");
         waitForGreen.addParameter("wait_for_nodes", numOfNodes);
         waitForGreen.addParameter("wait_for_status", "green");
+        waitForGreen.addParameter("wait_for_active_shards", activeShards);
```
heemin32: Can we just use "all"

```suggestion
        waitForGreen.addParameter("wait_for_active_shards", "all");
```

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -151,13 +151,14 @@ public abstract class BaseNeuralSearchIT extends OpenSearchSecureRestTestCase {
...
     private static final long TIMEOUT = 10000;
+    protected String numOfNode;
 
     @Before
     public void setupSettings() {
         threadPool = setUpThreadPool();
         clusterService = createClusterService(threadPool);
         final IndexNameExpressionResolver indexNameExpressionResolver = new IndexNameExpressionResolver(new ThreadContext(Settings.EMPTY));
-
+        numOfNode = System.getProperty("cluster.number_of_nodes", "1");
```
heemin32: Why do we set the number of node as 1 here?

chishui: oh, it's the default value if numNodes is not set

chishui: in multi-node testing from workflow or locally, this could be any number

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
chishui: It's "0" by default for health API, don't want to change the behavior for all other unrelated usage. WDYT?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
heemin32: Why do we need to set it explicitly if it is already a default value?

src/test/java/org/opensearch/neuralsearch/query/HybridQueryAggregationsIT.java
```diff
@@ -496,6 +496,8 @@ private void prepareResourcesForNestegAggregationsScenario(String index) throws
             String ingestBulkPayload = Files.readString(Path.of(classLoader.getResource("processor/ingest_bulk.json").toURI()))
                 .replace("\"{indexname}\"", "\"" + index + "\"");
 
+            // wait until cluster is healthy before running ingest to reduce flakiness
+            waitForClusterHealthGreen(numOfNode, 60, "all");
```
heemin32: Why this test is special that it have to pass "all" while others test does not pass the same value?

chishui: Your point makes sense, I'll change "all" as default

chishui:
cluster status could be yellow as replica shard can't be assigned to the same node as primary shard when we run single node test:
```
  1> [2025-08-15T09:41:46,101][INFO ][o.o.n.BaseNeuralSearchIT ] [testNestedAggs_whenMultipleShardsAndConcurrentSearchDisabled_thenSuccessful] retry: 2
  1> [2025-08-15T09:42:26,120][INFO ][o.o.n.q.HybridQueryAggregationsIT] [testNestedAggs_whenMultipleShardsAndConcurrentSearchDisabled_thenSuccessful] {"index":"test-hybrid-
nested-aggs-multi-doc-index","shard":0,"primary":false,"current_state":"unassigned","unassigned_info":{"reason":"INDEX_CREATED","at":"2025-08-15T07:40:35.279Z","last_allocat
ion_status":"no_attempt"},"can_allocate":"no","allocate_explanation":"cannot allocate because allocation is not permitted to any of the nodes","node_allocation_decisions":[{
"node_id":"NYlCvB51Qx6oT7ZNJRrncw","node_name":"integTest-0","transport_address":"127.0.0.1:49498","node_attributes":{"testattr":"test","shard_indexing_pressure_enabled":"tr
ue"},"node_decision":"no","weight_ranking":1,"deciders":[{"decider":"same_shard","decision":"NO","explanation":"a copy of this shard is already allocated to this node [[test
-hybrid-nested-aggs-multi-doc-index][0], node[NYlCvB51Qx6oT7ZNJRrncw], [P], s[STARTED], a[id=6HuvC2OPTV6EbpPrX2W_ag]]"}]}]}
```

I add a logic to change num_of_replica to 0 if there is only one node.

PR #1513, opened 2025-08-14
Support the sparse two phase processor for the semantic field.

### LINKED ISSUE #1352: [FEATURE] Semantic Field Enhancement - Support neural_sparse_two_phase_processor
In the initial launch we allow users to use a sparse model for the semantic field and later use neural query to query against the sparse embedding. But we have neural_sparse_two_phase_processor which can speed up the search against the sparse embedding by dividing the original method of scoring all documents with all tokens into two steps. We should also support it when users try to use a neural query to search against the sparse embedding generated for a semantic field.

Refactor the neural_sparse_two_phase_processor to also support the neural query against a sparse embedding of a semantic field.

N/A

bzhangam:
### Description
Support the sparse two phase processor for the semantic field.

### Related Issues
Resolves #1352 

### Check List


src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -164,17 +158,18 @@ public static void initialize(MLCommonsClientAccessor mlClient) {
     @Getter(AccessLevel.PACKAGE)
     @Setter(AccessLevel.PACKAGE)
     private Supplier<float[]> vectorSupplier;
-    private QueryBuilder filter;
+    private QueryBuilder queryfilter;
```
bzhangam, line 161: Rename it to avoid lombok Accessors auto generated function conflicting with the existing filter(QueryBuilder filterToBeAdded) function.

src/main/java/org/opensearch/neuralsearch/query/AbstractNeuralQueryBuilder.java
```diff
@@ -0,0 +1,123 @@
...
+    public String fieldName() {
+        return fieldName;
+    }
+
+    public QB fieldName(String fieldName) {
+        this.fieldName = fieldName;
+        return (QB) this;
+    }
+
+    public final String queryText() {
```
bzhangam, line 74: Need to define the set and get functions for the fields manually because there is an issue for lombok Accessors to auto generate these functions when we have subclass and parent abstract class structure.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -1207,12 +1279,41 @@ public String getWriteableName() {
...
+        // If raw tokens are provided directly in the query, split them without additional processing
+        if (Objects.nonNull(this.queryTokensMapSupplier)) {
+            Map<String, Float> tokens = queryTokensMapSupplier.get();
+            // Splitting tokens based on a threshold value: tokens greater than the threshold are stored in v1,
+            // while those less than or equal to the threshold are stored in v2.
+            Tuple<Map<String, Float>, Map<String, Float>> splitTokens = PruneUtils.splitSparseVector(pruneType, pruneRatio, tokens);
+            this.queryTokensMapSupplier(splitTokens::v1);
+            copy.queryTokensMapSupplier(splitTokens::v2);
+        } else {
+            copy.modelIdToTwoPhaseSharedQueryTokenSupplier(() -> modelIdToTwoPhaseSharedQueryToken);
```
bzhangam, line 1319: Need to use a supplier to delay setting the modelIdToTwoPhaseSharedQueryToken for the copy since now we haven't invoked the inference call to generate the tokens for the original query builder.

src/main/java/org/opensearch/neuralsearch/query/AbstractNeuralQueryBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+
+    // both dense and sparse embedding can use
+    protected String modelId;
+    protected String fieldName;
+    protected String queryText;
+
+    // sparse embedding only
+    protected String searchAnalyzer;
+    protected Supplier<Map<String, Float>> queryTokensMapSupplier;
+    protected NeuralSparseQueryTwoPhaseInfo neuralSparseQueryTwoPhaseInfo = new NeuralSparseQueryTwoPhaseInfo();
```
heemin32, line 47: For sparse embedding only field, I think it is better to define it in sparse query builder.
Also, see if we can inject `NeuralSparseQueryTwoPhaseInfo` from outside.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -201,6 +196,9 @@ public static class Builder {
         private Map<String, Supplier<Map<String, Float>>> modelIdToQueryTokensSupplierMap;
         private Boolean isSemanticField = false;
         private NeuralQueryBuildStage buildStage;
+        private NeuralSparseQueryTwoPhaseInfo neuralSparseQueryTwoPhaseInfo = new NeuralSparseQueryTwoPhaseInfo();;
```
heemin32: Can we delay the creation of this instance by creating it inside `build()` method so that we can avoid creation of this instance when it is provided from outside?

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
heemin32, line 1319: For this case, we don't need to update `this.queryTokensMapSupplier`?

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -1207,12 +1279,41 @@ public String getWriteableName() {
...
+                new NeuralSparseQueryTwoPhaseInfo(NeuralSparseQueryTwoPhaseInfo.TwoPhaseStatus.PHASE_TWO, pruneRatio, pruneType)
+            );
+
+        // If raw tokens are provided directly in the query, split them without additional processing
+        if (Objects.nonNull(this.queryTokensMapSupplier)) {
+            Map<String, Float> tokens = queryTokensMapSupplier.get();
+            // Splitting tokens based on a threshold value: tokens greater than the threshold are stored in v1,
+            // while those less than or equal to the threshold are stored in v2.
+            Tuple<Map<String, Float>, Map<String, Float>> splitTokens = PruneUtils.splitSparseVector(pruneType, pruneRatio, tokens);
+            this.queryTokensMapSupplier(splitTokens::v1);
```
heemin32, line 1314: nit. Not ideal because the method name indicate just getting copy of query builder but this actually modify this.queryTokensMapSupplier.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
junqiu-lei: How about use @Accessors(fluent = false) to disable fluent accessors for conflicting fields?

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
junqiu-lei: Remove extra semicolon

src/main/java/org/opensearch/neuralsearch/query/AbstractNeuralQueryBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Abstract class defines the common fields and behavior shared by the neural query builder and the neural sparse query
+ * builder.
+ * @param <QB>
+ */
+@Getter
+@Setter
+@NoArgsConstructor
+public abstract class AbstractNeuralQueryBuilder<QB extends AbstractNeuralQueryBuilder<QB>> extends AbstractQueryBuilder<QB>
```
junqiu-lei, line 29: We might want to take a eye for the code change performance once merged from nightly run benchmark for [Neural Sparse Search](https://benchmarks.opensearch.org/app/dashboards?#/view/01b25d30-5216-11f0-b39d-2363e77eb224?embed=true&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-7d,to:now))&_a=(description:'',filters:!(),fullScreenMode:!f,options:(hidePanelTitles:!f,useMargins:!t),query:(language:kuery,query:''),timeRestore:!f,title:'Neural%20Search%20Sparse%20Search%20Dashboard',viewMode:view)&show-time-filter=true&hide-filter-bar=true) and [Neural Semantic Search](https://benchmarks.opensearch.org/app/dashboards?#/view/ee8742a0-52b7-11f0-b39d-2363e77eb224?embed=true&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-7d,to:now))&_a=(description:'',filters:!(),fullScreenMode:!f,options:(hidePanelTitles:!f,useMargins:!t),query:(language:kuery,query:''),timeRestore:!f,title:'Neural%20Search%20Semantic%20Search%20Dashboard',viewMode:view)&show-time-filter=true&hide-filter-bar=true)

src/main/java/org/opensearch/neuralsearch/query/AbstractNeuralQueryBuilder.java
bzhangam: These fields are for sparse embedding but will be used by both the neural query builder and neural sparse query builder. Because neural query builder needs to support the semantic field with the sparse embedding.

We are injecting the neuralSparseQueryTwoPhaseInfo from outside. Here is just a default value to avoid null.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
bzhangam: sure

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
bzhangam: yes. we delay the update to the later phase when the tokens are available.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
bzhangam: I can rename it.

src/main/java/org/opensearch/neuralsearch/query/AbstractNeuralQueryBuilder.java
bzhangam: It should not impact them if we don't enable the sparse two phase processor.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
heemin32: Might be better to have as comment in the code?

src/main/java/org/opensearch/neuralsearch/query/AbstractNeuralQueryBuilder.java
heemin32: Got it. I am thinking if there might be a better way. Like, we have dense vector query, and sparse vector query. Than, semantic text query has both dense query builder and sparse query builder as variable to delegate. Just my thought.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
bzhangam: We still need a function to set and get the filter so it's better to rename it to continue to use accessors.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
bzhangam: ok

heemin32: Can we somehow consolidate the logic so that even for the case when token is already available, can we delay the update to the later phase so that both logic consistent in update behavior each other?
By looking at the code, it is hard to understand when the splitting is happening for the second case.

Issue #1516, opened 2025-08-15
[FEATURE] Batch Inference Support for Semantic Highlighting


junqiu-lei:
## Problem Statement

The current semantic highlighting implementation processes documents sequentially with individual ML inference calls per document. While this approach works well for small result sets, it presents optimization opportunities for larger search results. When processing multiple documents that require semantic highlighting, each document triggers a separate ML inference request, which doesn't fully utilize the batch processing capabilities (example: AWS Sagemaker GPU based endpoint). This creates potential for performance improvements, especially in scenarios with high document counts or concurrent search requests.

## Solution Overview

The proposed batch inference approach transforms semantic highlighting to process multiple documents in a single ML inference call. Instead of individual calls per document, the system collects all documents requiring highlighting within a shard and processes them together in one batch request. This leverages the SearchResponseProcessor framework for asynchronous batch processing, where the processor collects text content, constructs a batch inference request, and maps results back to corresponding documents. Performance improvements reduce ML inference overhead from N individual calls to 1 batch call per shard.

## Technical Implementation

The feature builds on OpenSearch capabilities for the System Generated Search Pipeline approach outlined in https://github.com/opensearch-project/OpenSearch/issues/19062, which enables automatic processor creation from highlight configurations. ML Commons already provides batch inference capabilities for remote models, so no additional ML infrastructure changes are required.

The Neural Search plugin will implement SemanticHighlightingResponseProcessor with batch inference logic. It can have response-level document batching, configurable batch processing modes through a `batch_inference` parameter. The processor will leverage existing ML Commons batch inference APIs through MLCommonsClientAccessor to process multiple documents efficiently in a single inference call.

## API Compatibility

The feature maintains complete backward compatibility with existing semantic highlighting API. Users continue using familiar highlight configuration syntax with no changes required to existing queries or applications. For existing models that don't support batch inference, the behavior remains unchanged with individual document processing. Batch processing is only enabled when using models that support batch inference capabilities and when explicitly configured with the `batch_inference` parameter.

```json
{
  "query": { "match": { "text": "symptoms" } },
  "highlight": {
    "fields": { "text": { "type": "semantic" } },
    "options": { 
      "model_id": "your-model-id",
      "batch_inference": true
    }
  }
}
```

PR #1520, opened 2025-08-19
Support semantic highlighting with processor

junqiu-lei:
### Description
Support batch semantic highlighting with processor 

### Related Issues
Resolves https://github.com/opensearch-project/neural-search/issues/1516

### Check List


junqiu-lei:
Not yet add integ test since the batch inference only support with remote model, we could run remote model integ test by one of the options:
1. Use local host model in CI, similar like https://github.com/opensearch-project/search-relevance/pull/173
2. Use pre-defined .env from https://github.com/opensearch-project/.github/issues/372 to run against endpoint deployed on AWS Sagemaker.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
```diff
@@ -0,0 +1,390 @@
...
+        } catch (Exception e) {
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "failed to extract highlight query text: %s", e.getMessage()), e);
+        }
+    }
+
+    /**
+     * Extract field text from search hit using ProcessorUtils
+     */
+    private String extractFieldText(SearchHit hit, String fieldName) {
+        return ProcessorUtils.getValueFromSource(hit.getSourceAsMap(), fieldName).map(Object::toString).orElse(null);
```
bzhangam: What if the returned the value is not a string? e.g. a list of strings for nested case

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
```diff
@@ -413,4 +422,93 @@ public static int getMaxTokenCount(
...
+    public static String extractQueryTextFromBuilder(QueryBuilder queryBuilder) {
+        if (queryBuilder == null) {
+            throw new IllegalArgumentException(String.format(Locale.ROOT, "query builder cannot be null"));
+        }
+
+        if (queryBuilder instanceof MatchQueryBuilder) {
+            return ((MatchQueryBuilder) queryBuilder).value().toString();
+        }
+
+        if (queryBuilder instanceof NeuralQueryBuilder) {
```
bzhangam: It's possible that we receive the rewritten request in the response processor. Not sure why we have a [logic](https://github.com/opensearch-project/OpenSearch/blob/696ed65af93db7c3f9ff5a32e05dfa13764c385c/server/src/main/java/org/opensearch/search/pipeline/Pipeline.java#L129) to copy the search request if there is a search request processor. In that case the response processor will receive the original query builder. Otherwise it will receive the rewritten query builder which can be NeuralKnnQueryBuilder.

vibrantvarun: QQ: What if the highlighter is called from inner_hits? I suppose it will break

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
```diff
@@ -0,0 +1,390 @@
...
+                }
+            }
+
+            if (requests.isEmpty()) {
+                responseListener.onResponse(response);
+                return;
+            }
+
+            if (batchInference) {
+                mlClientAccessor.batchInferenceSentenceHighlighting(modelId, requests, ActionListener.wrap(batchResults -> {
```
bzhangam: How to manage the batch size? Should we allow users to configure the batch size? Or we can start with a default batch size?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
```diff
@@ -0,0 +1,390 @@
...
+                        batchResults,
+                        validHits,
+                        semanticHighlightField
+                    );
+                    long totalTime = System.currentTimeMillis() - startTime;
+                    highlightedResponse = ProcessorUtils.updateResponseTookTime(highlightedResponse, totalTime);
+                    responseListener.onResponse(highlightedResponse);
+                }, error -> { handleProcessingError(error, response, responseListener); }));
+            } else {
+                processSequentialRequests(requests, validHits, 0, response, semanticHighlightField, responseListener, startTime);
```
bzhangam: This can be slower than the existing way that we highlight during the fetch phase which can be done currently on different shards. Should we keep doing this in the fetch phase?

Or would it be better to always do batch inference if we know our model can handle it?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
junqiu-lei: the request batch size is dynamic and equal to query hits doc size.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
junqiu-lei: We should keep both batch and single inference semantic highlighting by the processor. The fetch phase concurrent inference by large shards number could have uncertainty heavy load to model server. We can resolve it by introducing configurable thread number in this processor for single inference for controlled thread number request.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
```diff
@@ -0,0 +1,390 @@
...
+            String queryText = extractQueryText(request);
+
+            processSemanticHighlighting(response, queryText, semanticHighlightField, responseListener, startTime);
+        } catch (Exception e) {
+            handleProcessingError(e, response, responseListener);
+        }
+    }
+
+    @Override
+    public SearchResponse processResponse(SearchRequest request, SearchResponse response) {
```
heemin32: QQ. When is this method method called? In other word, how is it determined if `processResponseAsync` is called or `processResponse` is called?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
heemin32: We might need a limit in batch size here?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
heemin32: It may not actually be slower. For example, consider two shards distributed across two nodes. With a size of 10, each node makes 10 sequential model.predict calls. In comparison, the coordinator node would also make 10 sequential model.predict calls in this code. This means the latency could even improve due to reduced load on model server, or in the worst case, remain about the same.

heemin32:
> Not yet add integ test since the batch inference only support with remote model, we could run remote model integ test by one of the options:
> 
> 1. Use local host model in CI, similar like [Add local Ollama setup and testing scripts for native OpenAI API compatibility search-relevance#173](https://github.com/opensearch-project/search-relevance/pull/173)
> 2. Use pre-defined .env from [[GitHub Request] Create github env for running remote integration tests for neural search .github#372](https://github.com/opensearch-project/.github/issues/372) to run against endpoint deployed on AWS Sagemaker.

We could still test single highlighting functionality with local model?

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
```diff
@@ -413,4 +422,93 @@ public static int getMaxTokenCount(
...
+                response.getScrollId(),
+                response.getTotalShards(),
+                response.getSuccessfulShards(),
+                response.getSkippedShards(),
+                newTook,
+                response.getShardFailures(),
+                response.getClusters(),
+                response.pointInTimeId()
+            );
+        } catch (Exception e) {
```
heemin32: Few questions.
1. Why do we need to catch exception here? I don't see any code which might throw an exception?
2. If exception handling is necessary, do we have unit test testing this case?
3. If exception handling is necessary, shouldn't we also log error message?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -516,4 +547,71 @@ private void retryableExecuteAgent(
...
+            }));
+        } catch (Exception e) {
+            listener.onFailure(e);
+        }
+    }
+
+    /**
+     * Parse the ML output for batch highlighting results
+     */
+    private List<List<Map<String, Object>>> parseBatchHighlightingOutput(MLOutput mlOutput) {
```
heemin32, line 939: I see zero error handling in this method. Can we add? 
1. when mlOutput instance of ModelTensorOutput == false
2. when highlightsList.get(0) instanceof List == false
3. when mlModelTensers.isEmpty() == true
4. when highlightsList.isEmpty() == ture
5. when docHighlithgs instanceof List == false

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
```diff
@@ -0,0 +1,390 @@
...
+                String highlightedText = applyPositionHighlights(text, highlights);
+
+                Map<String, HighlightField> highlightFields = Optional.ofNullable(hit.getHighlightFields()).orElse(new HashMap<>());
+
+                HighlightField highlightField = new HighlightField(fieldName, new Text[] { new Text(highlightedText) });
+                highlightFields.put(fieldName, highlightField);
+                hit.highlightFields(highlightFields);
+            }
+        } catch (Exception e) {
+            log.error(String.format(Locale.ROOT, "error applying highlights to hit %s", hit.getId()), e);
```
heemin32: You don't need to use String.format for logging.
```suggestion
            log.error("error applying highlights to hit {}", hit.getId()), e);
```

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
```diff
@@ -0,0 +1,390 @@
...
+     * Apply batch highlight results to search response using valid hits list
+     */
+    private SearchResponse applyBatchHighlightResults(
+        SearchResponse response,
+        List<List<Map<String, Object>>> batchResults,
+        List<SearchHit> validHits,
+        String semanticHighlightField
+    ) {
+        try {
+            for (int i = 0; i < validHits.size() && i < batchResults.size(); i++) {
```
heemin32: Shouldn't we validate that `validHits.size() == batchResults.size()` and throw an exception otherwise?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
```diff
@@ -0,0 +1,390 @@
...
+     * Apply position-based highlights to text using ProcessorUtils for validation
+     */
+    private String applyPositionHighlights(String text, List<Map<String, Object>> highlights) {
+        try {
+            List<Map<String, Object>> validHighlights = new ArrayList<>();
+            for (Map<String, Object> highlight : highlights) {
+                Object startObj = highlight.get(START_KEY);
+                Object endObj = highlight.get(END_KEY);
+
+                if (ProcessorUtils.isNumeric(startObj) && ProcessorUtils.isNumeric(endObj)) {
```
heemin32: Shouldn't we need to add error log if this condition fails?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
```diff
@@ -0,0 +1,390 @@
...
+                return text;
+            }
+
+            StringBuilder result = new StringBuilder(text);
+            for (int i = validHighlights.size() - 1; i >= 0; i--) {
+                Map<String, Object> highlight = validHighlights.get(i);
+                int start = ((Number) highlight.get(START_KEY)).intValue();
+                int end = ((Number) highlight.get(END_KEY)).intValue();
+
+                if (start >= 0 && end <= text.length() && start < end) {
```
heemin32: Shouldn't we need to add error log if this condition fails?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
```diff
@@ -0,0 +1,390 @@
...
+            SearchResponse finalResponse = ProcessorUtils.updateResponseTookTime(response, totalTime);
+            responseListener.onResponse(finalResponse);
+            return;
+        }
+
+        // Process current request
+        mlClientAccessor.inferenceSentenceHighlighting(requests.get(currentIndex), ActionListener.wrap(highlightResults -> {
+            try {
+                applySingleHighlightResults(validHits.get(currentIndex), highlightResults, semanticHighlightField);
+                // Process next request recursively
```
heemin32: Can we avoid using a recursive approach here? For large results, it would consume stack memory unnecessarily.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
```diff
@@ -0,0 +1,390 @@
...
+
+    /**
+     * Extract query text from search request for semantic highlighting
+     */
+    private String extractQueryText(SearchRequest request) {
+        try {
+            return Optional.ofNullable(request.source())
+                .map(SearchSourceBuilder::query)
+                .map(ProcessorUtils::extractQueryTextFromBuilder)
+                .orElse("");
```
heemin32: If it is empty, shouldn't we throw an exception?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
```diff
@@ -0,0 +1,390 @@
...
+            return response;
+        }
+    }
+
+    /**
+     * Apply single highlight results to a hit
+     */
+    private void applySingleHighlightResults(SearchHit hit, List<Map<String, Object>> highlightResults, String fieldName) {
+        try {
+            if (highlightResults != null && !highlightResults.isEmpty()) {
```
heemin32: We might need to throw exception for each else case?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
junqiu-lei: `processResponseAsync()` is always called. `processResponse()` is only called indirectly through the default async implementation when we choose to implement the synchronous version.

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
junqiu-lei: > QQ: What if the highlighter is called from inner_hits? I suppose it will break

We can get the inner object from response, and then process it, will look into it to support.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
junqiu-lei: Sure, we can have a batch size limit for safety.

Maybe we can start setting `DEFAULT_MAX_INFERENCE_BATCH_SIZE` to 100 and automatically split large result sets into multiple batches. This handles typical queries under 100 hits efficiently while preventing large requests from overwhelming the ML model. And making it configurable via settings allows tuning for different model capacities.

junqiu-lei:
> > Not yet add integ test since the batch inference only support with remote model, we could run remote model integ test by one of the options:
> > 
> > 1. Use local host model in CI, similar like [Add local Ollama setup and testing scripts for native OpenAI API compatibility search-relevance#173](https://github.com/opensearch-project/search-relevance/pull/173)
> > 2. Use pre-defined .env from [[GitHub Request] Create github env for running remote integration tests for neural search .github#372](https://github.com/opensearch-project/.github/issues/372) to run against endpoint deployed on AWS Sagemaker.
> 
> We could still test single highlighting functionality with local model?

Yes, we can still test single highlighting functionality with local model, while the other main function are from batch ability.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -516,4 +547,71 @@ private void retryableExecuteAgent(
...
+        final int retryTime,
+        final ActionListener<List<Map<String, Object>>> listener
+    ) {
+        try {
+            MLInputDataset inputDataset = new QuestionAnsweringInputDataSet(inferenceRequest.getQuestion(), inferenceRequest.getContext());
+            MLInput mlInput = new MLInput(FunctionName.QUESTION_ANSWERING, null, inputDataset);
+
+            mlClient.predict(inferenceRequest.getModelId(), mlInput, ActionListener.wrap(mlOutput -> {
+                try {
+                    List<Map<String, Object>> result = processHighlightingOutput((ModelTensorOutput) mlOutput);
```
chishui: - `processHighlightingOutput` basically has the same implementation with `buildMapResultFromResponse` can we just use `buildMapResultFromResponse`?
- We already have `retryableInferenceSentenceHighlighting`, `retryableInferenceSentencesWithSingleVectorResult`, `retryableInferenceSimilarityWithVectorResult`, `retryableInferenceSentencesWithVectorResult`, `retryableInferenceSentencesWithMapResult`, now a  new `retryableBatchInferenceSentenceHighlighting`, there should be chances to refactor and remove duplicate code (just communicating ideas, not suggesting refactor in this PR)

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -516,4 +547,71 @@ private void retryableExecuteAgent(
...
+                    Object highlightsObj = dataMap.get(SemanticHighlightingConstants.HIGHLIGHTS_KEY);
+                    List<?> highlightsList = (List<?>) highlightsObj;
+
+                    // Check if this is a batch response (list of lists)
+                    if (!highlightsList.isEmpty() && highlightsList.get(0) instanceof List) {
+                        // This is a batch response - each element is a list of highlights for one document
+                        for (Object docHighlights : highlightsList) {
+                            if (docHighlights instanceof List) {
+                                @SuppressWarnings("unchecked")
+                                List<Map<String, Object>> highlights = (List<Map<String, Object>>) docHighlights;
```
chishui: Do you need to check if items in docHighlights are type of `Map<String, Object>`?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -516,4 +547,71 @@ private void retryableExecuteAgent(
...
+                                results.add(highlights);
+                            }
+                        }
+                    }
+                }
```
chishui, line 993: nit: maybe use early quit such as `if (mlModelTensors.isEmpty()) { continue; }` to avoid deep nested code block to increase code readability.

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
```diff
@@ -413,4 +422,93 @@ public static int getMaxTokenCount(
...
+            }
+
+            return text.toString();
+        }
+
+        if (queryBuilder instanceof NestedQueryBuilder nestedQuery) {
+            return extractQueryTextFromBuilder(nestedQuery.query());
+        }
+
+        // Throw exception for unsupported query types
```
chishui: Curious, other query builders are not supported? What about compound query builders such as: `DisMaxQueryBuilder`, `BoostingQueryBuilder` etc?

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
```diff
@@ -413,4 +422,93 @@ public static int getMaxTokenCount(
...
+    }
+
+    /**
+     * Update SearchResponse with additional processing time
+     */
+    public static SearchResponse updateResponseTookTime(SearchResponse response, long additionalTimeMs) {
+        try {
+            long originalTook = response.getTook().millis();
+            long newTook = originalTook + additionalTimeMs;
+            return new org.opensearch.action.search.SearchResponse(
```
chishui: nit: no need to use full path

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
chishui: `extractSemanticHighlightField` orElse uses null, but here uses empty string, make them consistent?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
```diff
@@ -0,0 +1,390 @@
...
+            log.error(String.format(Locale.ROOT, "sequential inference failed for request %d", currentIndex), error);
+            handleProcessingError(error, response, responseListener);
+        }));
+    }
+
+    /**
+     * Apply highlights to a specific search hit
+     */
+    private void applyHighlightsToHit(SearchHit hit, List<Map<String, Object>> highlights, String fieldName) {
+        try {
```
chishui: hit could be null?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
```diff
@@ -0,0 +1,390 @@
...
+
+                if (start >= 0 && end <= text.length() && start < end) {
+                    result.insert(end, postTag);
+                    result.insert(start, preTag);
+                }
+            }
+
+            return result.toString();
+
+        } catch (Exception e) {
```
chishui: It seems that try catch is used on many non-exceptional code blocks, this could be inefficient. If we can simply check for null, empty or similar cases, just use checks and remove try catch.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessorFactory.java
```diff
@@ -0,0 +1,83 @@
...
+                ignoreFailure,
+                modelId,
+                mlClientAccessor,
+                batchInference,
+                preTag,
+                postTag
+            );
+
+        } catch (Exception e) {
+            throw new IOException("Failed to create semantic highlighting processor", e);
```
chishui: I don't think you need to catch any exception or throw any exception here. Since you provide default values for all parameters, I think only exceptions could throw are mismatched argument type, and it's already handled by `ConfigurationUtils.readXXX`

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
junqiu-lei: updated

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
junqiu-lei: For now, we focused on the clear, unambiguous cases for simple text queries. The challenge for compound queriesis is how to handle multiple query clauses - should all be combined for highlighting, or should different clauses be weighted differently?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
junqiu-lei: hit is valid in this step after `processSemanticHighlighting`

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
junqiu-lei: > processHighlightingOutput basically has the same implementation with buildMapResultFromResponse can we just use buildMapResultFromResponse?

Resolved.

> We already have retryableInferenceSentenceHighlighting, retryableInferenceSentencesWithSingleVectorResult, retryableInferenceSimilarityWithVectorResult, retryableInferenceSentencesWithVectorResult, retryableInferenceSentencesWithMapResult, now a new retryableBatchInferenceSentenceHighlighting, there should be chances to refactor and remove duplicate code (just communicating ideas, not suggesting refactor in this PR)

I saw you created https://github.com/opensearch-project/neural-search/pull/1522, thanks!

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingResponseProcessor.java
junqiu-lei: Similar case for nested type with inner hit https://github.com/opensearch-project/neural-search/pull/1520#discussion_r2289490623. Will look into it.

.github/workflows/CI.yml
```diff
@@ -119,16 +119,68 @@ jobs:
...
       - uses: actions/checkout@v4
+
       - name: Set up JDK ${{ matrix.java }}
         uses: actions/setup-java@v4
         with:
           java-version: ${{ matrix.java }}
           distribution: temurin
-      - name: Build and Run Tests
+
+      - name: Setup TorchServe with Model
```
owaiskazi19: We should have a separate CI for model deployment and run the integ tests of semantic highlighter and only run integ tests for the same in it. Similar to https://github.com/owaiskazi19/neural-search/pull/3/files

.github/actions/torchserve-setup/action.yml
```diff
@@ -0,0 +1,306 @@
...
+          echo "Using handler from: ${{ inputs.handler-path }}"
+          cp "${{ inputs.handler-path }}" handler.py
+        else
+          echo "ERROR: Handler file not found at ${{ inputs.handler-path }}"
+          exit 1
+        fi
+
+        # Download model files from HuggingFace
+        echo "Downloading model files..."
+        BASE_URL="https://huggingface.co/opensearch-project/opensearch-semantic-highlighter-v1/resolve/main"
```
owaiskazi19: Should be passed as env variable or some dynamic way. That way we can deploy any model

.github/workflows/CI.yml
junqiu-lei: The semantic highlighter model is light to use in the existing CI.yml. But this is a good idea that in the future when we extend the local host larger model, dedicated separated CI can avoid OOM.

Btw, the memory usage for the host highlighting model:
```
  Memory Usage Progression:

  1. Container Start (base): 114.8 MiB (0.72%)
  2. After Model Loading: 919.5 MiB (5.75%) - +805 MiB for model
  3. After Single Inference: 942.5 MiB (5.89%) - +23 MiB
  4. After Batch Inference: 961.1 MiB (6.01%) - +18.6 MiB
```

.github/actions/torchserve-setup/action.yml
junqiu-lei: Sure

.github/workflows/CI.yml
owaiskazi19: We should treat this as a new framework for deploying other remote model and a new CI would be useful here

junqiu-lei:
> Is there a way for developers to run tests with a remote model locally? It would be really inconvenient if they could only validate tests through GitHub CI without the ability to run them on their own machine.

Yes, I can add a script to run the model server at local dev env, and then use it in integ test:
```
export TORCHSERVE_ENDPOINT="http://localhost:8080"

# Run specific test
./gradlew integTest --tests "org.opensearch.neuralsearch.highlight.SemanticHighlightingIT"
```

heemin32:
> > Is there a way for developers to run tests with a remote model locally? It would be really inconvenient if they could only validate tests through GitHub CI without the ability to run them on their own machine.
> 
> Yes, I can add a script to run the model server at local dev env, and then use it in integ test:
> 
> ```
> export TORCHSERVE_ENDPOINT="http://localhost:8080"
> 
> # Run specific test
> ./gradlew integTest --tests "org.opensearch.neuralsearch.highlight.SemanticHighlightingIT"
> ```

I’m not sure how feasible it is with reasonable effort, but ideally users could simply trigger a new Gradle task, for example `./gradlew remoteTest`, to run all tests that require a remote model. This task would handle spinning up and shutting down the local server automatically, so users wouldn’t need to run a separate script to launch it.

.github/workflows/CI.yml
junqiu-lei: Updated, you can follow up DEVELOPER_GUIDE from this PR to onboarding other models.

DEVELOPER_GUIDE.md
```diff
@@ -218,6 +228,120 @@ In case remote cluster is secured it's possible to pass username and password wi
...
+
+    @Before
+    public void setUp() {
+        // Check for TorchServe endpoint availability
+        String endpoint = System.getenv("TORCHSERVE_ENDPOINT");
+        if (endpoint == null) {
+            endpoint = System.getProperty("tests.torchserve.endpoint");
+        }
+
+        if (endpoint != null && RemoteModelTestUtils.isRemoteEndpointAvailable(endpoint)) {
```
bzhangam, line 303: Compared with using RemoteModelTestUtils it may be better to have a `BaseNeuralSearchRemoteModelIT` extending the `BaseNeuralSearchIT `. In that class we can put all the common things for remote model IT.

And I thinking get the endpoint should be a common function.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/SemanticHighlightingIT.java
```diff
@@ -0,0 +1,247 @@
...
+        assertTrue("Semantic highlighting should produce highlights for field: " + field, foundHighlight);
+    }
+
+    @SuppressWarnings("unchecked")
+    private Map<String, Object> searchWithHighlight(String index, QueryBuilder queryBuilder, String field, String modelId)
+        throws Exception {
+        // For BWC tests, we'll just use the standard search with the query
+        // Semantic highlighting isn't directly supported by the base search method
+        // The test will verify highlighting exists in response
+        return search(index, queryBuilder, null, 2);
```
bzhangam: How can this trigger the highlight function?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -2899,4 +2902,44 @@ protected void deleteDocById(String index, String docId) throws IOException {
         );
         assertOK(response);
     }
+
+    protected String createSemanticHighlightingPipeline(String pipelineName, String modelId, String fieldName, boolean batchInference)
```
bzhangam: I think it's better to have a dedicated base class for semantic highlighting to hold these common functions.

DEVELOPER_GUIDE.md
junqiu-lei: RemoteModelTestUtils is not only being used for NeuralSearchIT's embedding model, it could also be used for llm model for agentic search, sparse model for sparse search and etc. in the future. We'd better separate the remote model utils from IT.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/SemanticHighlightingIT.java
junqiu-lei: Umm, let me update, working on the the bwc area

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
```diff
@@ -321,6 +340,16 @@ public List<Setting<?>> getSettings() {
         );
     }
 
+    @Override
+    public Settings additionalSettings() {
+        return Settings.builder()
```
bzhangam: Will this merge the setting or override it? I think we need to do merge.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
```diff
@@ -24,6 +26,11 @@ public final class NeuralSearchSettings {
     public static final int MINIMUM_INDEX_THREAD_QTY = 1;
     public static final int MAXIMUM_INDEX_THREAD_QTY = 1024;
 
+    // Constants for system-generated factories setting
+    public static final String SYSTEM_GENERATED_FACTORIES_KEY = "cluster.search.enabled_system_generated_factories";
```
bzhangam: We can simply reference the setting defined in the core.

src/main/java/org/opensearch/neuralsearch/highlight/processor/SemanticHighlightingProcessor.java
```diff
@@ -0,0 +1,281 @@
...
+            String batchValidationError = enrichedConfig.validateBatchInference();
+            if (batchValidationError != null) {
+                enrichedConfig = enrichedConfig.withValidationError(batchValidationError);
+            }
+
+            // Additional validation with model type
+            enrichedConfig = validator.validate(enrichedConfig, response);
+
+            if (!enrichedConfig.isValid()) {
+                if (ignoreFailure) {
```
bzhangam: This is already handled in core. No need to handle it here.

src/main/java/org/opensearch/neuralsearch/highlight/processor/SemanticHighlightingProcessor.java
```diff
@@ -0,0 +1,281 @@
...
+@Log4j2
+public class SemanticHighlightingProcessor implements SearchResponseProcessor, SystemGeneratedProcessor {
+
+    private final boolean ignoreFailure;
+    private final MLCommonsClientAccessor mlClientAccessor;
+    private final HighlightConfigExtractor configExtractor;
+    private final HighlightValidator validator;
+    private final HighlightContextBuilder contextBuilder;
+    private final String tag;
+    private final String description;
```
bzhangam: It's good to have a dedicated description for this processor.

src/main/java/org/opensearch/neuralsearch/highlight/strategies/BatchHighlighter.java
```diff
@@ -0,0 +1,183 @@
...
+        );
+
+        if (requests.size() <= maxBatchSize) {
+            processSingleBatch(context, responseListener);
+        } else {
+            processMultipleBatches(context, responseListener);
+        }
+    }
+
+    private void processSingleBatch(HighlightContext context, ActionListener<SearchResponse> responseListener) {
```
bzhangam: what's the benefit to have this function for single batch? would it be cleaner we always use processMultipleBatches to handle it?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
junqiu-lei: Created a BaseSemanticHighlightingIT extended from BaseNeuralSearchIT for it's dedicated functions.

Removed "createSemanticHighlightingPipeline" since it's used after we apply default system generated pipeline.

src/main/java/org/opensearch/neuralsearch/highlight/processor/SemanticHighlightingProcessor.java
junqiu-lei: Yes, it uses it's own description:
```
        this.description = SemanticHighlightingConstants.DEFAULT_PROCESSOR_DESCRIPTION;

```

src/main/java/org/opensearch/neuralsearch/highlight/strategies/BatchHighlighter.java
junqiu-lei: It's a trade-off for production performance especially since single batch is likely the most frequent case. Single batch path uses applyBatchResults() which is straightforward while we cover large size request cases pass into processMultipleBatches. The change was come from previous comment from Heemin.

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
junqiu-lei: Good point, since OpenSearch 3.0 we introduced semantic highlighting, if user need explicitly enable this setting manually, it can be a breaking change. We need find a safe way to enable this system default processor.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/SemanticHighlightingIT.java
junqiu-lei: Updated, waiting bwc test CI runner disk issue being resolved(unrelated to this PR) to pass tests

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
```diff
@@ -44,17 +54,43 @@ public boolean canHighlight(MappedFieldType fieldType) {
...
     public HighlightField highlight(FieldHighlightContext fieldContext) {
+        // Extract batch_inference option from field options
+        Map<String, Object> options = fieldContext.field.fieldOptions().options();
+        boolean batchInference = extractBatchInference(options);
+
+        if (batchInference) {
+            // Check if system processor is enabled
+            if (!isSystemProcessorEnabled()) {
+                // When batch mode is requested but system processor is not enabled, throw an exception
+                String errorMessage = "Batch inference for semantic highlighting requires the '"
```
bzhangam: Use string.format. And use the setting to access the key `search.pipeline.enabled_system_generated_factories`.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
```diff
@@ -78,12 +114,31 @@ public HighlightField highlight(FieldHighlightContext fieldContext) {
...
+        if (clusterService == null) {
+            clusterService = NeuralSearchClusterUtil.instance().getClusterService();
+        }
+
+        Settings clusterSettings = clusterService.state().metadata().settings();
+        List<String> enabledFactories = clusterSettings.getAsList(
+            SearchPipelineService.ENABLED_SYSTEM_GENERATED_FACTORIES_SETTING.getKey()
+        );
+
+        boolean isEnabled = enabledFactories != null && enabledFactories.contains(SemanticHighlightingConstants.SYSTEM_FACTORY_TYPE);
```
bzhangam: We also support * to enable all factories.

src/main/java/org/opensearch/neuralsearch/highlight/batch/processor/SemanticHighlightingProcessor.java
```diff
@@ -0,0 +1,300 @@
...
+
+            // Check if basic fields are present
+            if (!config.hasRequiredFields()) {
+                log.debug("Missing required fields for semantic highlighting");
+                responseListener.onResponse(response);
+                return;
+            }
+
+            // This processor only handles batch inference
+            if (!config.isBatchInference()) {
```
bzhangam, line 76: I think in our offline discussion we agreed to use the processor to handle the single inference if the processor is available. But here seems like we still want to use the old way to process it. The old way has a better performance?

junqiu-lei: Right. If we use processor(if enabled) to process both single and batch inference, I found that we actually need additionally call with client to get connector information after getModel call, `model` and `connector` objects are separate objects, which can lead to higher overall latency.

And after we have to keep existing highlighter path for single inference, we can directly always use it for single inference for bwc. 

For performance improvement, I think using processor for dedicated batch is more clear.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
junqiu-lei: good point, updated

src/main/java/org/opensearch/neuralsearch/processor/util/ProcessorUtils.java
junqiu-lei: for inner object, the existing semantic highlighter remained to support. We can later raise another PR to support it for batch.

src/main/java/org/opensearch/neuralsearch/highlight/batch/config/HighlightContextBuilder.java
```diff
@@ -0,0 +1,127 @@
...
+public class HighlightContextBuilder {
+
+    /**
+     * Build a HighlightContext from configuration and response
+     * @param config the validated highlight configuration
+     * @param response the search response
+     * @param startTime the start time of processing
+     * @return built context ready for processing
+     */
+    public HighlightContext build(HighlightConfig config, SearchResponse response, long startTime) {
```
vibrantvarun, line 31: I see a lot of debug statements. Can we remove some?

vibrantvarun (commented):
Overall, looks good. But just a suggestion, In future can we raise smaller PR's? it makes easy to review. As this one is too big to review in one go and validating the logics.

src/main/java/org/opensearch/neuralsearch/highlight/batch/config/HighlightContextBuilder.java
junqiu-lei: reduced some

vibrantvarun (approved):

vibrantvarun:
LGTM

Issue #1444, opened 2025-07-14
[Design] Seismic Memory Management


yuye-aws:
# Problem Statement

The Seismic algorithm (see [**RFC**](https://github.com/opensearch-project/neural-search/issues/1335) and [**Design**](https://github.com/opensearch-project/neural-search/issues/1390) for background knowledge) has yielded promising results on MS Marco V1 and V2 dataset (see [**here**](https://github.com/opensearch-project/neural-search/issues/1335) for preliminary benchmark results). For faster access to sparse vector data, the Seismic index stores the data in memory. The following figure illustrates how the Seismic data is stored. The cache stores two parts of the Seismic data, posting clustering and forward index. For data persistence, both parts are stored in Lucene files.

<img width="1277" height="542" alt="Image" src="https://github.com/user-attachments/assets/b2f658f8-4843-46f8-a458-b2f465df8515" />

In practice, the OpenSearch cluster may not have enough memory to fit in all the data. On one hand, we apply quantization on the sparse vector data. We’ll release the quantization solution later. On the other hand, users should be able to protect their OpenSearch cluster from running out of memory. They can have enough flexibility to configure how the Seismic data is loaded into memory. We’re writing this design document on memory management of the Seismic algorithm.

To integrate the Seismic algorithm in OpenSearch, the memory management has the following challenges:

1. **Memory limitation:** The Seismic algorithm requires substantial memory resources. On MS Marco V2 dataset with **138 million** documents, we conducted tests with 3 instances, with each one containing **46 million** documents. The memory usage reaches **64.4GB** in each instance. This level of memory consumption can exceed available resources in many production environments, especially when dealing with large-scale datasets.
2. **Dynamic workload:** OpenSearch clusters typically handle varying workloads. A static memory allocation approach for Seismic data would be inefficient, necessitating a dynamic solution that can adapt to changing demands. For example, during force merge operations, memory requirements spike temporarily but intensely. Query patterns also vary widely. Users may run concurrent searches across different indices and fields, each requiring different portions of the Seismic data to be memory-resident. Some indices might experience heavy query traffic during business hours while others are accessed primarily during off-peak times, creating shifting memory demand patterns throughout the day.
3. **Maintaining a good balance between performance and availability:** While maintaining high search performance is critical, the system must also ensure stability under memory pressure. The challenge lies in designing a memory management system that can deliver optimal performance when resources are available while gracefully degrading when memory is constrained, all without compromising the cluster's stability or the correctness of search results.

[...]

Issue #1390, opened 2025-06-16
[Design] Approximate Sparse Nearest-Neighbor Search  in OpenSearch


chishui:
# Approximate Sparse Nearest-Neighbor Search  in OpenSearch

> **_NOTE:_**This is a design doc for this [RFC](https://github.com/opensearch-project/neural-search/issues/1335).
 
## Background

Since version 2.11, OpenSearch has supported neural sparse retrieval as a novel semantic search approach. Leveraging inverted index technology, this method achieves efficiency comparable to traditional BM25. However, the rapid advancement of generative AI has dramatically reshaped the requirements for vector databases and semantic search capabilities.

Organizations are now handling vector collections that have grown exponentially in scale, presenting significant challenges for neural sparse queries. At these unprecedented data volumes, query performance can deteriorate substantially, resulting in response times that fail to meet customer expectations. To address this critical performance gap, we must enhance the neural sparse retrieval system to maintain consistent query speeds even as data scales to billions of vectors.

ANN (approximate nearest neighbor) algorithms effectively trade minor query accuracy for substantial performance gains. In version 2.15, OpenSearch introduced a neural sparse ANN algorithm—two-phase query—which boosts query speed by [4.15x to 6.87x](https://opensearch.org/blog/introducing-a-neural-sparse-two-phase-algorithm/). We've continued exploring additional algorithms to improve neural sparse vector query performance. Recently, we identified a promising approach in the paper “[Efficient Inverted Indexes for Approximate Retrieval over Learned Sparse Representations](https://arxiv.org/abs/2404.18812)” which introduces the SEISMIC algorithm. This algorithm significantly outperforms BigANN competition winners. Our proof of concept implementation within OpenSearch verified both SEISMIC's performance effectiveness and its feasibility for integration as an OpenSearch plugin. In this document, we propose a design for integrating the SEISMIC algorithm into OpenSearch.

## Brief Introduction on SEISMIC Algorithm

Below is a concise overview of the SEISMIC algorithm. For comprehensive details, please refer to the original paper.

### Building Index

* **Posting Lists with Clusters**: SEISMIC enhances the standard inverted index structure. It first prunes each posting list to retain only documents with the most significant values (controlled by parameter λ). The remaining documents in each posting list are then grouped into clusters using a simplified k-means variant, with the number of clusters determined by parameter β. For each cluster, a summary vector is generated using the maximum value of each token across all documents in that group. Parameter α further refines these summary vectors by removing less significant tokens.
[...]

Issue #1335, opened 2025-05-20
[RFC] Sparse ANN Algorithm: Seismic


yuye-aws:
# Background

Learned sparse representations have been a key feature in OpenSearch's **[neural-search plugin](https://github.com/opensearch-project/neural-search)** since version 2.11, offering powerful semantic search capabilities while maintaining the efficiency advantages over dense vectors. As these sparse embeddings have gained popularity among users, the growth in index size presents scalability challenges.

Traditional exact retrieval methods for sparse vectors, despite their **99%+** accuracy, suffer from increasing query latency as collections grow, particularly for datasets exceeding one million documents. This degradation in query throughput can significantly impact user experience. In OpenSearch 2.15, we took an important first step toward addressing this challenge with a feature named **[two-phase search processor](https://opensearch.org/blog/introducing-a-neural-sparse-two-phase-algorithm/)**. By dynamically pruning tokens with negligible weights, this approach reduces computational load while preserving search relevance.

With this same goal of efficient and high-quality retrieval, we’re planning to implement the ANN (**A**pproximate **N**earest **N**eighbor) search algorithm named [**Seismic**](https://dl.acm.org/doi/10.1145/3626772.3657769) (**S**pilled Clust**e**ring of **I**nverted Lists with **S**ummaries for **M**aximum **I**nner Produ**c**t Search). The Seismic algorithm is specifically designed to capitalize on the unique statistical properties of neural sparse representations. Given Seismic, OpenSearch users can significantly reduce query latency with little impact on search quality. In the next few sections, we will briefly walk through the Seismic algorithm and present benchmark results.

# Seismic algorithm

The Seismic algorithm demonstrates remarkable efficiency, performing one to two orders of magnitude faster than conventional exact search methods. In their evaluation using the MS MARCO dataset containing **8.8 million** documents, the authors reported impressive performance metrics: achieving a mean query latency of just **187 microseconds** while maintaining **90%** recall compared to exact search methods. This significant reduction in processing time, without substantial compromise to search quality, highlights Seismic's potential for high-performance information retrieval applications.

![Image](https://github.com/user-attachments/assets/651a76c4-bcac-43f3-ad85-fec3adc6eac6)
<p align="center">
  <br>
  <strong>Figure I. The design of Seismic index from <a href="https://dl.acm.org/doi/10.1145/3626772.3657769">paper</a> </strong>
</p>

[...]

PR #1524, opened 2025-08-24
Seismic: cache basic classes

yuye-aws:
### Description
This is the third PR of SEISMIC feature. In this PR, we submit the basic classes related to cache, including:
1. Cached data structure for forward index and clustered posting.
2. Circuit breaker manager to protect the OpenSearch cluster from running out of JVM memory
3. LRU (Least Recently Used) Algorithm to evict cached data when full.

### Recommended order to review this PR:

- **CircuitBreakerManager**: manages circuit breaker for cached data
- **CacheKey**:  locates cached data by segment info and field name
- **SparseCache**, **ForwardIndexCache** and **ClusteredPostingCache**: cache for forward index and clustered posting cache
- **CacheableClusteredPostingWriter** and **CacheableSparseVectorWriter**: two interfaces to support erase method for cached data
- **ClusteredPostingCacheItem** and **ForwardIndexCacheItem**: store the cached data for a specific field within a segment, where the circuitBreakerTriggerHandler triggers the LRU cache eviction
- **CacheGatedForwardIndexReader**, **CacheGatedPostingsReader** and **SparseTermsLuceneReader**: reader to fetch data from cache or disk
- **AbstractLruCache**, **LruCacheKey**, **LruDocumentCache** and **LruTermCache**: LRU algorithm to evict cached data
- Unit tests

- [x] Basic classes
- [x] Cache
- [ ] Codec
- [ ] Query
- [ ] BWCs
- [ ] Interfaces
    - [ ] Indexing and query
    - [ ] Cache: Memory stats
    - [ ] Cache: Circuit breaker
    - [ ] Cache: Warm up and clear cache API

### Related Issues
RFC: #1335 
High level design: #1390 
Memory management design: #1444

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedForwardIndexReader.java
+++ b/src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedForwardIndexReader.java
@@ -63,7 +63,7 @@ public class CacheGatedForwardIndexReader implements SparseVectorReader {
         }
 
         vector = luceneReader.read(docId);
-        
+
         if (vector != null) {
             cacheWriter.insert(docId, vector);
         }
--- a/src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedPostingsReader.java
+++ b/src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedPostingsReader.java
@@ -66,7 +66,7 @@ public class CacheGatedPostingsReader implements ClusteredPostingReader {
         }
 
         clusters = luceneReader.read(fieldName, term);
-        
+
         if (clusters != null) {
             cacheWriter.insert(term, clusters.getClusters());
         }
--- a/src/test/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedForwardIndexReaderTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedForwardIndexReaderTests.java
@@ -12,7 +12,6 @@ import org.opensearch.neuralsearch.sparse.accessor.SparseVectorWriter;
 import org.opensearch.neuralsearch.sparse.data.SparseVector;
 
 import static org.mockito.ArgumentMatchers.anyInt;
-import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.when;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.any;
@@ -68,24 +67,6 @@ public class CacheGatedForwardIndexReaderTests extends AbstractSparseTestBase {
     }
 
     /**
-     * Tests the read method when first cache read cache return null.
-     * This scenario verifies the double check lock for lucene reader
-     */
-    public void test_read_whenVectorNotInCacheFirstTime() throws IOException {
-        when(cacheReader.read(anyInt())).thenReturn(null).thenReturn(testSparseVector);
-        when(luceneReader.read(anyInt())).thenReturn(null);
-
-        CacheGatedForwardIndexReader reader = new CacheGatedForwardIndexReader(cacheReader, cacheWriter, luceneReader);
-
-        SparseVector result = reader.read(testDocId);
-
-        assertNotNull(result);
-        verify(cacheReader, times(2)).read(testDocId);
-        verify(luceneReader, never()).read(anyInt());
-        verify(cacheWriter, times(1)).insert(anyInt(), any(SparseVector.class));
-    }
-
-    /**
      * Tests the read method when both cache and Lucene storage return null.
      * This scenario verifies that the method correctly handles the case where the
      * requested vector does not exist in either storage.
@@ -99,7 +80,7 @@ public class CacheGatedForwardIndexReaderTests extends AbstractSparseTestBase {
         SparseVector result = reader.read(testDocId);
 
         assertNull(result);
-        verify(cacheReader, times(2)).read(testDocId);
+        verify(cacheReader).read(testDocId);
         verify(luceneReader).read(testDocId);
         verify(cacheWriter, never()).insert(anyInt(), any(SparseVector.class));
     }
--- a/src/test/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedPostingsReaderTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedPostingsReaderTests.java
@@ -20,7 +20,6 @@ import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
@@ -108,24 +107,6 @@ public class CacheGatedPostingsReaderTests extends AbstractSparseTestBase {
      * This scenario verifies that the method correctly handles the case where the
      * requested posting clusters do not exist in either storage.
      */
-    public void test_read_whenClustersNotInCacheFirstTime() throws IOException {
-        when(cacheReader.read(any(BytesRef.class))).thenReturn(null).thenReturn(testPostingClusters);
-        when(luceneReader.read(anyString(), any(BytesRef.class))).thenReturn(null);
-
-        CacheGatedPostingsReader reader = new CacheGatedPostingsReader(testFieldName, cacheReader, cacheWriter, luceneReader);
-        PostingClusters result = reader.read(testTerm);
-
-        assertNotNull(result);
-        verify(cacheReader, times(2)).read(testTerm);
-        verify(luceneReader, never()).read(anyString(), any(BytesRef.class));
-        verify(cacheWriter).insert(eq(testTerm), eq(testPostingClusters.getClusters()));
-    }
-
-    /**
-     * Tests the read method when both cache and Lucene storage return null.
-     * This scenario verifies that the method correctly handles the case where the
-     * requested posting clusters do not exist in either storage.
-     */
     public void test_read_whenClustersNotInCacheAndLucene() throws IOException {
         when(cacheReader.read(any(BytesRef.class))).thenReturn(null);
         when(luceneReader.read(anyString(), any(BytesRef.class))).thenReturn(null);
@@ -134,7 +115,7 @@ public class CacheGatedPostingsReaderTests extends AbstractSparseTestBase {
         PostingClusters result = reader.read(testTerm);
 
         assertNull(result);
-        verify(cacheReader, times(2)).read(testTerm);
+        verify(cacheReader).read(testTerm);
         verify(luceneReader).read(testFieldName, testTerm);
         verify(cacheWriter, never()).insert(any(BytesRef.class), any());
     }
@@ -155,7 +136,7 @@ public class CacheGatedPostingsReaderTests extends AbstractSparseTestBase {
         PostingClusters result = reader.read(testTerm);
 
         assertEquals(testPostingClusters, result);
-        verify(cacheReader, times(2)).read(testTerm);
+        verify(cacheReader).read(testTerm);
         verify(luceneReader).read(testFieldName, testTerm);
         verify(cacheWriter).insert(eq(testTerm), eq(testPostingClusters.getClusters()));
     }
@@ -212,7 +193,7 @@ public class CacheGatedPostingsReaderTests extends AbstractSparseTestBase {
         PostingClusters result = reader.read(emptyTerm);
 
         assertEquals(testPostingClusters, result);
-        verify(cacheReader, times(2)).read(emptyTerm);
+        verify(cacheReader).read(emptyTerm);
         verify(luceneReader).read(testFieldName, emptyTerm);
         verify(cacheWriter).insert(eq(emptyTerm), eq(testPostingClusters.getClusters()));
     }
@@ -260,7 +241,7 @@ public class CacheGatedPostingsReaderTests extends AbstractSparseTestBase {
         PostingClusters result = reader.read(specialTerm);
 
         assertEquals(testPostingClusters, result);
-        verify(cacheReader, times(2)).read(specialTerm);
+        verify(cacheReader).read(specialTerm);
         verify(luceneReader).read(testFieldName, specialTerm);
         verify(cacheWriter).insert(eq(specialTerm), eq(testPostingClusters.getClusters()));
     }
@@ -280,7 +261,7 @@ public class CacheGatedPostingsReaderTests extends AbstractSparseTestBase {
         });
 
         assertEquals("Test IO Exception", exception.getMessage());
-        verify(cacheReader, times(2)).read(testTerm);
+        verify(cacheReader).read(testTerm);
         verify(luceneReader).read(testFieldName, testTerm);
         verify(cacheWriter, never()).insert(any(BytesRef.class), any());
     }


src/main/java/org/opensearch/neuralsearch/sparse/codec/SparsePostingsConsumer.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.lucene.index.Fields;
+import org.apache.lucene.codecs.NormsProducer;
+import lombok.extern.log4j.Log4j2;
+import org.apache.lucene.codecs.FieldsConsumer;
+
+/**
+ * This class is responsible for writing sparse postings to the index
+ */
+@Log4j2
+public class SparsePostingsConsumer extends FieldsConsumer {
```
yuye-aws, line 17: This class will be implemented in the codec PR. I am including this class in this PR because **SparseTermsLuceneReader** depends on it.

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseTermsLuceneReader.java
```diff
@@ -0,0 +1,182 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * This class read terms and clustered posting from lucene index.
+ * It stores the posting to the cache data structure.
+ */
+@Log4j2
+public class SparseTermsLuceneReader extends FieldsProducer {
```
yuye-aws, line 34: This class is needed by **CacheGatedPostingsReader**.

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparsePostingsConsumer.java
```diff
@@ -0,0 +1,23 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.sparse.codec;
+
+import lombok.extern.log4j.Log4j2;
+
+/**
+ * This class is responsible for writing sparse postings to the index
+ */
+public class SparsePostingsConsumer {
```
yuye-aws, line 10: This class will be implemented in the codec PR. I am including this class in this PR because **SparseTermsLuceneReader** depends on it.

yuye-aws:
Hi maintainers, this PR is ready for review now!

build.gradle
```diff
@@ -321,6 +324,7 @@ def _numNodes = findProperty('numNodes') as Integer ?: 1
 test {
     include '**/*Tests.class'
     systemProperty 'tests.security.manager', 'false'
+    systemProperty "jdk.attach.allowAttachSelf", true
```
yuye-aws, line 327: These dependencies are included to support mocking static function configured in the following file. 
```
src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```

The allowAttachSelf has already been used by the k-NN plugin: https://github.com/opensearch-project/k-NN/blob/183684243ded8fa120ae0d88761f4dc4d31b1e9e/build.gradle#L445

build.gradle
```diff
@@ -321,6 +321,10 @@ def _numNodes = findProperty('numNodes') as Integer ?: 1
 test {
     include '**/*Tests.class'
     systemProperty 'tests.security.manager', 'false'
+    filter {
+        // TODO: include sparse tests
+        excludeTestsMatching "org.opensearch.neuralsearch.sparse.codec.SparseTermsLuceneReaderTests.*"
+    }
```
yuye-aws, line 326: This test class is temporarily excluded since the neural-search repo does not support mock static. It calls for code change from this PR: https://github.com/chishui/neural-search/pull/99/files. Unfortunately, it will alter the mock behavior on other existing classes, resulting in the failure of other UTs. This issue will be fixed by the Codec PR.

src/main/java/org/opensearch/neuralsearch/sparse/cache/AbstractLruCache.java
```diff
@@ -0,0 +1,124 @@
...
+ * based on least recently used policy.
+ *
+ * @param <Key> The type of key used for cache entries
+ */
+@Log4j2
+public abstract class AbstractLruCache<Key extends LruCacheKey> {
+
+    /**
+     * Map to track access with LRU ordering
+     * We use Map instead of set because only linked hash map supports tracking the access order ot tiems
```
zhichao-aws: > the access order ot tiems

typo?

src/main/java/org/opensearch/neuralsearch/sparse/cache/CircuitBreakerManager.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Manages circuit breaker functionality for neural search operations.
+ * This class provides a centralized way to track and manage memory usage for neural search
+ * operations. It helps prevent out-of-memory situations by tracking memory allocations and
+ * notify when memory usage exceeds configured thresholds.
+ * The circuit breaker will be set upon plugin initialization.
+ */
+@Log4j2
+public class CircuitBreakerManager {
```
zhichao-aws, line 23: I don't quite understand why we need this class

1. for all method, it just delegate to `CircuitBreaker`, except the error handling of `addMemoryUsage`. If it's just to neat the calling of `addMemoryUsage`, can we make it a util method instead of class?
2. The CircuitBreakerManager class is binded with only one `CircuitBreaker` instance via static field. Is it possible that we'll add more circuit breakers? (e.g. we'll have more sparse ANN algos, or more fine-grained CB control) For currect design it's not easy for extension

src/main/java/org/opensearch/neuralsearch/sparse/cache/CircuitBreakerManager.java
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * Adds memory usage for neural search operations
+     *
+     * @param bytes The number of bytes to add to the circuit breaker
+     * @param label A label to identify the operation in case of circuit breaking
+     * @return false when the limit would be exceeded
+     */
+    public static boolean addMemoryUsage(long bytes, String label) {
+        try {
+            circuitBreaker.addEstimateBytesAndMaybeBreak(bytes, label);
```
zhichao-aws, line 40: should we add NPE check? same below

zhichao-aws: And for this method, should we check bytes is non negative?

src/main/java/org/opensearch/neuralsearch/sparse/cache/SparseCache.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Remove a specific index from cache.
+     * This method delegates to the CacheSparseVectorForwardIndex implementation
+     * to clean up resources associated with the specified cache key.
+     * This function is marked as synchronized for thread safety.
+     *
+     * @param key The CacheKey that identifies the index to be removed
+     */
+    public synchronized void removeIndex(@NonNull CacheKey key) {
```
zhichao-aws: Can we use `computeIfPresent` instead of `synchronized` for better performance?

src/main/java/org/opensearch/neuralsearch/sparse/cache/ForwardIndexCache.java
```diff
@@ -0,0 +1,35 @@
...
+    private ForwardIndexCache() {
+        CircuitBreakerManager.addWithoutBreaking(RamUsageEstimator.shallowSizeOf(cacheMap));
+    }
+
+    public static ForwardIndexCache getInstance() {
+        return INSTANCE;
+    }
+
+    @NonNull
+    public ForwardIndexCacheItem getOrCreate(@NonNull CacheKey key, int docCount) {
```
zhichao-aws, line 28: Why `getOrCreate` not a method of  `SparseCache`?

src/main/java/org/opensearch/neuralsearch/sparse/cache/ForwardIndexCache.java
```diff
@@ -0,0 +1,35 @@
...
+
+import lombok.NonNull;
+import org.apache.lucene.util.RamUsageEstimator;
+
+/**
+ * Registry class to manage forward index within cache.
+ * Given the limited cache introduced by CacheKey, we will simply use the addWithoutBreaking
+ * method to ensure success insertion in registry.
+ */
+public class ForwardIndexCache extends SparseCache<ForwardIndexCacheItem> {
```
zhichao-aws, line 15: Assume we don't use this cache anymore after we write some data into it. Will the memory never get released without restart JVM?

src/main/java/org/opensearch/neuralsearch/sparse/cache/ForwardIndexCacheItem.java
```diff
@@ -0,0 +1,153 @@
...
+            long ramBytesUsed = vector.ramBytesUsed();
+
+            if (!CircuitBreakerManager.addMemoryUsage(ramBytesUsed, CIRCUIT_BREAKER_LABEL)) {
+                if (circuitBreakerTriggerHandler != null) {
+                    circuitBreakerTriggerHandler.accept(ramBytesUsed);
+                }
+
+                // Try again after eviction
+                if (!CircuitBreakerManager.addMemoryUsage(ramBytesUsed, CIRCUIT_BREAKER_LABEL)) {
+                    log.warn("Failed to add to cache even after eviction, vector will not be cached");
```
zhichao-aws, line 106: If the cache is disabled by setting the threshold to 0 or a small number, will we generate too much warning log? I assume we'll try read/write cache for every doc traversed by query

src/main/java/org/opensearch/neuralsearch/sparse/cache/ForwardIndexCacheItem.java
```diff
@@ -0,0 +1,153 @@
...
+                }
+            }
+
+            // Record access to update LRU status
+            LruDocumentCache.DocumentKey documentKey = new LruDocumentCache.DocumentKey(cacheKey, docId);
+            LruDocumentCache.getInstance().updateAccess(documentKey);
+
+            // Only update memory usage if we actually inserted a new document
+            if (sparseVectors.compareAndSet(docId, null, vector)) {
+                usedRamBytes.addAndGet(ramBytesUsed);
```
zhichao-aws, line 117: If the `compareAndSet` fail, should we release the memory in CircuitBreakerManager? Otherwise there are inconsistency

src/main/java/org/opensearch/neuralsearch/sparse/cache/ForwardIndexCacheItem.java
```diff
@@ -0,0 +1,153 @@
...
+
+        /**
+         * Removes a sparse vector from the cached forward index.
+         *
+         * @param docId The document ID of the sparse vector to be removed from the forward index
+         * @return The number of RAM bytes freed by removing this sparse vector
+         */
+        @Override
+        public long erase(int docId) {
+            if (docId >= sparseVectors.length()) {
```
zhichao-aws: docId<0 || docId >= sparseVectors.length()

src/main/java/org/opensearch/neuralsearch/sparse/cache/ClusteredPostingCacheItem.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Consumer;
+
+/**
+ * This class manages the cache postings for sparse vectors. It provides methods to write and read postings from cache.
+ * It is used by the SparsePostingsConsumer and SparsePostingsReader classes.
+ */
+@Log4j2
+public class ClusteredPostingCacheItem implements ClusteredPosting, Accountable {
```
zhichao-aws, line 29: similar comments as ForwardIndexCacheItem

src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedForwardIndexReader.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * A cache-gated forward index reader that implements a two-tier read strategy for sparse vectors.
+ * Cache misses are automatically populated from the underlying storage.
+ */
+public class CacheGatedForwardIndexReader implements SparseVectorReader {
+
+    /**
+     * A no-op implementation of SparseVectorReader that always returns null.
+     */
+    private static final SparseVectorReader emptySparseVectorReader = docId -> null;
```
zhichao-aws: suggest renaming it using styles for constants. E.g. NOOP_READER, NOOP_WRITER

src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedForwardIndexReader.java
```diff
@@ -0,0 +1,70 @@
...
+     * Read Strategy:
+     * 1. First attempts to read from the cache
+     * 2. On cache miss, reads from Lucene storage
+     * 3. Automatically populates the cache with the retrieved vector
+     *
+     * @param docId the document ID for which to retrieve the sparse vector
+     * @return the sparse vector associated with the document ID, or null if the vector does not exist
+     * @throws IOException if an I/O error occurs while reading
+     */
+    public SparseVector read(int docId) throws IOException {
```
zhichao-aws, line 59: nit: add Override annotation

src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedPostingsReader.java
```diff
@@ -0,0 +1,88 @@
...
+     * @throws IOException if an I/O error occurs while reading
+     */
+    @Override
+    public PostingClusters read(BytesRef term) throws IOException {
+        PostingClusters clusters = cacheReader.read(term);
+        if (clusters != null) {
+            return clusters;
+        }
+        // if cluster does not exist in cache, read from lucene and populate it to cache
+        synchronized (luceneReader) {
```
zhichao-aws: why we need synchronized here but don't need for CacheGatedForwardIndexReader? If the read method is not thread-safe, should we add synchronized on `read` method? At least we should note this in method comments

src/main/java/org/opensearch/neuralsearch/sparse/cache/AbstractLruCache.java
yuye-aws: Ack. Will fix it.

src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedPostingsReader.java
zhichao-aws: And we can double check by reading cache again here to avoid duplicate lucene read. (n thread read same term at the same time, lucene read times n -> 1)

src/main/java/org/opensearch/neuralsearch/sparse/cache/CircuitBreakerManager.java
yuye-aws: > should we add NPE check? same below

No need. Because circuit breaker will be set upon plugin initialization

yuye-aws: > And for this method, should we check bytes is non negative?

There is certainly no harm to check, but no gain. Additionally, circuitBreaker.addEstimateBytesAndMaybeBreak accepts a negative value. If we turn to the singleton option, then the negative check will be saved.

src/main/java/org/opensearch/neuralsearch/sparse/cache/ForwardIndexCache.java
yuye-aws: Due to parameter difference, for ForwardIndex, we need to further specify the docCount

src/main/java/org/opensearch/neuralsearch/sparse/cache/ForwardIndexCache.java
yuye-aws: The memory will be released when user calls the clear cache API.

src/main/java/org/opensearch/neuralsearch/sparse/cache/ForwardIndexCacheItem.java
yuye-aws: You're right. Excessive log should definitely be avoided. @chishui is currently working on it. The solution will be included in the future PRs

src/main/java/org/opensearch/neuralsearch/sparse/cache/ForwardIndexCacheItem.java
yuye-aws: The `compareAndSet` fail when means the **sparseVectors[docId]** is not null, meaning that vector has been inserted. In this case, we'll simply skipping inserting the new vector. The memory stats should not be updated.

src/main/java/org/opensearch/neuralsearch/sparse/cache/ForwardIndexCacheItem.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedForwardIndexReader.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedForwardIndexReader.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedPostingsReader.java
yuye-aws: Confirming with @chishui

chishui: it's not thread-safe for luceneReader as it has a file pos which tracks the location of reading items. For cache object read, it's already thread-safe. You are right about CacheGatedForwardIndexReader, it should has the same logic.

yuye-aws: Great catch @zhichao-aws , for forward index, we can have multi thread reading the same doc ID. It's definitely safer to declare **synchronized**. In fact, we reply on the advanceExact method in **BinaryDocValues** class. This read may alter the file pointer. I have updated both files with synchronized keyword and comments

src/main/java/org/opensearch/neuralsearch/sparse/cache/CircuitBreakerManager.java
yuye-aws: > I don't quite understand why we need this class

You concern definitely makes sense. Let me share more context with you. This class is first introduced to store and access the CircuitBreaker instance. 
The neural search implements the [circuit breaker plugin](https://github.com/opensearch-project/OpenSearch/blob/669ec8dbe6e6a0afdd2a1c6f44e4d5f2c26dfba8/server/src/main/java/org/opensearch/plugins/CircuitBreakerPlugin.java#L45) from OpenSearch core. In the NeuralSearch class, upon plugin initialization, we fetch an instance of CircuitBreaker and then set to the Manager class.

In addition to making it a util class, we can make it a singleton class. Does it make sense to you? 

```
@Getter
@Log4j2
public class CircuitBreakerManager {

    private static CircuitBreakerManager instance;

    private CircuitBreaker circuitBreaker;

    public static synchronized CircuitBreakerManager getInstance() {
        if (instance == null) {
            instance = new CircuitBreakerManager();
        }
        return instance;
    }

    public synchronized void initialize(@NonNull CircuitBreaker circuitBreaker) {
        this.circuitBreaker = circuitBreaker;
    }
}
```

yuye-aws: > If it's just to neat the calling of `addMemoryUsage`, can we make it a util method instead of class?

Since this class has a member variable named circuitBreaker, we cannot make it util

yuye-aws: > The CircuitBreakerManager class is binded with only one `CircuitBreaker` instance via static field. Is it possible that we'll add more circuit breakers? (e.g. we'll have more sparse ANN algos, or more fine-grained CB control) For currect design it's not easy for extension

Unfortunately,  we cannot add more circuit breaker due to the limitation of the circuit breaker plugin. However, there is actually no need for a 2nd sparse circuit breaker. Circuit breaker is introduced to protect OpenSearch cluster from running out of heap memory. If we have more sparse ANN algos, they can simply register their memory usage in the current circuit breaker.

If you wish a fine-grained CB control, it's not the job of circuit breaker. Circuit breaker is for memory protection. Once the CB is triggered, we just prevent the OpenSearch user from inserting data into the cache. Additionally, new sparse algos can introduce their own neural stats. The neural stats will be included in the future PR.

src/main/java/org/opensearch/neuralsearch/sparse/cache/SparseCache.java
yuye-aws: For **computeIfPresent**, we still need to wrap the whole function for atomic operation. I don't see any gain in performance.

```
cacheMap.computeIfPresent(key, (k, value) -> {
    long ramBytesUsed = value.ramBytesUsed() + RamUsageEstimator.shallowSizeOf(k);
    CircuitBreakerManager.getInstance().getCircuitBreaker().addWithoutBreaking(-ramBytesUsed);
    LruTermCache.getInstance().removeIndex(k);
    LruDocumentCache.getInstance().removeIndex(k);
    return null; // Returning null removes the entry
});
```

src/main/java/org/opensearch/neuralsearch/sparse/cache/ClusteredPostingCacheItem.java
yuye-aws: Replied above

yuye-aws:
Updated the PR and replied all @zhichao-aws 's comments. Can you kindly review again?

src/main/java/org/opensearch/neuralsearch/sparse/cache/CircuitBreakerManager.java
zhichao-aws: Okay, if we can have only one circuit breaker in plugin then it makes sense.

I'm fine using current design, but not recomment to write like below. There will be performance issue if multiple threads call get instance class
```
public static synchronized CircuitBreakerManager getInstance() {
        if (instance == null) {
            instance = new CircuitBreakerManager();
        }
        return instance;
    }
```

src/main/java/org/opensearch/neuralsearch/sparse/cache/SparseCache.java
zhichao-aws: Current design add lock to the whole map. It has no difference to use a normal HashMap with lock.

For ConcurrentHashMap, it has more fine-grained bin-level locking and has better performance.

https://stackoverflow.com/questions/31577096/concurrenthashmap-which-concurrent-features-improved-in-jdk8

zhichao-aws: https://javaconceptoftheday.com/how-concurrenthashmap-works-internally-after-java-8/

src/main/java/org/opensearch/neuralsearch/sparse/cache/ForwardIndexCache.java
zhichao-aws: For a Cache abstract class, it should have write method besides read and remove. Otherwise it's incomplete.

There are solutions for parameter difference problem. For example, pass the creator lambda to abstract class
```
// SparseCache.java
protected @NonNull T getOrCreate(@NonNull CacheKey key, @NonNull Function<CacheKey, T> creator) {
    return cacheMap.computeIfAbsent(key, k -> {
        T value = creator.apply(k);
        CircuitBreakerManager.addWithoutBreaking(RamUsageEstimator.shallowSizeOf(k));
        return value;
    });
}

// ForwardIndexCache.java
@NonNull
public ForwardIndexCacheItem getOrCreate(@NonNull CacheKey key, int docCount) {
    return super.getOrCreate(key, k -> new ForwardIndexCacheItem(k, docCount));
}

// ClusteredPostingCache.java
@NonNull
public ClusteredPostingCacheItem getOrCreate(@NonNull CacheKey key) {
    return super.getOrCreate(key, k -> new ClusteredPostingCacheItem(k));
}
```

src/main/java/org/opensearch/neuralsearch/sparse/cache/ForwardIndexCache.java
zhichao-aws: I want to make sure these behaviors are aligned with some existing components, e.g. k-NN plugin? I'm not sure this is the correct experience

src/main/java/org/opensearch/neuralsearch/sparse/cache/ForwardIndexCacheItem.java
zhichao-aws: Good. We can add a todo here. There are already some logger have rate-limits. E.g. DeprecationLogger

yuye-aws: Ack. I will add a TODO. We'll definitely finish it before the Circuit Breaker API available

src/main/java/org/opensearch/neuralsearch/sparse/cache/ForwardIndexCache.java
yuye-aws: Yeah it is aligned. You can refer to the k-NN plugin settings: https://docs.opensearch.org/latest/vector-search/settings/

src/main/java/org/opensearch/neuralsearch/sparse/cache/ForwardIndexCache.java
yuye-aws: Ack.

src/main/java/org/opensearch/neuralsearch/sparse/cache/SparseCache.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/sparse/cache/ForwardIndexCacheItem.java
zhichao-aws: we need to release the memory registered in circuit breaker

src/main/java/org/opensearch/neuralsearch/sparse/cache/CircuitBreakerManager.java
yuye-aws: How about discard the lazy initialization?

```
@Getter
@Log4j2
public class CircuitBreakerManager {

    private static CircuitBreakerManager instance = new CircuitBreakerManager();

    private CircuitBreaker circuitBreaker;

    public static CircuitBreakerManager getInstance() {
        return instance;
    }

    public synchronized void initialize(@NonNull CircuitBreaker circuitBreaker) {
        this.circuitBreaker = circuitBreaker;
    }
}
```

zhichao-aws: I'm OK for this

src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedPostingsReader.java
zhichao-aws: If we do need synchronize here, then I'd recommend double-check locking for better performance

```
synchronized (luceneReader) {
    PostingClusters cached = cacheReader.read(term);
    if (cached != null) {
        clusters = cached;
    } else {
        clusters = luceneReader.read(fieldName, term);
    }
}
 ```

File: src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseTermsLuceneReader.java [hunk 19]
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    @Override
+    public void close() throws IOException {
+        IOUtils.close(this.termsIn, this.postingIn);
+    }
+
+    @Override
+    public void checkIntegrity() throws IOException {
+
```
zhichao-aws: Why is this method blank? Can we add some comments?

src/main/java/org/opensearch/neuralsearch/sparse/cache/ForwardIndexCacheItem.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseTermsLuceneReader.java
yuye-aws: Thanks. Should add some codec util stuff.

src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedPostingsReader.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/sparse/cache/ClusteredPostingCacheItem.java
yuye-aws: Also updated code just like above

src/main/java/org/opensearch/neuralsearch/sparse/cache/AbstractLruCache.java
```diff
@@ -0,0 +1,124 @@
...
+     */
+    public void evict(long ramBytesToRelease) {
+        if (ramBytesToRelease <= 0) {
+            return;
+        }
+
+        long ramBytesReleased = 0;
+
+        // Continue evicting until we've freed enough memory or the cache is empty
+        while (ramBytesReleased < ramBytesToRelease) {
```
bzhangam: Between get and evict there can be another thread mutate the map. The Collections.synchronizedMap only help synchronize the individual operation. For compound operations here we need manually make them sinchronized.

bzhangam: What if there is an issue that we cannot release enough bytes we expect to release? I don't think it's a good idea to be stuck in a infinite loop here. We should catch that case and handle it properly.

src/main/java/org/opensearch/neuralsearch/sparse/cache/CircuitBreakerManager.java
```diff
@@ -0,0 +1,81 @@
...
+        } catch (CircuitBreakingException circuitBreakingException) {
+            log.debug(
+                String.format(
+                    Locale.ROOT,
+                    "Cannot insert cluster into cache due to circuit breaker exception: %s",
```
bzhangam: Since this CircuitBreaker is designed for the whole neural search plugin I think we cannot assume the operation to add the memory is always inserting cluster.

yuye-aws: Make sense. And the use case is always inserting data. Changed the error log

src/main/java/org/opensearch/neuralsearch/sparse/cache/AbstractLruCache.java
yuye-aws: Let's take a look the the whole code. If the cache is empty, **leastRecentlyUsedKey** will be null and break the while loop.

```
// Continue evicting until we've freed enough memory or the cache is empty
while (ramBytesReleased < ramBytesToRelease) {
    // Get the least recently used item
    Key leastRecentlyUsedKey = getLeastRecentlyUsedItem();

    if (leastRecentlyUsedKey == null) {
        // Cache is empty, nothing more to evict
        break;
    }

    // Evict the item and track bytes freed
    ramBytesReleased += evictItem(leastRecentlyUsedKey);
}
```

yuye-aws: Good point. Multi-threading is definitely an issue. As you can see from the evict method comment, I mentioned that **For thread safety, please use synchronized when calling this method**. Below is an example when calling the evict function.

```
this.circuitBreakerTriggerHandler = (ramBytesUsed) -> {
    synchronized (LruTermCache.getInstance()) {
        LruTermCache.getInstance().evict(ramBytesUsed);
    }
};
```

src/main/java/org/opensearch/neuralsearch/sparse/cache/AbstractLruCache.java
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Abstract LRU cache implementation for sparse vector caches.
+ * This class provides common functionality for managing eviction of cache entries
+ * based on least recently used policy.
+ *
+ * @param <Key> The type of key used for cache entries
+ */
+@Log4j2
+public abstract class AbstractLruCache<Key extends LruCacheKey> {
```
zane-neo, line 23: Did we investigated on guava cache, can we reuse guava cache instead of reinventing the wheels?

src/main/java/org/opensearch/neuralsearch/sparse/cache/CircuitBreakerManager.java
zane-neo: Is this method possibly accessed by multiple threads? Do we need synchronization?

src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedForwardIndexReader.java
```diff
@@ -0,0 +1,78 @@
...
+        synchronized (luceneReader) {
+            SparseVector cached = cacheReader.read(docId);
+            if (cached != null) {
+                vector = cached;
+            } else {
+                vector = luceneReader.read(docId);
+            }
+        }
+        if (vector != null) {
+            cacheWriter.insert(docId, vector);
```
zane-neo, line 68: Do we need synchronization for this `insert` here?  I see the insert is checking circuit breaker and multiple threading access can break the checking.

src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedForwardIndexReader.java
```diff
@@ -0,0 +1,78 @@
...
+     * @throws IOException if an I/O error occurs while reading
+     */
+    @Override
+    public SparseVector read(int docId) throws IOException {
+        SparseVector vector = cacheReader.read(docId);
+        if (vector != null) {
+            return vector;
+        }
+        // synchronize luceneReader for thread safety, use double-check lock for better performance
+        synchronized (luceneReader) {
```
zane-neo: Why do we need synchronization for read?

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseTermsLuceneReader.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * This class read terms and clustered posting from lucene index.
+ * It stores the posting to the cache data structure.
+ */
+@Log4j2
+public class SparseTermsLuceneReader extends FieldsProducer {
+    private final Map<String, Map<BytesRef, Long>> fieldToTerms = new HashMap<>();
```
zane-neo, line 35: Is it possible to use ConcurrentHashMap here?

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseTermsLuceneReader.java
```diff
@@ -0,0 +1,183 @@
...
+        Map<BytesRef, Long> termsMapping = fieldToTerms.get(field);
+        if (termsMapping == null) {
+            return null;
+        }
+        if (!termsMapping.containsKey(term)) {
+            return null;
+        }
+        long offset = termsMapping.get(term);
+        List<DocumentCluster> clusters = readClusters(offset);
+        if (clusters == null || clusters.isEmpty()) {
```
zane-neo:
```suggestion
        if (clusters.isEmpty()) {
```

src/main/java/org/opensearch/neuralsearch/sparse/cache/AbstractLruCache.java
zane-neo: You should ensure your method thread safe if it's a multi-threading case, you can't depends on the user to read your code or comments before they use.

src/main/java/org/opensearch/neuralsearch/sparse/cache/AbstractLruCache.java
yuye-aws: Yes. I have checked the Guava cache. It supports LRU and I can specify custom weights based on the ram bytes used function. This is a naive example of Guava cache:

```
private Cache<K, V> buildCache(long memoryBytes) {
    return CacheBuilder.newBuilder()
        .maximumWeight(memoryBytes)
        .weigher(new Weigher<K, V>() {
            @Override
            public int weigh(K key, V value) {
                long ramBytesUsed = value.ramBytesUsed();
                return (int) Math.min(ramBytesUsed, Integer.MAX_VALUE);
            }
        })
        .expireAfterAccess(expirationTimeMinutes, TimeUnit.MINUTES)
        .recordStats()
        .build();
}
```

The key point is: we actually have a two-level cache. For low level, classes like **ClusteredPostingCacheItem** only focuses on a single field within a single segment. From high level, classes like **SparseCache** maintains a global map to cache item instances. To specify the maximum weight, the guava cache has to be implemented globally. Since you're already aware of the thread-safety issue, it's actually sub-optimal to **synchronize** a global map.

Besides, guava cache may temporarily exceed its memory limit due to asynchronous eviction. I don't like the edge cases when circuit breaker triggers quite a lot but the cache is still full.

Additionally, the cache is going to be optimized in the future, which calls for a finer-grained approach to remove items. For example, the current implementation is based on both term level and document level. If we, in the future, managed to optimize the cache by term level. If a term gets removed from cache, then remove all its corresponding documents.

src/main/java/org/opensearch/neuralsearch/sparse/cache/CircuitBreakerManager.java
yuye-aws: Yes. It's accessed by multiple threads. The thread safety is guaranteed by the ChildMemoryCircuiter class with Atomic Long.

```
public class ChildMemoryCircuitBreaker implements CircuitBreaker {
    private final AtomicLong used;
    private final AtomicLong trippedCount;
}
```

src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedForwardIndexReader.java
yuye-aws: I think you're considering the case when multiple threads is writing the same document ID. By our implementation, we use **AtomicReferenceArray** to ensure thread safety.
```
sparseVectors.compareAndSet(docId, null, vector)
```

As for the circuit breaker, it would be safer to reject inserting than to insert more item than the memory limit.

src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedForwardIndexReader.java
yuye-aws: By our implementation, the lucene reader for forward index is actually a binary doc values class. It uses the **advanceExact** method to read sparse vectors from Lucene. Unfortunately, this method may alter the file index pointer. We're synchronizing it to avoid read error.

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseTermsLuceneReader.java
yuye-aws: It's actually a segment level lucene reader. You can consider this map as a read-only view. It reads the terms, if needed, and populate into the map. I don't see crucial thread safety error.

src/main/java/org/opensearch/neuralsearch/sparse/cache/AbstractLruCache.java
yuye-aws: Ack. It would be better to mark the evict method as **synchronized**.

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseTermsLuceneReader.java
zane-neo: You can change the map to ConcurrentHashMap and add the synchronization to `readClusters`, you can have a test first.

src/main/java/org/opensearch/neuralsearch/sparse/cache/AbstractLruCache.java
```diff
@@ -0,0 +1,126 @@
...
+
+        return doEviction(key);
+    }
+
+    /**
+     * Removes all entries for a specific cache key when an index is removed.
+     *
+     * @param cacheKey The cache key to remove
+     */
+    public void removeIndex(@NonNull CacheKey cacheKey) {
```
model-collapse, line 114: 1）if this is related to removing a opensearch index. Definitely we cannot call it remove index because it is a general class. 
2) removing by cachekey is o(N) can we also index the entries by hashkey?

src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedForwardIndexReader.java
model-collapse: This is serious because you are making lucene reading access at least for this field inconcurrent. We need a formal evaluation of the performance.

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseTermsLuceneReader.java
yuye-aws: I find out that it suffices to only add synchronization to `readClusters` function. I'll update the same change to the lucene forward index reader. As for the ConcurrentHashMap, the read function is doing read-only operation. Thread-safety won't be a problem.

yuye-aws: Hi @zhichao-aws , I have also canceled the synchronized (luceneReader) when reading forward index from lucene. Let's have a sync if possbile

bzhangam (approved):

src/main/java/org/opensearch/neuralsearch/sparse/cache/CircuitBreakerManager.java
model-collapse, line 40: This function name is weird, may be check BreakWithEstimatedBytes

src/main/java/org/opensearch/neuralsearch/sparse/cache/SparseCache.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Remove a specific index from cache.
+     * This method delegates to the CacheSparseVectorForwardIndex implementation
+     * to clean up resources associated with the specified cache key.
+     * This function is marked as synchronized for thread safety.
+     *
+     * @param key The CacheKey that identifies the index to be removed
+     */
+    public void removeIndex(@NonNull CacheKey key) {
```
model-collapse, line 32: If you really love this function name. Use onIndexRemoval

src/main/java/org/opensearch/neuralsearch/sparse/cache/SparseCache.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+
+/**
+ * Abstract base class for sparse data caching implementations.
+ * Provides common functionality for caching forward index and clustered postings.
+ * @param <T> The type of Accountable objects stored in the cache
+ */
+public abstract class SparseCache<T extends Accountable> implements Accountable {
```
model-collapse, line 20: This class is just a cache. NOTHING to do with sparse. Use MemMonitoredCache

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseTermsLuceneReader.java
model-collapse: Let's consider the impact to concurrent search. When it's enabled this will be accessed by multiple threads.

src/main/java/org/opensearch/neuralsearch/sparse/cache/LruTermCache.java
```diff
@@ -0,0 +1,44 @@
...
+    private LruTermCache() {
+        super();
+    }
+
+    public static LruTermCache getInstance() {
+        return INSTANCE;
+    }
+
+    @Override
+    protected long doEviction(TermKey termKey) {
```
model-collapse, line 25: Accessing only global single instance in a member function? Is that right?

src/main/java/org/opensearch/neuralsearch/sparse/cache/AbstractLruCache.java
yuye-aws: 1. It's not removing an OpenSearch index. It's actually removing segment. This operation will remove the cache entries with this cacheKey.
2. It can be a optimization. Let's work with a naive cache solution and optimize later

src/main/java/org/opensearch/neuralsearch/sparse/cache/CacheGatedForwardIndexReader.java
yuye-aws: For the forward index, Zirui has tested the impact of **synchronized** on lucene reader. The latency impact for P50, P90 and P99 is minimal

src/main/java/org/opensearch/neuralsearch/sparse/cache/CircuitBreakerManager.java
yuye-aws: The function name is provided by OpenSearch circuit breaker instance.

```
/**
 * add bytes to the breaker and maybe trip
 * @param bytes number of bytes to add
 * @param label string label describing the bytes being added
 * @return the number of "used" bytes for the circuit breaker
 * @throws CircuitBreakingException if the breaker tripped
 */
double addEstimateBytesAndMaybeBreak(long bytes, String label) throws CircuitBreakingException;
```

src/main/java/org/opensearch/neuralsearch/sparse/cache/SparseCache.java
yuye-aws: Ack. Will be updated in the next PR.

src/main/java/org/opensearch/neuralsearch/sparse/cache/SparseCache.java
yuye-aws: Ack. Will be updated in the next PR.

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseTermsLuceneReader.java
yuye-aws: Ack. It will be a benchmark item

src/main/java/org/opensearch/neuralsearch/sparse/cache/LruTermCache.java
yuye-aws: I think you are referring to **ClusteredPostingCache.getInstance()**. We're using it to locate the cache item class. I'm not sure what you are concerned with.

PR #1528, opened 2025-08-27
Enhance: support mocking final functions and classes

yuye-aws:
### Description

This PR provides an unit test enhancement where developers can use mock and final classes. This is achieved by declaring **mock-maker-inline** in the mock maker file. In build.gradle file, I have introduced some common dependencies. In addition to that, some current UT will fail due to the mock inline change. To pass the CI check, I have also included the corresponding changes.

### Related Issues

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -292,7 +292,7 @@ dependencies {
     testImplementation "org.opensearch.plugin:parent-join-client:${opensearch_version}"
     testImplementation 'org.assertj:assertj-core:3.24.2'
     testImplementation group: 'net.bytebuddy', name: 'byte-buddy', version: "${versions.bytebuddy}"
-    testImplementation group: 'org.objenesis', name: 'objenesis', version: '3.3'
+    testImplementation group: 'org.objenesis', name: 'objenesis', version: "${versions.objenesis}"
     testImplementation group: 'net.bytebuddy', name: 'byte-buddy-agent', version: "${versions.bytebuddy}"
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -291,6 +291,9 @@ dependencies {
     testImplementation fileTree(dir: knnJarDirectory, include: ["opensearch-knn-${opensearch_build}.jar", "remote-index-build-client-${opensearch_build}.jar"])
     testImplementation "org.opensearch.plugin:parent-join-client:${opensearch_version}"
     testImplementation 'org.assertj:assertj-core:3.24.2'
+    testImplementation group: 'net.bytebuddy', name: 'byte-buddy', version: "${versions.bytebuddy}"
+    testImplementation group: 'org.objenesis', name: 'objenesis', version: '3.3'
+    testImplementation group: 'net.bytebuddy', name: 'byte-buddy-agent', version: "${versions.bytebuddy}"
 }
 
 // In order to add the jar to the classpath, we need to unzip the
@@ -320,10 +323,7 @@ def _numNodes = findProperty('numNodes') as Integer ?: 1
 test {
     include '**/*Tests.class'
     systemProperty 'tests.security.manager', 'false'
-    filter {
-        // TODO: include sparse tests
-        excludeTestsMatching "org.opensearch.neuralsearch.sparse.codec.SparseTermsLuceneReaderTests.*"
-    }
+    systemProperty "jdk.attach.allowAttachSelf", true
 }
 
 // Setting up Integration Tests
--- a/src/test/java/org/opensearch/neuralsearch/processor/InferenceProcessorTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/processor/InferenceProcessorTests.java
@@ -7,7 +7,11 @@ package org.opensearch.neuralsearch.processor;
 import lombok.Getter;
 import org.junit.Before;
 import org.mockito.ArgumentCaptor;
+import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
+import org.opensearch.cluster.ClusterState;
+import org.opensearch.cluster.metadata.IndexMetadata;
+import org.opensearch.cluster.metadata.Metadata;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.core.action.ActionListener;
@@ -26,7 +30,7 @@ import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.RETURNS_DEEP_STUBS;
+import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
@@ -37,7 +41,14 @@ public class InferenceProcessorTests extends InferenceProcessorTestCase {
     private MLCommonsClientAccessor clientAccessor;
     private Environment environment;
 
-    private ClusterService clusterService = mock(ClusterService.class, RETURNS_DEEP_STUBS);
+    @Mock
+    private ClusterService clusterService;
+    @Mock
+    private ClusterState clusterState;
+    @Mock
+    private Metadata metadata;
+    @Mock
+    private IndexMetadata indexMetadata;
 
     private static final String TAG = "tag";
     private static final String TYPE = "type";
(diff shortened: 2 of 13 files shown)


yuye-aws:
Here lists the UT needs to be updated.

```
Tests with failures:
 - org.opensearch.neuralsearch.processor.InferenceProcessorTests.test_batchExecuteWithNull_allFailedValidation
 - org.opensearch.neuralsearch.processor.InferenceProcessorTests.test_doBatchExecute_exception
 - org.opensearch.neuralsearch.processor.InferenceProcessorTests.test_batchExecute_happyCase
 - org.opensearch.neuralsearch.processor.InferenceProcessorTests.test_batchExecute_sort
 - org.opensearch.neuralsearch.processor.InferenceProcessorTests.test_batchExecute_subBatches
 - org.opensearch.neuralsearch.processor.NeuralSparseTwoPhaseProcessorTests.testProcessRequest_whenTwoPhaseEnabledWithNeuralQuerySparseEmbedding_thenSuccess
 - org.opensearch.neuralsearch.query.HybridQueryWeightTests.testSubQueries_whenMultipleEqualSubQueries_thenSuccessful
 - org.opensearch.neuralsearch.search.query.HybridCollectorManagerTests.testRescoreWithConcurrentSegmentSearch_whenMatchedDocsAndRescore_thenSuccessful
 - org.opensearch.neuralsearch.util.HybridQueryUtilTests.testIsHybridQueryWrappedInBooleanMustQueryWithFilters_NonFilterSecondClause_thenSuccessful
 - org.opensearch.neuralsearch.query.NeuralQueryBuilderRewriteTests.testRewriteTargetSemanticRankFeatures_whenMultipleTargetIndices_thenSuccess
 - org.opensearch.neuralsearch.util.HybridQueryUtilTests.testIsHybridQueryCheck_whenHybridWrappedIntoBoolWithDlsRules_thenSuccess
 - org.opensearch.neuralsearch.util.HybridQueryUtilTests.testIsHybridQueryCheck_whenNoHybridQuery_thenSuccess
 - org.opensearch.neuralsearch.util.HybridQueryUtilTests.testIsHybridQueryWrappedInBooleanMustQueryWithFilters_whenFirstClauseNotMust_thenSuccessful
 - org.opensearch.neuralsearch.util.HybridQueryUtilTests.testIsHybridQueryWrappedInBooleanMustQueryWithFilters_whenSingleHybridMustClause_thenSuccessful
 - org.opensearch.neuralsearch.util.HybridQueryUtilTests.testIsHybridQueryWrappedInBooleanMustQueryWithFilters_whenHybridMustWithValidFilters_thenSuccessful
 - org.opensearch.neuralsearch.util.HybridQueryUtilTests.testIsHybridQueryWrappedInBooleanQuery_whenHybridMustClauseAndFilterClauses_thenSuccessful
 - org.opensearch.neuralsearch.util.HybridQueryUtilTests.testIsHybridQueryCheck_whenQueryIsHybridQueryInstance_thenSuccess
 - org.opensearch.neuralsearch.util.HybridQueryUtilTests.testIsHybridQueryCheck_whenHybridWrappedIntoBoolAndNoNested_thenSuccess
 - org.opensearch.neuralsearch.query.NeuralQueryBuilderRewriteTests.testRewrite_whenTargetKnnAndOnShardDirectly_thenKnnQueryBuilder
 - org.opensearch.neuralsearch.query.NeuralQueryBuilderRewriteTests.testRewriteTargetSemanticRankFeatures_whenMultipleTargetIndicesOneWithAnalyzer_thenSuccess
 - org.opensearch.neuralsearch.query.NeuralQueryBuilderRewriteTests.testRewriteTargetSemanticKnn_whenOnShardDirectly_thenNestedKnnQueryBuilder
 - org.opensearch.neuralsearch.query.NeuralQueryBuilderRewriteTests.testRewriteTargetSemanticKnn_whenMultipleTargetIndices_thenNestedKnnQueryBuilder
 - org.opensearch.neuralsearch.transport.NeuralStatsResponseTests.test_constructor
 - org.opensearch.neuralsearch.stats.NeuralStatsInputTests.test_streamInput
```

yuye-aws:
All UT failures have been addressed. This PR is ready for review now

yuye-aws:
The gradle check error fails with the following error.

```
> Task :forbiddenApisTest FAILED
Forbidden method invocation: java.lang.String#formatted(java.lang.Object[]) [Uses default locale]
  in org.opensearch.neuralsearch.processor.rerank.ByFieldRerankProcessorTests (ByFieldRerankProcessorTests.java:830)
Forbidden method invocation: java.lang.String#formatted(java.lang.Object[]) [Uses default locale]
  in org.opensearch.neuralsearch.processor.rerank.ByFieldRerankProcessorTests (ByFieldRerankProcessorTests.java:871)
Forbidden method invocation: java.lang.String#formatted(java.lang.Object[]) [Uses default locale]
  in org.opensearch.neuralsearch.processor.rerank.ByFieldRerankProcessorTests (ByFieldRerankProcessorTests.java:910)
Scanned 232 class file(s) for forbidden API invocations (in 0.87s), 3 error(s).
```

Updating ByFieldRerankProcessorTests to fix it

chishui (approved):

build.gradle
```diff
@@ -291,6 +291,9 @@ dependencies {
     testImplementation fileTree(dir: knnJarDirectory, include: ["opensearch-knn-${opensearch_build}.jar", "remote-index-build-client-${opensearch_build}.jar"])
     testImplementation "org.opensearch.plugin:parent-join-client:${opensearch_version}"
     testImplementation 'org.assertj:assertj-core:3.24.2'
+    testImplementation group: 'net.bytebuddy', name: 'byte-buddy', version: "${versions.bytebuddy}"
+    testImplementation group: 'org.objenesis', name: 'objenesis', version: '3.3'
```
zhichao-aws: we can use ${versions.objenesis} for consistency

yuye-aws: Ack

chishui:
> Enabling this capability could encourage greater use of static methods

I really don't feel this holds true. Static methods are something you can't avoid. I understand people say static are not good practice, it's just sometimes you have to use it and in that situation, you need to have a way to test it to increase coverage and increase the reliability of your code. You can have a coding standard or PR check list to call out that "whenever you use static, think twice and see if you have non-static alternatives. ". However, enable the capability to test static is a different thing.

yuye-aws:
> Yes, I'm trying to understand how the behavior will be different. I.e. are all mocked objects be different, or part of the mocked objects. And what's the detailed difference it introduces

Let's take a simple example from **NeuralStatsResponseTests**

```
// Before
// Booleans as bytes
when(mockStreamInput.readByte()).thenReturn((byte) 1).thenReturn((byte) 0);

// After
when(mockStreamInput.readBoolean()).thenReturn(true).thenReturn(false);
```

The **readBoolean** method is actually calling **readByte**. With inline mock, the mocked objects function does not call the readByte any more. That's why I need to update the mocking behavior.

zane-neo:
> The mock inline maker is not only designed for static, it also allows us to mock **final** classes which is critical for us especially when we need to mock many lucene classes which is marked final.
> 
> > Enabling this capability could encourage greater use of static methods
> 
> I really don't feel this holds true. Static methods are something you can't avoid. I understand people say static are not good practice, it's just sometimes you have to use it and in that situation, you need to have a way to test it to increase coverage and increase the reliability of your code. You can have a coding standard or PR check list to call out that "whenever you use static, think twice and see if you have non-static alternatives. ". However, enable the capability to test static is a different thing.

Agree but I personally is trying to reduce the static method usage as possible as I can. I wonder do we have any methods we have to make it static instead of making it a non-static method in xxxHelper class?

yuye-aws:
OpenSearch currently uses Junit 4: https://github.com/opensearch-project/OpenSearch/blob/2eb9b53ae6f3e5a97aac2e1e62f6113589024680/gradle/libs.versions.toml#L73

yuye-aws:
> For static methods, I don’t think they should need to be mocked for testing. If mocking becomes necessary, it might be a sign the method should be non-static instead.
> 
> For final classes, an alternative is to create a wrapper class that holds an instance of the final class and use that wrapper in tests.
> 
> That said, I’m also fine if we can avoid mocking static methods.

1. For static methods, if we cannot make a static method from non-neural-search plugin non-static.
2. For final classes, your proposed alternative may work, but there is no harm to allow mocking final classes?
3. Avoid mocking static methods can be considered in another PR, where different options can be listed

src/test/java/org/opensearch/neuralsearch/processor/InferenceProcessorTests.java
```diff
@@ -50,6 +61,10 @@ public class InferenceProcessorTests extends InferenceProcessorTestCase {
     @Before
     public void setup() {
         MockitoAnnotations.openMocks(this);
+        when(clusterService.state()).thenReturn(clusterState);
```
zane-neo, line 64: Why do we need to expand the RETURNS_DEEP_STUBS?

yuye-aws: With mock-inline configured, the expression **clusterService.state().clusterState.metadata().metadata.index(anyString()).indexMetadata.getSettings()** will throw null pointer exception. Therefore, I updated the code to explicitly mock each step.

PR #1360, opened 2025-05-29
Add include_individual_nodes param to stats API

q-andy:
### Description
Adds a new parameter to stats API, `include_individual_nodes`, default `true`. When set to `false`, includes individual node-level stats from the stats response (but still includes aggregated `all_nodes` response)

## With flag disabled (default true)
```
GET /_plugins/_neural/stats?include_individual_nodes=false

{
	"_nodes": {
		"total": 1,
		"successful": 1,
		"failed": 0
	},
	"cluster_name": "integTest",
	"info": {
		"cluster_version": "3.1.0",
		"processors": {
			"ingest": {
				"text_chunking_delimiter_processors": 0,
				"text_chunking_fixed_length_processors": 0,
				"text_embedding_processors_in_pipelines": 0,
				"text_chunking_processors": 0
			}
		}
	},
	"all_nodes": {
		"semantic_highlighting": {
			"semantic_highlighting_request_count": 0
		},
		"processors": {
			"ingest": {
				"text_chunking_executions": 0,
				"text_embedding_executions": 0,
				"text_chunking_fixed_length_executions": 0,
				"text_chunking_delimiter_executions": 0
			}
		}
	}
}
```

## With flag enabled (default true)
```
GET /_plugins/_neural/stats
{
	"_nodes": {
		"total": 1,
		"successful": 1,
		"failed": 0
	},
	"cluster_name": "integTest",
	"info": {
		"cluster_version": "3.1.0",
		"processors": {
			"ingest": {
				"text_chunking_delimiter_processors": 0,
				"text_chunking_fixed_length_processors": 0,
				"text_embedding_processors_in_pipelines": 0,
				"text_chunking_processors": 0
			}
		}
	},
	"all_nodes": {
		"semantic_highlighting": {
			"semantic_highlighting_request_count": 0
		},
		"processors": {
			"ingest": {
				"text_chunking_executions": 0,
				"text_embedding_executions": 0,
				"text_chunking_fixed_length_executions": 0,
				"text_chunking_delimiter_executions": 0
			}
		}
	},
	"nodes": {
		"9XVcfxaHTjeuK1I65oZmPg": {
			"semantic_highlighting": {
				"semantic_highlighting_request_count": 0
			},
			"processors": {
				"ingest": {
					"text_chunking_executions": 0,
					"text_embedding_executions": 0,
					"text_chunking_fixed_length_executions": 0,
					"text_chunking_delimiter_executions": 0
				}
			}
		}
	}
}
```

### Related Issues
[...]

PR #1536, opened 2025-08-29
SEISMIC: Circuit breaker and memory stats

yuye-aws:
### Description

This is the follow-up cache PR of SEISMIC feature. In this PR, we submit the cache related APIs related to cache, including:

1. Four memory stats: sparse_memory_usage_percentage , sparse_memory_usage, clustered_posting_usage, and forward_index_usage 
2. Circuit breaker can be configured through cluster setting APIs.

Here is the example to call the circuit breaker cluster setting API:

```
PUT _cluster/settings
{
  "persistent": {
    "plugins.neural_search.circuit_breaker.limit": "30%",
    "plugins.neural_search.circuit_breaker.overhead": "1.01"
  }
}
```

Example response:

```
GET _cluster/settings

{
  "persistent": {
    "plugins": {
      "neural_search": {
        "circuit_breaker": {
          "limit": "30%",
          "overhead": "1.01"
        }
      }
    }
  },
  "transient": {}
}
```

Here is the example to call the neural stats API:

```
PUT /_cluster/settings
{
  "persistent": {
    "plugins.neural_search.stats_enabled": "true"
  }
}

GET /_plugins/_neural/stats/
```

You can then found memory stats in the node:

```
memory": {
  "sparse": {
    "sparse_memory_usage_percentage": "0.00%",
    "sparse_memory_usage": "128b",
    "clustered_posting_usage": "64b",
    "forward_index_usage": "64b"
  }
}
```

There is also a parameter named `include_metric`:

```
GET /_plugins/_neural/stats/?include_metric=false
```

### Recommended order to review this PR:

- NeuralSearch and NeuralSearchSettings: circuit breaker cluster settings
- metrics folder: all the Seismic memory stats
- cache folder: refactor so that circuit breaker won't produce excessive logs
- Unit tests

- [x] Basic classes
- [x] Cache
- [ ] Codec
- [ ] Query
- [ ] BWCs
- [ ] Interfaces
    - [ ] Indexing and query
    - [x] Cache: Memory stats and circuit breaker
    - [ ] Cache: Warm up and clear cache API

### Related Issues
RFC: #1335 
High level design: #1390 
Memory management design: #1444

### Check List


yuye-aws:
Hi maintainers, this PR is ready for review now!

src/main/java/org/opensearch/neuralsearch/sparse/cache/AbstractLruCache.java
```diff
@@ -111,7 +111,7 @@ protected long evictItem(Key key) {
      *
      * @param cacheKey The cache key to remove
      */
-    public void removeIndex(@NonNull CacheKey cacheKey) {
+    public void onIndexRemoval(@NonNull CacheKey cacheKey) {
```
yuye-aws, line 114: Update this according to @model-collapse 's comment: https://github.com/opensearch-project/neural-search/pull/1524#discussion_r2308887952

src/main/java/org/opensearch/neuralsearch/sparse/cache/MemMonitoredCache.java
```diff
@@ -17,7 +17,7 @@
  * Provides common functionality for caching forward index and clustered postings.
  * @param <T> The type of Accountable objects stored in the cache
  */
-public abstract class SparseCache<T extends Accountable> implements Accountable {
+public abstract class MemMonitoredCache<T extends Accountable> implements Accountable {
```
yuye-aws, line 20: Update this according to @model-collapse 's comment: https://github.com/opensearch-project/neural-search/pull/1524#discussion_r2308890403

src/main/java/org/opensearch/neuralsearch/stats/NeuralStatsInput.java
```diff
@@ -133,16 +151,19 @@ public NeuralStatsInput(StreamInput input) throws IOException {
...
         if (isClusterOnOrAfterMinReqVersionForStatCategoryFiltering()) {
             includeIndividualNodes = input.readBoolean();
             includeAllNodes = input.readBoolean();
             includeInfo = input.readBoolean();
+            includeMetrics = input.readBoolean();
         } else {
             includeIndividualNodes = true;
             includeAllNodes = true;
             includeInfo = true;
+            includeMetrics = true;
```
q-andy: This might cause BWC tests to fail since the StreamInput/Output will be incompatible during rolling upgrade

You may need a separate `if (isClusterOnOrAfterMinReqVersionForMetricStats()) { ... } else { //default values }` block here. Since ff during cluster upgrade, you have a mixed node cluster, if an upgraded node receives a StreamInput from an older version node without MetricStats then it will fail to read the input correctly.

 It may need a helper similar to https://github.com/opensearch-project/neural-search/blob/a30c7b1203e48734e8ce81ddc63649bbfaf8077d/src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java#L59-L61

src/main/java/org/opensearch/neuralsearch/transport/NeuralStatsNodeResponse.java
```diff
@@ -32,18 +35,25 @@ public class NeuralStatsNodeResponse extends BaseNodeResponse implements ToXCont
      */
     public NeuralStatsNodeResponse(StreamInput in) throws IOException {
         super(in);
-        this.stats = in.readMap(input -> input.readEnum(EventStatName.class), TimestampedEventStatSnapshot::new);
+        this.eventStats = in.readMap(input -> input.readEnum(EventStatName.class), TimestampedEventStatSnapshot::new);
+        this.metricStats = in.readMap(input -> input.readEnum(MetricStatName.class), MemoryStatSnapshot::new);
```
q-andy: This might also fail BWC due to same reason as above

src/main/java/org/opensearch/neuralsearch/transport/NeuralStatsResponse.java
```diff
@@ -80,39 +87,49 @@ public NeuralStatsResponse(StreamInput in) throws IOException {
...
     @Override
     public void writeTo(StreamOutput out) throws IOException {
         super.writeTo(out);
         Map<String, Object> downcastedInfoStats = (Map<String, Object>) (Map) (infoStats);
+        Map<String, Object> downcastedMetricStats = (Map<String, Object>) (Map) (metricStats);
         Map<String, Object> downcastedAggregatedNodeStats = (Map<String, Object>) (Map) (aggregatedNodeStats);
         Map<String, Object> downcastedNodeIdToNodeEventStats = (Map<String, Object>) (Map) (nodeIdToNodeEventStats);
 
         out.writeMap(downcastedInfoStats);
+        out.writeMap(downcastedMetricStats);
```
q-andy: This might also fail BWC due to same reason as above

src/main/java/org/opensearch/neuralsearch/stats/metrics/MetricStatsManager.java
```diff
@@ -0,0 +1,54 @@
...
+     */
+    public synchronized static MetricStatsManager instance() {
+        if (INSTANCE == null) {
+            INSTANCE = new MetricStatsManager();
+        }
+        return INSTANCE;
+    }
+
+    /**
+     * Retrieves snapshots of specified event statistics.
```
q-andy: nit: update comment

src/main/java/org/opensearch/neuralsearch/stats/metrics/MetricStatsManager.java
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * @param statsToRetrieve Set of event stat names to retrieve data for
+     * @return Map of event stat names to their current snapshots
+     */
+    public Map<MetricStatName, MemoryStatSnapshot> getStats(EnumSet<MetricStatName> statsToRetrieve) {
```
q-andy, line 45: Are there any other metric stat types planned other than MemoryStats? If so, you might want ot make the Map more generic to include other types of StatSnapshots not just MemoryStatSnapshots

src/main/java/org/opensearch/neuralsearch/transport/NeuralStatsResponse.java
```diff
@@ -49,11 +52,13 @@ public class NeuralStatsResponse extends BaseNodesResponse<NeuralStatsNodeRespon
     public NeuralStatsResponse(StreamInput in) throws IOException {
         super(new ClusterName(in), in.readList(NeuralStatsNodeResponse::readStats), in.readList(FailedNodeException::new));
         Map<String, StatSnapshot<?>> castedInfoStats = (Map<String, StatSnapshot<?>>) (Map) in.readMap();
+        Map<String, StatSnapshot<?>> castedMetricStats = (Map<String, StatSnapshot<?>>) (Map) in.readMap();
         Map<String, StatSnapshot<?>> castedAggregatedNodeStats = (Map<String, StatSnapshot<?>>) (Map) in.readMap();
         Map<String, Map<String, StatSnapshot<?>>> castedNodeIdToNodeEventStats = (Map<String, Map<String, StatSnapshot<?>>>) (Map) in
             .readMap();
 
         this.infoStats = castedInfoStats;
+        this.metricStats = castedMetricStats;
```
q-andy: This might also fail BWC due to same reason as above

q-andy (commented):
Thanks for the metric stat changes, this covers a gap in the original design :)

Regarding the format of the response, why have `"64b"` and `"30.3%"` versus having them as numbers like `64.0` and `30.3`? For example, in knn stats, the stats are like `graph_memory_usage_percent: 32.1`. So if they need to be monitored and graphed, they do not have to convert from string to float to parse. E.g. response: https://docs.opensearch.org/latest/vector-search/api/knn/#example-request 

Also see my comments on BWCs. Monitoring memory usage may be important during cluster upgrade, so it's a good idea to make sure it works. Maybe my changes [here](https://github.com/opensearch-project/neural-search/pull/1360/files#diff-a702c89f622f31cc723db52edddfdb1a1a4da84b7798f4fbd9ab1dc350968b48) might be useful as reference

src/main/java/org/opensearch/neuralsearch/stats/NeuralStatsInput.java
yuye-aws: Thanks. I think that's why the BWC test is currently failing

src/main/java/org/opensearch/neuralsearch/transport/NeuralStatsNodeResponse.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/transport/NeuralStatsResponse.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/stats/metrics/MetricStatsManager.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/stats/metrics/MetricStatsManager.java
yuye-aws: > Are there any other metric stat types planned other than MemoryStats?

Not yet. It's easy to extend. If more stats type in addition to memory stat is coming, I can introduce an abstract class named MetricStatSnapshot

src/main/java/org/opensearch/neuralsearch/transport/NeuralStatsResponse.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/sparse/cache/ClusteredPostingCache.java
```diff
@@ -12,20 +12,30 @@
  * Given the limited cache introduced by CacheKey, we will simply
  * use the addWithoutBreaking method to ensure success insertion in cacheMap.
  */
-public class ClusteredPostingCache extends SparseCache<ClusteredPostingCacheItem> {
+public class ClusteredPostingCache extends MemMonitoredCache<ClusteredPostingCacheItem> {
 
-    private static final ClusteredPostingCache INSTANCE = new ClusteredPostingCache();
+    private static ClusteredPostingCache INSTANCE;
```
zhichao-aws: we should add `volatile` here

https://dev.to/devcorner/thread-safe-singleton-in-java-understanding-volatile-and-double-checked-locking-3d1a

src/main/java/org/opensearch/neuralsearch/stats/metrics/MetricStatsManager.java
```diff
@@ -0,0 +1,54 @@
...
+public class MetricStatsManager {
+    private static MetricStatsManager INSTANCE;
+
+    /**
+     * Returns the singleton instance of EventStatsManager.
+     * Creates a new instance with default settings if one doesn't exist.
+     *
+     * @return The singleton instance of EventStatsManager
+     */
+    public synchronized static MetricStatsManager instance() {
```
zhichao-aws: We can use DCL for better performance and code consistency

src/main/java/org/opensearch/neuralsearch/sparse/cache/MemoryUsageManager.java
```diff
@@ -0,0 +1,61 @@
...
+     */
+    public void setLimit(ByteSizeValue limit) {
+        memoryUsageTracker.setCanRecordIncrementChecker(getInternalAndCbPredicate(limit));
+    }
+
+    @VisibleForTesting
+    BiPredicate<Long, Long> getInternalAndCbPredicate(ByteSizeValue limit) {
+        return (bytes, targetedTotalBytes) -> {
+            // check if memory limit exceeds, then check if circuit breaker is triggered
+            return targetedTotalBytes <= limit.getBytes() && CircuitBreakerManager.addMemoryUsage(bytes, SEISMIC);
```
zhichao-aws: since we're doing 2 checking for same purpose, is it possible that there are inconsistency? E.g. RamBytesRecorder add bytes usage, but circuit breaker reject. And the number will be different

src/main/java/org/opensearch/neuralsearch/sparse/cache/RamBytesRecorder.java
```diff
@@ -0,0 +1,81 @@
...
+        totalBytes.addAndGet(bytes);
+        return true;
+    }
+
+    /**
+     * Records bytes without validation and executes post-action.
+     * @param bytes byte count to add
+     * @param postAction action to execute after recording
+     */
+    public synchronized void safeRecord(long bytes, Consumer<Long> postAction) {
```
zhichao-aws: suggest to use consistent naming. -> recordWithoutBreaking

src/main/java/org/opensearch/neuralsearch/stats/metrics/MemoryStat.java
```diff
@@ -0,0 +1,71 @@
...
+                return ForwardIndexCache.getInstance().ramBytesUsed();
+            case MetricStatName.MEMORY_SPARSE_CLUSTERED_POSTING_USAGE:
+                return ClusteredPostingCache.getInstance().ramBytesUsed();
+            default:
+                throw new IllegalArgumentException(String.format(Locale.ROOT, "Metric stat not found: %s", statName));
+        }
+    }
+
+    @Override
+    public synchronized String getValue() {
```
zhichao-aws: Why make it synchronized and set byteSize AtomicLong? Can we declare byteSize method-scope here?

src/main/java/org/opensearch/neuralsearch/sparse/cache/MemoryUsageManager.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+        return instance;
+    }
+
+    /**
+     * Sets the memory limit for sparse vector operations.
+     *
+     * @param limit the maximum memory limit
+     */
+    public void setLimit(ByteSizeValue limit) {
```
zhichao-aws: currently it's only called during node initialization. We should also update this limit when users change the CBcluster settings, correct?

src/main/java/org/opensearch/neuralsearch/sparse/cache/ClusteredPostingCache.java
zhichao-aws: And I think we should make code style consistent. If we use DCL here, we should also apply to ForwardIndexCache

src/main/java/org/opensearch/neuralsearch/stats/metrics/MemoryStat.java
```diff
@@ -0,0 +1,71 @@
...
+        this.byteSize.set(getByteSize());
+        if (statName == MetricStatName.MEMORY_SPARSE_MEMORY_USAGE_PERCENTAGE) {
+            MemoryMXBean memoryMXBean = ManagementFactory.getMemoryMXBean();
+            long heapMaxBytes = memoryMXBean.getHeapMemoryUsage().getMax();
```
zhichao-aws: > Returns the maximum amount of memory in bytes; -1 if undefined

Should we add some special checks if it returns -1?

src/main/java/org/opensearch/neuralsearch/sparse/cache/MemMonitoredCache.java
```diff
@@ -29,12 +29,12 @@ public abstract class SparseCache<T extends Accountable> implements Accountable
...
      * @param key The CacheKey that identifies the index to be removed
      */
-    public void removeIndex(@NonNull CacheKey key) {
+    public void onIndexRemoval(@NonNull CacheKey key) {
         cacheMap.computeIfPresent(key, (k, value) -> {
-            long ramBytesUsed = value.ramBytesUsed() + RamUsageEstimator.shallowSizeOf(k);
-            CircuitBreakerManager.releaseBytes(ramBytesUsed);
-            LruTermCache.getInstance().removeIndex(k);
-            LruDocumentCache.getInstance().removeIndex(k);
+            long ramBytesUsed = value.ramBytesUsed() + RamUsageEstimator.shallowSizeOf(key);
```
zhichao-aws: we should use `k` here for better readability.

src/main/java/org/opensearch/neuralsearch/sparse/cache/RamBytesRecorder.java
zhichao-aws: And can you explain why we need synchronized here? Is the postAction not thread-safe?

src/main/java/org/opensearch/neuralsearch/stats/metrics/MemoryStatSnapshot.java
```diff
@@ -0,0 +1,114 @@
...
+            return null;
+        }
+
+        MetricStatName name = null;
+        long totalByteSize = 0;
+
+        for (MemoryStatSnapshot stat : snapshots) {
+            // Mixed version clusters may have nodes that return null stat snapshots not available on older versions.
+            // If so, exclude those from aggregation
+            if (stat == null) {
```
zhichao-aws, line 75: if all `stat` are null, `name` will be null. And we'll create a MemoryStatSnapshot with `null` as name. We'll have NPE when call toXContent

src/main/java/org/opensearch/neuralsearch/sparse/cache/AccountableTracker.java
```diff
@@ -0,0 +1,29 @@
...
+public abstract class AccountableTracker implements Accountable {
+    private final RamBytesRecorder recorder = new RamBytesRecorder(0);
+
+    /**
+     * Records the number of bytes used by an object.
+     *
+     * @param bytes number of bytes to record
+     */
+    public void recordUsedBytes(long bytes) {
+        recorder.record(bytes);
```
zhichao-aws: why don't use safe record here? as we don't have any handler if the recorder rejects

src/main/java/org/opensearch/neuralsearch/stats/NeuralStatsInput.java
zhichao-aws: We need to resolve it before merge the PR

src/main/java/org/opensearch/neuralsearch/sparse/cache/ClusteredPostingCache.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/stats/metrics/MetricStatsManager.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/sparse/cache/RamBytesRecorder.java
yuye-aws: > suggest to use consistent naming. -> recordWithoutBreaking

Ack. I personally prefer recordWithoutValidation because we do not trip like the circuit breaker

src/main/java/org/opensearch/neuralsearch/sparse/cache/MemoryUsageManager.java
yuye-aws: Yes. It's possible. For example, the left check passed but the parent circuit breaker trips. 

The consistency is ensured in **record** method in RamBytesRecorder

src/main/java/org/opensearch/neuralsearch/sparse/cache/MemMonitoredCache.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/sparse/cache/RamBytesRecorder.java
yuye-aws: > And can you explain why we need synchronized here? Is the postAction not thread-safe?

Nice catch! The postAction acts like the **addWithoutBreaking** method from circuit breaker, which is thread-safe. This method do not need to marked as **synchronized**

src/main/java/org/opensearch/neuralsearch/stats/metrics/MemoryStat.java
yuye-aws: Nice catch. The **synchronized** is introduced to ensure consistency between the string value and long value. A method-scope byteSize variable suffices. I'll remove the **synchronized**

src/main/java/org/opensearch/neuralsearch/sparse/cache/MemoryUsageManager.java
yuye-aws: Thanks. I found this PR also needs to include setLimitAndOverhead. The function registerSettingsCallbacks in NeuralSearchSettingsAccessor has been updated

src/main/java/org/opensearch/neuralsearch/sparse/cache/AccountableTracker.java
yuye-aws: Nice catch. Updated with `recorder.recordWithoutValidation(bytes, null);`

src/main/java/org/opensearch/neuralsearch/stats/metrics/MemoryStatSnapshot.java
yuye-aws: I think you are referring to the case where all stat is null from the snapshots Collection. The NPE will be thrown from this line:
```
builder.field(StatSnapshot.STAT_TYPE_FIELD, statName.getStatType().getTypeString())
```

Updated with this line:
```
builder.field(StatSnapshot.STAT_TYPE_FIELD, statName == null ? null : statName.getStatType().getTypeString());
```

src/main/java/org/opensearch/neuralsearch/stats/metrics/MemoryStat.java
yuye-aws: Ack

zhichao-aws (approved):

src/main/java/org/opensearch/neuralsearch/stats/NeuralStatsInput.java
yuye-aws: @q-andy BWC test gets passed after updating the PR. Resolving this thread now.

src/main/java/org/opensearch/neuralsearch/stats/NeuralStatsInput.java
```diff
@@ -144,6 +168,11 @@ public NeuralStatsInput(StreamInput input) throws IOException {
             includeAllNodes = true;
             includeInfo = true;
         }
+        if (isClusterOnOrAfterMinReqVersionForMetricStats()) {
+            includeMetrics = input.readBoolean();
+        } else {
+            includeMetrics = true;
+        }
```
q-andy, line 172: nit: I recommend combining these into a single if block at the bottom. Since the order of the input/output is only for developer use, we don't have to keep the similar fields next to each other. To me its cleaner to group them by the version check instead of by similar fields. Otherwise when we add new fields, it will make the code less readable

src/main/java/org/opensearch/neuralsearch/transport/NeuralStatsResponse.java
```diff
@@ -49,11 +53,16 @@ public class NeuralStatsResponse extends BaseNodesResponse<NeuralStatsNodeRespon
     public NeuralStatsResponse(StreamInput in) throws IOException {
         super(new ClusterName(in), in.readList(NeuralStatsNodeResponse::readStats), in.readList(FailedNodeException::new));
         Map<String, StatSnapshot<?>> castedInfoStats = (Map<String, StatSnapshot<?>>) (Map) in.readMap();
+        Map<String, StatSnapshot<?>> castedMetricStats = Map.of();
+        if (isClusterOnOrAfterMinReqVersionForMetricStats()) {
+            castedMetricStats = (Map<String, StatSnapshot<?>>) (Map) in.readMap();
+        }
```
q-andy: same comment about ordering in a single if block as above here

src/main/java/org/opensearch/neuralsearch/transport/NeuralStatsResponse.java
```diff
@@ -80,39 +94,51 @@ public NeuralStatsResponse(StreamInput in) throws IOException {
...
         super.writeTo(out);
         Map<String, Object> downcastedInfoStats = (Map<String, Object>) (Map) (infoStats);
+        Map<String, Object> downcastedMetricStats = (Map<String, Object>) (Map) (metricStats);
         Map<String, Object> downcastedAggregatedNodeStats = (Map<String, Object>) (Map) (aggregatedNodeStats);
         Map<String, Object> downcastedNodeIdToNodeEventStats = (Map<String, Object>) (Map) (nodeIdToNodeEventStats);
 
         out.writeMap(downcastedInfoStats);
+        if (isClusterOnOrAfterMinReqVersionForMetricStats()) {
+            out.writeMap(downcastedMetricStats);
+        }
```
q-andy: same here

src/main/java/org/opensearch/neuralsearch/transport/NeuralStatsTransportAction.java
```diff
@@ -178,6 +208,29 @@ private Map<String, StatSnapshot<?>> aggregateNodesResponses(
...
+                }
+            }
+
+            MemoryStatSnapshot aggregatedMetricSnapshot = MemoryStatSnapshot.aggregateMetricSnapshots(memoryStatSnapshotCollection);
+
+            // Skip adding null event stats. This happens when a node id parameter is invalid.
+            if (aggregatedMetricSnapshot != null) {
+                aggregatedMap.put(metricStatName.getFullPath(), aggregatedMetricSnapshot);
+            }
+        }
```
q-andy, line 218: If user specifies include_all_nodes=true but include_metrics=false, does metrics still appear in aggregated response? Wondering if we need an extra condition to check include_metrics flag to exclude metrics from aggregate response here

heemin32:
> Readability makes sense, my thought was that user might want to alarm on something like memory usage for operational work. I'm fine either way.

I agree with @q-andy. The APIs are mainly designed for system, not for people.

src/main/java/org/opensearch/neuralsearch/stats/NeuralStatsInput.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/transport/NeuralStatsResponse.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/transport/NeuralStatsResponse.java
yuye-aws: Ack

yuye-aws:
> I agree with @q-andy. The APIs are mainly designed for system, not for people.

Ack. I'll update this PR so that all usage metrics will be in kilos

src/main/java/org/opensearch/neuralsearch/transport/NeuralStatsTransportAction.java
yuye-aws: The metrics won't appear in the aggregated response. This is expected because aggregateNodesResponses will only get called when isIncludeAllNodes is true
```
if (request.getNeuralStatsInput().isIncludeAllNodes()) {
    aggregatedNodeStats = aggregateNodesResponses(
        responses,
        request.getNeuralStatsInput().getEventStatNames(),
        request.getNeuralStatsInput().getMetricStatNames()
    );
}
```

src/main/java/org/opensearch/neuralsearch/sparse/SparseIndexEventListener.java
```diff
@@ -0,0 +1,43 @@
...
+import org.opensearch.index.shard.IndexEventListener;
+import org.opensearch.index.shard.IndexShard;
+import org.opensearch.indices.cluster.IndicesClusterStateService;
+import org.opensearch.neuralsearch.sparse.cache.ClusteredPostingCache;
+import org.opensearch.neuralsearch.sparse.cache.CacheKey;
+import org.opensearch.neuralsearch.sparse.cache.ForwardIndexCache;
+import org.opensearch.neuralsearch.sparse.mapper.SparseTokensFieldType;
+
+@AllArgsConstructor
+public class SparseIndexEventListener implements IndexEventListener {
```
zane-neo: unused class

src/main/java/org/opensearch/neuralsearch/stats/metrics/MetricStat.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Interface for metric stats. These contain logic to store and update ongoing metric information.
+ */
+public interface MetricStat {
+    /**
+     * Returns a single point in time value associated with the stat.
+     * @return the value of the stat
+     */
+    Double getValue();
```
zane-neo: I see only the `getStatSnapshot` method is using this method, why do we need it in interface?

src/main/java/org/opensearch/neuralsearch/stats/metrics/MetricStatName.java
```diff
@@ -0,0 +1,112 @@
...
+     * @param path the unique path of the stat
+     * @param statType the category of stat
+     */
+    MetricStatName(String nameString, String path, MetricStatType statType, Version version) {
+        this.nameString = nameString;
+        this.path = path;
+        this.statType = statType;
+        this.version = version;
+
+        switch (statType) {
```
zane-neo: [nit] single switch can be replaced with if.

src/main/java/org/opensearch/neuralsearch/sparse/SparseIndexEventListener.java
yuye-aws: Ack. This class will be introduced in the next PR

src/main/java/org/opensearch/neuralsearch/stats/metrics/MetricStatName.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/stats/metrics/MetricStat.java
yuye-aws: Ack. This can be removed

zane-neo (approved):

PR #1554, opened 2025-09-09
SEISMIC: index thread quantity setting API

zirui-song-18:
### Description
In this PR, I implemented a framework of how to control thread quantity. The idea of this design can be seen in this [HLD](https://github.com/opensearch-project/neural-search/issues/1390#:~:text=4.-,Cluster%20Setting,-Parameter). With this feature, users can decide how many CPU cores will be involved during `force_merge`. The default value of `index_thread_qty` is half of available CPU cores. An example usage command will be shown below, and more details about it can be seen in this [blog](https://docs.opensearch.org/docs/latest/install-and-configure/configuring-opensearch/index/#updating-cluster-settings-using-the-api).
```
PUT _cluster/settings
{
  "transient" : {
    "neural.sparse.algo_param.index_thread_qty" : 48
  }
}
```

 - [x] Basic classes
 - [x] Cache
 - [x] Codec
 - [x] Query
 - [ ] BWCs
 - [ ] **Interfaces**

**We target version 3.3 release**.

### Related Issues
RFC: #1335 
High level design: #1390 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
+++ b/src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
@@ -20,6 +20,9 @@ public final class NeuralSearchSettings {
 
     public static final String SPARSE_ALGO_PARAM_INDEX_THREAD_QTY = "neural.sparse.algo_param.index_thread_qty";
     public static final String NEURAL_CIRCUIT_BREAKER_NAME = "neural_search";
+    public static final int DEFAULT_INDEX_THREAD_QTY = 1; // Choosing 1 as default value to protect safety
+    public static final int MINIMUM_INDEX_THREAD_QTY = 1;
+    public static final int MAXIMUM_INDEX_THREAD_QTY = 1024;
 
     /**
      * Specifies the initial memory limit for the parent circuit breaker.
@@ -92,9 +95,9 @@ public final class NeuralSearchSettings {
 
     public static Setting<Integer> SPARSE_ALGO_PARAM_INDEX_THREAD_QTY_SETTING = Setting.intSetting(
         SPARSE_ALGO_PARAM_INDEX_THREAD_QTY,
-        1, // Choosing 1 as default value to protect safety
-        1, // 1 means that user did not give specific thread quantity
-        1024,
+        DEFAULT_INDEX_THREAD_QTY,
+        MINIMUM_INDEX_THREAD_QTY,
+        MAXIMUM_INDEX_THREAD_QTY,
         Setting.Property.NodeScope,
         Setting.Property.Dynamic
     );


src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
```diff
@@ -8,14 +8,22 @@
...
 /**
  * Class defines settings specific to neural-search plugin
+ * DEFAULT_INDEX_THREAD_QTY: -1 represents that user did not give a specific thread quantity
+ * MAX_INDEX_THREAD_QTY: Initial max value, will be updated based on actual CPU cores
  */
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 public final class NeuralSearchSettings {
 
+    public static final int DEFAULT_INDEX_THREAD_QTY = -1;
+    public static int MAX_INDEX_THREAD_QTY = 1024;
```
yuye-aws: Can we initialize this variable as: `OpenSearchExecutors.allocatedProcessors(settings)`?

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
```diff
@@ -87,6 +95,34 @@ public final class NeuralSearchSettings {
         Setting.Property.Dynamic
     );
 
+    public static Setting<Integer> SPARSE_ALGO_PARAM_INDEX_THREAD_QTY_SETTING = Setting.intSetting(
+        SPARSE_ALGO_PARAM_INDEX_THREAD_QTY,
+        SPARSE_DEFAULT_ALGO_PARAM_INDEX_THREAD_QTY,
+        -1, // -1 means that user did not give specific thread quantity
```
yuye-aws: This -1 indicates the mean value. Shall we restrict that users should allocate at least one thread?

yuye-aws:
@zirui-song-18 Please fix the failing CI

src/test/java/org/opensearch/neuralsearch/plugin/NeuralSearchTests.java
```diff
@@ -73,6 +77,11 @@
 
 public class NeuralSearchTests extends OpenSearchQueryTestCase {
 
+    private static final int NEW_THREAD_COUNT = 10;
+    private static final int CURRENT_THREAD_COUNT = 6;
+    private static final int CUSTOM_THREAD_QTY = 12;
```
yuye-aws: CUSTOM_THREAD_QTY may exceed the number of CPUs on the Github CI check. Please make sure this never exceeds

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
zirui-song-18: No, we can't. When an OpenSearch cluster starts, it will first access this `NeuralSearchSettings` class before we can access to `settings`. But don't worry, once the cluster reaches `getExecutorBuilders`, we will update this variable. It is safe.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
zirui-song-18: Yes, we already restricted. once the cluster reaches `getExecutorBuilders`, we will update this variable by `updateThreadQtySettings`, where `minValue` is 1

src/test/java/org/opensearch/neuralsearch/plugin/NeuralSearchTests.java
zirui-song-18: Yes, that's why it just triggered CI failing. Now I've changed it to an environment-depending value, it should be fine now.

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
```diff
@@ -192,6 +196,12 @@ public Collection<Object> createComponents(
         infoStatsManager = new InfoStatsManager(NeuralSearchClusterUtil.instance(), settingsAccessor, pipelineServiceUtil);
         EventStatsManager.instance().initialize(settingsAccessor);
         this.xContentRegistry = xContentRegistry;
+        clusterService.getClusterSettings()
+            .addSettingsUpdateConsumer(
+                NeuralSearchSettings.SPARSE_ALGO_PARAM_INDEX_THREAD_QTY_SETTING,
```
zane-neo: Is this thread quantity used only by force_merge?

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
```diff
@@ -87,6 +95,34 @@ public final class NeuralSearchSettings {
...
+        MAX_INDEX_THREAD_QTY = maxThreadQty;
+        SPARSE_DEFAULT_ALGO_PARAM_INDEX_THREAD_QTY = threadQty;
+        SPARSE_ALGO_PARAM_INDEX_THREAD_QTY_SETTING = Setting.intSetting(
+            SPARSE_ALGO_PARAM_INDEX_THREAD_QTY,
+            SPARSE_DEFAULT_ALGO_PARAM_INDEX_THREAD_QTY,
+            1,
+            MAX_INDEX_THREAD_QTY,
+            Setting.Property.NodeScope,
+            Setting.Property.Dynamic
+        );
```
zane-neo: Why do we need this?

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
```diff
@@ -87,6 +95,34 @@ public final class NeuralSearchSettings {
...
+        MAX_INDEX_THREAD_QTY,
+        Setting.Property.NodeScope,
+        Setting.Property.Dynamic
+    );
+
+    public static int updateThreadQtySettings(Settings settings) {
+        int maxThreadQty = OpenSearchExecutors.allocatedProcessors(settings);
+        int threadQty = SPARSE_ALGO_PARAM_INDEX_THREAD_QTY_SETTING.get(settings);
+        if (threadQty == DEFAULT_INDEX_THREAD_QTY) {
+            threadQty = Math.max(maxThreadQty / 2, 1);
```
zane-neo: Is this the recommend value from us? Do we have a benchmark for different values?

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
```diff
@@ -87,6 +95,34 @@ public final class NeuralSearchSettings {
...
+    public static Setting<Integer> SPARSE_ALGO_PARAM_INDEX_THREAD_QTY_SETTING = Setting.intSetting(
+        SPARSE_ALGO_PARAM_INDEX_THREAD_QTY,
+        SPARSE_DEFAULT_ALGO_PARAM_INDEX_THREAD_QTY,
+        -1, // -1 means that user did not give specific thread quantity
+        MAX_INDEX_THREAD_QTY,
+        Setting.Property.NodeScope,
+        Setting.Property.Dynamic
+    );
+
+    public static int updateThreadQtySettings(Settings settings) {
```
zane-neo: Can we keep a simple setting and move the calculation of `threadQty` to the `updateThreadPoolSize` where it final used?

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
zirui-song-18: Yes, only used in force_merge

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
zirui-song-18: Because we want to set default thread_qty as half available CPU cores within OpenSearch, which can be defined manually in config. So, we will have to obtain `settings` in `getExecutorBuilders`, then update our initial `SPARSE_ALGO_PARAM_INDEX_THREAD_QTY_SETTING`

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
zirui-song-18: We take half of available CPU cores to prevent running out all resources. As for its impact, half setting will increase force merging time by around 10%, as part of that work is IO intensive. So this setting should be acceptable.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
zirui-song-18: Actually that's not a good idea, because `updateThreadQtySettings` is not the function for users to change thread count, but for the system to set a default value, half CPU cores. The function `updateThreadPoolSize` is the actual one to handle user's request, no dealing with initial setting.

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
yuye-aws: I think I will also be used by warmup and clear_cache operation

zirui-song-18: Thanks for @yuye-aws 's reminder. Yes, this thread pool will also be used by `warmUp` and `clearCache`.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
zane-neo: Acceptable is not a recommend value, a default value should be optimal, at least, we can use the empirical formula: `available_core * (1 + wait_time/compute_time)`

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
zane-neo: Then can we make the calculation into the `updateThreadPoolSize` method?

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
zirui-song-18: I'm a bit confused with your formula. It means that `available_core` will multiply with a number larger than 1, so we will assign more threads than available cores?

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
zirui-song-18: I don't think that's related. Because `updateThreadPoolSize` is a consumer, which is only called when users give a command to update new thread quantity. And the calculation above is to decide a default value, they are two things.

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
```diff
@@ -192,6 +196,12 @@ public Collection<Object> createComponents(
...
+                ClusterTrainingExecutor::updateThreadPoolSize
+            );
```
yuye-aws: Let's move this to a centralized place: registerSettingsCallbacks in NeuralSearchSettingsAccessor

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
yuye-aws: > Actually that's not a good idea, because `updateThreadQtySettings` is not the function for users to change thread count, but for the system to set a default value, half CPU cores.

I understand. The function name "update" is confusing. Can you update to **initailizeThreadQtySettings**?

zirui-song-18: That sounds good. Ack

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
zirui-song-18: Ack

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
zane-neo: Yes, in IO intensive case, we prefer to have more threads to make it has optimal performance (even threads number is bigger than core numbers).

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
zane-neo: Let's remove this method and move the calculation to the method that setting the thread pool value.

zirui-song-18: Yes. Removed now.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
zirui-song-18: I see. Since this is still partially computation intensive, I will set the default to 1 for safety.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
```diff
@@ -12,10 +12,16 @@
...
 /**
  * Class defines settings specific to neural-search plugin
+ * DEFAULT_INDEX_THREAD_QTY: -1 represents that user did not give a specific thread quantity
+ * MAX_INDEX_THREAD_QTY: Initial max value, will be updated based on actual CPU cores
  */
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 public final class NeuralSearchSettings {
 
+    public static final int DEFAULT_INDEX_THREAD_QTY = 1;
+    public static int MAX_INDEX_THREAD_QTY = 1024;
```
zane-neo: [nit]This can be final

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
```diff
@@ -12,10 +12,16 @@
...
+    public static final String SPARSE_ALGO_PARAM_INDEX_THREAD_QTY = "neural.sparse.algo_param.index_thread_qty";
+    public static Integer SPARSE_DEFAULT_ALGO_PARAM_INDEX_THREAD_QTY = DEFAULT_INDEX_THREAD_QTY;
```
zane-neo: This can be deleted as it's same to `DEFAULT_INDEX_THREAD_QTY`

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
```diff
@@ -12,10 +12,16 @@
 
 /**
  * Class defines settings specific to neural-search plugin
+ * DEFAULT_INDEX_THREAD_QTY: -1 represents that user did not give a specific thread quantity
+ * MAX_INDEX_THREAD_QTY: Initial max value, will be updated based on actual CPU cores
  */
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 public final class NeuralSearchSettings {
 
+    public static final int DEFAULT_INDEX_THREAD_QTY = 1;
```
zane-neo: In fact we can remove all these constants and directly put the values into the `SPARSE_ALGO_PARAM_INDEX_THREAD_QTY_SETTING`, also add comments there why we choose 1 as default value.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
zirui-song-18: Ack

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
zirui-song-18: Ack

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
zirui-song-18: Ack

zane-neo (approved):

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
```diff
@@ -97,32 +92,13 @@ public final class NeuralSearchSettings {
 
     public static Setting<Integer> SPARSE_ALGO_PARAM_INDEX_THREAD_QTY_SETTING = Setting.intSetting(
         SPARSE_ALGO_PARAM_INDEX_THREAD_QTY,
-        SPARSE_DEFAULT_ALGO_PARAM_INDEX_THREAD_QTY,
-        -1, // -1 means that user did not give specific thread quantity
-        MAX_INDEX_THREAD_QTY,
+        1, // Choosing 1 as default value to protect safety
+        1, // 1 means that user did not give specific thread quantity
+        1024,
```
yuye-aws: [nit] Can you mark 1 as DEFAULT_INDEX_THREAD_QTY and 1024 as MAX_INDEX_THREAD_QTY?

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettingsAccessor.java
```diff
@@ -53,9 +54,9 @@ private void registerSettingsCallbacks(ClusterService clusterService) {
                 MemoryUsageManager.getInstance().setLimitAndOverhead(limit, overhead);
             });
         clusterService.getClusterSettings()
-            .addSettingsUpdateConsumer(
-                NeuralSearchSettings.SPARSE_ALGO_PARAM_INDEX_THREAD_QTY_SETTING,
-                ClusterTrainingExecutor::updateThreadPoolSize
-            );
+            .addSettingsUpdateConsumer(NeuralSearchSettings.SPARSE_ALGO_PARAM_INDEX_THREAD_QTY_SETTING, (setting) -> {
+                int maxThreadQty = OpenSearchExecutors.allocatedProcessors(settings);
```
yuye-aws, line 58: Can you move int maxThreadQty = OpenSearchExecutors.allocatedProcessors(settings); to ClusterTrainingExecutor.updateThreadPoolSize?

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettingsAccessor.java
```diff
@@ -33,10 +34,10 @@ public class NeuralSearchSettingsAccessor {
     public NeuralSearchSettingsAccessor(ClusterService clusterService, Settings settings) {
         isStatsEnabled = NeuralSearchSettings.NEURAL_STATS_ENABLED.get(settings);
         isAgenticSearchEnabled = NeuralSearchSettings.AGENTIC_SEARCH_ENABLED.get(settings);
-        registerSettingsCallbacks(clusterService);
+        registerSettingsCallbacks(clusterService, settings);
     }
 
-    private void registerSettingsCallbacks(ClusterService clusterService) {
+    private void registerSettingsCallbacks(ClusterService clusterService, Settings settings) {
```
yuye-aws, line 40: [nit] Seems the settings is not used

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
zirui-song-18: Ack

PR #1557, opened 2025-09-10
SEISMIC: Query interface

chishui:
### Description
This PR contains:
1. Wire SparseAnnQueryBuilder to NeuralSparseQueryBuilder
2. Add MLAlgoParams to MLCommonAccessor so that for seismic we specify the response should be with token id instead of token
3. In SparseEncodingProcessor, if texts are batched inferencing, and some texts need token response, some need token id response, add logic to separate these two.
4. Throw no supporting exception when running neural sparse two-phase query against seismic field
5. Add ITs for Seismic index and query features.

##### Create index
```json
{
  "settings": {
      "index": {
          "sparse": true
      }
  },
    "mappings": {
        "properties": {
            "passage_embedding":{
                "type": "sparse_tokens",
                "method": {
                    "name": "seismic",
                    "parameters": {                
                        "n_postings": 160,
                        "summary_prune_ratio": 0.4,
                        "cluster_ratio": 0.1,
                        "approximate_threshold": 1000000
                    }
                }
            }
        }
    }
}
```

##### Query
```json
POST {{index}}/_search
{
    "query": {
        "neural_sparse": {
            "passage_embedding": {
                "query_tokens": {
                    "1000": 1.0
                },
                "method": {
                    "top_k": 10,
                    "heap_factor": 1.0,
                    "k": 10,
                    "filter": {
                        "bool": {
                            "must": [
                                {
                                    "match": {
                                        "passage_text": "tree"
                                    }
                                }
                            ]
                        }
                    }
               }
          }
     }
}
```

- [x] Basic classes
- [x] Codec
- [x] Query
- [x] Cache
- [ ] BWCs
- [x] **Interfaces**

### Related Issues
RFC: #1335 
Design: #1390


src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
```diff
@@ -258,7 +259,8 @@ public Map<String, Processor.Factory> getProcessors(Processor.Parameters paramet
                 parameters.client,
                 clientAccessor,
                 parameters.env,
-                parameters.ingestService.getClusterService()
+                parameters.ingestService.getClusterService(),
+                new SparseFieldUtils(parameters.ingestService.getClusterService())
```
yuye-aws: SparseFieldUtils can directly have access to cluster service via: 

```
NeuralSearchClusterUtil.instance().getClusterService();
```

There is no need for SparseFieldUtils to hold a member variable for cluster service

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -310,7 +305,7 @@ protected List<DataForInference> getDataForInference(List<IngestDocumentWrapper>
 
     @Getter
     @AllArgsConstructor
-    protected static class DataForInference {
+    public static class DataForInference {
```
yuye-aws: [nit] Seems that protected is enough

src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
```diff
@@ -178,4 +298,118 @@ public void subBatchExecute(List<IngestDocumentWrapper> ingestDocumentWrappers,
...
+                }
+            }
+        };
+    }
+
+    private boolean isSparseAnnField(Set<String> sparseAnnFields, String field) {
+        // we don't support nested sparse ann field
+        int nestedDotIndex = field.indexOf('.');
+        return nestedDotIndex == -1 && sparseAnnFields.contains(field);
+    }
```
yuye-aws, line 403: The OpenSearch user can created nested Seismic field with the following index mapping. What would be the expected behavior to ingest sparse embeddings?

```
PUT test_seismic_nested
{
  "settings": {
    "index": {
      "sparse": "true"
    }
  },
  "mappings": {
    "properties": {
      "text": {
        "type": "text"
      },
      "passage_chunk_embedding": {
        "type": "nested",
        "properties": {
          "sparse": {
            "type": "sparse_tokens",
            "method": {
              "name": "seismic",
              "parameters": {
                "n_postings": 160,
                "cluster_ratio": 0.075,
                "summary_prune_ratio": 0.4,
                "approximate_threshold": 25000
              }
            }
          }
        }
      }
    }
  }
}
```

src/main/java/org/opensearch/neuralsearch/sparse/common/SparseFieldUtils.java
```diff
@@ -0,0 +1,67 @@
...
+    private final ClusterService clusterService;
+
+    /**
+     * Retrieves all sparse ANN fields from a given index.
+     *
+     * @param index The name of the index
+     * @return A set of field names that are configured as sparse token fields, or an empty set if none exist
+     */
+    @SuppressWarnings("unchecked")
+    public Set<String> getSparseAnnFields(String index) {
```
yuye-aws: This function is actually checking the Seismic field. In the future, we may have more sparse ANN algorithms. How about renaming ad getSeismicAnnFields. This comment also applies to the sparse ann query builder.

src/main/java/org/opensearch/neuralsearch/sparse/common/SparseFieldUtils.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+        Object properties = mappingMetadata.sourceAsMap().get("properties");
+        if (!(properties instanceof Map)) {
+            return Collections.emptySet();
+        }
+        Set<String> sparseAnnFields = new HashSet<>();
+        Map<String, Object> fields = (Map<String, Object>) properties;
+        for (Map.Entry<String, Object> field : fields.entrySet()) {
+            Map<String, Object> fieldMap = (Map<String, Object>) field.getValue();
+            Object type = fieldMap.get("type");
```
yuye-aws, line 56: Can we check nested field?

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -568,4 +645,24 @@ private static boolean isClusterOnOrAfterMinReqVersionForAnalyzer() {
...
+     * Lazy load of SparseFieldUtils
+     * @return {@link SparseFieldUtils}
+     */
+    @VisibleForTesting
+    SparseFieldUtils getSparseFieldUtils() {
+        if (sparseFieldUtils == null) {
+            sparseFieldUtils = new SparseFieldUtils(NeuralSearchClusterUtil.instance().getClusterService());
+        }
+        return sparseFieldUtils;
+    }
```
yuye-aws: This may be other class using sparseFieldUtils. How about moving the instantiation code to the sparseFieldUtils class? sparseFieldUtils can be marked as singleton with double checked locking

src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
```diff
@@ -66,14 +80,16 @@ public SparseEncodingProcessor(
         OpenSearchClient openSearchClient,
         MLCommonsClientAccessor clientAccessor,
         Environment environment,
-        ClusterService clusterService
+        ClusterService clusterService,
+        SparseFieldUtils sparseFieldUtils
```
yuye-aws: No need to pass in a util class instance. Once SparseFieldUtils is declared to be a singleton, we can directly fetch it

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ## [Unreleased 3.x](https://github.com/opensearch-project/neural-search/compare/main...HEAD)
 
 ### Features
+- [SEISMIC] Support SEISMIC, a new sparse ANN algorithm
```
yuye-aws: We can include the existing PR links?

yuye-aws (commented):
A two-phase IT fails: org.opensearch.neuralsearch.processor.NeuralSparseTwoPhaseProcessorIT.testNeuralQuerySemanticFieldSparseEmbedding_whenTwoPhaseWithSearchAnalyzer_thenGetExpectedScore

src/test/java/org/opensearch/neuralsearch/sparse/SparseIndexingIT.java
```diff
@@ -178,6 +229,462 @@ public void testSparseTokensFieldWithAdditionParameters() throws IOException {
...
+            Map.of("1000", 0.1f),
+            filter
+        );
+
+        Map<String, Object> searchResults = search(TEST_INDEX_NAME, neuralSparseQueryBuilder, 20);
+        assertNotNull(searchResults);
+        assertTrue(getHitCount(searchResults) <= 15);
+    }
+
+    public void testSearchDocumentsWithTwoPhaseSearchProcessorThenThrowException() throws Exception {
```
yuye-aws: This IT class have a few test cases related to search. How about creating a separated IT class named: **SparseSearchingIT**?

src/test/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessorTests.java
```diff
@@ -798,6 +834,172 @@ public void test_execute_emptyInferenceList_successful() {
         verify(resultHandler).accept(resultCallback.capture());
     }
 
+    public void testExecute_with_all_seismic_field() {
+        Map<String, Object> sourceAndMetadata = new HashMap<>();
+        sourceAndMetadata.put(IndexFieldMapper.NAME, "my_index");
+        sourceAndMetadata.put(KEY1, "value1");
+        sourceAndMetadata.put(KEY2, "value2");
```
yuye-aws: nit: VALUE1 and VALUE2

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilderTests.java
```diff
@@ -438,29 +498,38 @@ public void testToXContent() {
...
+        testStreams(true, true);
         testStreamsWithQueryTokensOnly();
     }
 
     public void testStreams_whenMinVersionIsBeforeDefaultModelId_thenSuccess() {
         setUpClusterService(Version.V_2_12_0);
-        testStreams(false);
+        testStreams(false, false);
         testStreamsWithQueryTokensOnly();
     }
```
yuye-aws, line 549: Need one more test case with version where seismic is not supported and default model id is supported

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilderTests.java
```diff
@@ -819,8 +974,8 @@ public void testDoToQuery_successfulDoToQuery() {
         doAnswer(invocation -> mockedMappedFieldType).when(mockedQueryShardContext).fieldMapper(any());
 
         BooleanQuery.Builder targetQueryBuilder = new BooleanQuery.Builder();
-        targetQueryBuilder.add(FeatureField.newLinearQuery(FIELD_NAME, "hello", 1.f), BooleanClause.Occur.SHOULD);
-        targetQueryBuilder.add(FeatureField.newLinearQuery(FIELD_NAME, "world", 2.f), BooleanClause.Occur.SHOULD);
+        targetQueryBuilder.add(FeatureField.newLinearQuery(FIELD_NAME, "1000", 1.f), BooleanClause.Occur.SHOULD);
+        targetQueryBuilder.add(FeatureField.newLinearQuery(FIELD_NAME, "2000", 2.f), BooleanClause.Occur.SHOULD);
```
yuye-aws: Any reason why change the word token into int token? The sparseEncodingQueryBuilder can still work with word token

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
chishui: I know, it's for better testability.

src/main/java/org/opensearch/neuralsearch/sparse/common/SparseFieldUtils.java
chishui: We have quite a few such namings, let's make refactoring when we actually support a second algorithm.

src/main/java/org/opensearch/neuralsearch/sparse/common/SparseFieldUtils.java
chishui: I don't think we are supposed to support nested field.

src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
chishui: I don't think we are supposed to support nested field.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
chishui: It's to better testability. I'm reluctant to introduce another singleton class if not necessary.

src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
chishui: It's dependency injection for better testability.

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilderTests.java
chishui: they are unrelated and tested separately

src/test/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilderTests.java
chishui: restored

src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
```diff
@@ -121,10 +136,114 @@ public void doExecute(
...
+        List<IngestDocumentWrapper> ingestDocumentWrappers,
+        List<String> inferenceList,
+        List<DataForInference> dataForInferences,
+        Consumer<List<IngestDocumentWrapper>> handler
+    ) {
+        if (CollectionUtils.isEmpty(ingestDocumentWrappers)) {
+            handler.accept(ingestDocumentWrappers);
+            return;
+        }
+        // https://tiny.amazon.com/imoy8qta ingest documents in a batch belong to the same index
```
zhichao-aws: We should use public links here

src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
```diff
@@ -121,10 +136,114 @@ public void doExecute(
...
+        Object indexObj = ingestDocumentWrappers.getFirst().getIngestDocument().getSourceAndMetadata().get(INDEX_FIELD);
```
zhichao-aws, line 151: In below logics there is a prerequiste that all ingest documents are from same index.

Is it possible that these docs are from different index?

chishui: No, the implementation I linked categorize docs from same index and send them to processor

src/main/java/org/opensearch/neuralsearch/sparse/common/SparseFieldUtils.java
yuye-aws: Can you add a TODO on sparse ann naming?

chishui: There are quite a few, when we need to support another algorithm, we'll need to refactor these anyway.

yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zhichao-aws: Currently for every search request, it will create a new instance of SparseFieldUtils. What's the concern to build it a singleton or make this method static?

chishui: Static functions don't have good testability. The community is strongly against mocking static function and does not recommend using static functions [ref1](https://github.com/opensearch-project/OpenSearch/issues/14420), [ref2](https://github.com/opensearch-project/neural-search/pull/1528). 

I'll make it a singleton and inject the instance to each class to avoid creating instance for every query.

src/main/java/org/opensearch/neuralsearch/sparse/common/SparseFieldUtils.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * Utility class for operations related to sparse fields in neural search indices.
+ */
+@NoArgsConstructor(access = lombok.AccessLevel.PACKAGE)
```
zane-neo: Since it's a util class, shouldn't we make it `PRIVATE`?

src/main/java/org/opensearch/neuralsearch/sparse/common/SparseFieldUtils.java
zane-neo: This is a util class, can we make this method static so that we don't need it a singleton?

src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
```diff
@@ -121,10 +136,115 @@ public void doExecute(
...
+            for (Map.Entry<String, Object> entry : dataForInference.getProcessMap().entrySet()) {
+                if (isSparseAnnField(sparseAnnFields, entry.getKey())) {
+                    tokenIdProcessMap.put(entry.getKey(), entry.getValue());
+                } else {
+                    wordProcessMap.put(entry.getKey(), entry.getValue());
+                }
+            }
+            List<String> tokenIdList = createInferenceList(tokenIdProcessMap);
+            List<String> wordList = createInferenceList(wordProcessMap);
+            splitDataResponse.tokenIdResponseInferenceList.addAll(tokenIdList);
```
zane-neo: Shouldn't we use getter instead of exposing the field as public?

src/main/java/org/opensearch/neuralsearch/sparse/common/SparseFieldUtils.java
chishui: Will update

src/main/java/org/opensearch/neuralsearch/processor/SparseEncodingProcessor.java
chishui: good point!

src/main/java/org/opensearch/neuralsearch/sparse/common/SparseFieldUtils.java
chishui: Please check this comment https://github.com/opensearch-project/neural-search/pull/1557#discussion_r2339029011

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
zane-neo: I understand we're against mocking static, but from this code, it seems we don't need to mock static to test. Can you let me know which part you've concern and have to mock?

chishui: Please find all mock code `mock(SparseFieldUtils.class);`

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -568,4 +647,20 @@ private static boolean isClusterOnOrAfterMinReqVersionForAnalyzer() {
...
+
+    private boolean isSeismicSupported() {
+        return NeuralSearchClusterUtil.instance().getClusterMinVersion().onOrAfter(MINIMAL_SUPPORTED_VERSION_SEISMIC);
+    }
+
+    private boolean isSeismicFieldType(MappedFieldType fieldType) {
+        return isSeismicSupported() && Objects.nonNull(fieldType) && SparseTokensFieldType.isSparseTokensType(fieldType.typeName());
+    }
+
+    @VisibleForTesting
```
zane-neo: This method is only used [here](https://github.com/opensearch-project/neural-search/pull/1557/files#diff-c2736fa58eb7d17649ddcad13edab06a2f89cf888d8165d1fb568d09a948596dR444), it can be replaced to `NeuralSearchClusterUtil.instance().getClusterService()` directly, also I didn't see this been covered by any test. I think this can be deleted.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
```diff
@@ -82,6 +89,7 @@ protected NeuralSparseTwoPhaseProcessor(
             );
         }
         this.maxWindowSize = maxWindowSize;
+        this.clusterService = NeuralSearchClusterUtil.instance().getClusterService();
```
zane-neo, line 92: Since we're retrieving the clusterService from a util class, so it seems we can use this `NeuralSearchClusterUtil.instance().getClusterService()` anywhere we need `clusterService`.

zane-neo: [Out of the scope] This `NeuralSearchClusterUtil` is not elegant we shouldn't have this class at the first place, a recommend way to get clusterService is to set it to target class in `NeuralSearch`. The `instance` is not performant here as it uses `synchronized` in method level.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
chishui: It's a overwritten getter for clusterService, will remove annotation and make it public. Because initiating clusterService through constructor requires many changes, so just override it with lazy initiation.

src/main/java/org/opensearch/neuralsearch/processor/NeuralSparseTwoPhaseProcessor.java
chishui: It's for testability. synchronized could be refactored to only kick in when it's not initialized.

Issue #1525, opened 2025-08-25
[RFC] Design for Next Phase of Agentic Search


owaiskazi19:
# Design for Next Phase of Agentic Search

## Background

In 3.2 we released experimental version of [Agentic Search](https://docs.opensearch.org/latest/vector-search/ai-search/agentic-search/), where we are relying on queries generated from LLM with a [default prompt](https://github.com/opensearch-project/ml-commons/blob/main/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningPromptTemplate.java). The issue with such DSL is the accuracy and giving more control to LLM. To make the design more reliable and robust, we need to use the LLM to identify the intent of the query, understand it and fill a structure for us. This way the queries generated are structured and would require less validation as well.

The next phase centers the LLM on intent identification and template filling instead of freeform authoring. We standardize on search templates as the exchange format between Agent and the LLM. We can include 2 different configurations or modes:

* **Default Search Template (OpenSearch-provided)**: A powerful, general template that covers the bulk of use cases—multiple lexical fields, optional neural retrieval, rich filters, and multi-sort. The LLM fills only the fields required by the question and returns a filled template with no placeholders.
* **User-Defined Search Templates**: Users supply a curated set of stored templates (via OpenSearch Search Template APIs). The LLM picks the best one (two-step LLM: select then fill) and returns a filled template with no placeholders.

This approach would improve accuracy, reduces validation surface, and preserves engine-side control of structure, safety, and performance while maintaining natural-language usability for end users.

## Use Cases for Agentic Search

* **Query Understanding and Intent Recognition** - Makes search more accessible through natural language and search quality through query understanding from LLM. This understanding goes beyond simple keyword matching - it comprehends relationships between terms and the user's search objectives.
* **Query Optimization** - Optimizes queries for better performance. The optimization considers various factors such as field importance and search performance.
* **Search Enhancement Features** - Through its template-based approach, Agentic Search enriches queries with advanced search features that users might not explicitly request or know about. The system can automatically incorporate relevant search enhancements such as field boosting, multi-field searching.
* **Semantic Understanding -** The semantic understanding capability enables Agentic Search to grasp the meaning and context behind search queries. When processing a natural language query, the system considers not just the literal words but their semantic implications

#### Functional Requirement

* User registers an Agent with QueryPlanningTool and optionally WebSearchTool.
* User issues an agentic query with context and agent_id; may pass query_fields as hints.
[...]

PR #1568, opened 2025-09-16
[Agentic Search] Convert agentic query translator processor to system-generated processor

owaiskazi19:
### Description
Convert agentic query translator processor to system-generated processor

```
{
    "query": {
        "agentic": {
            "query_text": "List all species",
            "agent_id": "jf-WUZkBl9tG0YB4F8-A",
            "query_fields": ["species", "petal_length_in_cm"]
        }
    }
}
```

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
@@ -5,6 +5,7 @@
 package org.opensearch.neuralsearch.processor;
 
 import com.google.gson.Gson;
+import lombok.AllArgsConstructor;
 import lombok.extern.log4j.Log4j2;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.common.xcontent.XContentType;
@@ -20,7 +21,8 @@ import org.opensearch.neuralsearch.stats.events.EventStatName;
 import org.opensearch.neuralsearch.stats.events.EventStatsManager;
 import org.opensearch.neuralsearch.util.NeuralSearchClusterUtil;
 import org.opensearch.search.builder.SearchSourceBuilder;
-import org.opensearch.search.pipeline.AbstractProcessor;
+import org.opensearch.search.pipeline.SystemGeneratedProcessor;
+import org.opensearch.search.pipeline.ProcessorGenerationContext;
 import org.opensearch.search.pipeline.Processor;
 import org.opensearch.search.pipeline.SearchRequestProcessor;
 import org.opensearch.search.pipeline.PipelineProcessingContext;
@@ -31,29 +33,29 @@ import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 
-import static org.opensearch.ingest.ConfigurationUtils.readStringProperty;
-
 @Log4j2
-public class AgenticQueryTranslatorProcessor extends AbstractProcessor implements SearchRequestProcessor {
+public class AgenticQueryTranslatorProcessor implements SearchRequestProcessor, SystemGeneratedProcessor {
 
     public static final String TYPE = "agentic_query_translator";
     private static final int MAX_AGENT_RESPONSE_SIZE = 10_000;
     private final MLCommonsClientAccessor mlClient;
-    private final String agentId;
     private final NamedXContentRegistry xContentRegistry;
-    private static final Gson gson = new Gson();;
+    private final String tag;
+    private final String description;
+    private final boolean ignoreFailure;
+    private static final Gson gson = new Gson();
 
     AgenticQueryTranslatorProcessor(
         String tag,
         String description,
         boolean ignoreFailure,
         MLCommonsClientAccessor mlClient,
-        String agentId,
         NamedXContentRegistry xContentRegistry
     ) {
-        super(tag, description, ignoreFailure);
+        this.tag = tag;
+        this.description = description;
+        this.ignoreFailure = ignoreFailure;
         this.mlClient = mlClient;
-        this.agentId = agentId;
         this.xContentRegistry = xContentRegistry;
     }
 
@@ -80,12 +82,7 @@ public class AgenticQueryTranslatorProcessor extends AbstractProcessor implement
 
         // Validate that agentic query is used alone without other search features
         if (hasOtherSearchFeatures(sourceBuilder)) {
-            String errorMessage = String.format(
-                Locale.ROOT,
-                "Agentic search blocked - Invalid usage with other search features - Agent ID: [%s], Query: [%s]",
-                agentId,
-                agenticQuery.getQueryText()
-            );
+            String errorMessage = "Agentic search blocked - Invalid usage with other search features";
             requestListener.onFailure(new IllegalArgumentException(errorMessage));
             return;
         }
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
```diff
@@ -80,12 +82,7 @@ public void processRequestAsync(
 
         // Validate that agentic query is used alone without other search features
         if (hasOtherSearchFeatures(sourceBuilder)) {
-            String errorMessage = String.format(
-                Locale.ROOT,
-                "Agentic search blocked - Invalid usage with other search features - Agent ID: [%s], Query: [%s]",
-                agentId,
-                agenticQuery.getQueryText()
-            );
+            String errorMessage = "Agentic search blocked - Invalid usage with other search features";
```
bzhangam: Should we make this error message more clear to tell users what are other search features?

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
```diff
@@ -221,11 +225,7 @@ public AgenticQueryTranslatorProcessor create(
                     "Agentic search is currently disabled. Enable it using the 'plugins.neural_search.agentic_search_enabled' setting."
                 );
             }
-            String agentId = readStringProperty(TYPE, tag, config, "agent_id");
-            if (agentId == null || agentId.trim().isEmpty()) {
-                throw new IllegalArgumentException("agent_id is required for agentic_query_translator processor");
-            }
-            return new AgenticQueryTranslatorProcessor(tag, description, ignoreFailure, mlClient, agentId, xContentRegistry);
+            return new AgenticQueryTranslatorProcessor(tag, description, ignoreFailure, mlClient, xContentRegistry);
```
bzhangam: nit: We can use a better description for the processor otherwise it will use a default description for system generated processor. Even though today we will not show it to users but in future we may want to display this info.

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -69,6 +71,7 @@ public AgenticSearchQueryBuilder(StreamInput in) throws IOException {
         super(in);
         this.queryText = in.readString();
         this.queryFields = in.readOptionalStringList();
+        this.agentId = in.readOptionalString();
```
bzhangam, line 74: Just want to confirm we don't need to worry about the backward compatibility since it's an experimental feature?

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -171,8 +188,15 @@ protected QueryBuilder doRewrite(QueryRewriteContext queryRewriteContext) throws
 
     @Override
     protected Query doToQuery(QueryShardContext context) throws IOException {
+        // This should not be reached if the system-generated processor is working correctly
+        if (agentId == null || agentId.trim().isEmpty()) {
+            throw new IllegalStateException(
+                "Agentic search query requires an agent_id. Provide agent_id in the query or ensure the agentic_query_translator processor is configured."
```
bzhangam: Should we remove "ensure the agentic_query_translator processor is configured" since this check is just for agentId?

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -171,8 +188,15 @@ protected QueryBuilder doRewrite(QueryRewriteContext queryRewriteContext) throws
...
+            );
+        }
         throw new IllegalStateException(
-            "Agentic search query must be used as top-level query, not nested inside other queries. Should be used with agentic_query_translator search processor"
+            "Agentic search query must be processed by the agentic_query_translator system processor before query execution. "
```
bzhangam, line 204: I think we can add a validation in the AgenticSearchQueryBuilder to ensure the processor is enabled in the cluster setting `cluster.search.enabled_system_generated_factories`.

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
owaiskazi19: Yes, that's correct since it was experimental in 3.2

owaiskazi19:
Integ tests failing because of https://github.com/opensearch-project/ml-commons/issues/4172

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
owaiskazi19: I have added a new generic method `NeuralSearchSettingsAccessor` for all the system generated processor to use to verify if they are enabled in the factories

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettingsAccessor.java
```diff
@@ -59,4 +64,13 @@ private void registerSettingsCallbacks(ClusterService clusterService, Settings s
                 ClusterTrainingExecutor.updateThreadPoolSize(maxThreadQty, setting);
             });
     }
+
+    /**
+     * Checks if the system processor is enabled
+     * @return true if the processor is enabled in cluster settings
+     */
+    public boolean isSystemGenerateProcessorEnabled(String processor) {
+        String enabledFactories = String.valueOf(clusterService.getClusterSettings().get(SYSTEM_GENERATED_PIPELINE_SETTINGS));
```
bzhangam, line 73: `ENABLED_SYSTEM_GENERATED_FACTORIES_SETTING ` is a list of strings. And I think it's better to get the value through `ENABLED_SYSTEM_GENERATED_FACTORIES_SETTING.get(clusterService.getSettings())` which will return a list of string to you.

And we also need to check if it has "*" which will enabled all the system processors.

owaiskazi19:
Based on the discussion with the team again, we have decided to go with the current approach of user-defined pipeline to avoid overloading the request with agent id. Will keep this PR open and will add another processor to this PR later. Keeping this on hold until then

owaiskazi19:
> Could you tell more about the concern regarding agent id overloading?

For every search request, we need to pass an agent_id:
```
{
    "query": {
        "agentic": {
            "query_text": "List all species",
            "agent_id": "jf-WUZkBl9tG0YB4F8-A",
            "query_fields": ["species", "petal_length_in_cm"]
        }
    }
}
```
While we can define a pipeline as a one-time setup, we can have a system-generated pipeline if users want to use an agentic query for small purposes. However, overall, having a user-defined one-time pipeline would be a better choice here.

heemin32:
> While we can define a pipeline as a one-time setup, we can have a system-generated pipeline if users want to use an agentic query for small purposes. However, overall, having a user-defined one-time pipeline would be a better choice here.

That’s debatable. If putting agent-id in the query isn’t a good idea, then the same could be said about query_field? Both can just be set in the pipeline. Personally, I feel passing agent-id in the query is simpler than creating and attaching a processor.
As the api is called by program but not by person, passing agent-id every time is not that bad imo.

owaiskazi19:
> I feel passing agent-id in the query is simpler than creating and attaching a processor. As the api is called by program but not by person, passing agent-id every time is not that bad imo.

Agreed and we can provide both the options of system generated processor and user defined processor. Will add both the types in this same PR.

> then the same could be said about query_field? 

it's an optional field for better context.

PR #1579, opened 2025-09-24
[Agentic Search] Add conversational agent support for Agentic Query

owaiskazi19:
### Description
Add conversational agent support for Agentic Query. Based on agent type, select the right agent

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
@@ -555,6 +555,9 @@ public class MLCommonsClientAccessor {
         // Add index names if present
         String[] indices = request.indices();
         if (indices != null && indices.length > 0) {
+            if (type == MLAgentType.FLOW && indices.length > 1) {
+                throw new IllegalArgumentException("Flow agent does not support multiple indices");
+            }
             parameters.put("index_name", type == MLAgentType.FLOW ? indices[0] : Arrays.toString(indices));
         }
 
--- a/src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
@@ -976,6 +976,28 @@ public class MLCommonsClientAccessorTests extends OpenSearchTestCase {
         assertEquals("Unsupported agent type: unsupported", exceptionCaptor.getValue().getMessage());
     }
 
+    public void testExecuteAgent_FlowAgentWithMultipleIndices() throws Exception {
+        final String agentId = "test-agent-id";
+        final ActionListener<String> listener = mock(ActionListener.class);
+
+        // Mock SearchRequest with multiple indices
+        SearchRequest mockRequest = mock(SearchRequest.class);
+        when(mockRequest.indices()).thenReturn(new String[] { "index1", "index2" });
+
+        // Mock AgenticSearchQueryBuilder
+        AgenticSearchQueryBuilder mockQuery = mock(AgenticSearchQueryBuilder.class);
+        when(mockQuery.getQueryText()).thenReturn("test query");
+
+        AgentInfoDTO agentInfo = new AgentInfoDTO("flow", false, false);
+
+        IllegalArgumentException exception = expectThrows(
+            IllegalArgumentException.class,
+            () -> accessor.executeAgent(mockRequest, mockQuery, agentId, agentInfo, mock(NamedXContentRegistry.class), listener)
+        );
+
+        assertEquals("Flow agent does not support multiple indices", exception.getMessage());
+    }
+
     private ModelTensorOutput createConversationalAgentResponse() {
         final List<ModelTensors> tensorsList = new ArrayList<>();
         final List<ModelTensor> mlModelTensorList = new ArrayList<>();


src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -460,45 +472,258 @@ public void inferenceSentenceHighlighting(
...
-     * Execute agent with provided parameters and return DSL query string.
-     *
-     * @param agentId    the agent ID to execute
-     * @param parameters the parameters to pass to the agent
-     * @param listener   the listener to be called with the DSL query result
+     * Get agent type from agent ID
+     */
+    public void getAgentType(@NonNull String agentId, @NonNull ActionListener<Map<String, Object>> listener) {
+        retryableGetAgentType(agentId, 0, listener);
+    }
```
rithinpullela: nit: the name `getAgentDetails` might suit the function more since we are getting more than agent type?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -460,45 +472,258 @@ public void inferenceSentenceHighlighting(
...
+            MLOutput mlOutput = (MLOutput) response.getOutput();
+            String result = null;
+            if (type == MLAgentType.FLOW) {
+                result = extractFlowAgentResult(mlOutput);
+            } else if (type == MLAgentType.CONVERSATIONAL) {
+                Map<String, String> conversationalResult = extractConversationalAgentResult(mlOutput, xContentRegistry);
+                result = conversationalResult.get("dsl_query");
+            } else {
+                listener.onFailure(new IllegalArgumentException("Unsupported agent type: " + agentType));
+                return;
```
rithinpullela: Should we fail early for this case before executing the agent? Because other agents like Conversational Flow and PER might fail given these parameters.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -460,45 +472,258 @@ public void inferenceSentenceHighlighting(
...
     /**
-     * Execute agent with provided parameters and return DSL query string.
-     *
-     * @param agentId    the agent ID to execute
-     * @param parameters the parameters to pass to the agent
-     * @param listener   the listener to be called with the DSL query result
+     * Get agent type from agent ID
+     */
+    public void getAgentType(@NonNull String agentId, @NonNull ActionListener<Map<String, Object>> listener) {
+        retryableGetAgentType(agentId, 0, listener);
```
vibrantvarun: Why retry time is 0?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -460,45 +472,258 @@ public void inferenceSentenceHighlighting(
...
-    ) {
-        retryableExecuteAgent(agentId, parameters, 0, listener);
+        @NonNull SearchRequest request,
+        @NonNull AgenticSearchQueryBuilder agenticQuery,
+        @NonNull String agentId,
+        @NonNull Map<String, Object> agentInfo,
+        @NonNull NamedXContentRegistry xContentRegistry,
+        @NonNull ActionListener<String> listener
+    ) throws IOException {
+        retryableExecuteAgent(request, agenticQuery, agentId, agentInfo, xContentRegistry, 0, listener);
```
vibrantvarun, line 525: Why retry time is 0?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -460,45 +472,258 @@ public void inferenceSentenceHighlighting(
...
+
+            result.put("hasSystemPrompt", hasSystemPrompt);
+            result.put("hasUserPrompt", hasUserPrompt);
+
+            listener.onResponse(result);
+        }, e -> RetryUtil.handleRetryOrFailure(e, retryTime, () -> retryableGetAgentType(agentId, retryTime + 1, listener), listener)));
+    }
+
+    /**
+     * Execute agent with automatic detection of agent type
```
vibrantvarun, line 509: As it is a public method. Please add params in java doc.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -460,45 +472,258 @@ public void inferenceSentenceHighlighting(
...
+                        continue;
+                    }
+
+                    try (
+                        XContentParser parser = XContentType.JSON.xContent()
+                            .createParser(xContentRegistry, null, new BytesArray(responseJsonString).streamInput())
+                    ) {
+
+                        if (parser.currentToken() == null) {
+                            parser.nextToken();
```
vibrantvarun, line 685: nit: This method looks to big. Can we break it down in smaller methods?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -460,45 +472,258 @@ public void inferenceSentenceHighlighting(
...
+     * Execute get agent with retry logic
+     */
+    private void retryableGetAgentType(String agentId, int retryTime, ActionListener<Map<String, Object>> listener) {
+        mlClient.getAgent(agentId, ActionListener.wrap(mlAgent -> {
+            if (mlAgent == null) {
+                listener.onFailure(new IllegalStateException("Agent not found"));
+                return;
+            }
+
+            Map<String, Object> result = new HashMap<>();
```
bzhangam: Can we build a DTO for the result since we know what's the possible fields we can have in the result.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -460,45 +472,258 @@ public void inferenceSentenceHighlighting(
...
+            return;
+        }
+
+        Map<String, String> parameters = new HashMap<>();
+        parameters.put("question", agenticQuery.getQueryText());
+
+        // Add index names if present
+        String[] indices = request.indices();
+        if (indices != null && indices.length > 0) {
+            Object indexValue = (type == MLAgentType.FLOW) ? indices[0] : indices;
```
bzhangam: What if we have more than one index for Flow?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -460,45 +472,258 @@ public void inferenceSentenceHighlighting(
...
+            parameters.put("index_name", indexValue.toString());
```
bzhangam: What will the string be like when we use indices? Won't the default toString function return the reference? Should we use `Arrays.toString(indices)`?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -460,45 +472,258 @@ public void inferenceSentenceHighlighting(
...
+                            parser.nextToken();
+                        }
+
+                        if (parser.currentToken() != XContentParser.Token.START_OBJECT) {
+                            throw new IllegalStateException("Expected START_OBJECT in response, but got: " + parser.currentToken());
+                        }
+
+                        Map<String, Object> responseMap = parser.map();
+                        Object dslQueryObj = responseMap.get("dsl_query");
+                        Object stepsSummaryObj = responseMap.get("agent_steps_summary");
```
bzhangam, line 694: We are not using this field `agent_steps_summary ` do we need to extract it? If we do can we build a DTO to hold the data?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
owaiskazi19: Good call

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
owaiskazi19: because increasing it to +1 [here](https://github.com/opensearch-project/neural-search/pull/1579/files#diff-5db7b5fe327a13e217e83a2669b98db04da9a799aa6e22d3e885365d86fd576eR594)

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
owaiskazi19: same as above

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
owaiskazi19: Good call. Cleaner approach

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
owaiskazi19: Our default agent for agentic search is conversational. We thought of removing support for flow agent but since we released experimental feature with flow agent so keeping it for just 1 index since it uses just QueryPlanningTool. For conversational agent it uses ListIndexTool for selecting the right set of indices or from here if provided by the user.
We would cover this in doc that flow agent would work only for 1 index

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
owaiskazi19: Yes, realized after I raised the PR. Pushed the change

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
owaiskazi19: Not in this PR but will be using agent_steps_summary in the next PR with an ext builder to show summary of the agent tracing and will create DTO for it as well

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
bzhangam: Should we throw an exception if we do run into multiple indices? Otherwise we silently ignore the other indices which users may think it as an error.

owaiskazi19:
BWC tests failing are unrelated
```
 ERROR][o.o.m.a.r.TransportRegisterModelAction] [neuralSearchBwcCluster-restart-2] Failed to register model
»  org.opensearch.transport.RemoteTransportException: [neuralSearchBwcCluster-restart-0][127.0.0.1:34257][cluster:admin/opensearch/mlinternal/forward]
»  Caused by: org.opensearch.core.common.io.stream.NotSerializableExceptionWrapper: m_l_limit_exceeded_exception: Disk Circuit Breaker is open, please check your resources!
»  	at org.opensearch.ml.utils.MLNodeUtils.checkOpenCircuitBreaker(MLNodeUtils.java:63) ~[?:?]
»  	at org.opensearch.ml.model.MLModelManager.checkAndAddRunningTask(MLModelManager.java:829) ~[?:?]
```

vibrantvarun (approved):
Looks good to me.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
owaiskazi19: Yeah we can. Pushed the change and a test

bzhangam (approved):

PR #1580, opened 2025-09-24
Seismic: Implement Dynamic Max Float Value For Byte Quantization

yuye-aws:
### Description

This PR is an enhancement on the Seismic byte quantization feature. It allows the Seismic user to configure the max float value to consider during ingestion and query. Here is the current quantization function which linearly quantizes a float value into byte value range.

```
// The maximum float value to consider
private static final float MAX_FLOAT_VALUE = 3.0f;

public static byte quantizeFloatToByte(float value) {
    // Ensure the value is within the specified range
    value = Math.max(0.0f, Math.min(MAX_FLOAT_VALUE, value));
    
    // Scale the value to fit in the byte range (0-255)
    // Note: In Java, byte is signed (-128 to 127), but we'll use the full precision
    value = (value * MAX_UNSIGNED_BYTE_VALUE) / MAX_FLOAT_VALUE;
    
    // Round to nearest integer and cast to byte
    return (byte) Math.round(value);
}
```

In practice, the sparse embeddings can be generated from user's own sparse model or inverse document frequency value in doc-only model. **We may not assume that most float value falls within the range of [0, 3].** This PR introduces two properties within the index mapping named: **quantization_ceiling_search** and **quantization_ceiling_ingest**. This is an example index mapping:

```
{
  "settings": {
    "index": {
      "sparse": "true"
    }
  },
  "mappings": {
    "properties": {
      "sparse_embedding": {
        "type": "sparse_vector",
        "method": {
          "name": "seismic",
          "parameters": {
            "quantization_ceiling_search": 16,
            "quantization_ceiling_ingest": 16,
            "n_postings": 160,
            "cluster_ratio": 0.075,
            "summary_prune_ratio": 0.4,
            "approximate_threshold": 100000
          }
        }
      },
      "id": {
        "type": "keyword"
      }
    }
  }
}
```

### Related Issues

### Check List


src/main/java/org/opensearch/neuralsearch/sparse/algorithm/seismic/Seismic.java
```diff
@@ -103,6 +105,44 @@ public ValidationException validateMethod(SparseMethodContext sparseMethodContex
             }
             parameters.remove(APPROXIMATE_THRESHOLD_FIELD);
         }
+        if (parameters.containsKey(QUANTIZATION_CEILING_INGEST_FIELD)) {
+            try {
+                String fieldValueString = parameters.get(QUANTIZATION_CEILING_INGEST_FIELD).toString();
```
model-collapse, line 110: Why we have to convert string to int/float everytime? Why can we store the configure in int?

model-collapse (changes requested):

src/main/java/org/opensearch/neuralsearch/sparse/algorithm/seismic/Seismic.java
yuye-aws: It's because the MethodComponentContext store the parameters like this: Map<String, Object> parameters. This method validates the parameters in map. We introduce this map because there maybe unnecessary parameters specified by the user.

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseDocValuesProducer.java
```diff
@@ -48,7 +50,8 @@ public NumericDocValues getNumeric(FieldInfo field) throws IOException {
      */
     @Override
     public BinaryDocValues getBinary(FieldInfo field) throws IOException {
-        return new SparseBinaryDocValuesPassThrough(this.delegate.getBinary(field), this.getState().segmentInfo);
+        ByteQuantizer byteQuantizer = ByteQuantizerUtil.getByteQuantizerIngest(field);
```
model-collapse: Still need to check the call frequency.

src/main/java/org/opensearch/neuralsearch/sparse/query/SparseQueryWeight.java
```diff
@@ -51,10 +53,14 @@ public SparseQueryWeight(
         IndexSearcher searcher,
         ScoreMode scoreMode,
         float boost,
+        float quantizationCeilSearch,
+        float quantizationCeilIngest,
         ForwardIndexCache forwardIndexCache
     ) throws IOException {
         super(query);
         this.boost = boost;
+        this.quantizationCeilSearch = quantizationCeilSearch;
```
model-collapse: Need to clarify why these parameter is not implemented in scorer.

yuye-aws: The parameter **quantizationCeilSearch** and **quantizationCeilIngest** are introduced in the weight object to construct a **SimScorer** object. When creating the scorer object ( **OrderedPostingWithClustersScorer** and **ExactMatchScorer**), it passes in the **SimScorer** object to scorer in the **selectScorer()** function.

It does not matter whether the SimScorer is instantiated in weight or scorer. They just work equivalently.

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
```diff
@@ -554,10 +554,10 @@ Map<String, Float> getQueryTokens(QueryShardContext context) {
 
     @Override
     protected Query doToQuery(QueryShardContext context) throws IOException {
-        final MappedFieldType ft = context.fieldMapper(fieldName);
-        boolean isSeismic = isSeismicFieldType(ft);
+        final MappedFieldType fieldType = context.fieldMapper(fieldName);
```
chishui: Nitpick (non-blocking): this is a large PR, better to avoid making unnecessary changes like this

src/main/java/org/opensearch/neuralsearch/sparse/quantization/ByteQuantizerUtil.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.lucene.index.FieldInfo;
+import org.apache.lucene.search.similarities.Similarity;
+
+import static org.opensearch.neuralsearch.sparse.common.SparseConstants.QUANTIZATION_CEILING_INGEST_FIELD;
+import static org.opensearch.neuralsearch.sparse.common.SparseConstants.Seismic.DEFAULT_QUANTIZATION_CEILING_INGEST;
+
+/**
+ * Utility class for float -> byte quantization
+ */
+public final class ByteQuantizerUtil {
```
chishui: ByteQuantizationUtil

src/main/java/org/opensearch/neuralsearch/sparse/algorithm/seismic/Seismic.java
```diff
@@ -103,6 +105,44 @@ public ValidationException validateMethod(SparseMethodContext sparseMethodContex
...
+                float quantizationCeilValue = NumberUtils.createFloat(fieldValueString);
+                if (quantizationCeilValue <= 0) {
+                    errorMessages.add(String.format(Locale.ROOT, "Parameter [%s] must be positive", QUANTIZATION_CEILING_INGEST_FIELD));
```
chishui: "must be positive float number"

src/main/java/org/opensearch/neuralsearch/sparse/algorithm/seismic/Seismic.java
```diff
@@ -103,6 +105,44 @@ public ValidationException validateMethod(SparseMethodContext sparseMethodContex
...
+                );
+            }
+            parameters.remove(QUANTIZATION_CEILING_INGEST_FIELD);
+        }
+        if (parameters.containsKey(QUANTIZATION_CEILING_SEARCH_FIELD)) {
+            try {
+                String fieldValueString = parameters.get(QUANTIZATION_CEILING_SEARCH_FIELD).toString();
+                float quantizationCeilValue = NumberUtils.createFloat(fieldValueString);
+                if (quantizationCeilValue <= 0) {
+                    errorMessages.add(String.format(Locale.ROOT, "Parameter [%s] must be positive", QUANTIZATION_CEILING_SEARCH_FIELD));
```
chishui: same here

src/main/java/org/opensearch/neuralsearch/sparse/codec/ClusteredPostingTermsWriter.java
```diff
@@ -71,6 +72,13 @@ public class ClusteredPostingTermsWriter extends PushPostingsWriterBase {
     private SegmentWriteState state;
     private DocValuesProducer docValuesProducer;
     private final CodecUtilWrapper codecUtilWrapper;
+    private ByteQuantizer byteQuantizer;
+
+    @Override
+    public void setField(FieldInfo fieldInfo) {
+        super.setField(fieldInfo);
+        byteQuantizer = ByteQuantizerUtil.getByteQuantizerIngest(fieldInfo);
```
chishui: what's the point of doing this? Can you pass byteQuantizer through constructor?

src/main/java/org/opensearch/neuralsearch/sparse/quantization/ByteQuantizerUtil.java
```diff
@@ -0,0 +1,95 @@
...
+
+    private ByteQuantizerUtil() {} // no instance of this utility class
+
+    // Use full precision of byte (0-255)
+    public static final float MAX_UNSIGNED_BYTE_VALUE = 255.0f;
```
chishui: use int 255 to represent max byte value

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseDocValuesProducer.java
yuye-aws: Nice comment! The getBinary function happens during per (field * segment). It won't happen for each sparse vector read.

src/main/java/org/opensearch/neuralsearch/sparse/quantization/ByteQuantizer.java
```diff
@@ -4,92 +4,32 @@
...
-     * This represents the ratio between the square of the maximum float value (3.0²) and
-     * the square of the maximum byte value (255²), which is the scaling factor needed for
-     * dot products of quantized vectors.
-     */
-    public static final float SCORE_RESCALE_RATIO = MAX_FLOAT_VALUE * MAX_FLOAT_VALUE / MAX_UNSIGNED_BYTE_VALUE / MAX_UNSIGNED_BYTE_VALUE;
+    private final float ceilValue;
 
-    private ByteQuantizer() {} // no instance of this utility class
+    public ByteQuantizer(float ceilValue) {
+        if (ceilValue < 0) {
```
chishui: `<=0`? positive value

src/main/java/org/opensearch/neuralsearch/sparse/quantization/ByteQuantizerUtil.java
```diff
@@ -0,0 +1,95 @@
...
+     * Get a byte quantizer object during ingestion, the max float value is parsed from MAX_FLOAT_VALUE_INGEST_FIELD
+     * @param fieldInfo {@link FieldInfo}
+     * @return {@link ByteQuantizer}
+     */
+    public static ByteQuantizer getByteQuantizerIngest(FieldInfo fieldInfo) {
+        if (fieldInfo == null) {
+            return new ByteQuantizer(DEFAULT_QUANTIZATION_CEILING_INGEST);
+        }
+        String stringValue = fieldInfo.getAttribute(QUANTIZATION_CEILING_INGEST_FIELD);
+        float quantizationCeilIngest = stringValue == null ? DEFAULT_QUANTIZATION_CEILING_INGEST : NumberUtils.createFloat(stringValue);
```
chishui: Use `StringUtils.isEmpty()` to check both null and empty string

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseBinaryDocValuesPassThrough.java
```diff
@@ -27,6 +28,9 @@ public class SparseBinaryDocValuesPassThrough extends BinaryDocValues implements
     @Getter
     private final SegmentInfo segmentInfo;
 
+    /** The byte quantizer instance */
```
chishui: nit: single line comment use`//`

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseDocValuesProducer.java
```diff
@@ -48,7 +54,11 @@ public NumericDocValues getNumeric(FieldInfo field) throws IOException {
      */
     @Override
     public BinaryDocValues getBinary(FieldInfo field) throws IOException {
-        return new SparseBinaryDocValuesPassThrough(this.delegate.getBinary(field), this.getState().segmentInfo);
+        if (!byteQuantizerMap.containsKey(field)) {
```
chishui: I feel the overall solution is not very clean, I'd prefer to initialize the quantizer close to where it's been used and we don't need to store or passing it around as long as the class has access to field. Even it has not access to field, it's better to pass field instead of passing a quantizer.

src/main/java/org/opensearch/neuralsearch/sparse/query/SparseAnnQueryBuilder.java
```diff
@@ -232,11 +240,24 @@ private SparseQueryContext constructSparseQueryContext() {
...
     @Override
     public Query doToQuery(QueryShardContext context) throws IOException {
-        final MappedFieldType ft = context.fieldMapper(fieldName);
-        validateFieldType(ft);
+        final MappedFieldType fieldType = context.fieldMapper(fieldName);
+        validateFieldType(fieldType);
 
         SparseQueryContext sparseQueryContext = constructSparseQueryContext();
 
+        float quantizationCeilSearch = DEFAULT_QUANTIZATION_CEILING_SEARCH;
```
chishui: can't you get them from fieldInfo attributes, after you set them to fieldinfo attributes, fieldinfo attributes should be the single source of truth.

src/main/java/org/opensearch/neuralsearch/sparse/query/SparseAnnQueryBuilder.java
```diff
@@ -248,9 +269,11 @@ public Query doToQuery(QueryShardContext context) throws IOException {
         }
         return new SparseVectorQuery.SparseVectorQueryBuilder().fieldName(fieldName)
             .queryContext(sparseQueryContext)
-            .queryVector(new SparseVector(integerTokens))
+            .queryVector(new SparseVector(integerTokens, new ByteQuantizer(quantizationCeilSearch)))
             .fallbackQuery(fallbackQuery)
             .filter(filterQuery)
+            .quantizationCeilSearch(quantizationCeilSearch)
```
chishui: can you not passing them but get them when it's used

chishui (commented):
I think this PR needs a refactor to avoid touching too many classes which leads to high risk

src/main/java/org/opensearch/neuralsearch/query/NeuralSparseQueryBuilder.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/sparse/quantization/ByteQuantizerUtil.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/sparse/algorithm/seismic/Seismic.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/sparse/algorithm/seismic/Seismic.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/sparse/codec/ClusteredPostingTermsWriter.java
yuye-aws: The point is, clustered posting terms writer may interact with multiple fields. The byte quantizer will change. Ie still need the byte quantizer in startDoc function. I can change it to the map implementation

src/main/java/org/opensearch/neuralsearch/sparse/quantization/ByteQuantizerUtil.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/sparse/quantization/ByteQuantizer.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/sparse/quantization/ByteQuantizerUtil.java
yuye-aws: Ack

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseBinaryDocValuesPassThrough.java
yuye-aws: Ack. Will also change comment for delegate and segmentInfo

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseDocValuesProducer.java
yuye-aws: Let's have a quick discussion

src/main/java/org/opensearch/neuralsearch/sparse/query/SparseAnnQueryBuilder.java
yuye-aws: Here, I get the value from methodComponentContext. What do you think is the difference

src/main/java/org/opensearch/neuralsearch/sparse/query/SparseAnnQueryBuilder.java
yuye-aws: Not sure whether you are referring to quantizationCeilSearch or the byte quantizer. Let's have a quick call

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseDocValuesProducer.java
yuye-aws: Pass in fieldinfo into SparseBinaryDocValuesPassThrough

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseDocValuesProducer.java
yuye-aws: Per (field * segment) during merge

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseDocValuesProducer.java
yuye-aws: Need to check the initialization of quantizer

src/main/java/org/opensearch/neuralsearch/sparse/query/SparseAnnQueryBuilder.java
yuye-aws: Need to check whether I can get a field info object

src/main/java/org/opensearch/neuralsearch/sparse/query/SparseAnnQueryBuilder.java
yuye-aws: If we can get a field info object, then pass it into the weight object

src/main/java/org/opensearch/neuralsearch/sparse/codec/ClusteredPostingTermsWriter.java
yuye-aws: Still need the byte quantizer in startDoc function

src/main/java/org/opensearch/neuralsearch/sparse/codec/SparseDocValuesProducer.java
yuye-aws: Changed the implementation to passing the field info into the SparseBinaryDocValuesPassThrough class. Besides, I verified that all initialized quantizer is close to usage.

src/main/java/org/opensearch/neuralsearch/sparse/query/SparseAnnQueryBuilder.java
yuye-aws: Implemented with field info implementation. I added a try catch block to deal with NPE

src/main/java/org/opensearch/neuralsearch/sparse/query/SparseQueryWeight.java
yuye-aws: By @chishui 's comment. I just need to update the boost for the scorer

src/main/java/org/opensearch/neuralsearch/sparse/common/SparseConstants.java
```diff
@@ -29,5 +31,7 @@ public static final class Seismic {
         public static final int DEFAULT_APPROXIMATE_THRESHOLD = 1000000;
         public static final float DEFAULT_POSTING_PRUNE_RATIO = 0.0005f;
         public static final int DEFAULT_POSTING_MINIMUM_LENGTH = 160;
+        public static final float DEFAULT_QUANTIZATION_CEILING_INGEST = 3.0f;
```
chishui, line 34: Any reason we use different default values for ingest and query?

src/main/java/org/opensearch/neuralsearch/sparse/codec/ClusteredPostingTermsWriter.java
```diff
@@ -185,7 +189,11 @@ public void startDoc(int docID, int freq) throws IOException {
         if (docID == -1) {
             throw new IllegalStateException("docId must be set before startDoc");
         }
-        docWeights.add(new DocWeight(docID, ByteQuantizer.quantizeFloatToByte(ValueEncoder.decodeFeatureValue(freq))));
+        if (!byteQuantizerMap.containsKey(fieldInfo)) {
```
model-collapse: Let's don't search map in startDoc, create the quantizer each time we set a field.

src/main/java/org/opensearch/neuralsearch/sparse/query/SparseAnnQueryBuilder.java
```diff
@@ -232,11 +241,26 @@ private SparseQueryContext constructSparseQueryContext() {
...
-        final MappedFieldType ft = context.fieldMapper(fieldName);
-        validateFieldType(ft);
+        final MappedFieldType fieldType = context.fieldMapper(fieldName);
+        validateFieldType(fieldType);
 
         SparseQueryContext sparseQueryContext = constructSparseQueryContext();
 
+        // different field infos should have the same value for quantization ceiling search
+        float quantizationCeilSearch = DEFAULT_QUANTIZATION_CEILING_SEARCH;
+        try {
```
chishui: nit: move a function?

chishui: Thought (non-blocking): I think we can refactor SparseVector to be able to do a lazy quantization, so that we can set quantizer in Weight not in query builder

src/main/java/org/opensearch/neuralsearch/sparse/common/SparseConstants.java
yuye-aws: Yes. For ingestion, the sparse weights are generated via model. Most of them does not exceed 3.0. For search, there are both bi-encoder and doc-only model. It would be safer to set a larger value, namely 16.0 as the default value

src/main/java/org/opensearch/neuralsearch/sparse/query/SparseAnnQueryBuilder.java
yuye-aws: > I think we can refactor SparseVector to be able to do a lazy quantization

Without immediate quantization. There would be more JVM memory consumption because sparse vector needs to store the floats instead of bytes. Have you found any quick solution?

yuye-aws: > nit: move a function?

Ack

src/main/java/org/opensearch/neuralsearch/sparse/query/SparseAnnQueryBuilder.java
```diff
@@ -317,4 +329,21 @@ private static Map<String, Float> preprocessQueryTokens(Map<String, Float> token
...
+            for (LeafReaderContext leafContext : context.searcher().getTopReaderContext().leaves()) {
+                FieldInfos fieldInfos = leafContext.reader().getFieldInfos();
+                FieldInfo fieldInfo = fieldInfos.fieldInfo(fieldName);
+                if (fieldInfo != null) {
+                    quantizationCeilSearch = ByteQuantizationUtil.getCeilingValueSearch(fieldInfo);
+                    break;
+                }
+            }
+        } catch (Exception e) {
+            log.error(String.format(Locale.ROOT, "Failed to get quantization ceiling search value for field [%s]", fieldName), e);
```
chishui: nit: `log.error("Failed to get quantization ceiling search value for field [{}]", fieldName, e);`

src/main/java/org/opensearch/neuralsearch/sparse/query/SparseQueryWeight.java
```diff
@@ -114,12 +116,17 @@ public long cost() {
     @VisibleForTesting
     Scorer selectScorer(SparseVectorQuery query, LeafReaderContext context, SegmentInfo segmentInfo) throws IOException {
         SparseVectorReader cacheGatedForwardIndexReader = SparseVectorReader.NOOP_READER;
+        FieldInfo fieldInfo = context.reader().getFieldInfos().fieldInfo(query.getFieldName());
+        float rescaledBoost = boost * ByteQuantizationUtil.getCeilingValueIngest(fieldInfo) * ByteQuantizationUtil.getCeilingValueSearch(
+            fieldInfo
```
chishui, line 121: break line from `ByteQuantizationUtil.getCeilingValueSearch`?

src/main/java/org/opensearch/neuralsearch/sparse/query/SparseAnnQueryBuilder.java
yuye-aws: This is the log4j guide: https://logging.apache.org/log4j/2.x/manual/api.html#best-practice-exception

src/main/java/org/opensearch/neuralsearch/sparse/codec/ClusteredPostingTermsWriter.java
yuye-aws: Updated

src/main/java/org/opensearch/neuralsearch/sparse/query/SparseQueryWeight.java
yuye-aws: This is mandated by spotless apply

src/main/java/org/opensearch/neuralsearch/sparse/query/SparseAnnQueryBuilder.java
chishui: I was saying you don't need to use a String.format within a log

model-collapse (approved):

src/main/java/org/opensearch/neuralsearch/sparse/query/SparseAnnQueryBuilder.java
yuye-aws: Updated

yuye-aws:
Thanks this **[PR](https://github.com/opensearch-project/neural-search/pull/1584)** from @chishui to fix the BWC test. The BWC CI is able to pass now

Issue #1582, opened 2025-09-24
[BUG] Normalization Processor's BWC tests are failing


owaiskazi19:
### What is the bug?
```
HybridSearchIT > testNormalizationProcessor_whenIndexWithMultipleShards_E2EFlow FAILED
    org.opensearch.client.ResponseException: method [PUT], host [http://127.0.0.1:42429/], URI [/neuralsearch-bwc-testnormalizationprocessor_whenindexwithmultipleshards_e2eflow/_doc/0?refresh=true], status line [HTTP/1.1 403 Forbidden]
    {"error":{"root_cause":[{"type":"status_exception","reason":"Failed to Validate Access for ModelId VTySfZkBB0UbS6zCKcKi"}],"type":"status_exception","reason":"Failed to Validate Access for ModelId VTySfZkBB0UbS6zCKcKi"},"status":403}
        at __randomizedtesting.SeedInfo.seed([EC7D0DDCF3C7409E:45CE3FD286EB857A]:0)
        at app//org.opensearch.client.RestClient.convertResponse(RestClient.java:501)
        at app//org.opensearch.client.RestClient.performRequest(RestClient.java:384)
        at app//org.opensearch.client.RestClient.performRequest(RestClient.java:359)
        at app//org.opensearch.neuralsearch.BaseNeuralSearchIT.addDocument(BaseNeuralSearchIT.java:2120)
        at app//org.opensearch.neuralsearch.bwc.restart.HybridSearchIT.addDocuments(HybridSearchIT.java:93)
        at app//org.opensearch.neuralsearch.bwc.restart.HybridSearchIT.validateNormalizationProcessor(HybridSearchIT.java:67)
        at app//org.opensearch.neuralsearch.bwc.restart.HybridSearchIT.testNormalizationProcessor_whenIndexWithMultipleShards_E2EFlow(HybridSearchIT.java:46)
```

```
 » WARN ][a.d.u.c.CudaUtils        ] [neuralSearchBwcCluster-restart-0] Access denied during loading cudart library.
» ERROR][o.o.m.a.p.TransportPredictionTaskAction] [neuralSearchBwcCluster-restart-0] Failed to Validate Access for ModelId VTySfZkBB0UbS6zCKcKi
»  org.opensearch.ml.common.exception.MLLimitExceededException: Disk Circuit Breaker is open, please check your resources!
[...]

PR #1504, opened 2025-08-08
Disable ML-Commons disk circuit breaker in tests

chishui:
### Description
Many tests fails as disk circuit breaker is triggered in ML Commons. Disable this setting.

### Related Issues
Resolves #1494

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/neural-search/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1583, opened 2025-09-24
Disable ml-commons circuit breaker in bwc tests

owaiskazi19:
### Description
Ref: https://github.com/opensearch-project/neural-search/pull/1504

### Related Issues
Resolves https://github.com/opensearch-project/neural-search/issues/1582

### Check List

### DIFF
--- a/qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/AbstractRestartUpgradeRestTestCase.java
+++ b/qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/AbstractRestartUpgradeRestTestCase.java
@@ -8,8 +8,8 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Locale;
 import java.util.Optional;
-
 import org.junit.Before;
+
 import org.opensearch.common.settings.Settings;
 import static org.opensearch.neuralsearch.util.TestUtils.NEURAL_SEARCH_BWC_PREFIX;
 import static org.opensearch.neuralsearch.util.TestUtils.CLIENT_TIMEOUT_VALUE;
@@ -23,11 +23,6 @@ import org.opensearch.test.rest.OpenSearchRestTestCase;
 public abstract class AbstractRestartUpgradeRestTestCase extends BaseUpgradeTestCase {
 
     @Before
-    public void setUp() throws Exception {
-        super.setUp();
-        tryUpdateClusterSettings("plugins.ml_commons.disk_free_space_threshold", -1);
-    }
-
     protected String getIndexNameForTest() {
         // Creating index name by concatenating "neural-bwc-" prefix with test method name
         // for all the tests in this sub-project
@@ -36,12 +31,14 @@ public abstract class AbstractRestartUpgradeRestTestCase extends BaseUpgradeTest
 
     @Override
     protected final Settings restClientSettings() {
+        tryUpdateClusterSettings("plugins.ml_commons.disk_free_space_threshold", -1);
         return Settings.builder()
             .put(super.restClientSettings())
             // increase the timeout here to 90 seconds to handle long waits for a green
             // cluster health. the waits for green need to be longer than a minute to
             // account for delayed shards
             .put(OpenSearchRestTestCase.CLIENT_SOCKET_TIMEOUT, CLIENT_TIMEOUT_VALUE)
+            .put("plugins.ml_commons.disk_free_space_threshold", -1)
             .build();
     }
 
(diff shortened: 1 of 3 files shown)


vibrantvarun:
Looks good to me. Will wait for @chishui approval once.

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/AbstractRestartUpgradeRestTestCase.java
```diff
@@ -30,12 +31,14 @@ protected String getIndexNameForTest() {
 
     @Override
     protected final Settings restClientSettings() {
+        tryUpdateClusterSettings("plugins.ml_commons.disk_free_space_threshold", -1);
```
chishui, line 34: Do we need this setting in `AbstractRollingUpgradeTestCase` as well?

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/AbstractRestartUpgradeRestTestCase.java
```diff
@@ -36,6 +37,7 @@ protected final Settings restClientSettings() {
             // cluster health. the waits for green need to be longer than a minute to
             // account for delayed shards
             .put(OpenSearchRestTestCase.CLIENT_SOCKET_TIMEOUT, CLIENT_TIMEOUT_VALUE)
+            .put("plugins.ml_commons.disk_free_space_threshold", -1)
```
chishui, line 41: Issue: this setting is supported since version 2.16, for lower versions, an exception could throw

owaiskazi19: A better way would be to just call `tryUpdateClusterSettings("plugins.ml_commons.disk_free_space_threshold", -1);` here?

yuye-aws (approved):

qa/restart-upgrade/src/test/java/org/opensearch/neuralsearch/bwc/restart/AbstractRestartUpgradeRestTestCase.java
owaiskazi19: @chishui tests are still failing with `tryUpdateClusterSettings`

chishui: Looks like the update setting request is only sent to version 2.9 but not 3.2, could you try to move the call to `@Before` function?

owaiskazi19: Figured out 
`restClientSettings()` is client-side only. these settings are applied to the REST test client, not cluster settings.
OS might ignore unrecognized settings on the client side silently. So even if the node doesn't recognize this setting (on <2.16), it likely doesn't complain since it's not enforced in client-side logic.

chishui: looks like disk circuit breaker was introduced in [2.4](https://github.com/opensearch-project/ml-commons/blob/266bcfe75da0f5d716d42d07d74358102065a50a/release-notes/opensearch-ml-common.release-notes-2.4.0.0.md?plain=1#L32) , and making it a cluster setting was since 2.16, so for versions before 2.16, it doesn't seem to be adjustable, maybe we need to check why the test used all disk space.

chishui: ml-commons' [CB checks](https://github.com/opensearch-project/ml-commons/blob/2.9/plugin/src/main/java/org/opensearch/ml/breaker/DiskCircuitBreaker.java) if there are 5GB free disk left, but there are only 13G left at the beginning of `NeuralSearch Restart-Upgrade BWC Tests` of CI. If we can increase the disk size of CI node, that'll be a quick fix, but I'm not sure if it's even feasible. Another option is try to delete files/libraries which won't be used by BWC tests to reduce used disk size (59G). 

```bash
Run df -h
Filesystem      Size  Used Avail Use% Mounted on
overlay          72G   59G   13G  82% /
tmpfs            64M     0   64M   0% /dev
shm              64M     0   64M   0% /dev/shm
/dev/root        72G   59G   13G  82% /node
tmpfs           3.2G  1.2M  3.2G   1% /run/docker.sock
tmpfs           7.9G     0  7.9G   0% /proc/acpi
tmpfs           7.9G     0  7.9G   0% /proc/scsi
tmpfs           7.9G     0  7.9G   0% /sys/firmware
```

chishui: Create #1584 to increase CI node disk size.

yuye-aws:
Closing this PR in favor of https://github.com/opensearch-project/neural-search/pull/1584

PR #1584, opened 2025-09-25
Add a CI action to clean up disk and apply to BWC

chishui:
### Description
ml-commons' [CB checks](https://github.com/opensearch-project/ml-commons/blob/2.9/plugin/src/main/java/org/opensearch/ml/breaker/DiskCircuitBreaker.java) if there are 5GB free disk left, but there are only 13G left at the beginning of `NeuralSearch Restart-Upgrade BWC Tests` of CI. 

Add a CI action to clean up disk and apply to BWC which increases 4-5 GB disk size. It removes:
1. CodeQL
2. Go // configurable, with default true
3. Python // configurable, with default true
4. libgcc // configurable, with default true

```bash
Filesystem      Size  Used Avail Use% Mounted on
overlay          72G   60G   12G  84% /
tmpfs            64M     0   64M   0% /dev
shm              64M     0   64M   0% /dev/shm
/dev/root        72G   60G   12G  84% /__t
tmpfs           3.2G  1.2M  3.2G   1% /run/docker.sock
tmpfs           7.9G     0  7.9G   0% /proc/acpi
tmpfs           7.9G     0  7.9G   0% /proc/scsi
tmpfs           7.9G     0  7.9G   0% /sys/firmware

after run:

Run df -h
Filesystem      Size  Used Avail Use% Mounted on
overlay          72G   57G   16G  79% /
tmpfs            64M     0   64M   0% /dev
shm              64M     0   64M   0% /dev/shm
/dev/root        72G   57G   16G  79% /__t
tmpfs           3.2G  1.2M  3.2G   1% /run/docker.sock
tmpfs           7.9G     0  7.9G   0% /proc/acpi
tmpfs           7.9G     0  7.9G   0% /proc/scsi
tmpfs           7.9G     0  7.9G   0% /sys/firmware
```

### Related Issues
Resolves #1583 #1582

### Check List

### DIFF
--- a/src/test/java/org/opensearch/neuralsearch/query/NeuralQueryBuilderTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/query/NeuralQueryBuilderTests.java
@@ -1082,8 +1082,7 @@ public class NeuralQueryBuilderTests extends OpenSearchTestCase {
         KNNQueryBuilder returnedKNNQueryBuilder = (KNNQueryBuilder) result;
         assertEquals("Field name should match", FIELD_NAME, returnedKNNQueryBuilder.fieldName());
         assertArrayEquals("Vector should match", testVector, (float[]) returnedKNNQueryBuilder.vector(), 0.0f);
-        assertEquals("K should match", neuralQueryBuilder.k().intValue(), returnedKNNQueryBuilder.getK());
-
+        assertEquals("K should match", neuralQueryBuilder.k().intValue(), (int) returnedKNNQueryBuilder.getK());
     }
 
     @SneakyThrows
@@ -1146,7 +1145,8 @@ public class NeuralQueryBuilderTests extends OpenSearchTestCase {
             DELTA_FOR_FLOATS_ASSERTION
         );
         // For radial search, KNNQueryBuilder defaults K to 0 when not specified
-        assertEquals("K should be 0 for radial search when not specified", 0, returnedKNNQueryBuilder.getK());
+        Integer k = returnedKNNQueryBuilder.getK();
+        assertEquals("K should be 0 for radial search when not specified", 0, k == null ? 0 : k.intValue());
     }
 
     public void testPrepareTwoPhase_Query_whenRawTokensAvailable() {


.github/workflows/backwards_compatibility_tests_workflow.yml
```diff
@@ -36,6 +36,10 @@ jobs:
       BWC_VERSION_RESTART_UPGRADE: ${{ matrix.bwc_version }}
 
     steps:
+      - name: "node-cleanup"
+        run: |
+          rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc /__t/CodeQL
```
yuye-aws: I checked the rolling upgrade. After test, the there is only 150M disk space left. Can you also add it to rolling upgrade? It also works for integration tests and multi-node integration tests

chishui: added rolling upgrade, disk size for integration tests and multi-node integration tests shouldn't be an issue. Although one multi-node integration tests still fails, haven't looked into it.

yuye-aws: I understand multi-node IT is not currently failing due to limited disk space. Similar to restart upgrade and rolling upgrade, it also uses 3 nodes for testing. That's why I'm afraid the disk space, in the future, will not be enough.

yuye-aws:
> https://github.com/marketplace/actions/runner-cleanup-action (it says can free almost 30 GB of disk space)

Nice finding! I think, in long term, it would be a better solution

chishui:
> https://github.com/marketplace?query=disk

I tried a couple of them, neither of them works out of the box.
- https://github.com/marketplace/actions/runner-cleanup-action try to remove java and exit due to using sudo which is not available.
- https://github.com/jlumbroso/free-disk-space fails to remove dotnet and claims to free 16GB disk size, but when run our test, the disk size still shows 13GB. gradlew check on windows fails.

There might be one that works, but it takes times and efforts to try them out one by one. I'd recommend we start from one we know that we don't need (CodeQL), then we can gradually add more unused libraries in that list without accidentally impact any logic.

owaiskazi19:
@chishui restart bwc tests are still failing
```
» ERROR][o.o.m.m.MLModelManager   ] [neuralSearchBwcCluster-restart-2] No controller is deployed because the model kM6UgZkBW71TGiEuzs6O is expected not having an enabled model controller. Please use the create model controller api to create one if this is unexpected.
```
can we rerun and verify if it is flaky?

owaiskazi19:
@chishui you can refer my PR to fix the failing tests [here](https://github.com/opensearch-project/neural-search/pull/1585/files#diff-8c5a2adb190ee2ccada7dab80c23580149227abfbf758383fe50635c54c4a95dR1149-R1150). We can try to get this merge to resolve the bwc failures

yuye-aws:
Can you update the PR title and description?

heemin32 (approved):

PR #1585, opened 2025-09-26
[Agentic Search] Add support for agent summary and memory id for conversational agent

owaiskazi19:
### Description
This PR adds support for agent summary and memory id for conversational agent.
 - Adds a SearchExtBuilder to show agent summary and memory_id
 ```
 "ext": {
        "agent_steps_summary": "I have these tools available: [ListIndexTool, IndexMappingTool, query_planner_tool]\nThird I used: query_planner_tool — qpt.question: \"Is there any in blue color?\"; index_name_provided: \"products-index\"\nValidation: qpt output is valid JSON and reflects the user request.",
        "memory_id": "2GMahJkBlF3USoHtuIpW"
    }
 ```
 - Adds a new search response processor for agent summary
 ```
 {
    "request_processors": [
        {
            "agentic_query_translator": {
                "agent_id": "HmRwfZkB60g4p301w7iv"
            }
        }
    ],
    "response_processors": [
        {
            "agentic_context": {
                "agent_steps_summary": true,
                "dsl_query": true
            }
        }
    ]
}
 ```
 
 - Adds memory_id to the query
 ```
 {
    "query": {
        "agentic": {
            "query_text": "Is there any in blue color",
            "memory_id": "2GMahJkBlF3USoHtuIpW"
        }
    }
}
 ```
 - Flipping the experimental flag

### Related Issues

### Check List


src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -583,15 +589,20 @@ private void retryableExecuteAgent(
...
+            String memoryId = null;
             if (type == MLAgentType.FLOW) {
-                result = extractFlowAgentResult(mlOutput);
+                dslQuery = extractFlowAgentResult(mlOutput);
             } else if (type == MLAgentType.CONVERSATIONAL) {
                 Map<String, String> conversationalResult = extractConversationalAgentResult(mlOutput, xContentRegistry);
-                result = conversationalResult.get("dsl_query");
+                dslQuery = conversationalResult.get("dsl_query");
+                agentStepsSummary = conversationalResult.get("agent_steps_summary");
+                memoryId = conversationalResult.get("memory_id");
```
bzhangam: Can we make them as constants since we use them in multiple places?

src/main/java/org/opensearch/neuralsearch/query/ext/AgentStepsSearchExtBuilder.java
```diff
@@ -0,0 +1,90 @@
...
+
+import lombok.Getter;
+
+/**
+ * SearchExtBuilder for agent steps summary in agentic search responses
+ */
+@AllArgsConstructor
+public class AgentStepsSearchExtBuilder extends SearchExtBuilder {
+
+    public final static String PARAM_FIELD_NAME = "agent_steps_summary";
```
bzhangam: Use a better name? `PARAM_FIELD_NAME ` is not descriptive.

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
```diff
@@ -103,14 +103,18 @@ private boolean hasOtherSearchFeatures(SearchSourceBuilder sourceBuilder) {
...
+        PipelineProcessingContext requestContext,
         ActionListener<SearchRequest> requestListener
     ) {
         // First get agent type and prompts info
         mlClient.getAgentDetails(agentId, ActionListener.wrap(agentInfo -> {
             mlClient.executeAgent(request, agenticQuery, agentId, agentInfo, xContentRegistry, ActionListener.wrap(agentResponse -> {
                 try {
+                    String dslQuery = agentResponse.getDslQuery();
+                    String agentStepsSummary = agentResponse.getAgentStepsSummary();
+                    String memoryId = agentResponse.getMemoryId();
```
bzhangam, line 116: If the memory id is passed in the query do we need to validate the memoryId in the agentResponse should match that one?

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
```diff
@@ -57,6 +58,7 @@ public final class AgenticSearchQueryBuilder extends AbstractQueryBuilder<Agenti
     private static final int MAX_QUERY_LENGTH = 1000;
     public String queryText;
     public List<String> queryFields;
+    public String memoryId;
```
bzhangam, line 61: Why we choose to pass the agentId through the request processor but pass the memoryId through the query?

src/main/java/org/opensearch/neuralsearch/query/ext/AgentStepsSearchExtBuilder.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        return (obj instanceof AgentStepsSearchExtBuilder)
+            && Objects.equals(agentStepsSummary, ((AgentStepsSearchExtBuilder) obj).agentStepsSummary)
+            && Objects.equals(memoryId, ((AgentStepsSearchExtBuilder) obj).memoryId);
+    }
+
+    public static AgentStepsSearchExtBuilder fromXContent(XContentParser parser) throws IOException {
```
bzhangam: I think we are not using this function. Would it be better to throw the unsupported exception here?

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
owaiskazi19: No because a different memory_id can also be passed. It need not to be the same returned

src/main/java/org/opensearch/neuralsearch/query/AgenticSearchQueryBuilder.java
owaiskazi19: agentId would be added just once after the setup to get agentic search ready but memory id is something a user can add to continue the conversation from previous context or not provide it to start a fresh context multiple time or can provide a new memory id to continue from a different context. It is not part of the setup but more of the query

src/main/java/org/opensearch/neuralsearch/query/ext/AgentStepsSearchExtBuilder.java
owaiskazi19: It is not an overriden method, just added it if needed for future and added tests for this method

owaiskazi19: Exception is thrown from the [initialization](https://github.com/opensearch-project/neural-search/pull/1585/files#diff-0b5f74541932d2881f876b8452af44942e80abd0c316662a1d5e2c44810e74caR363) itself

bzhangam: I think we don't want to parse the data from the search request. And I don't think in near future we will want to use it. So it's better to remove it to avoid having some unused codes.

src/main/java/org/opensearch/neuralsearch/processor/AgentStepsResponseProcessor.java
```diff
@@ -0,0 +1,111 @@
...
+    @Override
+    public SearchResponse processResponse(SearchRequest request, SearchResponse response, PipelineProcessingContext requestContext) {
+        if (requestContext == null) {
+            return response;
+        }
+
+        Object agentStepsSummary = requestContext.getAttribute("agent_steps_summary");
+        Object memoryId = requestContext.getAttribute("memory_id");
+
+        if (!(agentStepsSummary instanceof String) || ((String) agentStepsSummary).trim().isEmpty()) {
```
heemin32: Any reason why we don't add memory id when steps summary does not exist?

src/main/java/org/opensearch/neuralsearch/processor/AgentStepsResponseProcessor.java
```diff
@@ -0,0 +1,111 @@
...
+            return response;
+        }
+
+        String memoryIdStr = (memoryId instanceof String) ? (String) memoryId : null;
```
heemin32: Shouldn't we throw exception when memoryId is not string type? Also same for the steps summary.
```
if (agentSteps Summary instanceof String == false) {
  throw new RuntimeException("");
}

if (memoryId instanceof String == false) {
  throw new RuntimeException("");
}
```

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
heemin32: In which case does it return different memory id?

heemin32:
If user create agent steps response processors, both step summary and memory id will be returned in ext block? 
```
   "response_processors": [
       {
           "agent_steps": {}
       }
   ]
```
Is there a way for user to see only the memory id?
For memory id, they might want to know it always so that they can use it for their next request but not the agent steps.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -583,15 +592,20 @@ private void retryableExecuteAgent(
...
         mlClient.execute(FunctionName.AGENT, agentMLInput, ActionListener.wrap(response -> {
             MLOutput mlOutput = (MLOutput) response.getOutput();
-            String result = null;
+            String dslQuery = null;
+            String agentStepsSummary = null;
+
+            String memoryId = null;
             if (type == MLAgentType.FLOW) {
-                result = extractFlowAgentResult(mlOutput);
+                dslQuery = extractFlowAgentResult(mlOutput);
```
heemin32, line 605: We might want to throw exception when memory id is provided with flow type? Otherwise, user might not know if their request is utilizing memory or not if it get ignored silently?

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
owaiskazi19: This validation is done on ml commons side to return the same memory id as passed in the parameters. Should we also do it on our end?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
owaiskazi19: Good catch. Will update and add a test

src/main/java/org/opensearch/neuralsearch/query/ext/AgentStepsSearchExtBuilder.java
owaiskazi19: Sounds good, will remove the method

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
heemin32: I guess it depends on our API spec. Whether we want to allow different memoryId to be returned or not. If user expect the memory id to be same, I think we should validate it here as well.

heemin32:
> This is a nice idea. How about the processor configuration like
```
{
  "response_processors": [
    {
      "agent_context": {
        "agent_steps_summary": false,
        "memory_id": true,
        "dsl_query": true
      }
    }
  ]
}
```
This is good for `agent_steps_summary` and `dsl_query`. However, for `memory_id`, it should be handled differently, in my opinion.

For a conversational agent, we should enforce the user to always provide a `memory_id`, either as an empty value (for a new conversation) or as the `memory_id` returned from the previous request. Also, when it is a conversational agent, we must always return the `memory_id` so that it can be used for the next request without the need to enable it in the response processor.

Is there a case where the user does not use a memory_id with a conversational agent?

heemin32:
> > Is there a case where the user does not use a memory_id with a conversational agent?
> 
> Yes, where user wants to start a fresh search/ conversation. For memory_id we can always return it if the processor is enabled

I mean, is there a case where user use conversational agent without enabling the processor? Not just for the very first conversation.

owaiskazi19:
> I mean, is there a case where user use conversational agent without enabling the processor? Not just for the very first conversation.

Yes, if user doesn't enable the processor than extBuilder won't be enabled in the searchResponse and thus no memory id will be shown. To get the memory id by default the extBuilder should be enabled to show memory_id

heemin32:
> > I mean, is there a case where user use conversational agent without enabling the processor? Not just for the very first conversation.
> 
> Yes, if user doesn't enable the processor than extBuilder won't be enabled in the searchResponse and thus no memory id will be shown. To get the memory id by default the extBuilder should be enabled to show memory_id

If they don't utilize the memory, whey do they want to use conversational agent?

owaiskazi19:
> If they don't utilize the memory, whey do they want to use conversational agent?

The support for conversational agent is for smart tool calling by the agent, for ex: to use the right index from ListIndexTool, calling IndexMappingTool for fetching index mapping for that index and then calling websearchtool if the query asked is based on latest data. 
With flow agent the agent doesn't have it's own intelligence but just a sequential flow of tools present.

With the ext support through the response, user would have an option to continue the conversation as well with the same context.

src/main/java/org/opensearch/neuralsearch/processor/AgentStepsResponseProcessor.java
owaiskazi19: Added memory_id always when the processor is enabled

heemin32 (approved):

PR #1600, opened 2025-10-06
[Agentic Search] Use a single model for Agent and Query Planner Tool

rithin-pullela-aws:
### Description
1. Get's the model ID from agent and provides it as a parameter during Agent's execution.
The Model ID parameter will be picked up by the Query Planner tool and used for query generation.
Note: It is not possible for Query Planner tool to get the model ID parameter unless explicitly provided during tool registration(extra step for cx, bad experience) or when provided during Agent Execution(like this case).

2. Added a function `removeTrailingDecimalZeros` which removes the trailing zeros in decimals (eg. 12.0 -> 12)
The LLMs when generating the query always tend to add decimals even though we explicitly prompt them to generate only integers. Hence this function processes the generated DSL to replace the trailing 0s in decimal using regex

Reason for this behavior is well captured across research related to training data bias, style intertia of the LLMs etc
Example paper: https://arxiv.org/html/2408.08656v1

### Related Issues

### Check List


src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
```diff
@@ -946,7 +964,7 @@ public void testExecuteAgent_ConversationalAgent() throws Exception {
             return null;
         }).when(client).execute(any(), any(), any());
 
-        AgentInfoDTO agentInfo = new AgentInfoDTO(agentType, hasSystemPrompt, false);
+        AgentInfoDTO agentInfo = new AgentInfoDTO(agentType, hasSystemPrompt, false, null);
```
owaiskazi19: Instead of passing null for model id, you can pass a "test-model-id"

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
```diff
@@ -180,6 +183,17 @@ private void executeAgentAsync(
         }));
     }
 
+    private String removeTrailingDecimalZeros(String query) {
```
owaiskazi19: This can be done in MLCommonsClientAccessor and a final query should be return from there to the processor

mingshl:
I am thinking the parameters from agent registration is usually carried over in tool execution,

have you tried stop point at this line, and see if the `llm.model_id` is existed in `originalParameters`. I am thinking you don't need to explicitly return the model id, the originalParameters ideally should carry over the `llm.model_id`

if so, you can do null check, if 'model_id' is null and 'llm.model_id' is not null, you can take 'model_id' = 'llm.model_id'

https://github.com/opensearch-project/ml-commons/blob/47c22b51e33a176e5c3c3bd0afe61feb5dc2ac61/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java#L125

rithin-pullela-aws:
> I am thinking the parameters from agent registration is usually carried over in tool execution,
> 
> have you tried stop point at this line, and see if the `llm.model_id` is existed in `originalParameters`. I am thinking you don't need to explicitly return the model id, the originalParameters ideally should carry over the `llm.model_id`
> 
> if so, you can do null check, if 'model_id' is null and 'llm.model_id' is not null, you can take 'model_id' = 'llm.model_id'
> 
> https://github.com/opensearch-project/ml-commons/blob/47c22b51e33a176e5c3c3bd0afe61feb5dc2ac61/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/QueryPlanningTool.java#L125

The llm.modelID is not passed to the Tool's context. Hence the model ID must be passed as a parameter either during registration of the agent in tool definition or during execution.

mingshl:
I prefer to have the llm id to pass over to all the tools within ml-commons, and you can try to do this 

in chat_agent, there is a constructLLMParams, which returns the parameters within lllm block, , you also pass over the root llm id. so this conversation agent model_id can be reused in all the tools. 


https://github.com/opensearch-project/ml-commons/blob/47c22b51e33a176e5c3c3bd0afe61feb5dc2ac61/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java#L797

bzhangam:
> I prefer to have the llm id to pass over to all the tools within ml-commons, and you can try to do this
> 
> in chat_agent, there is a constructLLMParams, which returns the parameters within lllm block, , you also pass over the root llm id. so this conversation agent model_id can be reused in all the tools.
> 
> https://github.com/opensearch-project/ml-commons/blob/47c22b51e33a176e5c3c3bd0afe61feb5dc2ac61/ml-algorithms/src/main/java/org/opensearch/ml/engine/algorithms/agent/MLChatAgentRunner.java#L797

I also think if we can manage this model id within ml-common plugin it's better to do it that way.

owaiskazi19 (approved):

bzhangam (approved):

heemin32 (approved):

mingshl:
I am not blocking this PR, but I have an advice of this removeTrailingDecimalZeros, which is a helpful method on the agentic search use case, eventually, I am hoping any modifications to tools output should stay in ml-commons, so that it can be reused by different use cases. There is a[ ML processor](https://github.com/opensearch-project/ml-commons/pull/4260) to address any post tool use and post predict processing, you can consider refactoring this after ML processor is introduced.

PR #1603, opened 2025-10-07
Add Jvector support to neural-search plugin

### LINKED ISSUE #1593: [PROPOSAL] Integrate JVector plugin support to Neural Search
### What are you proposing?
We are using the [JVector](https://github.com/datastax/jvector) vector query engine, and would like to add support to the neural search plugin to be able to use jvector vector engine via the [opensearch-jvector](https://github.com/opensearch-project/opensearch-jvector) plugin.

This topic was previously discussed here: https://github.com/opensearch-project/k-NN/issues/2605
While there is future work to simplify this area by creating a common interface for vector plugins/engines (knn, jvector etc.), the current work change will enable jvector users to take advantage of neural-search plugin.

Today, there is a hard dependency of neural-search plugin to use the k-nn plugin. K-nn and JVector have very similar interfaces, and we would like our users to be able to depend on the JVector plugin instead of k-nn. 

No changes.

N/A

N/A

N/A

N/A

JVector is a well established vector search engine, and is already integrated into the opensearch ecosystem via the opensearch-jvector plugin. This feature will allow users running hybrid search queries to take advantage of the neural-search plugin, and use the jvector engine as well. 

This feature will switch the k-nn dependency to jvector for users who enable it. The performance characteristics of their queries would be dependent on the jvector engine.

akash-shankaran:
### Description
Add support for JVector query engine to neural search plugin. This provides the user an option to use JVector instead of the k-nn plugin. 

### Related Issues
Resolves #1593 

### Check List


heemin32:
Hi. We’re releasing only one flavor of the neural plugin that uses k-NN but not jVector. This means customers who want to use the neural plugin with jVector will have to build the artifact themselves, which isn’t ideal. As a result, this PR may not provide much practical value. What are your thoughts?

martin-gaievski:
Hi @akash-shankaran, thanks for the contribution and for driving the JVector integration discussion.

This PR has been in draft since October 2025 with no recent activity. Before we proceed, I wanted to share some thoughts on the approach:

Two-flavor build approach (swapping k-NN for JVector at build time) has a fundamental limitation - OpenSearch ships a single distribution artifact, so only one flavor can be included. Users wanting the other flavor would need to build from
source, which limits the practical value significantly. It also introduces ongoing maintenance burden since every change to neural-search would need to be validated against both engines.
The right long-term path is the core interface migration proposed in [RFC #20050](https://github.com/opensearch-project/OpenSearch/issues/20050), once KNN interfaces move to OpenSearch core, neural-search can depend on those interfaces andwork with any vector engine (k-NN, JVector, or future engines) without separate builds. This was agreed upon in [issue #18](https://github.com/opensearch-project/opensearch-project/.github/issues/18).

Given the above, I'd suggest we close this PR and focus efforts on advancing RFC #20050 instead. That would provide a clean, maintainable solution that benefits the entire ecosystem.

Do you have any updates on your plans here, or thoughts on this direction? If we don't hear back in the next couple of weeks, we'll go ahead and close this PR.

PR #1605, opened 2025-10-07
Add support for asymmetric embedding models

fen-qin:
### Description
This PR is a rebasing on top of #710 

## E2E Tests
* register a local asymmetric model
```json
curl -XPOST "http://localhost:9200/_plugins/_ml/models/_register" -H 'Content-Type: application/json' -d'
{
  "name": "traced_small_model",
  "version": "1.0.0",
  "model_format": "TORCH_SCRIPT",
  "model_task_type": "text_embedding",
  "model_content_hash_value": "e13b74006290a9d0f58c1376f9629d4ebc05a0f9385f40db837452b167ae9021",
  "model_group_id": "ytHnwJkBA2y8vkecT0nP",
  "model_config": {
    "model_type": "bert",
    "embedding_dimension": 768,
    "framework_type": "sentence_transformers",
    "passage_prefix" : "passage: ",
    "query_prefix" : "query: ",
    "all_config": "{\"architectures\":[\"BertModel\"],\"max_position_embeddings\":512,\"model_type\":\"bert\",\"num_attention_heads\":12,\"num_hidden_layers\":6}"
  },
  "url": "https://github.com/opensearch-project/ml-commons/blob/2.x/ml-algorithms/src/test/resources/org/opensearch/ml/engine/algorithms/text_embedding/traced_small_model.zip?raw=true"
}'
{"task_id":"zNHnwJkBA2y8vkeckEnm","status":"CREATED"}
```

* test the model
    * The embedding model chosen is asymmetric. To use it, you must declare whether the input is of type QUERY or of type PASSAGE.
```json
curl -XPOST "http://localhost:9200/_plugins/_ml/_predict/text_embedding/3zfSwJkB9yHsUFukoFGN" -H 'Content-Type: application/json' -d'
{
  "parameters": {
    "content_type": "query"
  },
  "text_docs": ["What day is it today?"],
  "target_response": ["sentence_embedding"]
}'
{"inference_results":[{"output":[{"name":"sentence_embedding","data_type":"FLOAT32","shape":[768],"data":[0.4277453,0.36682454,0.57426196,0.53548354,0.5545153,0.68228644,0.37400416,0.6416417,0.6905142,0.5195829,0.46663672,0.3864537,0.419837,0.45926508,0.4904398,0.61876655,0.5276075,0.5057253,0.44570318,0.6018752,0.55649084,0.4044214,0.46216413, ....]}]}]}
```

* create an index
```json
curl -XPUT "http://localhost:9200/nyc_facts" -H 'Content-Type: application/json' -d'
{
  "settings": {
    "index.knn": true,
    "default_pipeline": "asymmetric_embedding_ingest_pipeline",
    "knn.algo_param.ef_search": 100
  },
  "mappings": {
    "properties":  {
      "fact_embedding": {
        "type": "knn_vector",
        "dimension": 768,
        "method": {
          "name": "hnsw",
          "space_type": "l2",
          "engine": "faiss",
          "parameters": {
            "ef_construction": 128,
            "m": 24
          }
        }
      }
    }
  }
}'
{"acknowledged":true,"shards_acknowledged":true,"index":"nyc_facts"}
```

* create an ingest pipeline
```json
curl -XPUT "http://localhost:9200/_ingest/pipeline/asymmetric_embedding_ingest_pipeline" -H 'Content-Type: application/json' -d'
{
    "description": "ingest passage text and generate a embedding using an asymmetric model",
    "processors": [
        {
            "ml_inference": {

                "model_input": "{\"text_docs\":[\"${input_map.text_docs}\"],\"target_response\":[\"sentence_embedding\"],\"parameters\":{\"content_type\":\"query\"}}",
                "function_name": "text_embedding",
                "model_id": "ztHnwJkBA2y8vkeckUng",
                "input_map": [
                    {
                        "text_docs": "description"
                    }
                ],
                "output_map": [
                    {
                        "fact_embedding": "$.inference_results[0].output[0].data",
                        "embedding_size": "$.inference_results.*.output.*.shape[0]"
                    }
                ]
            }
        }
    ]
}'
{"acknowledged":true}
```

* ingest data
```json
curl -XPOST "http://localhost:9200/_bulk" -H 'Content-Type: application/json' -d'
{ "index": { "_index": "nyc_facts" } }
{ "title": "Central Park", "description": "A large public park in the heart of New York City, offering a wide range of recreational activities." }
{ "index": { "_index": "nyc_facts" } }
{ "title": "Empire State Building", "description": "An iconic skyscraper in New York City offering breathtaking views from its observation deck." }
{ "index": { "_index": "nyc_facts" } }
{ "title": "Statue of Liberty", "description": "A colossal neoclassical sculpture on Liberty Island, symbolizing freedom and democracy in the United States." }
{ "index": { "_index": "nyc_facts" } }
{ "title": "Brooklyn Bridge", "description": "A historic suspension bridge connecting Manhattan and Brooklyn, offering pedestrian walkways with great views." }
{ "index": { "_index": "nyc_facts" } }
{ "title": "Times Square", "description": "A bustling commercial and entertainment hub in Manhattan, known for its neon lights and Broadway theaters." }
{ "index": { "_index": "nyc_facts" } }
{ "title": "Yankee Stadium", "description": "Home to the New York Yankees, this baseball stadium is a historic landmark in the Bronx." }
{ "index": { "_index": "nyc_facts" } }
{ "title": "The Bronx Zoo", "description": "One of the largest zoos in the world, located in the Bronx, featuring diverse animal exhibits and conservation efforts." }
{ "index": { "_index": "nyc_facts" } }
{ "title": "New York Botanical Garden", "description": "A large botanical garden in the Bronx, known for its diverse plant collections and stunning landscapes." }
{ "index": { "_index": "nyc_facts" } }
{ "title": "Flushing Meadows-Corona Park", "description": "A major park in Queens, home to the USTA Billie Jean King National Tennis Center and the Unisphere." }
{ "index": { "_index": "nyc_facts" } }
{ "title": "Citi Field", "description": "The home stadium of the New York Mets, located in Queens, known for its modern design and fan-friendly atmosphere." }
{ "index": { "_index": "nyc_facts" } }
{ "title": "Rockefeller Center", "description": "A famous complex of commercial buildings in Manhattan, home to the NBC studios and the annual ice skating rink." }
{ "index": { "_index": "nyc_facts" } }
{ "title": "Queens Botanical Garden", "description": "A peaceful, beautiful botanical garden located in Flushing, Queens, featuring seasonal displays and plant collections." }
{ "index": { "_index": "nyc_facts" } }
{ "title": "Arthur Ashe Stadium", "description": "The largest tennis stadium in the world, located in Flushing Meadows-Corona Park, Queens, hosting the U.S. Open." }
{ "index": { "_index": "nyc_facts" } }
{ "title": "Wave Hill", "description": "A public garden and cultural center in the Bronx, offering stunning views of the Hudson River and a variety of nature programs." }
{ "index": { "_index": "nyc_facts" } }
{ "title": "Louis Armstrong House", "description": "The former home of jazz legend Louis Armstrong, located in Corona, Queens, now a museum celebrating his life and music." }'
```

* create a search pipeline
```json
curl -XPUT "http://localhost:9200/_search/pipeline/asymmetric_embedding_search_pipeline" -H 'Content-Type: application/json' -d'
{
   "description": "ingest passage text and generate a embedding using an asymmetric model",
   "request_processors": [
      {
        "ml_inference": {
            "query_template": "{\"size\": 3,\"query\": {\"knn\": {\"fact_embedding\": {\"vector\": ${query_embedding},\"k\": 4}}}}",
            "function_name": "text_embedding",
            "model_id": "3zfSwJkB9yHsUFukoFGN",
            "model_input": "{ \"text_docs\": [\"${input_map.query}\"], \"target_response\": [\"sentence_embedding\"], \"parameters\" : {\"content_type\" : \"query\" } }",
            "input_map": [
               {
                  "query": "query.term.fact_embedding.value"
               }
            ],
            "output_map": [
               {
                  "query_embedding": "$.inference_results[0].output[0].data",
                  "embedding_size": "$.inference_results.*.output.*.shape[0]"
               }
            ]
         }
      }
   ]
}'
```

* run a query
```json
curl -XGET "http://localhost:9200/nyc_facts/_search" -H 'Content-Type: application/json' -d'
{
  "_source": {
    "excludes": [
      "fact_embedding"
    ]
  },
  "query": {
    "neural": {
      "fact_embedding": {
        "query_text": "What are some places for sports in NYC?",
        "model_id": "fpb3wJkBt8Q-zKDiG4F6",
        "boost": 1
      }
    }
  }
}'
```

### Related Issues
#620 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
@@ -30,6 +30,7 @@ import java.util.function.Supplier;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import org.opensearch.common.CheckedConsumer;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.common.Nullable;
 import org.opensearch.common.xcontent.XContentType;
@@ -50,6 +51,8 @@ import org.opensearch.ml.common.dataset.remote.RemoteInferenceInputDataSet;
 import org.opensearch.ml.common.input.MLInput;
 import org.opensearch.ml.common.input.execute.agent.AgentMLInput;
 import org.opensearch.ml.common.input.parameter.MLAlgoParams;
+import org.opensearch.ml.common.model.MLModelConfig;
+import org.opensearch.ml.common.model.TextEmbeddingModelConfig;
 import org.opensearch.ml.common.output.MLOutput;
 import org.opensearch.ml.common.output.model.ModelResultFilter;
 import org.opensearch.ml.common.output.model.ModelTensor;
@@ -80,6 +83,8 @@ import lombok.extern.log4j.Log4j2;
 @Log4j2
 public class MLCommonsClientAccessor {
     private final MachineLearningNodeClient mlClient;
+    private final Map<String, Boolean> modelAsymmetryCache = new ConcurrentHashMap<>();
+
     private static final Gson gson = new Gson();
 
     /**
@@ -127,13 +132,16 @@ public class MLCommonsClientAccessor {
         @NonNull final TextInferenceRequest inferenceRequest,
         @NonNull final ActionListener<List<List<Number>>> listener
     ) {
-        retryableInference(
-            inferenceRequest,
-            0,
-            () -> createMLTextInput(inferenceRequest.getTargetResponseFilters(), inferenceRequest.getInputTexts()),
-            this::buildVectorFromResponse,
-            listener
-        );
+        checkModelAsymmetryAndThenPredict(inferenceRequest.getModelId(), listener::onFailure, isAsymmetric -> {
+            MLAlgoParams params = isAsymmetric ? inferenceRequest.getMlAlgoParams() : null;
+            retryableInference(
+                inferenceRequest,
+                0,
+                () -> createMLTextInput(inferenceRequest.getTargetResponseFilters(), inferenceRequest.getInputTexts(), params),
+                this::buildVectorFromResponse,
+                listener
+            );
+        });
     }
 
     public void inferenceSentencesWithMapResult(
@@ -141,13 +149,16 @@ public class MLCommonsClientAccessor {
         @Nullable MLAlgoParams mlAlgoParams,
         @NonNull final ActionListener<List<Map<String, ?>>> listener
     ) {
-        retryableInference(inferenceRequest, 0, () -> {
-            MLInput input = createMLTextInput(null, inferenceRequest.getInputTexts());
-            if (mlAlgoParams != null) {
-                input.setParameters(mlAlgoParams);
-            }
-            return input;
-        }, this::buildMapResultFromResponse, listener);
+        checkModelAsymmetryAndThenPredict(inferenceRequest.getModelId(), listener::onFailure, isAsymmetric -> {
+            MLAlgoParams params = isAsymmetric ? (mlAlgoParams != null ? mlAlgoParams : inferenceRequest.getMlAlgoParams()) : mlAlgoParams;
+            retryableInference(
+                inferenceRequest,
+                0,
+                () -> createMLTextInput(null, inferenceRequest.getInputTexts(), params),
+                this::buildMapResultFromResponse,
+                listener
+            );
+        });
     }
 
     /**
@@ -159,13 +170,16 @@ public class MLCommonsClientAccessor {
      * @param listener         {@link ActionListener} which will be called when prediction is completed or errored out.
      */
     public void inferenceSentencesMap(@NonNull MapInferenceRequest inferenceRequest, @NonNull final ActionListener<List<Number>> listener) {
-        retryableInference(
-            inferenceRequest,
-            0,
-            () -> createMLMultimodalInput(inferenceRequest.getTargetResponseFilters(), inferenceRequest.getInputObjects()),
-            this::buildSingleVectorFromResponse,
-            listener
-        );
+        checkModelAsymmetryAndThenPredict(inferenceRequest.getModelId(), listener::onFailure, isAsymmetric -> {
+            MLAlgoParams params = isAsymmetric ? inferenceRequest.getMlAlgoParams() : null;
+            retryableInference(
+                inferenceRequest,
+                0,
+                () -> createMLMultimodalInput(inferenceRequest.getTargetResponseFilters(), inferenceRequest.getInputObjects(), params),
+                this::buildSingleVectorFromResponse,
+                listener
+            );
+        });
     }
 
     /**
@@ -221,10 +235,10 @@ public class MLCommonsClientAccessor {
         ));
     }
 
-    private MLInput createMLTextInput(final List<String> targetResponseFilters, List<String> inputText) {
+    private MLInput createMLTextInput(final List<String> targetResponseFilters, List<String> inputText, MLAlgoParams mlAlgoParams) {
         final ModelResultFilter modelResultFilter = new ModelResultFilter(false, true, targetResponseFilters, null);
         final MLInputDataset inputDataset = new TextDocsInputDataSet(inputText, modelResultFilter);
-        return new MLInput(FunctionName.TEXT_EMBEDDING, null, inputDataset);
+        return new MLInput(FunctionName.TEXT_EMBEDDING, mlAlgoParams, inputDataset);
     }
 
     private MLInput createMLTextPairsInput(final String query, final List<String> inputText) {
@@ -342,7 +356,11 @@ public class MLCommonsClientAccessor {
         }
     }
 
-    private MLInput createMLMultimodalInput(final List<String> targetResponseFilters, final Map<String, String> input) {
+    private MLInput createMLMultimodalInput(
+        final List<String> targetResponseFilters,
+        final Map<String, String> input,
+        MLAlgoParams mlAlgoParams
+    ) {
         List<String> inputText = new ArrayList<>();
         inputText.add(input.get(INPUT_TEXT));
         if (input.containsKey(INPUT_IMAGE)) {
@@ -350,7 +368,7 @@ public class MLCommonsClientAccessor {
         }
         final ModelResultFilter modelResultFilter = new ModelResultFilter(false, true, targetResponseFilters, null);
         final MLInputDataset inputDataset = new TextDocsInputDataSet(inputText, modelResultFilter);
-        return new MLInput(FunctionName.TEXT_EMBEDDING, null, inputDataset);
+        return new MLInput(FunctionName.TEXT_EMBEDDING, mlAlgoParams, inputDataset);
     }
 
     public void getModel(@NonNull final String modelId, @NonNull final ActionListener<MLModel> listener) {
@@ -1063,4 +1081,36 @@ public class MLCommonsClientAccessor {
         return highlights;
     }
 
+    /**
+     * Check if the model is asymmetric and then run the prediction. Model asymmetry is a concept
+     * that is specific to TextEmbeddingModelConfig. If the model is not a TextEmbeddingModel, then
+     * this check is not applicable.
+     *
+     * The asymmetry of a model is static for a given model. To avoid repeated checks for the same
+     * model, we cache the model asymmetry status. Non-TextEmbeddingModels are cached as false.
+     *
+     * @param modelId    The model id to check
+     * @param onFailure The action to take if the model cannot be retrieved
+     * @param runPrediction The action to take if the model is successfully retrieved
+     */
+    private void checkModelAsymmetryAndThenPredict(String modelId, Consumer<Exception> onFailure, Consumer<Boolean> runPrediction) {
+        CheckedConsumer<MLModel, ? extends Exception> checkModelAsymmetryListener = model -> {
+            MLModelConfig modelConfig = model.getModelConfig();
+            if (!(modelConfig instanceof TextEmbeddingModelConfig textEmbeddingModelConfig)) {
+                modelAsymmetryCache.computeIfAbsent(modelId, k -> false);
+                return;
+            }
+            final boolean isAsymmetricModel = textEmbeddingModelConfig.getPassagePrefix() != null
+                || textEmbeddingModelConfig.getQueryPrefix() != null;
+            modelAsymmetryCache.computeIfAbsent(modelId, k -> isAsymmetricModel);
+        };
+        if (modelAsymmetryCache.get(modelId) != null) {
+            runPrediction.accept(modelAsymmetryCache.get(modelId));
+        } else {
+            mlClient.getModel(modelId, null, ActionListener.<MLModel>wrap(mlModel -> {
+                checkModelAsymmetryListener.accept(mlModel);
+                runPrediction.accept(modelAsymmetryCache.get(modelId));
+            }, onFailure));
+        }
+    }
 }
--- a/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
@@ -91,7 +91,8 @@ import org.opensearch.neuralsearch.query.dto.NeuralQueryBuildStage;
 import org.opensearch.neuralsearch.stats.events.EventStatName;
 import org.opensearch.neuralsearch.stats.events.EventStatsManager;
 import org.opensearch.neuralsearch.util.NeuralSearchClusterUtil;
-
+import org.opensearch.ml.common.input.parameter.textembedding.AsymmetricTextEmbeddingParameters;
+import org.opensearch.ml.common.input.parameter.textembedding.AsymmetricTextEmbeddingParameters.EmbeddingContentType;
 import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -686,11 +687,8 @@ public class NeuralQueryBuilder extends AbstractNeuralQueryBuilder<NeuralQueryBu
             return rewriteQueryWithQueryShardContext(queryShardContext);
         }
 
-        // We can rewrite with a BaseQueryRewriteContext on the shard level:
-        // In SearchService do a quick rewrite to handle AliasFilters rewrite. This happens after the
-        // coordinate rewrite, and we do not need to rewrite the neural query for this case. We will do rewrite later
-        // with queryShardContext.
-        return this;
+        // Fall back to the old logic when both conversions return null
+        return rewriteQueryAgainstKnnField(queryRewriteContext);
     }
 
     private QueryBuilder rewriteQueryWithCoordinatorContext(@NonNull final QueryCoordinatorContext queryRewriteContext) {
@@ -918,7 +916,11 @@ public class NeuralQueryBuilder extends AbstractNeuralQueryBuilder<NeuralQueryBu
 
         queryRewriteContext.registerAsyncAction(
             ((client, actionListener) -> ML_CLIENT.inferenceSentencesMap(
-                MapInferenceRequest.builder().modelId(modelId()).inputObjects(inferenceInput).build(),
+                MapInferenceRequest.builder()
+                    .modelId(modelId())
+                    .inputObjects(inferenceInput)
+                    .mlAlgoParams(AsymmetricTextEmbeddingParameters.builder().embeddingContentType(EmbeddingContentType.QUERY).build())
+                    .build(),
                 ActionListener.wrap(floatList -> {
                     vectorSetOnce.set(vectorAsListToArray(floatList));
                     actionListener.onResponse(null);
--- a/src/test/resources/processor/UploadAsymmetricModelRequestBody.json
+++ b/src/test/resources/processor/UploadAsymmetricModelRequestBody.json
@@ -1,0 +1,17 @@
+{
+  "name": "traced_small_model",
+  "version": "1.0.0",
+  "model_format": "TORCH_SCRIPT",
+  "model_task_type": "text_embedding",
+  "model_content_hash_value": "e13b74006290a9d0f58c1376f9629d4ebc05a0f9385f40db837452b167ae9021",
+  "model_group_id": "%s",
+  "model_config": {
+    "model_type": "bert",
+    "embedding_dimension": 768,
+    "framework_type": "sentence_transformers",
+    "passage_prefix" : "passage: ",
+    "query_prefix" : "query: ",
+    "all_config": "{\"architectures\":[\"BertModel\"],\"max_position_embeddings\":512,\"model_type\":\"bert\",\"num_attention_heads\":12,\"num_hidden_layers\":6}"
+  },
+  "url": "https://github.com/opensearch-project/ml-commons/blob/2.x/ml-algorithms/src/test/resources/org/opensearch/ml/engine/algorithms/text_embedding/traced_small_model.zip?raw=true"
+}
--- a/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
@@ -25,6 +25,8 @@ import org.opensearch.index.mapper.IndexFieldMapper;
 import org.opensearch.ingest.AbstractBatchingProcessor;
 import org.opensearch.ingest.IngestDocument;
 import org.opensearch.ingest.IngestDocumentWrapper;
+import org.opensearch.ml.common.input.parameter.textembedding.AsymmetricTextEmbeddingParameters;
+import org.opensearch.ml.common.input.parameter.textembedding.AsymmetricTextEmbeddingParameters.EmbeddingContentType;
 import org.opensearch.ml.common.input.parameter.MLAlgoParams;
 import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
 import org.opensearch.neuralsearch.processor.optimization.InferenceFilter;
@@ -779,7 +781,11 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
         BiConsumer<IngestDocument, Exception> handler
     ) {
         mlCommonsClientAccessor.inferenceSentences(
-            TextInferenceRequest.builder().modelId(this.modelId).inputTexts(inferenceList).build(),
+            TextInferenceRequest.builder()
+                .modelId(this.modelId)
+                .inputTexts(inferenceList)
+                .mlAlgoParams(AsymmetricTextEmbeddingParameters.builder().embeddingContentType(EmbeddingContentType.PASSAGE).build())
+                .build(),
             ActionListener.wrap(vectors -> {
                 setVectorFieldsToDocument(ingestDocument, processMap, vectors);
                 handler.accept(ingestDocument, null);
--- a/src/main/java/org/opensearch/neuralsearch/processor/InferenceRequest.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/InferenceRequest.java
@@ -6,6 +6,7 @@ package org.opensearch.neuralsearch.processor;
 
 import java.util.List;
 
+import org.opensearch.ml.common.input.parameter.MLAlgoParams;
 import lombok.Builder;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
@@ -34,4 +35,9 @@ public abstract class InferenceRequest {
      */
     @Builder.Default
     private List<String> targetResponseFilters = List.of("sentence_embedding");
+    /**
+     * ML algorithm parameters for asymmetric models.
+     * Contains parameters like passage_prefix and query_prefix for asymmetric embedding models.
+     */
+    private MLAlgoParams mlAlgoParams;
 }
--- a/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/TextEmbeddingProcessor.java
@@ -19,6 +19,8 @@ import org.opensearch.core.common.util.CollectionUtils;
 import org.opensearch.env.Environment;
 import org.opensearch.ingest.IngestDocument;
 import org.opensearch.ingest.IngestDocumentWrapper;
+import org.opensearch.ml.common.input.parameter.textembedding.AsymmetricTextEmbeddingParameters;
+import org.opensearch.ml.common.input.parameter.textembedding.AsymmetricTextEmbeddingParameters.EmbeddingContentType;
 import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
 
 import lombok.extern.log4j.Log4j2;
@@ -105,7 +107,11 @@ public final class TextEmbeddingProcessor extends InferenceProcessor {
     @Override
     public void doBatchExecute(List<String> inferenceList, Consumer<List<?>> handler, Consumer<Exception> onException) {
         mlCommonsClientAccessor.inferenceSentences(
-            TextInferenceRequest.builder().modelId(this.modelId).inputTexts(inferenceList).build(),
+            TextInferenceRequest.builder()
+                .modelId(this.modelId)
+                .inputTexts(inferenceList)
+                .mlAlgoParams(AsymmetricTextEmbeddingParameters.builder().embeddingContentType(EmbeddingContentType.PASSAGE).build())
+                .build(),
             ActionListener.wrap(handler::accept, onException)
         );
     }
--- a/src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessorTests.java
@@ -51,6 +51,8 @@ import org.opensearch.ml.common.output.model.ModelTensors;
 import org.opensearch.ml.common.transport.execute.MLExecuteTaskResponse;
 import org.opensearch.neuralsearch.constants.TestCommonConstants;
 import org.opensearch.neuralsearch.processor.highlight.SentenceHighlightingRequest;
+import org.opensearch.neuralsearch.processor.TextInferenceRequest;
+import org.opensearch.neuralsearch.processor.MapInferenceRequest;
 import org.opensearch.test.OpenSearchTestCase;
 import org.opensearch.transport.NodeNotConnectedException;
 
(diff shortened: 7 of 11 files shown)


src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -686,11 +687,8 @@ protected QueryBuilder doRewrite(QueryRewriteContext queryRewriteContext) {
             return rewriteQueryWithQueryShardContext(queryShardContext);
         }
 
-        // We can rewrite with a BaseQueryRewriteContext on the shard level:
-        // In SearchService do a quick rewrite to handle AliasFilters rewrite. This happens after the
-        // coordinate rewrite, and we do not need to rewrite the neural query for this case. We will do rewrite later
-        // with queryShardContext.
-        return this;
+        // Fall back to the old logic when both conversions return null
```
heemin32: Could you tell about this change? I think this is a mistake during rebase?

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -918,7 +916,11 @@ private QueryBuilder rewriteQueryAgainstKnnField(QueryRewriteContext queryRewrit
 
         queryRewriteContext.registerAsyncAction(
             ((client, actionListener) -> ML_CLIENT.inferenceSentencesMap(
-                MapInferenceRequest.builder().modelId(modelId()).inputObjects(inferenceInput).build(),
+                MapInferenceRequest.builder()
+                    .modelId(modelId())
+                    .inputObjects(inferenceInput)
+                    .mlAlgoParams(AsymmetricTextEmbeddingParameters.builder().embeddingContentType(EmbeddingContentType.QUERY).build())
```
heemin32: I’m not sure if passing AsymmetricTextEmbeddingParameters here is the right choice, since we don’t yet know whether the model is asymmetric. I think we should define our own enum and pass the appropriate value once we’ve confirmed the model type.
Also, within MapInferenceRequest, it might be better to validate that the content type is not null to prevent accidental omissions.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -80,6 +83,8 @@
 @Log4j2
 public class MLCommonsClientAccessor {
     private final MachineLearningNodeClient mlClient;
+    private final Map<String, Boolean> modelAsymmetryCache = new ConcurrentHashMap<>();
```
heemin32: When will the cache get evicted? We should set some limit and evict when the limit exceed.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -127,27 +132,33 @@ public void inferenceSentences(
...
     ) {
-        retryableInference(inferenceRequest, 0, () -> {
-            MLInput input = createMLTextInput(null, inferenceRequest.getInputTexts());
-            if (mlAlgoParams != null) {
-                input.setParameters(mlAlgoParams);
-            }
-            return input;
-        }, this::buildMapResultFromResponse, listener);
+        checkModelAsymmetryAndThenPredict(inferenceRequest.getModelId(), listener::onFailure, isAsymmetric -> {
+            MLAlgoParams params = isAsymmetric ? (mlAlgoParams != null ? mlAlgoParams : inferenceRequest.getMlAlgoParams()) : mlAlgoParams;
```
heemin32: This logic is not correct. With this logic, the content type from inferenceRequest will get ignored when mlAlgoParams is null. I think mlAlgoParams should come from `inferenceRequest` now?

Otherwise, we need to build new MLAlgoParams by combining the value together.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -159,13 +170,16 @@ public void inferenceSentencesWithMapResult(
...
     public void inferenceSentencesMap(@NonNull MapInferenceRequest inferenceRequest, @NonNull final ActionListener<List<Number>> listener) {
-        retryableInference(
-            inferenceRequest,
-            0,
-            () -> createMLMultimodalInput(inferenceRequest.getTargetResponseFilters(), inferenceRequest.getInputObjects()),
-            this::buildSingleVectorFromResponse,
-            listener
-        );
+        checkModelAsymmetryAndThenPredict(inferenceRequest.getModelId(), listener::onFailure, isAsymmetric -> {
+            MLAlgoParams params = isAsymmetric ? inferenceRequest.getMlAlgoParams() : null;
```
heemin32: QQ. what happens if we pass the MLALgoParams without verifying the actual model type? Will it throw an exception?

I think inferenceRequest better to hold if this is for passage or query. Then here before inference we construct MLAlgoParams properly based on if the model is symmetric or not.
```
MLAlgoParams params = inferenceRequest.createMlAlgoParams(isAsymmetric);
```

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -1063,4 +1081,36 @@ private List<Map<String, Object>> processDocumentHighlights(Object docHighlights
...
+     *
+     * The asymmetry of a model is static for a given model. To avoid repeated checks for the same
+     * model, we cache the model asymmetry status. Non-TextEmbeddingModels are cached as false.
+     *
+     * @param modelId    The model id to check
+     * @param onFailure The action to take if the model cannot be retrieved
+     * @param runPrediction The action to take if the model is successfully retrieved
+     */
+    private void checkModelAsymmetryAndThenPredict(String modelId, Consumer<Exception> onFailure, Consumer<Boolean> runPrediction) {
+        CheckedConsumer<MLModel, ? extends Exception> checkModelAsymmetryListener = model -> {
```
heemin32: This is unnecessary usage of consumer imo. Can we convert this as method for better readability?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -1063,4 +1081,36 @@ private List<Map<String, Object>> processDocumentHighlights(Object docHighlights
...
+                return;
+            }
+            final boolean isAsymmetricModel = textEmbeddingModelConfig.getPassagePrefix() != null
+                || textEmbeddingModelConfig.getQueryPrefix() != null;
+            modelAsymmetryCache.computeIfAbsent(modelId, k -> isAsymmetricModel);
+        };
+        if (modelAsymmetryCache.get(modelId) != null) {
+            runPrediction.accept(modelAsymmetryCache.get(modelId));
+        } else {
+            mlClient.getModel(modelId, null, ActionListener.<MLModel>wrap(mlModel -> {
```
heemin32: Although this is not very critical, this implementation will case multiple getModel call when the model is accessed for the very first time.

I wonder if we have to call `getModel` in async way when getModel is fast operation in general.
```
boolean isAsymmetricModel = modelAsymmetricCache.computeIfAbset(modelId, id -> {
  mlModel = mlClient.getModel(id);
  return isAsymmetricModel(mlModel);
}

params = inferenceRequest.createMlAlgoParam(isAsymmetricModel);
retryableInference(
                inferenceRequest,
                0,
                () -> createMLTextInput(null, inferenceRequest.getInputTexts(), params),
                this::buildMapResultFromResponse,
                listener
            );
```

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
fen-qin: clear the rebase errors

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
fen-qin: added evicted for cache

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
fen-qin: Yes. I'm thinking hard on how to parse AsymmetricTextEmbeddingParameters .Since isAsymmetricModel check is in MLCommonsClientAccessor, so I want to put conditional check related code in ml-commons accessor as possible. Here is the revision logics:
 - for query, asymmetric model would introduce content_type as `query`
 - for processor, asymmetric model would introduce content_type as `passage`

by doing so, we no longer need to parse content_type from NeuralQueryBuilder, removing all code changes in NeuralSearchBuilder. And we no longer need to rewrite the query with content_type.

but we still need to parse content_type from processor as a flag to differ if the calls coming from processor or query. So in ml-commons accessor, after getting isSymmetricModel boolean value, the ml algor params will be built following the logics: 
- For symmetric models: Returns null (no special parameters needed)
- For asymmetric models: Creates AsymmetricTextEmbeddingParameters with appropriate content type:
   - QUERY: for search queries (default if not specified)
   - PASSAGE: for document ingestion (explicitly set by processors)

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
fen-qin: same here

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
fen-qin: same here

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
fen-qin: yes. remove Consumer and simplify the isAsymmetricModel functions

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
fen-qin: with cache mechanism: 
- Cache hit: Immediately calls runPrediction.accept(cachedAsymmetry) - no model lookup
- Cache miss: Calls mlClient.getModel() asynchronously, caches result, then calls runPrediction.accept(isAsymmetric)
the implementation already handles async calls and avoids multiple getModel calls through caching.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
heemin32: Better to use in memory cache supported by com.github.ben-manes.caffeine or com.google.guava?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
heemin32: Typo. I wonder if we have to call getModel in sync way.

My point is, for the very first few model access requests, they all will end up getting model config.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -1063,4 +1086,116 @@ private List<Map<String, Object>> processDocumentHighlights(Object docHighlights
...
+        if (!isAsymmetric) {
+            return null;
+        }
+
+        // Use pre-set parameters if provided
+        if (inferenceRequest.getMlAlgoParams() != null) {
+            return inferenceRequest.getMlAlgoParams();
+        }
+
+        // Determine content type: QUERY (default for search) or PASSAGE (for ingestion)
```
heemin32: No default. We should throw exception when content type is not set.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -141,13 +152,18 @@ public void inferenceSentencesWithMapResult(
...
     ) {
-        retryableInference(inferenceRequest, 0, () -> {
-            MLInput input = createMLTextInput(null, inferenceRequest.getInputTexts());
-            if (mlAlgoParams != null) {
-                input.setParameters(mlAlgoParams);
-            }
-            return input;
-        }, this::buildMapResultFromResponse, listener);
+        checkModelAsymmetryAndThenPredict(inferenceRequest.getModelId(), listener::onFailure, isAsymmetric -> {
+            MLAlgoParams params = isAsymmetric
```
heemin32: Can we just remove mlAlgoParams and always receive it through request?

Also, we only need to set the content type as null if it is symmetric model. No more logic.

fen-qin: ideally, ml-commons should be the centralized place to handle any requests no matter it's symmetric or asymmetric. However, when calling predict API using symmetric model with AsymmetricTextEmbeddingParameters, it will throw exception. 

-Testing  Request
```
curl -XPOST "http://localhost:9200/_plugins/_ml/_predict/text_embedding/i4AyCJoBseJsgUCdfPXp" -H 'Content-Type: application/json' -d'
{
  "parameters": {
    "content_type": "query"
  },
  "text_docs": ["What day is it today?"],
  "target_response": ["sentence_embedding"]
}
'

{"error":{"root_cause":[{"type":"m_l_exception","reason":"Failed to inference TEXT_EMBEDDING model: i4AyCJoBseJsgUCdfPXp"}],"type":"m_l_exception","reason":"Failed to inference TEXT_EMBEDDING model: i4AyCJoBseJsgUCdfPXp","caused_by":{"type":"illegal_argument_exception","reason":"When passing AsymmetricTextEmbeddingParameters, the model requires to be registered with at least one of `query_prefix` or `passage_prefix`."}},"status":500}
```

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
fen-qin: this content_type was previously hard-coded:
- NeuralQueryBuilder (with content_type as query)
- TextEmbeddingProcessor (with content_type as passage)

In this revision, I tried to minimize the changes in NeuralQueryBuilder and doRewrite. There is no need to rewrite the query context when a value is hard-coded. Instead content_type is:
- defaulted as query if it's not defined for an asymmetric embedding model
- TextEmbeddingProcessor (with content_type as passage)

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
fen-qin: yes. let me updated to use google.guava CacheBuilder

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
fen-qin: Let me try to use more optimal solution to handle concurrency case for very first getModel scenario:
- Guava cache for results
```
private final Cache<String, Boolean> modelAsymmetryCache = CacheBuilder.newBuilder().maximumSize(1000).build();
```

- pending requests for coordination
  - first request trigger getModel
  - concurrent requests wait and share the result
  - all callbacks execute when model metadata arrives
```
private final Map<String, List<Consumer<Boolean>>> pendingRequests = new ConcurrentHashMap<>();
```

- Flow Explanation:
```
// Request 1 arrives
cached = modelAsymmetryCache.getIfPresent(modelId);  // null (cache miss)
callbacks = pendingRequests.compute(...);             // size=1, triggers getModel

// Request 2 arrives (concurrent)
cached = modelAsymmetryCache.getIfPresent(modelId);  // null (still fetching)
callbacks = pendingRequests.compute(...);             // size=2, waits

// Request 3 arrives (concurrent)
cached = modelAsymmetryCache.getIfPresent(modelId);  // null (still fetching)
callbacks = pendingRequests.compute(...);             // size=3, waits

// getModel completes
modelAsymmetryCache.put(modelId, true);              // Cache result
pending = pendingRequests.remove(modelId);           // Remove callbacks
pending.forEach(cb -> cb.accept(true));              // Execute all 3 callbacks

// Request 4 arrives (later)
cached = modelAsymmetryCache.getIfPresent(modelId);  // true (cache hit!)
runPrediction.accept(true);                          // Immediate execution
```

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
heemin32: Thanks. At least, we could remove `mlAlgoParams` here and receive everything from request? Then, this method will be same as `inferenceSentencesMap` method.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
heemin32: I suggest not to focus on minimizing the change. 
If we remove inferenceRequest.getEmbeddingContentType and handle everything inside mlAlgoParams, that would be more cleaner.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
fen-qin: yes. The logic is handled in this function

```
MLAlgoParams createMLAlgoParams(boolean isAsymmetric, InferenceRequest inferenceRequest)
```

- if isAsymmetric == false, which means this is a symmetric model
   - will remove `mlAlgoParams` with contentType, and read everything from request
- if isAsymmetric == true, which means this is an asymmetric model
   - will read `mlAlgoParams` with contentType if there is one, or use defaulted as QUERY

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
fen-qin: it's not only for minimizing the changes.
If a model is asymmetric, the mlAlgoParams is anyway generated and parsed from neural-search instead of the query terms. 

And the problem is where to generate these asymmetric mlAlgoParams:
- for ingestion processor, content_type = PASSAGE (there is no change on adding PASSAGE from text embedding processor)
- for query, content_type = QUERY (this is a change between previous revision and this revision)
   - adding hard-coded QUERY in NeuralQueryBuilder and rewrite query context 
   - adding as defaulted value in ml-commons accessor only if the model is asymmetric (preferred)

Why the 2nd option is more preferred ?
- for NeuralQueryBuilder level, we don't know if the model is asymmetric or symmetric, and if we hard-coded at NeuralQueryBuilder level, it will work. These content_type will be removed when requests come to ml-commons accessor level if it's symmetric model. But it put extra/unnecessary overhead for symmetric QueryBuilder.
- Since ml-commons accessor is the place where we check `isAsymmetricModel`, I think 2nd option makes it clear and reduce the redundant doRewrite for all symmetric QueryBuilder

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -80,6 +87,9 @@
 @Log4j2
 public class MLCommonsClientAccessor {
     private final MachineLearningNodeClient mlClient;
+    private final Cache<String, Boolean> modelAsymmetryCache = CacheBuilder.newBuilder().maximumSize(1000).build();
```
junqiu-lei: Could we implement this cache layer to be generic for other models as well?

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -1063,4 +1089,105 @@ private List<Map<String, Object>> processDocumentHighlights(Object docHighlights
...
+     *
+     * Asymmetric models have different prefixes for queries and passages.
+     * Non-TextEmbeddingModels are considered symmetric (return false).
+     *
+     * @param model The ML model to check
+     * @return true if model has query or passage prefix, false otherwise
+     */
+    private boolean isAsymmetricModel(MLModel model) {
+        MLModelConfig modelConfig = model.getModelConfig();
+        if (!(modelConfig instanceof TextEmbeddingModelConfig textEmbeddingModelConfig)) {
```
junqiu-lei: Does remote and local model has the same conditional check logic?

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
fen-qin: updated in the PR description, I've included two new fields into inferenceRequest
- mlAlgoParas: ML algorithms params need for symmetric or asymmetric
- embeddingContentType: as indicator to determine if the request coming from ingestion or query

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
fen-qin: Updated revision:
- for symmetric, return `inferenceRequest.getMLAlgoParams()` 
- for asymmetric, cast MLAlgorithm into [AsymmetricTextEmbeddingParameters](https://github.com/opensearch-project/ml-commons/blob/66bcde78cbb353298f1e633cc68198259595da4a/common/src/main/java/org/opensearch/ml/common/input/parameter/textembedding/AsymmetricTextEmbeddingParameters.java) with everything from `inferenceRequest.getMLAlgoParams()` along with `inferenceRequest.getEmbeddingContentType()`

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
fen-qin: I reviewed the OpenSearch documentation about asymmetric semantic search [here](https://docs.opensearch.org/latest/tutorials/vector-search/semantic-search/semantic-search-asymmetric/). 
- The asymmetric model is registered using a local zip file
- The model_config for asymmetric models differs from symmetric models by requiring two additional parameters:
  - query_prefix
  - passage_prefix
- I couldn't find any pre-built remote models specifically for asymmetric search. Please let me know if you happen to know any remote use cases and examples. 

```
POST /_plugins/_ml/models/_register
{
    "name": "e5-small-onnx",
    "version": "1.0.0",
    "description": "Asymmetric multilingual-e5-small model",
    "model_format": "ONNX",
    "model_group_id": "your_group_id",
    "model_content_hash_value": "your_model_zip_content_hash_value",
    "model_config": {
        "model_type": "bert",
        "embedding_dimension": 384,
        "framework_type": "sentence_transformers",
        "query_prefix": "query: ",
        "passage_prefix": "passage: ",
        "all_config": "{ \"_name_or_path\": \"intfloat/multilingual-e5-small\", \"architectures\": [ \"BertModel\" ], \"attention_probs_dropout_prob\": 0.1, \"hidden_size\": 384, \"num_attention_heads\": 12, \"num_hidden_layers\": 12, \"tokenizer_class\": \"XLMRobertaTokenizer\" }"
    },
    "url": "http://localhost:8080/intfloat-multilingual-e5-small-onnx.zip"
}
```

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -1063,4 +1054,111 @@ private List<Map<String, Object>> processDocumentHighlights(Object docHighlights
...
+            throw new IllegalArgumentException("embeddingContentType must be set to either QUERY or PASSAGE for asymmetric models");
+        }
+
+        AsymmetricTextEmbeddingParameters.EmbeddingContentType mlContentType = contentType == EmbeddingContentType.QUERY
+            ? AsymmetricTextEmbeddingParameters.EmbeddingContentType.QUERY
+            : AsymmetricTextEmbeddingParameters.EmbeddingContentType.PASSAGE;
+
+        MLAlgoParams presetParams = inferenceRequest.getMlAlgoParams();
+        AsymmetricTextEmbeddingParameters.AsymmetricTextEmbeddingParametersBuilder builder = presetParams != null
+            ? ((AsymmetricTextEmbeddingParameters) presetParams).toBuilder()
```
heemin32: We might need to check if inferenceRequest.getMlAlgoParams is `AsymmetricTextEmbeddingParameters` or not and throw exception appropriately.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -1063,4 +1054,111 @@ private List<Map<String, Object>> processDocumentHighlights(Object docHighlights
...
+        if (!isAsymmetric) {
+            return inferenceRequest.getMlAlgoParams();
+        }
+
+        EmbeddingContentType contentType = inferenceRequest.getEmbeddingContentType();
+        if (contentType == null) {
+            throw new IllegalArgumentException("embeddingContentType must be set to either QUERY or PASSAGE for asymmetric models");
+        }
+
+        AsymmetricTextEmbeddingParameters.EmbeddingContentType mlContentType = contentType == EmbeddingContentType.QUERY
```
heemin32: We can make a method in `EmbeddingContentType` to return corresponding AsymmetricTextEmbeddingParameters.EmbeddingContentType type.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -1063,4 +1054,111 @@ private List<Map<String, Object>> processDocumentHighlights(Object docHighlights
...
+     *     Returns pre-set parameters if provided, with embeddingContentType overridden from request,
+     *     otherwise creates new AsymmetricTextEmbeddingParameters with appropriate content type (QUERY or PASSAGE)
+     *
+     * @param isAsymmetric Whether the model is asymmetric (has different query/passage prefixes)
+     * @param inferenceRequest The inference request containing content type and any pre-set parameters
+     * @return MLAlgoParams from request or generated for asymmetric models
+     */
+    private MLAlgoParams createMLAlgoParams(boolean isAsymmetric, InferenceRequest inferenceRequest) {
+        if (!isAsymmetric) {
+            return inferenceRequest.getMlAlgoParams();
```
heemin32: We might need to throw exception if MLAlgoParams is `AsymmetricTextEmbeddingParameters`

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
heemin32: I think this is good point. Semantic search and sentence highlighting also read model metadata. We might be able to create single method to read it.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
junqiu-lei: Let's make sure if there are remote model available to use in this feature. In some prod cases, users are encouraged to apply with remote model for better performance. So we need consider both local and remote cases to handle in the ml client accessor.

The function names could be different at [here](https://github.com/opensearch-project/ml-commons/blob/main/common/src/main/java/org/opensearch/ml/common/FunctionName.java#L15).

The model meta data(modelAsymmetry) check also need to be considered for both remote and local model.

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
fen-qin: updated to generic model cache

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
fen-qin: I've added customized remote sagemaker model for asymmetric use case
- https://github.com/opensearch-project/opensearch-py-ml/pull/587


updated the PR to support both local and remote models

src/test/resources/processor/UploadAsymmetricModelRequestBody.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "traced_small_model",
```
junqiu-lei, line 2: I think we also need introduce remote model integ tests for this feature.

Maybe you can add in this or separate PR.

fen-qin: Agree. Added integration tests with torchserve remote models. 
Validate the changes with
```
# Start and setup everything
./src/test/resources/remote-models/torchserve/scripts/run.sh lifecycle setup

# Run integration tests
./gradlew integTest --tests="*AsymmetricLocalModelIT*" -Dtests.security.manager=false
./gradlew integTest --tests="*AsymmetricRemoteModelIT*" -Dtests.security.manager=false
```

junqiu-lei: We have `./gradlew remoteModelIntegTest` for remote model only integ test, so that `./gradlew integTest` won't trigger it.

You can follow: https://github.com/opensearch-project/neural-search/blob/main/DEVELOPER_GUIDE.md#running-tests-with-remote-models

And finally your test should be picked in the remote model integ test CI as well, like: https://github.com/opensearch-project/neural-search/actions/runs/19308732819/job/55223053982?pr=1605

fen-qin: yes. you're right. trying to run with `remoteModelIntegTest` so it can setup and stop torchserve in one command. Thanks, Junqiu. Here is the screenshot of all remote model integ tests results

Screenshot 2025-11-12 at 11 28 50 AM

junqiu-lei: I saw the CI model server memory usage is full after we onboarding the new model :
```

[INFO] ✓ Batch inference successful!
[INFO] Memory checkpoint - After batch inference:
  Memory: 1.995GiB / 2GiB
  Memory %: 99.77%
  CPU %: 0.06%

=== Memory Usage Summary ===
Memory: 1.995GiB / 2GiB
Memory %: 99.77%
CPU %: 0.05%
===========================
node{::remoteModelIntegTest-0}: ports files took 0.0 seconds
Starting to wait for cluster to form
```

For long term stability and avoid OOM, maybe increase to 3g here: https://github.com/opensearch-project/neural-search/blob/main/src/test/resources/remote-models/torchserve/scripts/run.sh#L107

junqiu-lei (approved):

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -232,30 +212,56 @@ private <T> void retryableInference(
...
+                }
             }
         }
         return vector;
     }
 
+    /**
+     * Extracts vectors from asymmetric remote embedding model response format.
+     * Handles the simplified format used by asymmetric E5 remote embedding models: [[emb1], [emb2], [emb3]]
+     * This format is specific to our SageMaker E5 asymmetric model implementation.
```
heemin32: Remove `our`

fen-qin: sure. let me clean up the comments once again. thanks, Heemin

PR #1610, opened 2025-10-09
[Agentic Search] Fix return code and sanitize agent id

owaiskazi19:
### Description
- Graceful error handling to return 4xx when incorrect agent id
- Sanitize agent id length and pattern check

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/AgenticContextResponseProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/AgenticContextResponseProcessor.java
@@ -60,40 +60,9 @@ public class AgenticContextResponseProcessor implements SearchResponseProcessor 
             return response;
         }
 
-        Object agentStepsSummary = requestContext.getAttribute(AGENT_STEPS_FIELD_NAME);
-        Object memoryId = requestContext.getAttribute(MEMORY_ID_FIELD_NAME);
-        Object dslQuery = requestContext.getAttribute(DSL_QUERY_FIELD_NAME);
-
-        if (agentStepsSummary != null && !(agentStepsSummary instanceof String)) {
-            log.error("Invalid type for agent_steps_summary: expected String, got {}", agentStepsSummary.getClass().getSimpleName());
-            agentStepsSummary = null;
-        }
-
-        if (memoryId != null && !(memoryId instanceof String)) {
-            log.error("Invalid type for memory_id: expected String, got {}", memoryId.getClass().getSimpleName());
-            memoryId = null;
-        }
-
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/neuralsearch/processor/AgenticContextResponseProcessor.java
```diff
@@ -57,21 +60,23 @@ public SearchResponse processResponse(SearchRequest request, SearchResponse resp
...
-        Object memoryId = requestContext.getAttribute("memory_id");
-        Object dslQuery = requestContext.getAttribute("dsl_query");
+        Object agentStepsSummary = requestContext.getAttribute(AGENT_STEPS_FIELD_NAME);
+        Object memoryId = requestContext.getAttribute(MEMORY_ID_FIELD_NAME);
+        Object dslQuery = requestContext.getAttribute(DSL_QUERY_FIELD_NAME);
 
         // Validate types when attributes are present
         if (agentStepsSummary != null && !(agentStepsSummary instanceof String)) {
-            throw new RuntimeException("agent_steps_summary must be a String, but got: " + agentStepsSummary.getClass().getSimpleName());
+            throw new IllegalArgumentException(
```
bzhangam: Should we fail fast in the translator processor? Besides are users able to fix the error? If no I don't think we should throw 4xx error.

owaiskazi19: Good point. Pushed the change

src/main/java/org/opensearch/neuralsearch/processor/AgenticContextResponseProcessor.java
```diff
@@ -57,21 +60,23 @@ public SearchResponse processResponse(SearchRequest request, SearchResponse resp
             return response;
         }
 
-        Object agentStepsSummary = requestContext.getAttribute("agent_steps_summary");
-        Object memoryId = requestContext.getAttribute("memory_id");
-        Object dslQuery = requestContext.getAttribute("dsl_query");
+        Object agentStepsSummary = requestContext.getAttribute(AGENT_STEPS_FIELD_NAME);
```
bzhangam: I find in the translator processor we are sure these fields are string and then set them in the requestContext so there is no need to validate if they are string again here. We also can consider to set a DTO directly in the context to make it more clear.

owaiskazi19: Removed the type check and removed the tests associated

PR #1538, opened 2025-09-01
SEISMIC: Warm Up & Clear Cache

zirui-song-18:
### Description
This is a PR of SEISMIC feature. In this PR, we submit related classes and UTs of `warmUp` and `clearCache` operations. This PR aims at addressing this [LLD](https://github.com/opensearch-project/neural-search/issues/1523) and memory management [design](https://github.com/opensearch-project/neural-search/issues/1444).

#### Recommended order to review this PR:
- NeuralSparseIndexShard: Direct operation of warm up and clear cache
- RestNeuralSparseWarmupHandler & ClearCacheHandler: RestHandler for SEISMIC cache operation APIs
- transport: It includes `Action`, `Request`, `Response`, and `TransportAction` classes
- Unit Tests

Here are our plan for submitting PRs:

 - [x] Basic classes
 - [x] Clustering algorithms & field mapper
 - [x] Cache
 - [ ] Codec
 - [ ] Query
 - [ ] BWCs
 - [ ] Interfaces
     - [ ] Indexing and query
     - [ ] Cache: Memory stats
     - [ ] Cache: Circuit breaker
     - [x] Cache: Warm up and clear cache API

**We target version 3.3 release**.

#### Note
Within this PR, IT is not included as some dependencies are not merged into this main branch. We will supplement IT once dependencies are merged.

`SparseBinaryDocValuesPassThrough` is imported and used in this PR which is a dependency used. You can check its details and relations in #1537 

### Related Issues
RFC: #1335 
Low level design: #1523 
Memory management design: #1444 

### Check List
- [x] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/neural-search/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1630, opened 2025-10-28
[Fix]: Sparse ANN security risk

zirui-song-18:
### Description
Previously, there is a security risk of Sparse ANN's warmup/clear_cache APIs, where there will be validation of whether an index is sparse or not. This validation was located at REST layer inside `prepareRequest`, which is before authz against `NeuralSparseXXXXAction.INSTANCE`. Thanks to Suhas Suresh's valid findings and @cwperks 's suggestions. Now, I moved this validation process into Transport layer, especially after authz has been done.
#### Comparison
Previous: 'sparse' validation in REST request → before access authz → Possible index information leakage

Now: REST request -> access authz -> TransportAction -> 'sparse' validation in `checkRequestBlock` -> `shardOperation`

With this change, we can resolve this security risk.

### Related Issues
#1538 

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - [Agentic Search] Add conversation search support with agentic search ([#1626](https://github.com/opensearch-project/neural-search/pull/1626))
 
 ### Bug Fixes
-- [SEISMIC]: Resolve a security risk of Sparse ANN - Move 'index.sparse' validation from REST to transport layer
+- [SEISMIC]: Resolve a security risk of Sparse ANN - Move 'index.sparse' validation from REST to transport layer. [PR1630](https://github.com/opensearch-project/neural-search/pull/1630)
 
 ### Infrastructure
 - Onboard to s3 snapshots ([#1618](https://github.com/opensearch-project/neural-search/pull/1618))


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ## [Unreleased 3.x](https://github.com/opensearch-project/neural-search/compare/main...HEAD)
 
 ### Features
+- [SEISMIC]: Resolve a security risk of Sparse ANN - Move 'index.sparse' validation from REST to transport layer
```
chishui: Move to "Bug Fixes"?

src/main/java/org/opensearch/neuralsearch/transport/TransportUtils.java
```diff
@@ -0,0 +1,51 @@
...
+import org.opensearch.neuralsearch.sparse.SparseSettings;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Locale;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/**
+ * A Utils class for transport layer operations
```
chishui: nit: "A utility .."

src/main/java/org/opensearch/neuralsearch/transport/TransportUtils.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Validates that all provided indices are sparse indices
+     * @param state ClusterState
+     * @param concreteIndices Array of concrete index names
+     * @param operationName Name of the operation for error messages
+     * @throws OpenSearchStatusException if any index is not a sparse index
+     */
+    public static void validateSparseIndices(ClusterState state, String[] concreteIndices, String operationName) {
+        List<String> invalidIndexNames = Arrays.stream(concreteIndices).filter(indexName -> {
+            IndexMetadata indexMetadata = state.metadata().index(indexName);
```
chishui: Your previous logic looks cleaner 
```
        List<String> invalidIndexNames = Arrays.stream(indices).filter(index -> {
            String sparseIndexSetting = Optional.ofNullable(clusterService)
                .map(ClusterService::state)
                .map(ClusterState::metadata)
                .map(metadata -> metadata.getIndexSafe(index))
                .map(IndexMetadata::getSettings)
                .map(settings -> settings.get(SPARSE_INDEX))
                .orElse(null);
```

CHANGELOG.md
zirui-song-18: Ack

src/main/java/org/opensearch/neuralsearch/transport/TransportUtils.java
zirui-song-18: Ack

src/main/java/org/opensearch/neuralsearch/transport/TransportUtils.java
zirui-song-18: Ack

src/main/java/org/opensearch/neuralsearch/transport/TransportUtils.java
```diff
@@ -0,0 +1,54 @@
...
+     * @param state ClusterState
+     * @param concreteIndices Array of concrete index names
+     * @param operationName Name of the operation for error messages
+     * @throws OpenSearchStatusException if any index is not a sparse index
+     */
+    public static void validateSparseIndices(ClusterState state, String[] concreteIndices, String operationName) {
+        List<String> invalidIndexNames = Arrays.stream(concreteIndices).filter(indexName -> {
+            Boolean isSparseIndex = Optional.ofNullable(state)
+                .map(ClusterState::metadata)
+                .map(metadata -> metadata.index(indexName))
```
chishui, line 34: Question: I saw you switch from safeIndex to index, have you checked the differences, and why one replaces the other?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - [Agentic Search] Add conversation search support with agentic search ([#1626](https://github.com/opensearch-project/neural-search/pull/1626))
 
 ### Bug Fixes
+- [SEISMIC]: Resolve a security risk of Sparse ANN - Move 'index.sparse' validation from REST to transport layer
```
yuye-aws: include the PR link

src/main/java/org/opensearch/neuralsearch/transport/TransportUtils.java
zirui-song-18: Let's take a look at the code difference:
```
    public IndexMetadata index(String index) {
        return (IndexMetadata)this.indices.get(index);
    }

    public IndexMetadata getIndexSafe(Index index) {
        IndexMetadata metadata = this.index(index.getName());
        if (metadata != null) {
            if (metadata.getIndexUUID().equals(index.getUUID())) {
                return metadata;
            } else {
                String var10005 = index.getUUID();
                throw new IndexNotFoundException(index, new IllegalStateException("index uuid doesn't match expected: [" + var10005 + "] but got: [" + metadata.getIndexUUID() + "]"));
            }
        } else {
            throw new IndexNotFoundException(index);
        }
    }
```
We can notice that `getIndexSafe` will throw `IndexNotFoundException`. With the new validation position, already passing authz, there should be no chance of `IndexNotFoundException`, and this chain calling way should be safe enough to handle null situation. Besides, `getIndexSafe` can bring with more overhead.

One valuable point of `getIndexSafe` might be its UUID comparison, this will be helpful when an index owns the same name but different UUID. Do you think whether we need this check? Maybe in reconstruct scenario? (First delete, then create a new index with same name)

CHANGELOG.md
zirui-song-18: Ack. Will add it.

src/main/java/org/opensearch/neuralsearch/transport/TransportUtils.java
zirui-song-18: Also another factor is that here we use `String` of indices, not `Index` objects.

chishui (approved):

PR #1647, opened 2025-11-06
Fix for Hybrid Search with Collapse when number of shards are greater than number of documents.

vibrantvarun:
### Description
Hybrid Search when used with collapse throws an error when number of documents are lower than number of shards. 
When there are no documents in the shard then the indexReader.numDocs returns 0. This eventually makes [totalNumDocs](https://github.com/opensearch-project/neural-search/blob/3.3/src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java#L72) as 0 and numHits as also 0. Therefore, later when the [condition of topNGroups is validated in HybridCollapseCollector](https://github.com/opensearch-project/neural-search/blob/3.3/src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java#L75) it throws as error because numHits are set as topNGroups.

With this PR, we basically do not block the operation of collapse on a shard when topNGroups are 0. As the shard does not have any document the collapse operation will become a NoOP on that shard.

Testing done for Single Shard, multiple shards.
Also tested when size is sent 0.
Also tested on index which 0 documents.

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - [Agentic Search] Extract agent summary based on models ([#1633](https://github.com/opensearch-project/neural-search/pull/1633))
 
 ### Bug Fixes
-- [SEISMIC]: Resolve a security risk of Sparse ANN - Move 'index.sparse' validation from REST to transport layer. [PR1630](https://github.com/opensearch-project/neural-search/pull/1630)
-- [SEISMIC IT]: Fix some failed IT cases
+- [SEISMIC]: Resolve a security risk of Sparse ANN - Move 'index.sparse' validation from REST to transport layer. ([#1630](https://github.com/opensearch-project/neural-search/pull/1630))
+- [SEISMIC IT]: Fix some failed IT cases ([#1649](https://github.com/opensearch-project/neural-search/pull/1649))
 - [Hybrid Search]: Fix for hybrid search collapse bug when there are no documents in a shard ([#1647](https://github.com/opensearch-project/neural-search/pull/1647))
 
 ### Infrastructure


src/test/java/org/opensearch/neuralsearch/e2e/HybridCollapseIT.java
```diff
@@ -33,12 +33,27 @@ public class HybridCollapseIT extends BaseNeuralSearchIT {
...
     public void setUp() throws Exception {
         super.setUp();
-        createTestIndex();
-        indexTestDocuments();
         createSearchPipeline(SEARCH_PIPELINE, "min_max", "arithmetic_mean", Map.of());
     }
 
-    public void testCollapse_whenE2E_thenSuccessful() {
+    public void testCollapse_withSingleShard_thenSuccessful() {
+        createTestIndex(1);
```
martin-gaievski: can you use constant and name it in a way it's clear we are passing number of shards, like NUMBER_OF_SHARDS_ONE and NUMBER_OF_SHARDS_FIVE. Alternatively you can create a wrapper method that will set required number of shards, like
```
createTestIndexWithFiveShards() {
   createTestIndex(5);
}
```

src/test/java/org/opensearch/neuralsearch/e2e/HybridCollapseIT.java
```diff
@@ -67,7 +82,7 @@ public void testCollapse_whenE2E_thenSuccessful() {
         assertTrue(isCollapseDuplicateRemoved(searchResponse.toString(), collapseDuplicate));
     }
 
-    public void testCollapse_whenE2E_andSortEnabled_thenSuccessful() {
+    private void testCollapse_whenE2E_andSortEnabled_thenSuccessful() {
```
martin-gaievski, line 87: why we are changing this?

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -72,30 +72,26 @@ public class HybridCollapsingTopDocsCollector<T> implements HybridSearchCollecto
         this.groupSelector = groupSelector;
         this.collapseField = collapseField;
         this.sort = groupSort;
-        if (topNGroups < 1) {
```
martin-gaievski, line 75: we have dropped this validation, is this intentional, what's the reason?

vibrantvarun: The whole purpose of this PR is to remove this condition because when numHits becomes 0, then topNGroups also becomes 0 and throws the error which we are facing while running integ test. By removing this condition, the normal workflow will run and it will handle the case when results are not present.

src/test/java/org/opensearch/neuralsearch/e2e/HybridCollapseIT.java
vibrantvarun: If they are public then they will also run as tests. I am already calling these method from the common method here https://github.com/opensearch-project/neural-search/pull/1647/files#diff-49f453d9424e2fbde9ba4f3bf555f8b7d71c0512ce6ee0359a903737fb038928R47. As it saves memory by not creating the same index and ingest same documents thrice.

heemin32: I don't think combining multiple tests into a single one is ideal, because it reduces the granularity of the test scope. When multiple behaviors are validated in one test, it becomes harder to understand what specifically failed and to maintain the test over time.

Instead we could do
```
createTestIndex(rand() % 1 ? NUMBER_OF_SHARDS_FIVE : NUMBER_OF_SHARDS_ONE);
```
Or do we really need a test with single shard?

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
heemin32: Does it mean `topNGroups must be >= 1 (got " + topNGroups + ")` is not correct? topNGroup can be zero or below zero?

vibrantvarun: Yes @heemin32 . It is not correct.

src/test/java/org/opensearch/neuralsearch/e2e/HybridCollapseIT.java
vibrantvarun: But if we are doing the exact same index with exact same documents three times earlier, and combine them in caller method reduces the task from 3 to 1.

Yes @heemin32 one shard case is needed because in Hybrid query the process of search is different in one shard when compared to multiple shard. The normalization process at the end in single shard scenario

heemin32: >But if we are doing the exact same index with exact same documents three times earlier, and combine them in caller method reduces the task from 3 to 1.

If you really want to optimize that index creation, you can create once using `@BeforeClass` annotation.

vibrantvarun: @heemin32 I think the current way is much better as we are achieving the same goal. Also if use @BeforeClass then we have use @AfterClass as well for cleanup.

heemin32: @martin-gaievski wdyt?

martin-gaievski: I see pros and cons for each of the approaches, for a single index per test class main concern is - index state can be changes inside the method. Because we are not changing index content in this particular test  I'm inclining towards a single creation and deletion of the index with `before` and `after` annotations respectively. We can create/delete both indices (with 1 and 5 shards).

PR #1655, opened 2025-11-10
[SEISMIC IT] fix IT failing

### LINKED ISSUE #1653: [BUG] NeuralSparseCacheOperationIT multi shard tests could fail if test env with non data nodes
### What is the bug?
NeuralSparseCacheOperationIT multi shard tests could fail if test env with setup like three data nodes and three master nodes as the tests expect memory usage to change after calling warmup and clearcache API. But with master nodes, they don't have seismic running on it and those API won't bring a change to memory usage.

Run integration tests against remote clusters with 3 data nodes and 3 master nodes.

integration should pass with different cluster configuration.

_Operating system, version._

zirui-song-18:
### Description
This PR further fixed two failed IT cases when there are dedicated Master/Data nodes.

### Related Issues
Resolves #1653 

### Check List

### DIFF
--- a/src/test/java/org/opensearch/neuralsearch/sparse/NeuralSparseCacheOperationIT.java
+++ b/src/test/java/org/opensearch/neuralsearch/sparse/NeuralSparseCacheOperationIT.java
@@ -13,6 +13,7 @@ import org.opensearch.client.Response;
 import org.opensearch.client.ResponseException;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.common.xcontent.XContentType;
+import org.opensearch.neuralsearch.SparseTestCommon;
 
 import java.util.List;
 import java.util.Map;
@@ -188,7 +189,7 @@ public class NeuralSparseCacheOperationIT extends SparseBaseIT {
                 nodesWithMemoryIncrease++;
             }
         }
-        assertEquals("Only data nodes should have memory increase", nodesWithMemoryIncrease, getDataNodeCount());
+        assertEquals("Only data nodes should have memory increase", nodesWithMemoryIncrease, SparseTestCommon.getDataNodeCount(client()));
     }
 
     /**
@@ -230,7 +231,7 @@ public class NeuralSparseCacheOperationIT extends SparseBaseIT {
                 nodesWithMemoryDecrease++;
             }
         }
-        assertEquals("Only data nodes should have memory decrease", nodesWithMemoryDecrease, getDataNodeCount());
+        assertEquals("Only data nodes should have memory decrease", nodesWithMemoryDecrease, SparseTestCommon.getDataNodeCount(client()));
     }
 
     /**


src/test/java/org/opensearch/neuralsearch/sparse/NeuralSparseCacheOperationIT.java
```diff
@@ -182,9 +182,13 @@ public void testWarmUpMultiShardReplicasCache() {
...
         assertTrue("Memory usage should increase after warm up", afterWarmUpSparseMemoryUsageSum > originalSparseMemoryUsageSum);
         assertEquals(originalSparseMemoryUsageStats.size(), afterWarmUpSparseMemoryUsageStats.size());
+        int nodesWithMemoryIncrease = 0;
         for (int i = 0; i < originalSparseMemoryUsageStats.size(); i++) {
-            assertTrue(afterWarmUpSparseMemoryUsageStats.get(i) > originalSparseMemoryUsageStats.get(i));
+            if (afterWarmUpSparseMemoryUsageStats.get(i) > originalSparseMemoryUsageStats.get(i)) {
+                nodesWithMemoryIncrease++;
+            }
         }
+        assertEquals("At least data nodes should have memory increase", nodesWithMemoryIncrease, getDataNodeCount());
```
yuye-aws: `At least data nodes should have memory increase` is guaranteed by `Memory usage should increase after warm up`. How about simply discarding the assertions

zirui-song-18: I think this assertion is to make sure every data nodes' memory work in our expected way. `Memory usage should increase after warm up` can only assert total memory.

yuye-aws: Sorry, I missed some context. After revisiting the recent PRs, I just have a simple question: How do you ensure every node has a memory increase?

zirui-song-18: Each data node will have memory change. That's why I modified here in this way. If we do not use `getDataNodeCount`, we might count Master node by mistake, then this will fail.

yuye-aws: No worries. Resolving this thread now.

bzhangam: How can we know the master node memory will not increase? Can we check the node role when we do this check?

zirui-song-18: This assertion here is to compare the number of data nodes and the number of nodes with memory increases. If they are equal, which means that only data nodes increase memory.

owaiskazi19: +1 to @bzhangam . We can also check for node.roles to be assure that only data nodes has increased memory

owaiskazi19:
```suggestion
        assertEquals("Only data nodes should have memory increase", nodesWithMemoryIncrease, getDataNodeCount());
```

src/test/java/org/opensearch/neuralsearch/sparse/NeuralSparseCacheOperationIT.java
```diff
@@ -220,9 +224,13 @@ public void testClearMultiShardReplicasCache() {
...
         assertTrue("Memory usage should decrease after clear cache", afterClearCacheSparseMemoryUsageSum < originalSparseMemoryUsageSum);
         assertEquals(originalSparseMemoryUsageStats.size(), afterClearCacheSparseMemoryUsageStats.size());
+        int nodesWithMemoryDecrease = 0;
         for (int i = 0; i < originalSparseMemoryUsageStats.size(); i++) {
-            assertTrue(afterClearCacheSparseMemoryUsageStats.get(i) < originalSparseMemoryUsageStats.get(i));
+            if (afterClearCacheSparseMemoryUsageStats.get(i) < originalSparseMemoryUsageStats.get(i)) {
+                nodesWithMemoryDecrease++;
+            }
         }
+        assertEquals("At least data nodes should have memory decrease", nodesWithMemoryDecrease, getDataNodeCount());
```
owaiskazi19:
```suggestion
        assertEquals("Only data nodes should have memory decrease", nodesWithMemoryDecrease, getDataNodeCount());
```

owaiskazi19: Same as above we can check for node.roles here to verify

zirui-song-18: Ack. Now this assertion makes sure that only data nodes behave in the expected way. It should be fine.

src/test/java/org/opensearch/neuralsearch/sparse/NeuralSparseCacheOperationIT.java
yuye-aws: This is a bit tricky here because the get memory stats API will not return node.roles. A workaround is "node_id". You can call cat nodes to obtain node_id for master nodes and data nodes.

zirui-song-18: I think latest version is a good choice. `getNodeCount()` directly return how many data nodes we have, where we only execute cache operations on those data nodes. Then in these two cases, we count how many nodes' memory increase/decrease, to assert whether that equals to the number of data nodes.

CHANGELOG.md
```diff
@@ -14,7 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Bug Fixes
 - [SEISMIC]: Resolve a security risk of Sparse ANN - Move 'index.sparse' validation from REST to transport layer. ([#1630](https://github.com/opensearch-project/neural-search/pull/1630))
-- [SEISMIC IT]: Fix some failed IT cases ([#1649](https://github.com/opensearch-project/neural-search/pull/1649))
+- [SEISMIC IT]: Fix some failed IT cases ([#1649](https://github.com/opensearch-project/neural-search/pull/1649), [#1655](https://github.com/opensearch-project/neural-search/pull/1655))
```
owaiskazi19: We should clear the complete changelog since we have released the enhancement and bug fixes. And add this as a new entry for the next release

zirui-song-18: Thanks for that advice. Now I cleaned all those released items, just left this new change.

CHANGELOG.md
```diff
@@ -8,17 +8,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
-- [Agentic Search] Extract agent summary based on models ([#1633](https://github.com/opensearch-project/neural-search/pull/1633))
 
 ### Bug Fixes
-- [SEISMIC]: Resolve a security risk of Sparse ANN - Move 'index.sparse' validation from REST to transport layer. ([#1630](https://github.com/opensearch-project/neural-search/pull/1630))
-- [SEISMIC IT]: Fix some failed IT cases ([#1649](https://github.com/opensearch-project/neural-search/pull/1649))
-- [Hybrid Search]: Fix for hybrid search collapse bug when there are no documents in a shard ([#1647](https://github.com/opensearch-project/neural-search/pull/1647))
+- [SEISMIC IT]: Fix some failed IT cases ([#1655](https://github.com/opensearch-project/neural-search/pull/1655))
 
 ### Infrastructure
-- Onboard to s3 snapshots ([#1618](https://github.com/opensearch-project/neural-search/pull/1618))
```
owaiskazi19, line 21: We don't have to remove this. This was added for the next release. Please add it back

zirui-song-18: Sure. Ack

yuye-aws:
The CI failed due to issue https://github.com/opensearch-project/neural-search/issues/1660. Will merge this PR after the fix in https://github.com/opensearch-project/neural-search/pull/1657

yuye-aws:
Hi @zirui-song-18 , I have just merged the BWC test PR. Can you rebase main to unblock the CI failure?

PR #1649, opened 2025-11-06
Sparse ANN: Fix failed IT cases

zirui-song-18:
### Description
This PR is to fix some failed IT cases.

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/neural-search/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1663, opened 2025-11-19
Add role assignment multi-node integ testing in CI

chishui:
### Description
Add role assignment multi-node integ testing in CI. 

Tests before this fix #1649 would fail with new multi node testing with role assignment which means it can discover these potential issues in tests.

### Related Issues
Resolves #[1656]

### Check List

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -137,7 +137,7 @@ jobs:
           java: [ 21 ]
           os: [ ubuntu-latest ]
           configureNodeRoles: [ false, true ]
-      name: Multi-Node Integ Test JDK${{ matrix.java }}, roles=${{ matrix.configureNodeRoles }}
+      name: Multi-Node Integ Test JDK${{ matrix.java }}, With Role Assignment ${{ matrix.configureNodeRoles }}
       runs-on: ${{ matrix.os }}
       steps:
         - uses: actions/checkout@v4


build.gradle
```diff
@@ -416,6 +417,19 @@ testClusters.integTest {
     plugin(project.tasks.bundlePlugin.archiveFile)
     // Cluster shrink exception thrown if we try to set numberOfNodes to 1, so only apply if > 1
     if (_numNodes > 1) numberOfNodes = _numNodes
+
+    // Configure node roles if enabled
+    if (_configureNodeRoles && _numNodes > 1) {
```
owaiskazi19, line 422: We don't have to check for `_numNodes` now since we are passing the `_configureNodeRoles` as true

build.gradle
```diff
@@ -416,6 +417,19 @@ testClusters.integTest {
...
+        nodes.eachWithIndex { node, index ->
+            if (index == 0) {
+                node.setting('node.roles', '["cluster_manager"]')
+            } else if (index == 1) {
+                node.setting('node.roles', '["cluster_manager", "data", "ingest", "ml"]')
```
owaiskazi19, line 427: We should keep this as `data` since we already covered `cluster_manager` above

build.gradle
```diff
@@ -416,6 +417,19 @@ testClusters.integTest {
...
+            } else {
+                node.setting('node.roles', '["data", "ingest"]')
```
owaiskazi19, line 429: Just ingest here. So that we can be sure that we are covering all the different type of nodes
```suggestion
                node.setting('node.roles', '["ingest"]')
```

build.gradle
chishui: It doesn't hurt to check it and in case _configureNodeRoles is miss used.

build.gradle
chishui: We need at least two cluster_manager roles in a 3 nodes cluster.

build.gradle
chishui: The current configuration covers 2/3 data nodes, 2/3 ingest nodes, 2/3 master nodes, 1/3 ml nodes. It should be fine to make it different comparing to default configuration.

chishui:
@owaiskazi19 Could you please review this again? Thanks!

build.gradle
```diff
@@ -416,6 +417,19 @@ testClusters.integTest {
     plugin(project.tasks.bundlePlugin.archiveFile)
     // Cluster shrink exception thrown if we try to set numberOfNodes to 1, so only apply if > 1
     if (_numNodes > 1) numberOfNodes = _numNodes
+
+    // Configure node roles if enabled
+    if (_configureNodeRoles && _numNodes > 1) {
+        nodes.eachWithIndex { node, index ->
+            if (index == 0) {
+                node.setting('node.roles', '["cluster_manager"]')
```
yuye-aws, line 425: The node with index 0 does not have the `data` role type. When we create an index with 3 primary shards, I'm afraid there would be a node with 2 shards. This is not expected.

chishui: A node can host multiple shards, there is only limitation that a replica can't be hosted on the same node with its primary shard. With 2 nodes and 3 shards and 1 replica, we can have shard1,2 and replica3 on data node1, and shard3 and replica1,2 on data node2

yuye-aws: By making this change, the testing behavior would be different. Without this PR, all 3 nodes can host 1 primary shard. I don't want this PR to change the environment behavior

chishui: The point IS to verify all tests can pass under a different setup. Our tests shouldn't only verify a "happy path", it should also verify a path used by real situation.

yuye-aws: In practice, for the node where index == 0, the real situation could be either cluster_manager role exclusively (case 1) or both cluster_manager and data role (case 2). This PR is introducing "a different setup", changing **from case 2 to case 1**.

I totally understand that this PR is created in light of IT issues mentioned in https://github.com/opensearch-project/neural-search/pull/1649. How about making this PR **append-only**, i.e., introduce a new CI test case where the first node is cluster_manager role exclusively?

build.gradle
zirui-song-18: > We need at least two cluster_manager roles in a 3 nodes cluster

Just wonder if there is any source claiming this limit? I once tried using only 1 `cluster_manager` in a 3-node cluster ("localhost:9300", "localhost:9301", "localhost:9302"), it can work.

build.gradle
chishui: > In practice, for the node where index == 0, the real situation could be either cluster_manager role exclusively (case 1) or both cluster_manager and data role (case 2). This PR is introducing "a different setup", changing from case 2 to case 1.

This is a new setup, it doesn't change existing setup. I don't quite follow what you mean by case 2 to case 1, node1 only has cluster_manager. And why one is better than the other?

yuye-aws: Thanks. Resolving this thread now

build.gradle
chishui: Key Takeaways:
OpenSearch uses a strict majority (> 50%) quorum for cluster decisions
The VotingConfiguration class tracks which nodes can vote on cluster state changes
The hasQuorum() method enforces that more than half of the configured voting nodes must agree
For 3 nodes, you need 2 to form a quorum (2/3 > 50%)

https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/cluster/coordination/CoordinationMetadata.java#L408

owaiskazi19: Good point. 
> I once tried using only 1 cluster_manager in a 3-node cluster ("localhost:9300", "localhost:9301", "localhost:9302"), it can work.

It can work but the cluster would be unstable and would be stuck in no eligible cluster manager node errors if the only cluster manager nodes went down. Can cause writes and updates failure too

zirui-song-18: Thanks for that explanation!

chishui:
> > Roles are logged from here: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/node/Node.java#L601-L610
> 
> Yes, wanted to see logs around if this setup is working fine with different roles of node when invoked. Can share local logs too. Like
> 
> ```
> [integTest-0] node name [integTest-0], node ID [_8xLTPeMRcm2LC3J9IuHZg], cluster name [integTest], roles [ingest, remote_cluster_client, data, cluster_manager]
> ```

This is logs from local:
```
[2025-11-22T10:10:17,848][INFO ][o.o.n.Node               ] [integTest-0] node name [integTest-0], node ID [D5p6fBc7Tp-fLex827bgBA], cluster name [integTest], roles [cluster_manager]
[2025-11-22T10:10:40,156][INFO ][o.o.n.Node               ] [integTest-1] node name [integTest-1], node ID [6ryqzTIoS5mVszz0Xmz74Q], cluster name [integTest], roles [ingest, data, cluster_manager, ml]
[2025-11-22T10:10:59,740][INFO ][o.o.n.Node               ] [integTest-2] node name [integTest-2], node ID [xeUkALRgRWq48Od0oMNqxg], cluster name [integTest], roles [ingest, data]
```

owaiskazi19 (approved):
Thanks @chishui!

Issue #1664, opened 2025-11-19
Agentic Search: Support of `_source.excludes`


mateon01:
When executing an agentic query, the excludes in the _source are not applied.

Request:
```json
POST products/_search?search_pipeline=agentic-pipeline
{
  "explain": true,
  "_source": {
    "excludes": ["description_embedding"]
  },
  "query": {
    "agentic": {
      "query_text": "부드러운 필기감을 가진 볼팬"
    }
  }
}
```
Response : 
```json
{
  "took": 9353,
  "timed_out": false,
  "_shards": {
    "total": 3,
    "successful": 3,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 100,
      "relation": "eq"
    },
    "max_score": 1.2996008,
    "hits": [
      {
        "_index": "products",
        "_id": "43",
        "_score": 1.2996008,
        "_source": {
          "rating": 4.8,
          "description": "한국의 국민 볼펜. 부드러운 필기감.",
          "review_count": 15234,
          "created_at": "2024-05-20",
          "description_embedding": [
            -0.0629369,
            0.022355685,
            -0.009653951,
            -0.0049639,
            0.00020278986,
            -0.01612839,
            -0.034849353,
            -0.024070702,
            0.036264434,
            -0.049788043,
            0.013337143,
            -0.03353658,
...
```

PR #1669, opened 2025-11-24
[Agentic Search] Preserve source parameters for agentic query

owaiskazi19:
### Description
Preserve source parameters for agentic query like exclude, include

### Related Issues
Resolves https://github.com/opensearch-project/neural-search/issues/1664

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
@@ -158,8 +158,10 @@ public class AgenticQueryTranslatorProcessor extends AbstractProcessor implement
                         if (originalExtBuilders != null && !originalExtBuilders.isEmpty()) {
                             newSourceBuilder.ext(originalExtBuilders);
                         }
-                        // Preserve original request parameters
-                        preserveOriginalParameters(originalSourceBuilder, newSourceBuilder);
+                        // Preserve source field selection
+                        if (originalSourceBuilder != null && originalSourceBuilder.fetchSource() != null) {
+                            newSourceBuilder.fetchSource(originalSourceBuilder.fetchSource());
+                        }
                         request.source(newSourceBuilder);
                     }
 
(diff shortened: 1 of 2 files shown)


vibrantvarun (approved):
code looks good.

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
```diff
@@ -188,6 +191,66 @@ private void executeAgentAsync(
         }));
     }
 
+    private void preserveOriginalParameters(SearchSourceBuilder original, SearchSourceBuilder target) {
```
bzhangam: Probably it's better to use the stream to do this copy?

owaiskazi19: I intentionally avoided using streams here to keep the code more readable and straightforward. Since we’re just iterating and applying simple operations without any complex transformations or filtering, traditional loops are easier to follow and maintain. Using streams here wouldn’t add significant benefit and could make debugging slightly harder.

bzhangam: I think as long as we are good to copy over all the fields using the stream can make it cleaner and also easier to maintain. No need to add new code for new field.

chishui: 1. Can you not use `shallowCopy` function?
2. Do you need all those condition checks? Some fields for example with null by default, then doing null check and assigning null directly makes no difference, right? Asking because this function introduces many branches and UT doesn't seem to cover all branches, and to cover them all requires quite some efforts.

vibrantvarun: Actually, I think in the stream implementation we need to create a field type checker then create the checks accordingly. For example size, from , timeout are integers so there condition is not equal to -1. fetchSource(), storeFields are objects so there condition is != null. In stream also we have to first detect the type of the field and then validate it and then put it in target. So I don' there can be a generic method which will solve the problem. Also, the format of the response does not change. It is same since opensearch is in existence.

bzhangam: Why we need to check if the field exist or not. I think we can simply copy over all the info since here we just want to preserve the original search source and copy it to a new one to process.

PR #1701, opened 2025-12-16
Fix: memory usage track

yuye-aws:
### Description

Fix memory usage tracking during cache entry creation

### Related Issues
Resolves potential excessive warning logs from neural circuit breaker
```
[neural_search] New used memory 21474906 [20.4mb] for data of [seismic] would be larger than configured breaker: 21474836 [20.4mb], breaking
```
### Check List

### DIFF
--- a/src/test/java/org/opensearch/neuralsearch/sparse/AbstractSparseTestBase.java
+++ b/src/test/java/org/opensearch/neuralsearch/sparse/AbstractSparseTestBase.java
@@ -7,6 +7,7 @@ package org.opensearch.neuralsearch.sparse;
 import lombok.SneakyThrows;
 import org.apache.lucene.index.SegmentInfo;
 import org.junit.Before;
+import org.mockito.internal.util.MockUtil;
 import org.opensearch.neuralsearch.query.OpenSearchQueryTestCase;
 import org.opensearch.neuralsearch.sparse.accessor.SparseVectorReader;
 import org.opensearch.neuralsearch.sparse.cache.CacheKey;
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/neuralsearch/sparse/cache/MemMonitoredCache.java
```diff
@@ -53,7 +53,9 @@ public void onIndexRemoval(@NonNull CacheKey key) {
     protected T getOrCreate(@NonNull CacheKey key, @NonNull Function<CacheKey, T> creator) {
         return cacheMap.computeIfAbsent(key, k -> {
             T value = creator.apply(k);
-            CircuitBreakerManager.addWithoutBreaking(RamUsageEstimator.shallowSizeOf(k));
```
chishui, line 56: Is this line the only place which misses the globaltracker?

yuye-aws: Yes. After this PR, all **CircuitBreakerManager.addWithoutBreaking** call will use the global tracker

chishui: We'd better to add a mechanism to ensure CircuitBreakerManager.addWithoutBreaking can not be called without global tracker or make the global tracker threshold slightly lower than circuit breaker to avoid this issue to happen again. I'm OK to leave it to future PRs.

chishui (approved):

zhichao-aws (approved):

Issue #1696, opened 2025-12-15
[BUG] Hybrid Query with Collapse is not able to categorize on nested integer field


vibrantvarun:
### What is the bug?
The hybrid query with collapse is not able to categorize under a reference if the collapse is applied on integer field.

It throws following error and is not able to identify the collapse field.

```
{
  "error": {
    "root_cause": [],
    "type": "search_phase_execution_exception",
    "reason": "The phase has failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [],
    "caused_by": {
      "type": "illegal_argument_exception",
      "reason": "Could not determine collapse value type from input data"
    }
  },
  "status": 400
}
```

### How can one reproduce the bug?
1. Create a index with a following nested field

```
        "user": {
          "type": "nested",
          "properties": {
            "age": {
              "type": "integer"
            },
            "firstname": {
              "type": "keyword"
            },
            "lastname": {
              "type": "keyword"
            }
          }
        },
```
2. Index a few documents and apply collapse with inner hits on "user.age" field.
3. It will throw error of "Could not determine collapse value type from input data"

### What is the expected behavior?
Out of all the documents, it should take one field value as reference and add all inner hits under that.

For example if the documents are having ages 37, 41, 43. Then out of a 3 it will pick one age as reference and add other two in that to avoid ambiguity. 

Try the same thing with boolean query for understanding it better.

### What is your host/environment?
OS 3.2

Issue #1695, opened 2025-12-15
[BUG] Hybrid Query with Collapse and inner hits throws error "Could not determine collapse value type from input data"


vibrantvarun:
### What is the bug?
Collapse with inner hits when applied with Hybrid query fails when the collapse field value does not exist in the document with the following error

```
{
  "error": {
    "root_cause": [],
    "type": "search_phase_execution_exception",
    "reason": "The phase has failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [],
    "caused_by": {
      "type": "illegal_argument_exception",
      "reason": "Could not determine collapse value type from input data"
    }
  },
  "status": 400
}
```
Try the same subqueries with bool and must | should clause and it will work fine. 

### How can one reproduce the bug?
1. Ingest few documents and intentionally do not add value in some the documents in the collapsed field.
2. Try collapse alone.
3. Try collapse with inner hits

### What is the expected behavior?
The response should contain the hits under collapsed field. For the documents that do not contain value of the collapsed filed are categorized under null collapse value. 

### What is your host/environment?
OpenSearch 3.2

vibrantvarun:
@krisfreedain I am already working on this. In fact the PR is also merged.

Issue #1687, opened 2025-12-10
[BUG] Hybrid search with inner hits error "failed to expand hits" on 3.3.2


AdAhAHaH:
### Versions:
- OpenSearch 3.3.2
- Opensearch Dashboard 3.3.0
(Windows environment)

### Context:
I have a knn index and I need to do a hybrid search collapsing on a keyword field while also retrieving the inner hits. Recently there have been news regarding this matter and starting with the 3.2 release it was officially documented as a supported feature for the hybrid queries (link to the official documentation: [Retrieving inner hits for collapsed hybrid query results](https://docs.opensearch.org/latest/vector-search/ai-search/hybrid-search/collapse/#retrieving-inner-hits-for-collapsed-hybrid-query-results)).

### Bug:
When I try using the collapse with inner hits feature in combination with the hybrid query I get the error:
```
{
  "error": {
    "root_cause": [],
    "type": "search_phase_execution_exception",
    "reason": "failed to expand hits",
    "phase": "expand",
    "grouped": true,
    "failed_shards": [],
    "caused_by": {
      "type": "search_phase_execution_exception",
      "reason": "all shards failed",
      "phase": "query",
      "grouped": true,
      "failed_shards": [
        {
          "shard": 0,
          "index": "innerhits_expansion_error_index",
          "node": "fMKRQ9WpTxWJENS0O5hI6w",
          "reason": {
            "type": "e_o_f_exception",
            "reason": "read past EOF (pos=2147483647): MemorySegmentIndexInput(path=\"C:\\ZWeb\\OpenSearch\\opensearch\\data\\nodes\\0\\indices\\X3dJFJyZTfie9IdrxqrSOg\\0\\index\\_0.cfs\") [slice=_0.nvd] [slice=randomaccess]"
          }
        }
      ],
      "caused_by": {
        "type": "e_o_f_exception",
        "reason": "read past EOF (pos=2147483647): MemorySegmentIndexInput(path=\"<my_path>\") [slice=_0.nvd] [slice=randomaccess]",
        "caused_by": {
          "type": "e_o_f_exception",
          "reason": "read past EOF (pos=2147483647): MemorySegmentIndexInput(path=\"<my_path>\") [slice=_0.nvd] [slice=randomaccess]"
        }
      }
    }
  },
  "status": 500
}
```
The error only comes up when I try to also retrieve the inner hits. It seems that the expansion of the inner hits is failing for some reason.

### Steps to reproduce the bug:
```
[...]

PR #1702, opened 2025-12-16
Fix bugs in Hybrid Query with collapse 

vibrantvarun:
### Description
This PR contains fixes the following issues:

### Scenario 1 EOF exception

1. As per OpenSearch behaviour, for each collapsed hit it will fan out Boolean query containing hybrid query in must clause and inner hits query in filter clause on each shard.
2. During search, it will load DefaultBulkScorer for the Boolean query and for subquery it will load HybridQueryScorer.
3. Due to scorer being HybridQueryScorer and iterator being default docIdIterator, they both go out of sync.

As per the Hybrid Query design:
1. HQ can only be top level query clause. 
2. HybridQueryScorer needs HybridBulkScorer to execute it. However, in this bug it loads DefaultBulkScorer.

Fix:
When the Boolean query with Hybrid Query and Filter clause is sent, then transform the boolean query by remove hybrid clause and replacing it with boolean with should clause. See below:
The subquery 1 and subquery 2 are from hybrid queries clause.
```
bool
    must
        bool
           should
              subquery 1
              subquery 2
    filter
```
It will bring the scorer and the iterator in sync.

### Scenario 2 
If the collapsed field is not present in any document, then for that OpenSearch will fan out Boolean query containing hybrid query in must clause and inner hits query in `must_not` clause on each shard.

Currently, it throws an exception because it is not able to detect Boolean wth HQ and must_not clause.

### Scenario 3 Exception *class_cast_exception*
We see following issue. 
```
"reason": "class org.apache.lucene.search.TopDocs cannot be cast to class org.apache.lucene.search.TopFieldDocs (org.apache.lucene.search.TopDocs and org.apache.lucene.search.TopFieldDocs are in unnamed module of loader 'app')"
```
This issue arises when there are multiple segments in a shard and the topDocs are being merged in TopDocsMerger class. During the merging process, instead of creating CollapseTopFieldDocs it creates TopFieldDocs when collapse is applied. Due to this later in the normalization process, the above exception is thrown.

**Fix:**

Modify mergeScoreDocs method in TopDocsMerger to inculcate merging of collapseValues if collapse is applied. After getting mergeCollapseValues, create CollapseTopFieldDocs out of it.

### Scenario 4 Exception *Could not determine collapse value type from input data*
1. Index few document with the field `author`.
5. Index another document but without the field `author`.
As per OpenSearch behaviour, If the collapse is applied on a field and that field is missing from the document then that document is categorized under `null `group. When collapse + inner hits is applied with Hybrid query then the collapse fails with above exception because it does not categorize `null` group value. In this PR we fixed it by allowing null group value.

### Scenario 5 Exception *Could not determine collapse value type from input data*
If the collapse is applied on a nested integer field, then OpenSearch behaviour for other queries(except hybrid) is select a field value and put all inner hits under that. In short, if there is a ambiguity in deciding in which group the inner hits should be categorized then OpenSearch resolves by putting it in query hit. In hybrid query with collapse, it throws above exception because there is check if the collapse group value is either ByteRef or Long then do the collapse or throw exception. In this case, the group value will be `null`.

**Fix:**

In this PR, we have removed that check to allow `null `group.

### Related Issues
https://github.com/opensearch-project/neural-search/issues/1695
https://github.com/opensearch-project/neural-search/issues/1696
https://github.com/opensearch-project/neural-search/issues/1687

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/search/query/TopDocsMerger.java
+++ b/src/main/java/org/opensearch/neuralsearch/search/query/TopDocsMerger.java
@@ -6,8 +6,6 @@ package org.opensearch.neuralsearch.search.query;
 
 import com.google.common.annotations.VisibleForTesting;
 import lombok.RequiredArgsConstructor;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
 import org.apache.lucene.search.FieldDoc;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.SortField;
@@ -29,7 +27,6 @@ import org.opensearch.neuralsearch.search.query.HybridQueryScoreDocsMerger.Merge
  */
 @RequiredArgsConstructor
 public class TopDocsMerger {
-    private static final Logger log = LogManager.getLogger(TopDocsMerger.class);
     private HybridQueryScoreDocsMerger docsMerger;
     private SortAndFormats sortAndFormats;
     private CollapseContext collapseContext;
@@ -76,19 +73,25 @@ public class TopDocsMerger {
         TotalHits mergedTotalHits = getMergedTotalHits(source, newTopDocs);
 
         MergeResult mergeResult;
+        ScoreDoc[] sourceScoreDocs = source.topDocs.scoreDocs;
+        ScoreDoc[] newScoreDocs = newTopDocs.topDocs.scoreDocs;
+
         if (isCollapseEnabled()) {
-            CollapseTopFieldDocs sourceCollapseTopFieldDocs = (CollapseTopFieldDocs) source.topDocs;
-            CollapseTopFieldDocs newCollapseTopFieldDocs = (CollapseTopFieldDocs) newTopDocs.topDocs;
+            if (!(source.topDocs instanceof CollapseTopFieldDocs sourceCollapseTopFieldDocs)) {
+                throw new IllegalStateException("Collapse enabled but source TopDocs is not an instance of CollapseTopFieldDocs");
+            }
+            if (!(newTopDocs.topDocs instanceof CollapseTopFieldDocs newCollapseTopFieldDocs)) {
+                throw new IllegalStateException("Collapse enabled but new TopDocs is not an instance of CollapseTopFieldDocs");
+            }
             mergeResult = getMergedScoreDocs(
-                source.topDocs.scoreDocs,
-                newTopDocs.topDocs.scoreDocs,
+                sourceScoreDocs,
+                newScoreDocs,
                 sourceCollapseTopFieldDocs.collapseValues,
                 newCollapseTopFieldDocs.collapseValues
             );
         } else {
-            mergeResult = getMergedScoreDocs(source.topDocs.scoreDocs, newTopDocs.topDocs.scoreDocs, null, null);
+            mergeResult = getMergedScoreDocs(sourceScoreDocs, newScoreDocs, null, null);
         }
-
         return new TopDocsAndMaxScore(getTopDocs(mergeResult, mergedTotalHits), Math.max(source.maxScore, newTopDocs.maxScore));
     }
 


src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryScoreDocsMerger.java
```diff
@@ -47,27 +59,53 @@ public T[] merge(final T[] sourceScoreDocs, final T[] newScoreDocs, final Compar
...
+        if (isCollapseEnabled) {
+            if (sourceScoreDocs.length != Objects.requireNonNull(sourceCollapseValues, "collapse values cannot be null").length) {
+                throw new IllegalArgumentException(
+                    "cannot merge collapse values because the number of elements does not match score docs count"
+                );
+            }
+
+            if (newScoreDocs.length != Objects.requireNonNull(newCollapseValues, "collapse values cannot be null").length) {
+                throw new IllegalArgumentException(
+                    "cannot merge collapse values because the number of elements does not match score docs coun"
```
martin-gaievski:
```suggestion
                    "cannot merge collapse values because the number of elements does not match score docs count"
```

src/main/java/org/opensearch/neuralsearch/util/HybridQueryUtil.java
```diff
@@ -50,30 +52,73 @@ public static boolean isHybridQueryWrappedInBooleanQuery(final SearchContext sea
...
-     * @param booleanClauses a list of boolean clauses from a boolean query
-     * @return true if the clauses represent a hybrid query wrapped in a boolean must clause with the rest of the clauses being filters
+     * @param booleanClauses list of clauses from the main query
+     * @return query if the clauses represent a hybrid query wrapped in a boolean must clause with the rest of the clause being a filter or must_not
      */
-    public static boolean isHybridQueryWrappedInBooleanMustQueryWithFilters(List<BooleanClause> booleanClauses) {
-        boolean isFirstClauseMustHybrid = !booleanClauses.isEmpty()
-            && booleanClauses.getFirst().occur() == BooleanClause.Occur.MUST
-            && booleanClauses.getFirst().query() instanceof HybridQuery;
+    public static Query isHybridQueryWrappedInBooleanMustQueryWithFilters(List<BooleanClause> booleanClauses) {
```
martin-gaievski: let's call it `transformHybridQueryWrappedInBooleanMustQuery`, it's different from the previous method, instead of flagging if hybrid query is part of the argument, it return query or null. Semantically those are two different things

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryScoreDocsMerger.java
```diff
@@ -114,4 +169,7 @@ private boolean compareCondition(
             return comparator.compare((T) oldScoreDoc, (T) secondScoreDoc) >= 0;
         }
     }
+
+    public record MergeResult<T extends ScoreDoc>(T[] scoreDocs, Object[] collapseValues) {
```
martin-gaievski, line 179: please add class description, this is public class

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryScoreDocsMerger.java
```diff
@@ -47,27 +59,53 @@ public T[] merge(final T[] sourceScoreDocs, final T[] newScoreDocs, final Compar
...
         }
+
+        if (isCollapseEnabled) {
+            if (sourceScoreDocs.length != Objects.requireNonNull(sourceCollapseValues, "collapse values cannot be null").length) {
+                throw new IllegalArgumentException(
+                    "cannot merge collapse values because the number of elements does not match score docs count"
+                );
+            }
+
+            if (newScoreDocs.length != Objects.requireNonNull(newCollapseValues, "collapse values cannot be null").length) {
```
martin-gaievski, line 78: this if condition and the one above it - they are throwing identical exceptions. I suggest let's add logging that clearly identifies which one of those two cases throw an exception

martin-gaievski (commented):
Code looks good overall, except for several minor comments

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryScoreDocsMerger.java
vibrantvarun: ack

src/main/java/org/opensearch/neuralsearch/util/HybridQueryUtil.java
vibrantvarun: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryScoreDocsMerger.java
vibrantvarun: ack

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryScoreDocsMerger.java
vibrantvarun: ack

src/main/java/org/opensearch/neuralsearch/search/query/TopDocsMerger.java
```diff
@@ -6,24 +6,33 @@
...
+import org.opensearch.search.collapse.CollapseContext;
 import org.opensearch.search.sort.SortAndFormats;
+import org.opensearch.neuralsearch.search.query.HybridQueryScoreDocsMerger.MergeResult;
 
 /**
  * Utility class for merging TopDocs and MaxScore across multiple search queries
  */
 @RequiredArgsConstructor
 public class TopDocsMerger {
+    private static final Logger log = LogManager.getLogger(TopDocsMerger.class);
```
martin-gaievski: we switched to declarative logger import instead of explicit declaration, @Log4j2

src/main/java/org/opensearch/neuralsearch/search/query/TopDocsMerger.java
```diff
@@ -64,11 +74,22 @@ public TopDocsAndMaxScore merge(final TopDocsAndMaxScore source, final TopDocsAn
...
         }
         TotalHits mergedTotalHits = getMergedTotalHits(source, newTopDocs);
-        TopDocsAndMaxScore result = new TopDocsAndMaxScore(
-            getTopDocs(getMergedScoreDocs(source.topDocs.scoreDocs, newTopDocs.topDocs.scoreDocs), mergedTotalHits),
-            Math.max(source.maxScore, newTopDocs.maxScore)
-        );
-        return result;
+
+        MergeResult mergeResult;
+        if (isCollapseEnabled()) {
```
martin-gaievski, line 79: there is minor duplication of logic in the following section. This can be written like this:
```
ScoreDoc[] sourceScoreDocs = source.topDocs.scoreDocs;
ScoreDoc[] newScoreDocs = newTopDocs.topDocs.scoreDocs;

if (isCollapseEnabled()) {
    if (!(source.topDocs instanceof CollapseTopFieldDocs)) {
        throw new IllegalStateException("Collapse enabled but TopDocs not CollapseTopFieldDocs");
    }
    CollapseTopFieldDocs sourceCollapse = (CollapseTopFieldDocs) source.topDocs;
    mergeResult = getMergedScoreDocs(sourceScoreDocs, newScoreDocs, 
        sourceCollapse.collapseValues, newCollapse.collapseValues);
} else {
    mergeResult = getMergedScoreDocs(sourceScoreDocs, newScoreDocs, null, null);
}
```

heemin32 (approved):

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -81,25 +81,32 @@ public boolean searchWith(
...
-                    validateNestedBooleanQuery(booleanClause.query(), getMaxDepthLimit(searchContext));
-                }
+            // If Collapse with Inner hits is applied with HybridQuery then it for each collapsed group,
+            // opensearch will fan out new query in the form of HybridQuery wrapped under Boolean Query with filters.
+            // if the field is not present in the document, then it will contain a must_not clause for the inner_hits condition. Either of
+            // filter clause or must not clause can exist.
+            // In this case, the bulkScorer will be DefaultBulkScorer and the scorer will be hybridQueryScorer.
+            // Therefore, we need to create a new Boolean query by removing hybrid query and add it subqueries in should clause to bring
+            // docIdIterator and BulkScorer in sync.
+            Query transformedBooleanQuery = transformHybridQueryWrappedInBooleanMustQuery(booleanClauses);
```
bzhangam, line 97: After the transform we will not use hybridQueryScorer to be in sync with the DefaultBulkScorer? Will this impact how we score the query?

vibrantvarun: No. The hybrid query can only applied to parent docs. It does not work at inner hits level. 

Following query
```
bool
    must
         hybrid
               subquery 1
               subquery 2
    filter
         //inner hits query
```         
gets converted to
```
bool
      must
          bool
              should
                   subquery 1
                   subquery 2
      filter
         //inner hits query
```

So what it does instead of load hybridQueryScorer for inner hits query it loads bool query scorer. The above two query achieve the same purpose

PR #1704, opened 2025-12-18
Support Two-Phase Query in sparse_vector field

### LINKED ISSUE #1700: [PROPOSAL] Support Two-Phase Query in sparse_vector field
### What are you proposing?
In 3.3, we introduced a "sparse_vector" field which we run sparse ANN algorithm on it.  To maintain the balance between having high query performance and reducing building index overhead, we introduced a hybrid segment mechanism for this field that only when a segment has enough documents then we build the sparse ANN algorithm segment, otherwise, we fall back to rank_features segment and use plain neural sparse query.

Neural sparse two-phase algorithm was introduced in 2.15 and proven to be a very efficient approach comparing to plain neural sparse query when querying rank_features field. However, due to its limitation, "sparse_vector" does not support two-phase query on its rank_features segment parts.

Here, I'm proposing a solution to apply two-phase query algorithm on the rank_features segment of "sparse_vector" field.
Query performance optimization is like a forever topic. It's even more critical especially when we are able to index billions of sparse vectors.

A query on "sparse_vector" field is dispatched to both sparse ANN segments and rank_features segments as it has a hybrid segment pattern. SEISMIC query algorithm is run on sparse ANN segments and plain neural sparse query (a boolean query internally) is run on rank_features segments. Either users enable or not enable concurrent segment search, the plain neural sparse query would be the bottleneck. Users can definitely tune `approximate_threshold` to only use sparse ANN segments or even call `forcemerge` API to generate a single sparse ANN segment. These can boost query performance, but users have to endure the high CPU consumption during index building which can last for a while.

This proposal tries to optimize the "bottleneck" by adopting the two-phase query algorithm on rank_features segments to improve the overall query performance without sacrificing the cluster resources for indexing building. 

```
GET /my-sparse-ann-index/_search
{
  "query": {
    "neural_sparse": {
      "sparse_embedding": {
        "query_text": "machine learning algorithms",
        "model_id": "your_sparse_model_id",
        "method_parameters": {
          "k": 10,
          "top_n": 10,
          "heap_factor": 1.0,
          "two_phase_enabled": true,
          "two_phase_parameter": {
	          "prune_type": "max_ratio",
	          "prune_ratio": 0.5,
	          "expansion_rate": 5.0,
	          "max_window_size": 1000
          }
        }
      }
    }
  }
}
```

We'll introduce new parameters for two-phase query. "two_phase_enabled" indicates whether we'll use two-phase algorithm query for rank_features segments, it's false by default. "two_phase_parameter" section is exactly the same with same parameter defined in `neural_sparse_two_phase_processor` search processor to be consistent. Please refer to [link](https://docs.opensearch.org/latest/search-plugins/search-pipelines/neural-sparse-query-two-phase-processor/#request-body-fields) for details.
No.

As the default value of two_phase_enabled is false, when it's not used in query, we'll fall back to current path without impacting any existing logic.

### Proposed Solution

Image

Unlike the current two phase search processor logic, which after phase 1, documents with partial scores across segments are reranked together. We want to implement a segment level two-phase algorithm.

We'll introduce a new `TwoPhaseScorer`. It's a compositor scorer containing two sub scorers: phase 1 scorer and phase 2 scorer. Phase 1 scorer, a boolean scorer, constructed with high weight query tokens, is run ahead with any filtering. It retrieves top k' documents ranked with their partial scores. k' is query size times expansion_rate. Then we align phase 2 boolean scorer with top k' documents to fulfill the distance score with low weight query tokens.

TwoPhaseScorer will returns a final top k (sparse ann query size, defined in `method_parameters.k`) documents with highest full distance scores.

There is a scorer selector to select which scorer to run based on segment doc size. We'll add a logic to choose when to use TwoPhaseScorer.

chishui:
### Description
1. Support Two-Phase Query in sparse_vector field
2. minor: fix a back-log issue, replace using cost() with score() for score retrieval function.

### Related Issues
Resolves #1700 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/sparse/query/PhaseOneLeafCollector.java
+++ b/src/main/java/org/opensearch/neuralsearch/sparse/query/PhaseOneLeafCollector.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.sparse.query;
+
+import org.apache.commons.lang3.tuple.Pair;
+import org.apache.lucene.search.LeafCollector;
+import org.apache.lucene.search.Scorable;
+import org.apache.lucene.util.LongHeap;
+import org.apache.lucene.util.NumericUtils;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+
+/**
+ * A LeafCollector for phase one of two-phase sparse vector search.
+ * Collects top-scoring documents using a min-heap, encoding doc IDs and scores into longs for efficient storage.
+ */
+public class PhaseOneLeafCollector implements LeafCollector {
+    private Scorable scorer;
+    private final LongHeap resultHeap;
+
+    /**
+     * @param resultSize maximum number of top results to retain
+     */
+    public PhaseOneLeafCollector(int resultSize) {
+        this.resultHeap = new LongHeap(resultSize);
+    }
+
+    @Override
+    public void setScorer(Scorable scorer) throws IOException {
+        this.scorer = scorer;
+    }
+
+    @Override
+    public void collect(int doc) throws IOException {
+        float score = this.scorer.score();
+        long encoded = encode(doc, score);
+        resultHeap.insertWithOverflow(encoded);
+    }
+
+    /**
+     * Returns collected results as an iterator sorted by doc ID ascending.
+     */
+    public ResultsDocValueIterator<Float> getPhaseOneResults() {
+        List<Pair<Integer, Float>> results = new ArrayList<>(resultHeap.size());
+        for (int i = 1; i <= resultHeap.size(); i++) {
+            long encoded = resultHeap.get(i);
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/neuralsearch/sparse/query/SparseAnnQueryBuilder.java
```diff
@@ -258,14 +321,40 @@ public Query doToQuery(QueryShardContext context) throws IOException {
...
             .queryVector(new SparseVector(integerTokens, new ByteQuantizer(quantizationCeilSearch)))
             .fallbackQuery(fallbackQuery)
+            .rankFeaturesPhaseOneQuery(twoPhaseQueries.get(0))
+            .rankFeaturesPhaseTwoQuery(twoPhaseQueries.get(1))
+            .sparseQueryTwoPhaseInfo(sparseQueryTwoPhaseInfo)
             .filter(filterQuery)
             .build();
     }
 
+    private List<Query> buildTwoPhaseQueries(QueryShardContext context) throws IOException {
```
yuye-aws: [nit] seems that context is not used

Issue #1705, opened 2025-12-19
[BUG] Hybrid Query drops Top relevant documents from Knn query


vibrantvarun:
### What is the bug?
When the search request contains a hybrid query with two subqueries a match and a knn subquery then it drops high relevant documents in the final search result.

### How can one reproduce the bug?
Index some good amount of data - 5k documents on a single shard index. Here specifically take single shard index to understand the bug depth after reproducing it. 
1. Create a match query that only returns 1 document. 
2. Create a knn query.  
3. Do knn query standalone and keep the results aside.
4. Add match and knn under hybrid clause and perform the search.
5. There is high change that top rated knn documents might be missing.

### What is the expected behavior?
It should return the result with highest relevance possible after combing them.

### What is your host/environment?
OS 3.3

PR #1706, opened 2025-12-19
Fix postion overflow in hybrid bulk scorer

vibrantvarun:
### Description

This PR contains the fix of position overflow of docId in HybridBulkScorer.
Earlier the method getNextDocIdCandidate finds the maxDocId and saves in the variable nextDoc.
Lets consider the following example to understand better how the bug is reproduced.
 Match query docId = [30004]
 Knn query docId = [4295, 4586, 37007]

### 1st iteration
Consider in a segment the maximum docId number is 38000

Scoring starts:
Scoring 1st window where min =0 and max = 4096
The [advance method ](https://github.com/opensearch-project/neural-search/blob/3.3/src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java#L158)will move the docIdIterator to first docId of each queries respectively. The docIds array becomes [30004, 4295]

The [allDocIdUsed](https://github.com/opensearch-project/neural-search/blob/d22991e2dbf4c4794967dde2c0384f2459ecd203/src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java#L172C60-L173C1) becomes false both the docIds are greater than 4096. Therefore, we move to get [getNextDocIdCandidate](https://github.com/opensearch-project/neural-search/blob/3.3/src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java#L181C17-L181C38).
It returns the max of both the Id's in the docId array i.e. 30004

### 2nd iteration
Scoring 2nd window where min = 30004 and max  30004+Interval_size
The max value returned from the first Iteration will become min for the next iteration as per [CancellableBulkScorer](https://github.com/opensearch-project/OpenSearch/blob/3.3/server/src/main/java/org/opensearch/search/internal/CancellableBulkScorer.java#L66).
So we start considering docIds that are greater than 30004 and **completely skip knn query** docId's 4295 and 4586/

The HQ results of query phase will contain docId 30004 and 37007.

This results in missing docIds that could have high scores.

### Fix
1. nextDoc variable instead of being a single variable should be an array.
2. It stores nextDoc that has results for each query.
3. Take minimum of both the docId's.

### Related Issues
https://github.com/opensearch-project/neural-search/issues/1705

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
@@ -6,6 +6,7 @@ package org.opensearch.neuralsearch.processor.normalization;
 
 import java.math.BigDecimal;
 import java.math.RoundingMode;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -124,9 +125,8 @@ public class RRFNormalizationTechnique implements ScoreNormalizationTechnique, E
         int topDocsIndex,
         TriConsumer<DocIdAtSearchShard, Float, Integer> scoreProcessor
     ) {
-        for (int position = 0; position < topDocs.scoreDocs.length; position++) {
-            ScoreDoc scoreDoc = topDocs.scoreDocs[position];
-            float normalizedScore = calculateNormalizedScore(position);
+        for (ScoreDoc scoreDoc : topDocs.scoreDocs) {
+            float normalizedScore = calculateNormalizedScore(Arrays.asList(topDocs.scoreDocs).indexOf(scoreDoc));
             DocIdAtSearchShard docIdAtSearchShard = new DocIdAtSearchShard(scoreDoc.doc, searchShard);
             scoreProcessor.apply(docIdAtSearchShard, normalizedScore, topDocsIndex);
             scoreDoc.score = normalizedScore;
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java
@@ -179,19 +179,14 @@ public class HybridBulkScorer extends BulkScorer {
     }
 
     private int getNextDocIdCandidate(final int[] docsIds) {
-        int[] nextDoc = new int[docsIds.length];
-        Arrays.fill(nextDoc, -1);
-        int minDocIdForNextIteration = Integer.MAX_VALUE;
-        for (int i = 0; i < nextDoc.length; i++) {
-            if (docIds[i] != DocIdSetIterator.NO_MORE_DOCS) {
-                nextDoc[i] = Math.max(nextDoc[i], docIds[i]);
-                if (nextDoc[i] < minDocIdForNextIteration) {
-                    minDocIdForNextIteration = nextDoc[i];
-                }
+        int minDocIdForNextIteration = DocIdSetIterator.NO_MORE_DOCS;
+        for (int docId : docsIds) {
+            if (docId < minDocIdForNextIteration) {
+                minDocIdForNextIteration = docId;
             }
         }
 
-        return minDocIdForNextIteration == -1 ? DocIdSetIterator.NO_MORE_DOCS : minDocIdForNextIteration;
+        return minDocIdForNextIteration;
     }
 
     /**


src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
```diff
@@ -125,8 +124,9 @@ private void processTopDocsEntry(
         int topDocsIndex,
         TriConsumer<DocIdAtSearchShard, Float, Integer> scoreProcessor
     ) {
-        for (ScoreDoc scoreDoc : topDocs.scoreDocs) {
-            float normalizedScore = calculateNormalizedScore(Arrays.asList(topDocs.scoreDocs).indexOf(scoreDoc));
```
vibrantvarun, line 129: This was n^2 complexity due to usage of indexOf

martin-gaievski: Let's do this in a separate PR, this makes things complicated if we need to troubleshoot issues. I understood this is only optimization

src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java
```diff
@@ -179,13 +179,19 @@ private boolean allDocIdsUsed(int[] docsIds, int max) {
     }
 
     private int getNextDocIdCandidate(final int[] docsIds) {
-        int nextDoc = -1;
-        for (int doc : docsIds) {
-            if (doc != DocIdSetIterator.NO_MORE_DOCS) {
-                nextDoc = Math.max(nextDoc, doc);
+        int[] nextDoc = new int[docsIds.length];
+        Arrays.fill(nextDoc, -1);
+        int minDocIdForNextIteration = Integer.MAX_VALUE;
```
martin-gaievski: initialize with max value may be not safe, isn't NO_MORE_DOCS a better option?

src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java
```diff
@@ -179,13 +179,19 @@ private boolean allDocIdsUsed(int[] docsIds, int max) {
...
+        for (int i = 0; i < nextDoc.length; i++) {
+            if (docIds[i] != DocIdSetIterator.NO_MORE_DOCS) {
```
martin-gaievski: should this read from `docsIds`. Same for the next line

src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java
```diff
@@ -179,13 +179,19 @@ private boolean allDocIdsUsed(int[] docsIds, int max) {
...
+                nextDoc[i] = Math.max(nextDoc[i], docIds[i]);
+                if (nextDoc[i] < minDocIdForNextIteration) {
+                    minDocIdForNextIteration = nextDoc[i];
+                }
             }
         }
-        return nextDoc == -1 ? DocIdSetIterator.NO_MORE_DOCS : nextDoc;
+
+        return minDocIdForNextIteration == -1 ? DocIdSetIterator.NO_MORE_DOCS : minDocIdForNextIteration;
```
martin-gaievski: is this scenario even possible: `minDocIdForNextIteration == -1`? I think this is dead code now

src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java
```diff
@@ -179,13 +179,19 @@ private boolean allDocIdsUsed(int[] docsIds, int max) {
     }
 
     private int getNextDocIdCandidate(final int[] docsIds) {
-        int nextDoc = -1;
-        for (int doc : docsIds) {
-            if (doc != DocIdSetIterator.NO_MORE_DOCS) {
-                nextDoc = Math.max(nextDoc, doc);
+        int[] nextDoc = new int[docsIds.length];
```
martin-gaievski: do we really need `nextDoc` array? It's filled with `-1`, then `Math.max(-1, docIds[i])` is always `docIds[i]` (since doc IDs are non-negative). And array is never used for anything else
I think following code is doing the same, but is simpler:
```
private int getNextDocIdCandidate(final int[] docsIds) {
    int minDocId = Integer.MAX_VALUE;
    for (int doc : docsIds) {
        if (doc != DocIdSetIterator.NO_MORE_DOCS && doc < minDocId) {
            minDocId = doc;
        }
    }
    return minDocId == Integer.MAX_VALUE ? DocIdSetIterator.NO_MORE_DOCS : minDocId;
}
```

src/test/java/org/opensearch/neuralsearch/query/HybridBulkScorerTests.java
```diff
@@ -106,4 +106,107 @@ public void testScoreWithInvalidRange() throws IOException {
...
+        when(mockIterator2.docID()).thenReturn(-1);
+        when(mockIterator2.advance(0)).thenReturn(DocIdSetIterator.NO_MORE_DOCS);
+        when(mockIterator2.nextDoc()).thenReturn(DocIdSetIterator.NO_MORE_DOCS);
+
+        // Execute scoring
+        int result = bulkScorer.score(mockLeafCollector, null, 0, MAX_DOC);
+
+        // Should return NO_MORE_DOCS when all scorers are exhausted
+        assertEquals("Should return NO_MORE_DOCS when all scorers exhausted", DocIdSetIterator.NO_MORE_DOCS, result);
+    }
```
martin-gaievski, line 211: we need more test cases, but that can be added in later PRs:
1. multi-window iteration test
test that multiple windows are processed correctly with advancing doc IDs:
2. window boundary edge cases
3. sparse scorer (don't confuse with sparse query)  test

src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java
vibrantvarun: Ah Infact we can even get rid of this condition

doc != DocIdSetIterator.NO_MORE_DOCS if we initialize by DocIdSetIterator.NO_MORE_DOCS.

src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
vibrantvarun: ack.

src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java
vibrantvarun: yes. Integer.MaxValue and No_more_docs value is same

src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java
vibrantvarun: Change the code itself.

src/test/java/org/opensearch/neuralsearch/query/HybridBulkScorerTests.java
vibrantvarun: ack.

src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java
vibrantvarun: Changed the code itself/

heemin32 (approved):

PR #1713, opened 2025-12-25
[Agentic Search] Select explicit index for Agentic Query if returned from ListIndexTool

owaiskazi19:
### Description
Select explicit index for Agentic Query if returned from ListIndexTool otherwise can search on all indices

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
@@ -858,21 +858,15 @@ public class MLCommonsClientAccessor {
                         }
 
                         // Extract agent steps based on model type
-                        if (isClaude) {
-                            Map<String, String> claudeResult = extractClaudeAgentSteps(modelResponseMap);
-                            if (claudeResult.containsKey(STEPS_FIELD) && !claudeResult.get(STEPS_FIELD).isEmpty()) {
-                                agentSteps.add(claudeResult.get(STEPS_FIELD));
+                        if (isClaude || isOpenAI) {
+                            Map<String, String> extractedResult = isClaude
+                                ? extractClaudeAgentSteps(modelResponseMap)
+                                : extractOpenAIAgentSteps(modelResponseMap);
+                            if (extractedResult.containsKey(STEPS_FIELD) && !extractedResult.get(STEPS_FIELD).isEmpty()) {
+                                agentSteps.add(extractedResult.get(STEPS_FIELD));
                             }
-                            if (claudeResult.containsKey(INDEX_NAME_FIELD)) {
-                                result.put(SELECTED_INDEX, claudeResult.get(INDEX_NAME_FIELD));
-                            }
-                        } else if (isOpenAI) {
-                            Map<String, String> openAIResult = extractOpenAIAgentSteps(modelResponseMap);
-                            if (openAIResult.containsKey(STEPS_FIELD) && !openAIResult.get(STEPS_FIELD).isEmpty()) {
-                                agentSteps.add(openAIResult.get(STEPS_FIELD));
-                            }
-                            if (openAIResult.containsKey(INDEX_NAME_FIELD)) {
-                                result.put(SELECTED_INDEX, openAIResult.get(INDEX_NAME_FIELD));
+                            if (extractedResult.containsKey(INDEX_NAME_FIELD)) {
+                                result.put(SELECTED_INDEX, extractedResult.get(INDEX_NAME_FIELD));
                             }
                         }
                     } catch (Exception e) {


src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -921,54 +938,104 @@ private static Map<String, Object> extractJsonObjectFromString(String text) {
...
         if (content != null && !content.isBlank()) {
             steps.add(content);
         }
-        return steps;
+
+        // Extract index_name from tool_calls
+        List<Map<String, Object>> toolCalls = (List<Map<String, Object>>) message.get("tool_calls");
+        if (toolCalls != null) {
+            for (Map<String, Object> toolCall : toolCalls) {
+                Map<String, Object> function = (Map<String, Object>) toolCall.get("function");
```
rithinpullela, line 1017: Should we also validate that the tool name is `QueryPlanningTool`?

src/main/java/org/opensearch/neuralsearch/ml/dto/AgentExecutionDTO.java
```diff
@@ -16,4 +16,5 @@ public class AgentExecutionDTO {
     private final String dslQuery;
     private final String agentStepsSummary;
     private final String memoryId;
+    private final String selectedIndex;
```
rithinpullela, line 17: Would this trigger BWC issues during B/G?

rithin-pullela-aws:
Thanks for this PR Owais, changes look good to me, added 2 minor comments

src/main/java/org/opensearch/neuralsearch/ml/dto/AgentExecutionDTO.java
owaiskazi19: Good call out. Added a new constructor

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
owaiskazi19: `index_name` would be associated with QPT only but still added the validation to be on the safer side

rithin-pullela-aws (approved):
Thanks for the feature Owais!

src/main/java/org/opensearch/neuralsearch/ml/dto/AgentExecutionDTO.java
```diff
@@ -4,16 +4,27 @@
...
+
+    public AgentExecutionDTO(String dslQuery, String agentStepsSummary, String memoryId) {
+        this(dslQuery, agentStepsSummary, memoryId, null);
+    }
+
+    // New constructor with selectedIndex
+    public AgentExecutionDTO(String dslQuery, String agentStepsSummary, String memoryId, String selectedIndex) {
```
vibrantvarun, line 24: Why we removed @AllArgsConstructor earlier we are creating constructor anyway with all fields?

owaiskazi19: Added a new constructor to avoid BWC failure

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -853,9 +859,21 @@ private Map<String, String> extractConversationalAgentResult(
 
                         // Extract agent steps based on model type
                         if (isClaude) {
-                            agentSteps.addAll(extractClaudeAgentSteps(modelResponseMap));
+                            Map<String, String> claudeResult = extractClaudeAgentSteps(modelResponseMap);
```
bzhangam: After the extract we are doing the same logic for claude and open AI so no need to duplicate them.

owaiskazi19: The logic is little different for both. For claude we are looking for extracted index in 
```
 Map<String, Object> toolUse = (Map<String, Object>) item.get("toolUse");
```

for openai 
```
                if (function != null && QUERY_PLANNING_TOOL.equals(function.get("name")) && function.containsKey("arguments")) {
```
and even after that the field names are different

bzhangam: I'm saying after the extract the logic to set the value in the agentSteps and the result is the same.

```
 if (claudeResult.containsKey(STEPS_FIELD) && !claudeResult.get(STEPS_FIELD).isEmpty()) {
      agentSteps.add(claudeResult.get(STEPS_FIELD));
  }
  if (claudeResult.containsKey(INDEX_NAME_FIELD)) {
      result.put(SELECTED_INDEX, claudeResult.get(INDEX_NAME_FIELD));
  }
```

and

```
if (openAIResult.containsKey(STEPS_FIELD) && !openAIResult.get(STEPS_FIELD).isEmpty()) {
    agentSteps.add(openAIResult.get(STEPS_FIELD));
}
if (openAIResult.containsKey(INDEX_NAME_FIELD)) {
    result.put(SELECTED_INDEX, openAIResult.get(INDEX_NAME_FIELD));
}
```

Can we simply do
```
if (isClaude || isOpenAI) {
  Map<String, String> result = isClaude ? extractClaudeAgentSteps(modelResponseMap) : extractOpenAIAgentSteps(modelResponseMap);
  if (result.containsKey(STEPS_FIELD) && !result.get(STEPS_FIELD).isEmpty()) {
      agentSteps.add(result.get(STEPS_FIELD));
  }
  if (result.containsKey(INDEX_NAME_FIELD)) {
      result.put(SELECTED_INDEX, result.get(INDEX_NAME_FIELD));
  }
}

```

owaiskazi19: I see. Makes sense. Pushed the change

Issue #1703, opened 2025-12-18
[BUG] Reciprocal Rank Fusion (RRF) drops high-ranked KNN-only documents while retaining lower-ranked ones


martin-gaievski:
When using hybrid queries with Reciprocal Rank Fusion (RRF) normalization, high-ranked KNN-only documents are systematically excluded from final results while lower-ranked KNN-only documents are included. This violates the fundamental mathematical properties of RRF.

### RRF score calculation

RRF assigns scores as: `RRF_score = 1 / (k + rank)`

For documents matching only the KNN subquery (not the keyword subquery), the RRF score is solely determined by their KNN rank. A document at KNN rank 1 has score `1/(60+1) = 0.0164`, which is strictly greater than a document at KNN rank 95 with score `1/(60+95) = 0.0065`.

**Invariant**: If a KNN-only document at rank N appears in final results, ALL KNN-only documents at ranks 1 through N-1 MUST also appear (they have higher scores).

### Observed behavior

This invariant is violated. Example from reproduction:

| KNN Rank | Expected in Top 100 | Actual Result |
|----------|---------------------|---------------|
| 1        |  Yes              |  Missing    |
| 5        |  Yes              |  Missing    |
| 11       |  Yes              |  Missing    |
| 95       |  Yes              |  Present    |

### Step 1: Create search pipeline with RRF

```json
PUT /_search/pipeline/rrf-pipeline
{
  "description": "RRF pipeline for hybrid search",
  "phase_results_processors": [
    {
      "score-ranker-processor": {
        "combination": {
          "technique": "rrf",
          "parameters": {
            "rank_constant": 60
          }
        }
      }
    }
  ]
}
```

### Step 2: Run knn query

This establishes the expected ranking:

```json
GET /your-index/_search
{
  "query": {
    "knn": {
      "your_embedding_field": {
        "vector": [<your-256-dim-embedding>],
        "k": 100
      }
    }
  },
  "size": 100,
  "_source": false
}
```

Save the document IDs and their ranks (1-100).

### Step 3: Run hybrid query with RRF-based pipeline

Use a keyword query that has minimal/no overlap with KNN results:

```json
GET /your-index/_search?search_pipeline=rrf-pipeline
{
  "query": {
    "hybrid": {
      "queries": [
        {
          "match": {
            "text_field": {
              "query": "rare_keyword_with_few_matches"
            }
          }
        },
        {
          "knn": {
            "your_embedding_field": {
[...]

PR #1718, opened 2026-01-02
Add subquery global rank for RRF score calculation & fix position overflow in HybridBulkScorer

vibrantvarun:
### Description
Currently, the RRF logic misses the calculation of RRF normalized score as per global rank of the document in the subquery across multiple shards. This PR adds the calculation of global rank and then use it during RRF score calculation.

This PR also fixes position overflow in HybridBulkScorer where [earlier the first document less than maximum document is used for calculating windowBaseScore](https://github.com/opensearch-project/neural-search/blob/3.3/src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java#L86). This lead to a position overflow in following scenario. 
```
max = 17100
int[] docIds = [16720, 12501];
// As per old logic
topDoc becomes 16720
windowBase = 16720 ~ 4095 = 16384

This led to miss documents from the range of 12501 - 16384 because the windowBase is 16384. 

Instead the logic should select topDoc as 12501.
```

### Related Issues
https://github.com/opensearch-project/neural-search/issues/1703#issuecomment-3700713893

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
@@ -77,9 +77,9 @@ public class RRFNormalizationTechnique implements ScoreNormalizationTechnique, E
     public void normalize(final NormalizeScoresDTO normalizeScoresDTO) {
         final List<CompoundTopDocs> queryTopDocs = normalizeScoresDTO.getQueryTopDocs();
 
-        Map<Integer, Map<String, Integer>> sortedDocIdsPerSubqueryByGlobalRank = !normalizeScoresDTO.isSingleShard()
-            ? sortDocumentsAsPerGlobalRankInIndividualQuery(queryTopDocs)
-            : Map.of();
+        Map<Integer, Map<String, Integer>> sortedDocIdsPerSubqueryByGlobalRank = normalizeScoresDTO.isSingleShard()
+            ? Map.of()
+            : sortDocumentsAsPerGlobalRankInIndividualQuery(queryTopDocs);
 
         for (int referenceShardId = 0; referenceShardId < queryTopDocs.size(); referenceShardId++) {
             processTopDocs(
@@ -147,15 +147,18 @@ public class RRFNormalizationTechnique implements ScoreNormalizationTechnique, E
     public Map<DocIdAtSearchShard, ExplanationDetails> explain(final ExplainDTO explainDTO) {
         final List<CompoundTopDocs> queryTopDocs = explainDTO.getQueryTopDocs();
         Map<DocIdAtSearchShard, List<Float>> normalizedScores = new HashMap<>();
-        Map<Integer, Map<String, Integer>> sortedDocIdsPerSubqueryByGlobalRank = !explainDTO.isSingleShard()
-            ? sortDocumentsAsPerGlobalRankInIndividualQuery(queryTopDocs)
-            : Map.of();
+        Map<Integer, Map<String, Integer>> sortedDocIdsPerSubqueryByGlobalRank = explainDTO.isSingleShard()
+            ? Map.of()
+            : sortDocumentsAsPerGlobalRankInIndividualQuery(queryTopDocs);
         for (int referenceShardId = 0; referenceShardId < queryTopDocs.size(); referenceShardId++) {
-            CompoundTopDocs compoundTopDocs = queryTopDocs.get(referenceShardId);
-            List<TopDocs> topDocsPerSubQuery = compoundTopDocs.getTopDocs();
+            CompoundTopDocs compoundQueryTopDocs = queryTopDocs.get(referenceShardId);
+            if (Objects.isNull(compoundQueryTopDocs)) {
+                continue;
+            }
+            List<TopDocs> topDocsPerSubQuery = compoundQueryTopDocs.getTopDocs();
             int numberOfSubQueries = topDocsPerSubQuery.size();
             processTopDocs(
-                compoundTopDocs,
+                compoundQueryTopDocs,
                 (docId, score, subQueryIndex) -> ScoreNormalizationUtil.setNormalizedScore(
                     normalizedScores,
                     docId,
@@ -185,9 +188,9 @@ public class RRFNormalizationTechnique implements ScoreNormalizationTechnique, E
         SearchShard searchShard = compoundQueryTopDocs.getSearchShard();
 
         for (int topDocsIndex = 0; topDocsIndex < topDocsList.size(); topDocsIndex++) {
-            Map<String, Integer> docIdToRankMap = !sortedDocIdMapPerSubqueryByGlobalRank.isEmpty()
-                ? sortedDocIdMapPerSubqueryByGlobalRank.get(topDocsIndex)
-                : Map.of();
+            Map<String, Integer> docIdToRankMap = sortedDocIdMapPerSubqueryByGlobalRank.isEmpty()
+                ? Map.of()
+                : sortedDocIdMapPerSubqueryByGlobalRank.get(topDocsIndex);
             processTopDocsEntry(topDocsList.get(topDocsIndex), searchShard, topDocsIndex, scoreProcessor, docIdToRankMap, referenceShardId);
         }
     }
@@ -202,8 +205,8 @@ public class RRFNormalizationTechnique implements ScoreNormalizationTechnique, E
     ) {
         for (int position = 0; position < topDocs.scoreDocs.length; position++) {
             ScoreDoc scoreDoc = topDocs.scoreDocs[position];
-            int rank = !docIdToRankMap.isEmpty() ? docIdToRankMap.get(scoreDoc.doc + "_" + referenceShardId) : position;
-            if (rank < 0) {
+            Integer rank = docIdToRankMap.isEmpty() ? position : docIdToRankMap.get(scoreDoc.doc + "_" + referenceShardId);
+            if (Objects.isNull(rank) || rank < 0) {
                 throw new IllegalStateException(
                     "Document not found in global ranking map: doc=" + scoreDoc.doc + ", shard=" + referenceShardId
                 );
--- a/src/test/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflowTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflowTests.java
@@ -9,6 +9,7 @@ import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.when;
 import static org.opensearch.neuralsearch.search.util.HybridSearchResultFormatUtil.createDelimiterElementForHybridSearchResults;
 import static org.opensearch.neuralsearch.search.util.HybridSearchResultFormatUtil.createStartStopElementForHybridSearchResults;
+import static org.opensearch.neuralsearch.plugin.NeuralSearch.EXPLANATION_RESPONSE_KEY;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -29,6 +30,7 @@ import org.opensearch.neuralsearch.processor.combination.ScoreCombinationFactory
 import org.opensearch.neuralsearch.processor.combination.ScoreCombiner;
 import org.opensearch.neuralsearch.processor.normalization.ScoreNormalizationFactory;
 import org.opensearch.neuralsearch.processor.normalization.ScoreNormalizer;
+import org.opensearch.neuralsearch.processor.explain.ExplanationPayload;
 import org.opensearch.search.SearchHit;
 import org.opensearch.search.SearchHits;
 import org.opensearch.search.SearchShardTarget;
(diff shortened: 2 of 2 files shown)


navneet1v:
@vibrantvarun were you able to find out the root cause of the problem and why the current ITs didn't catch the bug?

vibrantvarun:
@heemin32 @bzhangam Can I get some eyes on this.

src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
```diff
@@ -66,9 +77,54 @@ public RRFNormalizationTechnique(final Map<String, Object> params, final ScoreNo
...
+            );
+        }
+    }
+
+    private Map<Integer, List<String>> sortDocumentsAsPerGlobalRankInIndividualQuery(final List<CompoundTopDocs> queryTopDocs) {
+        // Map: subQueryIndex -> PQ that contains ScoreDocs from all shards
+        Map<Integer, PriorityQueue<ShardResultPerSubQuery>> scoreDocsPerSubquery = new HashMap<>();
+        for (int referenceShardId = 0; referenceShardId < queryTopDocs.size(); referenceShardId++) {
+            // List of subquery results
+            List<TopDocs> topDocs = queryTopDocs.get(referenceShardId).getTopDocs();
```
martin-gaievski: potentially we can have NPE here. In other places we do check for null objects, e.g. in `processTopDocs()`:
```java
if (Objects.isNull(compoundQueryTopDocs)) {
    return;  
}
```
If `queryTopDocs` contains a null element, NPE will occur during global ranking but the same null would be safely skipped during processing. It's better add null check in `sortDocumentsAsPerGlobalRankInIndividualQuery()`:
```java
for (int referenceShardId = 0; referenceShardId < queryTopDocs.size(); referenceShardId++) {
    CompoundTopDocs compoundTopDocs = queryTopDocs.get(referenceShardId);
    if (Objects.isNull(compoundTopDocs)) {
        continue;
    }
    List<TopDocs> topDocs = compoundTopDocs.getTopDocs();
    // ...
}
```

src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
```diff
@@ -115,18 +178,27 @@ private void processTopDocs(CompoundTopDocs compoundQueryTopDocs, TriConsumer<Do
...
+        TriConsumer<DocIdAtSearchShard, Float, Integer> scoreProcessor,
+        List<String> sortedDocIdsPerSubqueryByGlobalRank,
+        int shardId
     ) {
-        for (ScoreDoc scoreDoc : topDocs.scoreDocs) {
-            float normalizedScore = calculateNormalizedScore(Arrays.asList(topDocs.scoreDocs).indexOf(scoreDoc));
+        for (int position = 0; position < topDocs.scoreDocs.length; position++) {
+            ScoreDoc scoreDoc = topDocs.scoreDocs[position];
+            int rank = !sortedDocIdsPerSubqueryByGlobalRank.isEmpty()
+                ? sortedDocIdsPerSubqueryByGlobalRank.indexOf(scoreDoc.doc + "_" + shardId)
```
martin-gaievski: this should be good more majority of scenarios, but in rare case when indexOf return -1 this can lead to wrong results. If rank became -1 then later it will be translated to some score, and I believe this will be the highest score instead of error:
```
calculateNormalizedScore(-1) = 1 / (60 + (-1) + 1) = 1 / 60 = 0.0167
```
better change code to be more defensive:
```java
int rank = sortedDocIdsPerSubqueryByGlobalRank.indexOf(scoreDoc.doc + "_" + shardId);
if (rank < 0) {
    throw new IllegalStateException("Document not found in global ranking map: doc=" 
        + scoreDoc.doc + ", shard=" + shardId);
}
```

src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
```diff
@@ -66,9 +77,54 @@ public RRFNormalizationTechnique(final Map<String, Object> params, final ScoreNo
...
+            processTopDocs(
+                queryTopDocs.get(referenceShardId),
+                (docId, score, subQueryIndex) -> {},
+                sortedDocIdsPerSubqueryByGlobalRank,
+                referenceShardId
+            );
+        }
+    }
+
+    private Map<Integer, List<String>> sortDocumentsAsPerGlobalRankInIndividualQuery(final List<CompoundTopDocs> queryTopDocs) {
```
bzhangam: Do we need to check null for queryTopDocs? If no can we add @NonNull to make it clear?

src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
bzhangam: Should we also check null for topDocs?

src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
bzhangam: +1. And maybe we also can make sortedDocIdsPerSubqueryByGlobalRank as Map<String,Integer> so that we can find the rank faster as O(1) rather than through indexOf which is O(n).

src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java
```diff
@@ -82,12 +84,11 @@ public int score(LeafCollector collector, Bits acceptDocs, int min, int max) thr
     }
 
     private void scoreWindow(LeafCollector collector, Bits acceptDocs, int min, int max, int[] docIds) throws IOException {
-        // find the first document ID below the maximum threshold to establish the next scoring window boundary
-        int topDoc = -1;
+        // find the least maximum docId below the maximum threshold to establish the next scoring window boundary
+        int topDoc = DocIdSetIterator.NO_MORE_DOCS;
```
martin-gaievski: I think we can simplify more and re-use some of the existing code:
```java
private void scoreWindow(LeafCollector collector, Bits acceptDocs, int min, int max, int[] docIds) throws IOException {
    // find minimum docId across all scorers (reuse existing method)
    int minDocId = getNextDocIdCandidate(docIds);
    
    // If minimum docId is >= max, no documents to process in this window
    if (minDocId >= max) {
        return;
    }
    //rest of existing code
}
```

src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java
```diff
@@ -82,12 +84,11 @@ public int score(LeafCollector collector, Bits acceptDocs, int min, int max) thr
     }
 
     private void scoreWindow(LeafCollector collector, Bits acceptDocs, int min, int max, int[] docIds) throws IOException {
-        // find the first document ID below the maximum threshold to establish the next scoring window boundary
-        int topDoc = -1;
+        // find the least maximum docId below the maximum threshold to establish the next scoring window boundary
```
martin-gaievski: please also update comment on line below 
```
 // take the next match (at random) and find the window where it belongs
```
I guess random in this case means first matching, with your change it's min

src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java
vibrantvarun: The logic of getNextDocIdCandidate is different. There we find minimum docId.

Here we need least maximum docId.

src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java
vibrantvarun: It is not minimum. It is least maximum.

src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java
martin-gaievski: The word "least" meaning same as "minimum", they are same things.
in your current logic, method `scoreWindow` filters by max and then finds minimum:
```java
int topDoc = DocIdSetIterator.NO_MORE_DOCS;
for (int docId : docIds) {
    if (docId < max && docId < topDoc) {  // filter by max, then find minimum
        topDoc = docId;
    }
}
```
method `getNextDocIdCandidate()` find minimum only, and filtering is done after, by `Math.min()`:
```java
int minDocIdForNextIteration = DocIdSetIterator.NO_MORE_DOCS;
for (int docId : docsIds) {
    if (docId < minDocIdForNextIteration) {  // find minimum
        minDocIdForNextIteration = docId;
    }
}
```

some simple test case:
```
docIds = [16720, 12501, NO_MORE_DOCS], max = 17100
```
results from running class methods:
- scoreWindow
```
| 12501 | min({16720, 12501} where d < 17100) = 12501 |
```
- getNextDocIdCandidate
```
|12501 | min(16720, 12501) = 12501
```

src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java
martin-gaievski: it's the same thing, please see my answer on a previous comment

src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
vibrantvarun: done

src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
```diff
@@ -78,7 +77,7 @@ public RRFNormalizationTechnique(final Map<String, Object> params, final ScoreNo
     public void normalize(final NormalizeScoresDTO normalizeScoresDTO) {
         final List<CompoundTopDocs> queryTopDocs = normalizeScoresDTO.getQueryTopDocs();
 
-        Map<Integer, List<String>> sortedDocIdsPerSubqueryByGlobalRank = !normalizeScoresDTO.isSingleShard()
+        Map<Integer, Map<String, Integer>> sortedDocIdsPerSubqueryByGlobalRank = !normalizeScoresDTO.isSingleShard()
```
martin-gaievski: better not to use negation if. thats possible:
```suggestion
        Map<Integer, Map<String, Integer>> sortedDocIdsPerSubqueryByGlobalRank = normalizeScoresDTO.isSingleShard()
         ? Map.of()
         : sortDocumentsAsPerGlobalRankInIndividualQuery(queryTopDocs);
```

src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
```diff
@@ -115,18 +185,30 @@ private void processTopDocs(CompoundTopDocs compoundQueryTopDocs, TriConsumer<Do
...
-        TriConsumer<DocIdAtSearchShard, Float, Integer> scoreProcessor
+        TriConsumer<DocIdAtSearchShard, Float, Integer> scoreProcessor,
+        Map<String, Integer> docIdToRankMap,
+        int referenceShardId
     ) {
-        for (ScoreDoc scoreDoc : topDocs.scoreDocs) {
-            float normalizedScore = calculateNormalizedScore(Arrays.asList(topDocs.scoreDocs).indexOf(scoreDoc));
+        for (int position = 0; position < topDocs.scoreDocs.length; position++) {
+            ScoreDoc scoreDoc = topDocs.scoreDocs[position];
+            int rank = !docIdToRankMap.isEmpty() ? docIdToRankMap.get(scoreDoc.doc + "_" + referenceShardId) : position;
```
martin-gaievski: with the new Map object this code may lead to NPE. If object is not in the map, then get() will return `null`, and unboxing Integer->int will fail with NPE before reaching `if (rank < 0) {` line. You need to check for null case, for example you can change type from `int` to `Integer`
```java
Integer rank = !docIdToRankMap.isEmpty() 
    ? docIdToRankMap.get(scoreDoc.doc + "_" + referenceShardId) 
    : position;

if (rank == null || rank < 0) {
    throw new IllegalStateException(
        "Document not found in global ranking map: doc=" + scoreDoc.doc + ", shard=" + referenceShardId
    );
}
```

src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
```diff
@@ -82,31 +144,39 @@ public String techniqueName() {
...
-            if (Objects.isNull(compoundQueryTopDocs)) {
-                continue;
-            }
-            List<TopDocs> topDocsPerSubQuery = compoundQueryTopDocs.getTopDocs();
+        Map<Integer, Map<String, Integer>> sortedDocIdsPerSubqueryByGlobalRank = !explainDTO.isSingleShard()
+            ? sortDocumentsAsPerGlobalRankInIndividualQuery(queryTopDocs)
+            : Map.of();
+        for (int referenceShardId = 0; referenceShardId < queryTopDocs.size(); referenceShardId++) {
+            CompoundTopDocs compoundTopDocs = queryTopDocs.get(referenceShardId);
+            List<TopDocs> topDocsPerSubQuery = compoundTopDocs.getTopDocs();
```
martin-gaievski: this code is out of sync with `normalize()` method, null check is missing
```java
for (int referenceShardId = 0; referenceShardId < queryTopDocs.size(); referenceShardId++) {
    CompoundTopDocs compoundTopDocs = queryTopDocs.get(referenceShardId);
    if (Objects.isNull(compoundTopDocs)) {
        continue;
    }
    List<TopDocs> topDocsPerSubQuery = compoundTopDocs.getTopDocs();
    // ...
}
```

bzhangam (approved):

PR #1726, opened 2026-01-13
apply min_score to final results in hybrid query

### LINKED ISSUE #1164: [FEATURE] Apply min_score to normalized/combined results in hybrid query
Today hybrid query pushes `min_score` to the shard level, and that filter got applied to individual sub-queries and before the normalization and/or combination. This is confusing in some scenarios, where expected behavior is to apply the filter to the normalized/combined scores.

Example:

```
{
  "size": 100,
  "_source": [
    "item_id",
    "item_description"],
  "query": {
    "hybrid": {
      "queries": [
        {
          "knn": {
            "embedding": {
              "vector": [...],
              "k": 100
            }
          }
        },
        {
          "match": {
            "item_description": {
              "query": "Looking for something",
              "fuzziness": 1,
              "operator": "and"
            }
          }
        }
      ]
    }
  },
  "min_score": 1.5
}
```

`min_score` will cut off all the results from `knn` query because they are within [0..1.0] interval for normalized vectors. At the same time for `match` query this will cut off mostly only irrelevant documents, because top matching docs will have scores of 20 and greater.

`min_score` or equivalent new custom parameter that defines the cut off line for scores after normalization/combination. 

Some workaround are possible with custom scoring function

Don't confuse it with the lower bound feature of min/max technique https://github.com/opensearch-project/neural-search/issues/299

wdongyu:
### Description
Detail description in #1164 

### Related Issues
Resolves #1164

### Check List

### DIFF
--- a/src/test/java/org/opensearch/neuralsearch/search/query/HybridCollectorManagerTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/search/query/HybridCollectorManagerTests.java
@@ -1164,6 +1164,7 @@ public class HybridCollectorManagerTests extends OpenSearchQueryTestCase {
         Sort sort = new Sort(sortField);
         SortAndFormats sortAndFormats = new SortAndFormats(sort, docValueFormat);
         when(searchContext.sort()).thenReturn(sortAndFormats);
+        when(searchContext.minimumScore()).thenReturn(null);
 
         // Setup collapse context - THIS IS KEY
         CollapseContext collapseContext = mock(CollapseContext.class);


src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -110,10 +120,27 @@ private void combineShardScores(
...
     }
 
+    private static TotalHits.Relation getRelation(final int originalSize, final int filteredSize, final boolean useMinScore) {
+        TotalHits.Relation newRelation = null;
+        if (useMinScore) {
+            // - if min score is not null, then we update total hits to filter size, and make best effort to keep total hits relation.
+            // when filteredSize < originalSize, relation can be determined to EQUAL_TO;
+            // when filteredSize == originalSize, we can't determine the exact value above min score, so we change it to
+            // GREATER_THAN_OR_EQUAL_TO
+            newRelation = filteredSize < originalSize ? TotalHits.Relation.EQUAL_TO : TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO;
```
heemin32: When `filteredSize == originalSize`, shouldn't it be null?

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -306,11 +344,28 @@ public Map<Integer, float[]> getNormalizedScoresPerDocument(final List<TopDocs>
...
             .collect(Collectors.toMap(Map.Entry::getKey, entry -> scoreCombinationTechnique.combine(entry.getValue())));
+
+        return updateCombinedNormalizedScores(combinedNormalizedScoresByDocId, minScore);
+    }
+
+    private Map<Integer, Float> updateCombinedNormalizedScores(
+        final Map<Integer, Float> combinedNormalizedScoresByDocId,
+        final Float minScore
+    ) {
+        if (isMinScoreAvailable(minScore)) {
```
heemin32: We could filter in a single steps when minScore available?
```
private Map<Integer, Float> combineScoresAndGetCombinedNormalizedScoresPerDocument(
    final Map<Integer, float[]> normalizedScoresPerDocument,
    final ScoreCombinationTechnique scoreCombinationTechnique,
    final Float minScore
) {
    // 1. Prepare the stream
    var stream = normalizedScoresPerDocument.entrySet().stream();

    // 2. If minScore is available, combine AND filter in one pass
    if (isMinScoreAvailable(minScore)) {
        return stream
            .map(e -> Map.entry(e.getKey(), scoreCombinationTechnique.combine(e.getValue())))
            .filter(e -> e.getValue() >= minScore)
            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
    }

    // 3. If minScore is NOT available, just combine and collect
    return stream.collect(Collectors.toMap(
        Map.Entry::getKey, 
        e -> scoreCombinationTechnique.combine(e.getValue())
    ));
}
```

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -406,12 +474,14 @@ private List<ExplanationDetails> explainByShard(
...
         final Sort sort,
-        final Map<Integer, Float> combinedNormalizedScoresByDocId
+        final Map<Integer, Float> combinedNormalizedScoresByDocId,
+        final Float minScore
     ) {
         Collection<Integer> sortedDocsIds;
         if (sort != null) {
             List<TopDocs> topDocsPerSubQuery = compoundQueryTopDocs.getTopDocs();
-            sortedDocsIds = getSortedDocIdsBySortCriteria(getTopFieldDocs(sort, topDocsPerSubQuery), sort);
+            Predicate<Integer> docIdFilter = isMinScoreAvailable(minScore) ? combinedNormalizedScoresByDocId::containsKey : null;
```
heemin32: Could you help me to understand why we need this filter only when there is minScore defined?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Features
 - Add support for asymmetric embedding models([#1605](https://github.com/opensearch-project/neural-search/pull/1605))
+- Add support for min_score param in hybrid search
```
martin-gaievski: please follow the previous format and add issue num and link

src/main/java/org/opensearch/neuralsearch/search/query/HybridAggregationProcessor.java
```diff
@@ -23,6 +23,12 @@ public class HybridAggregationProcessor implements AggregationProcessor {
     public void preProcess(SearchContext context) {
         // Simply delegate the call
         delegateAggsProcessor.preProcess(context);
+
+        if (isHybridQuery(context.query(), context) && context.minimumScore() != null) {
+            // unset min_score, so it will not work for when executing sub-queries,
+            // and we will retrieve it from search source to filter out final results after normalization and combination
+            context.minimumScore(Float.MIN_VALUE);
```
martin-gaievski: using Float.MIN_VALUE may lead to a problem and incorrect search hits. Because this is smallest but positive float this will drop documents that are no match/score of 0.0. If document has some score in only one sub-query, but has no match/zero score in another sub-query it can be incorrectly filtered out. 
If you use Float.NEGATIVE_INFINITY this should solve the problem

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -142,6 +174,12 @@ private List<TopFieldDocs> getTopFieldDocs(final Sort sort, final List<TopDocs>
             // Check for scoreDocs length.
             // If scoreDocs length=0 then it means that no results are found for that particular subquery.
             if (topDocs.scoreDocs.length != 0) {
+                if (docIdFilter != null) {
+                    topDocs.scoreDocs = Arrays.stream(topDocs.scoreDocs)
```
martin-gaievski: this code changes the value of the original `topDocs.scoreDocs`, this can lead to different unwanted consequences like inconsistent data in concurrent calls, or result mismatch in explain flow. Better create new object and put filtered data into that new object:
```java
List<TopFieldDocs> topFieldDocs = new ArrayList<>();
    for (TopDocs topDocs : topDocsPerSubQuery) {
        if (topDocs.scoreDocs.length != 0) {
            ScoreDoc[] filteredScoreDocs = topDocs.scoreDocs;
            if (docIdFilter != null) {
                // create new array, not mutating original one
                filteredScoreDocs = Arrays.stream(topDocs.scoreDocs)
                    .filter(scoreDoc -> docIdFilter.test(scoreDoc.doc))
                    .toArray(ScoreDoc[]::new);
            }
            
            // new TopFieldDocs with filtered results
            TopFieldDocs original = (TopFieldDocs) topDocs;
            topFieldDocs.add(new TopFieldDocs(
                new TotalHits(filteredScoreDocs.length, TotalHits.Relation.EQUAL_TO),
                filteredScoreDocs,
                original.fields
            ));
        }
    }
    return topFieldDocs;
```

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -110,10 +120,27 @@ private void combineShardScores(
...
         );
     }
 
+    private static TotalHits.Relation getRelation(final int originalSize, final int filteredSize, final boolean useMinScore) {
+        TotalHits.Relation newRelation = null;
+        if (useMinScore) {
+            // - if min score is not null, then we update total hits to filter size, and make best effort to keep total hits relation.
+            // when filteredSize < originalSize, relation can be determined to EQUAL_TO;
+            // when filteredSize == originalSize, we can't determine the exact value above min score, so we change it to
+            // GREATER_THAN_OR_EQUAL_TO
```
martin-gaievski: I think it should be one additional clause: If the original relation was already `GREATER_THAN_OR_EQUAL_TO` and some filtering occurred, we should still use `GREATER_THAN_OR_EQUAL_TO` because we can't know the true total.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -116,6 +117,15 @@ private boolean getIsSingleShard(final NormalizationProcessorWorkflowExecuteRequ
         return searchPhaseContext.getNumShards() == 1 || request.fetchSearchResultOptional.isEmpty() == false;
     }
 
+    private Float getMinScore(final NormalizationProcessorWorkflowExecuteRequest request) {
+        final SearchPhaseContext searchPhaseContext = request.getSearchPhaseContext();
+        Float minScore = null;
+        if (searchPhaseContext.getRequest() != null && searchPhaseContext.getRequest().source() != null) {
+            minScore = searchPhaseContext.getRequest().source().minScore();
+        }
+        return minScore;
```
martin-gaievski, line 128: 'min_score` is user provided value, if there is no validation already present in core OS then we need special logic to make sure it's none of bad values: negative, Float.NaN, positive infinity etc.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -384,8 +448,12 @@ private List<ExplanationDetails> explainByShard(
...
+
+        // update combined scores with min score
+        combinedNormalizedScoresByDocId = updateCombinedNormalizedScores(combinedNormalizedScoresByDocId, minScore);
+
         // sort combined scores as per sorting criteria - either score desc or field sorting
-        Collection<Integer> sortedDocsIds = getSortedDocsIds(compoundQueryTopDocs, sort, combinedNormalizedScoresByDocId);
+        Collection<Integer> sortedDocsIds = getSortedDocsIds(compoundQueryTopDocs, sort, combinedNormalizedScoresByDocId, minScore);
 
         List<ExplanationDetails> listOfExplanations = new ArrayList<>();
         String combinationDescription = String.format(
```
martin-gaievski, line 478: we can do user a favor and add a note about min_score usage to explain output, something along these lines:
```java
String combinationDescription = String.format(
    Locale.ROOT,
    "%s combination of:%s",
    ((ExplainableTechnique) scoreCombinationTechnique).describe(),
    minScore != null ? String.format(" [filtered by min_score: %.4f]", minScore) : ""
);
```

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
wdongyu: when filteredSize == originalSize, it mean all results are above min score, and we don't know if there exist more results not collected, so new relation are set to `GREATER_THAN_OR_EQUAL_TO`.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
wdongyu: This is derived from the behaviour that minScore defined with a normal query and sort. In that case, the normal query will be executed and scored, followed by a minScore filtering, and final with sort.

src/main/java/org/opensearch/neuralsearch/search/query/HybridAggregationProcessor.java
vibrantvarun: Also remove this code from here and put in HybridQueryCollectorContextSpecFactory.createCollectorContextSpec as there already isHybridQuery check is there. We are already trying to eliminate this HybridAggregationProcessor class altogether and use the defaultAggregationProcessor altogether.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
vibrantvarun: @martin-gaievski @wdongyu first thing first this min score logic should not touch anything related to sorting.

When there is a sort criteria then the scores are showed as null at the search response. Therefore min_score has nothing to do with sorting.

Also, this thing is different from other queries. 
If we  apply min score in sorting then for sure shot we will drop results because the end user expecting results as per sort criteria not by any cutoff from the scores.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -384,8 +448,12 @@ private List<ExplanationDetails> explainByShard(
         Map<Integer, Float> combinedNormalizedScoresByDocId = normalizedScoresPerDoc.entrySet()
             .stream()
             .collect(Collectors.toMap(Map.Entry::getKey, entry -> scoreCombinationTechnique.combine(entry.getValue())));
+
+        // update combined scores with min score
+        combinedNormalizedScoresByDocId = updateCombinedNormalizedScores(combinedNormalizedScoresByDocId, minScore);
+
         // sort combined scores as per sorting criteria - either score desc or field sorting
-        Collection<Integer> sortedDocsIds = getSortedDocsIds(compoundQueryTopDocs, sort, combinedNormalizedScoresByDocId);
+        Collection<Integer> sortedDocsIds = getSortedDocsIds(compoundQueryTopDocs, sort, combinedNormalizedScoresByDocId, minScore);
```
vibrantvarun, line 475: 1. @wdongyu you already filtered docIds in this method combineScoresAndGetCombinedNormalizedScoresPerDocument added in the map. 

A)Then why we need to do it all over again
B) We don't touch sorting logic here as it is little different than regular queries.

A document with a valid sort value can have score less than min score.

cc: @heemin32 @martin-gaievski

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
vibrantvarun: Why we need this isMinScoreAvailable check when you already set the value in the searchContext earlier with Float min value.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -110,10 +120,27 @@ private void combineShardScores(
...
             getDocIdSortFieldsMap(compoundQueryTopDocs, combinedNormalizedScoresByDocId, sort),
             sort,
-            isSingleShard
+            isSingleShard,
+            newRelation
         );
     }
 
+    private static TotalHits.Relation getRelation(final int originalSize, final int filteredSize, final boolean useMinScore) {
+        TotalHits.Relation newRelation = null;
```
vibrantvarun: Why default value is null.

vibrantvarun (commented):
Overall, I have some concerns about the areas this PR touches. 

Especially sorting logic, default value of TotalHits.

Moreover, I think @martin-gaievski said we need more integ tests and also testing all the features this PR touches on larger dataset.

vibrantvarun:
Also one more thing, with pagination_depth in play and from and size this min_score will act as 4th parameter from the customer viewpoint. 

I think we should create an RFC first of the areas this feature should touch. How it be in case of collapse, explain, sorting.

One more serious concern:
In case of RRF if have higher rank_constant then all results are tighly bound and score are very near like 0.015, 0.0151 etc.

In that case, min_score will chop of most results.

cc: @martin-gaievski

src/main/java/org/opensearch/neuralsearch/search/query/HybridAggregationProcessor.java
wdongyu: > using Float.MIN_VALUE may lead to a problem and incorrect search hits. Because this is smallest but positive float this will drop documents that are no match/score of 0.0. If document has some score in only one sub-query, but has no match/zero score in another sub-query it can be incorrectly filtered out. If you use Float.NEGATIVE_INFINITY this should solve the problem

@martin-gaievski OK，I will check that.

wdongyu: @vibrantvarun From the code in QueryPhase, it will follow the process below:

```
aggregationProcessor.preProcess(searchContext);  // NOTICE: we unset minScore here
boolean rescore = executeInternal(searchContext, queryPhaseSearcher);

---> in func executeInternal():
  if (searchContext.minimumScore() != null) {
      // use minScore here to create a collector
  }
  ...
  boolean shouldRescore = queryPhaseSearcher.searchWith()
  
   ---> in func searchWith() and later:
   // we call QueryCollectorContextSpecRegistry.getQueryCollectorContextSpec
  
```

So if we move the code from `preProcess()` to later, `minScore` will take effect on the sub-queries. Please correct me if I miss something.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
wdongyu: > this code changes the value of the original `topDocs.scoreDocs`, this can lead to different unwanted consequences like inconsistent data in concurrent calls, or result mismatch in explain flow. Better create new object and put filtered data into that new object:
> 
> ```java
> List<TopFieldDocs> topFieldDocs = new ArrayList<>();
>     for (TopDocs topDocs : topDocsPerSubQuery) {
>         if (topDocs.scoreDocs.length != 0) {
>             ScoreDoc[] filteredScoreDocs = topDocs.scoreDocs;
>             if (docIdFilter != null) {
>                 // create new array, not mutating original one
>                 filteredScoreDocs = Arrays.stream(topDocs.scoreDocs)
>                     .filter(scoreDoc -> docIdFilter.test(scoreDoc.doc))
>                     .toArray(ScoreDoc[]::new);
>             }
>             
>             // new TopFieldDocs with filtered results
>             TopFieldDocs original = (TopFieldDocs) topDocs;
>             topFieldDocs.add(new TopFieldDocs(
>                 new TotalHits(filteredScoreDocs.length, TotalHits.Relation.EQUAL_TO),
>                 filteredScoreDocs,
>                 original.fields
>             ));
>         }
>     }
>     return topFieldDocs;
> ```

Thanks for the good correction. 👍

wdongyu: > @martin-gaievski @wdongyu first thing first this min score logic should not touch anything related to sorting.
> 
> When there is a sort criteria then the scores are showed as null at the search response. Therefore min_score has nothing to do with sorting.
> 
> Also, this thing is different from other queries. If we apply min score in sorting then for sure shot we will drop results because the end user expecting results as per sort criteria not by any cutoff from the scores.

@vibrantvarun Although scores results with a sort criteria are showed as null, but it indeed first apply min_score and then sort. I will copy the comment I post before:
> This is derived from the behaviour that minScore defined with a normal query and sort. In that case, the normal query will be executed and scored, followed by a minScore filtering, and final with sort.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
wdongyu: If the original relation was already `GREATER_THAN_OR_EQUAL_TO`, for example 10 results. And some filtering occurred:
- if we get 5 results, I think we should use `EQUAL_TO` as we get the exact number. The scores of uncollected docs will definitely be less than the lowest score, also the min_score.
- if we get 10 results,  we should still use `GREATER_THAN_OR_EQUAL_TO` because we can't know the true total.

The logic is the same as the code above.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
wdongyu: > A document with a valid sort value can have score less than min score.

That's really a big difference between hybrid query and regular queries.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
wdongyu: We will not set the value if the user doesn't define the `min_score`.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
wdongyu: We use it for updating the final query top docs:

```
if (newRelation != null) {
    int filteredSize = sortedScores.size();
    compoundQueryTopDocs.setTotalHits(new TotalHits(filteredSize, newRelation));
} else {
    compoundQueryTopDocs.setTotalHits(getTotalHits(topDocsPerSubQuery, maxHits));
}
```

- if newRelation is not null, it means user defines a `min_score`, and we will use the newRelation
- if newRelation is null, it means we don't have a `min_score`, and we will follow the original logic

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
martin-gaievski: yeah, you're correct, no need in additional clause

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
heemin32: If my understanding is correct, here we are overriding the relation. If the original relation is already `EQUAL_TO`, even if `filteredSize == originalSize`, it should be still `EQUAL_TO`

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
heemin32: Inside `getTopFieldDocs` method, it filter doc which does not exist in `combinedNormalizedScoresByDocId`. Why it is required when minScore is defined?

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
wdongyu: I miss this clause, will add it.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
wdongyu: > Inside `getTopFieldDocs` method, it filter doc which does not exist in `combinedNormalizedScoresByDocId`. Why it is required when minScore is defined?

In this situation where minScore is defined with a sort, I find the behaviour in other queries is that it first filters out results using minScore, and then sort. And if minScore is not defined, we just sort docs without filtering.

But according to the discussion in other conversations, as @vibrantvarun @martin-gaievski mentioned, maybe we should only support minScore when sorting with `_score`

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
vibrantvarun: Yes but only for when sort criteria is score and is ascending. For descending opensearch by default does it and it is take care of by the normal query workflow.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
heemin32: No need to add new clause. We can simply set it as null I guess?

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
heemin32: We can block min_score when using sort. Let me ask the following question just to clarify my understanding.

In the code, we don’t explicitly filter documents based on minScore; instead, we filter out any document whose ID is not present in `combinedNormalizedScoresByDocId`. What does this do with filtering based on the minimum score?

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
wdongyu: I will refactor this snippet.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
wdongyu: The minScore is expected to be applied after normalization and combination, the filtered doc ID reflects on the key of `combinedNormalizedScoresByDocId` (which can be derived from the func `updateCombinedNormalizedScores`). So the predicate is used to filter out the doc ID with scores less than the minScore.

The reason why we don't explicitly filter documents based on minScore, is that we can only obtain combination scores from `combinedNormalizedScoresByDocId`, not from original `compoundQueryTopDocs `.

heemin32: Make sense. Thanks.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
wdongyu: @martin-gaievski Do we regard a negative `min_score` as an invalid value?

src/test/java/org/opensearch/neuralsearch/query/HybridQueryIT.java
wdongyu, line 1240: It seems weird that `TEST_MULTI_DOC_INDEX_NAME_ONE_SHARD` has 3 shards, and `TEST_MULTI_DOC_INDEX_NAME` has 1 shard.

heemin32: Yes. It does not seems correct.

owaiskazi19: @wdongyu thanks for pointing this out. Please change it to 1

wdongyu: @heemin32 @owaiskazi19 In this pull, I prefer to keep it unchanged to avoid introducing too many irrelevant test codes. We can later create a new pull request to fix that.

src/main/java/org/opensearch/neuralsearch/search/query/HybridAggregationProcessor.java
```diff
@@ -23,6 +23,12 @@ public class HybridAggregationProcessor implements AggregationProcessor {
     public void preProcess(SearchContext context) {
         // Simply delegate the call
         delegateAggsProcessor.preProcess(context);
+
+        if (isHybridQuery(context.query(), context) && context.minimumScore() != null) {
+            // unset min_score, so it will not work for when executing sub-queries,
+            // and we will retrieve it from search source to filter out final results after normalization and combination
+            context.minimumScore(Float.NEGATIVE_INFINITY);
```
vibrantvarun, line 30: Can't we set this to null? Why NEGATIVE_INFINITY. Later anyway you are fetching the min score value from searchPhaseContext

vibrantvarun: If we set it to null, it will altogether save use from adding extra filter collector.
```
 if (searchContext.minimumScore() != null) {
                // apply the minimum score after multi collector so we filter aggs as well
                collectors.add(createMinScoreCollectorContext(searchContext.minimumScore()));
                // this collector can filter documents during the collection
                hasFilterCollector = true;
            }

File: src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java [hunk 13]
```diff
@@ -142,7 +195,22 @@ private List<TopFieldDocs> getTopFieldDocs(final Sort sort, final List<TopDocs>
             // Check for scoreDocs length.
             // If scoreDocs length=0 then it means that no results are found for that particular subquery.
             if (topDocs.scoreDocs.length != 0) {
-                topFieldDocs.add((TopFieldDocs) topDocs);
+                ScoreDoc[] filteredScoreDocs = topDocs.scoreDocs;
```
vibrantvarun, line 198: It should be FieldDoc []. You may have to typecast it.

src/main/java/org/opensearch/neuralsearch/search/query/HybridAggregationProcessor.java
wdongyu: > Can't we set this to null? Why NEGATIVE_INFINITY. Later anyway you are fetching the min score value from searchPhaseContext

@vibrantvarun We can't because the param for setting minimum score needs to be `float` not `Float`. That's a workaround for setting this `Float.NEGATIVE_INFINITY` to collect all docs with all possible scores.

```
public abstract SearchContext minimumScore(float minimumScore);
```

vibrantvarun: Float is not a primitive data type. It is a object. Can't we set it to null?

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
wdongyu: Yes, it should be `FieldDoc[]`, but we don't access any fields in `FieldDoc` from `filteredScoreDocs`, just keep it unchanged will not lose its actual type. 

```
topFieldDocs.add(
    new TopFieldDocs(
        new TotalHits(filteredScoreDocs.length, TotalHits.Relation.EQUAL_TO), 
        filteredScoreDocs,
        original.fields
    )
);
```

src/main/java/org/opensearch/neuralsearch/search/query/HybridAggregationProcessor.java
wdongyu: > Float is not a primitive data type. It is a object. Can't we set it to null?

The getter return a `Float` type, but setter need a primitive data type `float`. So we can't set it to null.

```
public abstract SearchContext minimumScore(float minimumScore);

public abstract Float minimumScore();
```

src/main/java/org/opensearch/neuralsearch/search/query/HybridAggregationProcessor.java
```diff
@@ -23,6 +23,12 @@ public class HybridAggregationProcessor implements AggregationProcessor {
     public void preProcess(SearchContext context) {
         // Simply delegate the call
         delegateAggsProcessor.preProcess(context);
+
+        if (isHybridQuery(context.query(), context) && context.minimumScore() != null) {
```
vibrantvarun: One last optimization needed. isHybridQuery check is expensive therefore it should only run when minScore is enabled.
```suggestion
        if (context.minimumScore() != null && isHybridQuery(context.query(), context)) {
```

vibrantvarun:
@wdongyu checks are failing

wdongyu:
> @wdongyu checks are failing

@vibrantvarun Not sync from main, I have fixed that.

wdongyu:
> @wdongyu could you also add documentation PR?

you mean updating documentation with min_score usage? I am working on that.

PR #1740, opened 2026-01-27
Fix CollapseTopFieldDocs logic when a segment has no collapsed fieldDocs but has totalHits > 0

vibrantvarun:
### Description
This PR fixes a critical bug in the hybrid search functionality that occurs during concurrent segment search when collapse is enabled. The issue manifests as an IllegalStateException with the message: "Collapse enabled but source TopDocs is not an instance of CollapseTopFieldDocs".

#### Root Cause
The bug occurs in HybridSearchCollectorResultUtil.getNewTopFieldDocs() when processing results from collectors that have processed documents globally (getTotalHits() > 0) but have empty field documents in all sub-query results after collapse/grouping processing.

#### Specific Scenario
- Concurrent segment search creates multiple collectors processing different segments
- Collector A processes documents (increments totalHitCount) but after collapse processing, all sub-queries end up with empty FieldDoc[] arrays
- Collector B processes documents normally and returns valid CollapseTopFieldDocs
- During result reduction, the system attempts to merge results from both collectors

#### When This Happens
- Hybrid queries with collapse functionality enabled
- Concurrent segment search where some segments have no matching documents after collapse processing

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorResultsUtilParams.java
+++ b/src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorResultsUtilParams.java
@@ -42,7 +42,7 @@ public class HybridCollectorResultsUtilParams {
             this.docValueFormats = new DocValueFormat[] { DocValueFormat.RAW };
         }
 
-        if (isSortEnabled || isCollapseEnabled) {
+        if ((isSortEnabled || isCollapseEnabled) && sortFields == null) {
             this.sortFields = DEFAULT_SORT_FIELDS;
         }
         this.rescoreContexts = searchContext.rescore();


src/main/java/org/opensearch/neuralsearch/search/query/util/HybridSearchCollectorResultUtil.java
```diff
@@ -86,13 +86,97 @@ public TopDocsAndMaxScore getTopDocsAndMaxScore() throws IOException {
...
+                sortFields == null ? new SortField[] { new SortField(null, SortField.Type.SCORE) } : sortFields,
+                new Object[0]
+            );
+        }
+
+        // for a single shard case we need to do score processing at coordinator level.
+        // this is workaround for current core behaviour, for single shard fetch phase is executed
+        // right after query phase and processors are called after actual fetch is done
+        // find any valid doc Id, or set it to -1 if there is not a single match
+        int delimiterDocId = collapseTopFieldDocs.stream()
```
martin-gaievski: it's somewhat duplicate of the similar logic in getNewTopFieldDocs(). It's better extracting this into a private helper method:

```java
private int findDelimiterDocId(List<? extends TopDocs> topDocs) {
    return topDocs.stream()
        .filter(Objects::nonNull)
        .filter(topDoc -> Objects.nonNull(topDoc.scoreDocs))
        .map(topDoc -> topDoc.scoreDocs)
        .filter(scoreDoc -> scoreDoc.length > 0)
        .map(scoreDoc -> scoreDoc[0].doc)
        .findFirst()
        .orElse(-1);
}
```

src/main/java/org/opensearch/neuralsearch/search/query/util/HybridSearchCollectorResultUtil.java
```diff
@@ -86,13 +86,97 @@ public TopDocsAndMaxScore getTopDocsAndMaxScore() throws IOException {
...
+        final List<TopFieldDocs> collapseTopFieldDocs,
+        final SortField sortFields[]
+    ) {
+        String collapseField = hybridSearchCollectorResultsDTO.getSearchContext().collapse().getFieldName();
+        if (Objects.isNull(collapseTopFieldDocs)) {
+            return new CollapseTopFieldDocs(
+                collapseField,
+                totalHits,
+                new FieldDoc[0],
+                sortFields == null ? new SortField[] { new SortField(null, SortField.Type.SCORE) } : sortFields,
```
martin-gaievski: default sort field creation is repeated multiple times, extract to constant:

```java
private static final SortField[] DEFAULT_SCORE_SORT = new SortField[] { new SortField(null, SortField.Type.SCORE) };
```

src/main/java/org/opensearch/neuralsearch/search/query/util/HybridSearchCollectorResultUtil.java
```diff
@@ -127,67 +190,25 @@ private TopDocs getNewTopFieldDocs(final TotalHits totalHits, final List<TopFiel
...
+            List<FieldDoc> fieldDocsPerQuery = new ArrayList<>();
+            for (ScoreDoc scoreDoc : topFieldDoc.scoreDocs) {
+                fieldDocsPerQuery.add((FieldDoc) scoreDoc);
+            }
+            result.add(createFieldDocDelimiterElementForHybridSearchResults(delimiterDocId, sortFieldsForDelimiterResults));
+            result.addAll(fieldDocsPerQuery);
         }
+        result.add(createFieldDocStartStopElementForHybridSearchResults(delimiterDocId, sortFieldsForDelimiterResults));
+        FieldDoc[] fieldDocs = result.toArray(new FieldDoc[0]);
+        return new TopFieldDocs(totalHits, fieldDocs, sortFields);
```
bzhangam, line 198: Why here we use `sortFields` directly without null check?

vibrantvarun: Because when sorting is enabled or collapse is enabled then sortFields object will come. 

when sort is enabled. It will be TopFieldDocs

1. https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorResultsUtilParams.java#L46

When collapse is enabled. It will be CollapseTopFieldDocs

2. https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollectorFactory.java#L53

vibrantvarun: In case of valid results, we always get sortfields https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopFieldDocSortCollector.java#L114

bzhangam: Since when `sortFields` is null we always want to use `DEFAULT_SORT_FIELDS` let's move this logic to `getSortedTopDocsAndMaxScore` so that we don't need worry about it in the following steps.

src/main/java/org/opensearch/neuralsearch/search/query/util/HybridSearchCollectorResultUtil.java
```diff
@@ -86,35 +87,97 @@ public TopDocsAndMaxScore getTopDocsAndMaxScore() throws IOException {
...
-        if (Objects.isNull(topFieldDocs)) {
-            return new TopFieldDocs(totalHits, new FieldDoc[0], sortFields);
+    private TopDocs getCollapseTopFieldDocs(
+        final TotalHits totalHits,
+        final List<TopFieldDocs> collapseTopFieldDocs,
+        final SortField sortFields[]
+    ) {
+        String collapseField = hybridSearchCollectorResultsDTO.getSearchContext().collapse().getFieldName();
+        if (Objects.isNull(collapseTopFieldDocs)) {
+            return new CollapseTopFieldDocs(
```
bzhangam: nit: Normalize empty-result creation for CollapseTopFieldDocs and TopFieldDocs can make the code cleaner.

PR #1742, opened 2026-01-29
Fix array_index_out_of_bound_exception in case of docsPerGroupPerSubQuery greater or lesser than numHits

vibrantvarun:
### Description
Earlier, we introduced [`index.neural_search.hybrid_collapse_docs_per_group_per_subquery` setting](https://github.com/opensearch-project/neural-search/pull/1471) to control the collection of results per collapsegroup. But, it throws ArrayIndexOutOfBoundsException when setting is not set on default (i.e.=0) and have value either lower or greater than numHits.

The issue happens due to collector collect the docId as per setting value but the logic of queue size full validates it with numHits. In other words, the documents will keep on adding in the queue until queue is full. That queue size will be equal to setting value. But to check whether the queue is full or not the logic will be validated as per numHits aka size rather than setting value.

### Check List

### DIFF
--- a/src/test/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollectorTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollectorTests.java
@@ -356,7 +356,7 @@ public class HybridCollapsingTopDocsCollectorTests extends HybridCollectorTestCa
             sort,
             5,  // topNGroups
             new HitsThresholdChecker(TOTAL_HITS_UP_TO),
-            1   // docsPerGroupPerSubQuery = 1 (the problematic value)
+            10   // docsPerGroupPerSubQuery = 1 (the problematic value)
         );
 
         Weight weight = mock(Weight.class);
@@ -383,10 +383,10 @@ public class HybridCollapsingTopDocsCollectorTests extends HybridCollectorTestCa
             assertEquals(20, collapseTopFieldDocs.totalHits.value());
 
             // Key validation: Despite 20 documents in same group, only 1 should be returned
-            assertEquals("Should have exactly 1 document due to docsPerGroupPerSubQuery=1", 1, collapseTopFieldDocs.scoreDocs.length);
+            assertEquals("Should have exactly 10 document due to docsPerGroupPerSubQuery=10", 10, collapseTopFieldDocs.scoreDocs.length);
 
             // Verify the collapse value
-            assertEquals("Should have exactly 1 collapse value", 1, collapseTopFieldDocs.collapseValues.length);
+            assertEquals("Should have exactly 10 collapse value", 10, collapseTopFieldDocs.collapseValues.length);
             assertEquals("samegroup", ((BytesRef) collapseTopFieldDocs.collapseValues[0]).utf8ToString());
         }
 


src/test/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollectorTests.java
```diff
@@ -89,7 +89,7 @@ public void testKeywordCollapse_whenCollectAndTopDocs_thenSuccessful() throws IO
 
         for (CollapseTopFieldDocs collapseTopFieldDocs : topDocs) {
             assertEquals(1000, collapseTopFieldDocs.totalHits.value());
-            assertEquals(TOP_N_GROUPS * TOP_N_GROUPS, collapseTopFieldDocs.scoreDocs.length);
+            assertEquals(DOCS_PER_GROUP_PER_SUBQUERY * TOP_N_GROUPS, collapseTopFieldDocs.scoreDocs.length);
```
vibrantvarun, line 92: Document per subquery x numHits

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -470,7 +470,7 @@ private void addNewEntry(T groupValue, int subQueryNumber, int doc, float score,
                 updateMaps(comparators, fieldValueLeafTrackers, compoundScores);
 
                 // Check if the queue is full for this subquery
-                if (slot == (numHits - 1)) {
+                if (slot == (docsPerGroupPerSubQuery - 1)) {
                     boolean[] queueFullArray = queueFullMap.get(groupValue);
                     queueFullArray[subQueryNumber] = true;
                     queueFullMap.put(groupSelector.copyValue(), queueFullArray);
```
bzhangam, line 476: Wonder why we use `groupValue` to get the data but use `groupSelector.copyValue()` to set the value?

src/test/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollectorTests.java
```diff
@@ -227,6 +227,174 @@ public void testCollapse_whenZeroDocsPerGroupPerSubQuery_thenSuccessful() throws
...
+            // The key assertion: no ArrayIndexOutOfBoundsException was thrown during collection
+            // This validates that the queue full detection fix works correctly
+        }
+
+        reader.close();
+        writer.close();
+        directory.close();
+    }
+
+    public void testCollapse_whenDocsPerGroupPerSubQueryEqualsOne_thenQueueFullDetectionWorks() throws IOException {
```
bzhangam, line 330: What's the diff between this test and the above one? I think both test the case that `docsPerGroupPerSubQuery` is less than `topNGroups` and the queue full detection should work. Should we just keep one?

Also wonder if `docsPerGroupPerSubQuery` is larger than `topNGroups` how will it work? And do we need to add a test case for it?

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
vibrantvarun: Yes you are correct here. But I do not want to touch it in scope of this PR as collapse feature is already tested. We can change it another dedicated PR. Because at the critical release time I do not want to experiment with it.

vibrantvarun: Also the change would be required at earlier places as well. So definitely after removing a proper testing would be needed.

```
private void initializeLeafComparators(T groupValue, int numSubQueries, HybridSubQueryScorer compoundQueryScorer)
                throws IOException {
                LeafFieldComparator[] comparators = comparatorsMap.get(groupValue);
                FieldValueHitQueue<FieldValueHitQueue.Entry>[] compoundScores = groupQueueMap.get(groupValue);

                for (int subQueryNumber = 0; subQueryNumber < numSubQueries; subQueryNumber++) {
                    LeafFieldComparator[] leafFieldComparators = compoundScores[subQueryNumber].getComparators(context);
                    int[] reverseMuls = compoundScores[subQueryNumber].getReverseMul();

                    if (leafFieldComparators.length == 1) {
                        reverseMulMap.put(groupSelector.copyValue(), reverseMuls[0]);
                        comparators[subQueryNumber] = leafFieldComparators[0];
                    } else {
                        reverseMulMap.put(groupSelector.copyValue(), 1);
                        comparators[subQueryNumber] = new MultiLeafFieldComparator(leafFieldComparators, reverseMuls);
                    }
                    comparators[subQueryNumber].setScorer(compoundQueryScorer);
                }

                comparatorsMap.put(groupSelector.copyValue(), comparators);
                initializeLeafComparatorsPerSegmentOnceMap.put(groupSelector.copyValue(), false);
            }
```

martin-gaievski: +1, let's stick to the focused scope in the PR, any refactoring or change in approach can be done later

heemin32 (approved):

bzhangam:
I'm good to merge this PR and we can address the comment in another PR.

PR #1738, opened 2026-01-26
Fixed runtime error when number of shards greater than default batch reduce size

martin-gaievski:
### Description
Added request filter that is programmatically increase the number of batched reduce shards setting  DEFAULT_BATCHED_REDUCE_SIZE/512. This way the merge of Shard results by the opensearch core is delayed. This is critical for hybrid query with current design because such merges corrupt the internal format of results (multiple collections per sub-query) and lead to runtime exception (see original issue for detailed scenario). 
Request filter works at a single search request level, no global changes so other query types are not effected. Side effect of this change can be an increase in latency for hybrid query when index has more than 512 shards.  

I've manually tested scenario described in original GH issue. Automation is not feasible because it's failing with such high number of shards.

```
{
    "took": 64,
    "timed_out": false,
    "_shards": {
        "total": 513,
        "successful": 513,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "total": {
            "value": 93,
            "relation": "eq"
        },
        "max_score": 1.0,
        "hits": [
            {
                "_index": "test_index",
                "_id": "EhFg_JsBpmoBypRqgN4-",
                "_score": 1.0,
                "_source": {
                    "id": "s100",
                    "passage_text": "Garibaldian quickset",
                    "imdb": 9.8,
                    "actor": "jackie"
                }
            },
            {
                "_index": "test_index",
                "_id": "ExFg_JsBpmoBypRqgN4-",
                "_score": 1.0,
                "_source": {
                    "id": "s101",
                    "passage_text": "Garibaldian quickset quickset",
                    "imdb": 10,
                    "actor": "jackie"
                }
            },....
}
```

### Related Issues
Resolves https://github.com/opensearch-project/neural-search/issues/1733

### Check List
- [X] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [X] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/neural-search/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #1733, opened 2026-01-19
[BUG] hybrid search on a index or alias with more 512 shards causes exception


wdongyu:
### What is the bug?
Run a hybrid search on a index or alias, which has more than 512 shards, will cause exception `topDocs already consumed`.

### How can one reproduce the bug?
1. create a index with 513 shards
```
PUT test_index
{
  "mappings": {
    "properties": {
      "imdb": {
        "type": "float"
      },
      "actor": {
        "type": "keyword"
      }
    }
  },
  "settings": {
    "number_of_shards": 513
  }
}
```
2. bulk data
```
POST _bulk?refresh
{"index": {"_index": "test_index"}}
{"id": "s9", "passage_text": "Stenocarpus mynpachtbrief", "imdb": 7.7, "actor": "jackie"}
{"index": {"_index": "test_index"}}
{"id": "s10", "passage_text": "weirangle meritorious", "imdb": 2.5, "actor": "anil"}
{"index": {"_index": "test_index"}}
{"id": "s11", "passage_text": "clotbur plagiostome", "imdb": 2.5, "actor": "ranbir"}
{"index": {"_index": "test_index"}}
{"id": "s12", "passage_text": "guider perityphlic", "imdb": 8.0, "actor": "anil"}
{"index": {"_index": "test_index"}}
{"id": "s13", "passage_text": "amidoazobenzene bibliopegy nymphosis", "imdb": 2.5, "actor": "sanjay"}
{"index": {"_index": "test_index"}}
{"id": "s14", "passage_text": "sepiarian antipode unpadded", "imdb": 9.8, "actor": "anil"}
{"index": {"_index": "test_index"}}
{"id": "s15", "passage_text": "portraiture hemihyperidrosis mongery", "imdb": 8.0, "actor": "salman"}
{"index": {"_index": "test_index"}}
{"id": "s16", "passage_text": "pseudoasymmetrical glucolipide nonangling", "imdb": 6.6, "actor": "abhishek"}
{"index": {"_index": "test_index"}}
{"id": "s17", "passage_text": "supraseptal snitch", "imdb": 9.9, "actor": "ranveer"}
{"index": {"_index": "test_index"}}
{"id": "s18", "passage_text": "ultragaseous factious", "imdb": 9.9, "actor": "salman"}
{"index": {"_index": "test_index"}}
{"id": "s19", "passage_text": "shrewstruck redemptor uninquisitive", "imdb": 9.8, "actor": "anil"}
[...]

PR #1745, opened 2026-01-30
Fix null instance handling for empty and skipped shards

martin-gaievski:
### Description
This fix addresses scenario when shard does not return any results and as a result hybrid search is failing with runtime exception. This is special/edge case when shard result is `QuerySearchResult.nullInstance()`. Can happen in production for following reasons:

- Empty shards, shard has 0 documents
- No matching documents
- Query rewrite to match_no_docs
- Skipped shards, pre-filter optimization skips entire shard
- Index with sparse data, most shards empty, few have data
- Routing mismatch, custom routing can exclude shards

Unlike the issue fixed in https://github.com/opensearch-project/neural-search/pull/1738 this can happen with any number of shards.

Fix is to filter out null instance results before processing, regardless of why they occurred

### Related Issues
https://github.com/opensearch-project/neural-search/issues/1733

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - [HYBRID]: Fix runtime error when number of shards greater than default batch reduce size ([#1738](https://github.com/opensearch-project/neural-search/pull/1738))
 - [HYBRID]: Fix CollapseTopFieldDocs logic when a segment has no collapsed fieldDocs but has totalHits > 0 ([#1740](https://github.com/opensearch-project/neural-search/pull/1740))
 - [HYBRID]: Fix array_index_out_of_bound_exception in case of docsPerGroupPerSubQuery greater or lesser than numHits ([#1742](https://github.com/opensearch-project/neural-search/pull/1742))
+- [HYBRID]: Fix null instance handling for empty and skipped shards ([#1745](https://github.com/opensearch-project/neural-search/pull/1745))
 
 ### Infrastructure
 - [BWC]: Enable BWC tests after upgrading to Grade 9 ([#1729](https://github.com/opensearch-project/neural-search/pull/1729))
--- a/src/test/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflowTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflowTests.java
@@ -770,13 +770,16 @@ public class NormalizationProcessorWorkflowTests extends OpenSearchTestCase {
         nullInstanceResult.setShardIndex(1);
         querySearchResults.add(nullInstanceResult);
 
+        // Third element is an actual null in the list - tests the Objects.nonNull() check
+        querySearchResults.add(null);
+
         SearchPhaseContext searchPhaseContext = mock(SearchPhaseContext.class);
         SearchRequest searchRequest = mock(SearchRequest.class);
         SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
         searchSourceBuilder.from(0);
         when(searchPhaseContext.getRequest()).thenReturn(searchRequest);
         when(searchRequest.source()).thenReturn(searchSourceBuilder);
-        when(searchPhaseContext.getNumShards()).thenReturn(2);
+        when(searchPhaseContext.getNumShards()).thenReturn(3);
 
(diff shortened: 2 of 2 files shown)


src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -405,4 +414,19 @@ private List<Integer> unprocessedDocIds(final List<QuerySearchResult> querySearc
...
+    /**
+     * Filters out null instance results from the query search results.
+     * Null instances occur when:
+     * 1. A shard has no matching documents (match_no_docs)
+     * 2. A shard's query was rewritten to empty
+     * 3. Partial reduce consumed the topDocs during batched shard processing (>512 shards)
+     * @param querySearchResults the list of QuerySearchResult from all shards
+     * @return list containing only valid (non-null) results
+     */
+    private List<QuerySearchResult> filterValidResults(final List<QuerySearchResult> querySearchResults) {
```
vibrantvarun, line 427: @martin-gaievski If we filter null results, then I think it will
 disrupt the ranking of list in the List QuerySearchResult .

For example we use for loop to get shardId at many places. if we remove this null result then that for loop will disrupt the shard Id logic. Because in multi shard scenario we don't know the precise shardId. We use reference shardId with positioning logic. That positioning logic will fail.

I think we need an integ test here.

vibrantvarun: For example folliowing logics https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java#L99

martin-gaievski:
> @martin-gaievski a naive question how can QuerySearchResult be null if we have this consumedDocs check which checks if topDocsAndMaxScore is null then handle it by creating non null QuerySearchResult https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/search/query/util/HybridSearchCollectorResultUtil.java#L60 ?

Good question, thank you. There are two different code paths in different phases:

- `HybridSearchCollectorResultUtil` (data node - query phase). This runs on the data node during query phase when our HybridQueryPhaseSearcher merges segment results. It correctly creates a valid QuerySearchResult with TopDocsAndMaxScore set.
- partial reduce (coordinator node - after query phase)

This bug occurs after query phase completes, when results are being collected on the coordinator node:

The code you've mentioned, `HybridSearchCollectorResultUtil` runs in the neural-search plugin on data nodes, while partial reduce runs in OpenSearch core on the coordinator - they're completely different execution contexts.

check out these key classes in OpenSearch core

- [QueryPhaseResultConsumer](org.opensearch.action.search.QueryPhaseResultConsumer) - does partial reduce
- [QuerySearchResult.consumeTopDocs()](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/query/QuerySearchResult.java#L177-L184) - nullifies topDocs

[here](https://github.com/opensearch-project/neural-search/issues/1733#issuecomment-3825283223) you can find real life stack trace reported from live opensearch system

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
martin-gaievski: No, it will not fail, this is incorrect assumption. Fix is safe because referenceShardId is not the actual OpenSearch shard ID - it's a loop index used as a discriminator. Real shard info is preserved , actual shard information comes from `SearchShard`, which is built from `QuerySearchResult.getSearchShardTarget()`:

```java
SearchShard searchShard = compoundQueryTopDocs.getSearchShard();
```

Why the fix is safe

```
Before filtering:
  querySearchResults = [Result0, NullInstance, Result2]
  
After filtering:  
  querySearchResults = [Result0, Result2]
  queryTopDocs      = [CompoundTopDocs0, CompoundTopDocs1]  // From filtered list
  
Both loops iterate [0, 1]:
  - sortDocuments: builds map with keys "docId_0", "docId_1"
  - processTopDocs: looks up same keys "docId_0", "docId_1"
```

Positions match because same list is used consistently. 

Adding new integ test will increase confidence but to me it's out of scope for this change, I've done some research in scope of https://github.com/opensearch-project/neural-search/pull/1738 and found that test clusters with large number of shards are failing. It will take time to find a proper working solution for this problem.

PR #1753, opened 2026-02-06
Fix Hybrid search with Collapse relevancy bugs

vibrantvarun:
### Description
This PR fixes following logic in Hybrid search with collapse functionality
1. In the earlier logic, if the document is not a part of subquery and has 0 score reported in the collector's collect method then also it was adding in the result list. Therefore, disrupting normalization logic by considering those documents as well which are not valid results. In the new logic, we add a check of score==0 in the collect method to add only valid documents
2. In the earlier logic, if the hybrid search without collapse results a result as top document then after applying collapse it does not return the most relevant result as top document. The reason was while adding values in collapseValueToTopDocMap (i.e a HashMap) use to disrupt the order. Therefore, now we have switched to LinkedHashMap.
3. In the earlier logic, the collapseStrategy was used to determined twice once in CollapseExecuter and once in CollapseDataCollector. Determining the strategy is a expensive operation, therefore, in this PR we have optimized it by doing it once in the starting itself.

### Check List

### DIFF
--- a/src/test/java/org/opensearch/neuralsearch/processor/collapse/CollapseExecuterTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/processor/collapse/CollapseExecuterTests.java
@@ -335,47 +335,4 @@ public class CollapseExecuterTests extends OpenSearchTestCase {
         // Should have 2 groups: one for 100L and one for null
         assertEquals(2, result);
     }
-
-    /**
-     * Test executeCollapse verifies that BytesRef uses KeywordStrategy
-     * and any other type (including null) uses NumericStrategy
-     */
-    public void testExecuteCollapse_whenStringFieldType_thenUsesNumericStrategy() {
-        // String.class is not BytesRef.class, so it should use NumericStrategy
-        // This test verifies the logic: only BytesRef.class uses KeywordStrategy
-        CollapseTopFieldDocs collapseTopFieldDocs = new CollapseTopFieldDocs(
-            "text_field",
-            new TotalHits(2, TotalHits.Relation.EQUAL_TO),
-            new ScoreDoc[] {
-                new FieldDoc(1, 1.0f, new Object[] { 1.0f, "value1" }),
-                new FieldDoc(2, 0.8f, new Object[] { 0.8f, "value2" }) },
-            new SortField[] { SortField.FIELD_SCORE },
-            new Object[] { "value1", "value2" }
-        );
-
-        CompoundTopDocs compoundTopDocs = new CompoundTopDocs(
-            new TotalHits(2, TotalHits.Relation.EQUAL_TO),
-            List.of(collapseTopFieldDocs),
-            true,
-            SEARCH_SHARD
-        );
-
-        List<CompoundTopDocs> collapseQueryTopDocs = List.of(compoundTopDocs);
-        List<QuerySearchResult> collapseQuerySearchResults = List.of(mock(QuerySearchResult.class));
-        Sort collapseSort = new Sort(SortField.FIELD_SCORE);
-        CombineScoresDto combineScoresDto = mock(CombineScoresDto.class);
-
-        CollapseDTO collapseDTO = new CollapseDTO(
-            collapseQueryTopDocs,
-            collapseQuerySearchResults,
-            collapseSort,
-            true,
-            combineScoresDto,
-            String.class  // String.class is not BytesRef.class, so uses NumericStrategy
-        );
-
-        // Execute collapse - should use NumericStrategy (not KeywordStrategy)
-        int result = collapseExecutor.executeCollapse(collapseDTO);
-        assertEquals(2, result);
-    }
 }


src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseDataCollector.java
```diff
@@ -29,7 +30,7 @@
 @Log4j2
 public class CollapseDataCollector<T> {
 
-    private final Map<T, FieldDoc> collapseValueToTopDocMap = new HashMap<>();
+    private final Map<T, FieldDoc> collapseValueToTopDocMap = new LinkedHashMap<>();
```
vibrantvarun, line 33: Need to switch to LinkedHashMap to preserve the order of documents after collapse. 

Earlier:

Before collapse if order is following
```
docId 1 score 0.5 collapsefield ChoclateCake
docId 7 score 0.5 collapsefield ChoclateCake
docId 2 score 0.5 collapsefield Vanilla cake
docId 3 score 0.5 collapseField null
```
After collapse it used to become. It use to not preserve the order on same score.
```
docId 3 score 0.5 collapseField null
docId 2 score 0.5 collapsefield Vanilla cake
docId 1 score 0.5 collapsefield ChoclateCake
```

Now after switching to LinkedHashMap collapse it returns the same order.
```
docId 1 score 0.5 collapsefield ChoclateCake
docId 2 score 0.5 collapsefield Vanilla cake
docId 3 score 0.5 collapseField null
```

src/test/java/org/opensearch/neuralsearch/e2e/HybridCollapseIT.java
```diff
@@ -55,9 +56,8 @@ public void testCollapse_withSingleShard_thenSuccessful() {
         testCollapse_whenE2E_andSortEnabled_thenSuccessful();
         testCollapse_whenE2EWithInnerHits_thenSuccessful();
 
-        // For min_score=0.5005f, it filters out no docs; for min_score=1.0f, it filters out 1 doc.
-        testCollapse_whenE2E_withMinScore_thenSuccessful(0.5005f, 2, 3);
-        testCollapse_whenE2E_withMinScore_thenSuccessful(1.0f, 1, 2);
```
vibrantvarun, line 60: Earlier due to scoring logic was not working as expected so this integ test was valid but as now we fixed the logic after that we don't need testCollapse_whenE2E_withMinScore_thenSuccessful(1.0f, 1, 2); test case

vibrantvarun:
@martin-gaievski @bzhangam Could you guys review the PR?

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchCollapseUtil.java
```diff
@@ -0,0 +1,79 @@
...
+        outer: for (CompoundTopDocs queryTopDoc : queryTopDocs) {
+            // Handle null CompoundTopDocs entries
+            if (queryTopDoc == null) {
+                continue;
+            }
+
+            List<TopDocs> topDocsList = queryTopDoc.getTopDocs();
+            if (topDocsList != null && !topDocsList.isEmpty()) {
+                for (TopDocs topDocs : topDocsList) {
+                    CollapseTopFieldDocs collapseTopFieldDocs = (CollapseTopFieldDocs) topDocs;
```
martin-gaievski, line 58: just to make sure - is it safe to typecast here directly without checking instanceof?

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseDataCollector.java
```diff
@@ -117,9 +99,16 @@ private void processCollapseDoc(ScoreDoc scoreDoc, int shardIndex) {
             );
         }
 
-        if (fieldDoc.fields == null || fieldDoc.fields.length == 0) {
-            log.info("Field doc 'fields' attribute does not contain any values");
-            return;
+        if (fieldDoc.fields == null) {
+            throw new IllegalStateException(
```
martin-gaievski, line 103: do we really need to fail the whole query in this scenario? from the end user perspective can they fix this somehow, is it better to give partially incorrect results vs failing and not giving the result at all?

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseExecutor.java
```diff
@@ -22,19 +21,19 @@ public class CollapseExecutor {
      * @return The total number of documents that were collapsed during the operation
      */
     public int executeCollapse(CollapseDTO collapseDTO) {
-        boolean isKeywordCollapse = isCollapseOnKeywordField(collapseDTO);
-        CollapseStrategy collapseStrategy = isKeywordCollapse
+        Class<?> collapseFieldType = collapseDTO.getCollapseFieldType();
+        /**
+         * Ideally, the fieldType should be derived from CollapseContext but currently there is no way of providing searchContext in phase results processor.
+         * Through searchPhaseContext we can only get collapseBuilder which only contains the fieldName. To create collapseContext from builder we need QueryShardContext which is not present in phaseResultsProcessor.
+         * As per current code, if the group value is null when the document does not have collapse field then we cannot determine the field type. Therefore, we always follow Numeric strategy in that case.
```
martin-gaievski, line 28: curious why NumericStrategy is the right default

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseDTO.java
```diff
@@ -34,31 +33,32 @@ public class CollapseDTO {
     private String collapseField;
     private CompoundTopDocs updatedCollapseTopDocs;
     private int collapseShardIndex;
+    private Class<? extends Object> collapseFieldType;
```
martin-gaievski: nit: simpler `Class<?>` should work the same

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
```diff
@@ -105,6 +105,7 @@ public void execute(final NormalizationProcessorWorkflowExecuteRequest request)
             .fromValueForSingleShard(getFromValueIfSingleShard(request))
             .isSingleShard(isSingleShard)
             .minScore(minScore)
+            .isCollapseEnabled(request.searchPhaseContext.getRequest().source().collapse() != null)
```
martin-gaievski, line 108: nice improvement. I hope source() cannot return null

src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchCollapseUtil.java
vibrantvarun: I mean for unit test I can check that condition. But as now collapse code path is entirely dedicated to collapse so it should not be a problem.

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseDataCollector.java
vibrantvarun: Here the logic is little tricky.

Initially we get CollapseTopFieldDocs. That object has 2 important objects FieldDoc and CollapseValues.

FieldDoc has sortValues

Now during Normalization phase ScoreCombination process we append collapseValues at the end of sort fieldValues. This we do because later in the CollapseDataCollector where actual collapse logic happens we need this collapseValues to merge FieldDocs. It is a tiny hack we did during initial feature launch to avoid extra computation and keep the logic simple.

Now in our case the fields length has to be minimum 2 provided the sort criteria is score and one collapse value (even it is null then it will null value collapse array). So for collapse to function we need fields. There is no partially incorrect result. Moreover, this is just a safety check for unit test. In real, this scenario will not happen.

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseExecutor.java
vibrantvarun: Earlier also it was Numeric strategy. The only thing was it was silent. 

TBH, it can be any strategy. The logic is running generic behind the scenes and both handle Objects. So they can be null.

src/main/java/org/opensearch/neuralsearch/processor/NormalizationProcessorWorkflow.java
vibrantvarun: Yeah. It Initializes before calling normalization phase so we are good here.

src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseDataCollector.java
vibrantvarun: Code References:

https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java#L240

vibrantvarun: In case, if sortfield is not present in document and also collapseValue is also not present then the response will be as follows

In following example, The collapse is is applied on item field and sort on price field. Both are not present in the document. 

```
  {
                "_index": "collapse_test",
                "_id": "3",
                "_score": null,
                "_source": {
                    "category": "cakes",
                    "price": 12,
                    "item": "Vanilla Cake"
                },
                "fields": {
                    "item": [
                        "Vanilla Cake"
                    ]
                },
                "sort": [
                    12.0
                ]
            },
            {
                "_index": "collapse_test",
                "_id": "5",
                "_score": null,
                "_source": {
                    "category": "cakes"
                },
                "sort": [
                    "-Infinity"
                ]
            }

File: src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchCollapseUtil.java (hunk 3 above)
Comment (vibrantvarun) [reply src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchCollapseUtil.java:58]: Added the check now.

Comment (martin-gaievski) [reply src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchCollapseUtil.java:58]: ack

File: src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseExecutor.java (hunk 5 above)
Comment (martin-gaievski) [reply src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseExecutor.java:28]: I see, let's keep the numeric then

File: src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseDataCollector.java (hunk 4 above)
Comment (martin-gaievski) [reply src/main/java/org/opensearch/neuralsearch/processor/collapse/CollapseDataCollector.java:103]: makes sense, thx for explanation

File: src/main/java/org/opensearch/neuralsearch/search/util/HybridSearchCollapseUtil.java [hunk 8]
```diff
@@ -0,0 +1,80 @@
...
+                }
+            }
+        }
+
+        if (collapseValue instanceof BytesRef) {
+            // Keyword Type
+            return BytesRef.class;
+        } else if (collapseValue instanceof Long) {
+            // Numeric Type
+            return Long.class;
```
bzhangam, line 75: `collapseValue` only can be `BytesRef` and `Long`? If not do we intend to return null for the unknown type?

vibrantvarun: yes. It will be a null group.

Collapse feature only works for Keyword fields and Numeric fields. If the field is not present in the document then it will be categorize in null category.

martin-gaievski (approved):

PR #1762, opened 2026-02-18
Add size-based (byte-aware) batching for neural search ingest processors

### LINKED ISSUE #1761: [PROPOSAL] Size based batching
### What are you proposing?

Introduce size-based (byte-aware) batching for neural search ingest processors (`TextEmbeddingProcessor`, `SparseEncodingProcessor`, and `SemanticFieldProcessor`) as an alternative to the current static document-count-based `batch_size`. Instead of batching a fixed number of documents regardless of their content size, the processor would accumulate documents into a batch until a configurable byte threshold (`batch_size_bytes`) is reached. This ensures that the actual payload sent to ML Commons for inference remains predictable and bounded, regardless of individual document sizes.

A hybrid mode is also proposed where both `batch_size` (max document count) and `batch_size_bytes` (max byte size) can be configured together — whichever limit is hit first triggers the batch to be dispatched.

- The original batch ingestion RFC ([neural-search#598](https://github.com/opensearch-project/neural-search/issues/598)) and the OpenSearch core RFC ([OpenSearch#12457](https://github.com/opensearch-project/OpenSearch/issues/12457)) both discuss batch size configuration but only in terms of document count. Community feedback on these RFCs has highlighted that static document counts do not account for payload variance.
- [neural-search#743](https://github.com/opensearch-project/neural-search/issues/743) implemented batch ingestion for `TextEmbeddingProcessor` and `SparseEncodingProcessor` with a static `batch_size`, and benchmarks showed 77% ingestion time reduction. However, users with heterogeneous document sizes report suboptimal batching — either hitting ML Commons memory limits with large documents or underutilizing capacity with small ones.
- The analogous problem is well-documented in the broader search ecosystem. For example, [Elasticsuite#3239](https://github.com/Smile-SA/elasticsuite/issues/3239) describes how batches exceeding size limits cause indexing failures, forcing users to either over-provision infrastructure or drastically reduce batch sizes.
- ML service providers (OpenAI, SageMaker, Bedrock, Cohere) impose rate limits based on tokens/bytes per request, not document count. A size-aware batching strategy aligns more naturally with these upstream constraints.

- When **ingesting documents of varying sizes through a neural search pipeline**, a **user** wants **the batch size to adapt to the actual data volume** so they **don't hit ML Commons memory limits or remote model rate limits with oversized batches, and don't underutilize throughput with undersized batches**.

- When **bulk-indexing a corpus with heterogeneous document lengths** (e.g., a mix of short product titles and long article bodies), a **cluster administrator** wants **predictable and bounded inference request sizes** so they **can capacity-plan their ML nodes and remote model endpoints without needing to tune batch_size per index or document type**.

- When **using remote models with token-based or byte-based rate limits** (e.g., OpenAI, SageMaker, Bedrock), a **developer** wants **the ingest processor to respect payload size constraints automatically** so they **avoid throttling errors (HTTP 429) that cause ingestion failures and retries**.

- When **operating a multi-tenant cluster where different indices have different document profiles**, an **operations engineer** wants **a single batching configuration that works well across all indices** so they **don't need to maintain per-index pipeline configurations with different batch_size values**.

The feature introduces a new optional parameter `batch_size_bytes` to neural search ingest processors. It works alongside the existing `batch_size` parameter.

**Pipeline Configuration (REST API):**

```json
PUT _ingest/pipeline/my-embedding-pipeline
{
  "processors": [
    {
      "text_embedding": {
        "model_id": "my-model-id",
        "field_map": {
        },
        "batch_size": 100,
        "batch_size_bytes": 102400
      }
    }
  ]
}
```

| Parameter | Type | Default | Description |
|---|---|---|---|
| `batch_size` | integer | Existing default (from `AbstractBatchingProcessor`) | Maximum number of documents per batch. Existing behavior, unchanged. |
| `batch_size_bytes` | integer | `-1` (disabled) | Maximum cumulative size in bytes of inference texts per batch. When set to a positive value, enables size-based batching. |

**Behavior:**
- If only `batch_size` is set: current behavior, unchanged.
- If only `batch_size_bytes` is set: batches are formed purely by byte size, with a floor of 1 document per batch (a single document that exceeds the byte limit is still sent as its own batch).
- If both are set: whichever limit is reached first triggers the batch to be dispatched. This is the recommended hybrid mode.
- The byte measurement is taken on the `inferenceList` (the actual text strings sent to ML Commons), not the entire document payload. This accurately reflects the size of the inference request.

**Index-level setting for semantic fields:**

A corresponding index-level setting will be added:

```
index.neural_search.semantic_ingest_batch_size_bytes
```

This mirrors the existing `index.neural_search.semantic_ingest_batch_size` setting for the `SemanticFieldProcessor`.

**No changes to existing REST APIs.** The `_ingest/pipeline` API already accepts arbitrary processor parameters. This is purely an additive configuration option.

No new security considerations. The feature does not introduce new APIs, endpoints, or authentication flows. It operates within the existing ingest pipeline security model. The byte threshold is a local configuration parameter that does not affect data access patterns.

One consideration: excessively large `batch_size_bytes` values could lead to high memory usage on ingest nodes. The implementation should enforce a reasonable upper bound and document recommended values.

#### Are there any breaking changes to the API?

No. This is a purely additive change. The new `batch_size_bytes` parameter is optional and defaults to `-1` (disabled), preserving existing behavior for all current users. Existing pipelines with only `batch_size` configured will continue to work identically.

1. **Default experience (no change):** Users who don't set `batch_size_bytes` see identical behavior to today. The static `batch_size` continues to control batching.

2. **Size-based batching:** Users set `batch_size_bytes` in their processor configuration. During bulk ingestion, the processor accumulates documents into a batch, measuring the cumulative byte size of the inference texts. When the byte threshold is reached, the batch is dispatched to ML Commons. This naturally produces smaller batches for large documents and larger batches for small documents.

3. **Hybrid batching (recommended):** Users set both `batch_size` and `batch_size_bytes`. The processor dispatches a batch when either limit is reached first. This provides dual safety bounds — preventing both excessively large payloads and excessively large document counts.

4. **Observability:** Batch dispatch metrics (batch count, average batch size in docs, average batch size in bytes) should be exposed through the existing neural search stats API (`_plugins/_neural_search/stats`) to help users tune their configuration.

**Example scenarios with `batch_size: 100, batch_size_bytes: 102400` (100KB):**

| Document profile | Effective batch size | Limiting factor |
|---|---|---|
| 100 docs × 500 bytes each (50KB total) | 100 docs | `batch_size` hit first |
| 5 docs × 25KB each (125KB total) | 4 docs | `batch_size_bytes` hit at ~100KB |
| 1 doc × 200KB | 1 doc | Floor of 1 doc (exceeds byte limit) |
| 1000 docs × 100 bytes each (100KB total) | 100 docs | `batch_size` hit first |

No. Existing behavior is preserved by default. Users opt in to size-based batching by explicitly configuring `batch_size_bytes`.

**Why build it:**
- The current static `batch_size` is a known pain point for users with heterogeneous document sizes. It forces a compromise: set it low enough for the largest documents (wasting throughput on small documents) or set it high for small documents (risking failures on large documents).
- Remote ML model providers increasingly enforce token/byte-based rate limits. Size-aware batching aligns the ingest processor with these real-world constraints, reducing throttling errors and improving ingestion reliability.
- This is a low-risk, high-value improvement. It's additive, backward-compatible, and addresses a fundamental limitation in the current batching model.
- The `SemanticFieldProcessorFactory` already has a [TODO to make batch size configurable](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/processor/factory/SemanticFieldProcessorFactory.java#L43). Size-based batching would be a natural extension of that work.

**Reasons not to:**
- Adds configuration complexity. Users now have two knobs to tune instead of one. Mitigation: good defaults and documentation. The hybrid mode with both limits is straightforward to reason about.
- Byte measurement has a small runtime cost (measuring string byte lengths). This is negligible compared to the inference call itself.
- The batching split currently happens in OpenSearch core's `AbstractBatchingProcessor`. Implementing size-based splitting may require either a core change or a secondary split within the plugin layer. The plugin-layer approach is more practical for an initial implementation but may result in sub-batches within the core's document-count batches.

**Implementation approach:**

The most practical approach is to implement size-based splitting within the neural search plugin layer, inside `InferenceProcessor.doSubBatchExecute()`, rather than modifying OpenSearch core's `AbstractBatchingProcessor`. This avoids a core dependency and keeps the change scoped to the neural search plugin.

1. **Configuration parsing:** Update processor factories (`TextEmbeddingProcessorFactory`, `SparseEncodingProcessorFactory`, `SemanticFieldProcessorFactory`) to read `batch_size_bytes` from the processor config and pass it to the processor constructors.

2. **Byte-aware batch splitting:** In `InferenceProcessor`, after `constructInferenceTexts()` builds the flat inference list, split it into sub-batches based on cumulative byte size (using `String.getBytes(StandardCharsets.UTF_8).length` or a more efficient estimation). Each sub-batch is sent to `doBatchExecute()` independently.

3. **Result stitching:** Collect results from all sub-batches and concatenate them before passing to `batchExecuteHandler()` for mapping back to documents.

4. **Stats integration:** Add batch size metrics (doc count and byte size per batch) to the neural search stats framework.

**Assumptions and constraints:**
- Byte measurement is based on the UTF-8 encoded size of the inference text strings, not the full document payload.
- A single document that exceeds `batch_size_bytes` is always sent as its own batch (floor of 1).
- The existing `batch_size` from `AbstractBatchingProcessor` in OpenSearch core continues to control the initial document-level batching. The size-based splitting is a secondary split within each core-level batch. A future OpenSearch core enhancement could unify both into a single batching layer.
- No dependency on other feature work, though this complements the ongoing semantic field batch size configurability work ([neural-search#1438](https://github.com/opensearch-project/neural-search/pull/1438)).

**Estimated scope:** Medium. The core logic change is contained within `InferenceProcessor` and the processor factories. Testing requires benchmarking with varied document sizes to validate the adaptive behavior.

1. **Should `batch_size_bytes` be measured on raw text or estimated tokens?** Byte size is simpler and model-agnostic, but token count would be more accurate for token-based rate limits. An initial implementation using bytes is recommended, with token estimation as a future enhancement.

2. **Should there be a cluster-level default for `batch_size_bytes`?** Similar to `index.neural_search.semantic_ingest_batch_size`, a cluster or index-level setting could provide a default that individual pipelines can override.

3. **Should the core `AbstractBatchingProcessor` be enhanced to support size-based batching natively?** This would be cleaner architecturally but requires an OpenSearch core change. The plugin-layer approach is recommended for the initial implementation, with a potential core RFC to follow.

4. **What is the right default value for `batch_size_bytes`?** This depends on typical ML Commons deployment configurations and remote model limits. Benchmarking across common model providers (SageMaker, Bedrock, OpenAI-compatible endpoints) would inform a sensible default. Candidates include 100KB, 256KB, or 512KB.

5. **Should the feature support per-field byte budgets for multi-field processors?** In the current design, the byte budget applies to the entire inference list across all fields in the `field_map`. Per-field budgets would add complexity but could be useful for processors with many mapped fields of varying sizes.

arjunkumargiri:
### Description
Add size-based (byte-aware) batching for neural search ingest processors

https://github.com/opensearch-project/neural-search/issues/1761

### Related Issues
Resolves [#[Issue number to be closed when this PR is merged]](https://github.com/opensearch-project/neural-search/issues/1761)

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
@@ -44,6 +44,8 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
@@ -66,6 +68,26 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
     public static final String FIELD_MAP_FIELD = "field_map";
     public static final String SKIP_EXISTING = "skip_existing";
     public static final boolean DEFAULT_SKIP_EXISTING = false;
+    public static final String BATCH_SIZE_BYTES_FIELD = "batch_size_bytes";
+    public static final int DEFAULT_BATCH_SIZE_BYTES = -1;
+    /**
+     * Maximum allowed value for batch_size_bytes. Set to 100MB to match OpenSearch's default
+     * http.max_content_length, beyond which the request would be rejected by the HTTP layer anyway.
+     */
+    public static final int MAX_BATCH_SIZE_BYTES = 100 * 1024 * 1024; // 100MB
+    /**
+     * When batch_size_bytes is configured but batch_size is not explicitly set by the user,
+     * we auto-elevate batch_size to this value so that the byte-based limit becomes the primary
+     * effective batching constraint. 1000 is the maximum batch size supported by OpenSearch's
+     * ingest pipeline batching framework (AbstractBatchingProcessor), so this effectively makes
+     * batch_size_bytes the sole constraint for typical workloads.
+     *
+     * <p>For models that enforce both a payload size limit AND a per-request document-count limit
+     * (e.g. Cohere Embed supports at most 96 texts per request), users should set batch_size
+     * explicitly alongside batch_size_bytes. The processor will then respect whichever limit is
+     * hit first.
+     */
+    public static final int AUTO_BATCH_SIZE_FOR_BYTE_BATCHING = 1000;
     private static final BiFunction<Object, Object, Object> REMAPPING_FUNCTION = (v1, v2) -> {
         if (v1 instanceof Collection && v2 instanceof Collection) {
             ((Collection) v1).addAll((Collection) v2);
@@ -92,11 +114,46 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
 
     private final Environment environment;
     private final ClusterService clusterService;
+    private final int batchSizeBytes;
+
+    /** Returns the effective batch_size configured for this processor. Exposed for testing. */
+    @VisibleForTesting
+    public int getBatchSize() {
+        return batchSize;
+    }
 
     public InferenceProcessor(
         String tag,
         String description,
         int batchSize,
+        String type,
+        String listTypeNestedMapKey,
+        String modelId,
+        Map<String, Object> fieldMap,
+        MLCommonsClientAccessor clientAccessor,
+        Environment environment,
+        ClusterService clusterService
+    ) {
+        this(
+            tag,
+            description,
+            batchSize,
+            DEFAULT_BATCH_SIZE_BYTES,
+            type,
+            listTypeNestedMapKey,
+            modelId,
+            fieldMap,
+            clientAccessor,
+            environment,
+            clusterService
+        );
+    }
+
+    public InferenceProcessor(
+        String tag,
+        String description,
+        int batchSize,
+        int batchSizeBytes,
         String type,
         String listTypeNestedMapKey,
         String modelId,
@@ -115,6 +172,84 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
         this.mlCommonsClientAccessor = clientAccessor;
         this.environment = environment;
         this.clusterService = clusterService;
+        this.batchSizeBytes = batchSizeBytes;
+    }
+
+    /**
+     * Validates the batch_size_bytes configuration value and resolves the effective batch_size
+     * to use when constructing a processor. This shared helper is intended to be called from
+     * processor factory implementations to avoid duplicating validation logic.
+     *
+     * <p>Validation rules:
+     * <ul>
+     *   <li>If batch_size_bytes is the sentinel value ({@link #DEFAULT_BATCH_SIZE_BYTES}), it is
+     *       disabled and no validation is performed.</li>
+     *   <li>If batch_size_bytes is set, it must be a positive integer.</li>
+     *   <li>If batch_size_bytes is set, it must not exceed {@link #MAX_BATCH_SIZE_BYTES} (100MB).</li>
+     * </ul>
+     *
+     * <p><b>Batch size resolution:</b>
+     * <ul>
+     *   <li>If the user did not explicitly set batch_size ({@code userSetBatchSize} is false),
+     *       batch_size is auto-elevated to {@link #AUTO_BATCH_SIZE_FOR_BYTE_BATCHING}
+     *       ({@code Integer.MAX_VALUE}), making the byte limit the sole effective constraint.
+     *       This is the correct behavior for models with only a payload size limit (e.g. Amazon
+     *       Bedrock, Claude, most OpenAI-compatible endpoints).</li>
+     *   <li>If the user explicitly set batch_size, it is preserved as-is. The processor will
+     *       then dispatch a batch when either the document-count limit or the byte limit is
+     *       reached first. This is the correct configuration for models that enforce both limits
+     *       (e.g. Cohere Embed: max 96 texts per request and a payload size limit).</li>
+     * </ul>
+     *
+     * <p><b>Why {@code userSetBatchSize} is needed:</b>
+     * {@code AbstractBatchingProcessor.Factory.create()} removes {@code batch_size} from the
+     * config map via {@code ConfigurationUtils.readIntProperty} before calling
+     * {@code newProcessor()}, so it is impossible to detect user intent inside
+     * {@code newProcessor()} directly. Factories must override {@code create()} to peek at the
+     * config map before calling {@code super.create()} and pass the result here.
+     *
+     * @param processorType    the processor type string, used in error messages
+     * @param tag              the processor tag, used in error messages
+     * @param batchSizeBytes   the raw batch_size_bytes value read from config
+     * @param batchSize        the batch_size value resolved by AbstractBatchingProcessor
+     * @param userSetBatchSize true if the user explicitly included batch_size in the pipeline config
+     * @return the effective batch_size to pass to the processor constructor
+     * @throws IllegalArgumentException if batch_size_bytes is invalid
+     */
+    public static int validateBatchSizeBytesAndResolveEffectiveBatchSize(
+        String processorType,
+        String tag,
+        int batchSizeBytes,
+        int batchSize,
+        boolean userSetBatchSize
+    ) {
+        if (batchSizeBytes == DEFAULT_BATCH_SIZE_BYTES) {
+            return batchSize;
+        }
+        if (batchSizeBytes <= 0) {
+            throw new IllegalArgumentException(
+                "[" + processorType + "] [" + tag + "] [" + BATCH_SIZE_BYTES_FIELD + "] must be a positive integer when specified"
+            );
+        }
+        if (batchSizeBytes > MAX_BATCH_SIZE_BYTES) {
+            throw new IllegalArgumentException(
+                "["
+                    + processorType
+                    + "] ["
+                    + tag
+                    + "] ["
+                    + BATCH_SIZE_BYTES_FIELD
+                    + "] must not exceed "
+                    + MAX_BATCH_SIZE_BYTES
+                    + " bytes (100MB)"
+            );
+        }
+        // batch_size_bytes is enabled. If the user did not explicitly set batch_size, auto-elevate
+        // it so the byte limit is the sole effective constraint.
+        if (!userSetBatchSize) {
+            return AUTO_BATCH_SIZE_FOR_BYTE_BATCHING;
+        }
+        return batchSize;
     }
 
     private void validateEmbeddingConfiguration(Map<String, Object> fieldMap) {
@@ -208,9 +343,23 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
 
     /**
      * This is a helper function for subBatchExecute, which invokes doBatchExecute for given inference list.
+     * When batch_size_bytes is configured, the inference list is further split into sub-batches based on
+     * cumulative byte size of the texts. Each sub-batch is sent to doBatchExecute independently and results
+     * are stitched back together before being passed to batchExecuteHandler.
+     *
+     * <p><b>Failure semantics:</b> When batch_size_bytes is enabled, multiple sub-batches are dispatched
+     * in parallel. If any sub-batch fails, the entire set of ingestDocumentWrappers is failed via
+     * {@code updateWithExceptions}. This is consistent with the existing behavior when batch_size_bytes
+     * is disabled (a single doBatchExecute failure fails all documents in the batch). However, users
+     * should be aware that with auto-elevated batch_size (when only batch_size_bytes is set and
+     * batch_size is not explicitly configured), a larger number of documents may be grouped into a
+     * single internal batch, so a single inference failure will affect more documents than it would
+     * with a small explicit batch_size.
+     *
      * @param ingestDocumentWrappers a list of IngestDocuments in a batch.
      * @param inferenceList a list of String for inference.
      * @param dataForInferences a list of data for inference, which includes ingestDocumentWrapper, processMap, inferenceList.
+     * @param handler a callback handler to handle inference results.
      */
     protected void doSubBatchExecute(
         List<IngestDocumentWrapper> ingestDocumentWrappers,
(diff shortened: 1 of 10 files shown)


src/main/java/org/opensearch/neuralsearch/processor/factory/TextEmbeddingProcessorFactory.java
```diff
@@ -54,10 +58,17 @@ protected AbstractBatchingProcessor newProcessor(String tag, String description,
         String modelId = readStringProperty(TYPE, tag, config, MODEL_ID_FIELD);
         Map<String, Object> fieldMap = readMap(TYPE, tag, config, FIELD_MAP_FIELD);
         boolean skipExisting = readBooleanProperty(TYPE, tag, config, SKIP_EXISTING, DEFAULT_SKIP_EXISTING);
+        int batchSizeBytes = ConfigurationUtils.readIntProperty(TYPE, tag, config, BATCH_SIZE_BYTES_FIELD, DEFAULT_BATCH_SIZE_BYTES);
```
bzhangam, line 88: I think we should validate the value should be greater than 0.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -66,6 +70,18 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
...
+    /**
+     * When batch_size_bytes is configured but batch_size is not explicitly set by the user,
+     * we auto-elevate batch_size to this value so that the byte-based limit becomes the
+     * effective batching constraint. Without this, the core default batch_size of 1 would
+     * cause documents to be sent one at a time, making batch_size_bytes a no-op.
+     * OpenSearch bulk requests are bounded by http.max_content_length (default 100MB),
+     * not by document count, so 1000 is a safe upper bound.
+     */
+    public static final int AUTO_BATCH_SIZE_FOR_BYTE_BATCHING = 1000;
+    public static final int CORE_DEFAULT_BATCH_SIZE = 1;
```
bzhangam: I think we already has DEFAULT_BATCH_SIZE why not use it directly?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -221,10 +271,85 @@ protected void doSubBatchExecute(
...
+    @VisibleForTesting
+    static List<List<String>> splitInferenceListByBytes(List<String> inferenceList, int maxBytes) {
+        List<List<String>> subBatches = new ArrayList<>();
+        List<String> currentBatch = new ArrayList<>();
+        long currentBatchBytes = 0;
+
+        for (String text : inferenceList) {
+            long textBytes = text.getBytes(StandardCharsets.UTF_8).length;
+
+            if (!currentBatch.isEmpty() && currentBatchBytes + textBytes > maxBytes) {
```
bzhangam, line 447: What if a single text is longer than the maxBytes? I think we probably should throw an error to let customers either increase the maxBytes or chunk the text?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -221,10 +271,85 @@ protected void doSubBatchExecute(
...
+        if (subBatches.size() == 1) {
+            // Only one sub-batch, no need for stitching logic
+            doBatchExecute(subBatches.get(0), results -> {
+                batchExecuteHandler(results, dataForInferences, originalOrder);
+                handler.accept(ingestDocumentWrappers);
+            }, exception -> { updateWithExceptions(ingestDocumentWrappers, handler, exception); });
+            return;
+        }
+
+        // Multiple sub-batches: execute sequentially and stitch results
```
bzhangam: I think we should invoke ML Commons in parallel rather than sequentially. While bytesSize gives us better control over batching, it shouldn’t negatively impact ingestion throughput.

For example, if we have 100 documents, we currently process them as 10 batches of 10 documents and send 10 ML Commons requests in parallel. However, if we switch to batching based on bytesSize and still end up with 10 batches, but invoke ML Commons sequentially, ingestion performance will degrade significantly.

So even if we adopt bytesSize for smarter batching, we should preserve parallel execution to avoid slowing down ingestion.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
arjunkumargiri: I'd push back on throwing an error here. This runs during ingestion — throwing would fail the entire bulk request because one document happened to have a large text field. That's a poor user experience, especially since the text is still perfectly valid for inference.

The current behavior (floor of 1 item per batch) is intentional: the oversized text gets its own batch and is processed normally, just not grouped with others.

src/main/java/org/opensearch/neuralsearch/processor/factory/TextEmbeddingProcessorFactory.java
arjunkumargiri: Good catch — added validation in both `TextEmbeddingProcessorFactory` and `SparseEncodingProcessorFactory`. If the user explicitly sets `batch_size_bytes` to 0 or a negative value (other than the default sentinel `-1`), we now throw `IllegalArgumentException` at pipeline creation time

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
arjunkumargiri: The reason `CORE_DEFAULT_BATCH_SIZE` exists is that `AbstractBatchingProcessor` doesn't expose its default as a public constant we can reference — the `newProcessor` method receives the already-resolved `batchSize` int, so there's no way to distinguish \"user explicitly set batch_size=1\" from \"it defaulted to 1\". We need this sentinel value to detect the unset case. I could rename it to something clearer like `INFERRED_DEFAULT_BATCH_SIZE` if that helps, but the concept itself is necessary for the auto-elevation logic to work correctly.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
arjunkumargiri: Great point — changed to parallel execution.

arjunkumargiri:
@bzhangam agreed, this PR does not focus on `semantic` field. Will update https://github.com/opensearch-project/neural-search/issues/1761 with details that semantic field support needs to be implemented.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
bzhangam: Would it be better to modify `AbstractBatchingProcessor ` to expose the default batch size? My concern is if we change the default batch size in `AbstractBatchingProcessor ` but we still think the default size is 1 then we will behave incorrectly. Besides we also can expose a function to help us distinguish if user explicitly set the batch size or we are using the default size to make things more clear.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
bzhangam: We should document this behavior clearly so users can properly chunk their data when they encounter this situation.

The main concern is that we are not strictly respecting the maxBytes value they configure. If their model cannot process the input and throws an error, users may perceive this as a bug — since they would reasonably expect the input size to never exceed the maxBytes limit they specified.

Making this limitation explicit in the documentation will help set the right expectations and avoid confusion.

proposals/size-based-batching-proposal.md
```diff
@@ -0,0 +1,151 @@
+## What/Why
```
bzhangam: Curious what's the intention to add the proposal here? Make it easier for AI tool to understand the codebase in future?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -221,10 +270,89 @@ protected void doSubBatchExecute(
...
+                    // All sub-batches done — stitch results in original order
+                    List<Object> allResults = new ArrayList<>();
+                    for (List<?> batchResult : orderedResults) {
+                        allResults.addAll(batchResult);
+                    }
+                    batchExecuteHandler(allResults, dataForInferences, originalOrder);
+                    handler.accept(ingestDocumentWrappers);
+                }
+            }, exception -> {
+                if (firstException.compareAndSet(null, exception)) {
```
bzhangam, line 415: We should also explicitly call out that we now fail the entire batch if any inference error occurs.

Previously, we also failed the entire batch when an error happened, but the behavior was easier for users to understand because the effective batch size was clear — either 1 (default) or the value they explicitly configured.

For example, with 10 documents and the default batch size of 1. If one document failed, the other 9 could still succeed. Failures were isolated to individual documents.

However, with the new behavior, even though the default configured batch size remains 1, we internally process up to 1000 documents in a single batch. If one document fails inference, the entire internal batch fails. As a result, all documents in that batch will fail, which is a significant behavioral difference.

This change could be surprising to users and should be clearly documented to avoid confusion.

src/main/java/org/opensearch/neuralsearch/processor/factory/TextEmbeddingProcessorFactory.java
```diff
@@ -54,10 +58,20 @@ protected AbstractBatchingProcessor newProcessor(String tag, String description,
...
```
chishui, line 88: Duplicate code, what about moving it out to a shared location?

proposals/size-based-batching-proposal.md
```diff
@@ -0,0 +1,151 @@
...
+      }
+    }
+  ]
+}
+```
+
+| Parameter | Type | Default | Description |
+|---|---|---|---|
+| `batch_size` | integer | Existing default (from `AbstractBatchingProcessor`) | Maximum number of documents per batch. Existing behavior, unchanged. |
+| `batch_size_bytes` | integer | `-1` (disabled) | Maximum cumulative size in bytes of inference texts per batch. When set to a positive value, enables size-based batching. |
```
chishui: Should you limit the upper bound of this parameter?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -66,6 +69,18 @@ public abstract class InferenceProcessor extends AbstractBatchingProcessor {
...
     public static final boolean DEFAULT_SKIP_EXISTING = false;
+    public static final String BATCH_SIZE_BYTES_FIELD = "batch_size_bytes";
+    public static final int DEFAULT_BATCH_SIZE_BYTES = -1;
+    /**
+     * When batch_size_bytes is configured but batch_size is not explicitly set by the user,
+     * we auto-elevate batch_size to this value so that the byte-based limit becomes the
+     * effective batching constraint. Without this, the core default batch_size of 1 would
+     * cause documents to be sent one at a time, making batch_size_bytes a no-op.
+     * OpenSearch bulk requests are bounded by http.max_content_length (default 100MB),
+     * not by document count, so 1000 is a safe upper bound.
```
chishui: Different models may support different maximum batch size, e.g. Cohere only supports 96 texts at most https://docs.cohere.com/reference/embed. 1000 as default could cause users of these models unable to run inference.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
```diff
@@ -221,10 +270,89 @@ protected void doSubBatchExecute(
...
+     * @return a list of sub-batches
+     */
+    @VisibleForTesting
+    static List<List<String>> splitInferenceListByBytes(List<String> inferenceList, int maxBytes) {
+        List<List<String>> subBatches = new ArrayList<>();
+        List<String> currentBatch = new ArrayList<>();
+        long currentBatchBytes = 0;
+
+        for (String text : inferenceList) {
+            long textBytes = text.getBytes(StandardCharsets.UTF_8).length;
```
martin-gaievski: This is sub-optimal in terms of memory use. Consider a realistic scenario: bulk request with 1000 documents, each document has a 10KB text field. SplitInferenceListByBytes() allocates 1000 × 10KB = ~10MB of throwaway byte arrays just to measure lengths.
This is called for every text string in the inference list. String.getBytes() allocates a brand new byte[] array, copies the encoded bytes into it, and then the code only reads .length before discarding the array. 

you can address this in two ways: simpler option, for most ML inference text (English, CJK, etc.), text.length() * 3 is a safe conservative upper bound (worst case for BMP characters). This over-estimates slightly but is a one-liner and zero-allocation.
Or if you need to be precise, compute UTF8 byte length without any allocation by iterating the string's characters directly. UTF-8 encoding rules are deterministic based on codepoint value:

```java
long utf8ByteLength(String text) {
    long bytes = 0;
    for (int i = 0; i < text.length(); i++) {
        char c = text.charAt(i);
        if (c <= 0x7F) {
            bytes++;
        } else if (c <= 0x7FF) {
            bytes += 2;
        } else if (Character.isHighSurrogate(c)) {
            bytes += 4; // supplementary character = 4 UTF-8 bytes
            i++;        // skip low surrogate
        } else {
            bytes += 3;
        }
    }
    return bytes;
}
```

martin-gaievski: This is sub-optimal in terms of memory use. Consider a realistic scenario: bulk request with 1000 documents, each document has a 10KB text field. SplitInferenceListByBytes() allocates 1000 × 10KB = ~10MB of throwaway byte arrays just to measure lengths.
This is called for every text string in the inference list. String.getBytes() allocates a brand new byte[] array, copies the encoded bytes into it, and then the code only reads .length before discarding the array.

you can address this in two ways: simpler option, for most ML inference text (English, CJK, etc.), text.length() * 3 is a safe conservative upper bound (worst case for BMP characters). This over-estimates slightly but is a one-liner and zero-allocation.
Or if you need to be precise, compute UTF8 byte length without any allocation by iterating the string's characters directly. UTF-8 encoding rules are deterministic based on codepoint value:

```java
long utf8ByteLength(String text) {
    long bytes = 0;
    for (int i = 0; i < text.length(); i++) {
        char c = text.charAt(i);
        if (c <= 0x7F) {
            bytes++;
        } else if (c <= 0x7FF) {
            bytes += 2;
        } else if (Character.isHighSurrogate(c)) {
            bytes += 4; // supplementary character = 4 UTF-8 bytes
            i++;        // skip low surrogate
        } else {
            bytes += 3;
        }
    }
    return bytes;
}
```

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
arjunkumargiri: Makes sense. Will remove CORE_DEFAULT_BATCH_SIZE and will check if customer has explicitly set batch_size. If batch_size is not explicitly set then will only consider byte size for processing.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
arjunkumargiri: Users can set batch_size in addition to batch_byte_size to define an upper threshold. Do you think it is a valid solution?

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
arjunkumargiri: Addressed — replaced String.getBytes(StandardCharsets.UTF_8).length with a utf8ByteLength() helper that computes UTF-8 byte length by iterating the string's characters directly, with no byte array allocation. The implementation follows the UTF-8 encoding rules by codepoint range exactly as suggested.

src/main/java/org/opensearch/neuralsearch/processor/InferenceProcessor.java
chishui: It is a solution, but I mean it might break user's experiences that they might see broken inferencing once upgrade.

heemin32:
I am wondering if we should consider another approach which does not depends on the processor batch size entirely.
For example, queue the requests and trigger the inference it in batch when either the queue is full or it has been Xms(100ms? 500ms?) since the first request queued. This way, regardless of processor batch size, it will be handled in batch with optimal size.

arjunkumargiri:
Thanks for the suggestion @heemin32! I looked into this and here's my thinking:
  
 Model limits are primarily token/size-based, not count-based. The fundamental constraint customers hit is payload size — not timing or queue depth. A queue-based approach with a time trigger (flush after Xms) still needs to answer "when is the queue full?" — and that answer is bytes/tokens, which is exactly what batch_size_bytes provides.
  
Concerns with the queue + timer approach:
  
  1. Latency penalty for small requests — A customer indexing a single document would pay 100-500ms waiting for a timer that will never fill the queue. batch_size_bytes has zero added latency.
  2. Loss of predictability — Customers can't reason about batch boundaries. When a model rejects a payload, they need to know what knob to turn. An explicit batch_size_bytes gives them that control.
  3. Scope — A shared queue across ingest requests is a significant architectural change (thread coordination, pipeline isolation concerns, backpressure). It's better suited as a separate enhancement.
  
That said, I think the queue-based optimization is a valid complementary feature for the real-time trickle-indexing use case (many concurrent single-doc requests). But for the bulk ingest use case — where customers are hitting model size limits today — explicit batch_size_bytes is the right tool.

heemin32:
> Thanks for the suggestion @heemin32! I looked into this and here's my thinking:
> 
> Model limits are primarily token/size-based, not count-based. The fundamental constraint customers hit is payload size — not timing or queue depth. A queue-based approach with a time trigger (flush after Xms) still needs to answer "when is the queue full?" — and that answer is bytes/tokens, which is exactly what batch_size_bytes provides.
> 
> Concerns with the queue + timer approach:
> 
> 1. Latency penalty for small requests — A customer indexing a single document would pay 100-500ms waiting for a timer that will never fill the queue. batch_size_bytes has zero added latency.
> 2. Loss of predictability — Customers can't reason about batch boundaries. When a model rejects a payload, they need to know what knob to turn. An explicit batch_size_bytes gives them that control.
> 3. Scope — A shared queue across ingest requests is a significant architectural change (thread coordination, pipeline isolation concerns, backpressure). It's better suited as a separate enhancement.
> 
> That said, I think the queue-based optimization is a valid complementary feature for the real-time trickle-indexing use case (many concurrent single-doc requests). But for the bulk ingest use case — where customers are hitting model size limits today — explicit batch_size_bytes is the right tool.

1. 100ms is small( let me use 100ms :) ) compared to the time taken for inference. Additionally, this penalty is unlikely to be noticeable in a production environment with a moderate volume of requests.
2. Today, they have two knobs: batch size and byte size. With the queue approach, it simplifies to a single knob—the byte size of the queue.
3. I agree that this is a larger change requiring more engineering effort to implement the queue approach. I guess we could go with current approach and evaluate queue approach later. That said, I think this queue based batch inference in ml-common side could benefit many feature out of the box; text embedding processor, text-image embedding processor, sparse embedding processor, semantic field, and more.

bzhangam:
@arjunkumargiri We will have this [PR](https://github.com/opensearch-project/ml-commons/pull/4898) in ml-common to better solve the issue. So this PR can be discarded.

bzhangam:
@arjunkumargiri The https://github.com/opensearch-project/ml-commons/pull/4898 is merged so I think we can close this PR since the issue is solved through that one?

PR #1763, opened 2026-02-19
Fix Hybrid Search collapse result missing issue

vibrantvarun:
### Description
This PR introduces fix for multiple issues in hybrid search collapse.
## Workflow of addNewEntry
When a doc is reported to collect method, it is first added in addNewEntry if the queue is not full. When no sort criteria is provided, then collapse takes sort criteria by score in descending order.
### Issue 1
Earlier, when the new entry is added it value is updated in the comparators by following code. comparators[subQueryNumber].copy(slot, doc); Comparator is a instance of RelevancyComparator of lucene. The copy method has following definition.
```
@Override
    public void copy(int slot, int doc) throws IOException {
      scores[slot] = scorer.score();
      assert !Float.isNaN(scores[slot]);
    }
```
Here the scorer is HybridSubQueryScorer. The HSQS score() method is as follows. 
```
 @Override
    public float score() throws IOException {
        // for scenarios when scorer is needed (like in aggregations) for one doc id return sum of sub-query scores
        float totalScore = 0.0f;
        for (float score : subQueryScores) {
            totalScore += score;
        }
        return totalScore;
    }
```
The above code always gives total score (combination of both subqueries). Therefore, the comparators after copy method always had a total score for comparing individual subquery results.
For example
If for a docId the subquery scores are (0.7, 0.3] then comparators will have 1.0. It will use 1.0 score in the priority queue heap which will always gives wrong order of results in subquery perspective.

#### Fix
Introduce HybridLeafFieldComparator which wraps the current subquery score to provide in copy method. So that the minHeap in priority queue maintains the correct order.

### Issue 2
Once the queue is marked full in addNewEntry method then the comparator bottom value is not set. The bottom value basically acts a baseline to compare the new value in updateExistingEntry method. 

For example lets say size = 10. The priority queue is full. The first element in minHeap is 83 which is a minimum element. Now in the updateExistingEntry a new value if lower than 83 arrives then it will be a no-op. Greater than 83 will only be added to an existing heap.

This value 83 should be added as bottom entry for the comparison to occur. In the current code, that bottom value is not set. 

#### Fix
Sets the bottom value after fieldValueLeafTrackers[subQueryNumber] = compoundScores[subQueryNumber].add(bottomEntry); where fieldValueLeafTrackers[subQueryNumber] always contains the minHeap first element.

## Workflow of updateExistingEntry

The following condition evaluates whether the existing entry should be updated or not.
```
 if (reverseMulMap.get(groupValue) * comparators[index].compareBottom(doc) > 0) {
```

In this the compareBottom internally calls HybridSubQueryScorer.score() method to get the current score which will again be a total score as we understood in Issue 1 fix.
```
 @Override
    public int compareBottom(int doc) throws IOException {
      float score = scorer.score();
      assert !Float.isNaN(score);
      return Float.compare(score, bottom);
    }
```
### Issue 3
As the scorer.score() gets the total score from all queries to compare against a individual sub query score. Therefore, the comparison disrupts the order and misses results.

#### Fix
As soon as the updateExistingMethod is called, add the new score in HybridLeafFieldComparator. That will eventually used a right score in compareBottom to check whether any value should be updated or not. 

If the compare condition evaluates to false, then again update the HybridLeafFieldComparator to last bottom Value before comparison. So that it can be used for checking another upcoming values.

## Workflow of getTopDocs
The logic of getTopDocs is for each subquery results is to get the topNGroups. For each group, get the fieldDocs as per sorting criteria. The count of fieldDocs depends of the setting docsPerGroupSubquery.

### Issue 4
The logic of getting topNGroups per subquery is flawed. Because GroupPriorityQueue<T> lessthan logic is wrong. 

#### FIx
Fix the GroupPriorityQueue less than logic to correct get topNGroups.

### Issue 5
From each group, in earlier logic the fieldDocs are just added to array list. Basically, the list doesn't contain fieldDocs as per sort criteria. Therefore, it disrupts the normalization logic later. 

For example.

Group 1
doc 1, score -> 70000

Group 2
doc 2 , score -> 73000
doc 3 , score -> 68000
doc 4 , score -> 72000

With the buggy logic it will just add the fieldsDocs in following order
doc 1, score -> 70000, group1
doc 2 , score -> 73000, group 2
doc 3 , score -> 68000, group 2
doc 4 , score -> 72000, group 2

The right order should be (if the sort criteria is _score)
doc 2 , score -> 73000, group 2
doc 4 , score -> 72000, group 2
doc 1, score -> 70000, group1
doc 3 , score -> 68000, group 2

#### Fix
Introduce TreeMap with comparator logic as per sort criteria to arrange fieldDocs as per right sort order.

### Reason 1
In the hybrid search without collapse we use HybridTopDocCollector, where we maintain the [minScores per subquery.](https://github.com/opensearch-project/neural-search/blob/3.5/src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopScoreDocCollector.java#L164) This optimization is called during docId capture of only those documents which are[ above the minScore](https://github.com/opensearch-project/neural-search/blob/3.5/src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java#L139). 

In the hybrid search with collapse we use HybridCollapsingTopDocCollector which does not have this minscore optimization. Therefore, we end up adding all relevant results and later all are sent for normalization. Therefore, it impacts the final response. Because more the results per subquery it impacts the normalization logic.

### Reason 2
In hybrid search without collapse, we collect numHits equivalent number of top Hits. 
In hybrid search with collapse, we collect numHits equivalent number of top groups. Each group may have multiple results, which will be added in getTopDocs method. Therefore again, the number of results sent for normalization will be different in both the cases.

### Related Issues

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Bug Fixes
 - [HYBRID]: Fix relevancy bugs in hybrid query collapse ([#1753](https://github.com/opensearch-project/neural-search/pull/1753))
+- [HYBRID]: Fix profiler support for hybrid query by unwrapping ProfileScorer to access HybridSubQueryScorer ([#1754](https://github.com/opensearch-project/neural-search/pull/1754))
 - [HYBRID]: Fix missing results and ranking issue in hybrid query collapse([#1763](https://github.com/opensearch-project/neural-search/pull/1763))
 
 ### Infrastructure
--- a/src/main/java/org/opensearch/neuralsearch/search/collector/HybridLeafCollector.java
+++ b/src/main/java/org/opensearch/neuralsearch/search/collector/HybridLeafCollector.java
@@ -7,11 +7,16 @@ package org.opensearch.neuralsearch.search.collector;
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.extern.log4j.Log4j2;
+import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.LeafCollector;
 import org.apache.lucene.search.Scorable;
+import org.apache.lucene.search.Scorer;
+import org.opensearch.neuralsearch.query.HybridQueryScorer;
 import org.opensearch.neuralsearch.query.HybridSubQueryScorer;
+import org.opensearch.search.profile.ProfilingWrapper;
 
 import java.io.IOException;
+import java.util.List;
 import java.util.Objects;
 
 /**
@@ -22,30 +27,104 @@ public abstract class HybridLeafCollector implements LeafCollector {
     @Getter(AccessLevel.PACKAGE)
     HybridSubQueryScorer compoundQueryScorer;
 
+    // When profiling is enabled, HybridBulkScorer is bypassed and DefaultBulkScorer is used instead.
+    // In that case we find HybridQueryScorer through the scorer hierarchy and use it to populate
+    // per-sub-query scores on each collect() call.
+    @Getter(AccessLevel.PACKAGE)
+    HybridQueryScorer hybridQueryScorer;
+
     @Override
     public void setScorer(Scorable scorer) throws IOException {
         if (scorer instanceof HybridSubQueryScorer) {
+            // Normal path: HybridBulkScorer sets HybridSubQueryScorer directly
             compoundQueryScorer = (HybridSubQueryScorer) scorer;
         } else {
-            compoundQueryScorer = getHybridQueryScorer(scorer);
+            // Try to find HybridSubQueryScorer in the hierarchy (existing behavior)
+            compoundQueryScorer = findScorerInHierarchy(scorer, HybridSubQueryScorer.class);
+
             if (Objects.isNull(compoundQueryScorer)) {
-                log.error("cannot find scorer of type HybridQueryScorer in a hierarchy of scorer {}", scorer);
+                // Profiler path: DefaultBulkScorer is used, so we look for HybridQueryScorer instead.
+                // When found, we create a HybridSubQueryScorer adapter to bridge the two paths.
+                hybridQueryScorer = findHybridQueryScorer(scorer);
+                if (Objects.nonNull(hybridQueryScorer)) {
+                    int numSubQueries = hybridQueryScorer.getSubScorers().size();
+                    compoundQueryScorer = new HybridSubQueryScorer(numSubQueries);
+                    log.debug(
+                        "profiler mode detected: created HybridSubQueryScorer adapter from HybridQueryScorer with {} sub-queries",
+                        numSubQueries
+                    );
+                } else {
+                    log.error("cannot find scorer of type HybridSubQueryScorer in a hierarchy of scorer {}", scorer);
+                }
             }
         }
     }
 
-    private HybridSubQueryScorer getHybridQueryScorer(final Scorable scorer) throws IOException {
+    /**
+     * When in profiler mode (HybridQueryScorer path), populate the HybridSubQueryScorer's score array
+     * from the HybridQueryScorer's current sub-matches. This must be called before reading scores
+     * in collect().
+     */
+    void populateScoresFromHybridQueryScorer() throws IOException {
+        if (Objects.isNull(hybridQueryScorer) || Objects.isNull(compoundQueryScorer)) {
+            return;
+        }
+        compoundQueryScorer.resetScores();
+        float[] subQueryScores = compoundQueryScorer.getSubQueryScores();
+        int currentDoc = hybridQueryScorer.docID();
+        List<Scorer> subScorers = hybridQueryScorer.getSubScorers();
+        for (int i = 0; i < subScorers.size(); i++) {
+            Scorer subScorer = subScorers.get(i);
+            if (Objects.nonNull(subScorer) && subScorer.docID() == currentDoc && currentDoc != DocIdSetIterator.NO_MORE_DOCS) {
+                subQueryScores[i] = subScorer.score();
+            }
+        }
+    }
+
+    /**
+     * Finds HybridQueryScorer in the scorer hierarchy. Uses ProfilingWrapper interface to unwrap
+     * profiling wrappers (like ProfileScorer) without reflection.
+     */
+    @SuppressWarnings("unchecked")
+    private HybridQueryScorer findHybridQueryScorer(final Scorable scorer) throws IOException {
         if (Objects.isNull(scorer)) {
             return null;
         }
-        if (scorer instanceof HybridSubQueryScorer) {
-            return (HybridSubQueryScorer) scorer;
+        if (scorer instanceof HybridQueryScorer) {
+            return (HybridQueryScorer) scorer;
+        }
+        // when profiling is enabled, ProfileScorer implements ProfilingWrapper
+        if (scorer instanceof ProfilingWrapper) {
+            Scorer wrappedScorer = ((ProfilingWrapper<Scorer>) scorer).getDelegate();
+            if (Objects.nonNull(wrappedScorer)) {
+                return findHybridQueryScorer(wrappedScorer);
+            }
+        }
+        // Also check children for other wrapper types
+        for (Scorable.ChildScorable childScorable : scorer.getChildren()) {
+            HybridQueryScorer found = findHybridQueryScorer(childScorable.child());
+            if (Objects.nonNull(found)) {
+                return found;
+            }
+        }
+        return null;
+    }
+
+    /**
+     * Traverses the scorer tree via getChildren() to find a scorer of the specified type.
+     */
+    @SuppressWarnings("unchecked")
+    private <T extends Scorable> T findScorerInHierarchy(final Scorable scorer, final Class<T> targetType) throws IOException {
+        if (Objects.isNull(scorer)) {
+            return null;
+        }
+        if (targetType.isInstance(scorer)) {
+            return (T) scorer;
         }
         for (Scorable.ChildScorable childScorable : scorer.getChildren()) {
-            HybridSubQueryScorer hybridQueryScorer = getHybridQueryScorer(childScorable.child());
-            if (Objects.nonNull(hybridQueryScorer)) {
-                log.debug("found hybrid query scorer, it's child of scorer {}", childScorable.child().getClass().getSimpleName());
-                return hybridQueryScorer;
+            T found = findScorerInHierarchy(childScorable.child(), targetType);
+            if (Objects.nonNull(found)) {
+                return found;
             }
         }
         return null;
--- a/src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopScoreDocCollector.java
+++ b/src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopScoreDocCollector.java
@@ -133,6 +133,9 @@ public class HybridTopScoreDocCollector implements HybridSearchCollector {
 
         @Override
         public void collect(int doc) throws IOException {
+            // In profiler mode, populate scores from HybridQueryScorer before reading them
+            populateScoresFromHybridQueryScorer();
+
             HybridSubQueryScorer compoundQueryScorer = getCompoundQueryScorer();
             if (Objects.isNull(compoundQueryScorer)) {
                 return;
--- a/src/test/java/org/opensearch/neuralsearch/query/HybridQueryIT.java
+++ b/src/test/java/org/opensearch/neuralsearch/query/HybridQueryIT.java
@@ -24,9 +24,15 @@ import java.util.stream.IntStream;
 
 import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.commons.lang3.RandomUtils;
+import org.apache.hc.core5.http.io.entity.EntityUtils;
 import org.apache.lucene.search.join.ScoreMode;
 import org.junit.Before;
+import org.opensearch.client.Request;
+import org.opensearch.client.Response;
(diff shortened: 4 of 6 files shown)


src/test/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollectorTests.java
```diff
@@ -88,7 +88,7 @@ public void testKeywordCollapse_whenCollectAndTopDocs_thenSuccessful() throws IO
         assertEquals(1, topDocs.size());  // One for each sub-query
 
         for (CollapseTopFieldDocs collapseTopFieldDocs : topDocs) {
-            assertEquals(1000, collapseTopFieldDocs.totalHits.value());
+            assertEquals(100, collapseTopFieldDocs.totalHits.value());
```
vibrantvarun, line 92: Earlier these values were wrong as per logic.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -78,6 +80,7 @@ public class HybridCollapsingTopDocsCollector<T> implements HybridSearchCollecto
 
         for (int i = 0; i < sortFields.length; ++i) {
             SortField sortField = sortFields[i];
+            isSortByScore = SortField.Type.SCORE.equals(sortField.getType());
```
martin-gaievski: `isSortByScore` flag is reassigned on every iteration. Only the last sort field's type determines the flag value. For a compound sort like `Sort(SortField.FIELD_SCORE, new SortField("price", INT))`, the flag would be `false` even though score is part of the sort criteria.

While collapse with hybrid query currently defaults to `Sort(FIELD_SCORE)` (single field), this is a latent bug that will surface if compound sorts are ever used.

I suggest we do following instead: 

```java
isSortByScore = false; 
for (int i = 0; i < sortFields.length; ++i) {
    SortField sortField = sortFields[i];
    if (SortField.Type.SCORE.equals(sortField.getType())) {
        isSortByScore = true;
    }
    this.reversed[i] = sortField.getReverse() ? -1 : 1;
}
```

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -354,20 +418,17 @@ public void collect(int doc) throws IOException {
...
                     if (isQueueFull(groupValue, subQueryNumber)) {
-                        updateExistingEntry(groupValue, subQueryNumber, doc);
+                        updateExistingEntry(groupValue, subQueryNumber, doc, score);
                     } else {
+                        // Retrieve the array of collected hits for the current group
+                        // Use groupValue for immediate map lookups - safe because we're not storing it
+                        int[] collectedHitsForCurrentSubQuery = collectedHitsPerSubQueryMap.get(groupValue);
+                        int slot = collectedHitsForCurrentSubQuery[subQueryNumber];
+                        collectedHitsForCurrentSubQuery[subQueryNumber]++;
+                        collectedHitsPerSubQueryMap.put(groupValue, collectedHitsForCurrentSubQuery);
```
martin-gaievski, line 437: with this change code only counts documents that were successfully added to a queue slot. Once the queue is full, no further increments happen even though new documents continue to be evaluated and may replace bottom entries. I'm not sure this is correct.

In standard OpenSearch collapse behavior, `totalHits` reflects the total number of documents that matched the query (before collapse filtering), not the number of documents in the collapse result. Users who depend on `total.value` in the search response for pagination or UI display may see unexpectedly low counts.

We either should keep the old behavior or explicitly call out the change in docs

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -183,8 +199,12 @@ public List<CollapseTopFieldDocs> topDocs() throws IOException {
...
             }
 
-            ArrayList<ScoreDoc> fieldDocs = new ArrayList<>();
-            ArrayList<T> collapseValues = new ArrayList<>();
+            // Get comparators from any group (they're all the same type)
+            final FieldComparator<?>[] comparators = topGroupsQueue.top().queue.getComparators();
+            final int[] reverseMul = topGroupsQueue.top().queue.getReverseMul();
+
+            // Use TreeMap with simple comparator
+            TreeMap<FieldDoc, T> fieldDocsMap = new TreeMap<>(createFieldDocComparator(comparators, reverseMul));
```
martin-gaievski, line 210: TreeMap` uses the comparator for both ordering and equality. When the comparator returns `0`, the `TreeMap` considers two entries as the same key and the `put()` call overwrites the existing entry. 

We should add a final tie-breaker that guarantees uniqueness, such as a sequence counter or use `System.identityHashCode`

```java
private Comparator<FieldDoc> createFieldDocComparator(final FieldComparator<?>[] comparators, final int[] reverseMul) {
    return (a, b) -> {
        if (comparators != null) {
            for (int i = 0; i < comparators.length; i++) {
                FieldComparator<Object> comparator = (FieldComparator<Object>) comparators[i];
                int cmp = comparator.compareValues(a.fields[i], b.fields[i]);
                if (cmp != 0) {
                    return reverseMul[i] * cmp;
                }
            }
        }
        if (a.score != b.score) {
            return Float.compare(b.score, a.score);
        }
        if (a.doc != b.doc) {
            return Integer.compare(a.doc, b.doc);
        }
        // final tie-breaker to prevent TreeMap key collision
        return Integer.compare(System.identityHashCode(a), System.identityHashCode(b));
    };
}
```

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -429,10 +503,18 @@ private boolean isQueueFull(T groupValue, int index) {
...
-
+                float scoreOfLastTopEntry = 0;
+
+                // If the sort criteria is score then we need to update the currentSubQueryScore to compare it will last bottom entry value.
+                // We do this to evaluate whether it should be added in the priority queue of not.
+                if (isSortByScore) {
+                    assert comparators[index] instanceof HybridLeafFieldComparator;
+                    scoreOfLastTopEntry = ((HybridLeafFieldComparator) comparators[index]).getCurrentSubQueryScore();
+                    ((HybridLeafFieldComparator) comparators[index]).setCurrentSubQueryScore(score);
+                }
```
martin-gaievski, line 523: If `comparators[index].compareBottom(doc)` throws an `IOException`, the `else` block is never reached and the comparator's `currentSubQueryScore` remains set to the new (wrong) value. All subsequent comparisons for this group/sub-query would use a stale score.

Although such exception would likely propagate all the way up and terminate collection, the practical risk is low, but we can proactively do the defensive coding.

We can add a simple try/finally:

```java
float scoreOfLastTopEntry = 0;
if (isSortByScore) {
    scoreOfLastTopEntry = ((HybridLeafFieldComparator) comparators[index]).getCurrentSubQueryScore();
    ((HybridLeafFieldComparator) comparators[index]).setCurrentSubQueryScore(score);
}

boolean accepted = false;
try {
    accepted = reverseMulMap.get(groupValue) * comparators[index].compareBottom(doc) > 0;
} finally {
    if (!accepted && isSortByScore) {
        ((HybridLeafFieldComparator) comparators[index]).setCurrentSubQueryScore(scoreOfLastTopEntry);
    }
}

if (accepted) {
    // ... update queue logic ...
}
```

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
vibrantvarun: I added like that because it is understood that _score criteria cannot be set with any other sort critera.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
vibrantvarun: This collectedHitsPerSubQueryMap should only store those count which are a valid hit. 
In case when queue is full, we update existing entry. The count should still be same.

For example. Let say 10 hits are collected and queue is full. Then 11th hit if it valid it should replace one out 10. So collected count should still be 10. Same logic applies in Sorting here in [SimpleFieldCollector](https://github.com/opensearch-project/neural-search/blob/3.5/src/main/java/org/opensearch/neuralsearch/search/collector/SimpleFieldCollector.java#L50)

The total hits are counted updateHitCount method in https://github.com/opensearch-project/neural-search/blob/3.3/src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java#L350

The updateHitCount internally increments for totalHitCount. Which sets the totalHits later in HybridCollectorResultUtil hybridSearchCollector.getTotalHits()

```
    private TotalHits getTotalHits(final List<?> topDocs) {
        final TotalHits.Relation relation = hybridSearchCollectorResultsDTO
            .getTrackTotalHitsUpTo() == SearchContext.TRACK_TOTAL_HITS_DISABLED
                ? TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO
                : TotalHits.Relation.EQUAL_TO;

        if (topDocs == null || topDocs.isEmpty()) {
            return new TotalHits(0, relation);
        }

        return new TotalHits(hybridSearchCollector.getTotalHits(), relation);
```

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
vibrantvarun: ack.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
martin-gaievski: Oh, you are correct — I was conflating two separate counters:

1. `totalHitCount` (incremented in `updateHitCount()` for every document) — this feeds the user-facing `hits.total.value` via `getTotalHits()`, and is not affected by this change.

2. `collectedHitsPerSubQueryMap` — this is the internal per-sub-query count inside each `CollapseTopFieldDocs`, and your change to only count successfully queued entries (rather than all matched docs) is consistent with what `SimpleFieldCollector` does. Queue-full replacement doesn't increase the count, which is correct since the result set size stays the same.

Change in test assertion (e.g., `1000` → `100`) now make sense — they reflect the corrected per-sub-query count, not the user-facing total.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -255,25 +288,65 @@ protected boolean lessThan(GroupEntry<T> a, GroupEntry<T> b) {
...
+                    int cmp = comparator.compareValues(valueA, valueB);
+                    if (cmp != 0) {
+                        return reverseMul[i] * cmp;
+                    }
+                }
+            }
+
+            // Tie-breaker: score then doc ID
+            if (a.score != b.score) {
+                return Float.compare(a.score, b.score);
```
martin-gaievski: order of sort was different prior to this change, now it became ascending, lower score first. 
Is this intentionally, what is the correct order - desc or asc (my understanding it's desc)?

vibrantvarun: Oh yeah I missed it. Thanks for pointing it out.

vibrantvarun: @martin-gaievski after thinking more the following suggestion of yours is not possible. 
```
return Integer.compare(System.identityHashCode(a), System.identityHashCode(b));

```
The reason being 
```
// BAD: Non-deterministic
FieldDoc doc1 = new FieldDoc(5, 1.0f, fields);
FieldDoc doc2 = new FieldDoc(5, 1.0f, fields);
// System.identityHashCode(doc1) != System.identityHashCode(doc2)
// Even though they represent the same document!
```

vibrantvarun: This treemap will be called for per subquery result. And one document cannot occur twice in one subquery result.

martin-gaievski: I see, if it's at sub-query level that's fine if we replace one instance with other, although that shouldn't happen

vibrantvarun:
> > > Overall code looks good, please create new issue to address the drift in total hits count between w/o collapse modes, and the issue for opensearch documentation to make sure current behavior is described clearly in case customer raises questions.
> > 
> > 
> > @martin-gaievski Create two new issues
> > 
> > 1. For fixing the callouts mentioned in this PR related to collapse [[FEATURE] Hybrid Search with and without collapse should have consistent behaviour #1772 (comment)](https://github.com/opensearch-project/neural-search/issues/1772#issue-3986010581)
> > 2. One regarding minscore with hybrid collapse [[FEATURE] Hybrid Search with collapse does not return right total hits when min_score is applied #1773](https://github.com/opensearch-project/neural-search/issues/1773)
> 
> Please open issue to track the update in [opensearch documentation](https://github.com/opensearch-project/documentation-website/issues), we should clearly communicate how total hits works for hybrid query with collapse

@martin-gaievski Created a issue for documentation
https://github.com/opensearch-project/documentation-website/issues/11944

PR #1775, opened 2026-02-24
Dynamically fetch cluster node count at runtime

### LINKED ISSUE #1774: [Integ Tests] Pull number of the nodes during the runtime
### What is the bug?
In `HybridQueryAggregationsIT` we do `numOfNodes = System.getProperty("cluster.number_of_nodes", "1");`. It works when we run it against the local cluster but it will not work when we run it against a remote cluster and we don't set that parameter correctly.

Run it against a remote cluster without setting system property "cluster.number_of_nodes" properly.

It's better that we retrieve the number of the nodes info during the runtime to ensure we are using the right number of the test cluster. We can rely on a simple `_cluster/health` API call to pull the actual number of the nodes of the test cluster.

_Operating system, version._

utafrali:
Replace hardcoded system property lookup with a dynamic query to the cluster health API. This ensures integration tests accurately reflect the actual number of nodes in the cluster at runtime rather than relying on a potentially stale or missing system property.

Fixes #1774


heemin32:
@utafrali Thanks for the contribution. Unfortunately, it is duplication of https://github.com/opensearch-project/neural-search/pull/1776

Can we close this PR?

PR #1776, opened 2026-02-24
Fixed integ tests to read num of nodes dynamically

martin-gaievski:
### Description
Fixed the way of getting number of real nodes for integration tests. Instead of hardcoding it via system property locally to the test runner we read them from the actual cluster.

Manually confirmed that scenario described in #1774 works

### Related Issues
https://github.com/opensearch-project/neural-search/issues/1774

### Check List

### DIFF
--- a/src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
+++ b/src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
@@ -156,21 +156,19 @@ public abstract class BaseNeuralSearchIT extends OpenSearchSecureRestTestCase {
     private static final long TIMEOUT = 10000;
     protected String numOfNodes;
 
+    @SneakyThrows
     @Before
     public void setupSettings() {
         threadPool = setUpThreadPool();
         clusterService = createClusterService(threadPool);
         final IndexNameExpressionResolver indexNameExpressionResolver = new IndexNameExpressionResolver(new ThreadContext(Settings.EMPTY));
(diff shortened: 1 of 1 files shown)


src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -161,7 +161,16 @@ public void setupSettings() {
...
         final IndexNameExpressionResolver indexNameExpressionResolver = new IndexNameExpressionResolver(new ThreadContext(Settings.EMPTY));
-        numOfNodes = System.getProperty("cluster.number_of_nodes", "1");
+        // Dynamically determine node count from the cluster to handle remote test clusters correctly.
+        // Falls back to the system property if the cluster is not yet reachable (e.g., during startup).
+        // See https://github.com/opensearch-project/neural-search/issues/1774
+        try {
+            int actualNodeCount = getClusterNodeCount();
+            numOfNodes = String.valueOf(actualNodeCount);
+        } catch (Exception e) {
+            numOfNodes = System.getProperty("cluster.number_of_nodes", "1");
```
bzhangam: Should we simply fail the test? When we cannot get the response for the cluster health API it's highly possible the cluster is broken and the test will not work.

martin-gaievski: Makes sense, this is not the production business logic, failing fast is better. Pushed new commit

Issue #1773, opened 2026-02-24
[FEATURE] Hybrid Search with collapse does not return right total hits when min_score is applied


vibrantvarun:
### Is your feature request related to a problem?
Current the hybrid search with collapse and min_Score has following logic
```
private static TotalHits getTotalHits(
        final TotalHits existingTotalHits,
        final int originalSize,
        final int filteredSize,
        final boolean useMinScore
    ) {
        TotalHits newTotalHits;
        if (useMinScore) {
            // - if min score is not null, then we update total hits to filter size, and make best effort to keep total hits relation.
            // when filteredSize < originalSize, relation can be determined to EQUAL_TO;
            // when filteredSize == originalSize, we use the relation in existingTotalHits, which mean:
            // - EQUAL_TO: no more docs, can be determined to EQUAL_TO
            // - GREATER_THAN_OR_EQUAL_TO: can't determine the exact value above min score

            if (filteredSize < originalSize) {
                newTotalHits = new TotalHits(filteredSize, TotalHits.Relation.EQUAL_TO);
            } else {
                newTotalHits = new TotalHits(filteredSize, existingTotalHits.relation());
            }
        } else {
            newTotalHits = existingTotalHits;
        }
        return newTotalHits;
    }
```

The filtered size is not the right total hits. Instead it should be derived from existingTotalHits.

Issue #1772, opened 2026-02-24
[FEATURE] Hybrid Search with and without collapse should have consistent behaviour


vibrantvarun:
### Is your feature request related to a problem?
Currently, PR https://github.com/opensearch-project/neural-search/pull/1763 fixes the result missing issue from hybrid search with collapse. But the following reasons needs to be addressed to make the hybrid search with and without collapse in sync.

### Two important callouts
Hybrid Search with collapse will not have exact same results from hybrid search without collapse.
####  Reason 1
In the hybrid search without collapse we use HybridTopDocCollector, where we maintain the [minScores per subquery.](https://github.com/opensearch-project/neural-search/blob/3.5/src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopScoreDocCollector.java#L164) This optimization is called during docId capture of only those documents which are[ above the minScore](https://github.com/opensearch-project/neural-search/blob/3.5/src/main/java/org/opensearch/neuralsearch/query/HybridBulkScorer.java#L139).

In the hybrid search with collapse we use HybridCollapsingTopDocCollector which does not have this minscore optimization. Therefore, we end up adding all relevant results and later all are sent for normalization. Therefore, it impacts the final response. Because more the results per subquery it impacts the normalization logic.

#### Reason 2
In hybrid search without collapse, we collect numHits equivalent number of top Hits.
In hybrid search with collapse, we collect numHits equivalent number of top groups. Each group may have multiple results, which will be added in getTopDocs method. Therefore again, the number of results sent for normalization will be different in both the cases.

### What solution would you like?
_A clear and concise description of what you want to happen._

### What alternatives have you considered?
_A clear and concise description of any alternative solutions or features you've considered._

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

PR #1787, opened 2026-03-07
Fix response of hybrid search with collapse same as without collapse

vibrantvarun:
### Description
Hybrid query with collapse produces different scores and total hits compared to hybrid query without collapse.

#### Root Causes

1. The previous collector used per-group FieldValueHitQueue maps — one queue per collapse group per sub-query. During topDocs(), it selected the top-N groups and emitted all their field docs. Since each group could hold multiple docs, the total field docs sent for normalization could exceed numHits. This caused score differences because the normalization processor received a different set of documents than the non-collapse path.

2. The collector lacked the minScore optimization that HybridTopScoreDocCollector already has. Without it, every document was accepted regardless of competitiveness, inflating both the per-group doc counts and totalHits relative to the non-collapse path.

3. The minScore threshold was computed from the new bottom entry's score after updateTop(), rather than from the evicted entry's score. This set the threshold too aggressively, causing HybridBulkScorer to skip documents that should have been counted — producing a totalHits discrepancy (e.g. 380 vs 382).

#### Solution
1. Replace the per-group queue maps with a flat per-sub-query FieldValueHitQueue of size numHits, mirroring the pattern used by HybridTopScoreDocCollector. Each entry carries docId, score, sort field values, and the associated collapse group value. The actual collapse deduplication (one result per group) happens downstream in CollapseDataCollector, not in this collector. This ensures the collector emits exactly numHits docs per sub-query, matching the non-collapse path's behavior for normalization.

2. Add minScore threshold feedback. When sorting by score and a doc is evicted from the queue, the evicted entry's score (captured before overwrite) is propagated to HybridSubQueryScorer.getMinScores(). This allows HybridBulkScorer to skip non-competitive documents early, matching the optimization in HybridTopScoreDocCollector.

3. Remove the docsPerGroupPerSubQuery parameter from the constructor and factory methods since it is no longer relevant with the flat queue approach.

### Related Issues
[1773](https://github.com/opensearch-project/neural-search/issues/1773)
[1772](https://github.com/opensearch-project/neural-search/issues/1772)

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - [HYBRID]: Fix missing results and ranking issue in hybrid query collapse([#1763](https://github.com/opensearch-project/neural-search/pull/1763))
 - [HYBRID]: Fix HybridQueryDocIdStream by adding intoArray overridden method from upstream ([#1780](https://github.com/opensearch-project/neural-search/pull/1780))
 - [HYBRID]: Block hybrid query nested inside compound queries like function_score, constant_score, and script_score ([#1791](https://github.com/opensearch-project/neural-search/pull/1791))
-- [HYBRID]: Fix response of hybrid search with collapse same as without collapse ([#1787](https://github.com/opensearch-project/neural-search/pull/1787))
+- [HYBRID]: Replace per-group collection with flat queue in hybrid query collapse to fix score and totalHits inconsistencies ([#1787](https://github.com/opensearch-project/neural-search/pull/1787))
 
 ### Infrastructure
 - Fix integration test health check failures in remote clusters by dynamically discovering node count and using >= syntax for wait_for_nodes ([#1776](https://github.com/opensearch-project/neural-search/pull/1776))


src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -63,6 +64,11 @@ public class HybridCollapsingTopDocsCollector<T> implements HybridSearchCollecto
     @Setter
     TotalHits.Relation totalHitsRelation = TotalHits.Relation.EQUAL_TO;
     private HitsThresholdChecker hitsThresholdChecker;
+    // One GroupPriorityQueue per sub-query, each of size numHits
+    private GroupPriorityQueue<T>[] liveTopGroupsPerSubQuery;
+    // Per sub-query: which groups are currently in the top-K
+    private Map<T, GroupEntry<T>>[] liveGroupEntriesPerSubQuery;
```
martin-gaievski: can we use safer dynamic array (like ArrayList) instead of basic array

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -620,6 +628,80 @@ private void initializeQueue(int numSubQueries) throws IOException {
...
+             * When sorting by score, the evicted group's top score is used to update minScores.
+             */
+            private void updateLiveTopGroupForSubQuery(HybridSubQueryScorer compoundQueryScorer, T groupValue, int subQueryNumber)
+                throws IOException {
+                FieldValueHitQueue<FieldValueHitQueue.Entry>[] queues = groupQueueMap.get(groupValue);
+                if (queues == null || queues[subQueryNumber].size() == 0) {
+                    return;
+                }
+
+                GroupPriorityQueue<T> liveQueue = liveTopGroupsPerSubQuery[subQueryNumber];
```
martin-gaievski: Looks like `GroupPriorityQueue` heap invariant broken by mutable underlying queues, this potentially leads to incorrect group evictions. 

`GroupPriorityQueue` extends Lucene's `PriorityQueue` (a min-heap). `lessThan()` comparator reads `queue.top()` dynamically from the underlying `FieldValueHitQueue`. However, `PriorityQueue` does not re-heapify when the underlying state of existing elements mutates**.

Example failure scenario:
1. Group A inserted into the min-heap when its best score was 0.3 → positioned at root (weakest)
2. New docs arrive for Group A, pushing its best score to 0.9
3. Group A's heap position is not updated - still at root despite being strongest
4. Next insertion: `insertWithOverflow()` compares new element against `top()` (which is Group A at root). Since Group A's *live* score is 0.9, the comparison may correctly keep A - but the heap invariant is still violated for other elements
5. Over multiple insertions/evictions, stale heap ordering can cause wrong groups to be evicted

This is the same class of bug the PR aims to fix — incorrect group selection during collapse. A heap with violated invariants can silently produce wrong results.

We can address this properly if after early return for already-tracked groups, we add `liveTopGroupsPerSubQuery[subQueryNumber].updateTop()` to restore heap invariant when the current top element has been mutated. This matches how Lucene's own collectors handle mutable priority queue elements.

```java
if (liveEntries.containsKey(groupValue)) {
    // underlying queue mutated — restore heap invariant
    liveTopGroupsPerSubQuery[subQueryNumber].updateTop();
    return;
}

File: src/test/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollectorTests.java [hunk 3]
```diff
@@ -769,6 +755,302 @@ public void testCollapse_whenReverseSortOrder_thenCorrectOrdering() throws IOExc
...
+
+        for (int i = 0; i < scores.length; i++) {
+            hybridScorer.resetScores();
+            hybridScorer.getSubQueryScores()[0] = scores[i];
+            leafCollector.collect(i);
+        }
+
+        // Should not throw NPE or any exception
+        List<CollapseTopFieldDocs> topDocs = collector.topDocs();
+        assertEquals(1, topDocs.size());
```
martin-gaievski, line 966: add few assertions:
- which groups survived (groupA should be back in top-K after high-score re-entry)
- final scores match expectations
- correct number of results

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -63,6 +64,11 @@ public class HybridCollapsingTopDocsCollector<T> implements HybridSearchCollecto
     @Setter
     TotalHits.Relation totalHitsRelation = TotalHits.Relation.EQUAL_TO;
     private HitsThresholdChecker hitsThresholdChecker;
+    // One GroupPriorityQueue per sub-query, each of size numHits
+    private GroupPriorityQueue<T>[] liveTopGroupsPerSubQuery;
+    // Per sub-query: which groups are currently in the top-K
+    private Map<T, GroupEntry<T>>[] liveGroupEntriesPerSubQuery;
+    private float[] minScoreThresholds;
```
martin-gaievski, line 82: With these additional three structures we are increasing complexity, now we have dual priority queue per sub-query.  These exist alongside the already-complex `groupQueueMap`, `collectedHitsPerSubQueryMap`, `fieldValueLeafTrackersMap`, `comparatorsMap`, `firstComparatorMap`, `reverseMulMap`, `queueFullMap` (7 maps).

Could the old `topDocs()` approach be simplified instead? The old code already created a `GroupPriorityQueue` in `topDocs()` and selected top-N groups there. Two problems were: (a) too many docs emitted and (b) missing minScore. Could these be fixed more simply by:
- (a) Adding the `count >= numHits` cap in the existing `topDocs()` without live tracking, and
- (b) Adding `minScore` propagation in `updateExistingEntry()` when a within-group entry is evicted?

This would avoid the heap invariant issue entirely (since the queue would be built fresh in `topDocs()` without mutation). The trade-off is slightly less efficiency during collection (no early group eviction), but the correctness is easier to verify.

If live tracking during collect() is truly needed for performance with large group counts, then the heap invariant fix (#1) is mandatory.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -424,6 +426,10 @@ public void collect(int doc) throws IOException {
                         continue;
                     }
 
+                    if (isSortByScore && score <= 0 && score < minScoreThresholds[subQueryNumber]) {
```
martin-gaievski: add comment explaining that real competitive filtering happens upstream in bulk scorer class

martin-gaievski (commented):
This PR shows excellent RCA and overall approach to tackle the issue

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
vibrantvarun: > Could the old topDocs() approach be simplified instead? The old code already created a GroupPriorityQueue in topDocs() and selected top-N groups there. Two problems were: (a) too many docs emitted and (b) missing minScore. Could these be fixed more simply by:
> 
> (a) Adding the count >= numHits cap in the existing topDocs() without live tracking, and
> (b) Adding minScore propagation in updateExistingEntry() when a within-group entry is evicted?
> 

@martin-gaievski 
Yes found a better logic now.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
vibrantvarun: Good catch. Yeah I agree.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
vibrantvarun: Removed all unecessary maps as per new logic.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
vibrantvarun: Any specific reason behind this proposed change?

vibrantvarun: We can change but there is no benefit. The size is fixed at numSubQueries and never changes after initialization. Arrays are the right fit for fixed-size indexed access — no resizing, no boxing, no extra object overhead.

martin-gaievski: fixed size can change in later implementations, and cause index out of bounds exception. All listed pros are available with array backed implementation of List  - ArrayList

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -193,432 +171,262 @@ public List<CollapseTopFieldDocs> topDocs() throws IOException {
...
                 if (Objects.isNull(compoundQueryScorer)) {
                     return;
                 }
 
-                // Gets the collapse group value associated with the current document
+                // Get the collapse group value for this document
                 groupSelector.advanceTo(doc);
-                // currentValue() returns a potentially reusable reference - safe for immediate use
-                // but NOT safe for storing as map keys or persisting across document iterations
                 T groupValue = groupSelector.currentValue();
```
martin-gaievski: If I recall it correctly, in one of the previous PRs for collapse in hybrid query we bumped into one pattern: `currentValue()` returns a reference that gets reused/overwritten on the next `advanceTo()` call. Code later calls `copyGroupValue(groupValue)` when storing in `groupValueBySlot`, which does `BytesRef.deepCopyOf()` for BytesRef. This is correct for the storage path.

Here between `currentValue()` and `copyGroupValue()` (called inside `addNewEntry`/`updateExistingEntry`), `groupValue` reference is passed through the loop body. If any code between these points calls `groupSelector.advanceTo()` again, the reference would be stale. Currently safe because there's no intervening `advanceTo()` call, but fragile. 

If that's correct, I suggest we call `copyValue()` for safety.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -193,432 +171,262 @@ public List<CollapseTopFieldDocs> topDocs() throws IOException {
...
 
-            /**
-             * Collects a document and processes it based on its group value, scores, and sort.
-             *
-             * @param doc The document ID
-             * @throws IOException If an I/O error occurs
-             */
             @Override
             public void collect(int doc) throws IOException {
                 HybridSubQueryScorer compoundQueryScorer = getCompoundQueryScorer();
```
martin-gaievski, line 246: In this method we should call populateScoresFromHybridQueryScorer() from parent class HybridLeafCollector, same way HybridTopScoreDocCollector.HybridTopScoreLeafCollector.collect() does it.
I checked that the same behavior of 3 scenarios, will open new issue to address it (out of scope for this PR)

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollectorFactory.java
```diff
@@ -41,28 +39,22 @@ public static Collector createCollector(HybridCollectorFactoryDTO hybridCollecto
         int numHits = hybridCollectorFactoryDTO.getNumHits();
         FieldDoc after = hybridCollectorFactoryDTO.getAfter();
         if (collapseContext != null) {
-            // Collapse is applied
-            int docsPerGroupPerSubquery = HYBRID_COLLAPSE_DOCS_PER_GROUP_PER_SUBQUERY.get(
```
martin-gaievski, line 45: this is the only place in code where HYBRID_COLLAPSE_DOCS_PER_GROUP_PER_SUBQUERY setting is consumed. If we don't need it let's delete the actual setting and update [documentation](https://docs.opensearch.org/latest/vector-search/ai-search/hybrid-search/collapse/) to avoid customer confusion

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
martin-gaievski: Opened https://github.com/opensearch-project/neural-search/issues/1793

martin-gaievski (commented):
With this PR for scenario when one collapse group dominates sub-query's scores (many high-scoring docs in the same group) final result will have mostly documents from that group. That's fine with flat queue approach, we just need to state it clearly in the PR and issue, and update documentation, mark setting as deprecated or delete it. Initial motivation for making it configurable was performance, if collapse start retrieving too many documents and became slow user can set lower number. 

Example scenario I have in mind: 3 groups, numHits=5, sort by score

Sub-query 1 scores (docs arrive in this order):
- doc1: group=A, score=0.9
- doc2: group=A, score=0.85
- doc3: group=A, score=0.8
- doc4: group=A, score=0.75
- doc5: group=A, score=0.7
- doc6: group=B, score=0.65
- doc7: group=C, score=0.6

New code (flat queue, size=5):
- After doc1-doc5: queue = [A:0.9, A:0.85, A:0.8, A:0.75, A:0.7], full
- doc6 (B:0.65): 0.65 < bottom (0.7) → rejected. Group B lost.
- doc7 (C:0.6): 0.6 < bottom (0.7) → rejected. Group C lost.
- Result: only group A in output.

Old code (per-group queues, docsPerGroupPerSubQuery=5):
- Group A queue: [0.9, 0.85, 0.8, 0.75, 0.7]
- Group B queue: [0.65]
- Group C queue: [0.6]
- topDocs() with numHits=5: GroupPriorityQueue selects top 3 groups (only 3 exist). All survive.
- Emits: A's 5 docs + B's 1 doc + C's 1 doc = 7 docs total
- After downstream collapse: A:0.9, B:0.65, C:0.6 → 3 unique results

Old code with docsPerGroupPerSubQuery=1:
- Group A queue: [0.9] (only best kept)
- Group B queue: [0.65]
- Group C queue: [0.6]
- topDocs(): all 3 groups survive
- Emits: 3 docs
- After collapse: same 3 results 

Even with docsPerGroupPerSubQuery=1 (minimum setting), old code preserved all groups.`hybrid_collapse_docs_per_group_per_subquery ` setting only affected how many docs per group survived, not whether the group survived at all.

CHANGELOG.md
```diff
@@ -16,7 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - [HYBRID]: Fix missing results and ranking issue in hybrid query collapse([#1763](https://github.com/opensearch-project/neural-search/pull/1763))
 - [HYBRID]: Fix HybridQueryDocIdStream by adding intoArray overridden method from upstream ([#1780](https://github.com/opensearch-project/neural-search/pull/1780))
 - [HYBRID]: Block hybrid query nested inside compound queries like function_score, constant_score, and script_score ([#1791](https://github.com/opensearch-project/neural-search/pull/1791))
-
+- [HYBRID]: Fix response of hybrid search with collapse same as without collapse ([#1787](https://github.com/opensearch-project/neural-search/pull/1787))
```
martin-gaievski:
```suggestion
- [HYBRID]: Replace per-group collection with flat queue in hybrid query collapse to fix score and totalHits inconsistencies ([#1787](https://github.com/opensearch-project/neural-search/pull/1787))
```

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -193,432 +171,249 @@ public List<CollapseTopFieldDocs> topDocs() throws IOException {
...
                     if (score == 0) {
                         continue;
                     }
 
-                    // If the priority queue is full, replace the lowest scoring document per the comparator.
-                    // If the priority queue is not full, add the entry to the queue.
-                    if (isQueueFull(groupValue, subQueryNumber)) {
-                        updateExistingEntry(groupValue, subQueryNumber, doc, score);
+                    // Skip non-competitive docs when sorting by score
+                    if (isSortByScore && score <= 0 && score < minScoreThresholds[subQuery]) {
```
bzhangam, line 268: Curious why we want to keep the doc with score in (0, minScoreThresholds)?

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollectorFactory.java
```diff
@@ -41,28 +43,28 @@ public static Collector createCollector(HybridCollectorFactoryDTO hybridCollecto
         int numHits = hybridCollectorFactoryDTO.getNumHits();
         FieldDoc after = hybridCollectorFactoryDTO.getAfter();
         if (collapseContext != null) {
-            // Collapse is applied
             int docsPerGroupPerSubquery = HYBRID_COLLAPSE_DOCS_PER_GROUP_PER_SUBQUERY.get(
                 searchContext.indexShard().indexSettings().getSettings()
             );
+            if (docsPerGroupPerSubquery > 0) {
```
bzhangam, line 49: This is a breaking change and we need to update the documentation accordingly.

I think this setting was introduced to handle the large result sets. Will we run into some performance issue after removing it?

src/test/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollectorTests.java
```diff
@@ -769,6 +712,321 @@ public void testCollapse_whenReverseSortOrder_thenCorrectOrdering() throws IOExc
...
+     */
+    public void testCollapse_whenMoreGroupsThanNumHits_thenOnlyTopKGroupsSurvive() throws IOException {
+        Directory directory = newDirectory();
+        IndexWriter writer = new IndexWriter(directory, newIndexWriterConfig());
+
+        // 20 unique groups, TOP_N_GROUPS=5, so 15 groups should be evicted
+        for (int i = 0; i < 20; i++) {
+            addKeywordDoc(writer, i, "text" + i, 100 + i, "group" + i);
+        }
+        writer.forceMerge(1);
```
bzhangam, line 728: Should we add the test case without force merge to ensure we can handle multiple segments correctly?

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollectorFactory.java
vibrantvarun: Yes I am working on documentation. We have made this setting depreceated because earlier the logic was get all top groups and there field docs. This led to the documents which are irrelevant also to be part of final result. Now the logic is get all relevant docs and there associated groups.

src/test/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollectorTests.java
```diff
@@ -769,6 +712,321 @@ public void testCollapse_whenReverseSortOrder_thenCorrectOrdering() throws IOExc
         directory.close();
     }
 
+    /**
+     * Test that when more groups than numHits exist, only the top-K groups survive in topDocs.
+     */
+    public void testCollapse_whenMoreGroupsThanNumHits_thenOnlyTopKGroupsSurvive() throws IOException {
```
bzhangam, line 720: nit: We have very similar part of the tests(creating directory, writer, reader, search context, mocks) should we extract them into a helper function to reduce some duplicatioin?

vibrantvarun: I mean we can, but from a clean state perspective we should not mix two test resources.  In future, if any change has to made to specific tests then again refactoring has to be done for all tests.

src/test/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollectorTests.java
vibrantvarun: Yes I can add

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
vibrantvarun: The reason being minScoreThresholds starts with Float.MinValue. A tiny positive number when the value is not yet assigned then we don't want to accidently skip it. Same check we have in HybridTopDocsCollector as well.

bzhangam: If the minScoreThresholds = 0.5 and one doc is with the score 0.3. Should we skip the doc or keep? I feel here we want to skip the doc with the score less than minScoreThresholds. But with `score <= 0 &&` we only skip the doc with the score less than 0. Is this what we intend to do? Basically `score < minScoreThresholds` will only work when the score is less than 0.

vibrantvarun: The real min-score filtering for positive scores happens upstream in HybridBulkScorer, which checks compoundQueryScorer.getMinScores() before even calling collect(). The minScoreThresholds propagated to the scorer is what drives that early skip.

So in this collector, the threshold check is just a safety net for non-positive scores. The actual competitive filtering for positive scores is handled by:

compareBottom (in updateExistingEntry) when the queue is full
HybridBulkScorer skipping docs below minScores before they reach collect()

vibrantvarun: @bzhangam made the change in the condition.

cc: @martin-gaievski

bzhangam: Synced offline that we will address this issue in a separate PR.

Issue #1788, opened 2026-03-08
[BUG] Hybrid query inside function_score throws index_out_of_bounds_exception


martin-gaievski:
### What is the bug?
 When a hybrid query is nested inside a `function_score` query, the search fails with `index_out_of_bounds_exception: 2147483647 is out of bounds`. Value `Integer.MAX_VALUE` (`DocIdSetIterator.NO_MORE_DOCS`), indicating the scorer iterator has been exhausted but the `FunctionScoreQuery` still attempts to score at that position.
Bug is triggered specifically when the hybrid query contains a knn sub-query (supposedly any query typed based on knn, e.g. neural query). Hybrid queries with only text-based sub-queries (match, bool) inside `function_score` do not crash.

This query pattern worked in 2.19 but fails in 3.x. Problem was introduced by custom bulk scorer (PR #1289, 3.1) which changed the primary execution path from `DefaultBulkScorer` + `HybridQueryScorer` to `HybridBulkScorer`. When `function_score` wraps the hybrid query, `FunctionScoreQuery` creates its own `Weight` that uses `DefaultBulkScorer` (not `HybridBulkScorer`), causing a DISI/TPI desync between the scorer and iterator.

### How can one reproduce the bug?
1. create index with knn vector field

```json
PUT /test-func-score-bug
{
  "settings": {
    "number_of_shards": 2,
    "number_of_replicas": 0,
    "index.knn": true,
    "index.refresh_interval": "-1"
  },
  "mappings": {
    "properties": {
      "title": {"type": "text"},
      "ccu": {"type": "float"},
      "title_embedding": {
        "type": "knn_vector",
        "dimension": 3,
        "method": {"name": "hnsw", "space_type": "l2", "engine": "lucene"}
      }
    }
  }
}
```

2. create search pipeline

```json
PUT /_search/pipeline/test-pipeline
{
  "phase_results_processors": [{
    "normalization-processor": {
      "normalization": {"technique": "min_max"},
      "combination": {"technique": "arithmetic_mean"}
    }
  }]
}
```

3. ingest data, generate 10K documents with text, float, and vector fields. Ingest in two batches with a flush between them to create multiple segments.

4. run the failing query

```json
GET /test-func-score-bug/_search?search_pipeline=test-pipeline
{
  "query": {
    "function_score": {
      "functions": [{
        "filter": {"exists": {"field": "ccu"}},
        "field_value_factor": {
          "factor": 1,
          "field": "ccu",
[...]

Issue #1125, opened 2025-01-21
[Enhancement] Validate remaining Compound queries in HybridQueryPhaseSearcher


owaiskazi19:
### Is your feature request related to a problem?
Coming from [here](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java#L107). When the hybrid query is not a top level query but present as a nested or sub query in other compound queries like Boolean, Disjunction, Constant score, Function score or Boost. Such queries need to be validated and an IllegalArgumentException need to be thrown stating `hybrid query must be a top level query and cannot be wrapped into other queries`.

### What solution would you like?
Validation is already done for BooleanQuery [here](https://github.com/opensearch-project/neural-search/blob/main/src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java#L115). Other compound queries are still need to be validated.

### What alternatives have you considered?
Just validate Boolean queries.

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

owaiskazi19:
Function score query returns the below if hybrid query is nested

```json
{
  "query": {
    "function_score": {
      "query": { 
        "hybrid": {
          "play_name": "Hamlet"
        } 
      },
      "weight": "2"
    }
  }
}
```

```json
{
    "error": {
        "root_cause": [
            {
                "type": "parsing_exception",
                "reason": "Field is not supported by [hybrid] query",
                "line": 6,
                "col": 24
            }
        ],
        "type": "parsing_exception",
        "reason": "Field is not supported by [hybrid] query",
        "line": 6,
        "col": 24
    },
    "status": 400
}
```

owaiskazi19:
Constant score returns the below

```
{
  "query": {
    "constant_score": {
      "hybrid": {
        "match": {
          "text_entry": "Hamlet"
        }
      },
      "boost": 1.2
    }
  }
}

```

```
{
    "error": {
        "root_cause": [
            {
                "type": "parsing_exception",
                "reason": "[constant_score] query does not support [hybrid]",
                "line": 4,
                "col": 17
            }
        ],
        "type": "parsing_exception",
        "reason": "[constant_score] query does not support [hybrid]",
        "line": 4,
        "col": 17
    },
    "status": 400
}
```

PR #1791, opened 2026-03-10
Blocking compound and score search queries with nested hybrid queries

martin-gaievski:
### Description
When hybrid query is nested inside function_score, FunctionScoreQuery uses DefaultBulkScorer instead of HybridBulkScorer, causing DISI/TPI desync that results in index_out_of_bounds_exception (doc ID = Integer.MAX_VALUE). This is triggered specifically when knn sub-query is present inside the hybrid query. 

Hybrid query is designed to work as a top-level query only. For certain compound and score queries validation is missing, causing problems when control is given to the end user. 

With this change I'm adding a blocker for nested hybrid query when it's nested into following query types:
- function score
- script score 
- constant score

Plus I've added check for the multi-level nested hybrid queries, those are also blocked now. For example we can track now queries like this:
```
{
  "query": {
    "bool": {
      "must": [
        {
          "function_score": {
            "query": {
              "bool": {
                "should": [
                  {
                    "hybrid": {
                      "queries": [
                        { "match": { "title": "search terms" } },
                        { "term": { "category": "books" } }
                      ]
                    }
                  }
                ]
              }
            },
            "functions": [
              {
                "field_value_factor": {
                  "field": "popularity",
                  "modifier": "log1p"
                }
              }
            ]
          }
        }
      ]
    }
  }
}
```

### Related Issues
#1788 
change request for documentation https://github.com/opensearch-project/documentation-website/issues/12079
original issue regarding compound queries with nested hybrid query: https://github.com/opensearch-project/neural-search/issues/1125

### Check List
- [X] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

### DIFF
--- a/src/test/java/org/opensearch/neuralsearch/query/HybridQueryNestedBlockingIT.java
+++ b/src/test/java/org/opensearch/neuralsearch/query/HybridQueryNestedBlockingIT.java
@@ -1,0 +1,178 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.query;
+
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.containsString;
+import static org.opensearch.neuralsearch.util.TestUtils.TEST_DIMENSION;
+import static org.opensearch.neuralsearch.util.TestUtils.TEST_SPACE_TYPE;
+import static org.opensearch.neuralsearch.util.TestUtils.createRandomVector;
+
+import java.util.Collections;
+import java.util.Map;
+
+import org.junit.BeforeClass;
+import org.opensearch.client.ResponseException;
+import org.opensearch.index.query.BoolQueryBuilder;
+import org.opensearch.index.query.DisMaxQueryBuilder;
+import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.index.query.QueryBuilders;
+import org.opensearch.neuralsearch.BaseNeuralSearchIT;
+
+import com.google.common.primitives.Floats;
+import lombok.SneakyThrows;
+
+/**
+ * Integration tests verifying that hybrid queries nested inside other compound queries
+ * are correctly blocked with a clear error message. Hybrid query must always be the
+ * top-level query — nesting it inside function_score, constant_score, bool, dis_max,
+ * or any combination thereof is not supported.
+ */
+public class HybridQueryNestedBlockingIT extends BaseNeuralSearchIT {
+
+    private static final String TEST_INDEX = "test-hybrid-nested-blocking-index";
+    private static final String SEARCH_PIPELINE = "phase-results-hybrid-nested-blocking-pipeline";
+    private static final String TEST_KNN_VECTOR_FIELD = "test-knn-vector-1";
+    private static final String TEST_TEXT_FIELD = "test-text-field-1";
+    private static final String QUERY_TEXT = "hello";
+    private static final String QUERY_TEXT_2 = "world";
+    private static final String DOC_TEXT_1 = "Hello world";
+    private static final String DOC_TEXT_2 = "Hi to this place";
+    private static final String EXPECTED_ERROR = "hybrid query must be a top level query and cannot be wrapped into other queries";
+
+    private final float[] testVector1 = createRandomVector(TEST_DIMENSION);
+    private final float[] testVector2 = createRandomVector(TEST_DIMENSION);
+
+    @BeforeClass
+    @SneakyThrows
+    public static void setUpCluster() {
+        HybridQueryNestedBlockingIT instance = new HybridQueryNestedBlockingIT();
+        instance.initClient();
+        instance.updateClusterSettings();
+    }
+
+    @SneakyThrows
+    private void initializeIndexIfNotExist() {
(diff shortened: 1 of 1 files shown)


src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
```diff
@@ -150,6 +164,42 @@ private int getMaxDepthLimit(final SearchContext searchContext) {
         return MapperService.INDEX_MAPPING_DEPTH_LIMIT_SETTING.get(indexSettings).intValue();
     }
 
+    /**
+     * Check if a query contains a HybridQuery anywhere in its tree. This detects hybrid queries
+     * nested inside compound queries like function_score, script_score, constant_score, or boosting.
+     */
+    private boolean containsHybridQuery(final Query query) {
```
bzhangam: Have we handled the case hybrid query nested in another hybrid query?

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
bzhangam: Should we also have similar check for containsHybridQuery?

martin-gaievski: I think it's not needed. Unlike `validateNestedBooleanQuery` which traverses a branching tree (BooleanQuery with N clauses), `containsHybridQuery` follows a linear chain (each wrapper has exactly 1 inner
query). Depth is already bounded by index.mapping.depth.limit. Adding a limit could cause false negatives where a deeply-wrapped hybrid escapes validation.

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
martin-gaievski: It's not a concern for this PR. It's parseable but would take the `isHybridQuery` branch, never reaching `validateAndTransformQuery`. If we want to block it, that should be a parse-time check in
HybridQueryBuilder.fromXContent() as a separate issue. I think we may have addressed this in one of previous PRs

martin-gaievski: PR #1305 added validateHybridQuery() in HybridQueryUtil which checks direct sub-queries of a HybridQuery for nested HybridQuery instances. This is called from `searchWith()`:
```java 
phaseQuery = extractHybridQuery(searchContext, query);
validateHybridQuery((HybridQuery) phaseQuery);  // <- blocks hybrid-in-hybrid
```
Check is only one level deep, it checks if any direct sub-query of the outer hybrid is itself a HybridQuery. 
It doesn't recursively check deeper nesting like hybrid(bool(must: hybrid)).However, that deeper case is already covered by `validateNestedBooleanQuery` which recursively traverses BooleanQuery clauses looking for HybridQuery instances. So hybrid(bool(must: hybrid)) would be caught when the outer hybrid's sub-queries are executed, inner bool(must: hybrid) would go through validateAndTransformQuery -> BooleanQuery branch -> validateNestedBooleanQuery -> finds HybridQuery -> throws.


One real pre-existing gap: `containsHybridQuery` doesn't recurse into BooleanQuery children of wrapper queries. Function_score(bool(must: hybrid)) would not be caught, `containsHybridQuery`
sees FunctionScoreQuery -> inner is BooleanQuery -> not a recognized wrapper type -> returns false. 

This is a pre-existing gap in the validation, and could be filed as a follow-up.

martin-gaievski: I checked the code, looks like with such recursive traversal over query structure code may be not that complex, I'll run quick draft and if that confirms my finding I'll add such check to this PR, and we don't need follow up PR

martin-gaievski: Recursive checks for multiple depth level wasn't that hard, so I've added it to this PR, take a look

src/main/java/org/opensearch/neuralsearch/search/query/HybridQueryPhaseSearcher.java
bzhangam: Is it possible the inner query is a bool query and in that bool query we can have hybrid query? Will this be a case we care about?

martin-gaievski: yes, this query will be blocked, the chain of calls will be like this:
```
validateAndTransformQuery(query = FunctionScoreQuery)
  → query instanceof BooleanQuery? No
  → isCompoundQuery(query)? Yes (FunctionScoreQuery)
  → containsHybridQuery(FunctionScoreQuery, depth=20)
      → instanceof HybridQuery? No
      → instanceof FunctionScoreQuery? Yes → recurse into getSubQuery()
      → containsHybridQuery(BooleanQuery, depth=19)
          → instanceof HybridQuery? No
          → instanceof BooleanQuery? Yes → iterate clauses
              → containsHybridQuery(HybridQuery, depth=18)
                  → instanceof HybridQuery? Yes → return true
```

Issue #1795, opened 2026-03-17
[Enhancement] Fix optimization in HybridCollapsingTopDocCollector of non competitive doc


vibrantvarun:
### What is the bug?
Currently, HybridCollaspingTopDocCollector the following NonCompetitive doc logic
```
if (isSortByScore && (score <= 0 || score < minScoreThresholds[subQuery])) {
                        continue;
                    }
```
It has a dead code of score<minScoreThresholds[subquery]. Evaluate the minScoreThreshold logic in HybridTopScoreDocCollector logic and check if following condition can be changed to 
```
if (isSortByScore && (score <= 0 || score < minScoreThresholds[subQuery])) {
                        continue;
                    }
```

If the condition is changed to || then it will impact collectedHitsPerSubquery and which will lead to change in totalHits in CollapseTopFieldDocs.

PR #1796, opened 2026-03-17
Fix Competitive doc Id check with score in HybridTopScoreDocCollector and HybridCollapsingTopDocCollector

vibrantvarun:
### Description
This Pr fixes the dead competitiveDoc check in both HybridCollapsingTopScoreDocCollector and HybridCollapsingTopDocCollector by switch condition from && to ||.

Earlier all the positive scores used to participate in collectedHits but now only ones that are greater than minScoreThreshold will participate.

### Related Issues
[1795](https://github.com/opensearch-project/neural-search/issues/1795)

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
+++ b/src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
@@ -265,7 +265,7 @@ public class HybridCollapsingTopDocsCollector<T> implements HybridSearchCollecto
                     }
 
                     // Skip non-competitive docs when sorting by score
-                    if (isSortByScore && score <= 0 && score < minScoreThresholds[subQuery]) {
+                    if (isSortByScore && (score <= 0 || score < minScoreThresholds[subQuery])) {
                         continue;
                     }
 
--- a/src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopScoreDocCollector.java
+++ b/src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopScoreDocCollector.java
@@ -173,7 +173,7 @@ public class HybridTopScoreDocCollector implements HybridSearchCollector {
         }
 
         private boolean isNonCompetitiveScore(float score, int subQueryIndex) {
-            return score <= 0 && score < minScoreThresholds[subQueryIndex];
+            return score <= 0 || score < minScoreThresholds[subQueryIndex];
         }
 
         /**
--- a/src/test/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollectorTests.java
+++ b/src/test/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollectorTests.java
@@ -91,7 +91,7 @@ public class HybridCollapsingTopDocsCollectorTests extends HybridCollectorTestCa
 
         for (CollapseTopFieldDocs collapseTopFieldDocs : topDocs) {
             // With flat queue, totalHits counts all docs with score > 0 per sub-query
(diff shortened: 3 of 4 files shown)


bzhangam (approved):

martin-gaievski (changes requested):
blocking this until we have proof there is no drop in relevancy

martin-gaievski:
> Hey @martin-gaievski . Current Behavior with && Creates a Logic Gap The existing code:
> 
> ```
> if (isSortByScore && score <= 0 && score < minScoreThresholds[subQuery]) {
>     continue;
> }
> ```
> 
> This only filters when both conditions are true simultaneously. Consider this scenario:
> 
> 1. Queue fills up, and we evict a document with score = 0.5
> 2. minScoreThresholds[subQuery] is now set to 0.5
> 3. A new document arrives with score = 0.3 (positive but below threshold)
> 4. The check evaluates: 0.3 <= 0 (FALSE) && 0.3 < 0.5 (TRUE) → Overall FALSE
> 5. The document is NOT filtered early, even though it's below the competitive threshold
> 
> **Why This Matters** While there's a downstream filter that catches non-competitive documents (which is why this hasn't been reported as a bug), this early filter exists for performance optimization—to avoid unnecessary processing of documents that will be filtered out later anyway. If it doesn't work as intended, why keep confusing code that serves no purpose?
> 
> **The Fundamental Problem with &&** Since minScoreThresholds[subQuery] is dynamically updated as the queue fills (set to the evicted document's score), the && operator creates a logical flaw:
> 
> Both conditions must be met, but once minScoreThresholds rises above 0, the second condition (score < minScoreThresholds) becomes unreachable for positive scores because the first condition (score <= 0) fails This defeats the entire purpose of maintaining a dynamic threshold for early filtering
> 

Analysis of the AND condition being ineffective for positive scores is correct, we agree on that. However, the conclusion that this is a bug to fix misses the two-level architecture:

- BulkScorer (HybridBulkScorer) filters non-competitive documents before they reach collector. Document with score=0.3 when threshold=0.5 never reaches collect() in the normal path. Collector-level check is not "failing to filter", it never sees these documents.

- Collector-level check serves a different purpose: filtering score=0 (no-match) documents from the hybrid scoring system. minScoreThresholds exists at the collector level to be propagated up to the BulkScorer via compoundQueryScorer.getMinScores(), not to filter at the collector level.

- this change impacts collectedHitsPerSubQuery, which feeds into howMany in topDocsPerQuery(): `Math.min(collectedHitsPerSubQuery[i], queue.size())`. If collectedHitsPerSubQuery drops below queue.size(), valid top-K documents could be left in the queue and not returned to the normalization pipeline.

The change may be safe in practice (our testing showed identical results), but "safe" and "correct" are different things. Before approving, I'd like to see test results demonstrating that the howMany truncation
doesn't occur under any realistic scenario, particularly in profiler mode where the BulkScorer is bypassed.

> **Additional Concern: Negative Threshold Scenario** With &&: If minScoreThresholds could ever be negative, you'd keep documents with scores between the negative threshold and 0. Is that the intended behavior? This seems counterintuitive since scores ≤ 0 are generally considered non-competitive.

minScoreThresholds can never be negative. It's initialized to Float.MIN_VALUE (line 130): ≈1.4E-45, a tiny positive number, and it's only updated via Math.max: minScoreThresholds[subQueryIndex] = Math.max(minScoreThresholds[subQueryIndex], newThresholdScore). newThresholdScore comes from evicted documents: evictedScoreDoc.score. Evicted documents always have positive scores because score=0 documents are filtered by isNonCompetitiveScore before they enter the queue, they never get inserted, so they can never be evicted
Threshold starts positive (Float.MIN_VALUE) and only increases (Math.max). It's mathematically impossible for it to become negative.

heemin32:
No update for long. Closing this PR but keeping the issue.

PR #1800, opened 2026-03-19
[Agentic Search] Add support for embedding model id from agentic query translator processor

owaiskazi19:
### Description
- Add support for embedding model id from agentic query translator processor
- Added BWC tests for agentic search
- Added UTs
- Updated bwc version to 3.6.0-SNAPSHOT

```
{
    "request_processors": [
        {
            "agentic_query_translator": {
                "agent_id": "Z459B50B15L0E4dHJf8t",
                "embedding_model_id": "aoRdB50BKWay0x44HV2n"
            }
        }
    ],
```

### Related Issues
Resolves https://github.com/opensearch-project/neural-search/issues/1801

### Check List

### DIFF
--- a/.github/workflows/backwards_compatibility_tests_workflow.yml
+++ b/.github/workflows/backwards_compatibility_tests_workflow.yml
@@ -21,8 +21,8 @@ jobs:
       matrix:
         java: [ 21, 25 ]
         os: [ubuntu-latest]
-        bwc_version : [ "2.9.0","2.10.0","2.11.0","2.12.0","2.13.0","2.14.0","2.15.0","2.16.0","2.17.0","2.18.0","2.19.1","3.0.0", "3.1.0", "3.2.0", "3.3.0", "3.4.0" ]
-        opensearch_version : [ "3.5.0-SNAPSHOT" ]
+        bwc_version : [ "2.9.0","2.10.0","2.11.0","2.12.0","2.13.0","2.14.0","2.15.0","2.16.0","2.17.0","2.18.0","2.19.1","3.0.0", "3.1.0", "3.2.0", "3.3.0", "3.4.0", "3.5.0" ]
+        opensearch_version : [ "3.6.0-SNAPSHOT" ]
 
     name: NeuralSearch Restart-Upgrade BWC Tests
     runs-on: ${{ matrix.os }}
@@ -62,8 +62,8 @@ jobs:
       matrix:
         java: [21, 25]
         os: [ubuntu-latest]
-        bwc_version: [ "2.19.1","3.0.0", "3.1.0", "3.2.0", "3.3.0","3.4.0" ]
-        opensearch_version: [ "3.5.0-SNAPSHOT" ]
+        bwc_version: [ "2.19.1","3.0.0", "3.1.0", "3.2.0", "3.3.0","3.4.0", "3.5.0" ]
+        opensearch_version: [ "3.6.0-SNAPSHOT" ]
 
     name: NeuralSearch Rolling-Upgrade BWC Tests
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
```diff
@@ -700,6 +703,10 @@ private void retryableExecuteAgent(
             parameters.put("query_fields", gson.toJson(agenticQuery.getQueryFields()));
         }
 
+        if (embeddingModelId != null && !embeddingModelId.trim().isEmpty()) {
+            parameters.put("embedding_model_id", embeddingModelId);
```
bzhangam: Do we plan to support multiple embedding model ids? e.g. If one doc has multiple embedding fields and they are using different models.

vibrantvarun (approved):

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
joshpalis: This is an interesting case I haven't considered before, thanks for bringing this up. Currently agentic search supports only one embedding model ID, since we can only attach one embedding model to the `QueryPlanningTool` 

Perhaps @owaiskazi19 we can store model IDs in a dictionary and use the keys as parameters in the system prompt. An example request that this change would be able to handle would be something like this :
```
GET /movies-index/_search?search_pipeline=hybrid-search-pipeline
{
  "query": {
    "hybrid": {
      "queries": [
        {
          "neural": {
            "title_embedding": {
              "query_text": "sci-fi adventure in space",
              "model_id": "model-1",
              "k": 10
            }
          }
        },
        {
          "neural": {
            "plot_embedding": {
              "query_text": "sci-fi adventure in space",
              "model_id": "model-2",
              "k": 10
            }
          }
        }
      ]
    }
  },
  "size": 10
}
```

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
```diff
@@ -112,80 +126,109 @@ private void executeAgentAsync(
         PipelineProcessingContext requestContext,
         ActionListener<SearchRequest> requestListener
     ) {
+        // Validate embedding_model_id is only used on supported cluster versions at runtime
+        if (embeddingModelId != null
```
martin-gaievski: if embeddingModelId param is exposed to end user, shouldn't we validate value is it's not empty, similar to how it's done for agent_id

src/main/java/org/opensearch/neuralsearch/ml/MLCommonsClientAccessor.java
martin-gaievski: embedding_model_id is used as inlined value in multiple places, create new constant and use it instead

owaiskazi19: This is a good call. We can take this into future consideration to support multiple embedding models and provide more options for the users for a doc.

src/main/java/org/opensearch/neuralsearch/processor/AgenticQueryTranslatorProcessor.java
owaiskazi19: Was doing validation when storing embedding_model_id in the params but makes sense to have the same pattern as agent_id. Pushed the change

martin-gaievski: For agent_id we also doing few more checks: MAX_AGENT_ID_LENGTH and AGENT_ID_PATTERN regex check. This is a minor gap though, not a blocker

owaiskazi19: Added validation for length and pattern for embedding model id too

PR #1810, opened 2026-03-26
Fix semantic highlighter crash on documents with missing fields

### LINKED ISSUE #1809: Semantic highlighter throws on documents missing highlighted field
### Description

The semantic highlighter throws `IllegalArgumentException` when a search result document does not contain a field declared in the highlight configuration. This makes it impossible to use semantic highlighting in indices with mixed content types where documents have different fields.

### Steps to Reproduce

1. Create an index with documents of different content types (e.g., articles with `article_text`, podcasts with `podcast_description`, bible verses with `bible_verse_text`)
2. Configure semantic highlighting on multiple fields:

```json
{
  "query": { "match_all": {} },
  "highlight": {
    "fields": {
      "article_text": { "type": "semantic" },
      "podcast_description": { "type": "semantic" }
    },
    "options": { "model_id": "your-model-id" }
  }
}
```

3. The query fails with: `Field article_text is not found in the document` when a podcast document (which doesn't have `article_text`) is in the results.

### Root Cause

In `HighlightExtractorUtils.getFieldText()`, when `extractValue()` returns null (field doesn't exist on the document), an `IllegalArgumentException` is thrown:

```java
Object fieldTextObject = fieldContext.hitContext.sourceLookup().extractValue(fieldContext.fieldName, null);
if (fieldTextObject == null) {
        String.format(Locale.ROOT, "Field %s is not found in the document", fieldContext.fieldName)
    );
}
```

### Expected Behavior

The semantic highlighter should gracefully skip fields that don't exist on a document and return `null` (no highlight), similar to how the unified highlighter handles missing fields.

### Proposed Fix

In `HighlightExtractorUtils.getFieldText()`, return `null` instead of throwing when the field is missing. The calling code in `SemanticHighlighter.highlight()` already handles null returns from `getFieldText()` would need a null check before proceeding.

### Environment
- OpenSearch 3.5
- neural-search plugin (latest)

### Impact

This limitation prevents using semantic highlighting in any index with heterogeneous document types, which is a common pattern (e.g., multi-content-type search indices).

niek-mereu:
### Description

The semantic highlighter throws `IllegalArgumentException` when a document does not contain a field declared in the highlight configuration. This makes it impossible to use semantic highlighting in indices with mixed content types where documents have different fields.

### Changes

- `HighlightExtractorUtils.getFieldText()`: Return `null` instead of throwing when the field is missing, empty, or not a string. Changed from exceptions to debug-level log messages.
- `SemanticHighlighter.highlight()`: Added null check after `getFieldText()` to gracefully return null (skip highlighting) when the field doesn't exist on the document.

### Context

When using semantic highlighting across an index with heterogeneous document types (e.g. articles, podcasts, bible verses, each with different text fields), the highlighter crashes on the first document that doesn't have one of the declared highlight fields. The unified highlighter handles this gracefully by skipping missing fields; the semantic highlighter should behave the same way.

### Related Issues

Resolves #1809

### Check List
- [x] New functionality includes testing
- [ ] New functionality has been documented

For more information on following Developer Certification of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Bug Fixes
 - Fix semantic highlighter crash on documents with missing highlighted fields ([#1810](https://github.com/opensearch-project/neural-search/pull/1810))
+- [Text Chunking] Fix text chunking processor ignoring index max_token_count setting when ingesting via alias ([#1803](https://github.com/opensearch-project/neural-search/pull/1803))
 - Fix rerank processor unable to extract text from nested and dot-notation fields in document_fields ([#1805](https://github.com/opensearch-project/neural-search/pull/1805))
 - [HYBRID]: Fix relevancy bugs in hybrid query collapse ([#1753](https://github.com/opensearch-project/neural-search/pull/1753))
 - [Neural] Fix issue where remote symmetric models are not supported ([#1767](https://github.com/opensearch-project/neural-search/pull/1767))
--- a/src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/TextChunkingProcessor.java
@@ -14,11 +14,9 @@ import java.util.Objects;
 import java.util.Optional;
 
 import org.apache.commons.lang3.StringUtils;
-import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.env.Environment;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.index.analysis.AnalysisRegistry;
-import org.opensearch.index.IndexSettings;
 import org.opensearch.ingest.AbstractProcessor;
 import org.opensearch.ingest.IngestDocument;
 import org.opensearch.neuralsearch.processor.chunker.Chunker;
@@ -27,6 +25,7 @@ import org.opensearch.neuralsearch.processor.chunker.ChunkerFactory;
 import org.opensearch.neuralsearch.processor.chunker.DelimiterChunker;
 import org.opensearch.neuralsearch.processor.chunker.FixedCharLengthChunker;
 import org.opensearch.neuralsearch.processor.chunker.FixedTokenLengthChunker;
+import org.opensearch.neuralsearch.processor.util.ProcessorUtils;
 import org.opensearch.neuralsearch.stats.events.EventStatName;
 import org.opensearch.neuralsearch.stats.events.EventStatsManager;
 import org.opensearch.neuralsearch.util.ProcessorDocumentUtils;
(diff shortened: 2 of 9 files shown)


heemin32:
Could you also update the change log? https://github.com/opensearch-project/neural-search/blob/main/CHANGELOG.md

src/test/java/org/opensearch/neuralsearch/highlight/SemanticHighlighterTests.java
```diff
@@ -195,6 +195,44 @@ public void testSingleInferenceModeWithBatchInferenceFalse() throws Exception {
         verify(semanticHighlighterEngine, times(1)).getHighlightedSentences(any(), any(), any(), any(), any());
     }
 
+    public void testSingleInferenceModeReturnsNullWhenFieldMissing() throws Exception {
```
bzhangam, line 198: Can we also add a test for the batch inference mode to ensure the fix also works for it?

niek-mereu: The batch inference path is not affected by this bug. In `SemanticHighlighter.highlight()`, batch mode returns `null` at line 72 before `getFieldText()` is ever called. The actual batch processing happens in `HighlightContextBuilder.extractFieldText()` which already handles missing fields gracefully with a null check (line 42: `if (fieldText != null && !fieldText.isEmpty())`).

The existing test `testBatchInferenceModeWithSystemProcessorEnabled` already verifies this flow. Happy to add an explicit test with a missing field in batch mode if you still think it's needed.

bzhangam: I think you're correct that the batch inference has its own function to extract field text - 

https://github.com/opensearch-project/neural-search/blob/a7dafb604cd33d7592fc115b46ca708e0176f0f2/src/main/java/org/opensearch/neuralsearch/highlight/batch/config/HighlightContextBuilder.java#L73

However, I've noticed the logic differs between the two approaches:

Standard Highlighter Framework:

- Cannot support non-string fields
- Fails early when encountering lists or objects - returns null and silently skips highlighting (Need to double confirm)

Batch Inference Processor Approach:

- Can support non-string fields
- Can handle lists and objects successfully (Need to double confirm)

The User Experience Gap:

Currently, when a user tries to highlight a field that's a list:

The standard approach silently returns null and skips highlighting (no feedback to user)
The processor approach can actually highlight the list successfully

Proposed Improvement:

Rather than silently skipping, we should inform users when they attempt to highlight a list or object field using the standard approach. Should we try to throw an error(may not be easy to do this since we may fail in core before we get into the neural plugin) to let user know if they want to highlight non string field they can try the processor approach? Or we simply update the documentation to make this clear and let user figure it out by themselves?

heemin32: I think the behavioral differences between single and batch highlighting are better to be handled outside of this PR.

bzhangam: ok Then I'm good to approve this one and we can create another issue for the behavioral differences. - https://github.com/opensearch-project/neural-search/issues/1813

PR #1814, opened 2026-03-28
Fix gRPC integration test port discovery for reliable local and CI execution

martin-gaievski:
### Description
gRPC integration tests (HybridQueryGrpcIT) fail in two scenarios:

1. Local ./gradlew integTest — gRPC transport binds to the first available port in 9400-9500 ([code ref](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/transport/AuxTransport.java#L36)). When port
9400 is occupied (e.g., by VPN) or dual-stack binding shifts the IPv4 port, tests fail because GrpcTestHelper hardcodes port 9400.

2. Distribution integration tests (opensearch-build test.sh, or any run with -Dtests.rest.cluster) — the external cluster doesn't have gRPC transport configured (aux.transport.types not set), so all gRPC tests fail with IndexNotFoundException.

### Changes

build.gradle:
- Discover actual gRPC port from cluster log in doFirst block and pass it to tests via tests.grpc.port system property (only for local Gradle-managed cluster runs)
- Guard port discovery to not run when tests.rest.cluster is set (external cluster), this prevents discovering wrong cluster's port
- Bind gRPC to IPv4 loopback (grpc.bind_host=127.0.0.1) to prevent dual-stack port mismatch
- Preserve existing waitForAllConditions() call unconditionally

### Test execution matrix

| Environment | tests.rest.cluster | tests.grpc.port | gRPC tests |
|---|---|---|---|
| Local ./gradlew integTest | not set | auto-discovered | RUN |
| External cluster without gRPC | set | not set | SKIPPED |
| External cluster with gRPC | set | explicitly passed | RUN |

This fix is required to address failing grpc related integ tests 

```
./gradlew integTest --tests "org.opensearch.neuralsearch*Hybrid*Grpc*IT"

Tests with failures:
 - org.opensearch.neuralsearch.grpc.HybridQueryGrpcIT.testHybridQueryWithMixedQueryTypes
 - org.opensearch.neuralsearch.grpc.HybridQueryGrpcIT.testHybridQueryWithMaxSubQueries
 - org.opensearch.neuralsearch.grpc.HybridQueryGrpcIT.testGrpcConnectivityWithMatchAllQuery
 - org.opensearch.neuralsearch.grpc.HybridQueryGrpcIT.testHybridQueryWithLexicalKnnAndFilter
 - org.opensearch.neuralsearch.grpc.HybridQueryGrpcIT.testHybridQueryWithQueryName
 - org.opensearch.neuralsearch.grpc.HybridQueryGrpcIT.testHybridQueryWithPaginationDepth
 - org.opensearch.neuralsearch.grpc.HybridQueryGrpcIT.testBasicHybridQueryReturnsResults
 - org.opensearch.neuralsearch.grpc.HybridQueryGrpcIT.testErrorEmptyQueries
 - org.opensearch.neuralsearch.grpc.HybridQueryGrpcIT.testErrorExceedsMaxSubQueries
 - org.opensearch.neuralsearch.grpc.HybridQueryGrpcIT.testHybridQueryWithBoolFilter
 - org.opensearch.neuralsearch.grpc.HybridQueryGrpcIT.testHybridQueryWithLexicalAndKnnSubQueries
 - org.opensearch.neuralsearch.grpc.HybridQueryGrpcIT.testErrorNonDefaultBoost
 - org.opensearch.neuralsearch.grpc.HybridQueryGrpcIT.testHybridQueryWithFilterNoMatches
 - org.opensearch.neuralsearch.grpc.HybridQueryGrpcIT.testHybridQueryWithAllOptionalFields
 - org.opensearch.neuralsearch.grpc.HybridQueryGrpcIT.testHybridQueryWithFilterRestrictsResults
 - org.opensearch.neuralsearch.grpc.HybridQueryGrpcIT.testHybridQueryCombinesMultipleSubQueries

17 tests completed, 16 failed
```

I've verified locally (build min distribution with https://github.com/opensearch-project/opensearch-build) that tests on remote cluster are all green.

I've open issue in opensearch core to address this properly (port can be exposed via start script) https://github.com/opensearch-project/OpenSearch/issues/21027

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -394,18 +394,40 @@ integTest {
         // Set number of nodes system property to be used in tests
         systemProperty 'cluster.number_of_nodes', "${_numNodes}"
 
+        // There seems to be an issue when running multi node run or integ tasks with unicast_hosts
+        // not being written, the waitForAllConditions ensures it's written
+        getClusters().forEach { cluster ->
+            cluster.waitForAllConditions()
+        }
+
         // Discover the actual gRPC port from the cluster log.
         // The gRPC transport binds to the first available port in 9400-9500,
         // which may differ from 9400 due to port conflicts or dual-stack binding.
-        getClusters().forEach { cluster ->
-            cluster.waitForAllConditions()
-            def logFile = new File(cluster.nodes.first().workingDir.toFile(), "logs/integTest.log")
-            if (logFile.exists()) {
-                def matcher = (logFile.text =~ /Netty4GrpcServerTransport.*publish_address \{([^}]+):(\d+)\}/)
-                if (matcher.find()) {
-                    def grpcPort = matcher.group(2)
+        // Only discover when NOT using an external cluster (tests.rest.cluster),
+        // since the Gradle-managed cluster's gRPC port is irrelevant for external clusters.
+        // Note: For multi-node clusters, gRPC may run on any node; we parse the first node's log.
+        if (System.getProperty("tests.rest.cluster") == null) {
+            getClusters().forEach { cluster ->
+                def logFile = new File(cluster.nodes.first().workingDir.toFile(), "logs/integTest.log")
+                // Retry to handle race condition where gRPC transport hasn't written port to log yet
+                def grpcPort = null
+                for (int attempt = 0; attempt < 10 && grpcPort == null; attempt++) {
+                    if (logFile.exists()) {
+                        def matcher = (logFile.text =~ /Netty4GrpcServerTransport.*publish_address \{([^}]+):(\d+)\}/)
+                        if (matcher.find()) {
+                            grpcPort = matcher.group(2)
+                        }
+                    }
+                    if (grpcPort == null) {
+                        logger.lifecycle("Waiting for gRPC port in log file (attempt ${attempt + 1}/10)...")
+                        Thread.sleep(1000)
+                    }
+                }
+                if (grpcPort != null) {
                     systemProperty 'tests.grpc.port', grpcPort
                     logger.lifecycle("Discovered gRPC port: ${grpcPort}")
+                } else {
+                    logger.warn("Could not discover gRPC port from cluster log — gRPC tests will use default port ${9400}")
                 }
             }
         }
@@ -424,7 +446,8 @@ testClusters.integTest {
     // This is required for gRPC integration tests to work
     setting("aux.transport.types", "[transport-grpc]")
     // Bind gRPC to IPv4 loopback to avoid dual-stack port mismatch
-    setting("grpc.host", "127.0.0.1")
+    // where IPv6 takes 9400 and IPv4 gets 9401
+    setting("grpc.bind_host", "127.0.0.1")
 
     // Optionally install security
     if (System.getProperty("security.enabled") != null && System.getProperty("security.enabled") == "true") {
--- a/src/test/java/org/opensearch/neuralsearch/grpc/GrpcTestHelper.java
+++ b/src/test/java/org/opensearch/neuralsearch/grpc/GrpcTestHelper.java
@@ -44,6 +44,37 @@ public class GrpcTestHelper {
     public static final int DEFAULT_GRPC_PORT = 9400;
     public static final int DEFAULT_TIMEOUT_SECONDS = 30;
 
+    /**
+     * Check if gRPC transport is expected to be available.
+     * Returns true when either:
+     * - tests.grpc.port is explicitly set (Gradle-managed cluster discovered the port), or
+     * - tests.rest.cluster is NOT set (local testClusters run where gRPC is configured in build.gradle)
+     *
+     * Returns false when tests.rest.cluster is set but tests.grpc.port is not,
+     * indicating an external cluster that likely doesn't have gRPC transport configured.
+     *
+     * Use with {@code assumeTrue(GrpcTestHelper.isGrpcTransportConfigured())} in test setUp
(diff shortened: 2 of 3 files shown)


build.gradle
```diff
@@ -393,10 +393,21 @@ integTest {
...
         systemProperty 'cluster.number_of_nodes', "${_numNodes}"
-        // There seems to be an issue when running multi node run or integ tasks with unicast_hosts
-        // not being written, the waitForAllConditions ensures it's written
+
+        // Discover the actual gRPC port from the cluster log.
+        // The gRPC transport binds to the first available port in 9400-9500,
+        // which may differ from 9400 due to port conflicts or dual-stack binding.
         getClusters().forEach { cluster ->
             cluster.waitForAllConditions()
+            def logFile = new File(cluster.nodes.first().workingDir.toFile(), "logs/integTest.log")
```
will-hwang: how do we ensure grpc server has started and has its port assigned?

build.gradle
```diff
@@ -393,10 +393,21 @@ integTest {
...
+            if (logFile.exists()) {
+                def matcher = (logFile.text =~ /Netty4GrpcServerTransport.*publish_address \{([^}]+):(\d+)\}/)
```
will-hwang: [nit]`def matcher = (logFile.text =~ /Netty4GrpcServerTransport.*publish_address \{(127\.0\.0\.1):(\d+)\}/)
`

build.gradle
```diff
@@ -412,6 +423,8 @@ testClusters.integTest {
     // Enable gRPC transport for integration tests
     // This is required for gRPC integration tests to work
     setting("aux.transport.types", "[transport-grpc]")
+    // Bind gRPC to IPv4 loopback to avoid dual-stack port mismatch
+    setting("grpc.host", "127.0.0.1")
```
will-hwang: do we need this change?

will-hwang: when tested locally, `grpc.host` still had dual-stack binding issue

below setting worked for me instead: 
```
setting("grpc.bind_host", "127.0.0.1")
```

build.gradle
martin-gaievski: current code calls waitForAllConditions() unconditionally first (ensures cluster is fully started), then reads the log. gRPC server starts during cluster startup, so by the time waitForAllConditions()
returns, the port is bound and logged. Your comment was on the original code where waitForAllConditions() was inside the gRPC discovery block, with this PR it's separated and runs first.

build.gradle
martin-gaievski: changed setting to `grpc.bind_host`, it's needed to prevent dual-stack port mismatch.

martin-gaievski:
> if it is likely that port 9400 is occupied, should we change the default value here too?

DEFAULT_GRPC_PORT = 9400 is now only a fallback. In local runs, `tests.grpc.port` is auto-discovered and overrides the default via getGrpcPort(). In external cluster runs without gRPC, tests are skipped entirely. Default only matters if someone runs tests manually without the Gradle wrapper, keeping 9400 as the default is correct since that's the standard gRPC port.

build.gradle
martin-gaievski: IMO current regex `([^}]+):(\d+)` is more robust, it works regardless of what address gRPC publishes. If someone changes `grpc.bind_host` to something else (e.g., 0.0.0.0, a specific interface IP), or if the setting is removed in the future, the regex still works. Hardcoding 127.0.0.1 in the regex creates a hidden coupling between the setting("grpc.bind_host", "127.0.0.1") line and the regex pattern, change one without the other and port discovery silently breaks.

build.gradle
```diff
@@ -393,11 +393,31 @@ integTest {
...
+
+        // Discover the actual gRPC port from the cluster log.
+        // The gRPC transport binds to the first available port in 9400-9500,
+        // which may differ from 9400 due to port conflicts or dual-stack binding.
+        // Only discover when NOT using an external cluster (tests.rest.cluster),
+        // since the Gradle-managed cluster's gRPC port is irrelevant for external clusters.
+        if (System.getProperty("tests.rest.cluster") == null) {
+            getClusters().forEach { cluster ->
+                def logFile = new File(cluster.nodes.first().workingDir.toFile(), "logs/integTest.log")
+                if (logFile.exists()) {
```
owaiskazi19: Can add a small retry logic to avoid race condition for grpc to write port to log file

build.gradle
```diff
@@ -393,11 +393,31 @@ integTest {
...
         // There seems to be an issue when running multi node run or integ tasks with unicast_hosts
         // not being written, the waitForAllConditions ensures it's written
         getClusters().forEach { cluster ->
             cluster.waitForAllConditions()
         }
+
+        // Discover the actual gRPC port from the cluster log.
+        // The gRPC transport binds to the first available port in 9400-9500,
+        // which may differ from 9400 due to port conflicts or dual-stack binding.
+        // Only discover when NOT using an external cluster (tests.rest.cluster),
```
owaiskazi19, line 406: Let's add a note for multi node
```
Note: For multi-node clusters, gRPC may run on any node; we parse the first node's log
```

build.gradle
martin-gaievski: ack

build.gradle
martin-gaievski: ack

will-hwang (approved):
LGTM

PR #1824, opened 2026-04-01
Fix flaky integration test failure caused by ML memory circuit breaker during model deployment

### LINKED ISSUE #1823: [AUTOCUT] Integration Test Failed for neural-search-3.6.0
#### Details

| Platform | Dist | Arch | Dist Build No. | RC | Test Report | Workflow Run | Failing tests |
|----------|------|------|----------------|----|-------------|--------------|---------------|
| windows | zip | x64 | 11832 | 4 | https://ci.opensearch.org/ci/dbc/integ-test/3.6.0/11832/windows/x64/zip/test-results/11035/integ-test/test-report.yml | https://build.ci.opensearch.org/job/integ-test/11035/display/redirect | [Check metrics](null) |

Check out test report manifest linked above for steps to reproduce, cluster and integration test failure logs. For additional information checkout the [wiki](https://github.com/opensearch-project/opensearch-build/wiki/Testing-the-Distribution) and [OpenSearch Metrics Dashboard](https://metrics.opensearch.org/_dashboards/app/dashboards#/view/21aad140-49f6-11ef-bbdd-39a9b324a5aa).

Tagging the release owners to take a look @will-hwang

martin-gaievski:
### Description
Fix flaky integration test failure in distribution pipeline caused by ML memory circuit breaker tripping during model deployment.

When the circuit breaker trips (due to accumulated heap pressure from running many model load/undeploy cycles in a single JVM with limited heap), the model enters `DEPLOY_FAILED` state. Previously, `waitForModelToBeReady()` would only poll the model state without retrying the deploy, causing the test to time out after 30 polling attempts even though the circuit breaker may have already cleared (GC logs show memory recovering immediately after the failure).

This change adds retry logic: when `DEPLOY_FAILED` state is detected, the deploy operation is retried up to 3 times.

### Related Issues
Resolves #1823

This is a recurring issue pattern:
- #596 (Jun 2024) — same CB failure in distribution pipeline
- #1093 (Jan 2025) — same pattern

### Check List
- [x] New functionality includes testing
- [x] New functionality has been documented

### Changes
- `BaseNeuralSearchIT.waitForModelToBeReady()`: detect `DEPLOY_FAILED` state and retry deploy (up to 3 times)
- Added `MAX_DEPLOY_RETRIES = 3` constant
- Improved logging to include current model state

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * [Text Chunking] Fix text chunking processor ignoring index max_token_count setting when ingesting via alias ([#1803](https://github.com/opensearch-project/neural-search/pull/1803))
 
 ### Infrastructure
+* Fix flaky integration test failure caused by ML memory circuit breaker during model deployment in distribution pipeline ([#1824](https://github.com/opensearch-project/neural-search/pull/1824))
 
 ### Documentation
 
--- a/src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
+++ b/src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
@@ -153,6 +153,7 @@ public abstract class BaseNeuralSearchIT extends OpenSearchSecureRestTestCase {
     private static final Set<String> DEPLOYED_MODEL_IDS = ConcurrentHashMap.newKeySet();
     private static final int MAX_ATTEMPTS = 30;
     private static final int WAIT_TIME_IN_SECONDS = 2;
+    private static final int MAX_DEPLOY_RETRIES = 3;
     private static final long TIMEOUT = 10000;
     protected String numOfNodes;
 
@@ -365,12 +366,24 @@ public abstract class BaseNeuralSearchIT extends OpenSearchSecureRestTestCase {
     }
 
     protected void waitForModelToBeReady(String modelId) throws Exception {
+        int deployRetries = 0;
         for (int attempt = 0; attempt < MAX_ATTEMPTS; attempt++) {
-            if (isModelReadyForInference(modelId)) {
+            MLModelState state = getModelState(modelId);
+            if (MLModelState.LOADED.equals(state) || MLModelState.DEPLOYED.equals(state)) {
                 logger.info("Model {} is ready for inference after {} attempts", modelId, attempt + 1);
                 return;
             }
-            logger.info("Waiting for model {} to be ready. Attempt {}/{}", modelId, attempt + 1, MAX_ATTEMPTS);
+            if (MLModelState.DEPLOY_FAILED.equals(state) && deployRetries < MAX_DEPLOY_RETRIES) {
+                deployRetries++;
+                logger.info(
+                    "Model {} deploy failed (likely circuit breaker), retrying deploy {}/{}",
+                    modelId,
+                    deployRetries,
+                    MAX_DEPLOY_RETRIES
+                );
+                loadModel(modelId);
+            }
+            logger.info("Waiting for model {} to be ready. Attempt {}/{}, state: {}", modelId, attempt + 1, MAX_ATTEMPTS, state);
             Thread.sleep(WAIT_TIME_IN_SECONDS * 1000);
         }
         throw new RuntimeException("Model " + modelId + " failed to be ready for inference after " + MAX_ATTEMPTS + " attempts");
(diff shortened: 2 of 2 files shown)


src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -365,12 +366,24 @@ protected void loadModel(final String modelId) throws Exception {
...
-            logger.info("Waiting for model {} to be ready. Attempt {}/{}", modelId, attempt + 1, MAX_ATTEMPTS);
+            if (MLModelState.DEPLOY_FAILED.equals(state) && deployRetries < MAX_DEPLOY_RETRIES) {
+                deployRetries++;
+                logger.info(
+                    "Model {} deploy failed (likely circuit breaker), retrying deploy {}/{}",
+                    modelId,
+                    deployRetries,
+                    MAX_DEPLOY_RETRIES
+                );
+                loadModel(modelId);
```
owaiskazi19: Suggest adding wait time for circuit breaker recovery
```
Thread.sleep(500);
```

owaiskazi19: Do we want to surround loadModel with try/catch?

martin-gaievski: I'm not sure we need additional sleep time, loadModel() method already has built-in wait time, it polls the deploy task status up to 30 times with 1-second intervals before returning. After loadModel() returns, the outer loop also sleeps for 2 seconds before the next state check. So there's already 30+ seconds of recovery
time between retry attempts.

martin-gaievski: Good point. In circuit breaker scenario, loadModel() won't throw because checkComplete() treats error states as complete (returns true when error != null). Model just ends up in DEPLOY_FAILED state, and next loop iteration handles it. But a defensive try/catch is cheap and makes this more robust against unexpected failures. 

IMO it worth adding

owaiskazi19: Makes sense. `loadModel()` wait time should be enough for now then

owaiskazi19 (approved):

CHANGELOG.md
```diff
@@ -15,7 +15,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * [Text Chunking] Fix text chunking processor ignoring index max_token_count setting when ingesting via alias ([#1803](https://github.com/opensearch-project/neural-search/pull/1803))
 
 ### Infrastructure
+<<<<<<< fix/flaky-model-deploy-retry
```
bzhangam: Miss cleaning up when resolve the conflicts?

src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java
```diff
@@ -365,12 +366,28 @@ protected void loadModel(final String modelId) throws Exception {
...
+        int deployRetries = 0;
         for (int attempt = 0; attempt < MAX_ATTEMPTS; attempt++) {
-            if (isModelReadyForInference(modelId)) {
+            MLModelState state = getModelState(modelId);
+            if (MLModelState.LOADED.equals(state) || MLModelState.DEPLOYED.equals(state)) {
                 logger.info("Model {} is ready for inference after {} attempts", modelId, attempt + 1);
                 return;
             }
-            logger.info("Waiting for model {} to be ready. Attempt {}/{}", modelId, attempt + 1, MAX_ATTEMPTS);
+            if (MLModelState.DEPLOY_FAILED.equals(state) && deployRetries < MAX_DEPLOY_RETRIES) {
```
bzhangam, line 376: If the memory circuit breaker is trigger will retry help? Once some memory is released the circuit breaker will auto reset and allow us to deploy the model?

martin-gaievski: Yes, that's exactly the mechanism. CB auto-resets once GC frees enough memory, in the failure logs we can see heap dropping from 2.1GB to 699MB immediately after the CB trips. The problem is that by that poin model is already stuck in DEPLOY_FAILED state, and the old waitForModelToBeReady() just polls the state 30 times without ever re-issuing the deploy call. So even though the CB has cleared and a new deploy would succeed, nothing triggers it.

  The retry here detects DEPLOY_FAILED, calls loadModel() again (which internally polls for 30s), and the CB is already closed by then so the deploy succeeds. We've seen this same CB failure recur in #596 and #1093 — the retry bridges the gap between "CB auto-resets" and "model needs a new deploy call to recover."

bzhangam (approved):

PR #1830, opened 2026-04-08
fix: zscore returns 0.0 when sd is zero, arithmetic mean excludes zero scores

tejasae-afk:
Found two small bugs while going through the normalization and combination code.

`ZScoreNormalizationTechnique.normalizeSingleScore` returns `minScore` when standard deviation is 0, but the test helper at line 371 and the inline comment on line 286 both document the expected value as `0.0f` for that edge case.

`ArithmeticMeanScoreCombinationTechnique` uses `score >= 0.0` to include scores in the weighted mean, but the class javadoc explicitly says zero scores are excluded from N. The condition should be `score > 0.0` so a document matched by only some subqueries does not get its score diluted by the zeros from the others.

Happy to close if either of these is intentional, you know the codebase far better than I do.

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/combination/ArithmeticMeanScoreCombinationTechnique.java
@@ -46,7 +46,7 @@ public class ArithmeticMeanScoreCombinationTechnique implements ScoreCombination
         float sumOfWeights = 0;
         for (int indexOfSubQuery = 0; indexOfSubQuery < scores.length; indexOfSubQuery++) {
             float score = scores[indexOfSubQuery];
-            if (score >= 0.0) {
+            if (score > 0.0) {
                 float weight = scoreCombinationUtil.getWeightForSubQuery(weights, indexOfSubQuery);
                 score = score * weight;
                 combinedScore += score;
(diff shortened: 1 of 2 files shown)


martin-gaievski (changes requested):
This breaks score semantics for partial matches with min_max and l2. 
Score 0.0 in the scores array comes from `ScoreCombiner.getNormalizedScoresPerDocument()` which initializes `new float[topDocsPerSubQuery.size()]`, Java zero-initializes to 0.0f. This 0.0 represents "document did not match this sub-query."
With the current code (`>= 0.0`), unmatched sub-queries contribute their weight to the denominator, penalizing partial matches. With `> 0.0` in this PR, unmatched sub-queries are excluded entirely.

Regression example (2 sub-queries, equal weights 1.0):

| Document | Sub-query 0 | Sub-query 1 | Current score | PR score |
|---|---|---|---|---|
| Doc A | 0.8 (matched) | 0.0 (not matched) | (0.8+0.0)/2 = 0.4 | 0.8/1 = 0.8 |
| Doc B | 0.5 (matched) | 0.5 (matched) | (0.5+0.5)/2 = 0.5 | (0.5+0.5)/2 = 0.5 |

Current ranking: Doc B > Doc A (0.5 > 0.4). PR ranking: Doc A > Doc B (0.8 > 0.5) => ranking is reversed.

Historical context: scores array was originally initialized to `-1.0f` via `Arrays.fill(scores, -1.0f)` (sentinel for "not matched"). `>= 0.0` check in ArithmeticMean correctly excluded -1.0 while including 0.0 (valid normalized score). PR #267 removed the `Arrays.fill` but left the stale comment "we initialize with -1.0." After removal, unmatched sub-queries get 0.0 (Java default), and `>= 0.0` includes them in the denominator. This has been the behavior since OS 2.10 — changing it now is a breaking change to established scoring semantics.

Geometric and harmonic mean use `> 0` because `ln(0)` is undefined (geometric) and `1/0` is undefined (harmonic). They must skip zeros for mathematical correctness. ArithmeticMean has no such mathematical constraint, it can and should include zeros.

PR #1858, opened 2026-05-19
Fix batch semantic highlighting on inner_hits fields

### LINKED ISSUE #1840: [FEATURE] Support InnerHits for SemanticHighlighting batch processing
Yes.

This issue is initially found in 3.3:

**The setup:**

In OpenSearch Nested `chunks` is being used — documents store text in nested fields, retrieved via inner_hits
Baseline query needs are: 10 docs × 10 chunks each = up to 100 inference calls per query

**The problem:**

Semantic highlighting on `chunks.text` inside inner_hits works fine in synchronous mode however it doesn't scale as it synchronously makes an individual request at a time and with 100 hits (10 docs x 10 chunks) query response time is significantly impacted. 

Adding `batch_inference: true` causes it to silently fail — response comes back in ~6ms with no highlights and no errors, confirming the model is never called
This makes batch mode essential for us at that scale, but it's currently broken for nested architectures

**Root cause:**

- The config extraction (HighlightConfigBuilder) also only reads the top-level highlighter() from the search source, missing highlight configs specified within inner_hits of query. The processor never even sees the highlight configuration.
- The batch processor (SemanticHighlightingProcessor) operates as a SearchResponseProcessor and only processes top-level hits. It doesn’t traverse inner_hits from search hit.

**Example query that fails:**
```
JSON
{
  "query": {
    "nested": {
      "path": "chunks",
      "query": {
        "match": {
          "chunks.text": "your search term"
        }
      },
      "inner_hits": {
        "highlight": {
          "fields": {
            "chunks.text": {
              "type": "semantic"
            }
          },
          "options": {
            "model_id": "<model-id>",
            "batch_inference": true
          }
        }
      }
    }
  }
```

- Support extracting highlight config from `inner_hits` of query in SemanticHighlightingProcessor
- Support semantic highlighting in batch processor for query response in `inner_hits` case

_A clear and concise description of any alternative solutions or features you've considered._

RexyZhang2001:
## Description

Batch semantic highlighting silently produced no highlights when the `type: semantic` declaration was placed inside `nested.inner_hits.highlight` rather than at the top-level `highlight` block. Customers running a chunked-style architecture with `batch_inference: true` saw the request return ~6ms with no highlights and no errors.

This PR fixes the issue by improving the existing system-generated batch processor and adding a request-level opt-in. No new custom processor and no user-configured search pipeline are required.

## Changes

- **`SemanticHighlighterExtBuilder` (new)** — request-level ext block `"ext": { "semantic_highlighting_batch": true }`. Boolean opt-in for batch semantic highlighting. Reads via `FetchContext.getSearchExt(NAME)` so the fetch-phase highlighter can yield, and via the request's ext list so the factory can decide whether to engage.
- **`SemanticHighlightingFactory`** — uses a cheap candidate check before walking the query tree: returns false immediately when neither the ext block is set nor a top-level `type: semantic` field is declared. Only candidates have the tree-walk latency; non-semantic-highlighting searches see zero added latency.
- **`HighlightConfigResolver` (new)** — walks the query tree (`NestedQueryBuilder`, `BoolQueryBuilder`, `DisMaxQueryBuilder`, `ConstantScoreQueryBuilder`, `BoostingQueryBuilder`, `FunctionScoreQueryBuilder`, `ScriptScoreQueryBuilder`) so `type: semantic` declarations inside `inner_hits` are discovered, with their nested path and bucket name captured from the enclosing `NestedQueryBuilder` and `InnerHitBuilder`.
- **`SemanticHighlightingProcessor`** — runtime path now uses the resolver to enumerate every target (top-level + inner hits), expands per-row inference units against actual hits, runs one batch inference call (paginated when row count exceeds `max_inference_batch_size`), and writes spans into each hit's `highlightFields`.
- **`SemanticHighlighter` (fetch phase)** — yields whenever either `ext.semantic_highlighting_batch: true` or the legacy field-level `batch_inference: true` is set, so existing 3.x users on `batch_inference` are unaffected. When the system-generated factory is disabled in cluster settings, the fetch-phase logs a WARN and returns null (200 OK with no highlight, instead of throwing).
- **`HighlightContext` / `HighlightContextBuilder` / `HighlightResultApplier`** — switched to aligned parallel lists (`fieldNames`, `preTags`, `postTags`, `noMatchSizes`, `encoders`) so a single batch call can carry rows for multiple fields with their own tags / encoder / no_match_size. Inner-hit `_source` is keyed by leaf name, so the applier and builder both fall back from dotted path (`chunks.text`) to leaf (`text`).
- **`ProcessorUtils.extractQueryTextFromBuilder`** — added unwrappers for `BoolQueryBuilder`, `DisMaxQueryBuilder`, `ConstantScoreQueryBuilder`, `BoostingQueryBuilder` (positive clause only), `FunctionScoreQueryBuilder`, `ScriptScoreQueryBuilder`. Without these, requests wrapping a `match` in any compound container silently produced no highlights — same failure mode as the primary bug.
- **`SemanticHighlighter` non-batch path** — now honors the standard `encoder: html` and `no_match_size` highlight options (previously only the batch path did).
- **Removed** internal helpers replaced by the resolver: `HighlightConfigBuilder`, `HighlightValidator`, `HighlightRequestPreparer`, plus their test files. Not part of any public API.
- **`batch_inference: true` is preserved** as the legacy yield signal so existing 3.x users continue to work unchanged.

## Related Issues

Resolves #1840

## Testing

**Unit tests** — added/updated:

- `HighlightConfigResolverTests` — top-level only, inner_hits only (customer-bug shape), inner_hits inside bool (must/should/filter/mustNot) / dis_max / constant_score / boosting / function_score / script_score, custom `inner_hits.name`, mixed top-level + inner_hits, tag/options resolution, highlight_query priority, global options merge, non-semantic fields ignored, query text extraction failure graceful fallback
- `HighlightContextBuilderTests` — top-level expansion, nested expansion, custom bucket name, mixed targets, missing model_id skip, null response, empty inner-hit text skip, max_batch_size from int/string options, no_match_size and encoder propagation
- `HighlightResultApplierTests` — single span, multi-span ordering, html encoder escaping non-tag text, no_match_size fallback, leaf-name fallback for nested fields, slice-based apply, batch size mismatch
- `SemanticHighlightingProcessorTests` — processResponse throws UnsupportedOperationException, pass-through when no targets / null query text / empty context / missing model_id, single batch ML call + highlight apply, ML client error propagation, ignoreFailure returns original response, getter coverage
- `SemanticHighlightingFactoryTests` — cheap candidate check (ext-only, top-level-only, neither, both), `ext: false` not overriding top-level semantic field
- `SemanticHighlighterExtBuilderTests` — boolean parsing, stream round-trip
- `SemanticHighlighterTests` — fetch-phase yield via ext, fetch-phase yield via legacy `batch_inference`, ext disabled does not yield, graceful WARN when system factory disabled, encoder/no_match_size in the non-batch path, no_match_size with string value, no_match_size with html encoder
- `ProcessorUtilsQueryTextTests` — new compound-query unwrappers

Full unit-test run: **2451 tests, 0 failures, 0 skipped**.

**Integration tests** — added in `SemanticHighlightingRemoteModelIT`:

- `testBatchSemanticHighlightingWithNestedInnerHits` — customer bug repro using `batch_inference: true`
- `testBatchSemanticHighlightingWithExtBlockOnNestedInnerHits` — same shape via the new ext block
- `testSemanticHighlightingBatchModeWithoutSystemProcessor` — graceful degradation when the system factory is not enabled

### Step 1. Create the index

**Request**

```json
PUT /research_papers
{
  "settings": {
    "number_of_shards": 1,
    "number_of_replicas": 0
  },
  "mappings": {
    "properties": {
      "summary":  { "type": "text" },
      "abstract": { "type": "text" },
      "chunks": {
        "type": "nested",
        "properties": { "text": { "type": "text" } }
      },
      "sections": {
        "type": "nested",
        "properties": { "body": { "type": "text" } }
      }
    }
  }
}
```

**Response**

```json
{ "acknowledged": true, "shards_acknowledged": true, "index": "research_papers" }
```

The mapping has two top-level text fields (`summary`, `abstract`) and two independent nested paths (`chunks`, `sections`). This covers every combination exercised by the four tests.

### Step 2. Index two documents

**Request**

```json
POST /research_papers/_bulk?refresh=true
{ "index": { "_id": "paper_1" } }
{ "summary": "Recent breakthroughs in neurodegenerative disease therapy include monoclonal antibodies. Aducanumab targets amyloid plaques to slow cognitive decline. Lecanemab is another anti-amyloid therapy approved by the FDA.", "abstract": "We review modern pharmacological options for Alzheimer disease. Cholinesterase inhibitors remain first-line for symptomatic relief, while disease-modifying therapies such as aducanumab address underlying pathology.", "chunks": [ { "text": "Alzheimer disease is a progressive neurodegenerative disorder affecting millions of older adults." }, { "text": "Aducanumab and lecanemab are anti-amyloid antibodies. Cholinesterase inhibitors such as donepezil treat cognitive symptoms." }, { "text": "Future treatments include tau-targeting therapies and combination approaches with anti-inflammatory agents." } ], "sections": [ { "body": "Neurodegenerative diseases like Alzheimer and Parkinson share protein misfolding mechanisms." }, { "body": "Disease-modifying treatments for neurodegenerative disorders aim to slow progression rather than only manage symptoms." } ] }
{ "index": { "_id": "paper_2" } }
{ "summary": "Levodopa remains the cornerstone treatment for Parkinson disease. Deep brain stimulation can help patients with motor fluctuations. Newer therapies target alpha-synuclein aggregation.", "abstract": "This paper surveys treatments for Parkinson disease. We discuss dopaminergic therapy, surgical options, and emerging disease-modifying approaches.", "chunks": [ { "text": "Parkinson disease is the second most common neurodegenerative disorder worldwide." }, { "text": "Levodopa, dopamine agonists, and MAO-B inhibitors are mainstays of dopaminergic therapy." }, { "text": "Gene therapy and antibodies against alpha-synuclein are under active investigation as disease-modifying treatments." } ], "sections": [ { "body": "Parkinson disease is characterized by loss of dopaminergic neurons in the substantia nigra." }, { "body": "Treatment of Parkinson disease combines pharmacological, surgical, and rehabilitation strategies." } ] }
```

Two documents indexed, each with a `summary`, an `abstract`, three `chunks` entries, and two `sections` entries.

### Step 3. Enable the system-generated factory (one-time cluster setting)

**Request**

```json
PUT /_cluster/settings
{
  "persistent": {
    "cluster.search.enabled_system_generated_factories": ["semantic-highlighter"]
  }
}
```

**Response**

```json
{ "acknowledged": true }
```

### Test 1 — One top-level semantic field

**Request**

```json
POST /research_papers/_search
{
  "query": {
    "match": { "summary": "treatments for neurodegenerative diseases" }
  },
  "highlight": {
    "fields": {
      "summary": { "type": "semantic" }
    },
    "options": {
      "model_id": "<remote-model-id>",
      "batch_inference": true
    }
  }
}
```

**Response**

```json
{
  "took": 279,
  "_shards": { "total": 1, "successful": 1, "failed": 0 },
  "hits": {
    "total": { "value": 2, "relation": "eq" },
    "hits": [
      {
        "_id": "paper_2",
        "_source": { "...": "..." },
        "highlight": {
          "summary": [
            "<em>Levodopa remains the cornerstone treatment for Parkinson disease.</em> Deep brain stimulation can help patients with motor fluctuations. Newer therapies target alpha-synuclein aggregation."
          ]
        }
      },
      {
        "_id": "paper_1",
        "_source": { "...": "..." },
        "highlight": {
          "summary": [
            "Recent breakthroughs in neurodegenerative disease therapy include monoclonal antibodies. Aducanumab targets amyloid plaques to slow cognitive decline. <em>Lecanemab is another anti-amyloid therapy approved by the FDA.</em>"
          ]
        }
      }
    ]
  }
}
```

Note: this works without `ext` because the legacy `batch_inference: true` plus a top-level `type: semantic` field triggers the cheap candidate check.

### Test 2 — One inner_hits semantic field (the customer bug scenario)

**Request**

```json
POST /research_papers/_search
{
  "query": {
    "nested": {
      "path": "chunks",
      "query": {
        "match": { "chunks.text": "treatments for neurodegenerative diseases" }
      },
      "inner_hits": {
        "size": 5,
        "highlight": {
          "fields": {
            "chunks.text": { "type": "semantic" }
          },
          "options": { "model_id": "<remote-model-id>" }
        }
      }
    }
  },
  "ext": { "semantic_highlighting_batch": true }
}
```

**Response**

```json
{
  "took": 367,
  "_shards": { "failed": 0 },
  "hits": {
    "total": { "value": 2 },
    "hits": [
      {
        "_id": "paper_2",
        "inner_hits": {
          "chunks": {
            "hits": {
              "hits": [
                {
                  "_nested": { "field": "chunks", "offset": 0 },
                  "_source": { "text": "Parkinson disease is the second most common neurodegenerative disorder worldwide." },
                  "highlight": {
                    "chunks.text": [
                      "<em>Parkinson disease is the second most common neurodegenerative disorder worldwide.</em>"
                    ]
                  }
                },
                {
                  "_nested": { "field": "chunks", "offset": 2 },
                  "_source": { "text": "Gene therapy and antibodies against alpha-synuclein are under active investigation as disease-modifying treatments." },
                  "highlight": {
                    "chunks.text": [
                      "<em>Gene therapy and antibodies against alpha-synuclein are under active investigation as disease-modifying treatments.</em>"
                    ]
                  }
                }
              ]
            }
          }
        }
      },
      {
        "_id": "paper_1",
        "inner_hits": {
          "chunks": {
            "hits": {
              "hits": [
                {
                  "_nested": { "field": "chunks", "offset": 0 },
                  "_source": { "text": "Alzheimer disease is a progressive neurodegenerative disorder affecting millions of older adults." },
                  "highlight": {
                    "chunks.text": [
                      "<em>Alzheimer disease is a progressive neurodegenerative disorder affecting millions of older adults.</em>"
                    ]
                  }
                },
                {
                  "_nested": { "field": "chunks", "offset": 2 },
                  "_source": { "text": "Future treatments include tau-targeting therapies and combination approaches with anti-inflammatory agents." },
                  "highlight": {
                    "chunks.text": [
                      "<em>Future treatments include tau-targeting therapies and combination approaches with anti-inflammatory agents.</em>"
                    ]
                  }
                }
              ]
            }
          }
        }
      }
    ]
  }
}
```

The customer bug is fixed: each paper's `chunks` inner hits carry semantic highlights on `chunks.text`. Pre-fix this returned empty highlights silently. The user adds one line — `"ext": { "semantic_highlighting_batch": true }` — and the system processor walks the query tree, finds the `inner_hits` target, and runs.

### Test 3 — One top-level + one inner_hits

**Request**

```json
POST /research_papers/_search
{
  "query": {
    "bool": {
      "should": [
        { "match": { "summary": "treatments for neurodegenerative diseases" } },
        {
          "nested": {
            "path": "chunks",
            "query": {
              "match": { "chunks.text": "treatments for neurodegenerative diseases" }
            },
            "inner_hits": {
              "size": 5,
              "highlight": {
                "fields": {
                  "chunks.text": { "type": "semantic" }
                },
                "options": { "model_id": "<remote-model-id>" }
              }
            }
          }
        }
      ]
    }
  },
  "highlight": {
    "fields": {
      "summary": { "type": "semantic" }
    },
    "options": { "model_id": "<remote-model-id>" }
  },
  "ext": { "semantic_highlighting_batch": true }
}
```

**Response**

```json
{
  "took": 292,
  "_shards": { "failed": 0 },
  "hits": {
    "total": { "value": 2 },
    "hits": [
      {
        "_id": "paper_2",
        "highlight": {
          "summary": [
            "<em>Levodopa remains the cornerstone treatment for Parkinson disease.</em> Deep brain stimulation can help patients with motor fluctuations. Newer therapies target alpha-synuclein aggregation."
          ]
        },
        "inner_hits": {
          "chunks": {
            "hits": {
              "hits": [
                {
                  "_nested": { "field": "chunks", "offset": 0 },
                  "highlight": {
                    "chunks.text": [
                      "<em>Parkinson disease is the second most common neurodegenerative disorder worldwide.</em>"
                    ]
                  }
                },
                {
                  "_nested": { "field": "chunks", "offset": 2 },
                  "highlight": {
                    "chunks.text": [
                      "<em>Gene therapy and antibodies against alpha-synuclein are under active investigation as disease-modifying treatments.</em>"
                    ]
                  }
                }
              ]
            }
          }
        }
      },
      {
        "_id": "paper_1",
        "highlight": {
          "summary": [
            "Recent breakthroughs in neurodegenerative disease therapy include monoclonal antibodies. Aducanumab targets amyloid plaques to slow cognitive decline. <em>Lecanemab is another anti-amyloid therapy approved by the FDA.</em>"
          ]
        },
        "inner_hits": {
          "chunks": {
            "hits": {
              "hits": [
                {
                  "_nested": { "field": "chunks", "offset": 0 },
                  "highlight": {
                    "chunks.text": [
                      "<em>Alzheimer disease is a progressive neurodegenerative disorder affecting millions of older adults.</em>"
                    ]
                  }
                },
                {
                  "_nested": { "field": "chunks", "offset": 2 },
                  "highlight": {
                    "chunks.text": [
                      "<em>Future treatments include tau-targeting therapies and combination approaches with anti-inflammatory agents.</em>"
                    ]
                  }
                }
              ]
            }
          }
        }
      }
    ]
  }
}
```

### Test 4 — Two top-level + two inner_hits

**Request**

```json
POST /research_papers/_search
{
  "query": {
    "bool": {
      "should": [
        { "match": { "summary":  "treatments for neurodegenerative diseases" } },
        { "match": { "abstract": "treatments for neurodegenerative diseases" } },
        {
          "nested": {
            "path": "chunks",
            "query": {
              "match": { "chunks.text": "treatments for neurodegenerative diseases" }
            },
            "inner_hits": {
              "highlight": {
                "fields": {
                  "chunks.text": { "type": "semantic" }
                },
                "options": { "model_id": "<id>" }
              }
            }
          }
        },
        {
          "nested": {
            "path": "sections",
            "query": {
              "match": { "sections.body": "treatments for neurodegenerative diseases" }
            },
            "inner_hits": {
              "name": "matching_sections",
              "highlight": {
                "fields": {
                  "sections.body": { "type": "semantic" }
                },
                "options": { "model_id": "<id>" }
              }
            }
          }
        }
      ]
    }
  },
  "highlight": {
    "fields": {
      "summary":  { "type": "semantic" },
      "abstract": { "type": "semantic" }
    },
    "options": { "model_id": "<id>" }
  },
  "ext": { "semantic_highlighting_batch": true }
}
```

**Response**

```json
{
  "took": 123,
  "_shards": { "failed": 0 },
  "hits": {
    "total": { "value": 2 },
    "hits": [
      {
        "_id": "paper_1",
        "highlight": {
          "summary": [
            "Recent breakthroughs in neurodegenerative disease therapy include monoclonal antibodies. Aducanumab targets amyloid plaques to slow cognitive decline. <em>Lecanemab is another anti-amyloid therapy approved by the FDA.</em>"
          ],
          "abstract": [
            "We review modern pharmacological options for Alzheimer disease. <em>Cholinesterase inhibitors remain first-line for symptomatic relief, while disease-modifying therapies such as aducanumab address underlying pathology.</em>"
          ]
        },
        "inner_hits": {
          "chunks": {
            "hits": {
              "hits": [
                {
                  "_nested": { "field": "chunks", "offset": 0 },
                  "highlight": {
                    "chunks.text": [
                      "<em>Alzheimer disease is a progressive neurodegenerative disorder affecting millions of older adults.</em>"
                    ]
                  }
                },
                {
                  "_nested": { "field": "chunks", "offset": 2 },
                  "highlight": {
                    "chunks.text": [
                      "<em>Future treatments include tau-targeting therapies and combination approaches with anti-inflammatory agents.</em>"
                    ]
                  }
                }
              ]
            }
          },
          "matching_sections": {
            "hits": {
              "hits": [
                {
                  "_nested": { "field": "sections", "offset": 1 },
                  "highlight": {
                    "sections.body": [
                      "<em>Disease-modifying treatments for neurodegenerative disorders aim to slow progression rather than only manage symptoms.</em>"
                    ]
                  }
                },
                {
                  "_nested": { "field": "sections", "offset": 0 },
                  "highlight": {
                    "sections.body": [
                      "<em>Neurodegenerative diseases like Alzheimer and Parkinson share protein misfolding mechanisms.</em>"
                    ]
                  }
                }
              ]
            }
          }
        }
      },
      {
        "_id": "paper_2",
        "highlight": {
          "summary": [
            "<em>Levodopa remains the cornerstone treatment for Parkinson disease.</em> Deep brain stimulation can help patients with motor fluctuations. Newer therapies target alpha-synuclein aggregation."
          ]
        },
        "inner_hits": {
          "chunks": {
            "hits": {
              "hits": [
                {
                  "_nested": { "field": "chunks", "offset": 0 },
                  "highlight": {
                    "chunks.text": [
                      "<em>Parkinson disease is the second most common neurodegenerative disorder worldwide.</em>"
                    ]
                  }
                },
                {
                  "_nested": { "field": "chunks", "offset": 2 },
                  "highlight": {
                    "chunks.text": [
                      "<em>Gene therapy and antibodies against alpha-synuclein are under active investigation as disease-modifying treatments.</em>"
                    ]
                  }
                }
              ]
            }
          },
          "matching_sections": {
            "hits": { "hits": [] }
          }
        }
      }
    ]
  }
}
```

### Test 5 — Legacy `batch_inference: true` (no ext, regression check)

Existing 3.x users who already opt into batch inference via the legacy field-level flag must keep working unchanged.

**Request**

```json
POST /research_papers/_search
{
  "query": {
    "match": { "summary": "treatments for neurodegenerative diseases" }
  },
  "highlight": {
    "fields": {
      "summary": {
        "type": "semantic",
        "options": {
          "model_id": "<remote-model-id>",
          "batch_inference": true
        }
      }
    }
  }
}
```

**Response**

```json
{
  "took": 142,
  "_shards": { "failed": 0 },
  "hits": {
    "total": { "value": 2 },
    "hits": [
      {
        "_id": "paper_2",
        "_source": { "...": "..." },
        "highlight": {
          "summary": [
            "<em>Levodopa remains the cornerstone treatment for Parkinson disease.</em> Deep brain stimulation can help patients with motor fluctuations. Newer therapies target alpha-synuclein aggregation."
          ]
        }
      },
      {
        "_id": "paper_1",
        "_source": { "...": "..." },
        "highlight": {
          "summary": [
            "Recent breakthroughs in neurodegenerative disease therapy include monoclonal antibodies. Aducanumab targets amyloid plaques to slow cognitive decline. <em>Lecanemab is another anti-amyloid therapy approved by the FDA.</em>"
          ]
        }
      }
    ]
  }
}
```

No `ext` flag, no top-level `options` block — just the legacy field-level `batch_inference: true`. The cheap candidate check still passes because a top-level `type: semantic` field is present, the system processor picks the request up, and the highlights come back exactly as before. No breaking change for existing 3.x users.

### Test 6 — Non-batch synchronous path (no ext, no `batch_inference`, regression check)

The original single-inference fetch-phase path must also keep working for users who never opted into batch.

**Request**

```json
POST /research_papers/_search
{
  "query": {
    "match": { "summary": "treatments for neurodegenerative diseases" }
  },
  "highlight": {
    "fields": {
      "summary": {
        "type": "semantic",
        "options": { "model_id": "<remote-model-id>" }
      }
    }
  }
}
```

**Response**

```json
{
  "took": 318,
  "_shards": { "failed": 0 },
  "hits": {
    "total": { "value": 2 },
    "hits": [
      {
        "_id": "paper_2",
        "_source": { "...": "..." },
        "highlight": {
          "summary": [
            "<em>Levodopa remains the cornerstone treatment for Parkinson disease.</em> Deep brain stimulation can help patients with motor fluctuations. Newer therapies target alpha-synuclein aggregation."
          ]
        }
      },
      {
        "_id": "paper_1",
        "_source": { "...": "..." },
        "highlight": {
          "summary": [
            "Recent breakthroughs in neurodegenerative disease therapy include monoclonal antibodies. Aducanumab targets amyloid plaques to slow cognitive decline. <em>Lecanemab is another anti-amyloid therapy approved by the FDA.</em>"
          ]
        }
      }
    ]
  }
}
```

Neither `ext` nor `batch_inference: true` is set, so the system processor's cheap candidate check yields false on the ext signal and the factory does not generate. The fetch-phase highlighter handles each field synchronously — same path as before this PR. Highlights are produced normally.

## Check List

- [ ] API changes companion pull request created.
- [x] Commits are signed per the DCO using --signoff.
- [ ] Public documentation issue/PR created.


src/test/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingRemoteModelIT.java
```diff
@@ -392,4 +396,178 @@ public void testSemanticHighlightingWithHybridQueryWithBatchEnabledWithRemoteMod
...
+    private static final String NESTED_TEST_INDEX = "test-semantic-highlight-nested-index";
+
+    /**
+     * Customer bug reproduction: nested query with inner_hits highlighting on
+     * {@code chunks.text} in batch mode. Before the fix the system processor only
+     * inspected the top-level highlight block and silently produced no highlights.
+     * After the fix the system processor walks the query tree, discovers the
+     * inner_hits target, and returns proper highlights for every inner hit.
+     */
+    public void testBatchSemanticHighlightingWithNestedInnerHits() throws Exception {
```
heemin32: How does this test work? We are not passing ext block but how nested field get highlighted?

heemin32: It is failing which is correct behavior. We need to remove this test case actually.
```
SemanticHighlightingRemoteModelIT > testBatchSemanticHighlightingWithNestedInnerHits STANDARD_ERROR
    REPRODUCE WITH: ./gradlew ':remoteModelIntegTest' --tests 'org.opensearch.neuralsearch.highlight.SemanticHighlightingRemoteModelIT.testBatchSemanticHighlightingWithNestedInnerHits' -Dtests.seed=DF1B157FD5255554 -Dtests.security.manager=false -Dtests.locale=en-AE -Dtests.timezone=Africa/Addis_Ababa -Druntime.java=21
REPRODUCE WITH: ./gradlew ':remoteModelIntegTest' --tests 'org.opensearch.neuralsearch.highlight.SemanticHighlightingRemoteModelIT.testBatchSemanticHighlightingWithNestedInnerHits' -Dtests.seed=DF1B157FD5255554 -Dtests.security.manager=false -Dtests.locale=en-AE -Dtests.timezone=Africa/Addis_Ababa -Druntime.java=21

SemanticHighlightingRemoteModelIT > testBatchSemanticHighlightingWithNestedInnerHits FAILED
    java.lang.AssertionError: expected at least one inner_hit highlight containing 'treatments'
        at __randomizedtesting.SeedInfo.seed([DF1B157FD5255554:A82CB44CB02D070E]:0)
        at org.junit.Assert.assertTrue(Assert.java:42)
        at org.opensearch.neuralsearch.highlight.SemanticHighlightingRemoteModelIT.assertInnerHitsHaveSemanticHighlight(SemanticHighlightingRemoteModelIT.java:571)
        at org.opensearch.neuralsearch.highlight.SemanticHighlightingRemoteModelIT.testBatchSemanticHighlightingWithNestedInnerHits(SemanticHighlightingRemoteModelIT.java:429)
```

src/main/java/org/opensearch/neuralsearch/highlight/batch/config/HighlightConfigResolver.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    private static String resolveQueryText(SearchRequest request) {
+        HighlightBuilder top = request.source().highlighter();
+        if (top != null && top.highlightQuery() != null) {
+            String text = tryExtract(top.highlightQuery());
+            if (text != null) return text;
+        }
+        QueryBuilder mainQuery = request.source().query();
+        if (mainQuery == null) return null;
```
heemin32, line 161: When we fail to resolve query text, shouldn't we notify customer instead of failing silently?

src/main/java/org/opensearch/neuralsearch/highlight/batch/config/HighlightContextBuilder.java
```diff
@@ -4,117 +4,209 @@
...
-                    .context(fieldText)
-                    .build();
-
-                requests.add(request);
-                validHits.add(hit);
+        if (topHits.length > 0 && config.hasTargets()) {
+            for (SemanticHighlightTarget target : config.getTargetsOrEmpty()) {
+                String modelId = stringOption(target, SemanticHighlightingConstants.MODEL_ID, null);
+                if (modelId == null) {
+                    log.debug("Target [{}] skipped: options.model_id is required", target.getFieldName());
```
heemin32: Isn't model Id mandatory value? Shouldn't we throw an exception when model id is null?

src/main/java/org/opensearch/neuralsearch/highlight/batch/processor/SemanticHighlightingProcessor.java
```diff
@@ -56,164 +63,138 @@ public void processResponseAsync(
...
+            if (!config.hasTargets()) {
                 responseListener.onResponse(response);
                 return;
             }
 
-            // This processor only handles batch inference
-            if (!config.isBatchInference()) {
-                log.debug("Non-batch inference should be handled by SemanticHighlighter directly");
+            if (config.getQueryText() == null) {
+                log.debug("Skipping semantic highlighting: could not extract query text");
```
heemin32: Shouldn't we throw an exception?

src/main/java/org/opensearch/neuralsearch/highlight/batch/processor/SemanticHighlightingProcessor.java
```diff
@@ -56,164 +63,138 @@ public void processResponseAsync(
...
     ) {
         long batchStartTime = System.currentTimeMillis();
-
-        mlClientAccessor.batchInferenceSentenceHighlighting(
-            config.getModelId(),
+        mlCommonsClientAccessor.batchInferenceSentenceHighlighting(
+            context.getModelId(),
             context.getRequests(),
-            context.getModelType(),
+            FunctionName.REMOTE,
```
heemin32: Why did we switch from `context.getModelType()` to `FunctionName.REMOTE`?

src/main/java/org/opensearch/neuralsearch/highlight/batch/processor/SemanticHighlightingProcessor.java
```diff
@@ -56,164 +63,138 @@ public void processResponseAsync(
...
+                );
                 return;
             }
 
-            // Build context and execute batch highlighting
-            executeBatchHighlighting(enrichedConfig, response, startTime, responseListener);
+            HighlightResultApplier applier = new HighlightResultApplier();
 
+            if (context.getRequests().size() <= context.getMaxBatchSize()) {
+                processSingleBatch(context, applier, responseListener);
```
heemin32: Why do we have `processSingleBatch`? Can everything be handled by `BatchExecutor`?

src/main/java/org/opensearch/neuralsearch/highlight/batch/utils/HighlightResultApplier.java
```diff
@@ -4,203 +4,232 @@
...
-
             if (start >= 0 && end <= text.length() && start < end) {
                 result.insert(end, postTag);
                 result.insert(start, preTag);
             }
         }
+        return result.toString();
+    }
 
+    private static String htmlEscape(String text) {
```
heemin32: I see exactly same `htmlEscape` and `htmlEncodePreservingTags` in `SemanticHighlighter` class. Can we remove duplication?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
```diff
@@ -41,38 +50,25 @@ public boolean canHighlight(MappedFieldType fieldType) {
...
-                    Locale.ROOT,
-                    "Batch inference for semantic highlighting is disabled. Enable it by adding '%s' to the '%s' cluster setting.",
-                    SemanticHighlightingConstants.SYSTEM_FACTORY_TYPE,
-                    SearchPipelineService.ENABLED_SYSTEM_GENERATED_FACTORIES_SETTING.getKey()
+                log.warn(
+                    "[SEMANTIC_HIGHLIGHT] Field [{}] is configured for batch semantic highlighting but the system-generated "
+                        + "processor is not enabled. Add '{}' to the cluster setting "
+                        + "'cluster.search.enabled_system_generated_factories' to enable batch highlighting.",
+                    fieldContext.fieldName,
+                    SemanticHighlightingConstants.SYSTEM_FACTORY_TYPE
```
heemin32: Let's throw an exception as before. Customer need to know that batch highlighting is not going to get applied. Otherwise, they might just think it is slow even when batch highlighting is enabled.

src/test/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingRemoteModelIT.java
heemin32: Please update `SemanticHighlightingLocalModelIT` as well. Otherwise, you won't be able to catch the failure in local easily.

src/test/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingRemoteModelIT.java
```diff
@@ -400,46 +385,10 @@ public void testSemanticHighlightingWithHybridQueryWithBatchEnabledWithRemoteMod
...
-                log.debug("Failed to delete nested index: {}", e.getMessage());
-            }
-        }
-    }
-
-    /**
-     * Same customer scenario but using the new {@code ext.semantic_highlighting_batch}
-     * boolean opt-in instead of {@code batch_inference: true}. The system
-     * processor must produce the same highlights.
+     * Customer-bug shape using the new {@code ext.semantic_highlighting_batch}
```
heemin32: I don't think this is not Customer-bug shape because the query form is different from what customer have used.

src/test/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingLocalModelIT.java
```diff
@@ -438,4 +425,122 @@ public void testSemanticHighlightingDisabledWhenFactoryNotEnabled() throws Excep
             updateClusterSettings("cluster.search.enabled_system_generated_factories", null);
         }
     }
+
+    /**
+     * Customer-bug shape on a local model: nested query with inner_hits highlighting.
```
heemin32: This is not customer-bug shape. The query used in this test is different from what customer used.

RexyZhang2001: Fixed the javadoc

src/test/java/org/opensearch/neuralsearch/highlight/SemanticHighlightingRemoteModelIT.java
RexyZhang2001: Fixed the javadoc

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
RexyZhang2001: Fixed

src/main/java/org/opensearch/neuralsearch/highlight/batch/processor/SemanticHighlightingProcessor.java
RexyZhang2001: Switched to only using `BatchExecutor`

src/main/java/org/opensearch/neuralsearch/highlight/batch/utils/HighlightResultApplier.java
RexyZhang2001: Moved both into `HighlightEncoders` and updated `SemanticHighlighter` to call them.

src/main/java/org/opensearch/neuralsearch/highlight/batch/processor/SemanticHighlightingProcessor.java
RexyZhang2001: Both works, switched back to `context.getModelType()` to stay consistent with existing code

src/main/java/org/opensearch/neuralsearch/highlight/batch/processor/SemanticHighlightingProcessor.java
RexyZhang2001: FIxed

src/main/java/org/opensearch/neuralsearch/highlight/batch/config/HighlightContextBuilder.java
RexyZhang2001: Fixed

src/main/java/org/opensearch/neuralsearch/highlight/batch/config/HighlightConfigResolver.java
RexyZhang2001: FIxed to throw an exception

heemin32 (approved):
LGTM. Thanks! Please follow up with documentation update.

src/main/java/org/opensearch/neuralsearch/highlight/batch/config/HighlightContextBuilder.java
```diff
@@ -4,117 +4,214 @@
...
-                SentenceHighlightingRequest request = SentenceHighlightingRequest.builder()
-                    .modelId(config.getModelId())
-                    .question(config.getQueryText())
-                    .context(fieldText)
-                    .build();
-
-                requests.add(request);
-                validHits.add(hit);
+        if (topHits.length > 0 && config.hasTargets()) {
+            for (SemanticHighlightTarget target : config.getTargetsOrEmpty()) {
```
martin-gaievski, line 40: Looks like mixed legacy and non-batch top-level scenario can do double process for highlights. If request that has a top-level semantic field with `batch_inference` *not* set, plus a separate field somewhere in the request that *does* set `batch_inference: true` then following happens: fetch-phase highlighter process top-level field synchronously (per-field `extractBatchInference(options)` is false and `ext` is absent); then `shouldGenerate` returns true (because `anyTargetRequestsBatch` returns true for the other field). Processor then iterates on `getTargetsOrEmpty()` which still includes the top-level target, builds a row for it, calls the model again, and `HighlightResultApplier.applyRow` overwrites the synchronous result via `hit.highlightFields(...)`.

You can skip targets that won't have been yielded by the fetch phase highlighter. 

```
      if (!target.isNested()
              && !config.isExtEnabled()
              && !HighlightExtractorUtils.extractBatchInferenceFromOptions(target.getOptions())) {
          continue;
      }
```

This requires `HighlightConfig` to carry `extEnabled` (resolver already knows it). Add a `boolean extEnabled` to `HighlightConfig` and set it in `HighlightConfigResolver.resolve`.

src/main/java/org/opensearch/neuralsearch/highlight/batch/config/HighlightContextBuilder.java
```diff
@@ -4,117 +4,214 @@
...
+        if (target.getOptions() == null) return fallback;
+        Object v = target.getOptions().get(key);
+        if (v instanceof String) {
+            String s = (String) v;
+            return s.isEmpty() ? fallback : s;
+        }
+        return fallback;
+    }
+
+    private static int intOption(SemanticHighlightTarget target, String key, int fallback) {
```
martin-gaievski, line 184: with this refactoring this became effectively a static-but-unused. Same applies to `stringOption`, in the method `resolveMaxBatchSize` we can call helper, but we practically re-implement same logic in place. Either genuinely support per-target model IDs (which the request structure suggests was intended, model_id appears under `inner_hits.highlight.options`, not at the request level), or constrain to a single model ID for a request and reject divergence with a clear error

src/main/java/org/opensearch/neuralsearch/highlight/batch/utils/HighlightResultApplier.java
```diff
@@ -4,203 +4,192 @@
...
 
     /**
-     * Apply position-based highlights to text
+     * Walks the validated span list and inserts the pre/post tags into the source text.
+     * Assumes the caller has already confirmed at least one valid span exists.
      */
-    private String applyPositionHighlights(String text, List<Map<String, Object>> highlights, String preTag, String postTag) {
-        List<Map<String, Object>> validHighlights = new ArrayList<>();
-
+    private static String applyPositionHighlights(String text, List<Map<String, Object>> highlights, String preTag, String postTag) {
```
martin-gaievski, line 167: looks like comparator is wrong when two spans have the same start. comparator sorts by `start` descending so that inserts from the right don't shift left-side offsets. But when two spans have the same `start`, order is unspecified, and inserting two `pre+post` pairs at the same start position produces nested tags whose order depends on insertion order.
Example: spans `[5,8]` and `[5,10]` would produce `<em><em>foo</em>foobar</em>` or `<em>foo<em>foobar</em></em>` depending on order — both invalid. Most semantic models won't return overlapping spans, so this is mostly theoretical, but it's worth being defensive.
You can address this by adding tiebreaker on `end` DESC

martin-gaievski (commented):
Few high level questions: 
- you need to plan change in documentation, please open PR and add link to this PR or related GH issue
- pagination boundary failure semantic is not clear. For example, if batch [0, 100) succeeds and batch [100, 200) fails, the response will already have highlights applied to docs 0–99 but not 100+. error path calls `handleError(e, context.getOriginalResponse(), responseListener)`. With `ignoreFailure: true`, this returns the original response. not the one with partial highlights applied. With `ignoreFailure: false`, it returns failure. Either way, partial mutations to `validHits` items via `hit.highlightFields(...)` persist on the SearchHit objects that are still in the response. Is this intended? If yes, document in comment that callers should treat partial state as possible after pagination errors.

src/main/java/org/opensearch/neuralsearch/highlight/batch/utils/HighlightResultApplier.java
heemin32: Good point, but I think we should avoid over-complicating this. We assume model not to return two spans with the same start position. We can document this if needed.

src/main/java/org/opensearch/neuralsearch/highlight/batch/config/HighlightContextBuilder.java
heemin32: @martin-gaievski I cannot understand the comment. `intOption` is used to fetch `noMatchSizes` which is common parameter in sentence highlighter in OpenSearch. See https://docs.opensearch.org/latest/search-plugins/searching-data/highlight/#highlighting-options

src/main/java/org/opensearch/neuralsearch/highlight/batch/config/HighlightContextBuilder.java
heemin32: Good catch. That said, I think we can keep current code. We're planning to deprecate `batch_inference` and remove it from the documentation, and mixing batch and non-batch is not something we intend to support.

heemin32:
> With ignoreFailure: true, this returns the original response. not the one with partial highlights applied. With ignoreFailure: false, it returns failure. Either way, partial mutations to validHits items via hit.highlightFields(...) persist on the SearchHit objects that are still in the response.

@martin-gaievski. For system generated search processor, ignoreFailure is always `false`. Anyway, I think there is no harm to return partially highlighted sentence?

src/main/java/org/opensearch/neuralsearch/highlight/batch/utils/HighlightResultApplier.java
martin-gaievski: let's document that this is our assumption, we can address this properly in follow up PR

martin-gaievski:
> > With ignoreFailure: true, this returns the original response. not the one with partial highlights applied. With ignoreFailure: false, it returns failure. Either way, partial mutations to validHits items via hit.highlightFields(...) persist on the SearchHit objects that are still in the response.
> 
> @martin-gaievski. For system generated search processor, ignoreFailure is always `false`. Anyway, I think there is no harm to return partially highlighted sentence?

yes, there is no harm, we can do deep dive later

martin-gaievski (approved):

PR #1862, opened 2026-05-22
Unify highlight tag application logic and add strict validation

heemin32:
### Description
Extract shared HighlightTagApplier utility to replace divergent implementations in SemanticHighlighterEngine (forward iteration) and HighlightResultApplier (reverse insertion). Both paths now use the same forward-iteration algorithm and throw on invalid model output (non-numeric, out-of-bounds, unsorted, duplicate start, or overlapping spans).

### Related Issues
N/A

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/highlight/single/SemanticHighlighterEngine.java
+++ b/src/main/java/org/opensearch/neuralsearch/highlight/single/SemanticHighlighterEngine.java
@@ -9,6 +9,7 @@ import org.apache.lucene.search.Query;
 import org.opensearch.OpenSearchException;
 import org.opensearch.neuralsearch.highlight.single.extractor.QueryTextExtractorRegistry;
 import org.opensearch.neuralsearch.highlight.utils.HighlightExtractorUtils;
+import org.opensearch.neuralsearch.highlight.utils.HighlightTagApplier;
 import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
 import org.opensearch.neuralsearch.processor.highlight.SentenceHighlightingRequest;
 import org.opensearch.search.fetch.subphase.highlight.FieldHighlightContext;
@@ -16,7 +17,6 @@ import org.opensearch.action.support.PlainActionFuture;
 import lombok.NonNull;
 import lombok.Builder;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -29,8 +29,6 @@ import java.util.Map;
 public class SemanticHighlighterEngine {
     private static final String MODEL_ID_FIELD = "model_id";
     private static final String MODEL_INFERENCE_RESULT_KEY = "highlights";
-    private static final String MODEL_INFERENCE_RESULT_START_KEY = "start";
-    private static final String MODEL_INFERENCE_RESULT_END_KEY = "end";
 
     @NonNull
     private final MLCommonsClientAccessor mlCommonsClient;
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/neuralsearch/highlight/single/SemanticHighlighterEngine.java
```diff
@@ -152,119 +150,10 @@ public String applyHighlighting(String context, Map<String, Object> highlightRes
...
-     * @return The highlight position map
-     * @throws OpenSearchException if the item cannot be cast to Map<String, Number>
-     */
-    private static Map<String, Number> getHighlightsPositionMap(Object item) {
-        try {
-            return (Map<String, Number>) item;
-        } catch (ClassCastException e) {
-            throw new OpenSearchException(String.format(Locale.ROOT, "Expect item to be map of string to number, but was: %s", item));
-        }
+        String result = HighlightTagApplier.applyTags(context, (List<Map<String, Object>>) highlightsObj, preTag, postTag);
```
owaiskazi19, line 156: Can have guarded check here to avoid the suppresCheck
``` java
 try {
              highlightSpans = (List<Map<String, Object>>) highlightsObj;
          } catch (ClassCastException e) {
              throw new OpenSearchException(
                  String.format(Locale.ROOT, "Expected highlights to be a list of maps, but got: %s", highlightsObj.getClass().getSimpleName())
              );
          }

          String result = HighlightTagApplier.applyTags(context, highlightSpans, preTag, postTag);
```

src/main/java/org/opensearch/neuralsearch/highlight/utils/HighlightTagApplier.java
```diff
@@ -0,0 +1,111 @@
...
+                        prevStart,
+                        prevEnd,
+                        currStart,
+                        spans.get(i)[1]
+                    )
+                );
+            }
+        }
+
+        StringBuilder result = new StringBuilder();
```
owaiskazi19: We can pre allocate the size of StringBuilder since know it from text.length to avoid reallocations

martin-gaievski (approved):
LGTM, thx

src/main/java/org/opensearch/neuralsearch/highlight/utils/HighlightTagApplier.java
```diff
@@ -0,0 +1,111 @@
...
+    public static String applyTags(String text, List<Map<String, Object>> highlights, String preTag, String postTag) {
+        if (highlights == null || highlights.isEmpty()) {
+            return null;
+        }
+
+        List<int[]> spans = new ArrayList<>(highlights.size());
+        for (Map<String, Object> highlight : highlights) {
+            Object startObj = highlight.get(SemanticHighlightingConstants.START_KEY);
+            Object endObj = highlight.get(SemanticHighlightingConstants.END_KEY);
+            if (!ProcessorUtils.isNumeric(startObj) || !ProcessorUtils.isNumeric(endObj)) {
```
owaiskazi19, line 43: We are fine from user perspective to throw an exception with one bad span and kill the entire highlighting?

heemin32: Yes. If the span is not correct, it means there is a bug in model wrapper code. Better to let the customer knows by throwing an exception.

src/main/java/org/opensearch/neuralsearch/highlight/single/SemanticHighlighterEngine.java
heemin32: At runtime, (List<Map<String, Object>>) highlightsObj just checks that highlightsObj is a List. The generic parameters <Map<String,Object>> are erased. Since we already verified highlightsObj instanceof List<?> above, the cast will never throw ClassCastException at that line. 

Also, even with this change, we cannot avoid suppressCheck.

owaiskazi19 (approved):

PR #1870, opened 2026-06-13
Improve hybrid query filter validation error message

### LINKED ISSUE #1811: [FEATURE] Improve the validation error info for the hybrid query filter field
Yes. Hybrid query supports a common filter since 3.0 but only expect the filter to be a query object which means it cannot support the filter like 
```
"filter": [
        {
          "term": { "field1": "value1" }
        },
        {
          "term": { "field2": "value2" }
        }
]
```

We should provide better error info rather than simply saying “Field is not supported by [hybrid] query”.

Besides we should also improve our doc to include this info and guide users that they can use filter like
```
"filter": {
  "bool": {
    "must": [
      {"term": {"field1": "value1"}},
      {"term": {"field2": "value2"}}
    ]
  }
}
```

https://docs.opensearch.org/latest/query-dsl/compound/hybrid/
https://opensearch.org/blog/introducing-common-filter-support-for-hybrid-search-queries/

Or we can consider support the array format.

venkateshwaracholan:
Description

Improve validation error messages for invalid hybrid query filter formats.

Currently, when the filter field is provided in an unsupported format (for example, an array), hybrid query parsing returns the generic error:

Field is not supported by [hybrid] query

This change adds a dedicated validation error for invalid filter formats and provides guidance to use a bool query with must clauses when combining multiple filters.

Related Issues

Fixes #1811 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
@@ -71,6 +71,8 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
     public static final String ERROR_MSG_QUERIES_REQUIRED = "[%s] requires 'queries' field with at least one clause";
     public static final String ERROR_MSG_MAX_QUERIES_EXCEEDED = "Number of sub-queries exceeds maximum supported by [%s] query";
     public static final String ERROR_MSG_BOOST_NOT_SUPPORTED = "[%s] query does not support [%s]";
+    public static final String ERROR_MSG_FILTER_MUST_BE_QUERY_OBJECT = "[%s] query's [%s] field must be a query object. "
+        + "For multiple filters, use a bool query with must clauses";
 
     public HybridQueryBuilder(StreamInput in) throws IOException {
         super(in);
@@ -186,12 +188,25 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
      *                          "text": "keyword"
      *                       }
      *                    }
-     *               ]
-     *               "filter":
+     *               ],
+     *               "filter": {
      *                  "term": {
      *                      "text": "keyword"
      *                  }
+     *               }
      *          }
+     *     }
+     * }
+     *
+     * To combine multiple filter clauses, wrap them in a bool query:
+     * {
+     *     "filter": {
+     *         "bool": {
+     *             "must": [
+     *                 {"term": {"field1": "value1"}},
+     *                 {"term": {"field2": "value2"}}
+     *             ]
+     *         }
      *     }
      * }
      *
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -71,6 +71,8 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
     public static final String ERROR_MSG_QUERIES_REQUIRED = "[%s] requires 'queries' field with at least one clause";
     public static final String ERROR_MSG_MAX_QUERIES_EXCEEDED = "Number of sub-queries exceeds maximum supported by [%s] query";
     public static final String ERROR_MSG_BOOST_NOT_SUPPORTED = "[%s] query does not support [%s]";
+    public static final String ERROR_MSG_FILTER_MUST_BE_QUERY_OBJECT = "[%s] query's [%s] field must be a query object. "
+        + "For multiple filters, use a bool query with must clauses";
```
heemin32: I think we can omit this part. `[%s] query's [%s] field must be a query object.` should be sufficient.

The phrase `use a bool query with must clauses` could be misleading, as it may give users the impression that they need to use a bool query instead of a hybrid query.

bzhangam: Potentially we can add some logic to convert an array of the filters to a bool query with must clauses to make this feature more smart.

venkateshwaracholan: Thanks for the suggestions. I can see how automatically converting an array of filters into a bool query could improve the user experience.

For this PR I tried to keep the scope limited to improving the validation error message and making the expected filter format clearer. Automatic conversion would introduce new parsing behavior and query semantics, so it may be better handled as a separate enhancement discussion if the community would like to support that pattern.

It's an interesting idea though, and I'd be happy to explore it further in a follow-up enhancement.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -440,4 +446,15 @@ public static void updateQueryStats(boolean hasFilter, boolean hasPagination, bo
...
     }
+
+    private static void throwUnsupportedFilterParsingException(XContentParser parser) {
+        throw new ParsingException(
+            parser.getTokenLocation(),
+            String.format(Locale.ROOT, ERROR_MSG_FILTER_MUST_BE_QUERY_OBJECT, NAME, FILTER_FIELD.getPreferredName())
+        );
+    }
+
+    private static void throwUnsupportedFieldParsingException(XContentParser parser, String fieldName) {
```
martin-gaievski, line 457: `fieldName` param is not used, better inject it into error message: `"Field [%s] is not supported by [%s] query"`

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -218,11 +232,7 @@ public static HybridQueryBuilder fromXContent(XContentParser parser) throws IOEx
                 } else if (FILTER_FIELD.match(currentFieldName, parser.getDeprecationHandler())) {
                     filter = parseInnerQueryBuilder(parser);
                 } else {
-                    log.error(String.format(Locale.ROOT, "[%s] query does not support [%s]", NAME, currentFieldName));
```
martin-gaievski, line 221: with dropped log.error this became inconsistent with `boost` branch in this method, we need to pick one approach - either leave log.error here or drop it from boost related branch

martin-gaievski (commented):
In addition to mentioned comments please add changelog entry

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
heemin32: We don't want to return customer provide value back to customer without validating it. That is why it was omitted in the past and that is why we logged field name instead of adding it in the exception message.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
venkateshwaracholan: Thanks. I've restored the `log.error()` call in the unsupported-field path so the behavior remains consistent with the existing boost validation path.

heemin32 (approved):

PR #1874, opened 2026-06-18
Push down hybrid filters to nested neural/kNN queries

### LINKED ISSUE #1759: [BUG] hybrid.filter not propagated as kNN pre-filter when neural query is inside a nested query
### What is the bug?

When using a hybrid query with `filter` (introduced in OpenSearch 3.0) combined with a `nested` neural/kNN subquery, the filter is not pushed down into the kNN engine as a pre-filter. Instead, it appears to be applied **after** the approximate nearest neighbor search has already selected its top-k candidates globally.

This means that if the target document's embedding is not in the global top-k nearest neighbors, it gets excluded from the kNN results entirely, even though it matches the filter. The filter effectively behaves as a post-filter for the neural subquery, while correctly behaving as a pre-filter for the keyword/match subquery.

This contradicts the documented behavior described in the [[common filter support blog post](https://opensearch.org/blog/introducing-common-filter-support-for-hybrid-search-queries/)](https://opensearch.org/blog/introducing-common-filter-support-for-hybrid-search-queries/), which states:

> *"In practice, this might be implemented by [...] using the filtering capabilities of the subquery type (for example, the k-NN engine can apply a pre-filter on documents before vector similarity search)."*

**Prerequisites:**
- OpenSearch 3.x cluster
- An index with nested embeddings (knn_vector + text) and a search pipeline with normalization-processor
- Enough documents (5k+) so that the target document is not in the global kNN top-k for the given query

**Index mapping (simplified):**
```json
{
  "embeddings": {
    "type": "nested",
    "properties": {
      "text": { "type": "text" },
      "embedding": { "type": "knn_vector", "dimension": 4096 }
    }
  }
}
```

**Steps to reproduce:**

1. **Standalone text query with bool filter** - returns the **1 document correctly**
```json
GET /my-index/_search
{
  "query": {
    "bool": {
      "must": [
        {
          "nested": {
            "path": "embeddings",
            "query": { "match": { "embeddings.text": "search terms" } }
          }
        }
      ],
      "filter": [ { "terms": { "_id": ["target-doc-id"] } } ]
    }
  }
}
```

2. **Standalone neural query with bool filter** - returns **0 results**
```json
GET /my-index/_search
{
  "query": {
    "bool": {
      "must": [
        {
          "nested": {
            "path": "embeddings",
            "query": {
              "neural": {
                "embeddings.embedding": {
                  "query_text": "search terms",
                  "model_id": "my-model-id",
                  "k": 100
                }
              }
            }
          }
        }
      ],
      "filter": [ { "terms": { "_id": ["target-doc-id"] } } ]
    }
  }
}
```
The document's embedding is not in the global top-100 nearest neighbors, so it is excluded before the filter is applied.

3. **Neural query with filter inside kNN**  returns the document **correctly**
```json
GET /my-index/_search
{
  "query": {
    "nested": {
      "path": "embeddings",
      "query": {
        "neural": {
          "embeddings.embedding": {
            "query_text": "search terms",
            "model_id": "my-model-id",
            "k": 100,
            "filter": { "terms": { "_id": ["target-doc-id"] } }
          }
        }
      }
    }
  }
}
```
Here the kNN engine correctly pre-filters to the target document before running ANN search.

4. **Hybrid query with `hybrid.filter`** - document found but only via text subquery, neural contributes nothing
```json
GET /my-index/_search?search_pipeline=my-pipeline
{
  "query": {
    "hybrid": {
      "filter": {
        "terms": { "_id": ["target-doc-id"] }
      },
      "queries": [
        {
          "nested": {
            "path": "embeddings",
            "query": { "match": { "embeddings.text": "search terms" } }
          }
        },
        {
          "nested": {
            "path": "embeddings",
            "query": {
              "neural": {
                "embeddings.embedding": {
                  "query_text": "search terms",
                  "model_id": "my-model-id",
                  "k": 100
                }
              }
            }
          }
        }
      ]
    }
  }
}
```
The text subquery finds the document (filter works correctly as pre-filter for BM25). The neural subquery does not find it because the `hybrid.filter` is not propagated into the kNN engine as a pre-filter, it is applied after the global top-k selection, at which point the document is already excluded.

When `hybrid.filter` is specified, it should be pushed down into the kNN/neural subquery as a **pre-filter**, so the kNN engine restricts its candidate set to the filtered documents before performing the approximate nearest neighbor search. This is the behavior documented in the common filter support blog post.

Currently, the filter is correctly propagated for BM25/match subqueries but **not** for neural/kNN subqueries when they are wrapped in a `nested` query.

### Workaround

Manually duplicate the filter inside the neural query's `filter` parameter:

```json
{
  "neural": {
    "embeddings.embedding": {
      "query_text": "...",
      "model_id": "...",
      "k": 100,
      "filter": { "terms": { "_id": ["target-doc-id"] } }
    }
  }
}
```

This forces proper kNN pre-filtering but defeats the purpose of the common filter feature.

- OpenSearch 3.x
- Index with nested knn_vector fields
- Search pipeline with min_max normalization and arithmetic_mean combination

### Related issues
https://github.com/opensearch-project/neural-search/issues/1705 has similar symptom (hybrid dropping relevant kNN documents) but different root cause (position overflow in HybridBulkScorer)

venkateshwaracholan:
## Summary

Fix `hybrid.filter` propagation for nested neural/kNN sub-queries.

When `hybrid.filter` is applied to a hybrid query containing a nested neural/kNN query, the filter was previously wrapped at the parent level and never reached the underlying kNN query. As a result, ANN search ran against the global candidate set and filtering behaved effectively as a post-filter.

This PR pushes the filter into the inner neural/kNN query so the kNN engine can pre-filter candidates before ANN search.

Fixes #1759.

## Root Cause

`HybridQueryBuilder.filter()` distributes filters by calling:

```java
query.filter(filter)
```

on each sub-query.

`NestedQueryBuilder` does not override `filter()` and inherits the default implementation from `AbstractQueryBuilder`, which wraps the query in a parent-level `BoolQueryBuilder`:

```java
BoolQueryBuilder b = new BoolQueryBuilder();
b.must(this);
b.filter(filter);
```

This keeps the filter outside the nested neural/kNN query, preventing it from reaching the kNN engine as a pre-filter.
This matches the tentative root cause identified by @martin-gaievski in #1759.

## Solution

Introduced `HybridQueryFilterUtil.applyFilterToSubQuery()` and applied it across all hybrid filter entry points:

* `HybridQueryBuilder.filter()`
* `HybridQueryBuilder.fromXContent()` (REST)
* `HybridQueryBuilderProtoUtils.fromProto()` (gRPC)

For nested sub-queries whose inner query supports vector pre-filtering (`NeuralQueryBuilder`, `KNNQueryBuilder`, `NeuralKNNQueryBuilder`), the filter is pushed down to the inner query.

All other query types retain the existing behavior.

Also added `NeuralKNNQueryBuilder.filter()` to delegate filter application to the underlying `KNNQueryBuilder`.

## Tests

Added:

* Unit tests validating filter pushdown for nested neural queries.
* Regression tests verifying non-vector nested queries retain existing behavior.
* REST parsing coverage.
* Integration test reproducing the issue scenario and verifying neural contribution is preserved when `hybrid.filter` is used.

### Validation

```bash
./gradlew test
./gradlew integTest --tests "org.opensearch.neuralsearch.query.HybridQueryNestedFilterIT"
```

## Notes

This change intentionally focuses on dense neural/kNN queries. Sparse ANN query types are left unchanged to avoid expanding scope and can be addressed separately if needed.

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Enhancements
 
 ### Bug Fixes
-
+* Fix hybrid.filter not propagated as kNN pre-filter for nested neural/kNN sub-queries ([#1874](https://github.com/opensearch-project/neural-search/pull/1874))
 ### Infrastructure
 
 ### Documentation
--- a/src/main/java/org/opensearch/neuralsearch/grpc/proto/request/search/query/HybridQueryBuilderProtoUtils.java
+++ b/src/main/java/org/opensearch/neuralsearch/grpc/proto/request/search/query/HybridQueryBuilderProtoUtils.java
@@ -9,6 +9,7 @@ import org.opensearch.core.xcontent.XContentParser;
 import org.opensearch.index.query.InnerHitContextBuilder;
 import org.opensearch.index.query.QueryBuilder;
 import org.opensearch.neuralsearch.query.HybridQueryBuilder;
+import org.opensearch.neuralsearch.util.HybridQueryFilterUtil;
 import org.opensearch.transport.grpc.spi.QueryBuilderProtoConverterRegistry;
 import org.opensearch.protobufs.HybridQuery;
 import org.opensearch.protobufs.QueryContainer;
@@ -97,7 +98,7 @@ public class HybridQueryBuilderProtoUtils {
             if (filter == null) {
                 compoundQueryBuilder.add(query);
             } else {
-                compoundQueryBuilder.add(query.filter(filter));
+                compoundQueryBuilder.add(HybridQueryFilterUtil.applyFilterToSubQuery(query, filter));
             }
 
             if (hasInnerHits == false) {
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
@@ -41,6 +41,7 @@ import lombok.experimental.Accessors;
 import lombok.extern.log4j.Log4j2;
 import org.opensearch.neuralsearch.stats.events.EventStatName;
 import org.opensearch.neuralsearch.stats.events.EventStatsManager;
+import org.opensearch.neuralsearch.util.HybridQueryFilterUtil;
 
 import static org.opensearch.neuralsearch.common.MinClusterVersionUtil.isClusterOnOrAfterMinReqVersionForPaginationInHybridQuery;
 
@@ -120,8 +121,8 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
         ListIterator<QueryBuilder> iterator = queries.listIterator();
         while (iterator.hasNext()) {
             QueryBuilder query = iterator.next();
-            // set the query again because query.filter(filter) can return new query.
-            iterator.set(query.filter(filter));
+            // set the query again because applyFilterToSubQuery can return new query.
+            iterator.set(HybridQueryFilterUtil.applyFilterToSubQuery(query, filter));
         }
         return this;
     }
@@ -281,7 +282,7 @@ public final class HybridQueryBuilder extends AbstractQueryBuilder<HybridQueryBu
             if (filter == null) {
                 compoundQueryBuilder.add(query);
             } else {
-                compoundQueryBuilder.add(query.filter(filter));
+                compoundQueryBuilder.add(HybridQueryFilterUtil.applyFilterToSubQuery(query, filter));
             }
 
             // Check if children have inner hits for stats
--- a/src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
@@ -403,6 +403,21 @@ public class NeuralKNNQueryBuilder extends AbstractQueryBuilder<NeuralKNNQueryBu
     }
 
     /**
+     * Applies a filter to the underlying KNN query so the kNN engine can pre-filter candidates.
+     *
+     * @param filterToBeAdded filter to apply
+     * @return a new builder with the filter applied to the underlying KNN query
+     */
+    @Override
+    public QueryBuilder filter(QueryBuilder filterToBeAdded) {
+        if (validateFilterParams(filterToBeAdded) == false) {
+            return this;
+        }
+        KNNQueryBuilder filteredKnnQueryBuilder = (KNNQueryBuilder) knnQueryBuilder.filter(filterToBeAdded);
+        return new NeuralKNNQueryBuilder(filteredKnnQueryBuilder, originalQueryText);
+    }
+
+    /**
      * Gets the name of this query for serialization purposes.
      *
      * @return The writeable name
--- a/src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/NeuralKNNQueryBuilder.java
@@ -414,7 +414,10 @@ public class NeuralKNNQueryBuilder extends AbstractQueryBuilder<NeuralKNNQueryBu
             return this;
         }
         KNNQueryBuilder filteredKnnQueryBuilder = (KNNQueryBuilder) knnQueryBuilder.filter(filterToBeAdded);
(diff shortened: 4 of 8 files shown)


bzhangam:
I believe the fix belongs in OpenSearch core (`[NestedQueryBuilder](https://github.com/opensearch-project/OpenSearch/blob/a1a29a4a26e5f6027696290dcf4755e20ba3f61d/server/src/main/java/org/opensearch/index/query/NestedQueryBuilder.java#L83)`) rather than as a utility function in neural-search.

### Why `NestedQueryBuilder` should override `filter()`

The `filter()` contract on `AbstractQueryBuilder` exists so that callers can generically push a filter down into any query builder. The default implementation (`bool { must: this, filter: f }`) is correct for most queries, but `NestedQueryBuilder` is a special case:

1. **`NestedQueryBuilder` does not support a top-level `filter` parameter** in its DSL — there is no `"nested": { "filter": ... }` syntax. The only semantically correct place for a filter is _inside_ the nested `query` clause.

2. **The default wrapping produces incorrect semantics** — wrapping the entire nested query in a parent-level bool+filter means the filter is applied to _parent_ documents after the nested join, not to the _nested documents_ before the inner query executes. This affects any inner query type, not just kNN.

3. **This is a general problem, not specific to hybrid queries** — any caller of `nestedQueryBuilder.filter(f)` will hit this bug. Fixing it in `HybridQueryFilterUtil` only patches one caller while leaving the core API broken for all others.

4. **Fix in hybrid query cannot handle nested query in a nested query** - it's possible we have multiple nested layers which require a nested query in another nested query. In that case the util function in the hybrid query cannot handle it.

### Proposed fix: override `filter()` in `NestedQueryBuilder`

The override should push the filter down into the inner query:

```java
// In NestedQueryBuilder.java (OpenSearch core)
@Override
public QueryBuilder filter(QueryBuilder filter) {
    if (validateFilterParams(filter) == false) {
        return this;
    }
    // NestedQueryBuilder doesn't support a top-level filter — push it into the inner query.
    QueryBuilder filteredInnerQuery = query.filter(filter);
    NestedQueryBuilder result = new NestedQueryBuilder(path, filteredInnerQuery, scoreMode, innerHitBuilder);
    result.ignoreUnmapped(ignoreUnmapped);
    result.boost(boost());
    result.queryName(queryName());
    return result;
}
```

This is correct because:
- It delegates to the inner query's own `filter()` implementation. If the inner query is a `KNNQueryBuilder`, it already overrides `filter()` to set kNN pre-filter. If it's a regular `BoolQueryBuilder` or `MatchQueryBuilder`, the default bool wrapping applies at the nested document level — which is the correct behavior.
- It preserves all metadata (`innerHitBuilder`, `ignoreUnmapped`, `boost`, `queryName`).
- It makes `NestedQueryBuilder.filter()` work correctly for **all callers** — hybrid, bool, or any future query that calls `filter()` generically.

### What this means for this PR

With the core fix in place, the `HybridQueryFilterUtil` class and the special-casing in `HybridQueryBuilder.filter()` / `fromXContent()` / `fromProto()` become unnecessary. The existing code path:

```java
query.filter(filter)  // in HybridQueryBuilder
```

would already do the right thing for nested queries, because `NestedQueryBuilder.filter()` would push down correctly.

I'd recommend:
1. Open a PR against [opensearch-project/OpenSearch](https://github.com/opensearch-project/OpenSearch) to add the `filter()` override to `NestedQueryBuilder`.
2. Once that's merged, this neural-search PR can be dropped entirely.

This avoids coupling neural-search to the internal structure of `NestedQueryBuilder` (the `copyNestedQueryBuilder` helper, `instanceof` checks, etc.) and fixes the problem at its source for all consumers.

venkateshwaracholan:
@bzhangam  Thanks for the suggestion. I've moved the fix into OpenSearch core as proposed: OpenSearch #22355.

The core change overrides `NestedQueryBuilder.filter()` to delegate to the inner query (`query.filter(filter)`), so the behavior works for all callers and naturally supports nested-in-nested queries.

Once the core fix is merged, I'll close this PR (or keep only any plugin-specific regression tests if they're still needed).

heemin32:
> @bzhangam Thanks for the suggestion. I've moved the fix into OpenSearch core as proposed: OpenSearch #22355.
> 
> The core change overrides `NestedQueryBuilder.filter()` to delegate to the inner query (`query.filter(filter)`), so the behavior works for all callers and naturally supports nested-in-nested queries.
> 
> Once the core fix is merged, I'll close this PR (or keep only any plugin-specific regression tests if they're still needed).

Keeping plugin specific test might be good. Thanks!

PR #1885, opened 2026-06-28
Align ML Commons FunctionName with registered model configuration

venkateshwaracholan:
### Description
Align the FunctionName used in ML Commons inference requests with the algorithm configured in the registered ML Commons model.

Previously, several inference paths in NeuralSearchMLInputBuilder and MLCommonsClientAccessor used hardcoded FunctionName values (for example, TEXT_EMBEDDING) regardless of the model's registered algorithm. Although ML Commons currently accepts these requests, the resulting MLInput did not accurately reflect the model configuration.

This change updates inference request construction to use the model's registered algorithm consistently across supported inference paths.

Changes Made
Production Changes
Added logic to resolve FunctionName from the registered model configuration.
Updated NeuralSearchMLInputBuilder to construct MLInput using the model's configured algorithm.
Updated MLCommonsClientAccessor to propagate model algorithm information when building inference requests.
Aligned FunctionName handling for:
Local models
Remote symmetric models
Remote asymmetric models
Sparse encoding models
Text similarity inference
Semantic highlighting inference

### Check List


dummy.txt
```diff
@@ -0,0 +1 @@
+test
```
yuye-aws: Is this txt file really needed?

heemin32:
Please update changelog.

dummy.txt
venkateshwaracholan: thanks, removed

heemin32:
Is there a known case where the current hardcoded `FunctionName` causes a failure? I don't see the benefit of resolving the algorithm from the model.

If ML Commons actually requires the `FunctionName` to match the registered algorithm, then ML Commons should resolve it internally from the model ID. The caller shouldn't need to pass it at all.

The hardcoded `FunctionName` serves as a safeguard: if a user accidentally configures the wrong model type for a feature (e.g., a text similarity model where a text embedding model is expected), ML Commons can reject the request due to the mismatch. With this change, we'd silently pass through whatever algorithm the model has, potentially producing invalid results without the user knowing they misconfigured the model.

heemin32:
@pyek-bot Does it mean, even this inference processor does not need to pass function name? https://docs.opensearch.org/latest/ingest-pipelines/processors/ml-inference/

heemin32:
I think we should keep the current approach. Fetching model metadata would introduce latency we'd prefer to avoid. It looks like the function name is used for local models but not for remote ones. Since we already know which model we're calling, we should hardcode the function name rather than reading it from model metadata to avoids the unnecessary overhead.

heemin32:
Closing the PR

PR #1915, opened 2026-07-28
HighlightConfigResolver: use QueryBuilderVisitor

nomoa:
### Description
instead of manually walking through the query tree prefer using a QueryBuilderVisitor.
This reduces the complexity of the code and the number of classes it depends on but also allows traversing queries that may not be available to the plugin classpath. It should also be a bit more futureproof.

### Check List
- ~[ ] New functionality has been documented.~
- ~[ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).~
- ~[ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,3 +20,4 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Maintenance
 
 ### Refactoring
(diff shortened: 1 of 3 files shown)


heemin32 (approved):
LGTM. Thanks.

PR #1917, opened 2026-07-28
Add SemanticHighlighterQueryEnricherProcessor

### LINKED ISSUE #1916: [FEATURE] Provide a search processor to enrich the semantic highligher model id
Unlike the neural query there does not seem to be a way to setup a search pipeline that will populate the model_id of the semantic highlighter must use.
In setups with multiple clusters (multi-region) a client might not know what cluster it is using, model ids being cluster specific this makes it hard for such clients to build a cluster agnostic query.
Storing search templates could be a solution but this might be hard for existing client logic to switch to a template approach.
This problem is nicely solved for the neural query with a processor that can be added to a search pipeline and this pipeline be made the default of the target index. This ultimately allows to completely decouple the query building logic from the actual model ids generated on a particular cluster.
 

A query processor should be available to enrich the model id of the semantic highlighter.

One may workaround this issue with stored templates but this generally involves a big change on existing clients.

nomoa:
### Description
A new search request processor to enrich the semantic highlighter model id.
Works the same way as the neural_query_enricher allowing to define a global or a per field model id.
The processor inspects the top level highlight config but also any highlight definitions added to nested query inner hits.

Limitations: it does not inspect aggregations (which could possibly be useful for top hits aggregation) and this is left as a possible followup. Main reason for not doing it now is that the batch inference does not support aggregations yet.

### Related Issues
Resolves #1916

### Check List
- ~[ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).~

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/stats/events/EventStatName.java
+++ b/src/main/java/org/opensearch/neuralsearch/stats/events/EventStatName.java
@@ -250,6 +250,13 @@ public enum EventStatName implements StatName {
         EventStatType.TIMESTAMPED_EVENT_COUNTER,
         Version.V_3_1_0
     ),
+    /** Tracks executions of the semantic highlighter query enricher processor */
+    SEMANTIC_HIGHLIGHTING_QUERY_ENRICHER_EXECUTIONS(
+        "semantic_highlighting_query_enricher_executions",
+        "processors.search",
+        EventStatType.TIMESTAMPED_EVENT_COUNTER,
+        Version.V_3_8_0
+    ),
     /** Tracks executions of the ML reranking processor */
     RERANK_ML_PROCESSOR_EXECUTIONS("rerank_ml_executions", "processors.search", EventStatType.TIMESTAMPED_EVENT_COUNTER, Version.V_3_1_0),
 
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/neuralsearch/stats/events/EventStatName.java
```diff
@@ -250,6 +250,13 @@ public enum EventStatName implements StatName {
         EventStatType.TIMESTAMPED_EVENT_COUNTER,
         Version.V_3_1_0
     ),
+    /** Tracks executions of the semantic highlighter query enricher processor */
+    SEMANTIC_HIGHLIGHTING_QUERY_ENRICHER_EXECUTIONS(
+        "semantic_highlighting_query_enricher_executions",
+        "processors.search",
+        EventStatType.TIMESTAMPED_EVENT_COUNTER,
+        Version.V_3_8_0
```
nomoa, line 258: unsure about this, if 3.8.0 has been released, this should probably be 3.9.0 but current dependencies still target 3.8

PR #1919, opened 2026-07-29
[Enhancement] Add model_selection parameter to semantic field to resolve model id from cluster settings

### LINKED ISSUE #1918: [Enhancement] Add language_option and model_type parameters to semantic field for automatic model setup
## Overview

Currently, creating a semantic field requires the customer to manually register, deploy, and provide a `model_id`. This is a significant barrier to adoption — especially for customers who just want semantic search and don't want to manage ML models.

This enhancement adds two optional parameters to the semantic field — `language_option` and `model_type` — that allow the system to automatically resolve, register, and deploy the appropriate pretrained model. The customer describes *what they want* (language + model type), and the system handles the *how* (model selection, download, deployment).

## Proposed API

```json
PUT /my-index
{
  "mappings": {
    "properties": {
      "title": {
        "type": "semantic",
        "language_option": "ENGLISH",
        "model_type": "SPARSE"
      }
    }
  }
}
```

No `model_id` needed. The system resolves a managed pretrained model automatically.

## Parameters

| Parameter | Values | Default |
|---|---|---|
| `language_option` | `ENGLISH`, `MULTILINGUAL` | `ENGLISH` |
| `model_type` | `SPARSE`, `DENSE` | `SPARSE` |

**Defaults:** If neither `model_id` nor `language_option`/`model_type` is specified, the system defaults to `language_option: ENGLISH` + `model_type: SPARSE`.

**Conflict rule:** If `model_id` is specified alongside `language_option` or `model_type`, the request is rejected with a conflict error. These are mutually exclusive — either the customer supplies their own model, or the system resolves one from the language/model_type combination.

## Model Resolution Table

| language_option | model_type | Pretrained Model |
|---|---|---|
| ENGLISH | SPARSE | `amazon/neural-sparse/opensearch-neural-sparse-encoding-doc-v2-mini` |
| ENGLISH | DENSE | `huggingface/sentence-transformers/paraphrase-MiniLM-L3-v2` |
| MULTILINGUAL | SPARSE | `amazon/neural-sparse/opensearch-neural-sparse-encoding-multilingual-v1` |
| MULTILINGUAL | DENSE | `huggingface/sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2` |

### Architecture

Introduce a `SemanticModelResolver` interface with a default implementation (`PretrainedSemanticModelResolver`):

```
SemanticMappingTransformer
  ├── fields with model_id → existing path (fetch model metadata, expand companion field)
  └── fields with language_option/model_type (no model_id) → SemanticModelResolver.resolve()
        ├── in-memory cache hit → return model_id instantly
        ├── cache miss → search .plugins-ml-model for existing DEPLOYED model with same name
        │     ├── found → cache + return (no re-download)
        │     └── not found → register pretrained model + deploy + cache
        └── resolved model_id written into field config → existing expansion path continues
```

### Key Components

1. **`SemanticModelResolver` (interface)** — `resolve(language, modelType, listener)` returns a deployed `model_id` asynchronously. Also provides `validate(language, modelType)` for early conflict detection.

2. **`PretrainedSemanticModelResolver` (default implementation)** — Maps `language_option`/`model_type` to the pretrained model name+version, searches ml-commons for an existing deployed instance, and only registers+deploys on cache miss. Uses an in-memory `ConcurrentHashMap` cache for same-node fast-path.

3. **`SemanticMappingTransformer` changes** — Splits incoming fields into `model_id` fields (existing path) vs. managed fields (resolver path). After resolution, both paths converge on the same `fetchModelAndModifyMapping` logic.

4. **`SemanticFieldMapper` changes** — Accepts `language_option`/`model_type` as stored parameters; returns them in GET mapping response. Validates mutual exclusion with `model_id`.

5. **Plugin wiring** — `NeuralSearch.createComponents()` instantiates the resolver and injects it into the transformer via `setModelResolver()`.

### Optimization: Search-Before-Deploy

To avoid duplicate model deployments (across multiple semantic fields, node restarts, or multi-node clusters), the resolver searches the `.plugins-ml-model` index for an already-deployed model with the matching name before downloading/registering. This search uses `ThreadContext.stashContext()` to execute with system privileges (independent of calling user's permissions).

### Performance Note

The first CreateIndex with a new `language_option`/`model_type` combination will incur higher latency due to model download and deployment (can be 30–60s depending on model size). Subsequent calls with the same combination return near-instantly via cache or search-before-deploy. This is a one-time cost per model per cluster.

## Validation Rules

- `model_id` + (`language_option` or `model_type`) specified together → **400 conflict error**
- Unknown `language_option` value → **400 error** ("Unsupported language option")
- Unknown `model_type` value → **400 error** ("Unsupported model type")

## Scope / Non-Goals

This enhancement covers automatic model resolution for CreateIndex and PutMapping (adding new semantic fields). It does **not** include:
- Model lifecycle management (undeploy, delete) — out of scope for this issue

## GET Mapping Response

After creation, the mapping includes the resolved `model_id` alongside the customer's `language_option`/`model_type`:

```json
{
  "title": {
    "type": "semantic",
    "model_id": "<auto-resolved>",
    "language_option": "ENGLISH",
    "model_type": "SPARSE"
  }
}
```

yizheliu-amazon:
## Summary

Resolves #1918

Adds a `model_selection` parameter to the `semantic` field type so customers can create a semantic field **without specifying a `model_id`**. Instead of a raw model id, they declare *what they want* (language + model type) via `model_selection`, and the system resolves the `model_id` from operator-configured cluster settings.

```json
PUT /my-index
{
  "mappings": {
    "properties": {
      "title": {
        "type": "semantic",
        "model_selection": { "language_option": "ENGLISH", "model_type": "SPARSE" }
      }
    }
  }
}
```

`model_id` remains a top-level parameter. `model_selection` is a sibling nested object holding `language_option` (`ENGLISH` | `MULTILINGUAL`, default `ENGLISH`) and `model_type` (`SPARSE` | `DENSE`, default `SPARSE`).

## How it works

The plugin does **not** register or deploy any model. The operator deploys the model and points a cluster setting at it; the plugin resolves the `model_id` from that setting at index-creation/put-mapping time.

1. Operator deploys the model(s) they want to use for each profile (using the existing ML Commons register/deploy APIs — local or remote).
2. Operator maps each profile to a `model_id` via a dynamic cluster setting:
   ```
   plugins.neural_search.model_selection.model_id.sparse.english: <model-id>
   plugins.neural_search.model_selection.model_id.dense.english: <model-id>
   plugins.neural_search.model_selection.model_id.sparse.multilingual: <model-id>
   plugins.neural_search.model_selection.model_id.dense.multilingual: <model-id>
   ```
   This is an **affix** setting (`plugins.neural_search.model_selection.model_id.<model_type>.<language_option>`), so the mapping stays flexible for additional dimensions in the future. The value is read from node settings (`opensearch.yml`) merged with the persistent/transient cluster settings.
3. Customer creates/updates an index with a `model_selection` on the semantic field; the resolved `model_id` is written into the mapping and returned in the GET mapping response.

### Behavior
- **No shipped defaults (yet):** if no `model_id` is configured for the requested profile, index creation fails fast with a clear error telling the operator which setting to set. (Predictable defaults can be shipped once ml-commons supports stable/custom model ids — [ml-commons#1217](https://github.com/opensearch-project/ml-commons/issues/1217).)
- **Model existence + type check:** the resolved model must exist and its type (dense/sparse) must match the requested `model_type`, otherwise the request is rejected.
- **`model_id` + `model_selection` together:** accepted only if the supplied `model_id` equals the one resolved from `model_selection` (strict match); this keeps the GET→PUT mapping roundtrip working. A mismatch is rejected.
- **No cache, no system-index search, no auto-deploy** — resolution is a cheap, synchronous cluster-setting lookup.

## Key changes
- **`ModelSelection`** DTO — parses/validates the nested `model_selection` object (enum validation, defaults, serialization).
- **`SemanticModelResolver`** interface + **`DefaultSemanticModelResolver`** — synchronous resolution of `model_selection` → `model_id` from the cluster setting (no ml-commons call).
- **`SemanticModelSelectionSettingsAccessor`** + the `SEMANTIC_MODEL_SELECTION_MODEL_ID` affix setting — reads the current value from node + cluster settings.
- **`SemanticFieldMapper`** — replaces the top-level `language_option`/`model_type` params with a single `model_selection` object parameter (`model_id` unchanged).
- **`SemanticMappingTransformer`** — the single enforcement point: iteratively resolves `model_selection` fields, enforces the strict `model_id` match, and validates the model type where the model is loaded (single fetch per field). Documented as the enforcement point.
- **`NeuralSearch`** — wires the resolver + registers the setting.

## Test plan
- [x] Unit — `DefaultSemanticModelResolverTests`: missing-setting error, sparse/dense resolution.
- [x] Unit — `SemanticFieldMapperModelSelectionTests`: parse/serialize `model_selection`, defaults, invalid enum values, unsupported sub-key, coexistence with `model_id`, param count.
- [x] Unit — `SemanticMappingTransformerTests`: no-resolver failure, model_id/model_selection match → success, mismatch → fail, non-string model_id → fail, invalid `semantic_info_field_name` → fail, model-type mismatch → fail.
- [x] Integration — `SemanticFieldModelSelectionIT` (8 cases): no config → error; sparse resolve; dense resolve; model_id+model_selection match → success; mismatched model_id → error; model-type mismatch → error; non-string model_id → error; invalid `semantic_info_field_name` → error.
- [x] Existing tests pass (`SemanticFieldMapperTests`, `SemanticMappingTransformerTests`, `NeuralSearchTests`).
- [x] `./gradlew compileJava compileTestJava spotlessJavaCheck` passes.

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
+++ b/src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
@@ -12,11 +12,17 @@ import org.opensearch.index.mapper.MappingTransformer;
 import org.opensearch.ml.common.FunctionName;
 import org.opensearch.ml.common.MLModel;
 import org.opensearch.neuralsearch.constants.SemanticFieldConstants;
+import org.opensearch.neuralsearch.mapper.dto.ModelSelection;
 import org.opensearch.neuralsearch.ml.MLCommonsClientAccessor;
+import org.opensearch.neuralsearch.ml.resolver.SemanticModelResolver;
+import org.opensearch.neuralsearch.util.SemanticMLModelUtils;
 import lombok.NonNull;
+import lombok.Setter;
+import lombok.extern.log4j.Log4j2;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -24,6 +30,8 @@ import java.util.Set;
 import java.util.stream.Collectors;
 
 import static org.opensearch.neuralsearch.constants.MappingConstants.PROPERTIES;
+import static org.opensearch.neuralsearch.constants.SemanticFieldConstants.MODEL_ID;
+import static org.opensearch.neuralsearch.constants.SemanticFieldConstants.MODEL_SELECTION;
 import static org.opensearch.neuralsearch.constants.SemanticFieldConstants.SEMANTIC_INFO_FIELD_NAME;
 import static org.opensearch.neuralsearch.constants.SemanticFieldConstants.SPARSE_ENCODING_CONFIG;
 import static org.opensearch.neuralsearch.util.SemanticMappingUtils.collectSemanticField;
@@ -38,7 +46,28 @@ import static org.opensearch.neuralsearch.util.SemanticMappingUtils.validateSema
 /**
  * SemanticMappingTransformer transforms the index mapping for the semantic field to auto add the semantic info fields
  * based on the ML model id defined in the semantic field.
+ * <p>
+ * This transformer is the single enforcement point for the {@code model_id}/{@code model_selection} rules. The
+ * {@link org.opensearch.neuralsearch.mapper.SemanticFieldMapper} TypeParser intentionally accepts both together (so a
+ * GET mapping response, which contains the resolved {@code model_id} alongside {@code model_selection}, can be reapplied
+ * via PUT). It resolves {@code model_selection} to a {@code model_id}, and, when an explicit {@code model_id} is also
+ * supplied, verifies the two point to the same model. The mapper cannot perform the check itself as it has no access to
+ * the cluster-setting resolution at parse time.
+ * <p>
+ * The mapping transformers are invoked by the core {@code MappingTransformerRegistry} from all mapping-defining
+ * transport actions, so this transformer runs on the raw mapping for every path that can introduce or update a semantic
+ * field:
+ * <ul>
+ *   <li>{@code TransportCreateIndexAction} (create index)</li>
+ *   <li>{@code TransportPutMappingAction} (update mapping)</li>
+ *   <li>{@code TransportPutIndexTemplateAction} (legacy index template)</li>
+ *   <li>{@code TransportPutComponentTemplateAction} (component template)</li>
+ *   <li>{@code TransportPutComposableIndexTemplateAction} (composable index template)</li>
+ * </ul>
+ * Note that for the composable index template path the transformed mapping is persisted into the stored template, so a
+ * {@code model_selection} in a template is resolved to a concrete {@code model_id} at template-creation time.
  */
+@Log4j2
 public class SemanticMappingTransformer implements MappingTransformer {
     public final static Set<String> SUPPORTED_MODEL_ALGORITHMS = Set.of(
         FunctionName.TEXT_EMBEDDING.name(),
@@ -54,6 +83,9 @@ public class SemanticMappingTransformer implements MappingTransformer {
 
     private final MLCommonsClientAccessor mlClientAccessor;
     private final NamedXContentRegistry xContentRegistry;
+
+    @Setter
+    private volatile SemanticModelResolver modelResolver;
 
     public SemanticMappingTransformer(final MLCommonsClientAccessor mlClientAccessor, final NamedXContentRegistry xContentRegistry) {
         this.mlClientAccessor = mlClientAccessor;
@@ -142,13 +174,101 @@ public class SemanticMappingTransformer implements MappingTransformer {
 
             collectSemanticField(properties, semanticFieldPathToConfigMap);
 
-            validateSemanticFields(semanticFieldPathToConfigMap);
+            // Split fields into those that carry a model_selection (need model_id resolution from the cluster settings)
+            // and those that rely on an explicit model_id (existing path). Fields with neither model_selection nor
+            // model_id are kept in fieldsWithModelId so the existing validation surfaces the missing model_id error.
+            // Use a LinkedHashMap for the fields to resolve so resolution order is deterministic.
+            final Map<String, Map<String, Object>> fieldsToResolve = new LinkedHashMap<>();
+            final Map<String, Map<String, Object>> fieldsWithModelId = new HashMap<>();
 
-            fetchModelAndModifyMapping(semanticFieldPathToConfigMap, properties, listener);
+            for (Map.Entry<String, Map<String, Object>> entry : semanticFieldPathToConfigMap.entrySet()) {
+                final Map<String, Object> config = entry.getValue();
+                if (config.get(MODEL_SELECTION) != null) {
+                    fieldsToResolve.put(entry.getKey(), config);
+                } else {
+                    fieldsWithModelId.put(entry.getKey(), config);
+                }
+            }
+
+            validateSemanticFields(fieldsWithModelId);
+
+            // Resolve model_selection fields (a cheap, synchronous cluster-setting lookup) and record the requested
+            // ModelSelection per field so the model type can be verified later, when the model is loaded.
+            final Map<String, ModelSelection> modelSelectionByPath = new HashMap<>();
+            if (fieldsToResolve.isEmpty() == false) {
+                if (modelResolver == null) {
+                    throw new IllegalStateException(
+                        "Cannot resolve the model_selection because the SemanticModelResolver is not configured."
+                    );
+                }
+                for (final Map.Entry<String, Map<String, Object>> entry : fieldsToResolve.entrySet()) {
+                    resolveModelSelectionField(entry.getKey(), entry.getValue(), fieldsWithModelId, modelSelectionByPath);
+                }
+            }
+
+            if (fieldsWithModelId.isEmpty()) {
+                listener.onResponse(null);
+            } else {
+                fetchModelAndModifyMapping(fieldsWithModelId, modelSelectionByPath, properties, listener);
+            }
         } catch (Exception e) {
             listener.onFailure(e);
         }
 
+    }
+
+    private void resolveModelSelectionField(
+        @NonNull final String fieldPath,
+        @NonNull final Map<String, Object> fieldConfig,
+        @NonNull final Map<String, Map<String, Object>> fieldsWithModelId,
+        @NonNull final Map<String, ModelSelection> modelSelectionByPath
+    ) {
+        final ModelSelection modelSelection = new ModelSelection(MODEL_SELECTION, fieldConfig.get(MODEL_SELECTION));
+        final String resolvedModelId = modelResolver.resolve(modelSelection);
+        final Object suppliedModelId = fieldConfig.get(MODEL_ID);
+
+        // Log the resolution so drift is diagnosable. On an update that re-applies a model_selection field, this line
+        // records which model the field resolved to; if the resolved id differs from what an existing field was bound
+        // to, comparing successive log lines surfaces that the field moved off its previous model.
+        log.info(
+            "Resolved model_selection [language_option={}, model_type={}] to model_id [{}] for semantic field at [{}].",
+            modelSelection.getLanguageOption(),
+            modelSelection.getModelType(),
+            resolvedModelId,
+            fieldPath
+        );
+
+        if (suppliedModelId != null) {
+            if (suppliedModelId instanceof String == false) {
+                throw new IllegalArgumentException(
+                    String.format(Locale.ROOT, "The model_id for the semantic field at %s must be a string.", fieldPath)
+                );
+            }
+            // When both model_id and model_selection are provided, they must point to the same model.
+            if (resolvedModelId.equals(suppliedModelId) == false) {
+                throw new IllegalArgumentException(
+                    String.format(
+                        Locale.ROOT,
+                        "The provided model_id [%s] for the semantic field at %s does not match the model_id [%s] resolved "
+                            + "from model_selection [language_option=%s, model_type=%s]. Provide the matching model_id or omit it.",
+                        suppliedModelId,
+                        fieldPath,
+                        resolvedModelId,
+                        modelSelection.getLanguageOption(),
+                        modelSelection.getModelType()
+                    )
+                );
+            }
+        } else {
+            fieldConfig.put(MODEL_ID, resolvedModelId);
+        }
+
+        // Validate the resolved field consistently with the explicit model_id path (e.g. semantic_info_field_name
+        // format). Model selection fields skip the upfront validateSemanticFields call, so validate them here.
+        validateSemanticFields(Map.of(fieldPath, fieldConfig));
+
+        fieldsWithModelId.put(fieldPath, fieldConfig);
+        modelSelectionByPath.put(fieldPath, modelSelection);
     }
 
     private void validateSemanticFields(@NonNull final Map<String, Map<String, Object>> semanticFieldPathToConfigMap) {
@@ -172,13 +292,14 @@ public class SemanticMappingTransformer implements MappingTransformer {
 
     private void fetchModelAndModifyMapping(
         @NonNull final Map<String, Map<String, Object>> semanticFieldPathToConfigMap,
+        @NonNull final Map<String, ModelSelection> modelSelectionByPath,
         @NonNull final Map<String, Object> mappings,
         @NonNull final ActionListener<Void> listener
     ) {
         final Map<String, List<String>> modelIdToFieldPathMap = extractModelIdToFieldPathMap(semanticFieldPathToConfigMap);
 
         mlClientAccessor.getModels(modelIdToFieldPathMap.keySet(), modelIdToConfigMap -> {
-            modifyMappings(modelIdToConfigMap, mappings, modelIdToFieldPathMap, semanticFieldPathToConfigMap);
+            modifyMappings(modelIdToConfigMap, mappings, modelIdToFieldPathMap, semanticFieldPathToConfigMap, modelSelectionByPath);
             listener.onResponse(null);
         }, listener::onFailure);
     }
@@ -187,7 +308,8 @@ public class SemanticMappingTransformer implements MappingTransformer {
         @NonNull final Map<String, MLModel> modelIdToConfigMap,
         @NonNull final Map<String, Object> mappings,
         @NonNull final Map<String, List<String>> modelIdToFieldPathMap,
-        @NonNull final Map<String, Map<String, Object>> semanticFieldPathToConfigMap
+        @NonNull final Map<String, Map<String, Object>> semanticFieldPathToConfigMap,
+        @NonNull final Map<String, ModelSelection> modelSelectionByPath
     ) {
         for (String modelId : modelIdToFieldPathMap.keySet()) {
             final MLModel mlModel = modelIdToConfigMap.get(modelId);
@@ -195,6 +317,12 @@ public class SemanticMappingTransformer implements MappingTransformer {
             for (String fieldPath : fieldPathList) {
                 try {
                     final Map<String, Object> fieldConfig = semanticFieldPathToConfigMap.get(fieldPath);
+                    // For fields resolved from a model_selection, verify the resolved model's type matches the
+                    // requested model_type now that the model is loaded (single fetch, no separate resolve-time fetch).
+                    final ModelSelection modelSelection = modelSelectionByPath.get(fieldPath);
+                    if (modelSelection != null) {
+                        validateModelTypeMatches(modelSelection, mlModel, modelId);
+                    }
                     final Map<String, Object> semanticInfoConfig = createSemanticInfoField(mlModel, modelId, fieldConfig, fieldPath);
                     setSemanticInfoField(mappings, fieldPath, fieldConfig.get(SEMANTIC_INFO_FIELD_NAME), semanticInfoConfig);
                 } catch (IllegalArgumentException e) {
@@ -203,6 +331,25 @@ public class SemanticMappingTransformer implements MappingTransformer {
                     throw new RuntimeException(getModifyMappingErrorMessage(fieldPath, e.getMessage()), e);
                 }
             }
+        }
+    }
+
+    private void validateModelTypeMatches(
+        @NonNull final ModelSelection modelSelection,
+        @NonNull final MLModel mlModel,
+        @NonNull final String modelId
+    ) {
+        final boolean modelIsDense = SemanticMLModelUtils.isDenseModel(SemanticMLModelUtils.getModelType(mlModel));
+        if (modelIsDense != modelSelection.isDense()) {
+            throw new IllegalArgumentException(
+                String.format(
+                    Locale.ROOT,
+                    "The model [%s] is a %s model, which does not match the requested model_type [%s].",
+                    modelId,
+                    modelIsDense ? ModelSelection.DENSE : ModelSelection.SPARSE,
+                    modelSelection.getModelType()
+                )
+            );
         }
     }
 
--- a/src/main/java/org/opensearch/neuralsearch/ml/resolver/DefaultSemanticModelResolver.java
+++ b/src/main/java/org/opensearch/neuralsearch/ml/resolver/DefaultSemanticModelResolver.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.ml.resolver;
+
+import lombok.NonNull;
+import lombok.extern.log4j.Log4j2;
+import org.opensearch.neuralsearch.mapper.dto.ModelSelection;
+import org.opensearch.neuralsearch.settings.SemanticModelSelectionSettingsAccessor;
+
+import java.util.Locale;
+
+import static org.opensearch.neuralsearch.settings.NeuralSearchSettings.SEMANTIC_MODEL_SELECTION_MODEL_ID_PREFIX;
+
+/**
+ * Default {@link SemanticModelResolver} implementation. It resolves a {@link ModelSelection} to a model_id by reading
+ * the operator configured {@code plugins.neural_search.model_selection.model_id.<model_type>.<language_option>} cluster
+ * setting. It does not register, deploy, or fetch any model; the operator is expected to deploy the model and configure
+ * the setting. Verification that the resolved model exists and its type matches the requested model type happens later,
+ * where the model is loaded to build the semantic info field.
+ */
+@Log4j2
+public class DefaultSemanticModelResolver implements SemanticModelResolver {
+
+    private final SemanticModelSelectionSettingsAccessor settingsAccessor;
+
+    public DefaultSemanticModelResolver(@NonNull final SemanticModelSelectionSettingsAccessor settingsAccessor) {
+        this.settingsAccessor = settingsAccessor;
+    }
+
+    @Override
+    public String resolve(@NonNull final ModelSelection modelSelection) {
+        final String profileKey = modelSelection.getProfileKey();
+        final String modelId = settingsAccessor.getModelId(profileKey);
+
+        if (modelId == null) {
+            final String settingKey = SEMANTIC_MODEL_SELECTION_MODEL_ID_PREFIX + profileKey;
+            throw new IllegalArgumentException(
+                String.format(
+                    Locale.ROOT,
+                    "No model_id is configured for the model_selection [language_option=%s, model_type=%s]. "
+                        + "Deploy a %s model and set the cluster setting [%s] to its model_id, then retry.",
+                    modelSelection.getLanguageOption(),
+                    modelSelection.getModelType(),
+                    modelSelection.getModelType(),
+                    settingKey
+                )
+            );
+        }
+
+        return modelId;
+    }
+}
--- a/src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
+++ b/src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
@@ -49,6 +49,9 @@ import org.opensearch.index.mapper.Mapper;
 import org.opensearch.index.mapper.MappingTransformer;
 import org.opensearch.neuralsearch.mapper.SemanticFieldMapper;
 import org.opensearch.neuralsearch.mappingtransformer.SemanticMappingTransformer;
+import org.opensearch.neuralsearch.ml.resolver.DefaultSemanticModelResolver;
+import org.opensearch.neuralsearch.ml.resolver.SemanticModelResolver;
+import org.opensearch.neuralsearch.settings.SemanticModelSelectionSettingsAccessor;
 import org.opensearch.neuralsearch.processor.factory.SemanticFieldProcessorFactory;
 import org.opensearch.plugins.MapperPlugin;
 import org.opensearch.search.query.QueryCollectorContextSpecFactory;
@@ -192,6 +195,8 @@ public class NeuralSearch extends Plugin
     private PipelineServiceUtil pipelineServiceUtil;
     private InfoStatsManager infoStatsManager;
     private ClusterService clusterService;
+    private SemanticMappingTransformer semanticMappingTransformer;
+    private SemanticModelResolver modelResolver;
     private final SemanticHighlighter semanticHighlighter;
     private final ScoreNormalizationFactory scoreNormalizationFactory = new ScoreNormalizationFactory();
     private final ScoreCombinationFactory scoreCombinationFactory = new ScoreCombinationFactory();
@@ -241,6 +246,14 @@ public class NeuralSearch extends Plugin
 
         // Initialize the semantic highlighter
         this.semanticHighlighter.initialize(semanticHighlighterEngine);
+
+        // Create model resolver for semantic field model_selection support. It resolves the model_id from the
+        // plugins.neural_search.model_selection.model_id.* cluster settings.
+        SemanticModelSelectionSettingsAccessor modelSelectionSettingsAccessor = new SemanticModelSelectionSettingsAccessor(clusterService);
+        modelResolver = new DefaultSemanticModelResolver(modelSelectionSettingsAccessor);
+        if (semanticMappingTransformer != null) {
+            semanticMappingTransformer.setModelResolver(modelResolver);
+        }
 
         // Create and provide the Hybrid query converter for gRPC transport
         HybridQueryBuilderProtoConverter hybridQueryConverter = new HybridQueryBuilderProtoConverter();
@@ -366,7 +379,8 @@ public class NeuralSearch extends Plugin
             SparseSettings.IS_SPARSE_INDEX_SETTING,
             NeuralSearchSettings.SPARSE_ALGO_PARAM_INDEX_THREAD_QTY_SETTING,
             NEURAL_CIRCUIT_BREAKER_LIMIT,
-            NEURAL_CIRCUIT_BREAKER_OVERHEAD
+            NEURAL_CIRCUIT_BREAKER_OVERHEAD,
+            NeuralSearchSettings.SEMANTIC_MODEL_SELECTION_MODEL_ID
         );
     }
 
@@ -466,7 +480,11 @@ public class NeuralSearch extends Plugin
 
     @Override
     public List<MappingTransformer> getMappingTransformers() {
-        return List.of(new SemanticMappingTransformer(clientAccessor, xContentRegistry));
+        semanticMappingTransformer = new SemanticMappingTransformer(clientAccessor, xContentRegistry);
+        if (modelResolver != null) {
+            semanticMappingTransformer.setModelResolver(modelResolver);
+        }
+        return List.of(semanticMappingTransformer);
     }
 
     @Override
--- a/src/main/java/org/opensearch/neuralsearch/settings/SemanticModelSelectionSettingsAccessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/settings/SemanticModelSelectionSettingsAccessor.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.settings;
+
+import lombok.NonNull;
+import org.opensearch.cluster.service.ClusterService;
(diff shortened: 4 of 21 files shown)


src/main/java/org/opensearch/neuralsearch/ml/resolver/PretrainedSemanticModelResolver.java
```diff
@@ -0,0 +1,229 @@
...
+        mlClient.getTask(taskId, ActionListener.wrap(mlTask -> {
+            String modelId = mlTask.getModelId();
+            if (modelId != null && !modelId.isEmpty()) {
+                modelCache.put(cacheKey, modelId);
+                listener.onResponse(modelId);
+            } else {
+                // Task still in progress, schedule a retry
+                client.threadPool().generic().execute(() -> {
+                    try {
+                        Thread.sleep(POLL_INTERVAL_MS);
```
martin-gaievski, line 211: This holds a `generic` threadpool worker for the full 2s sleep, and this pool is shared across many cluster operations; a burst of CreateIndex requests (or several managed fields) can exhaust it and degrade cluster-wide responsiveness. 

Use a non-blocking scheduled retry and fail fast on a FAILED task instead of polling to timeout

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -142,15 +154,120 @@ public void transform(final Map<String, Object> mapping, final TransformContext
...
     }
 
+    private void resolveManagedFields(
+        @NonNull final Map<String, Map<String, Object>> managedFields,
+        @NonNull final Map<String, Map<String, Object>> fieldsWithModelId,
+        @NonNull final Map<String, Object> properties,
+        @NonNull final ActionListener<Void> listener
+    ) {
+        if (modelResolver == null) {
+            log.warn("SemanticModelResolver is not set, skipping managed semantic fields");
```
martin-gaievski, line 212: looks like we're completing successfully while leaving managed fields with no `model_id` and no `semantic_info`, so index is created with a non-functional semantic field, silently. And failure only surfaces later at ingest/search.
Shouldn't we fail fast instead: 
  ```java
  if (modelResolver == null) {
      listener.onFailure(new IllegalStateException(
          "Semantic model resolver is not initialized; cannot resolve language_option/model_type. "));
      return;
  }
  ```

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
```diff
@@ -242,6 +246,13 @@ public Collection<Object> createComponents(
         // Initialize the semantic highlighter
         this.semanticHighlighter.initialize(semanticHighlighterEngine);
 
+        // Create model resolver for semantic field language_option/model_type support
+        MachineLearningNodeClient mlNodeClient = new MachineLearningNodeClient(client);
```
martin-gaievski, line 250: We do have facade to abstract ml low level operations `MLCommonsClientAccessor`, I suggest you move ml client calls there and create new methods, that later will be called from `PretrainedSemanticModelResolver`. Otherwise will be multiplying classes that talk with low level ml-commons constructs which makes long term maintenance harder

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -142,15 +154,120 @@ public void transform(final Map<String, Object> mapping, final TransformContext
...
+
+        modelResolver.resolve(languageOption, modelTypeVal, ActionListener.wrap(resolvedModelId -> {
+            // Set model_id on the field config
+            fieldConfig.put(MODEL_ID, resolvedModelId);
+
+            // For DENSE models, set dense_embedding_config with method.engine=lucene to avoid faiss native lib requirement
+            if ("DENSE".equalsIgnoreCase(modelTypeVal) && !fieldConfig.containsKey(DENSE_EMBEDDING_CONFIG)) {
+                Map<String, Object> denseConfig = new HashMap<>();
+                Map<String, Object> method = new HashMap<>();
+                method.put("engine", "lucene");
```
martin-gaievski, line 258: can you move engine assignment/resolution to resolver/field-config layer rather than the transformer

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -142,15 +154,120 @@ public void transform(final Map<String, Object> mapping, final TransformContext
...
+                denseConfig.put("method", method);
+                fieldConfig.put(DENSE_EMBEDDING_CONFIG, denseConfig);
+            }
+
+            // Move this field into the fieldsWithModelId map for subsequent processing
+            fieldsWithModelId.put(fieldPath, fieldConfig);
+
+            // Continue resolving next field
+            resolveNextManagedField(iterator, fieldsWithModelId, properties, listener);
```
heemin32, line 267: We need some safeguard to avoid stack overflow with this recursive call or we need to change it to iteratively.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
heemin32, line 258: We should not force it to use lucene. CX should be able to choose faiss if they want.

src/main/java/org/opensearch/neuralsearch/ml/resolver/PretrainedSemanticModelResolver.java
```diff
@@ -0,0 +1,229 @@
...
+        searchForExistingModel(spec, cacheKey, listener);
+    }
+
+    private void searchForExistingModel(ModelSpec spec, String cacheKey, ActionListener<String> listener) {
+        BoolQueryBuilder query = QueryBuilders.boolQuery()
+            .must(QueryBuilders.termQuery(MODEL_NAME_KEYWORD_FIELD, spec.name()))
+            .must(QueryBuilders.termQuery(MODEL_ALGORITHM_FIELD, spec.algorithm().name()));
+
+        SearchSourceBuilder sourceBuilder = new SearchSourceBuilder().query(query).size(1);
+        SearchRequest searchRequest = new SearchRequest(ML_MODEL_INDEX).source(sourceBuilder);
```
heemin32, line 126: Isn't there ml common api to search model? Accessing system index directly doesn't seems as an ideal approach?

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -55,6 +64,9 @@ public class SemanticMappingTransformer implements MappingTransformer {
     private final MLCommonsClientAccessor mlClientAccessor;
     private final NamedXContentRegistry xContentRegistry;
 
+    @Setter
+    private volatile SemanticModelResolver modelResolver;
```
heemin32, line 68: Why volatile?

yizheliu-amazon: `modelResolver` is written once via `setModelResolver()` during plugin init (from `createComponents()` / `getMappingTransformers()`, on the node startup thread) and read later in `transform()`, which runs on request-handling threads. Without `volatile` there is no happens-before edge between that write and those reads, so a request thread could observe a stale `null`. `volatile` guarantees the published reference is visible once set.

It can't be `final` because the resolver is injected via a setter after the transformer is constructed — the transformer is created in `getMappingTransformers()` while the resolver is created in `createComponents()` (two-phase wiring). The referenced resolver is itself effectively immutable (its dependencies are `final`), so a single volatile reference is sufficient and cheaper than locking.

bzhangam:
The current PR description describes behavior that isn't in the code (auto-register/deploy, pretrained-model table, in-memory cache, `PretrainedSemanticModelResolver`, mutual-exclusion). Suggest replacing it with the description below so it matches the implementation.

src/main/java/org/opensearch/neuralsearch/ml/resolver/DefaultSemanticModelResolver.java
```diff
@@ -0,0 +1,103 @@
...
+                        modelSelection.getModelType(),
+                        settingKey
+                    )
+                )
+            );
+            return;
+        }
+
+        // Verify the configured model exists and its type matches the requested model type.
+        mlClientAccessor.getModel(modelId, ActionListener.wrap(mlModel -> {
```
bzhangam: Each `model_selection` field ends up fetching the model from ml-commons twice:

1. Here in `DefaultSemanticModelResolver.resolve()` we call `mlClientAccessor.getModel(modelId, ...)` just to check the resolved model's dense/sparse type matches the requested `model_type`.
2. The resolved field is then handed to `fetchModelAndModifyMapping()`, which calls
   `getModels(...)` again for the same `model_id` to build the semantic info config.

That's two round-trips per resolved field (fields with an explicit `model_id` only fetch once),
and the resolution is done serially across fields, so the extra calls add up.

Suggest splitting resolution from model-type validation:

- **In the resolver:** only resolve `model_id` from the cluster setting and do cheap, local
  validation (setting present, non-empty). Drop the `getModel` call entirely — no ml-commons
  dependency in the resolve path. The resolver can then return synchronously, which also removes the serial round-trip cost.

- **In `fetchModelAndModifyMapping` (the existing fetch path):** after the `MLModel` is fetched,
  add the new `model_type` vs `model_selection` check there — we already have the `MLModel` in
  hand, and `SemanticMLModelUtils.getModelType(...)` / `isDenseModel(...)` are already available.
  To do this we need to carry the requested `ModelSelection` (or just the expected dense/sparse
  flag) alongside the field config into that path so the validation knows what was requested.

Net effect: one fetch per field instead of two, no ml-commons call during resolution, and the
type-match validation lives next to the code that already loads the model. The user-facing error
message on mismatch can stay the same.

src/main/java/org/opensearch/neuralsearch/settings/SemanticModelSelectionSettingsAccessor.java
```diff
@@ -0,0 +1,49 @@
...
+     * @param profileKey profile key of the form {@code <model_type>.<language_option>} (lower cased), e.g. {@code sparse.english}
+     * @return the configured model id, or null when the operator has not configured this profile
+     */
+    public String getModelId(@NonNull final String profileKey) {
+        final String fullKey = SEMANTIC_MODEL_SELECTION_MODEL_ID_PREFIX + profileKey;
+        // Merge node settings (opensearch.yml), persistent, and transient cluster settings, in increasing order of
+        // precedence, so both static (opensearch.yml) and dynamically updated values are honored.
+        final Settings settings = Settings.builder()
+            .put(clusterService.getSettings())
+            .put(clusterService.state().getMetadata().persistentSettings())
```
bzhangam: Why not use `clusterService.state().getMetadata().settings()` which already merge the persistent and transient?

yizheliu-amazon: Done — switched to `clusterService.state().getMetadata().settings()`, which already merges the persistent and transient cluster settings. I kept `clusterService.getSettings()` as a base layer so statically-configured `opensearch.yml` values are honored too (`getMetadata().settings()` only covers cluster-state settings, not node settings).

src/main/java/org/opensearch/neuralsearch/ml/resolver/DefaultSemanticModelResolver.java
yizheliu-amazon: Refactored as suggested. The resolver now only reads the `model_id` from the cluster setting — a synchronous, local lookup with no ml-commons call. The `model_type` vs `model_selection` check moved into `fetchModelAndModifyMapping`, where the `MLModel` is already loaded, so each field is fetched once. The requested `ModelSelection` is carried per field into that path; the mismatch error message is unchanged.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
yizheliu-amazon: Done — when a `model_selection` field is present but the resolver is not configured, we now fail fast with an `IllegalStateException` instead of silently creating a non-functional semantic field.

src/main/java/org/opensearch/neuralsearch/plugin/NeuralSearch.java
yizheliu-amazon: The design has changed so this is no longer a concern: the resolver no longer talks to ml-commons at all — it is now a cluster-setting lookup (`plugins.neural_search.model_selection.model_id.*`). The single model fetch still goes through the existing `MLCommonsClientAccessor` facade in the transformer, so no new class talks to low-level ml-commons constructs.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
yizheliu-amazon: Removed entirely — the transformer no longer assigns/forces an engine. The resolved model just flows through the existing path, and the customer-provided `dense_embedding_config` is respected.

yizheliu-amazon: Removed the forced `lucene` engine. Customers can use faiss (or configure the engine via `dense_embedding_config`).

src/main/java/org/opensearch/neuralsearch/ml/resolver/PretrainedSemanticModelResolver.java
yizheliu-amazon: Approach changed — we no longer search the model system index at all. Resolution is now an operator-configured cluster-setting lookup (`plugins.neural_search.model_selection.model_id.<model_type>.<language_option>`), so there is no direct `.plugins-ml-model` access.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
yizheliu-amazon: Converted to an iterative loop. Resolution is now a synchronous cluster-setting lookup, so the per-field recursion is gone and there is no per-field stack growth.

yizheliu-amazon:
Thanks for the automated review. Notes on the three focus areas (as of the latest commit):

**1. Initialization Race** — Safe by design. `getMappingTransformers()` and `createComponents()` both run during node bootstrap, before the node serves client requests, and index recovery at startup does not run the mapping transformer (recovered mappings are already transformed). The resolver is wired in both hooks, so whichever runs second completes the wiring regardless of ordering — there's no path where a `model_selection` transform executes with a null resolver.

**2. Inconsistent Mismatch Detection** — This is intentional and now documented on `SemanticMappingTransformer`. The transformer is the single enforcement point: it runs on CreateIndex and PutMapping (the only paths that add/update a semantic field), resolves `model_selection` to a `model_id`, and verifies a match when an explicit `model_id` is also supplied. The `TypeParser` accepts both together on purpose so a GET-mapping response (which contains the resolved `model_id` alongside `model_selection`) can be reapplied via PUT. The mapper can't perform the check itself since it has no access to cluster-setting resolution at parse time; and already-stored mappings were validated at write time.

**3. Possibly Unrelated Change** — `NEURAL_CIRCUIT_BREAKER_OVERHEAD` is still registered in `getSettings()`; it only appears in the diff because the new `SEMANTIC_MODEL_SELECTION_MODEL_ID` line was added right after it — nothing was dropped. The one-line change in `NeuralSparseTwoPhaseProcessorTests` is required by `spotlessJavaCheck` (the prior form fails the current formatter), so it's a mandated reformat rather than scope creep.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -142,15 +163,92 @@ public void transform(final Map<String, Object> mapping, final TransformContext
...
     }
 
+    private void resolveModelSelectionField(
+        @NonNull final String fieldPath,
+        @NonNull final Map<String, Object> fieldConfig,
+        @NonNull final Map<String, Map<String, Object>> fieldsWithModelId,
+        @NonNull final Map<String, ModelSelection> modelSelectionByPath
+    ) {
+        final ModelSelection modelSelection = new ModelSelection(MODEL_SELECTION, fieldConfig.get(MODEL_SELECTION));
+        final String resolvedModelId = modelResolver.resolve(modelSelection);
```
martin-gaievski, line 227: that can be a source of split ground truth, we resolve on every transform before looking into supplied `model_id`, and then checking in they are equal (line 266, ` if (resolvedModelId.equals(suppliedModelId) == false) {`). transformer also stores resolved id into the stored mapping (241) while `model_selection`
stays in the mapping and is serialised back out => artifact the plugin produces contains both keys.

Example of such scenario:
1. `PUT _cluster/settings {"persistent":{"plugins.neural_search.model_selection.model_id.sparse.english":"M1"}}`
2. `PUT /docs {"mappings":{"properties":{"content":{"type":"semantic","model_selection":{}}}}}` => stored mapping is `{model_id: M1, model_selection:{ENGLISH,SPARSE}}`.
3. Operator rotates the profile to `M2`

After that any call that re-applies what is reads back, e.g. `PUT /docs/_mapping` will get code 400

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -142,15 +163,92 @@ public void transform(final Map<String, Object> mapping, final TransformContext
...
+                        suppliedModelId,
+                        fieldPath,
+                        resolvedModelId,
+                        modelSelection.getLanguageOption(),
+                        modelSelection.getModelType()
+                    )
+                );
+            }
+        } else {
+            fieldConfig.put(MODEL_ID, resolvedModelId);
```
martin-gaievski, line 263: IMO we need to make resolved binding effectively immutable for an existing field. on update path
reject a resolution that would change the stored `model_id` ("the model of an existing semantic field cannot be changed; reindex instead") and at minimum log at INFO on every resolution, so drift is diagnosable after it happens (`field path → profile key →resolved model_id`)

Scenario that is possible with current code: `docs` created while the profile pointed at sparse `M1`, 50M documents ingested (token weights from `M1` in `content_semantic_info.chunks.embedding`). Operator rotates to sparse `M2`. Someone re-applies `PUT /docs/_mapping {"properties":{"content":{"type":"semantic","model_selection":{"model_type":"SPARSE"}}}}`, or just adds an unrelated updateable setting to that field. Regenerated `semantic_info` block is byte-identical (`SemanticInfoConfigBuilder` embeds no model identity), so the merge returns 200, and field is now bound to `M2`. New/updated documents are embedded by `M2` and queries are inferred with `M2` while 50M existing vectors came from `M1`. Nothing in the response indicates that, and nothing is logged.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -38,7 +46,17 @@
...
  * based on the ML model id defined in the semantic field.
+ * <p>
+ * This transformer is the single enforcement point for the {@code model_id}/{@code model_selection} rules. The
+ * {@link org.opensearch.neuralsearch.mapper.SemanticFieldMapper} TypeParser intentionally accepts both together (so a
+ * GET mapping response, which contains the resolved {@code model_id} alongside {@code model_selection}, can be reapplied
+ * via PUT). The transformer runs on the raw mapping for CreateIndex and PutMapping - the only paths that add or update a
+ * semantic field - where it resolves {@code model_selection} to a {@code model_id}, and, when an explicit {@code model_id}
+ * is also supplied, verifies the two point to the same model. Because a semantic field can only be introduced through
+ * these paths, this is a sufficient enforcement point; the mapper cannot perform the check itself as it has no access to
+ * the cluster-setting resolution at parse time.
```
martin-gaievski, line 55: this javadoc is not completely accurate, `applyTransformers` is also called from `TransportPutIndexTemplateAction`,
`TransportPutComponentTemplateAction` and `TransportPutComposableIndexTemplateAction`, and
composable-template action persists transformed body. Doc should name all five transport actions

martin-gaievski (commented):
In addition to my code level comments:
- no new test/missing coverage for rotation, GET->PUT re-apply, `PUT _mapping`, templates, or a mapping merge of the new parameter, please add tests

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
bzhangam: Agreed. I think the intended contract should be: **once a `model_selection`-backed field is created, both `model_selection` and its resolved `model_id` are frozen.** Changing the model is a reindex, not a mapping update.

To get there I believe we need changes in **two places**:

1. **Transformer (required):** make it update-aware and resolve `model_selection` only for a *new* field. For a field that already exists in the committed mapping, do **not** re-resolve — carry forward the stored `model_id`. This is the load-bearing fix; the mapper can't do it because re-resolution happens before the mapper runs. This needs the transformer to see the committed mapping / target index (per-field new-vs-existing, since one PUT can add one field while leaving another untouched) —  `TransformContext` doesn't exposes that info today, but we can leverage it to access the info like target index and then access the existing index mapping.

2. **Field mapper (recommended backstop):**
   - mark `model_selection` `updateable=false` so changing the selection object is rejected declaratively;
   - in the existing `merge()` override, add a conditional check: if the existing field has a `model_selection`, reject an incoming `model_id` that differs from the stored one. (`merge()` already has both existing and incoming state, so this needs no core change and can guard the interim.)

Note the `model_id` immutability has to be *conditional* (only when `model_selection` is present) — we can't just flip the `updateable` flag, since plain `model_id` fields should keep their current update behavior.

Proposed per-field rule on the update path for an existing `model_selection` field:

| Incoming | Result |
|---|---|
| `model_selection` changed | reject (frozen) |
| `model_selection` same, `model_id` omitted | carry forward stored `model_id`, no re-resolve |
| `model_selection` same, `model_id` == stored | accept (GET→PUT roundtrip) |
| `model_selection` same, `model_id` != stored | reject (frozen) |

A couple of things worth confirming/covering with a test:
- create → drift the cluster setting → PUT the same mapping should keep the **original** `model_id` (not repoint);
- changing `model_selection` or `model_id` on an existing selection-backed field → rejected;

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
```diff
@@ -187,14 +286,21 @@ private void modifyMappings(
...
             final MLModel mlModel = modelIdToConfigMap.get(modelId);
             final List<String> fieldPathList = modelIdToFieldPathMap.get(modelId);
             for (String fieldPath : fieldPathList) {
                 try {
                     final Map<String, Object> fieldConfig = semanticFieldPathToConfigMap.get(fieldPath);
+                    // For fields resolved from a model_selection, verify the resolved model's type matches the
+                    // requested model_type now that the model is loaded (single fetch, no separate resolve-time fetch).
+                    final ModelSelection modelSelection = modelSelectionByPath.get(fieldPath);
+                    if (modelSelection != null) {
+                        validateModelTypeMatches(modelSelection, mlModel, modelId);
```
bzhangam, line 324: Wonder would it be better to do this validation when we update the cluster setting so that we fail fast?

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
yizheliu-amazon: Thanks @martin-gaievski — this is the intended behavior of the strict-match design, and I'd prefer to keep it. A few points on the scenario you described:

In your step 4, the reapply returns a 400 with:

> The provided model_id [M1] for the semantic field at content does not match the model_id [M2] resolved from model_selection [language_option=ENGLISH, model_type=SPARSE]. Provide the matching model_id or omit it.

That message states exactly what happened — the profile's model was rotated (the `sparse.english` cluster setting moved from M1 to M2) — and gives two concrete unblock actions: **omit `model_id`** (so it re-resolves to the current M2), or **set `model_id` to M2**. This is preferable to silently reapplying a mapping whose two keys now disagree, or silently swapping the field's model underneath the customer.

**Specifying both keys is not the recommended usage.** A semantic field should provide *either* `model_id` (customer-managed) *or* `model_selection` (system-resolved), not both. The "both present" state essentially only arises from a GET→PUT roundtrip of a system-resolved field, and it only breaks when an operator rotated the profile in between — a rare event. We'll call this out in the docs (provide one or the other).

**Existing indices are unaffected.** Rotating the cluster setting doesn't touch already-created indices — their resolved `model_id` is baked into the stored mapping and ingest/search keep working against it. Only an *explicit reapply* of the read-back mapping after a rotation hits the 400, by design.

So I'd keep strict-match: it validates that `model_id` and `model_selection` agree at write time, and in the rare post-rotation reapply case it fails with a clear, self-service fix rather than papering over a genuine inconsistency.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
yizheliu-amazon: Thanks @martin-gaievski @bzhangam — agreed the silent rebind is a real concern: existing vectors come from M1, new ones from M2, and the index holds a mix until re-embedded, with nothing surfaced.

I think for this PR we should keep `model_selection` consistent with the existing `model_id` behavior — **both remain updatable.** Two reasons:
- A `model_selection`-only immutability guard wouldn't actually fix it. The rebind manifests as a `model_id` change at merge time (the resolved id is written into `model_id`), and it's trivially bypassable by sending `model_id` directly. The only effective guard is at `model_id`, and changing that behavior is out of scope for this PR.
- Making one of the two mutable and the other immutable is a confusing contract. Consistent (both updatable) is clearest.

The model change is also recoverable and is a supported workflow: after changing the model, `update_by_query` (or reindex) re-embeds existing docs with the new model. So it's not necessarily permanent corruption — but I agree it's a footgun if done unknowingly.

Rather than ship a half-measure, I've opened #1963 to improve this holistically. Proposed direction there: a cluster/index setting that controls whether the model binding is updatable, **defaulting to `false` (immutable)** so safe-by-default is the default, checked at PutMapping, with an explicit opt-in for the rare intentional model change — giving backward compatibility without a hard breaking change. Plus resolution logging so any rebind is at least diagnosable.

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
yizheliu-amazon: Good catch — fixed. Updated the javadoc to name all five transport actions the core `MappingTransformerRegistry` invokes:

- `TransportCreateIndexAction`
- `TransportPutMappingAction`
- `TransportPutIndexTemplateAction`
- `TransportPutComponentTemplateAction`
- `TransportPutComposableIndexTemplateAction`

and I also called out that the composable index template path persists the transformed mapping, so a `model_selection` in a template is resolved to a concrete `model_id` at template-creation time.

yizheliu-amazon:
@martin-gaievski — added coverage for the paths you called out:

**Integration tests** (`SemanticFieldModelSelectionIT`):
- `testUpdateMapping_addModelSelectionField_thenResolveModelId` — adding a `model_selection` semantic field via `PUT _mapping` resolves the model_id.
- `testComposableTemplate_withModelSelection_thenResolveModelId` — a composable index template with a `model_selection` field resolves, and an index created from it carries the resolved model_id (covers the template transport-action path).
- `testReapplyMappingAfterProfileRotation_thenFail` — create with `model_selection`, rotate the profile's cluster setting, then reapply the read-back mapping → 400 mismatch (the rotation / GET→PUT-after-rotation case).

**Unit test** (`SemanticFieldMapperModelSelectionTests`):
- `testMerge_modelSelectionIsUpdatable` — mapping merge of the `model_selection` parameter.

GET→PUT re-apply with the profile *unchanged* (the happy roundtrip) is already covered by the `model_id` + `model_selection` match cases (`testCreateIndex_withModelSelectionAndMatchingModelId_thenSucceed` and the transformer unit test).

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
yizheliu-amazon: Good question. I looked into it and I don't think the settings-update path is a safe place for this particular check:

- Validating that the configured `model_id` **exists and has the matching sparse/dense type** requires an ML fetch (`getModel` against `.plugins-ml-model`), which is async cross-index I/O. A `Setting.Validator` runs synchronously on the cluster-manager thread during `PUT _cluster/settings` and is meant for pure value-level checks — doing a blocking ML lookup there would stall cluster-state updates (no precedent for it in core/neural-search). A `SettingsUpdateConsumer` runs *after* the value is applied, so it can only log a warning, not reject — i.e. not truly fail-fast.
- Even if we did it there, it'd be a point-in-time check: the model can be undeployed/deleted after the setting is set, so it could still resolve to a dead/wrong model at index creation. It would also force a strict "deploy the model first, then set the setting" ordering.

Where the check currently sits (`modifyMappings`) the model is already being fetched (`getModels`) to build the semantic-info field, so the type validation is essentially free (no extra round trip), reflects the model's actual state at the moment of use, and fails with a clear error at CreateIndex/PutMapping.

If we want an earlier operator signal we could add a best-effort `SettingsUpdateConsumer` that logs a WARN when a newly-configured model is missing/wrong-type, but I'd lean toward keeping just the authoritative index-time error (and the diagnosability logging tracked in #1963) rather than adding per-node ML calls on every settings change.

bzhangam (approved):

src/main/java/org/opensearch/neuralsearch/mappingtransformer/SemanticMappingTransformer.java
martin-gaievski: I agree with us not changing existing behavior for model_id updates in scope of this PR, but IMO we still need to add few things:
- logging, just a line telling which model a field resolved to, or that a resolution just moved a populated field off its previous model
- test on the update path. all new tests only create an index, nothing exercises re-applying a mapping, so today's behavior isn't pinned. When #1963 later flips the default to immutable, that flip should turn a test red indicating the change

yizheliu-amazon: Thanks — added both:

**Logging.** `SemanticMappingTransformer` now logs at INFO on every model_selection resolution:
`Resolved model_selection [language_option=..., model_type=...] to model_id [<id>] for semantic field at [<path>].`
So you can see which model a field resolved to, and comparing successive lines for the same field surfaces a resolution moving it off its previous model. Detecting the move authoritatively (comparing against the *stored* model_id) needs the old value, which only the mapper merge has — I've left that to #1963.

**Update-path tests pinning today's behavior** (both written so #1963's flip to immutable turns them red):
- Unit `SemanticFieldMapperModelSelectionTests#testMerge_modelSelectionField_modelIdRebindIsAllowedToday` — a model_selection field with `model_id` M1, merged with an update carrying M2, asserts the merged mapper carries **M2** (the binding is updatable today).
- IT `SemanticFieldModelSelectionIT#testReapplyModelSelectionOnly_afterRotation_silentlyRebinds` — create with `model_selection` (resolves M1) → rotate the profile setting to M2 → re-apply `model_selection`-only via `PUT _mapping` → asserts **200** and the field's `model_id` is now **M2** (the silent rebind).

Also, FWIW the model_id-only update path was already pinned by `SemanticFieldMapperTests#testFieldMapper_merge_shouldMergeDelegateFieldMapper` (merge M1 -> newModelId asserts the merged id is the new one).

martin-gaievski (approved):
Re-reviewed after latest changes. Logging on resolution and update-path tests (PUT _mapping, composable template, rotation rejection, canary for #1963) all look good. Approving.

yizheliu-amazon:
### CI status: the red checks are infra/upstream, not this change

The failing checks here are **not** caused by this PR. There are two independent, pre-existing issues:

**1. Eclipse JDT formatter P2-mirror timeout (the large majority of failures).**
Nearly every ubuntu Integ/BWC/remote-model/pre-commit job fails at *project configuration*, before any test runs:

```
A problem occurred configuring root project 'neural-search'.
> java.io.IOException: Failed to load eclipse jdt formatter:
      java.lang.RuntimeException: java.net.SocketTimeoutException: timeout
BUILD FAILED in ~30s
```

Spotless resolves the Eclipse JDT formatter through the `withP2Mirrors(... -> https://ci.opensearch.org/)` mirror in `formatter/formatting.gradle` (added in #1940), and that mirror is currently timing out from most runners. Because it fails at configuration time, it takes down every Gradle task on affected runners, as a per-runner network lottery. This is the same class of failure tracked in opensearch-project/OpenSearch#20826, and `main` is red on these same jobs on recent nightlies.

I've opened #1964 to fix this repo-wide (resolve the formatter directly from download.eclipse.org, matching the OpenSearch 3.x resolution).

Evidence it is not this change:
- `Gradle Check` (JDK 21 + 25, **Linux and Windows**), `Pre-commit Linux 21`, and `Integ Tests Windows 25` pass — build, unit tests, and `spotlessCheck` are all green.
- This PR touches only `SemanticField*` / `ModelSelection*` / resolver / settings files — zero build-config or formatter changes.

**2. Security-enabled highlighting test — 3.9.0-SNAPSHOT drift (also unrelated).**
When a runner does get past configuration, `Run Security Enabled Integration Tests` fails in `SemanticHighlightingLocalModelIT` with a 500 on `_ml/models/_register`:

```
invalid_definition_exception: Direct self-reference leading to cycle
    (through reference chain: org.opensearch.security.user.User["principal"])
```

This is a Jackson serialization incompatibility in the current `opensearch-security` / `opensearch-ml` `3.9.0-SNAPSHOT` builds, only on the security-enabled cluster. This PR touches no highlighting code, and the Secure-Cluster workflow passed on `main` at the same base commit on earlier nightlies (before the snapshot drifted).

**Ask:** once #1964 lands I'll rebase this PR on top. In the meantime, could a maintainer re-run the failed jobs (or proceed on the green required checks)? Thanks!

PR #1924, opened 2026-08-02
Block hybrid query when no search pipeline is configured

### LINKED ISSUE #1922: [PROPOSAL] Block hybrid query in case there are no search pipeline
### What are you proposing?
New validation that should detect if hybrid query is executed without any search pipeline attached, meaning there are none of the inlined, request param or index default pipeline that system can resolve for a search request.

Currently if there are no resolvable search pipeline then hybrid query will silently accepts it and return results in internal format that are not relevant to user query results.

Incorrect results that may be misleading, introduce fail fast behavior so incorrect search configuration can be fixed faster.

Same/no change in API

System may start returning error instead of incorrect result 

Pipeline configuration is manual and error prone process, by design hybrid query needs both components - query and pipeline, but they are not directly connected and misconfiguration is possible

We should be able to resolve pipelines at coordinator level inside hybrid query builder

AdityaWaskar:
### Description
Hybrid query currently executes even when no search pipeline is configured for the request, silently returning results in an internal, unnormalized format instead of failing clearly. This adds a coordinator-level check (in `HybridQuerySearchRequestFilter`) that rejects hybrid query requests when no search pipeline can be resolved — neither via the request (inline or `search_pipeline` param) nor as every target index's default search pipeline.

### Related Issues
Resolves #1922

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/search/HybridQuerySearchRequestFilter.java
+++ b/src/main/java/org/opensearch/neuralsearch/search/HybridQuerySearchRequestFilter.java
@@ -14,21 +14,31 @@ import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchType;
 import org.opensearch.action.support.ActionFilter;
 import org.opensearch.action.support.ActionFilterChain;
+import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.core.action.ActionResponse;
+import org.opensearch.index.IndexSettings;
 import org.opensearch.index.query.QueryBuilder;
 import org.opensearch.neuralsearch.query.HybridQueryBuilder;
 import org.opensearch.neuralsearch.util.HybridQueryUtil;
+import org.opensearch.neuralsearch.util.NeuralSearchClusterUtil;
+import org.opensearch.search.pipeline.SearchPipelineService;
 import org.opensearch.tasks.Task;
+
+import java.util.List;
 
 import lombok.extern.log4j.Log4j2;
 
 /**
- * An ActionFilter that automatically disables batched reduction for hybrid queries.
+ * An ActionFilter that validates hybrid query requests and disables batched reduction for them.
  *
  * This filter intercepts all search requests and checks if they contain a hybrid query.
  * If a hybrid query is detected with search_type=dfs_query_then_fetch, the request is rejected.
- * If a hybrid query is detected, it unconditionally sets batchedReduceSize to Integer.MAX_VALUE
- * to disable batched reduction, regardless of any user-specified value.
+ * If a hybrid query is detected but no search pipeline can be resolved for it (neither inline,
+ * via the search_pipeline request parameter, nor as every target index's default search
+ * pipeline), the request is rejected, since without a pipeline the normalization processor
+ * never runs and the response would otherwise be returned in an unnormalized, internal format.
+ * Otherwise, it unconditionally sets batchedReduceSize to Integer.MAX_VALUE to disable batched
+ * reduction, regardless of any user-specified value.
  *
  * This prevents the "topDocs already consumed" error that occurs when:
  * 1. Hybrid query is executed
@@ -39,8 +49,6 @@ import lombok.extern.log4j.Log4j2;
  * batched reduction is fundamentally incompatible with hybrid query processing.
  * The NormalizationProcessor requires access to all shard results simultaneously
  * to perform score normalization and combination.
- *
- * This filter works transparently without any pipeline or query configuration.
  *
  */
 @Log4j2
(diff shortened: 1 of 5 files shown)


AdityaWaskar:
One design decision worth confirming: for requests spanning multiple indices (or an alias/wildcard), this PR requires **every** matched index to have a default search pipeline configured (or the request itself to set one) before a hybrid query is allowed to proceed — checked via `NeuralSearchClusterUtil#getIndexMetadataList`.

I wasn't sure whether that's the right semantics, or whether the default-search-pipeline fallback should only apply to single-index requests (mirroring how I believe `SearchPipelineService` resolves default pipelines elsewhere). Happy to adjust based on what's intended here.

src/main/java/org/opensearch/neuralsearch/search/HybridQuerySearchRequestFilter.java
```diff
@@ -119,4 +131,33 @@ private boolean containsHybridQuery(SearchRequest searchRequest) {
...
+
+    /**
+     * Check whether a search pipeline can be resolved for this request, either from the request
+     * itself (inline or via the search_pipeline request parameter) or from the default search
+     * pipeline configured on every index the request targets.
+     *
+     * @param searchRequest the search request to check
+     * @return true if a non-noop pipeline can be resolved for this request
+     */
+    private boolean hasResolvableSearchPipeline(SearchRequest searchRequest) {
```
bzhangam, line 143: I think we should inspect the pipeline to confirm the normalization and combination processors are configured otherwise a search pipeline without those processors still cannot return the right hybrid query result.

owaiskazi19: +1 here. Please add a validation to check the normalization and combination processors attached and an associated test

martin-gaievski, line 143: this logic diverges from [SearchPipelineService](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/pipeline/SearchPipelineService.java#L464). Core method `resolvePipeline` checks `source().searchPipelineSource()` first, before the named param.

When Core resolves an inline-body ad-hoc pipeline, request's `searchPipelineSource` map is not drained — meaning it is still intact when this ActionFilter runs (the filter runs before `resolvePipeline`). An inline-object body sets `searchPipelineSource` (`SearchSourceBuilder.java`) and leaves `searchRequest.pipeline()` null (`RestSearchAction` only populates `pipeline()` from the `?search_pipeline=` param or the *string* form of `source().pipeline()`). So `isConfiguredPipeline(null)` is false, no index default exists, and the request is wrongly rejected with a 400, even core would build and run the inline normalization pipeline correctly. 

Example of failing request, where index has no default pipeline, no `?search_pipeline=` param:
```json
POST /my-index/_search
{
  "search_pipeline": {
    "phase_results_processors": [
      { "normalization-processor": {
          "combination":  { "technique": "arithmetic_mean" } } }
    ]
  },
  "query": { "hybrid": { "queries": [ {"match":{"text":"hello"}}, {"term":{"text":"place"}} ] } }
}
```

You can fail fast on the inline body before the named/default checks, mirroring core branch.

src/main/java/org/opensearch/neuralsearch/search/HybridQuerySearchRequestFilter.java
```diff
@@ -119,4 +131,33 @@ private boolean containsHybridQuery(SearchRequest searchRequest) {
...
+        String requestPipeline = searchRequest.pipeline();
+        if (isConfiguredPipeline(requestPipeline)) {
+            return true;
+        }
+
+        List<IndexMetadata> indexMetadataList = NeuralSearchClusterUtil.instance().getIndexMetadataList(searchRequest);
+        return indexMetadataList.isEmpty() == false && indexMetadataList.stream().allMatch(this::hasDefaultSearchPipeline);
```
martin-gaievski, line 150: this logic is different in core, and fails in multiple scenarios:
 - it collapses to `_none` → `NO_OP_PIPELINE` → no normalization runs → exact unnormalized/internal-format response we're trying to address. Intended behavior is to fail fast
 - multi-index mixed defaults → false rejection (over-strict vs. core). with your change we're skipping indices that have no explicit default rather than treating them as `_none`. 
 
I suggest you don't `allMatch` over raw `.get()` values. 
Most robust approach is to delegate to core so the filter cannot diverge. 
Alternatively fold per-index defaults with core's exact semantics — use `.exists()` to skip indices with no default, take the first default, collapse to `_none` if a later index has a different pipeline, then accept only if the folded id is a configured (non-`_none`) pipeline.

src/main/java/org/opensearch/neuralsearch/search/HybridQuerySearchRequestFilter.java
```diff
@@ -82,6 +90,10 @@ public <Request extends org.opensearch.action.ActionRequest, Response extends Ac
                     listener.onFailure(new IllegalArgumentException(HybridQueryUtil.HYBRID_QUERY_DFS_SEARCH_TYPE_NOT_SUPPORTED_MESSAGE));
                     return;
                 }
+                if (hasResolvableSearchPipeline(searchRequest) == false) {
+                    listener.onFailure(new IllegalArgumentException(HybridQueryUtil.HYBRID_QUERY_REQUIRES_SEARCH_PIPELINE_MESSAGE));
```
martin-gaievski, line 94: provided error message is over promising - we are checking for a simple search pipeline presence, not normalization related processors. Either add check for processor or relax the message and drop mentions of processors (which is much easier fix I believe)

martin-gaievski (commented):
Overall that's the right direction but you need to address comments, in addition to that add a single integ tests for failure path

PR #1926, opened 2026-08-04
[Sparse ANN] Fold sparse vector tokens into the signed-short range

### LINKED ISSUE #1925: [BUG] Sparse ANN throw NegativeArraySizeException for token index ranging from 32768-65535
### What is the bug?
In Sparse ANN index which the doc count exceeds `approximate_threshold`, when token index ranges from 32768 to 65535, the sparse ann query will throw `NegativeArraySizeException`. As the logic hashes token index to short type but uses 65535 which generates negative short value. It should use 32768.
 
Create index:
```
{
  "settings": {
      "index": {
          "sparse": true,
          "number_of_shards": 1,
          "number_of_replicas": 0
      }
  },
  "mappings": {
    "properties": {
      "passage_text":{
          "type": "text"
      },
      "passage_embedding":{
          "type": "sparse_vector",
          "method": {
              "name": "seismic",
              "parameters": {
                "n_postings": 4,
                "cluster_ratio": 0.4,
                "summary_prune_ratio": 0.5,
                "approximate_threshold": 4
              }
          }
      }
    }
  }
}
```
bulk
```
{ "index": { "_index": "{{index}}"} }
{ "passage_text": "apple tree", "passage_embedding": {"40000": 1.5, "2000": 2.5},"sparse_embedding": {"40000": 1.5, "2000": 2.5}}
{ "index": { "_index": "{{index}}"} }
{ "passage_text": "banana tree", "passage_embedding": {"2000": 0.5, "3000": 1.3}, "sparse_embedding": {"2000": 0.5, "3000": 1.3}}
{ "index": { "_index": "{{index}}"} }
{ "passage_text": "banana tree", "passage_embedding": {"4000": 3.5, "5000": 1.6}, "sparse_embedding": {"4000": 3.5, "5000": 1.6}}
{ "index": { "_index": "{{index}}"} }
{ "passage_text": "duck", "passage_embedding": {"2000": 3.5, "6000": 1.6}, "sparse_embedding": {"2000": 3.5, "6000": 1.6}}
```

query:
```
{    
    "query": {
        "neural_sparse": {
            "passage_embedding": {
                "query_tokens": {
                    "65538": 2.0
                }
            }
        }
    },
    "size": 5
}
```

It should be able to query token index > 32768

_Operating system, version._

chishui:
### Description
Token IDs are folded into [0, MODULUS_FOR_SHORT) via `token % 65536` and stored in a signed short[] to keep the memory footprint low. Reads went directly through tokens[i], so any token folding into [32768, 65535] was sign-extended back to a negative int, causing:

- getToken()/iterator() to return a negative token (e.g. 40000 -> -25536)
- toDenseVector() to throw NegativeArraySizeException on `maxToken + 1`
- dotProduct() to mis-score via a negative index / bad bounds check

Route every token read through a getToken(int) helper that widens with Short.toUnsignedInt. Writes are unchanged since the (short) cast is a lossless reinterpretation; only reads needed to treat the value as unsigned. Sorting in processListItems() is by the folded non-negative value, so ascending order and dotProduct()'s early-exit remain valid.

Note this path is reachable only when the pure-Java SEISMIC path runs (segment doc count >= approximate_threshold, default 1M); below that the query falls back to plain neural sparse, and the native engine passes raw ints to C++ where term_t is uint16_t. That gating is why the defect went unnoticed.

Adds unit tests covering the constructor/iterator round-trip, toDenseVector(), and dotProduct() for a token folding to a negative short.

### Related Issues
Resolves #1925 

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Bug Fixes
 * [SemanticHighlighter] Fix SemanticHighlighterExtBuilder.toXContent ([#1906](https://github.com/opensearch-project/neural-search/issues/1906)) (query-insights [#651](https://github.com/opensearch-project/query-insights/issues/651))
+* [Sparse ANN] Fold sparse vector tokens into the signed-short range (modulus 32768) so folded tokens are never sign-extended to a negative value when stored in short[]
 
 ### Infrastructure
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Bug Fixes
 * [SemanticHighlighter] Fix SemanticHighlighterExtBuilder.toXContent ([#1906](https://github.com/opensearch-project/neural-search/issues/1906)) (query-insights [#651](https://github.com/opensearch-project/query-insights/issues/651))
-* [Sparse ANN] Fold sparse vector tokens into the signed-short range (modulus 32768) so folded tokens are never sign-extended to a negative value when stored in short[]
+* [Sparse ANN] Fold sparse vector tokens into the signed-short range (modulus 32768) so folded tokens are never sign-extended to a negative value when stored in short[] ([#1926](https://github.com/opensearch-project/neural-search/pull/1926))
 
 ### Infrastructure
 
--- a/src/main/java/org/opensearch/neuralsearch/sparse/common/SparseConstants.java
+++ b/src/main/java/org/opensearch/neuralsearch/sparse/common/SparseConstants.java
@@ -19,7 +19,11 @@ public final class SparseConstants {
     public static final String CLUSTER_RATIO_FIELD = "cluster_ratio";
     public static final String APPROXIMATE_THRESHOLD_FIELD = "approximate_threshold";
     public static final String THREAD_POOL_NAME = "seismic_thread_pool";
-    public static final int MODULUS_FOR_SHORT = 65536;
+    // Tokens are stored in a signed short[] to keep the memory footprint low, so they are folded
+    // into the non-negative signed-short range [0, Short.MAX_VALUE] via this modulus. Using 32768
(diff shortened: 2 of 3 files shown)


chishui:
> ## PR Code Suggestions ✨
> Explore these optional code suggestions:
> 
> **Category**	**Suggestion                                                                                                                                    **	**Impact**
> Possible issue	
> Preserve backward compatibility for existing indices
> **Changing `MODULUS_FOR_SHORT` from 65536 to 32768 is a breaking change for any previously indexed sparse vectors: existing on-disk short[] token arrays were written under modulus 65536 and will now be interpreted under modulus 32768, silently corrupting recall for old indices. Consider a backward-compatibility path (e.g., versioned codec, migration, or reading old segments with the old modulus) rather than switching the constant unconditionally.**
> 
> [src/main/java/org/opensearch/neuralsearch/sparse/common/SparseConstants.java [22-26]](https://github.com/opensearch-project/neural-search/pull/1926/files#diff-c2c08c5fb01c7f4b930eb1217e3b12586de59fe134fa4bc78dfe9903cab35c71R22-R26)
> 
> ```diff
>  // Tokens are stored in a signed short[] to keep the memory footprint low, so they are folded
>  // into the non-negative signed-short range [0, Short.MAX_VALUE] via this modulus. Using 32768
>  // (not 65536) keeps every folded value <= Short.MAX_VALUE, so it round-trips without being
>  // sign-extended to a negative int on read.
> +// NOTE: Older segments were written with modulus 65536; ensure a compatibility path exists
> +// for reading them before removing the old value entirely.
>  public static final int MODULUS_FOR_SHORT = 32768;
> ```
> 
> Suggestion importance[1-10]: 7
> __
> 
> Why: The concern about backward compatibility is legitimate—changing the modulus from 65536 to 32768 would change how tokens are folded, potentially affecting previously indexed data. However, without knowing the release state of the feature, this may or may not be a real issue.
> 
> Medium

Previously, token id from 65536 to 32768 will cause the `NegativeArraySizeException`, and since the hash and casting to short is runtime not persisted, so changing MODULUS_FOR_SHORT to 32768 won't cause backward incompatibility.

yuye-aws (approved):

PR #1933, opened 2026-08-06
Implement base classes and enable fusion for hybrid query (limited to min_max and arithmetic mean)

### LINKED ISSUE #1930: [RFC] In-query fusion for hybrid search
## Overview

Hybrid query today is two coupled pieces: `hybrid` query clause (what to retrieve) and a search pipeline with a `normalization-processor` (how to fuse). That coupling, plus a non-standard internal result format, is the root of the query's current limitations — it must be the top-level query, it can't be nested or appear more than once, it's unusable without the right pipeline (and silently returns raw scores when one is missing), and several advanced features work only through hybrid-specific reimplementations.

This RFC proposes an evolution of the existing `hybrid` query — a **resolver mode** — that dispatches each sub-query as an independent leg, fuses the results on the coordinator, and then rewrites itself into a standard query that runs with no hybrid-specific code path and no required pipeline. It keeps the familiar `hybrid` query surface (zero migration) while removing the structural limits.

We are seeking community feedback on two things: **(1) overall direction** — evolving `hybrid` in place vs. a new construct — and **(2) the API surface** for turning the mode on and supplying fusion configuration.

> **Relationship to the retriever framework RFC (**[**#22495**](https://github.com/opensearch-project/OpenSearch/issues/22495)**).** That RFC proposes a new, higher-abstraction `retriever` framework for composable multi-stage retrieval. This RFC is **complementary, not competing**: it applies the same fan-out/fuse/self-erase execution idea to the *existing `hybrid` query* to give query-level flexibility (nestable, usable inside `bool`/`dis_max`/`function_score`, multiple instances per request, zero migration for current hybrid users), while the retriever framework serves the higher-level orchestration use cases (interleaving, A/B testing, multi-stage pipelines). The two can share the coordinator-side fusion machinery. 

## Motivation

Current architecture is close to the limit of what it can support. Limitations fall into three groups.

### 1. Limited composability

The `hybrid` query must be the top-level query. It cannot be nested inside `bool`, `dis_max`, or `function_score`, cannot appear more than once in a request, and cannot be rescored meaningfully (rescore applies inside a single sub-query, so it's ineffective when a document matches only one leg). Customers have asked for all of these (e.g. filtering/boosting around a hybrid block, combining hybrid with other compound queries).

```json
// Wanted today, but rejected — hybrid must be top-level
{ "query": { "bool": {
    "must":   [ { "hybrid": { "queries": [ /* match + neural */ ] } } ],
    "filter": [ { "term": { "in_stock": true } } ] } } }
```

### 2. Latency

Within each segment the sub-queries are scored sequentially in one collector, so cost trends toward the *sum* of the legs rather than the *max*. On an expensive kNN leg this is worse than the sum, because the kNN sub-query loses its native ANN scorer inside `HybridQuery`, and the classic path force-disables batched reduce (a coordinator barrier that grows with shard count). Many users expect combined latency ≈ the slowest leg plus a small overhead, and today's alternative is to run each sub-query as a separate request and fuse outside OpenSearch.

### 3. Manual error prone setup

"What to retrieve" (the query clause) and "how to fuse" (the normalization/combination technique and weights, in a search pipeline) live in two different places. A missing or wrong pipeline returns raw, uncombined scores with no error. New users must learn two coupled concepts to run one hybrid search.

### At a glance

| Area | Current state | With resolver mode |
| --- | --- | --- |
| Placement | Top-level only | Nestable in `bool`/`dis_max`/`function_score`; multiple per request |
| Fusion config | In a search pipeline (separate from the query) | Inline in the query, or read from the pipeline (zero migration) |
| Missing pipeline | Silent raw scores | Fail fast with a clear error |
| Rescore | Inside one sub-query (ineffective) | Standard rescore over the fused result |
| Latency (expensive leg) | Trends to sum of legs; batched reduce disabled | Legs run as parallel searches (≈ slowest leg); batched reduce kept |
| Result format | Non-standard internal format | Standard query results; all standard features apply |

## Proposal

Resolver mode replaces the coordinator post-processing pipeline with a **two-round, orchestration-based** model, fully inside the coordinator rewrite:

1. **Fan out.** The coordinator fires each sub-query as an independent leg (one `MultiSearch`, run in parallel), collecting each leg's top-window document ids and scores.
2. **Fuse.** It normalizes and combines the leg scores into one ranked list (reusing today's normalization/combination techniques — one relevance implementation).
3. **Self-erase.** It rewrites the `hybrid` query into a **standard `bool` query** that encodes the fused ranking, and the search proceeds as an ordinary query. Every standard feature (sort, pagination, collapse, aggregations, rescore) operates on it with no hybrid-specific code.

```
TODAY (pipeline post-processing)                RESOLVER MODE (orchestrate + self-erase)

  hybrid query ─► shards score all legs           hybrid{fusion} ─► coordinator rewrite
                  sequentially per segment                          │
                  │ (non-standard ScoreDoc format)                  ├─► round 1: fan out legs (parallel MultiSearch)
                  ▼                                                  │      collect ids + scores
   coordinator: normalization-processor                             ├─► fuse on coordinator (normalize + combine)
                (needs a search pipeline)                           │
                  ▼                                                  └─► round 2: self-erase into a standard
   reduce ─► fetch ─► response                                            bool query ─► normal query/fetch ─► response
```
The self-erased query is a standard `bool`: a scored **Top** (`constant_score(ids)^fusedScore`, one clause per ranked doc — this stamps the coordinator's fused score as the Lucene `_score`) plus an optional non-scoring **Tail** (`bool{ filter: {should: legs} }`) that defines the full match set so `total_hits` and aggregations stay correct. When the request needs neither exact totals nor aggregations/highlight, the Tail is dropped — a **fast path** (see below).

## API design options

The mechanism is the same across all options; they differ in how the mode is turned on and where fusion config comes from. All extend the existing `hybrid` query except Option 3.

### Option 1 - single overloaded `fusion` parameter on `hybrid` [Preferred]

One new optional field, `fusion`, on the existing `hybrid` query. Its **presence** turns on resolver mode; its **shape** carries the config. No second flag, no separate window parameter.

```json
// Zero-migration: read fusion config from the attached pipeline
POST /my-index/_search?search_pipeline=my_norm_pipeline
{ "query": { "hybrid": {
    "fusion": "pipeline",
    "queries": [ { "match": { "text": "climate policy" } },
                 { "neural": { "vec": { "query_text": "climate policy", "model_id": "abc", "k": 50 } } } ] } } }

// Self-describing: inline fusion config (wins over any pipeline)
POST /my-index/_search
{ "query": { "hybrid": {
    "queries": [ /* match + neural */ ],
    "fusion": {
      "combination":   { "technique": "arithmetic_mean", "parameters": { "weights": [0.3, 0.7] } },
      "window_size": 100 } } } }
```

Dispatch: `fusion` absent = classic path (unchanged); `"pipeline"` / `{source:"pipeline"}` = read from the attached pipeline; a `normalization`/`combination` block = inline config; `{}` or `{window_size:N}` = built-in defaults. Nestable and repeatable because presence rides in the query body.

**Pros:** 
- zero migration (existing tutorials, SDKs, dashboards, `default_pipeline` all keep working); 
- one parameter is both the on switch and the config; 
- self-contained request; 
- extensible. 

**Cons:** 
- slightly more parsing logic than a plain boolean flag; 
- "on but read-from-pipeline" case needs an explicit marker (`source:"pipeline"`) rather than an empty object, to avoid silently applying an unrelated index-default pipeline.

### Option 2 - `fusion_mode` flag plus a separate `fusion` block on `hybrid`

A `fusion_mode: "processor" | "query"` flag turns on the resolver; an optional separate `fusion` block (and a separate window parameter) carries inline config.

```json
{ "query": { "hybrid": {
    "fusion_mode": "query",
    "queries": [ /* match + neural */ ],
    "fusion": { "normalization": { "technique": "min_max" }, "combination": { "technique": "arithmetic_mean" } },
    "rank_window_size": 100 } } }
```

**Pros:** 
- on/off switch is explicit and obvious; 
- config block mirrors the processor JSON. 
**Cons:** 
- three related parameters where one would do; 
- more surface to validate for mutually-exclusive combinations.

### Option 3 - new top-level `resolver` query type

Ship the resolver as its own query type, all config inline, leaving classic `hybrid` untouched.

```json
{ "query": { "resolver": {
    "queries": [ /* match + neural */ ],
    "combination":   { "technique": "arithmetic_mean", "parameters": { "weights": [0.3, 0.7] } },
    "window_size": 100 } } }
```

**Pros:**
- clean slate, no legacy behavior to preserve, simplest internal logic

**Cons:** 
 - highest migration cost — the hybrid-search UX and relevance-tuning tooling (SRW) must be re-introduced; 
 - two query types to maintain (fixes + new features in both); 

### Option comparison

| Option | Surface | Config source | Migration | Verdict |
| --- | --- | --- | --- | --- |
| **1 (preferred)** | Extend `hybrid`, single `fusion` field | Inline / pipeline / defaults (presence + shape) | Zero | **Recommended** — one param, zero migration, self-describing |
| 2 | Extend `hybrid`, `fusion_mode` + `fusion` + window | Inline / pipeline | Zero | Works; more parameters than needed |
| 3 | New `resolver` query | Inline only | High | Cleanest logic, worst migration + double maintenance |

## Keeping the cluster safe

The resolver fans one user request into N leg searches plus a round-2 query, all issued from the coordinator. Two questions for the community:

- **Fan-out amplification.** One request becomes N+1 internal searches. We propose caps on leg count and `window_size`, and reusing the existing search-level safeguards (circuit breakers, task cancellation) rather than building new ones. Should the leg MultiSearch also honor `max_concurrent_search_requests`-style limits per request?
- **Coordinator load.** Fan-out orchestration, fusion arithmetic, and round-2 coordination all run on the coordinator. Measurements (shared separately) show the per-query coordinator CPU cost is within ~1.1–1.4× of classic at small windows and *below* classic at large windows (classic normalizes a larger candidate set on the coordinator), with no thread-pool rejections under load — so a shared/multi-tenant coordinator is not the bottleneck. We'd welcome input on additional safeguards for shared-coordinator deployments.

Unlike a raw MultiSearch, resolver's legs are internal and uniform (same target indices, id-only results), so per-request accounting is simpler; we don't propose exposing the legs as user-visible tasks.

## Fast path

For requests that only need the top-K documents (no exact `total_hits`, no aggregations/highlight), round 2 drops the Tail and becomes a pure `constant_score(ids)` lookup over the fused window — no full-index scan, no leg re-execution. This is the single biggest latency lever and is opt-in through the existing `track_total_hits: false` flag. Ranking is identical to the full path (same technique, same window, doc-for-doc); only the exact match count is given up.

```
round 2, standard:  bool{ Top (constant_score ids)  +  Tail (filter: should[legs]) }   ← exact totals/aggs
round 2, fast path: bool{ Top (constant_score ids) }                                   ← top-K only, lowest latency
```

## What this unlocks

- **Rescore after hybrid** — a standard rescorer over the fused result, feature community has asked for
- **Hybrid nested in compound queries** — filter/boost around a hybrid block; multiple hybrids per request.
- **Aggregations/sort/collapse without hybrid-specific caveats** — they run on a standard query.
- **Pipeline-free operation** — inline fusion config; a missing config fails fast instead of returning raw scores.
- **Shared fusion machinery** with the retriever framework ([#22495](https://github.com/opensearch-project/OpenSearch/issues/22495))
- Related requests: hybrid-query composability and per-query weighting asks (neural-search [#1005](https://github.com/opensearch-project/neural-search/issues/1005), [#1172](https://github.com/opensearch-project/neural-search/issues/1172)).

We propose the two share the coordinator fusion implementation so relevance stays consistent across both surfaces. Direction feedback on whether these should be developed as one component or two coordinated ones is welcome.

## Open questions

1. **API surface.** Which of the three options above should we adopt for turning on resolver mode and supplying fusion config? (We prefer Option 1 — single overloaded `fusion` param on `hybrid`.)
2. **Cluster safeguards.** Beyond leg-count / window caps and the existing circuit breakers, what additional protections do shared-coordinator / multi-tenant deployments need for the fan-out?

martin-gaievski:
### Description
Initial scope for resolver in hybrid query, targeting feature branch:
- foundation: fusion mode entry point, fan-out (round 1) and tail (round 2), self-erase query, utility classes for fusion spec
- refactor math classes for both classic and fusion into sharable utils
- one end-to-end test, UT

After this PR following query will work:

1. Inlined fusion configuration, including explicit `window` and `weights` params
```
  {
    "hybrid": {
      "queries": [ ... ],
      "fusion": {
        "combination": { "technique": "arithmetic_mean", "parameters": { "weights": [0.7, 0.3] } },
        "window_size": 100
      }
    }
  }
```

2. Read config from pipeline (zero migration path). 
```
  { "hybrid": { "queries": [ ... ], "fusion": {} } }
  { "hybrid": { "queries": [ ... ], "fusion": "pipeline" } }
  { "hybrid": { "queries": [ ... ], "fusion": { "source": "pipeline" } } }
```
these require an attached pipeline resolvable via inline body / ?search_pipeline= / index.search.default_pipeline

Intentional limitation: 

1. Rejected at parse (HTTP 400)

  - "fusion": true / any non-string, non-object
  - "fusion": "anything-but-pipeline" (string form must be "pipeline")
  - { "source": "pipeline", "normalization": {…} } — contradiction (read-from-pipeline vs inline)
  - unknown key: { "foo": 1 } (only source|normalization|combination|window_size allowed)
  - "window_size": 0 or negative
  - fusion together with pagination_depth

2. Parses, but fails fast at rewrite (HTTP 400)

  - Any technique other than min_max + arithmetic_mean: z_score, l2, geometric_mean, harmonic_mean, RRF (combination.technique: rrf). These parse fine but requireSupportedTechniques rejects them until a later PR.
  - fusion present but no resolvable config (no inline techniques, no pipeline, no index default).
  - window_size > index.max_result_window (the check added this session).

There will be follow up PRs with:
- more techniques under fusion + extended test coverage, this PR kept smaller intentionally to easier review. 
- hybrid{fusion} nested inside bool/dis_max/function_score. rewrite computes a topLevel flag and would self-erase Top-only when nested, but it's not validated end-to-end yet.
- fetch-time features (aggregations correctness, explain/profiler, PIT, gRPC) are coming later.

### Related Issues
Resolves #1930 

### Check List


src/main/java/org/opensearch/neuralsearch/fusion/rankdocs/FusedDoc.java
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Fused document identity: a Lucene-local {@code docId}, its fused {@code score}, and the {@code shardIndex} it
+ * belongs to. Deliberately a plain value — <b>not</b> {@code Writeable} — so no feature's wire-identity contract is
+ * baked into a resolver-authored type. It is a candidate for the retriever framework to lift into OpenSearch core when
+ * that framework's orchestration lands there; until then it lives resolver-locally.
+ *
+ * <p>Populated only on the {@code (shardIndex, docId)} fast path (PIT held). The default {@code _id}-keyed self-erase
+ * path ({@link org.opensearch.neuralsearch.query.HybridFusionQuery}) does not use it.
+ */
+public record FusedDoc(int docId, float score, int shardIndex) {
```
vibrantvarun: Why we need FusedDoc?  This class has same fields which ScoreDoc has.

src/main/java/org/opensearch/neuralsearch/fusion/rankdocs/FusedDocsScorerQuery.java
```diff
@@ -0,0 +1,80 @@
...
+
+    private final int[] docIds;
+    private final float[] scores;
+
+    /**
+     * @param docIds Lucene-local doc ids for this shard, ascending (the sorted-int scan relies on ordering)
+     * @param scores fused scores positionally aligned to {@code docIds}
+     */
+    public FusedDocsScorerQuery(int[] docIds, float[] scores) {
+        if (Objects.isNull(docIds) || Objects.isNull(scores) || docIds.length != scores.length) {
```
vibrantvarun: In which scenario, either would be null? Or the array lengths would be different. Asking to backtrack the scenarios.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -0,0 +1,263 @@
...
+            throw new IllegalStateException(
+                String.format(Locale.ROOT, "[hybrid] expected %d leg sub-search responses but got %d", legCount, items.length)
+            );
+        }
+        SearchHit[][] legHits = new SearchHit[legCount][];
+        int survivingLegs = 0;
+        for (int leg = 0; leg < legCount; leg++) {
+            MultiSearchResponse.Item item = items[leg];
+            if (item.isFailure()) {
+                log.warn("[hybrid] fused-mode sub-query {} dropped: {}", leg, item.getFailureMessage());
```
vibrantvarun: HybridFusionOrchestrator.groupLegHits drops a failed leg with only a log.warn and fuses the survivors. No _shards.failures, no response warning. A user can silently get single-leg (e.g.
  pure-BM25) relevance and never know. Classic hybrid surfaces shard failures normally. Recommend: at minimum propagate to response._shards.failures, or make partial-leg-failure behavior
  configurable (fail-fast vs. degrade).

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -326,6 +420,163 @@ protected QueryBuilder doRewrite(QueryRewriteContext queryShardContext) throws I
...
+                )
+            );
+        }
+        // Current scope (first working slice): min_max + arithmetic_mean only. Other techniques parse but are not wired
+        // into the coordinator fusion path yet — fail fast rather than silently mis-fuse.
+        requireSupportedTechniques(fusionSpec);
+
+        // Whether this query is the whole request (=> conditional Tail for accurate totals/aggs) or nested inside a
+        // container (=> Top-only, so an enclosing filter intersects at the query phase).
+        boolean topLevel = Objects.nonNull(searchRequest.source()) && searchRequest.source().query() == this;
```
vibrantvarun, line 484: This relies on reference identity — topLevel is only true when the fused rewrite runs on the exact instance still parked at source().query(). That holds today, but it's an unenforced  invariant: if a request-rewrite layer (search-request processor, security/field-masking, or a future core change) ever clones the query before this point, a genuinely top-level query would compare != and get misclassified as nested → forced Top-only → silently drops the Tail. The failure mode is nasty because scores and top hits still look correct; only total_hits and aggregations would be wrong (computed over just the fused window).

Could we either (a) add a comment documenting that == is intentional and the invariant it depends on, or (b) thread the nesting/top-level signal down explicitly (e.g. via the coordinator
  context) so it survives cloning? Since nested fusion isn't validated end-to-end yet, this isn't blocking — but worth hardening before that follow-up lands.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -326,6 +420,163 @@ protected QueryBuilder doRewrite(QueryRewriteContext queryShardContext) throws I
...
+        return marker;
+    }
+
+    /** True when the inline {@code fusion} block carries an actual config (not just {@code source: pipeline} / empty). */
+    private static boolean hasInlineConfig(final Map<String, Object> fusion) {
+        return fusion.containsKey(FUSION_KEY_NORMALIZATION) || fusion.containsKey(FUSION_KEY_COMBINATION);
+    }
+
+    /** Fail fast on techniques not yet wired into the coordinator fusion path (current scope: min_max + arithmetic_mean). */
+    private static void requireSupportedTechniques(final FusionSpec fusionSpec) {
```
vibrantvarun, line 551: Not an issue in this PR (only one pair is allowed). But when techniques expand, please have the fused path reuse ScoreNormalizationFactory's existing compatibility check rather than  re-implementing the normalization↔combination rules — otherwise the two paths' validation will drift.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -326,6 +420,163 @@ protected QueryBuilder doRewrite(QueryRewriteContext queryShardContext) throws I
...
+                    "[%s] query [%s] requires a normalization or score-ranker processor in the attached search pipeline "
+                        + "(inline body, ?search_pipeline=, or index.search.default_pipeline), or an inline fusion block; none was found",
+                    NAME,
+                    FUSION_FIELD.getPreferredName()
+                )
+            );
+        }
+        // Current scope (first working slice): min_max + arithmetic_mean only. Other techniques parse but are not wired
+        // into the coordinator fusion path yet — fail fast rather than silently mis-fuse.
+        requireSupportedTechniques(fusionSpec);
```
vibrantvarun, line 471: Weights are validated lazily inside ScoreCombinationUtil when the combiner is built in the async callback — i.e. after the leg MultiSearch already fired. A bad weights array (wrong count/ not summing to 1.0) wastes a full fan-out before erroring. Not blocking, but consider validating weight count/range at parse-time here (inline case) and before registerAsyncAction in doRewriteFused (pipeline case) to fail fast like the rest of the fusion validation.

src/test/java/org/opensearch/neuralsearch/query/FusionConfigResolverTests.java
```diff
@@ -0,0 +1,200 @@
...
+import org.opensearch.core.index.Index;
+import org.opensearch.core.xcontent.MediaTypeRegistry;
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.neuralsearch.util.NeuralSearchClusterUtil;
+import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.search.pipeline.PipelineConfiguration;
+import org.opensearch.search.pipeline.SearchPipelineMetadata;
+import org.opensearch.test.OpenSearchTestCase;
+
+public class FusionConfigResolverTests extends OpenSearchTestCase {
```
vibrantvarun, line 33: These branches in FusionConfigResolver.resolve look uncovered: 
(1) two indices with different index.search.default_pipeline collapsing to _none

(2) inline body + ?search_pipeline= both set → hard error. Worth adding — they're the error/edge paths most likely to regress silently.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryBuilderTests.java
```diff
@@ -361,6 +368,539 @@ public void testDoToQuery_whenTooManySubqueries_thenFail() {
         assertThat(exception.getMessage(), containsString("Number of sub-queries exceeds maximum supported"));
     }
 
+    // ---- resolver (fused) mode: `fusion` parameter parse / round-trip / validation (PR1, execution inert) ----
```
vibrantvarun, line 371: validateWindowSizeAgainstMaxResultWindow (window_size > index.max_result_window → 400) doesn't appear to have a test. Please add one.

src/main/java/org/opensearch/neuralsearch/fusion/CoordinatorScoreFusion.java
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Fuse legs with {@code min_max} normalization followed by the given combination technique.
+     *
+     * @param legRawScores        one entry per leg, each an {@code _id -> raw score} map of that leg's hits (order is
+     *                            the leg order; a doc absent from a leg's map did not match that leg)
+     * @param combinationTechnique the same combination technique classic would use (e.g. arithmetic_mean with weights)
+     * @return {@code _id -> fused score} for the union of all legs' docs
+     */
+    public static Map<String, Float> fuseMinMax(
```
vibrantvarun, line 52: Combination is already dynamic (picked by name via ScoreCombinationFactory), but normalization is frozen to min_max — because ```ScoreNormalizationTechnique.normalize(NormalizeScoresDTO)``` mutates shard-shaped CompoundTopDocs, which the coordinator's _id → score maps can't use.
 

Suggestion for the follow-up PRs: extract a small two-phase interface so normalization can be picked by name too —
```
  interface ScalarNormalizer {
      LegStats computeLegStats(float[] legScores);  // phase 1: summarize a leg
      float normalize(float score, LegStats stats); // phase 2: score a doc
}
```
Every technique already splits this way internally, so l2 ({l2Norm}) and z_score ({mean, stddev}) drop in cleanly. Then fuse looks the normalizer up by name and this hardcode + requireSupportedTechniques both go away.
  

 Two caveats: z_score stats must accumulate {count, sum, sumSq} to merge across shards (can't merge per-shard mean/stddev); and RRF is rank-based, so it won't fit this interface and needs  its own path.

 Not needed for this PR — just worth introducing the abstraction now so each future technique is a drop-in class rather than another branch here.

src/main/java/org/opensearch/neuralsearch/fusion/rankdocs/FusedDoc.java
martin-gaievski: I was planning to use it in later PRs for PIT, but agree it's not needed now. Will remove it

src/main/java/org/opensearch/neuralsearch/fusion/rankdocs/FusedDocsScorerQuery.java
martin-gaievski: in this PR neither can happen from a production path, because FusedDocsScorerQuery has no production caller yet. It's the deferred (shardIndex, docId)/PIT fast-path scorer; createWeight throws UnsupportedOperationException until PIT is implemented in upcoming PRs.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
martin-gaievski: Agreed, the `==` this is deliberate reference identity (comparing against source().query()), but you're right it's an undocumented invariant, and a clone upstream would silently downgrade a top-level query to Top-only: correct scores/top-hits, but wrong total_hits and aggregations (computed over just the fused window). 

I'll add a comment in this PR making explicit that the == is intentional. I'll defer option (b) to the follow-up that wires and validates nested fusion end-to-end, where that robustness actually earns its complexity, and leave a TODO referencing this thread.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
martin-gaievski: agreed. today requireSupportedTechniques is a single-pair scope gate (only min_max+arithmetic_mean are wired into the coordinator path), not a re-implementation of the compatibility matrix, so there's no drift yet.
When we extend fused support I'll have it delegate to ScoreNormalizationFactory.isTechniquesCompatible (TechniqueCompatibilityCheckDTO from the resolved FusionSpec) so both paths share one source of truth for normalization and combination rules.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
martin-gaievski: good catch, bad weights array indeed burns full MultiSearch before ScoreCombinationUtil errors. 
I'll add a fail-fast in this PR

src/test/java/org/opensearch/neuralsearch/query/FusionConfigResolverTests.java
martin-gaievski: that's not accurate, they both are covered
 inline-body + named-?search_pipeline= mutual-exclusion error is tested by testResolve_whenInlineAndNamedBothPresent_thenThrows (FusionConfigResolverTests), asserting the "Both named and inline search pipeline were specified" IllegalArgumentException. The two-indices-with-different-defaults collapse to _none is tested by testResolve_whenConflictingIndexDefaults_thenNull, which targets idx-a/idx-b with pipe-a/pipe-b defaults and asserts resolve returns null.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryBuilderTests.java
martin-gaievski: This is already tested by testDoRewriteFused_whenWindowSizeExceedsMaxResultWindow_thenFailsFast (HybridQueryBuilderTests.java): it sets index.max_result_window=100, drives a fused query with window_size=500 through doRewrite, and asserts the IllegalArgumentException whose message contains max_result_window, exact window_size > index.max_result_window → 400 path. Since validateWindowSizeAgainstMaxResultWindow is private static and called from the rewrite path, it's driven through doRewrite; the within-bounds companion testDoRewriteFused_whenWindowSizeWithinMaxResultWindow_thenProceeds covers the pass side.

src/main/java/org/opensearch/neuralsearch/fusion/CoordinatorScoreFusion.java
martin-gaievski: Agreed. Combination flows through ScoreCombinationFactory by name while normalization is hardcoded, since fuseMinMax calls MinMaxScoreNormalizer.normalizeSingleScore directly, because classic ScoreNormalizationTechnique mutates CompoundTopDocs in place and the coordinator only holds _id-to-score maps. 
Two-phase ScalarNormalizer (computeLegStats(float[]) → LegStats; normalize(score, stats) → float) is exactly what fuseMinMax already hand-rolls, so l2 and z_score slot in and requireSupportedTechniques could drop its normalization arm. 

One details on z_score: since we materialize each leg's full window as a float[] on the coordinator, its
mean/std compute directly in computeLegStats just like min/max — the {count, sum, sumSq} mergeability only matters if we later want population-accurate stats beyond the window (shard-side partials), which is orthogonal here; RRF stays on its own rank-based path. 

I'll defer the extraction to the follow-up that lands the second scalar technique, so we don't guess the LegStats shape before we have two consumers.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: ack, will be adding validation to fail fast

vibrantvarun: Curious, how are we failing fast here in case of failure on one leg?

martin-gaievski: we do not fail fast on a single leg failure as of now. I was still evaluating options how we can do it properly:
- unconditional fail fast - most straightforward way, but the transient single-leg hiccup fails whole query
- honor allow_partial_search_results: false -> fail fast; true -> degrade
- new pre-query key 
- warning header on degrade, emit a Warning/299 OpenSearch header naming the failed leg
I guess for this PR we can do first option and fail whole query immediately, then adjust in later PRs

vibrantvarun (approved):

src/main/java/org/opensearch/neuralsearch/query/HybridFusionQuery.java
```diff
@@ -0,0 +1,172 @@
...
+ *       scoring {@code should} clauses, so the fused window is returned in fused-score order.</li>
+ *   <li><b>Tail</b> — a {@code bool{ should: [sourceQuery...] }} added as a non-scoring {@code filter}. It matches the
+ *       full set of documents any sub-query matched, so {@code total_hits} and aggregations cover all matches (not just
+ *       the ranked window) and the highlighter has the sub-queries' terms available. Included only when the request
+ *       needs the full match set; omitted (Top-only) for plain top-K and for a nested fused query.</li>
+ * </ul>
+ *
+ * <p>This query is created internally by the coordinator self-erase and is never parseable from a search request.
+ */
+public class HybridFusionQuery extends AbstractQueryBuilder<HybridFusionQuery> {
```
bzhangam: I think this is a QueryBuilder rather than a Query. Can we rename it?

src/main/java/org/opensearch/neuralsearch/fusion/rankdocs/FusedDocsScorerQuery.java
```diff
@@ -0,0 +1,80 @@
...
+ * {@code (shardIndex, docId)} / PIT fast-path scorer; the default {@code _id}-keyed self-erase path
+ * ({@link org.opensearch.neuralsearch.query.HybridFusionQuery}, a {@code constant_score(IdsQuery)} bool) does not need
+ * it, so this class is <b>not wired to execution in PR1</b>.
+ *
+ * <p>Kept small and dependency-light (no {@code Writeable}, no plugin-specific types) so it is a clean candidate for
+ * the retriever framework to lift into OpenSearch core later; the resolver would then adopt the core class and retire
+ * this copy. The scoring {@link Weight}/iterator implementation is deferred to PR5, where the PIT/docId fast path is
+ * introduced — {@link #createWeight} intentionally throws until then.
+ */
+public final class FusedDocsScorerQuery extends Query {
```
bzhangam: Since we don't plan to use it in this PR maybe it's better to exclude it? And for the retriever framework we may also have a different implementation.

src/main/java/org/opensearch/neuralsearch/fusion/CoordinatorScoreFusion.java
```diff
@@ -0,0 +1,93 @@
...
+ *       that did not match a doc therefore contributes a {@code 0.0} slot, which the arithmetic-mean combiner counts
+ *       toward the denominator (its {@code score >= 0.0} participation rule), exactly as classic does.</li>
+ * </ul>
+ *
+ * <p>Current scope: {@code min_max} normalization + the supplied combination technique (arithmetic_mean). Other techniques
+ * (z_score, l2, RRF) join this path in a later change. This entry is not reachable from a live request yet — the
+ * coordinator rewrite wiring lands with the execution path.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class CoordinatorScoreFusion {
```
bzhangam, line 42: **Design suggestion: separation of concerns in `CoordinatorScoreFusion`**

Currently `fuseMinMax()` computes per-leg min/max bounds, normalizes scores, and combines them — all in one method. This works for min_max alone, but once z_score, l2, and RRF land on this path, we'd end up with separate `fuseMinMax` / `fuseZScore` / `fuseL2` / `fuseRRF` methods each with their own inline stats-gathering or rank-computation loop.

What if we extract a per-leg transformation step? Something like:

```java
interface ScoreTransformer {
    List<Float> transform(List<Float> rawScores);
}
```

Each implementation computes what it needs internally and returns values ready for combination:

| Technique | `transform()` does |
|-----------|-------------------|
| min_max | compute min/max, apply `(s-min)/(max-min)` |
| z_score | compute mean/stddev, apply `(s-mean)/stddev` |
| l2 | compute L2 norm, apply `s/norm` |
| RRF | ignore score values, return `1/(k+rank)` by position |

Then `CoordinatorScoreFusion` becomes technique-agnostic:

```java
public static Map<String, Float> fuse(
    List<Map<String, Float>> legRawScores,
    ScoreTransformer transformer,
    ScoreCombinationTechnique combiner
) { ... }
```

This pairs `ScoreTransformer` with the existing `ScoreCombinationTechnique` — one transforms per-leg scores, the other combines across legs. Adding a technique means implementing the interface, no changes to the coordinator. It also mirrors how the classic pipeline already models RRF (normalization step replaces scores with reciprocals, combination step sums them) — just with a name that doesn't imply "normalization" for rank-based techniques.

Not blocking — the current code is correct. Just thinking about extensibility for the upcoming techniques.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -326,6 +420,172 @@ protected QueryBuilder doRewrite(QueryRewriteContext queryShardContext) throws I
...
+                                fusionSpec,
+                                window,
+                                topLevel
+                            )
+                        );
+                        listener.onResponse(null);
+                    } catch (Exception e) {
+                        listener.onFailure(e);
+                    }
+                }, listener::onFailure)
```
bzhangam: Should we provide a better error message here since the error from multiSearch can be confusing since end users think they are doing hybrid query?

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -0,0 +1,274 @@
...
+     * leg's full window, so a leg silently truncated by a down shard would be as corrupting as a fully failed leg.
+     * Failing the leg instead lets {@link #groupLegHits} fail the whole request fast, rather than fusing over incomplete
+     * data — this is the leg-level half of the fail-fast contract.
+     */
+    static MultiSearchRequest buildLegMultiSearch(SearchRequest request, List<QueryBuilder> legs, int windowSize) {
+        MultiSearchRequest multiSearchRequest = new MultiSearchRequest();
+        for (QueryBuilder leg : legs) {
+            SearchSourceBuilder legSource = new SearchSourceBuilder().query(leg)
+                .size(windowSize)
+                .from(0)
```
bzhangam: Why from 0? Isn't the default behavior?

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -0,0 +1,274 @@
...
+     * <p>Legs disable partial results ({@code allowPartialSearchResults(false)}): fused relevance is computed across a
+     * leg's full window, so a leg silently truncated by a down shard would be as corrupting as a fully failed leg.
+     * Failing the leg instead lets {@link #groupLegHits} fail the whole request fast, rather than fusing over incomplete
+     * data — this is the leg-level half of the fail-fast contract.
+     */
+    static MultiSearchRequest buildLegMultiSearch(SearchRequest request, List<QueryBuilder> legs, int windowSize) {
+        MultiSearchRequest multiSearchRequest = new MultiSearchRequest();
+        for (QueryBuilder leg : legs) {
+            SearchSourceBuilder legSource = new SearchSourceBuilder().query(leg)
+                .size(windowSize)
```
bzhangam: For knn query do we need to use the windowSize as the k if k is not set in the knn query itself? knn query by default will use 10 as the k which may not match the windowSize.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -0,0 +1,274 @@
...
+                .from(0)
+                .fetchSource(false)
+                .trackTotalHits(false);
+            multiSearchRequest.add(
+                new SearchRequest(request.indices()).indicesOptions(request.indicesOptions())
+                    .source(legSource)
+                    .pipeline(SearchPipelineService.NOOP_PIPELINE_ID)
+                    .allowPartialSearchResults(false)
```
bzhangam: why hardcode to false? Shouldn't inherit from the hybrid query? Today in classic hybrid query we don't always enforce this.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -0,0 +1,274 @@
...
+        return legHits;
+    }
+
+    /**
+     * Fuse via the shared {@link CoordinatorScoreFusion} core (min_max + arithmetic_mean), then rank by fused score and
+     * cut to the window. Converts the coordinator's {@code SearchHit[][]} view into the {@code _id}-keyed per-leg maps
+     * the shared core consumes; a leg that matched nothing contributes an empty map (groupLegHits fails fast on failures,
+     * so every slot is non-null).
+     */
+    private static RankedDocs computeRankedDocs(SearchHit[][] legHits, FusionSpec fusion, int windowSize) {
```
bzhangam, line 152: **Potential issue: `_id` collision in multi-index searches**

In `computeRankedDocs`, leg hits are keyed purely by `hit.getId()`:

```java
Map<String, Float> byId = new LinkedHashMap<>();
for (SearchHit hit : hits) {
    byId.put(hit.getId(), hit.getScore());
}
```

For a single-index search this is fine — each `_id` is unique. But for multi-index searches (e.g., `POST /index-a,index-b/_search`), two different documents in different indices can share the same `_id`. In that case:

1. Within a single leg's map, the second doc would overwrite the first (last-write-wins).
2. Across legs, a doc from `index-a` in leg 0 and a different doc from `index-b` in leg 1 would be treated as the same document during fusion — their scores would be combined as if they're one entity.

The fused `HybridFusionQuery` then uses `IdsQueryBuilder` with those ids, which would match both documents, and both would get the same fused score.

Should the key be `_index + _id` (or use `hit.getIndex() + "#" + hit.getId()`) to avoid conflation? The downstream `IdsQueryBuilder` in the Top would also need to become index-aware (or use a `bool` with `terms` on `_id` + filter on `_index`).

This may be an acceptable limitation for the initial scope if multi-index hybrid queries are uncommon, but worth documenting or validating against.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -0,0 +1,274 @@
...
+                tail.add(leg);
+            }
+        }
+        return tail;
+    }
+
+    /** Legs whose Lucene match set is their own top-k (re-running them in the Tail = a redundant ANN pass). */
+    private static boolean isMaterializableLeg(QueryBuilder leg) {
+        String name = leg.getWriteableName();
+        return "knn".equals(name) || "neural".equals(name) || "neural_knn".equals(name);
```
bzhangam: why not use QueryBuilder name? what about neural_sparse?

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    private static boolean needsExecutionTail(SearchSourceBuilder source) {
```
bzhangam: **Question: does the Tail actually help `explain` and `profile`?**

`needsExecutionTail` includes `explain` and `profile` as reasons to include the Tail, but I'm not sure the Tail provides useful output for either:

**`explain`:** The user wants to understand how the fused score was computed (per-leg raw scores → normalization → combination). But the Tail puts the original legs inside a `filter` clause (non-scoring), so the explanation only shows "constant_score, boost=0.716" from the Top — not the per-leg breakdown. The actual fusion happened at the coordinator before the self-erased query was built, so it's not in the Lucene explain tree regardless of the Tail.

**`profile`:** The user wants to see where time was spent. With the Tail, profiling shows the shard re-executing the original legs inside the filter — which is redundant work (they already ran during the MultiSearch fan-out). It profiles the wrong thing: the second execution in the Tail, not the actual leg latency from the MultiSearch.

In both cases the Tail adds query-phase cost (re-executing legs) without providing the information the user is looking for. These might need a fusion-aware solution (e.g., storing the per-leg breakdown in the response) rather than the Tail.

Should `explain` and `profile` be removed from `needsExecutionTail`? Or is there a plan for fusion-aware explain/profile that would use the Tail differently?

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -0,0 +1,274 @@
...
+        MultiSearchResponse.Item[] items = multiSearchResponse.getResponses();
+        SearchHit[][] legHits = groupLegHits(items, legs.size());
+        RankedDocs ranked = computeRankedDocs(legHits, fusion, windowSize);
+        if (ranked.ids().length == 0) {
+            return new MatchNoneQueryBuilder();
+        }
+        boolean topOnly;
+        if (topLevel == false) {
+            topOnly = true; // nested: enclosing filter intersects at the query phase
+        } else if (needsExecutionTail(source) || legsHaveInnerHits(legs)) {
```
bzhangam: Two issues:

**1. `needsExecutionTail` doesn't cover all cases that need the Tail.**

It checks aggregations, explain, profile, highlighter — but total_hits is handled separately by `wantsTotalsBeyondWindow`. If the method is called "needsExecutionTail" you'd expect it to be the single source of truth for "do we need the Tail?". Instead, the decision is split across three conditions. It would be clearer as one method that returns the full decision.

**2. `legsHaveInnerHits` doesn't actually require the Tail.**

`extractInnerHitBuilders` reads from `sourceQueries` (the field on `HybridFusionQuery`), not from the Tail's filter clause. The fetch phase registers inner_hits contexts from `extractInnerHitBuilders` and does its own per-doc lookup. The Tail re-executing the nested query during query phase is wasted work — the fetch phase doesn't use that result.

So `legsHaveInnerHits` being a reason for the Tail seems incorrect. The `sourceQueries` list just needs to be populated (which it always is when we pass the legs to `HybridFusionQuery`), but it doesn't need to be in the executed filter.

src/main/java/org/opensearch/neuralsearch/query/FusionConfigResolver.java
```diff
@@ -0,0 +1,128 @@
...
+ *       (mutually exclusive, not prioritized).</li>
+ *   <li>Inline body block → read directly from the request source (an ad-hoc pipeline, never in cluster state).</li>
+ *   <li>Named param → read that pipeline's config from cluster-state metadata by id.</li>
+ *   <li>Index default ({@code index.search.default_pipeline}) → first concrete index's default wins; a second index
+ *       with a different default collapses to {@code _none}.</li>
+ *   <li>Else {@code _none} → no pipeline (returns null; the caller fails fast).</li>
+ * </ol>
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+final class FusionConfigResolver {
```
bzhangam, line 42: **Questions on `FusionConfigResolver`**

**1. Inline body pipeline is silently broken**

The comment in Step 2 acknowledges that core's `resolvePipeline` drains `searchPipelineSource()` before query rewrite runs, so `fromPipelineConfig({})` returns null and the request fails with "no fusion config found." The user sees an error but it's confusing — they DID specify a pipeline with a normalization processor in the body. The error doesn't explain that inline body pipelines are unsupported for fused mode.

Could we detect this case earlier (e.g., at parse time if `fusion: "pipeline"` coexists with an inline `search_pipeline` block) and throw a clear error like "inline body search pipelines are not supported with fused mode; use a named pipeline or specify the config directly in the `fusion` block"?

**2. Named pipeline doesn't exist — misleading error**

If `?search_pipeline=my-pipeline` references a pipeline that doesn't exist in cluster state, `pipelineConfigById` returns null → the error says "requires a normalization or score-ranker processor in the attached search pipeline." The actual problem is the pipeline doesn't exist, not that it's missing a processor. A more specific error ("pipeline 'my-pipeline' not found") would help debugging.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionQuery.java
martin-gaievski: makes sense, will rename it

src/main/java/org/opensearch/neuralsearch/fusion/rankdocs/FusedDocsScorerQuery.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
martin-gaievski: agree

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: Agree, legSource is a fresh SearchSourceBuilder whose from defaults to 0 at execution, so .from(0) is a no-op;

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: Good catch on the mechanics — knn leg's retrieval depth is governed by its k (collected per shard), not the leg's size, so raising size to window_size doesn't deepen recall; a neural leg with no k/max_distance/min_score defaults to k=10. This is identical to classic hybrid, which runs the sub-queries as one Lucene query and never rewrites k either. Since a stated goal of fused mode is to match classic relevance for the same hit set, I'd keep the same user-responsibility contract (set k ≥ window_size) instead of silently rewriting k. Otherwise that would diverge from classic and wouldn't apply to radial knn (min_score/max_distance have no k). Also I'll add a docs note; I hope that works.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: hardcoded false is intentional, leg-level half of the fail-fast contract (paired with groupLegHits throwing on any leg failure). It matters more here because min_max is computed per-leg over each leg's returned hits, so a leg silently truncated by a down shard (HTTP 200, fewer hits) would shift that leg's min/max and corrupt the fused ranking with no error. Classic hq in contrast runs all sub-queries in one Lucene query over a single shard set. 
I agree that honest end-state is to honor the request's allow_partial_search_results, but  `getter` is a nullable Boolean while the setter takes a primitive boolean, and at coordinator-rewrite time the outer flag isn't yet resolved to the cluster default.  Thus follow-up must propagate flag only when set and otherwise leave the leg unset (a bare allowPartialSearchResults(request.allowPartialSearchResults()) would NPE on the common unset case). 
I'd keep the safe false for this PR and track honoring the flag on the fail-fast thread.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: agree that's a real bug

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: agree on naming, it should be safe to switch to QueryBuilder constant as we do have dependency on all these plugins. 
`neural_sparse` should stay excluded. Materialization is only safe when a leg's full Lucene match set equals its returned top-k (true for HNSW kNN/neural). A neural_sparse leg's match set is every doc containing a query token — far larger than the window — so materializing to the window ids would drop the rest from the Tail and undercount total_hits/aggregations. Plus re-running it is cheap (no HNSW to re-walk); the check keys on getWriteableName() so it also correctly covers the seismic sparse-ANN variant (which keeps a lexical fallback). I'll expand the existing rationale comment at line 242 to make the "match-set == top-k" basis explicit.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: correct, Top is constant_score(ids)^fusedScore with min_max/arithmetic_mean already applied on the coordinator, so the Lucene tree has no fusion math to explain and the Tail is a non-scoring filter; for profile it re-executes legs that already ran in the fan-out. I'll drop explain and profile from needsExecutionTail. Safe: hits and fused scores come from the Top, and total_hits is still covered by wantsTotalsBeyondWindow; aggregations and highlight keep the Tail. Proper fusion-aware explain/profile is scoped to later PR.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: (1) Agreed, I'll consolidate needsExecutionTail / legsHaveInnerHits / wantsTotalsBeyondWindow into one needsTail(...); they're already OR'd so it's a readability win. 

(2) You're right on the mechanics: inner_hits are computed in the fetch phase per parent doc from the registered InnerHitContextBuilder's own query, so re-running the nested leg in the query-phase Tail is wasted. That catch is: sourceQueries isn't always populated — it's the same list that backs the Tail filter and is empty when we go Top-only, so I can't just drop legsHaveInnerHits or extractInnerHitBuilders would register nothing. Clean fix decouples the inner-hit registration source (always carry the leg builders) from the executed Tail (built only for aggs/totals/highlight), with a regression test — and should also cover the nested-fused case, where inner_hits are actually dropped today. Since inner_hits are correct now and this is an optimization, I'd land it as a focused follow-up.

src/main/java/org/opensearch/neuralsearch/query/FusionConfigResolver.java
martin-gaievski: On (1) you're right: core drains our inline search_pipeline map before coordinator rewrite, so we only see {} and emit the misleading "none found." Since searchPipelineSource() is still non-null when a block was attached, I'll turn that into a targeted rewrite-time error; genuinely supporting an inline body needs the companion core fix (deep-copy the config in resolvePipeline), which I'll track as a follow-up. 

On (2), though, a genuinely missing named or index-default pipeline id doesn't reach this code: core's resolvePipeline throws "Pipeline [x] is not defined" (before rewriteAndFetch), so the user already gets a clear error. Our generic message only fires when the pipeline exists but has no fusion processor (or the drained-inline case), so a "not found" branch would be unreachable. I'll instead sharpen the wording to say the resolved pipeline has no normalization/score-ranker processor.

src/main/java/org/opensearch/neuralsearch/fusion/CoordinatorScoreFusion.java
martin-gaievski: it's the same per-leg normalization seam we discussed for @vibrantvarun two-phase computeLegStats/normalize(score,stats) idea — let's converge on one abstraction rather than land two. Your single transform(List<Float>)→List<Float> is the better target: it generalizes to rank-based RRF (sort the leg, emit 1/(k+rank) by position), whereas the stats-split forces a per-score phase-2 that RRF can't satisfy. 

This PR only supports min_max + arithmetic_mean (requireSupportedTechniques rejects the rest), so there's one technique today, I'd abstract in PR that lands the second scalar technique, keeping ScoreCombinationTechnique as the combine half.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
bzhangam: If user explicitly defines it then we can pass that down to the leg. If not defined and we also not define it for the leg search it should auto pick up the cluster default. Unless we think hardcode it as false a better UX for hybrid query? But I feel we should follow the default behavior in OpenSearch to not hardcode it as false unless user explicitly set it.

src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
```diff
@@ -32,6 +32,7 @@ public final class MinClusterVersionUtil {
     public static final Version MINIMAL_SUPPORTED_VERSION_METRICS_STATS = Version.V_3_3_0;
     private static final Version MINIMAL_SUPPORTED_VERSION_NEURAL_KNN_QUERY_BUILDER = Version.V_3_0_0;
     private static final Version MINIMAL_SUPPORTED_VERSION_AGENTIC_EMBEDDING_MODEL_ID = Version.V_3_6_0;
+    public static final Version MINIMAL_SUPPORTED_VERSION_FUSED_MODE_IN_HYBRID_QUERY = Version.V_3_8_0;
```
heemin32, line 35: V_3_9_0 is available now

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: we cannot match existing opensearch query behavior to 100%, that's why I incline towards do it defensively in this PR. as per documentation setting is in effect if "request times out or a shard fails." For resolver that "shard fails" is not atomic state, same shard can fail for one leg/sub-query and succeed for another, so from top level query it looks like "partial shard results", not "this shard failed". In contrast existing hybrid query in such case will fail the whole shard.  
I suggest we keep it defensive false for now and I address that in later PRs, I'm planning one specifically for advanced query features like aggregations etc.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: discussed offline, with latest code we are breaking default behavior, which is allow partial results (`true` setting, https://docs.opensearch.org/latest/install-and-configure/configuring-opensearch/search-settings/). If `false` is hardcoded that will fail the whole top query independently of actual setting, including default. Thus I'm reverting to initial version so we keep the default behavior. @vibrantvarun  that something we discussed in https://github.com/opensearch-project/neural-search/pull/1933#discussion_r3752662934

src/main/java/org/opensearch/neuralsearch/common/MinClusterVersionUtil.java
martin-gaievski: 3.8 is feature branch baseline, I keep it before all dependencies/plugins up-vert to 3.9 to avoid build failures and instabilities with not yet released changes in main. Control gate is the final merge from feature branch to main after all development is completed, then we must update the flag to latest OS version.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: Added validation that we are running query against a single index: https://github.com/martin-gaievski/neural-search/blob/feature/fusion-hybrid-query/src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java#L586

bzhangam (approved):

PR #1942, opened 2026-08-13
[RRF] Compute rank scores with exact integer arithmetic instead of a BigDecimal per document

### LINKED ISSUE #1941: [PERF] RRF score calculation allocates a BigDecimal per document; an exact integer form is bit-identical and allocation-free
While refactoring the RRF code I noticed this calculation looked wasteful, since it allocates a throwaway `BigDecimal` on every call
([`RRFNormalizationTechnique.java#L221-L223`](https://github.com/opensearch-project/neural-search/blob/746c7df5e0c305c341565ce0f62346704f08b618/src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java#L221-L223)):

```java
private float calculateNormalizedScore(int position) {
    return BigDecimal.ONE.divide(BigDecimal.valueOf(rankConstant + position + 1), 10, RoundingMode.HALF_UP).floatValue();
}
```

It runs once per document, per subquery, per shard, on the coordinator node — so `shards × subqueries × hits` times per search request. That is ~600 calls for a shallow search (20 shards × 10 hits × 3 subqueries) and ~60,000 for a deep one (20 shards × 1000 hits × 3 subqueries), doubled when `explain` is enabled, since `explain(...)` re-runs the same `processTopDocs` traversal.

I initially assumed the fix would be either a precomputed table or plain `double`/`float` arithmetic, both of which change the output. But measuring it and digging into `BigDecimal`'s internals showed the whole expression reduces to `HALF_UP(10^10 / m) / 10^10` for a single integer `m = rankConstant + position + 1`, and both `10^10` and `2 × 10^10` fit in a `long` — so the rounding is expressible in exact integer arithmetic. The current code is **bit-for-bit identical** to:

```java
private static final long SCORE_SCALE = 10_000_000_000L;   // 10^10, the scale-10 denominator

private float calculateNormalizedScore(int position) {
    long m = (long) rankConstant + position + 1;
    long numerator = (2 * SCORE_SCALE + m) / (2 * m);      // HALF_UP(10^10 / m), exactly
    return (float) ((double) numerator / (double) SCORE_SCALE);
}
```

`(2·10^10 + m) / (2m)` is the standard exact integer form of round-half-up division. The quotient is at most `5 × 10^9 < 2^53`, and `10^10` is exact as a `double`, so the single narrowing to `float` reproduces exactly what `BigDecimal.floatValue()` produces. No allocation, and primitive integer and floating-point ops the JIT handles well.

### Measured improvement

JIT-warm, Corretto 21, allocation via `ThreadMXBean#getThreadAllocatedBytes`:

| domain of `m` | today | proposed |
|---|---|---|
| `[61, 260]` — default `rank_constant`, shallow results | 54 – 56 ns, 302 – 322 B/call | 4.9 ns, **0 B** |
| `[61, 20060]` — default `rank_constant`, deep results | 8.5 ns, 85 B/call | 2.6 ns, **0 B** |

Per request that is roughly **180 KB of garbage eliminated for a shallow search and ~5 MB for a deep one**, plus ~11× lower latency in the shallow case.

The saving is much larger at small `m` because `BigDecimal.floatValue()` has a cost cliff: when the scale-10 unscaled value exceeds `2^24 = 16777216` it takes a substantially more expensive conversion path, and `10^10 / 2^24 ≈ 596`, so every `m ≤ 596` pays it. Measured on adjacent windows: `m ∈ [560, 596]` costs 63.9 ns / 271 B, while `m ∈ [597, 633]` costs 20.5 ns / 40 B. **With the default `rank_constant` of 60 this means ranks 0–535 are all in the expensive band** — i.e. the entire result set for typical `size` values, so the default configuration is the worst case. This threshold is an internal of `BigDecimal` and is specified nowhere; the proposed fix removes any dependence on it. Above the cliff the expression still allocates ~40 B/call, which is where most of the aggregate garbage in deep searches comes from.

This is not a per-segment scoring path, so it is not urgent. It is worth doing because the fix is free, simpler than what is there now, and provably changes no output.

### Verification

Bit-identical over the **entire reachable domain**, not a sample. Since `rankConstant` and `position` are both `int`, every reachable `m` lies in `[2, Integer.MAX_VALUE]` (`m ≥ 2` because `rank_constant ≥ 1` and `position ≥ 0`). All **2,147,483,646** values were compared via `Float.floatToIntBits`: **zero mismatches on JDK 17, 21 and 26**, extremes included. Worth pinning a representative subset as a unit test so it cannot silently drift.

⚠️ **The `double` narrowing is load-bearing.** The apparently equivalent `(float) numerator / 1.0e10f` is *not* bit-identical — it mismatches for 167 values in the same domain, first at `m = 3`, because it rounds twice in binary. Existing tests compare scores at `DELTA_FOR_SCORE_ASSERTION = 0.001f`, so that regression would pass CI silently. The unit test should pin the `double` form
specifically.

This change deliberately preserves the existing scale-10 quantization, including the places where it is arguably wrong (e.g. at `m > 10⁴` ten decimal places is coarser than `float`'s ~7 significant digits). That is a separate, user-visible decision; happy to file it separately if maintainers want it considered.

### Method and caveats

Standalone single-file probes, JIT-warmed over the realistic domain, accumulating into a sink to prevent dead-code elimination. These are microbenchmarks isolated from the surrounding `PriorityQueue` sort, not JMH with fork isolation, so treat the ns figures as indicative — JDK 26 measured the `[61, 260]` window as high as 115 ns/call against JDK 21's 54 ns, which is more run-to-run variance than I would defend to a decimal place. The direction and order of magnitude were stable across every run; the allocation figures and the bit-identity result are exact. Probe sources available on request.

dbwiddis:
### Description

`RRFNormalizationTechnique#calculateNormalizedScore` allocated a throwaway `BigDecimal` per document, per subquery, per shard on the coordinator node — roughly 600 allocations for a shallow hybrid search (20 shards × 10 hits × 3 subqueries) and 60,000 for a deep one, doubled when `explain` is enabled.

The expression reduces to `HALF_UP(10^10 / d) / 10^10` for a single integer denominator `d = rankConstant + position + 1`. Both `10^10` and `2 × 10^10` fit in a `long`, so rounding HALF_UP to scale 10 is exactly expressible as the integer division `(2 × 10^10 + d) / (2 × d)`, with no `BigDecimal` and no allocation:

```java
private float calculateNormalizedScore(int position) {
    long denominator = (long) rankConstant + position + 1;
    long numerator = (2 * SCORE_SCALE_L + denominator) / (2 * denominator);
    return (float) (numerator / SCORE_SCALE_D);
}
```

**This is behavior-preserving, not an approximation.** The new form is bit-identical to the old one for every reachable denominator. Since `rankConstant` and `position` are both `int`, every reachable `d` lies in `[2, Integer.MAX_VALUE]`; all **2,147,483,646** values were compared via `Float.floatToIntBits` with **zero mismatches on JDK 17, 21 and 26**. No existing test or expected score changes.

Measured JIT-warm on Corretto 21, allocation via `ThreadMXBean#getThreadAllocatedBytes`:

| domain of `d` | before | after |
|---|---|---|
| `[61, 260]` — default `rank_constant`, shallow results | 54 – 56 ns, 302 – 322 B/call | **4.9 ns, 0 B** |
| `[61, 20060]` — default `rank_constant`, deep results | 8.5 ns, 85 B/call | **2.6 ns, 0 B** |

Per request that is roughly 180 KB of garbage eliminated for a shallow search and ~5 MB for a deep one.

The gain is largest at small `d` because `BigDecimal.floatValue()` takes a more expensive conversion path once the scale-10 numerator exceeds `2^24`, which is every `d ≤ 596` — so with the default `rank_constant` of 60, ranks 0 through 535 all pay it, i.e. the entire result set for typical `size` values. That threshold is a `BigDecimal` internal specified nowhere; this change removes any dependence on it.

Two notes for reviewers:

- **The final division is deliberately performed in `double`.** Narrowing the numerator to `float` first and dividing by `1.0e10f` rounds twice and is *not* bit-identical — it differs by one ULP for 167 denominators, starting at `d = 3`. The existing tests compare scores within `DELTA_FOR_ASSERTION` of `0.001f` and would not catch that, so `testNormalize_whenScoresComputed_thenBitIdenticalToBigDecimalReference` asserts on raw float bits instead. I verified the test fails on that exact substitution before committing.
- **The scale-10 quantization is preserved as-is**, including where it is arguably wrong (past `d ≈ 10⁴`, ten decimal places is coarser than `float`'s ~7 significant digits, so it discards real precision). Changing that would alter user-visible scores and is a separate decision; this PR deliberately changes no output.

Benchmark caveat: these were standalone single-file probes, JIT-warmed and accumulating into a sink to prevent dead-code elimination, but isolated from the surrounding `PriorityQueue` sort and not JMH with fork isolation — so treat the ns figures as indicative rather than exact. JDK 26 measured the `[61, 260]` window as high as 115 ns/call against JDK 21's 54 ns. The direction and order of magnitude were stable across every run; the allocation figures and the bit-identity result are exact. Probe sources available on request.

### Related Issues
Resolves #1941

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,4 +23,4 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Refactoring
 * [SemanticHighlighter] Traverse the query tree with a QueryBuilderVisitor instead of a "manual" walk ([#1915](https://github.com/opensearch-project/neural-search/pull/1915))
-* [RRF] Compute rank scores with exact integer arithmetic instead of allocating a BigDecimal per document ([#1941](https://github.com/opensearch-project/neural-search/issues/1941))
+* [RRF] Compute rank scores with exact integer arithmetic instead of allocating a BigDecimal per document ([#1942](https://github.com/opensearch-project/neural-search/pull/1942))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,3 +23,4 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Refactoring
 * [SemanticHighlighter] Traverse the query tree with a QueryBuilderVisitor instead of a "manual" walk ([#1915](https://github.com/opensearch-project/neural-search/pull/1915))
+* [RRF] Compute rank scores with exact integer arithmetic instead of allocating a BigDecimal per document ([#1941](https://github.com/opensearch-project/neural-search/issues/1941))
--- a/src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/normalization/RRFNormalizationTechnique.java
@@ -4,8 +4,6 @@
  */
 package org.opensearch.neuralsearch.processor.normalization;
 
-import java.math.BigDecimal;
-import java.math.RoundingMode;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
@@ -49,6 +47,9 @@ public class RRFNormalizationTechnique implements ScoreNormalizationTechnique, E
     private static final int MIN_RANK_CONSTANT = 1;
     private static final int MAX_RANK_CONSTANT = 10_000;
     private static final Range<Integer> RANK_CONSTANT_RANGE = Range.of(MIN_RANK_CONSTANT, MAX_RANK_CONSTANT);
+    // Rank scores are quantized to 10 decimal places, i.e. expressed as an integer numerator over 10^10.
+    private static final long SCORE_SCALE_L = 10_000_000_000L;
+    private static final double SCORE_SCALE_D = 1.0e10;
     @ToString.Include
     private final int rankConstant;
     // Comparator to compare ShardResultPerSubQuery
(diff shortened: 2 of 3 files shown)


martin-gaievski (approved):
looks good, thanks

PR #1943, opened 2026-08-13
In-query fusion advanced query features

### LINKED ISSUE #1930: [RFC] In-query fusion for hybrid search

martin-gaievski:
### Description
Adding advanced query features carry-over follow-ups from #1933.

- New feature support/improvements:
  - nested composition, Tail decision is now depth-independent
     nested leg is dispatched with `search_pipeline=_none`, so re-resolving the fusion config from the leg request found nothing and a pipeline-configured index returned a 500 for any nested fused query. the resolved spec is now projected onto the legs instead of
  re-resolved, and the "no fusion config" message names what the user has to do
  - inner_hits decoupled from the executed Tail
    split `sourceQueries` into `tailQueries` (executed) and `innerHitsQueries` (registered only). Top-only query now returns leg inner_hits without re-running the legs, and previously dropped nested-fused inner_hits now work
  - added setting to limit depth of nested hq to prevent resource over-consumption
  - Point-in-time (PIT) passthrough
    parity with PIT support with regular queries - user PIT reaches every leg, so all legs and the round-2 query read one immutable view instead of N+1 reader instants
  - pluggable normalization abstraction (ScalarNormalizer) - new low level abstraction for scalar normalizers 
  - window_size validated against indices.query.bool.max_clause_count parse-time 400 instead of a per-shard failure
- Cary over/feedback from foundational PR #1933 
  - allow_partial_search_results properly handled
    explicit request value propagates to every leg; unset stays unset so legs resolve cluster default like a normal search. degraded leg shifts its own min/max, so adding affected legs names to `Warning` header. wholly-failed leg remains a hard failure.
  - multi-index correctness: index-aware document identity
     `_id`-collision bug, documents are now keyed by `_index` + `_id`
 
### Related Issues
Resolves #1930

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/fusion/CoordinatorScoreFusion.java
+++ b/src/main/java/org/opensearch/neuralsearch/fusion/CoordinatorScoreFusion.java
@@ -4,6 +4,7 @@
  */
 package org.opensearch.neuralsearch.fusion;
 
+import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -34,43 +35,59 @@ import lombok.NoArgsConstructor;
  *       toward the denominator (its {@code score >= 0.0} participation rule), exactly as classic does.</li>
  * </ul>
  *
- * <p>Current scope: {@code min_max} normalization + the supplied combination technique (arithmetic_mean). Other techniques
- * (z_score, l2, RRF) join this path in a later change. This entry is not reachable from a live request yet — the
- * coordinator rewrite wiring lands with the execution path.
+ * <p>The normalization step is pluggable via {@link ScalarNormalizer} (resolved by name through
+ * {@link ScalarNormalizerFactory}); the combination step is the caller-supplied {@link ScoreCombinationTechnique}. This
+ * class owns only the shape-level work that is identical for every technique: per-leg normalization dispatch, the doc
+ * union, and building each doc's per-leg input array. Adding a technique therefore touches neither this class nor the
+ * orchestrator. Current fused-mode scope is {@code min_max} + arithmetic_mean; the caller gates the rest at rewrite.
+ *
+ * <p>Document keys are treated as <b>opaque strings</b> throughout — this class never parses or builds them — so the
+ * caller can change its document identity scheme (e.g. to disambiguate same-{@code _id} docs across indices) without
+ * touching fusion or any normalizer.
  */
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 public final class CoordinatorScoreFusion {
 
     /**
-     * Fuse legs with {@code min_max} normalization followed by the given combination technique.
+     * Fuse legs with {@code min_max} normalization followed by the given combination technique. Convenience wrapper over
+     * {@link #fuse(List, ScalarNormalizer, ScoreCombinationTechnique)}; this is the pairing the classic-vs-fused
+     * differential test pins.
      *
-     * @param legRawScores        one entry per leg, each an {@code _id -> raw score} map of that leg's hits (order is
+     * @param legRawScores        one entry per leg, each a {@code key -> raw score} map of that leg's hits (order is
      *                            the leg order; a doc absent from a leg's map did not match that leg)
      * @param combinationTechnique the same combination technique classic would use (e.g. arithmetic_mean with weights)
-     * @return {@code _id -> fused score} for the union of all legs' docs
+     * @return {@code key -> fused score} for the union of all legs' docs
      */
     public static Map<String, Float> fuseMinMax(
         final List<Map<String, Float>> legRawScores,
         final ScoreCombinationTechnique combinationTechnique
     ) {
+        return fuse(legRawScores, MinMaxScalarNormalizer.INSTANCE, combinationTechnique);
+    }
+
+    /**
+     * Fuse legs: normalize each leg with {@code normalizer}, then combine across legs per doc.
+     *
+     * @param legRawScores        one entry per leg, each a {@code key -> raw score} map of that leg's hits
+     * @param normalizer          per-leg normalization step (e.g. {@code min_max})
+     * @param combinationTechnique cross-leg combination step (e.g. arithmetic_mean with weights)
+     * @return {@code key -> fused score} for the union of all legs' docs
+     */
+    public static Map<String, Float> fuse(
+        final List<Map<String, Float>> legRawScores,
+        final ScalarNormalizer normalizer,
+        final ScoreCombinationTechnique combinationTechnique
+    ) {
         final int legCount = legRawScores.size();
 
-        // Per-leg min/max, seeded exactly as classic (getMinScores/getMaxScores)
-        // picking array over List as it's slightly faster, we know size and access by index has identical behavior
-        final float[] minPerLeg = new float[legCount];
-        final float[] maxPerLeg = new float[legCount];
-        for (int leg = 0; leg < legCount; leg++) {
-            float min = Float.MAX_VALUE;
-            float max = Float.MIN_VALUE;
-            for (float raw : legRawScores.get(leg).values()) {
-                min = Math.min(min, raw);
-                max = Math.max(max, raw);
-            }
-            minPerLeg[leg] = min;
-            maxPerLeg[leg] = max;
+        // Normalize each leg independently. A leg's map is already the merged across-shard set, so the normalizer sees
+        // every value it needs to compute its own statistics (min/max here) with no cross-shard merging.
+        final List<Map<String, Float>> normalizedPerLeg = new ArrayList<>(legCount);
+        for (Map<String, Float> leg : legRawScores) {
+            normalizedPerLeg.add(normalizer.normalizeLeg(leg));
         }
 
-        // Union of ids across legs, preserving first-seen order for deterministic output.
+        // Union of keys across legs, preserving first-seen order for deterministic output.
         final Set<String> allIds = new LinkedHashSet<>();
         for (Map<String, Float> leg : legRawScores) {
             allIds.addAll(leg.keySet());
@@ -81,9 +98,9 @@ public final class CoordinatorScoreFusion {
             // float[legCount] initialized to 0.0; only matching legs are filled (mirrors classic per-doc array).
             final float[] perLegNormalized = new float[legCount];
             for (int leg = 0; leg < legCount; leg++) {
-                Float raw = legRawScores.get(leg).get(id);
-                if (raw != null) {
-                    perLegNormalized[leg] = MinMaxScoreNormalizer.normalizeSingleScore(raw, minPerLeg[leg], maxPerLeg[leg]);
+                Float normalized = normalizedPerLeg.get(leg).get(id);
+                if (normalized != null) {
+                    perLegNormalized[leg] = normalized;
                 }
             }
             fused.put(id, combinationTechnique.combine(perLegNormalized));
--- a/src/main/java/org/opensearch/neuralsearch/fusion/CoordinatorScoreFusion.java
+++ b/src/main/java/org/opensearch/neuralsearch/fusion/CoordinatorScoreFusion.java
@@ -18,11 +18,11 @@ import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
 
 /**
- * Coordinator-side score fusion for the resolver (fused) mode, over an {@code _id}-keyed per-leg score view (shape
- * that coordinator has after fanning legs out as a {@code MultiSearch}). It deliberately reuses the same relevance math
- * as classic shard-side hybrid — {@link MinMaxScoreNormalizer} for normalization and the caller-supplied
+ * Coordinator-side score fusion for the resolver (fused) mode, over a per-leg score view keyed by document key (the
+ * shape the coordinator has after fanning legs out as a {@code MultiSearch}). It deliberately reuses the same relevance
+ * math as classic shard-side hybrid — {@link MinMaxScoreNormalizer} for normalization and the caller-supplied
  * {@link ScoreCombinationTechnique#combine(float[])} for combination — so that, for an identical hit set, this path and
- * classic produce identical fused scores to float precision. Only the data shape differs ({@code _id} map here vs.
+ * classic produce identical fused scores to float precision. Only the data shape differs (a key→score map here vs.
  * {@code CompoundTopDocs} there), never the arithmetic.
  *
  * <p>Two classic behaviors are reproduced exactly:
@@ -36,7 +36,7 @@ import lombok.NoArgsConstructor;
  * </ul>
  *
  * <p>The normalization step is pluggable via {@link ScalarNormalizer} (resolved by name through
- * {@link ScalarNormalizerFactory}); the combination step is the caller-supplied {@link ScoreCombinationTechnique}. This
+ * {@link ScalarNormalizers}); the combination step is the caller-supplied {@link ScoreCombinationTechnique}. This
  * class owns only the shape-level work that is identical for every technique: per-leg normalization dispatch, the doc
  * union, and building each doc's per-leg input array. Adding a technique therefore touches neither this class nor the
  * orchestrator. Current fused-mode scope is {@code min_max} + arithmetic_mean; the caller gates the rest at rewrite.
@@ -82,28 +82,28 @@ public final class CoordinatorScoreFusion {
 
         // Normalize each leg independently. A leg's map is already the merged across-shard set, so the normalizer sees
         // every value it needs to compute its own statistics (min/max here) with no cross-shard merging.
-        final List<Map<String, Float>> normalizedPerLeg = new ArrayList<>(legCount);
-        for (Map<String, Float> leg : legRawScores) {
-            normalizedPerLeg.add(normalizer.normalizeLeg(leg));
+        final List<Map<String, Float>> legNormalizedScores = new ArrayList<>(legCount);
+        for (Map<String, Float> legScores : legRawScores) {
+            legNormalizedScores.add(normalizer.normalizeLeg(legScores));
         }
 
         // Union of keys across legs, preserving first-seen order for deterministic output.
-        final Set<String> allIds = new LinkedHashSet<>();
-        for (Map<String, Float> leg : legRawScores) {
-            allIds.addAll(leg.keySet());
+        final Set<String> allKeys = new LinkedHashSet<>();
+        for (Map<String, Float> legScores : legRawScores) {
+            allKeys.addAll(legScores.keySet());
         }
 
         final Map<String, Float> fused = new LinkedHashMap<>();
-        for (String id : allIds) {
+        for (String key : allKeys) {
             // float[legCount] initialized to 0.0; only matching legs are filled (mirrors classic per-doc array).
-            final float[] perLegNormalized = new float[legCount];
+            final float[] docScoresPerLeg = new float[legCount];
             for (int leg = 0; leg < legCount; leg++) {
-                Float normalized = normalizedPerLeg.get(leg).get(id);
-                if (normalized != null) {
-                    perLegNormalized[leg] = normalized;
+                Float normalizedScore = legNormalizedScores.get(leg).get(key);
+                if (normalizedScore != null) {
+                    docScoresPerLeg[leg] = normalizedScore;
                 }
             }
-            fused.put(id, combinationTechnique.combine(perLegNormalized));
+            fused.put(key, combinationTechnique.combine(docScoresPerLeg));
         }
         return fused;
     }
--- a/src/main/java/org/opensearch/neuralsearch/fusion/ScalarNormalizerFactory.java
+++ b/src/main/java/org/opensearch/neuralsearch/fusion/ScalarNormalizerFactory.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.fusion;
+
+import java.util.Locale;
+import java.util.Map;
+import java.util.Objects;
+
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+
+/**
+ * Resolves a coordinator-side {@link ScalarNormalizer} by technique name — the extension point for widening fused-mode
+ * normalization support. Adding a technique is a new {@link ScalarNormalizer} plus one entry here; neither
+ * {@link CoordinatorScoreFusion} nor the orchestrator changes.
+ *
+ * <p>Only {@code min_max} is registered today, matching the fused-mode scope gate in
+ * {@code HybridQueryBuilder#requireSupportedTechniques} (which rejects other techniques earlier, at rewrite). This
+ * factory's throw is therefore a defense-in-depth backstop, not the user-facing validation.
+ *
+ * <p>When adding techniques, two things need a decision that this seam intentionally leaves open:
+ * <ul>
+ *   <li><b>{@code l2}/{@code z_score}</b> drop in as plain {@link ScalarNormalizer}s. But note
+ *       {@link CoordinatorScoreFusion} encodes "leg did not match this doc" as a {@code 0.0} slot, and {@code l2} can
+ *       legitimately normalize a real score to {@code 0.0}. That is harmless for arithmetic-mean (which counts
+ *       {@code >= 0.0} slots, matching classic) but ambiguous for geometric/harmonic mean, which skip {@code <= 0}
+ *       scores. Wiring those combiners likely needs an explicit presence mask rather than the {@code 0.0} sentinel —
+ *       changing the sentinel is a deliberate parity decision, since the current behavior is what the classic-vs-fused
+ *       differential test pins.</li>
+ *   <li><b>RRF</b> is rank-based and is modelled as {@code combination=rrf, normalization=none}, so it does not arrive
+ *       here under a normalization name. It fits the {@link ScalarNormalizer} shape (sort the leg, emit
+ *       {@code 1/(rank_constant + rank + 1)} by position) but needs its own routing plus a rank_constant parameter, and
+ *       the classic {@code (doc, shardId)} rank tie-break has no equivalent over the coordinator's already-merged view —
+ *       so a tie-break must be defined explicitly.</li>
+ * </ul>
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class ScalarNormalizerFactory {
+
+    private static final Map<String, ScalarNormalizer> NORMALIZERS = Map.of(
+        MinMaxScalarNormalizer.INSTANCE.techniqueName(),
+        MinMaxScalarNormalizer.INSTANCE
+    );
+
+    /**
+     * @param techniqueName normalization technique name from the resolved fusion config
+     * @return the normalizer for that technique
+     * @throws IllegalArgumentException when the technique has no coordinator-side implementation yet
+     */
+    public static ScalarNormalizer create(final String techniqueName) {
+        ScalarNormalizer normalizer = Objects.isNull(techniqueName) ? null : NORMALIZERS.get(techniqueName);
+        if (Objects.isNull(normalizer)) {
+            throw new IllegalArgumentException(
+                String.format(
+                    Locale.ROOT,
+                    "normalization technique [%s] is not supported in fused mode; supported: %s",
+                    techniqueName,
+                    NORMALIZERS.keySet()
+                )
+            );
+        }
+        return normalizer;
+    }
+}
--- a/src/main/java/org/opensearch/neuralsearch/query/CandidateScope.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/CandidateScope.java
@@ -141,9 +141,10 @@ final class CandidateScope {
             table,
             SEARCH_REQUEST,
             "scroll",
-            Disposition.NOT_PROPAGATED,
-            "round 1 is a one-shot candidate query; the scroll context belongs to the user request, which pages through "
-                + "Top then Tail the same way a size beyond the window already does"
+            Disposition.REJECTED,
+            "a scroll pages one reader snapshot to exhaustion, and that snapshot is round 2's alone: the legs that chose "
+                + "the window ran as one-shot searches against their own reader instants and are never re-run. Classic "
+                + "hybrid rejects scroll too — use point_in_time instead, which every leg and round 2 do share"
         );
         put(
             table,
@@ -207,7 +208,9 @@ final class CandidateScope {
             SEARCH_SOURCE,
             "sliceBuilder",
             Disposition.PROPAGATED,
-            "round 2 returns only the slice, so unsliced legs would fill the window with documents outside it"
+            "round 2 returns only the slice, so unsliced legs would fill the window with documents outside it and leave "
+                + "the slice an arbitrary fraction of it. A slice is legal only alongside a point_in_time or a scroll, "
+                + "and scroll is rejected, so the reachable shape is a slice over a PIT — inherited together with it"
         );
         put(
             table,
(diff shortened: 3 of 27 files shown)


src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -138,32 +169,85 @@ private static SearchHit[][] groupLegHits(MultiSearchResponse.Item[] items, int
...
     /**
-     * Fuse via the shared {@link CoordinatorScoreFusion} core (min_max + arithmetic_mean), then rank by fused score and
-     * cut to the window. Converts the coordinator's {@code SearchHit[][]} view into the {@code _id}-keyed per-leg maps
-     * the shared core consumes; a leg that matched nothing contributes an empty map (groupLegHits fails fast on failures,
-     * so every slot is non-null).
+     * Surface partially-degraded legs as a response {@code Warning} header. Uses the same mechanism as deprecation
+     * warnings, emitted from the coordinator rewrite's async callback so it rides the request's thread context onto the
+     * response.
+     */
+    private static void warnOnDegradedLegs(final List<Integer> degradedLegs) {
```
owaiskazi19, line 200: This is tested through UT for now but would be called with msearch async callback, can we add a IT with a shard failure to test this? That would help to understand that the header actually rides the request ThreadContext onto the Search response

src/test/java/org/opensearch/neuralsearch/query/HybridFusionOrchestratorTests.java
```diff
@@ -94,12 +117,53 @@ public void testBuildLegMultiSearch_perLegSourceShape() {
...
             assertFalse(source.fetchSource().fetchSource());
             assertEquals(SearchPipelineService.NOOP_PIPELINE_ID, leg.pipeline());
-            // Legs don't override allow_partial_search_results — left unset so each resolves the cluster default (true)
-            // at execution, like a normal search.
+            // Unset on the request → left unset on the leg so each resolves the cluster default (true) at execution.
             assertNull(leg.allowPartialSearchResults());
         }
     }
 
+    public void testBuildLegMultiSearch_whenAllowPartialExplicitlySet_thenPropagatedToLegs() {
```
owaiskazi19: This test prove false is copied onto each leg and that groupLegHits throws on item.isFailure(), but nothing ties them together. Consider a unit test (aMultiSearchResponse with one Item.isFailure()) asserting the whole-request failure under effective allow_partial=false, so both halves of claim 5 are pinned

src/main/java/org/opensearch/neuralsearch/query/HybridFusionQueryBuilder.java
```diff
@@ -33,55 +34,119 @@
...
+ * favour: core's {@code ExpandSearchPhase} re-runs {@code source().query()} per group, which here is this self-erased
+ * query, so every expanded member is scored by its own Top clause and therefore carries <b>its own fused score</b> —
+ * on the same scale as the group representative, and equal to the score it would receive in the ungrouped fused search.
+ * Classic instead re-runs the real sub-queries and reports their <b>raw, un-normalized</b> scores, which are on a
+ * different scale from the representative's normalized score (e.g. representative {@code 1.0} beside members
+ * {@code 198.0}). Leg-level {@code inner_hits} (a {@code nested}/{@code has_child} sub-query with its own
+ * {@code inner_hits} block) match classic exactly, because core re-runs the inner query there rather than the parent.
+ *
+ * <p>This query is created internally by the coordinator self-erase and is never parseable from a search request. Its
+ * wire form needs no version gate: the whole query type is new in the same version that introduced fused mode, and a
```
owaiskazi19, line 91: The wire gate protects the classic path, but a fused request during a rolling upgrade will hard-fail once it reaches a pre-3.8.0 node. The javadoc frames this as benign, but to a user it's a hard runtime error. Please state explicitly

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -138,32 +169,85 @@ private static SearchHit[][] groupLegHits(MultiSearchResponse.Item[] items, int
...
+        // no change here. The caller already rejected techniques outside the current scope at rewrite.
+        ScalarNormalizer normalizer = ScalarNormalizerFactory.create(fusion.normalizationTechnique());
+        Map<String, Float> combined = CoordinatorScoreFusion.fuse(legRawScores, normalizer, combination);
+        return toRankedDocs(combined, identityByKey, windowSize);
+    }
+
+    /**
+     * Fusion key for a hit: {@code _index}-qualified when the hit carries an index, else the bare {@code _id}.
+     *
+     * <p>Limitation in custom routing. {@code _index} + {@code _id} is not a total identity when custom routing is
```
owaiskazi19: Can we have a followup issue on this?

src/main/java/org/opensearch/neuralsearch/query/HybridFusionQueryBuilder.java
```diff
@@ -33,55 +34,119 @@
...
         this.ids = ids;
+        this.indices = indices;
         this.scores = scores;
-        this.sourceQueries = Objects.isNull(sourceQueries) ? new ArrayList<>() : sourceQueries;
+        this.tailQueries = Objects.isNull(tailQueries) ? new ArrayList<>() : tailQueries;
+        this.innerHitsQueries = Objects.isNull(innerHitsQueries) ? new ArrayList<>() : innerHitsQueries;
+    }
+
+    /** Single-index convenience: no index qualification, and the Tail legs are also the inner_hits source. */
+    public HybridFusionQueryBuilder(String[] ids, float[] scores, List<QueryBuilder> tailQueries) {
```
owaiskazi19: This constructor re-couples inner_hits to the Tail which the exact thing this PR decouples. Only tests use it; consider dropping it

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFusedModeNestedIT.java
```diff
@@ -0,0 +1,280 @@
...
+            + "\"combination\":{\"technique\":\"arithmetic_mean\",\"parameters\":{\"weights\":[0.5,0.5]}}},"
+            + "\"queries\":["
+            + fusedHybrid()
+            + ","
+            + leg()
+            + "]}}}";
+
+        List<String> ids = hitIds(searchRaw(body, 20));
+
+        assertFalse("fusion-of-fusion must return hits", ids.isEmpty());
```
owaiskazi19, line 248: We should assert total hits/aggs as well on fused level here

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFusedModeNestedIT.java
```diff
@@ -0,0 +1,280 @@
...
+    private Map<String, Integer> grpCounts(Map<String, Object> resp) {
+        Map<String, Integer> out = new LinkedHashMap<>();
+        for (String id : hitIds(resp)) {
+            out.merge(Integer.parseInt(id) <= WINDOW_LAST_ID ? "A" : "B", 1, Integer::sum);
+        }
+        return out;
+    }
+
+    @SuppressWarnings("unchecked")
+    private long totalHits(Map<String, Object> resp) {
```
owaiskazi19, line 417: Common across all IT, can be extracted to different file

owaiskazi19 (commented):
Nice improvement

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFusedModeNestedIT.java
martin-gaievski: ack

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFusedModeNestedIT.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/query/HybridFusionQueryBuilder.java
martin-gaievski: agree, will delete it and migrate ~9 test call sites to 5-arg form (pass List.of()/explicit inner_hits), should be straightforward

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: ack, will create a tracking issue

src/main/java/org/opensearch/neuralsearch/query/HybridFusionQueryBuilder.java
martin-gaievski: ack, will update the java docs

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: I'll check what can be done with IT, if that's feasible I'll add the test with a failure on one leg only

src/test/java/org/opensearch/neuralsearch/query/HybridFusionOrchestratorTests.java
martin-gaievski: this one can be addressed together with previous comment, if one shard-failure IT is feasible, we run it twice, allow_partial=true => 200 + Warning (pins comment 1), allow_partial=false => request fails

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: Created https://github.com/opensearch-project/neural-search/issues/1952 to track this

src/main/java/org/opensearch/neuralsearch/fusion/CoordinatorScoreFusion.java
```diff
@@ -81,9 +98,9 @@ public static Map<String, Float> fuseMinMax(
             // float[legCount] initialized to 0.0; only matching legs are filled (mirrors classic per-doc array).
             final float[] perLegNormalized = new float[legCount];
             for (int leg = 0; leg < legCount; leg++) {
-                Float raw = legRawScores.get(leg).get(id);
-                if (raw != null) {
-                    perLegNormalized[leg] = MinMaxScoreNormalizer.normalizeSingleScore(raw, minPerLeg[leg], maxPerLeg[leg]);
+                Float normalized = normalizedPerLeg.get(leg).get(id);
+                if (normalized != null) {
+                    perLegNormalized[leg] = normalized;
```
bzhangam: `normalizedPerLeg` and `perLegNormalized` are a little bit confusing. Should we use a more descriptive name? Maybe `normalizedPerLeg` -> `normalizedLegs` and `perLegNormalized` -> `normalizedScoresAcrossLegs`?

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -434,6 +462,12 @@ protected QueryBuilder doRewrite(QueryRewriteContext queryShardContext) throws I
      * </ul>
      */
     private QueryBuilder doRewriteFused(QueryRewriteContext queryRewriteContext) throws IOException {
+        // Only the match set is wanted (this builder is a leg inside an enclosing fused query's Tail): contribute what the
+        // legs match and do not fan them out again — they already ran, as the enclosing query's leg sub-search. Checked
+        // ahead of everything else: it is the one case where the correct answer is to do no work at all.
+        if (MatchSetRewriteContext.isMatchSetOnly(queryRewriteContext)) {
+            return matchSetQuery();
+        }
         // Round 2: the async self-erase already produced the standard query — swap to it (or stay put until it lands).
```
owaiskazi19, line 473: We have couple of guardrails for window and all but we should also have one for min node version check.  The version gate at L148/L165 only protects serialization of the fusion field. It does not stop the coordinator from self-erasing into a hybrid_fusion NamedWriteable and dispatching it to a not-yet-upgraded data node mid-rolling-upgrade

martin-gaievski: yes, that's reals risk, pre-3.8 node can't resolve the hybrid_fusion name at all, so it fails while deserializing the shard request. For Core that gonna be ordinary shard failure which under the default low_partial_search_results=true is response 200 with documents silently missing. If a replica of that shard sits on an upgraded node, retry clears the recorded failure, so it's a 200 with _shards.failed=0 and no signal at all.
I'll work on the change to address this

martin-gaievski: pushed https://github.com/opensearch-project/neural-search/pull/1943/changes/9d8c2a54a900638572d1c82e738f035d672fa8f5 to address this

src/main/java/org/opensearch/neuralsearch/fusion/ScalarNormalizerFactory.java
```diff
@@ -0,0 +1,66 @@
...
+        MinMaxScalarNormalizer.INSTANCE.techniqueName(),
+        MinMaxScalarNormalizer.INSTANCE
+    );
+
+    /**
+     * @param techniqueName normalization technique name from the resolved fusion config
+     * @return the normalizer for that technique
+     * @throws IllegalArgumentException when the technique has no coordinator-side implementation yet
+     */
+    public static ScalarNormalizer create(final String techniqueName) {
```
bzhangam: Would get a better name since we are not actually creating the normalizer here?

src/main/java/org/opensearch/neuralsearch/query/CandidateScope.java
```diff
@@ -0,0 +1,523 @@
...
+            Disposition.OVERRIDDEN,
+            "an inline pipeline body is not applied per leg; the leg pipeline is pinned to _none"
+        );
+        put(table, SEARCH_SOURCE, "searchPipeline", Disposition.OVERRIDDEN, "as searchPipelineSource");
+        put(table, SEARCH_SOURCE, "timeout", Disposition.PROPAGATED, "bounds a leg's per-shard work as the user intended");
+        put(
+            table,
+            SEARCH_SOURCE,
+            "sliceBuilder",
+            Disposition.PROPAGATED,
```
bzhangam, line 210: If we propagate the slice won't it slice the leg's result?

src/main/java/org/opensearch/neuralsearch/query/CandidateScope.java
```diff
@@ -0,0 +1,523 @@
...
+            Disposition.NOT_PROPAGATED,
+            "rescoring reorders round 2's top documents after fusion, which is the intended post-fusion rerank; on a leg "
+                + "it would change the candidates instead"
+        );
+        put(
+            table,
+            SEARCH_SOURCE,
+            "explain",
+            Disposition.NOT_PROPAGATED,
+            "round 2 can only explain the self-erased query, a known fused-mode limitation independent of the legs"
```
bzhangam: Why not propagate to the legs and build the hybrid explain from the legs' explain? At least make it a future enrichment?

src/main/java/org/opensearch/neuralsearch/query/CandidateScope.java
```diff
@@ -0,0 +1,523 @@
...
+            table,
+            SEARCH_SOURCE,
+            "includeNamedQueriesScore",
+            Disposition.NOT_PROPAGATED,
+            "matched_queries reporting, which the self-erased round-2 query cannot carry anyway"
+        );
+        put(table, SEARCH_SOURCE, "suggestBuilder", Disposition.NOT_PROPAGATED, "suggestions do not depend on the query");
+        put(table, SEARCH_SOURCE, "stats", Disposition.NOT_PROPAGATED, "propagating would count every leg into the user's stats groups");
+        put(table, SEARCH_SOURCE, "extBuilders", Disposition.NOT_PROPAGATED, "plugin response extensions, not selection");
+        put(table, SEARCH_SOURCE, "profile", Disposition.NOT_PROPAGATED, "profiles the user's search, not the candidate query");
```
bzhangam: Will we be able to view the leg query profile? Maybe another future enrichemtn?

src/main/java/org/opensearch/neuralsearch/query/CandidateScope.java
```diff
@@ -0,0 +1,523 @@
...
+        rejectUnsupported(request);
+        return new CandidateScope(request);
+    }
+
+    /**
+     * Fail the {@link Disposition#REJECTED} shapes. Each check tests for a value the user actually supplied, so a plain
+     * request never trips one, and the messages name the field the user wrote rather than the field this class keys on.
+     */
+    private static void rejectUnsupported(final SearchRequest request) {
+        // A remote hit's _index is "cluster:index", which no shard of the remote index matches, so the _index-qualified
```
bzhangam: If it's a cross cluster search will we even hit here? Shouldn't we dispatch the search request through the remote cluster client and the local cluster should not try to process it?

owaiskazi19 (approved):
LGTM! Thanks @martin-gaievski

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -130,40 +132,103 @@ private static SearchHit[][] groupLegHits(MultiSearchResponse.Item[] items, int
...
+     * used: the same {@code _id} can be written to different shards of one index under different routing values, giving
+     * two genuinely distinct documents that share this key and are therefore fused as one. Qualifying further is not
+     * possible from here — a leg's {@link SearchHit} exposes {@link SearchHit#getShard()} but no routing value, so the
+     * coordinator has nothing to add to the key, and the {@code _routing} metadata field (queryable in principle) would
```
bzhangam: Using `legSource.storedField("_routing"); ` will not give us the `_routing` info? If it will should we include it to build the key?

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -130,40 +132,103 @@ private static SearchHit[][] groupLegHits(MultiSearchResponse.Item[] items, int
...
+     * first have to be fetched per leg hit, which the id-only leg deliberately does not do. One possible way to resolve this
+     * is leg-fetch.
+     */
+    private static String documentKey(SearchHit hit) {
+        return Objects.isNull(hit.getIndex()) ? hit.getId() : hit.getIndex() + KEY_SEPARATOR + hit.getId();
```
bzhangam: Why we need to have a fallback for the case when hit has no index? If there is hit which means the index must exist.

src/main/java/org/opensearch/neuralsearch/fusion/CoordinatorScoreFusion.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/query/CandidateScope.java
martin-gaievski: nope, leq query is invisible for normal profiler result, one of future enhacements

src/main/java/org/opensearch/neuralsearch/fusion/ScalarNormalizerFactory.java
martin-gaievski: makes sense, I think we should rename the class as well

src/main/java/org/opensearch/neuralsearch/query/CandidateScope.java
martin-gaievski: yup, that's in the road map for future

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: javadoc is inaccurate, I'll correct it

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: agree, will push fix for this

src/main/java/org/opensearch/neuralsearch/query/CandidateScope.java
martin-gaievski: agree, and real issue is reachability, scroll is NOT_PROPAGATED (permitted, never copied to a leg) and PIT is propagated, so scroll+slice makes every leg shard throw "all shards failed" with a misleading cause. obvious slice && !pit guard keys on wrong field, the root cause is scroll, and existing hybrid explicitly refuses scroll. Resolver silently allows it. This is reachable via _delete_by_query/_reindex, not just hand-rolled requests. 

IMO we need to refuse scroll in rejectUnsupported, this converts an opaque failure into a 400 response

src/main/java/org/opensearch/neuralsearch/query/CandidateScope.java
martin-gaievski: rewrite actually precedes the cross cluster search split at [TransportSearchAction](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/search/TransportSearchAction.java#L456), so dispatching through remote-cluster client is. not needed. Code comment is actually wrong: stated reason is inverted (SearchHit.getIndex() returns the bare name; alias is spliced in at render — real breakage is shard-side SearchIndexNameMatcher), and the guard's message is unreachable for a literal cluster:index because concreteIndices throws no such index first.  *:idx wildcards do reach the guard, so it neither the dead code nor a defect, still a 400

bzhangam (approved):

PR #1944, opened 2026-08-13
[RRF] Extract RRF rank arithmetic into a shared helper and deduplicate the hybridization processors

dbwiddis:
### Description

Prep work for replicating RRF under the coordinator-side "fusion" path introduced by #1933, plus a
duplication cleanup that fell out of reading the two hybridization processors side by side. Every
pre-existing test passes unmodified.

**Now behavior-preserving throughout.** An earlier revision hoisted `RRFProcessor.recordStats` *and*
fixed the `Optional.of` NPE lurking in it in the same commit. Per review feedback, that fix — plus the
validation that makes the bad config a 400 at pipeline creation instead of a 500 on every query — is
now a **separate commit of its own** (`80ce5f91`), a straight port of #1949 against `main`. The
refactor commit on top of it moves code and nothing else. Once #1949 merges and `main` is forward-merged
into this feature branch, that ported commit drops out as already-applied.

#### 1. Extract `RRFScoreNormalizer`

#1933 established the pattern: it lifted the scalar min-max arithmetic out of
`MinMaxScoreNormalizationTechnique` into a stateless `MinMaxScoreNormalizer` so the shard-side and
coordinator-side paths call one implementation of the formula. This does the same for RRF, in the same
style (`public final`, static-only, private constructor, same package).

Moved out of `RRFNormalizationTechnique`:

| New member | Was |
|---|---|
| `DEFAULT_RANK_CONSTANT`, `PARAM_NAME_RANK_CONSTANT`, `MIN_RANK_CONSTANT`, `MAX_RANK_CONSTANT` | constants on the technique |
| `scoreForRank(rank, rankConstant)` | `private float calculateNormalizedScore(int)` |
| `resolveRankConstant(params)` | `private int getRankConstant(Map)` + `getParamAsInteger` |
| `validateRankConstant(int)` | `private void validateRankConstant(int)` |
| `drainToRanks(queue, keyFunction)` | the priority-queue drain inside `sortDocumentsAsPerGlobalRankInIndividualQuery` |
| `assignRanksByScoreDescending(scoresById)` | new — the coordinator-side entry point |

`drainToRanks` takes the `PriorityQueue` rather than building one from a collection, and the queue
construction stays in `RRFNormalizationTechnique` unchanged. The reason is only that this is a move:
restructuring the caller's queue assembly is a rewrite reviewers would have to re-verify, for no
functional gain.

To be precise about what the heap does and does not buy, since an earlier revision of this description
overstated it: ranks here do not depend on using a heap. Heap-drain order and stable-sort order diverge
only for elements a comparator treats as equal, and the classic comparator (`ScoreDoc.COMPARATOR` then
`referenceShardId`) is a *total* order over results — measured, it returns `0` only for identical
`(docId, shardId)`, which cannot occur across two distinct results. Sorting a list with the same
comparator would therefore yield identical ranks. The `PriorityQueue` is kept because it is what the code
already did, not because the semantics require it.

`RRFNormalizationTechnique` keeps only what is `CompoundTopDocs`-shaped — the traversal, the
`ShardResultPerSubQuery` record, the comparator, `describe()`. `DEFAULT_RANK_CONSTANT` and
`PARAM_NAME_RANK_CONSTANT` remain as public aliases sourced from the normalizer, since
`RRFProcessorFactory` and several tests reach for them through the technique (same as #1933 did for
`MIN_SCORE`/`MAX_SCORE`).

`RRFScoreCombinationTechnique` is **not** extracted, on purpose: coordinator-side fusion builds a real
`ScoreCombinationTechnique` from `ScoreCombinationFactory` and calls `combine(float[])` on it directly,
so the combination side is already shared. Nothing to do there.

#### 2. Deduplicate the two hybridization processors

`NormalizationProcessor` and `RRFProcessor` were byte-for-byte identical across ~70 lines
(`hybridizeScores`, `shouldSkipProcessor`, `isHybridQuery`, `getQueryPhaseSearchResults`,
`getFetchSearchResults`, `getBeforePhase`, `getAfterPhase`, `isIgnoreFailure`). That is now implemented
once in `AbstractScoreHybridizationProcessor`, driven by four hooks — `getNormalizationTechnique()`,
`getCombinationTechnique()`, `getNormalizationWorkflow()` (all satisfied by Lombok `@Getter`) and
`recordStats()`. Each subclass is now just its `TYPE`, its fields, and its stats incrementers. The log
message uses `getType()` instead of a hard-coded name.

#### Preparatory commits ported from `main`

The base commit here mirrors #1942 (opened against `main`) so the extracted `scoreForRank` doesn't
land in its old allocating form and immediately need rewriting. It replaces the per-document
`BigDecimal` with the exact integer equivalent — `HALF_UP` rounding to scale 10 is expressible as
`(2 * 10^10 + d) / (2 * d)`, verified **bit-identical for all 2,147,483,646 reachable denominators** on
three JDKs. Kept as a separate commit so it can be swapped for the merged form of #1942 if that lands
differently.

The second, `80ce5f91`, ports #1949: `RRFProcessorFactory` rejects any `combination.technique` but `rrf`
(HTTP 400 at `PUT _search/pipeline/...`), and `recordStats` uses `Optional.ofNullable`. Both are
pre-existing bugs on `main`, not artifacts of this refactor, which is why they are fixed there and merely
carried here so the refactor commit stays a move.

#### One finding worth flagging before RRF is wired into fusion

Multi-shard ranks **are** globally score-ordered and interleave across shards; the existing
`testNormalization_whenResultFromMultipleShardsMultipleSubQueries_thenSuccessful` looks like it asserts
shard-grouped ranks but cannot distinguish them — it compares within `DELTA_FOR_ASSERTION = 0.001f`
while adjacent RRF scores at `rank_constant = 60` differ by roughly `0.00026`. A new test pins the real
contract with exact equality.

This matters for fusion: classic multi-shard ranking breaks score ties by **Lucene docId**, which does
not exist on the coordinator, so `assignRanksByScoreDescending` uses ascending `_id` as the closest
analogue. Any tie will therefore rank differently between the two paths.

That has now been measured rather than left open, because it decides whether fused-mode RRF should chase
bit-exact parity with classic on ties. It should not. Taking three docs all tied at score `1.0` — which is
not exotic; a `constant_score`, `match_all` or filter-context leg ties *every* doc by construction — and
laying the same logical hit set out differently:

| Layout | Resulting rank order |
|---|---|
| 1 shard, Lucene docIds 5, 3, 9 | `doc3` → `doc5` → `doc9` |
| 3 shards, each with local docId 0 | `shard0` → `shard1` → `shard2` |
| 1 shard, docIds 9, 5, 3 (post-merge renumbering) | `doc3` → `doc5` → `doc9` |

Classic's tie order is `docId` ascending then `shardId` ascending, and both are physical storage
artifacts: docId is reassigned by segment merges and reindexing, shardId by re-sharding. So the same tied
documents can change relative order without any change to the data or the query. It is deterministic for
a fixed index state, but it is not a contract.

Coordinator-side `_id` ascending is deterministic *and* independent of segment layout, shard count and
insertion order (verified: scrambled insertion orders produce identical ranks). Fused-mode RRF will
therefore define `_id` ascending as its tie contract, and the parity tests will assert the arithmetic over
tie-free score sets plus one explicit test pinning the tie divergence as intended — rather than trying to
reproduce a Lucene internal that is not present in `SearchHit` in order to match an order that is not
stable anyway.

#### Follow-up not in this PR

`HybridQueryBuilder.requireSupportedTechniques(...)` still gates out `rrf`; `CoordinatorScoreFusion`
needs `fuseRrf(...)`; `HybridFusionOrchestrator.computeRankedDocs(...)` ignores `fusion.rankConstant()`;
`FusionSpec` carries its own `DEFAULT_RANK_CONSTANT = 60` / `RANK_CONSTANT_KEY` literals and accepts any
`Number` where `RRFProcessorFactoryTests` requires negative/too-large/non-numeric to be rejected; and
`BaseNeuralSearchIT.createRRFSearchPipeline` takes `weights` but no `rank_constant`.

### Related Issues

Related to #1933 (this targets its feature branch). Preparatory commits mirror #1942 (merged) and #1949 (open against `main`).

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,3 +23,4 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Refactoring
 * [SemanticHighlighter] Traverse the query tree with a QueryBuilderVisitor instead of a "manual" walk ([#1915](https://github.com/opensearch-project/neural-search/pull/1915))
+* [RRF] Extract the rank arithmetic into a shared RRFScoreNormalizer and hoist the workflow duplicated between NormalizationProcessor and RRFProcessor into AbstractScoreHybridizationProcessor ([#1944](https://github.com/opensearch-project/neural-search/pull/1944))
--- a/src/main/java/org/opensearch/neuralsearch/processor/AbstractScoreHybridizationProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/AbstractScoreHybridizationProcessor.java
@@ -4,19 +4,37 @@
  */
 package org.opensearch.neuralsearch.processor;
 
+import org.opensearch.action.search.QueryPhaseResultConsumer;
 import org.opensearch.action.search.SearchPhaseContext;
+import org.opensearch.action.search.SearchPhaseName;
 import org.opensearch.action.search.SearchPhaseResults;
+import org.opensearch.neuralsearch.processor.combination.ScoreCombinationTechnique;
+import org.opensearch.neuralsearch.processor.normalization.ScoreNormalizationTechnique;
 import org.opensearch.search.SearchPhaseResult;
+import org.opensearch.search.fetch.FetchSearchResult;
 import org.opensearch.search.internal.SearchContext;
 import org.opensearch.search.pipeline.PipelineProcessingContext;
 import org.opensearch.search.pipeline.SearchPhaseResultsProcessor;
+import org.opensearch.search.query.QuerySearchResult;
 
+import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Collectors;
+
+import com.google.common.annotations.VisibleForTesting;
+import lombok.extern.log4j.Log4j2;
+
+import static org.opensearch.neuralsearch.search.util.HybridSearchResultFormatUtil.isHybridQueryStartStopElement;
 
 /**
  * Base class for all score hybridization processors. This class is responsible for executing the score hybridization process.
  * It is a pipeline processor that is executed after the query phase and before the fetch phase.
+ * <p>
+ * Subclasses supply the techniques to apply and the stats to record; the harvesting of query phase results and
+ * the handoff to {@link NormalizationProcessorWorkflow} are identical for all of them and are implemented here.
  */
+@Log4j2
 public abstract class AbstractScoreHybridizationProcessor implements SearchPhaseResultsProcessor {
     /**
      * Method abstracts functional aspect of score normalization and score combination. Exact methods for each processing stage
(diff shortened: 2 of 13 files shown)


dbwiddis:
Went through all four suggestions. One is a real bug and is now fixed (601c83f); the other three don't hold up. Details so they're auditable rather than just dismissed.

### 1. `Optional.of` NPE in `RRFProcessor.recordStats` — valid, fixed

Confirmed, and reachable from ordinary user configuration rather than hypothetically. `RRFProcessorFactory` reads `combination.technique` as a free-form string and hands it to `ScoreCombinationFactory`, which registers `arithmetic_mean`, `harmonic_mean`, `geometric_mean` and `rrf` — while `combTechniqueIncrementers` maps only `rrf`. So this pipeline

```json
{ "score-ranker-processor": { "combination": { "technique": "arithmetic_mean" } } }
```

threw `NullPointerException` on every query. It predates this PR (it is on `main`), but this PR is what moves the line, so it is fixed here rather than left behind. Now `ofNullable`, pinned by a test that reproduces the NPE against the old form:

```
RRFProcessorTests > testProcess_whenCombinationTechniqueHasNoStatsIncrementer_thenSucceed FAILED
    java.lang.NullPointerException
        at java.base/java.util.Optional.of(Optional.java:117)
        at org.opensearch.neuralsearch.processor.RRFProcessor.recordStats(RRFProcessor.java:59)
        at org.opensearch.neuralsearch.processor.AbstractScoreHybridizationProcessor.hybridizeScores(...)
```

The rating of `Low` — "limited impact unless custom techniques appear" — understates this. No custom technique is required; three built-ins reach it.

### 2. Range boundary semantics — no such change was made, and the semantics are inclusive

The premise that this PR changed `Range.between` to `Range.of` isn't accurate. `Range.of(MIN_RANK_CONSTANT, MAX_RANK_CONSTANT)` was already present in `RRFNormalizationTechnique` on the base commit (a71331b4, line 51) and moved verbatim into `RRFScoreNormalizer`. Verified anyway against the version actually on the classpath (commons-lang3 3.18.0, resolved through the OpenSearch BOM):

```
of(1,10000).contains(1)     = true
of(1,10000).contains(10000) = true
of(1,10000).contains(0)     = false
of(1,10000).contains(10001) = false
of(1,10000).equals(between(1,10000)) = true
```

Inclusive on both ends, matching the "between 1 and 10000" message; `of` is the rename of `between` (deprecated in 3.13.0) and produces an equal range. This PR also closes the coverage gap the suggestion assumed existed: `RRFScoreNormalizerTests` now asserts `validateRankConstant` accepts both `MIN_RANK_CONSTANT` and `MAX_RANK_CONSTANT`, where the pre-existing factory tests only exercised `-1` and `50000`.

### 3. Delimiter / start-stop `ScoreDoc`s contaminating global ranks — not applicable

Two independent reasons.

First, no filtering was removed. The diff on `RRFNormalizationTechnique` only lifts the rank-drain block out into `drainToRanks`; the shard / sub-query / scoreDoc loops are unchanged from the base commit, so there is no "previously skipped, now added" transition to worry about.

Second, markers structurally cannot be present at that point. `CompoundTopDocs` consumes them as separators while splitting the flat shard array into per-sub-query `TopDocs`: the loop starts at index 2 to skip the leading start-stop pair, and on hitting `isHybridQueryDelimiterElement` / `isHybridQueryStartStopElement` it closes the current sub-query rather than adding the marker to `scoreDocList` (`CompoundTopDocs.java` 107-131). By the time `normalize()` runs, every `TopDocs` holds real hits only.

As a side note, the suggested replacement also removes the enclosing loops while referencing `scoreDocs` and `shardResults`, which are declared inside them.

### 4. Input validation in `scoreForRank` — declining

Both inputs are already constrained at their entry points. `rankConstant` passes through `resolveRankConstant` → `validateRankConstant`, so no technique instance can hold an out-of-range value; `rank` only ever originates from `drainToRanks`, which emits `0..n-1`. The guard would defend against a caller that cannot exist.

Against that, this is a per-document, per-sub-query, per-shard hot path — the preceding commit in this PR exists specifically to remove a `BigDecimal` allocation from it — so two always-false branches would cost something for no coverage gained. The constraints are stated in the Javadoc (`@param rank zero based rank of the document within the sub-query`), which is the alternative the suggestion itself offers.

martin-gaievski:
> Went through all four suggestions. One is a real bug and is now fixed ([601c83f](https://github.com/opensearch-project/neural-search/commit/601c83f3dad22b5c90940e043f584d68066a1525)); the other three don't hold up. Details so they're auditable rather than just dismissed.
> 
> ### 1. `Optional.of` NPE in `RRFProcessor.recordStats` — valid, fixed
> Confirmed, and reachable from ordinary user configuration rather than hypothetically. `RRFProcessorFactory` reads `combination.technique` as a free-form string and hands it to `ScoreCombinationFactory`, which registers `arithmetic_mean`, `harmonic_mean`, `geometric_mean` and `rrf` — while `combTechniqueIncrementers` maps only `rrf`. So this pipeline
> 
> ```json
> { "score-ranker-processor": { "combination": { "technique": "arithmetic_mean" } } }
> ```
> 
> threw `NullPointerException` on every query. It predates this PR (it is on `main`), but this PR is what moves the line, so it is fixed here rather than left behind. Now `ofNullable`, pinned by a test that reproduces the NPE against the old form:
> 
> ```
> RRFProcessorTests > testProcess_whenCombinationTechniqueHasNoStatsIncrementer_thenSucceed FAILED
>     java.lang.NullPointerException
>         at java.base/java.util.Optional.of(Optional.java:117)
>         at org.opensearch.neuralsearch.processor.RRFProcessor.recordStats(RRFProcessor.java:59)
>         at org.opensearch.neuralsearch.processor.AbstractScoreHybridizationProcessor.hybridizeScores(...)
> ```
> 
> The rating of `Low` — "limited impact unless custom techniques appear" — understates this. No custom technique is required; three built-ins reach it.
> 

What is our fix for this - silent acceptance or nicer error message? IMO we shouldn't accept incorrect configuration, and there are tow scenarios: if that's defined and read in from pipeline it needs to be validated in factory during pipeline creation; for in-query fusion mode if search config is part of the request then we validate it at query runtime.

martin-gaievski (commented):
Please add changelog item for this change

dbwiddis:
> What is our fix for this - silent acceptance or nicer error message? I

Probably a 400 error?  It was pre-existing, so whatever fix we apply needs to go against current main branch code, not just this feature.

> Please add changelog item for this change

How are we doing change logs on a feature branch?  Merging with the main branch changelog is going to create a lot of conflicts when we eventually try to merge together.

dbwiddis:
Submitted https://github.com/opensearch-project/neural-search/pull/1949/changes against `main` to avoid the NPE and return `BAD REQUEST`, and rolled that commit into this PR.

martin-gaievski:
> > What is our fix for this - silent acceptance or nicer error message? I
> 
> Probably a 400 error? It was pre-existing, so whatever fix we apply needs to go against current main branch code, not just this feature.

agree, I just checked main has problem, so NPE specific fix needs to go to main

dbwiddis:
> agree, I just checked main has problem, so NPE specific fix needs to go to main

YEp, as mentioned already filed #1949 for that.

That commit is folded into this one as well since they're on different branches.

martin-gaievski (approved):
Looks good, thanks

PR #1948, opened 2026-08-17
[RRF] Wire RRF into the coordinator-side fusion path

dbwiddis:
### Description

Enables `rrf` for the coordinator-side (fused) mode of the `hybrid` query, so an RRF user can move to fused mode without any change to their config and without any change to their scores.

> [!NOTE]
> Rebased onto `feature/fusion-hybrid-query` now that #1944 has merged, so this is down to its own four commits — no stacked prerequisite left to skip over. The rebase was content-neutral: the tree is byte-identical to the pre-rebase tip, which also confirms the squash of #1944 landed exactly what this branch was carrying.

Commits, bottom-up:

| commit | what |
|:--|:--|
| `641111fc` | **Behavior fix on #1933.** `FusionSpec` read `combination.parameters.rank_constant`, but `RRFProcessorFactory` — the ground truth for the `score-ranker-processor` shape — reads `rank_constant` directly off the `combination` clause and rejects anything but `weights` under `parameters`. So every user-supplied rank constant was silently ignored (fused mode always ranked at the default 60), and the shape `FusionSpec` documented was rejected by the real processor with a 400. Fused-mode-only code, never released. |
| `cc2441bc` | `CoordinatorScoreFusion.fuseRrf` plus the wiring. Reuses the helpers from #1944, so fused and classic compute rank scores through the same code. `rrf` is admitted only with normalization `none`; pairing it with a normalization technique is contradictory and is rejected rather than silently dropped. |
| `6b82bbb4` | Integration test asserting fused RRF against classic RRF directly. |
| `b6e0aa7c` | CHANGELOG entry. |

### The parity claim, and how it is checked

The point of the #1944 extraction is that the two paths **cannot** drift, so that is asserted rather than assumed, at three levels.

**1. Unit.** `CoordinatorScoreFusionRrfDifferentialTests` runs the same hit sets through `RRFNormalizationTechnique` + `RRFScoreCombinationTechnique` and through `fuseRrf`, and requires the scores to agree at `DELTA = 0.0f` — bit for bit, not within a tolerance. Any tolerance there would hide exactly the divergence the test exists to catch. Covers both classic ranking paths (positional single-shard and the cross-shard priority queue), partial leg overlap, weights, `rank_constant` bounds, and raw-score magnitude independence.

**2. Integration** — `HybridQueryFusedModeIT#testFusedMode_whenRrf_thenScoresMatchClassicScoreRankerPipeline`. One index, one pair of sub-queries, three requests: classic via a `score-ranker-processor` pipeline, fused with inline config, and fused reading *that same pipeline*. Requires identical documents and identical scores, for `rank_constant` 1 / 60 / 10000, unweighted and weighted `[0.3, 0.7]`. Verified non-vacuous by mutation — passing `rankConstant + 1` to the inline config fails the per-document assertion with a full score diff.

**3. Manual, against a live cluster** — below. Not a substitute for (2); it is how the two divergences were characterised, and it covers the multi-shard and narrow-window combinations the IT deliberately does not pin.

<details>
<summary><b>End-to-end run against a live cluster — classic RRF vs fused RRF</b></summary>

Both paths were run against a real single-node OpenSearch `3.8.0-SNAPSHOT` cluster (`./gradlew run`, which installs k-NN, ml-commons and this plugin at the matching version), over the same index with the same two sub-queries. Only the way RRF is requested differs:

| path | how RRF is requested |
|:--|:--|
| classic (shard-side) | `score-ranker-processor` search pipeline |
| fused, inline config | `fusion: {combination: {technique: rrf, rank_constant: K}}` |
| fused, pipeline config | `fusion: {source: pipeline}` — reads the *same* pipeline as the classic run |

Every returned `(_id, _score)` pair was compared for **exact float equality**, not within a delta.

**Result: exact match** across rank_constant 1 / 60 / 10000, unweighted and weighted `[0.3, 0.7]`, on 1 shard and on 3 shards, from both config sources.

```
## Part 1 — single shard

- leg A (title:alpha) order: 5, 1, 2, 3, 4, 9, 12, 11, 7 — within-leg score ties: none
- leg B (body:beta)  order: 6, 4, 3, 2, 9, 1, 12, 8, 11 — within-leg score ties: none

### rank_constant = 1
  MATCH   classic pipeline vs fused inline
  MATCH   classic pipeline vs fused reading THE SAME pipeline

### rank_constant = 60
  MATCH   classic pipeline vs fused inline
  MATCH   classic pipeline vs fused reading THE SAME pipeline

  Fused result (identical to classic):
  | rank | _id | _score      |
  |-----:|:----|:------------|
  |    0 | 4   | 0.031513646 |
  |    1 | 2   | 0.031498015 |
  |    2 | 3   | 0.031498015 |
  |    3 | 1   | 0.031280547 |
  |    4 | 9   | 0.03053613  |
  |    5 | 12  | 0.029850746 |
  |    6 | 11  | 0.029198635 |
  |    7 | 5   | 0.016393442 |
  |    8 | 6   | 0.016393442 |
  |    9 | 8   | 0.014705882 |
  |   10 | 7   | 0.014492754 |

### rank_constant = 10000
  MATCH   classic pipeline vs fused inline
  MATCH   classic pipeline vs fused reading THE SAME pipeline

### weights [0.3, 0.7], rank_constant = 60
  MATCH   classic pipeline vs fused inline
  MATCH   classic pipeline vs fused reading the same pipeline

## Part 2 — three shards, window (100) larger than the match set
  MATCH   classic pipeline vs fused inline, 3 shards
  MATCH   classic pipeline vs fused reading the same pipeline, 3 shards

## Part 3 — narrow window, at the documented `size <= window_size` contract
  MATCH   1 shard,  depth = window = size = 3
  DIFFER  3 shards, depth = window = size = 3   (divergence expected, see below)
      classic: [('4', 0.031544957), ('1', 0.031544957), ('3', 0.031513646)]
      fused  : [('4', 0.016393442), ('1', 0.016393442), ('3', 0.016129032)]

## Part 4 — the deliberate tie divergence
  (index written in DESCENDING _id order, and one leg a constant_score tying every doc at 1.0)
  classic (docId order within the tied leg):   ['12','6','9','11','4','3','8','2','1','10','7','5']
  fused   (_id ascending within the tied leg): ['1','4','3','2','6','12','11','9','8','10','5','7']
  Same document set: True.  Identical order: False.
```

Exactly two divergences exist. Both are below.

</details>

### Tie ordering: a deliberate decision that can change behavior

**Where two documents tie on score within one leg, fused mode can order them differently from classic.** This is the one place the arithmetic-identical paths produce a different answer, and it is a choice.

Classic breaks a within-leg tie by Lucene `docId`, then by shard id. Neither exists on the coordinator, which sees only merged `_id`-keyed results — but more to the point, **neither is a stable contract today**: `docId` is reassigned by segment merges and reindexing, and shard id by re-sharding. The same two tied documents can therefore swap places on the classic path with no change to the data or the query, purely as a function of storage layout and merge timing. There is no deterministic order there to preserve.

Fused mode breaks the tie by **`_id` ascending**: deterministic, and independent of physical layout, so tied documents get the same relative order on every replica, before and after a merge, and across a reindex. The trade is that a nondeterministic order is replaced by a deterministic one that is not always the order the nondeterministic one happened to produce — someone watching a specific pair of tied documents may see them swap once, and then never again. `_id` ordering is lexicographic on the string, so `"10"` sorts before `"2"`; that is unavoidable when the key is the `_id`, and it is stable, which is the property being bought.

This is documented on `fuseRrf` and pinned by a unit test that inserts tied documents in two different orders and asserts an identical result. The parity IT sidesteps it deliberately: its corpus gives every document a distinct (term frequency, length) pair per leg so neither leg ties, and the test **asserts that tie-freeness up front** — so a future scoring change surfaces as "the corpus developed a tie" rather than as an unexplained parity failure. Score comparison is also split into a per-document map check plus a separate score-*sequence* check, which keeps the assertion agnostic about which document wins a tie on the *fused* score.

### The other divergence: candidate pool under a truncating window

Where a window truncates across more than one shard, classic ranks the union of each shard's top-`pagination_depth` per leg (up to `depth × shards` documents); fused ranks the coordinator-merged global top-`window_size` per leg — a strictly smaller pool. This is inherent to fusing after the shard merge, not a difference in the RRF arithmetic (Part 3 above). The pools coincide on a single shard, or whenever the window holds the whole match set. Widening the fused pool is follow-up work; the parity IT uses a single shard and a non-truncating window so the comparison is well defined, rather than pinning today's limitation as desired behavior.

### Out of scope, deliberately

Fusion-aware `explain` / `profile`, multi-index fused queries, and the remaining normalization techniques (`z_score`, `l2`) are unchanged here and remain scoped to later changes. Each is rejected or documented rather than silently mis-fused.

### Related Issues

Part of #1941. Builds on #1944. Follows #1933.

### Check List


src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -162,7 +168,9 @@ private static RankedDocs computeRankedDocs(SearchHit[][] legHits, FusionSpec fu
             fusion.combinationTechnique(),
             weightsParams(fusion.weights())
         );
-        Map<String, Float> combined = CoordinatorScoreFusion.fuseMinMax(legRawScores, combination);
+        Map<String, Float> combined = FusionSpec.TECHNIQUE_RRF.equals(fusion.combinationTechnique())
```
bzhangam: Since we can have more different fuse techs it's better to make it a map<tech_name, tech_handler> to make it easier to extend it for other techs.

dbwiddis: Agreed, and done — there is no dispatch here at all now. `rrf` resolves through the `ScalarNormalizers` name -> factory map that the score-based techniques already use, so `CoordinatorScoreFusion` just normalizes then combines without knowing which technique it holds, and adding a technique is a new `ScalarNormalizer` plus one map entry.

It fits that contract because reciprocal rank fusion *is* a whole-leg normalization: rank comes from the leg's full value set, the same way `min_max` derives min/max from it. Classic models it identically — `RRFProcessorFactory` builds an `RRFNormalizationTechnique` for a processor with no normalization clause — so `FusionSpec` now resolves that shape to `normalization: rrf` and `fuseRrf` is gone (~60 lines of duplicated fusion loop with it). Since `rrf` is the only technique carrying a parameter, the map values are factories rather than singletons, mirroring `ScoreNormalizationFactory`.

src/main/java/org/opensearch/neuralsearch/query/FusionSpec.java
```diff
@@ -47,8 +52,10 @@ public final class FusionSpec {
     // Normalization techniques
     static final String NORMALIZATION_NONE = "none";
```
martin-gaievski: please delete this one, you just removed the only caller for this constant, it's official dead code now

martin-gaievski (commented):
we start accepting combination techniques other than arithmetic_mean, that's not correct and not backward compatible with classic hybrid query. Also in that case we are silently ignoring custom value of RRF parameter `rank_constant`, and falling back to it's default value 60.

src/main/java/org/opensearch/neuralsearch/query/FusionSpec.java
dbwiddis: Done in 90a0de2a. It was still the constructor's null-coalesce fallback, but now that `rrf` resolves to `normalization=rrf`, `"none"` no longer names anything the coordinator understands — that fallback could only produce a value the technique gate rejects. Constant deleted, field is `Objects.requireNonNull`.

dbwiddis:
Both real, both fixed in f4903944 — thanks.

**`rank_constant` silently ignored.** A classic-legal `normalization-processor` with `normalization: {technique: rrf, parameters: {rank_constant: N}}` + `arithmetic_mean` fused at 60 regardless of N: `fromNormalizationProcessor` hardcoded the default, and only the score-ranker path read the value (from the `combination` clause). Each shape now reads it where its own classic factory reads it — `normalization.parameters` for `normalization-processor`, the `combination` clause for `score-ranker-processor` — both via the shared `RRFScoreNormalizer.resolveRankConstant`, so an out-of-range value gets classic's 400 instead of a silent fallback.

**Accepting a combination technique classic rejects.** The `rrf + rrf` shortcut in `requireSupportedTechniques` was keyed on technique *names*, so it also admitted a `normalization-processor` asked to combine rrf-normalized scores by rrf — a pairing classic rejects through `ScoreNormalizationFactory`'s compatibility matrix. `FusionSpec` now carries the config `Shape` (`NORMALIZATION_PROCESSOR` / `SCORE_RANKER_PROCESSOR`) and the exemption keys on that, so only the score-ranker shape bypasses the matrix. The matrix stays the single source of truth — no second copy of the compatibility rules.

Three tests were written to fail first (`expected:<100> but was:<60>`, plus two missing-exception cases). Full unit suite green, and `HybridQueryFusedModeIT` 13/13 — including the tests that compare fused output against the classic pipeline directly.

martin-gaievski (approved):
Looks good, thanks for addressing all comments

bzhangam (approved):

PR #1949, opened 2026-08-18
[RRF] Reject combination techniques the score-ranker-processor cannot use

dbwiddis:
### Description

`score-ranker-processor` normalizes by rank, so `rrf` is the only combination technique that means anything to it. `RRFProcessorFactory` nonetheless accepted any technique registered in `ScoreCombinationFactory` — including the three means (`arithmetic_mean`, `harmonic_mean`, `geometric_mean`), which belong to `normalization-processor`, where they combine already-normalized scores.

Such a pipeline was accepted at creation and then **threw `NullPointerException` on every query**:

```java
// RRFProcessor
private final Map<String, Runnable> combTechniqueIncrementers = Map.of(
    RRFScoreCombinationTechnique.TECHNIQUE_NAME, ...   // rrf and nothing else
);

private void recordStats(ScoreCombinationTechnique combinationTechnique) {
    EventStatsManager.increment(EventStatName.RRF_PROCESSOR_EXECUTIONS);
    Optional.of(combTechniqueIncrementers.get(combinationTechnique.techniqueName()))  // NPE on a miss
        .ifPresent(Runnable::run);
}
```

`recordStats` runs unconditionally, before any normalization or combination work, so **no such pipeline could ever serve a query**. There is no working configuration to preserve, which is what makes rejecting the config strictly an improvement rather than a breaking change.

This came out of review feedback on #1944, but the bug predates that PR, so it is fixed here against `main`.

### Changes

1. **`RRFProcessorFactory` rejects any combination technique but `rrf`** → `IllegalArgumentException` from a processor factory, which OpenSearch surfaces as **HTTP 400 at `PUT _search/pipeline/...`** instead of HTTP 500 on every subsequent search. The guard sits before `createCombination`, and the message is a *superset* of the previous string:

   ```
   provided combination technique is not supported by [score-ranker-processor], supported technique is [rrf], got [arithmetic_mean]
   ```

   Because it still contains `provided combination technique is not supported`, the existing assertion in `testInvalidCombinationName_whenUnsupportedFunction_thenFail` holds unchanged — and an unregistered name like `my_function` now gets the more informative message too. One code path, no test churn.

2. **`recordStats` uses `Optional.ofNullable`.** With (1) in place the miss is unreachable from a pipeline definition, but `RRFProcessor`'s constructor is public, and stats bookkeeping should never fail a query. Defence in depth, not the primary fix.

### Verification

**The NPE was reproduced on unmodified `main` first**, so the claim above isn't inferred from reading the code:

```
RRFProcessorTests > testProcess_whenCombinationTechniqueHasNoStatsIncrementer_thenSucceed FAILED
    java.lang.NullPointerException
        at java.base/java.util.Optional.of(Optional.java:117)
        at org.opensearch.neuralsearch.processor.RRFProcessor.recordStats(RRFProcessor.java:159)
15 tests completed, 1 failed
```

Added tests:

| Test | What it pins down |
|---|---|
| `RRFProcessorFactoryTests.testInvalidCombinationName_whenTechniqueBelongsToNormalizationProcessor_thenFail` | Each of the three means is rejected, and the message names the rejected technique |
| `RRFProcessorFactoryTests.testCombinationName_whenRrfExplicitlyRequested_thenSuccessful` | The counterpart — an explicit `rrf` is still accepted, so the guard rejects only what never worked |
| `RRFProcessorTests.testProcess_whenCombinationTechniqueHasNoStatsIncrementer_thenSucceed` | The `ofNullable` path via direct construction (the reproduction above) |
| `RRFProcessorIT.testRRFPipelineCreation_whenCombinationTechniqueIsNotRrf_thenBadRequest` | **Asserts the 400 against a live cluster** rather than assuming the status code |

The IT was mutation-checked — with the guard disabled it fails for the right reason, so it is not vacuously green:

```
junit.framework.AssertionFailedError: Expected exception ResponseException but no exception was thrown
    at org.opensearch.neuralsearch.processor.RRFProcessorIT
       .testRRFPipelineCreation_whenCombinationTechniqueIsNotRrf_thenBadRequest(RRFProcessorIT.java:90)
```

With the guard in place, `RRFProcessorIT` is `tests="2" failures="0" errors="0"`, and the full unit suite is `BUILD SUCCESSFUL`.

### Deliberately not changed

`RRFProcessorFactory` also **silently ignores a `normalization` clause** (enshrined by the existing `testInvalidTechniqueType_whenPassingNormalization_thenSuccessful`). That is a second flavour of improper config, but unlike this one those pipelines *work correctly today* — the ignored clause defaults to `rrf` anyway — so rejecting them would break working configurations on upgrade. Worth a separate discussion rather than a drive-by change here.

### Related Issues

Follows up on review feedback in #1944.

### Check List
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/opensearch-project.github.io/blob/main/_community_members/DEVELOPER_CERTIFICATE_OF_ORIGIN.md).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * [Hybrid Query] Fix NoSuchElementException in hybrid query with sort/search_after when a shard returns no results ([#1939](https://github.com/opensearch-project/neural-search/pull/1939))
 * [SemanticHighlighter] Fix SemanticHighlighterExtBuilder.toXContent ([#1906](https://github.com/opensearch-project/neural-search/issues/1906)) (query-insights [#651](https://github.com/opensearch-project/query-insights/issues/651))
 * [Sparse ANN] Fold sparse vector tokens into the signed-short range (modulus 32768) so folded tokens are never sign-extended to a negative value when stored in short[] ([#1926](https://github.com/opensearch-project/neural-search/pull/1926))
+* [RRF] Reject a combination technique other than rrf when creating a score-ranker-processor, instead of accepting the pipeline and throwing NullPointerException on every query ([#1949](https://github.com/opensearch-project/neural-search/pull/1949))
 
 ### Infrastructure
 
--- a/src/main/java/org/opensearch/neuralsearch/processor/RRFProcessor.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/RRFProcessor.java
@@ -156,6 +156,8 @@ public class RRFProcessor extends AbstractScoreHybridizationProcessor {
 
     private void recordStats(ScoreCombinationTechnique combinationTechnique) {
         EventStatsManager.increment(EventStatName.RRF_PROCESSOR_EXECUTIONS);
-        Optional.of(combTechniqueIncrementers.get(combinationTechnique.techniqueName())).ifPresent(Runnable::run);
+        // ofNullable, not of: a technique with no incrementer must not fail the query. RRFProcessorFactory rejects any
+        // technique but rrf, so this is no longer reachable from a pipeline definition, but this constructor is public.
+        Optional.ofNullable(combTechniqueIncrementers.get(combinationTechnique.techniqueName())).ifPresent(Runnable::run);
     }
 }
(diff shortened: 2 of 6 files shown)


martin-gaievski (approved):
looks good, thanks!

PR #1956, opened 2026-08-20
[Hybrid Query] Gate collapse distinct-groups collection behind an opt-in index setting

### LINKED ISSUE #1947: [BUG] Hybrid query `collapse` under-returns groups
## Description

For a hybrid query, `HybridCollapsingTopDocsCollector` collects the top-`numHits` **documents** by score and labels each with its collapse group; it does not itself collapse by group (that is deferred to the downstream `CollapseDataCollector`, per the collector's own Javadoc). This is only correct if enough documents are collected that, after downstream per-group dedup, at least `numHits` **distinct groups** remain. The per-group over-collection that used to guarantee that (`index.neural_search.hybrid_collapse_docs_per_group_per_subquery`) is now a deprecated no-op, so when one group owns several of a shard's top-`numHits` documents, those duplicate slots crowd out other groups. After downstream dedup collapses the duplicates, the response contains **fewer groups than `size`** — silently dropping valid groups.

User-visible on a single-shard index: `size: 5` over 6 distinct collapse groups returns **4 hits**.

Reproduced on `main` /`3.8.0` branch. 

## Steps to reproduce

Single shard, 6 collapse groups, where group `gA` owns the three highest-scoring documents so it fills multiple slots in the collector's top-5 queue:

```
PUT tmp-collapse-dup — 1 shard, 0 replicas, index.search.default_pipeline = <min_max + arithmetic_mean>
mappings: { grp: keyword, s: integer }

docs (id → grp, s):
  0 → gA, 10      4 → gE, 60
  1 → gB, 90      5 → gF, 50
  2 → gC, 80      6 → gA, 95
  3 → gD, 70      7 → gA, 99
```

```json
POST tmp-collapse-dup/_search
{
  "size": 5,
  "query": { "hybrid": { "queries": [
    { "function_score": { "query": { "match_all": {} }, "field_value_factor": { "field": "s", "modifier": "none", "missing": 1 } } },
    { "function_score": { "query": { "match_all": {} }, "field_value_factor": { "field": "s", "modifier": "none", "missing": 1 } } }
  ] } },
  "collapse": { "field": "grp" }
}
```

### Expected

5 hits — the 5 best distinct groups by their top document: `gA`, `gB`, `gC`, `gD`, `gE`. (`gF`, whose best doc scores 50, is correctly the 6th and excluded.)

### Actual

4 hits. `gE` is missing even though it is a distinct group whose top document (60) outranks `gF` (50):

```
7:[gA]@1.0   1:[gB]@0.6896552   2:[gC]@0.3448276   3:[gD]@0.001
```

## Root cause

In `src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java`, collect dispatch keys only on whether the queue is full — never on whether the incoming document's collapse group already holds a slot:

```java
if (queueFull[subQuery]) {
    updateExistingEntry(compoundQueryScorer, subQuery, doc, score);  // replaces the queue BOTTOM
} else {
    addNewEntry(subQuery, doc, score);                               // allocates a NEW slot
}
```

`updateExistingEntry` replaces the weakest heap entry (acceptance is `comparator.compareBottom(doc) > 0`), not the entry for the doc's group — no group identity is consulted. `groupValueBySlot` is a write-only label, set *after* the slot is chosen and read only in `topDocs()`. `topDocs()` pops all `numHits` entries and emits each with its label, with no group collapse. `GroupSelector` (`CollapseDocSourceGroupSelector`) only extracts the current doc's value; `setGroups` throws `UnsupportedOperationException`. So no stage in this collector deduplicates by group — it is a top-`numHits`-documents collector, and one group can occupy multiple slots.

Trace for the reproduction above (`numHits = 5`, min-heap by score, single shard/segment), verified by simulation:

| step | queue (min-heap) | note |
|---|---|---|
| docs 0-4 | `gA(10) gB(90) gC(80) gD(70) gE(60)` | queue fills; bottom = `gA(10)` |
| doc 5 `gF(50)` | `gF(50) gB gC gD gE` | 50 > bottom `gA(10)` → replaces it |
| doc 6 `gA(95)` | `gA(95) gB gC gD gE` | 95 > bottom `gF(50)` → replaces it |
| doc 7 `gA(99)` | `gA(95) gB gC gD gA(99)` | 99 > bottom `gE(60)` → **`gE` evicted; `gA` now holds 2 slots** |

Retained 5 documents are the 5 highest-scoring: `gA(99), gA(95), gB(90), gC(80), gD(70)`. Because `gA` contributes two of them, only **4 distinct groups** fit, and the 5th-best distinct group `gE(60)` is pushed out. Downstream `CollapseDataCollector` (`:127-140`, a `Map<collapseValue, FieldDoc>` keeping the best doc per group) then merges the two `gA` slots into one — leaving 4 hits, `gE` gone. `CollapseDataCollector` is not a second bug; it correctly dedups but cannot recover a group the flat queue already evicted.

Correct collapse semantics require the top-`numHits` **groups** (by best document), not the top-`numHits` documents: either track group→slot in the collector and evict per group, or restore per-group over-collection so downstream dedup has enough documents to yield `numHits` distinct groups.

### Precise trigger condition

Per sub-query, per shard/segment, before coordinator merge: `numHits` (= `size`, or `pagination_depth`; equal to `size` when `from = 0`) is **strictly less than the number of distinct groups**, AND at least one group owns **≥ 2 of the top-`numHits` scored documents**. No per-group over-fetch compensates.

## Multi-shard behavior (scope)

The under-return is triggered **per shard**. Empirically (function_score numeric scoring, so results are shard-independent and deterministic):

| scenario | hits |
|---|---|
| hybrid + collapse, **1 shard** | **4** — `gA, gB, gC, gD` (`gE` dropped) ❌ |
| hybrid + collapse, **3 shards** | **5** — `gA, gB, gC, gD, gE` ✅ |
| non-hybrid collapse control, 3 shards | 5 ✅ |

On multiple shards, the `gA` documents spread across shards, so no single shard locally has one group owning multiple top-slots-that-evict-a-unique-group, and the coordinator's group-aware collapse merge assembles the correct 5 groups. So the reliable trigger is a **single-shard index** (or any shard that locally holds ≥ 2 competitive documents of one group within its top-`numHits`, where the evicted group has no representative on other shards). This bounds the real-world impact but does not eliminate it.

## Additional evidence — an existing unit test would hide this

`HybridCollapsingTopDocsCollectorTests.testCollapse_whenEvictedGroupReappears_thenNoError` funnels result groups through `Set<String> survivingGroups = new HashSet<>()` and asserts only membership and `collapseValues.length == numHits`. Duplicate collapse values in the collector output would therefore go undetected. It should additionally assert that `collapseValues` contains **no duplicate group values** and that the distinct-group count equals `numHits`. (The product bug itself is proven independently above by the code trace and the end-to-end reproduction; this test does not prove it, but its assertion shape is why it never flagged it.)

Two other unit tests currently **enshrine** the buggy behavior and must be updated by any fix:
- `testCollapse_whenAllDocsInSameGroup_thenQueueFillsCorrectly` (`:285-352`) asserts the collector returns 5 documents from a **single** group (per-document, not per-group).
- `testCollapse_whenMoreGroupsThanNumHits_thenOnlyTopKGroupsSurvive` (`:724-781`) passes only because it uses 20 **unique** groups (1 doc each), deliberately avoiding the crowding case.

## Separate, related test bug (seed-flaky)

The same test is also intermittently flaky, a distinct defect worth fixing alongside. It injects scores **by doc id**:

```java
float[] scores = { 0.1f, 0.9f, 0.8f, 0.7f, 0.6f, 0.5f, 0.95f, 0.99f };
for (int i = 0; i < scores.length; i++) { ...; leafCollector.collect(i); }
```

but assumes doc-id order equals insertion order. Lucene's `newIndexWriterConfig()` selects `MockRandomMergePolicy` only ~1% of the time (`rarely()`), which then **reverses** doc order ~50% of the time — roughly 0.5% of runs. `forceMerge(1)` guarantees a single segment, not insertion order. That low probability is exactly why the failure is seed-dependent rather than constant. Under `-Dtests.seed=628A981950B4CEF` the reversal fires, giving docId→group `0=A,1=A,2=F,3=E,4=D,5=C,6=B,7=A`; group maxes become `A=0.99,B=0.95,F=0.8,E=0.7,D=0.6,C=0.5`, so the top-5 keeps `{A,B,F,E,D}`, `groupC(0.5)` is evicted, `groupF` survives, and `assertFalse("groupF should be evicted...", ...)` fails. Fix by deriving each document's score from its group rather than its doc id, or by pinning the doc→group mapping explicitly.

Reproduce:

```
./gradlew ':test' --tests 'org.opensearch.neuralsearch.search.collector.HybridCollapsingTopDocsCollectorTests.testCollapse_whenEvictedGroupReappears_thenNoError' \
  -Dtests.seed=628A981950B4CEF -Dtests.security.manager=false -Dtests.locale=sms-Latn-FI -Dtests.timezone=Africa/Libreville
```

Fails on `main`; passes on random seeds.

## Suggested fix

1. **Collector:** collect top-`numHits` **groups**, not documents — track group→slot (e.g. `Map<Object, Integer>` per sub-query) so collect() updates the slot a group already owns and evicts the weakest *group*; or restore/replace the per-group over-collection so downstream dedup yields `numHits` distinct groups.

## Impact

`hybrid` + `collapse` query on a **single-shard index** (or where a shard locally holds ≥ 2 competitive documents of one group within its top-`size` window) can silently return fewer groups than `size`, dropping valid groups. In the extreme — all top-`size` documents belonging to one dominant group — as few as **1** group is returned. Multi-shard indices with documents spread across shards are generally masked by the coordinator collapse merge. Relevance-visible and silent: no error or warning.

Ystk-hsn:
### Description
Implements the decision from #1947: both collapse behaviors make sense but are mutually exclusive, so the existing behavior stays the default and the distinct-groups collection is opt-in. This adds
```
index.neural_search.hybrid_collapse_distinct_groups_enabled   (boolean, default false, IndexScope + Dynamic)
```

By default nothing changes — the collector keeps the top-`size` documents per sub-query, preserving score parity with the same hybrid query without collapse. Over the issue's dataset (6 groups on one shard, `groupA` owning the three top-scoring documents), `size: 5` returns 4 hits:

```
groupA, groupB, groupC, groupD        (groupE crowded out after deduplication)
```

Turning the setting on (dynamic, takes effect on the next request):

```json
PUT /my-index/_settings
{"index.neural_search.hybrid_collapse_distinct_groups_enabled": true}
```

makes the same search return 5 hits:

```
groupA, groupB, groupC, groupD, groupE
```

The two modes cannot be combined: they hand different document sets to per-sub-query normalization, so the same document can get different normalized scores under each mode — which is why this is a switch rather than a fix of the default.

**Implementation notes**
- `HybridCollapsingTopGroupsCollector` (new) — collects the top-`size` distinct groups per sub-query, mirroring the bookkeeping of Lucene's `FirstPassGroupingCollector`: a group map plus an ordered set once `size` groups exist, evicting the weakest group when a new competitive one arrives. When sorting by score, comparators are wrapped with `HybridLeafFieldComparator` so they read the sub-query's individual score rather than the sum. Emits one `FieldDoc` per surviving group, in the same `CollapseTopFieldDocs` encoding the existing collector uses — the coordinator-side normalization/deduplication pipeline is unchanged and no version gating is needed.
- `HybridCollectorFactory` — reads the setting and picks the collector; this is the only decision point, next to where the deprecated `hybrid_collapse_docs_per_group_per_subquery` setting is already read.
- `HybridCollectorManager` — registers the new collector type.
- `HybridCollapsingTopDocsCollector`, `HybridLeafFieldComparator` and the existing collector tests are byte-identical to `main` — the default path is untouched.

**Testing**
- Unit: `HybridCollapsingTopDocsCollectorTests` (default behavior, identical to `main`), `HybridCollapsingTopGroupsCollectorTests` (opt-in behavior), factory tests for the setting switch
- Integ: `HybridCollapseIT` covers both modes over the same skewed-groups dataset, including a test pinning the by-design default from the issue discussion

### Related Issues
Resolves #1947

### Check List


src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopGroupsCollector.java
```diff
@@ -0,0 +1,480 @@
...
+            if (isSortByScore) {
+                minScoreThresholds[subQueryIndex] = Math.max(minScoreThresholds[subQueryIndex], evictedScore);
+                compoundQueryScorer.getMinScores()[subQueryIndex] = Math.max(
+                    compoundQueryScorer.getMinScores()[subQueryIndex],
+                    evictedScore
+                );
+            }
+        }
+
+        private void collectExistingGroup(int doc, float score, CollectedGroup<T> group) throws IOException {
```
martin-gaievski: Seems per-leg representative election still splits a group's score. `CollectedGroup` holds `int topDoc`, each `SubQueryGroupCollector` is per sub-query and its score comparator is wrapped to read that leg's  individual score (`HybridLeafFieldComparator` fed by `setCurrentSubQueryScore(score)`). Election happens per leg in `collectExistingGroup` , same for `collectNewGroup`.
This scenario will be relatively common, it needs only 2+ legs and one group containing two documents that legs rank differently. Legs ranking documents differently is the point of hybrid search, so any multi-doc group will potential have this problem.

One way to fix this, and it's easier with the new opt-in mode - leg-independent election can go straight in new collector `collect(int doc)` method:

```java
// in the outer collect(), before the per-sub-query loop
float fusionProxy = 0f;
for (float s : subScoresByQuery) fusionProxy += s;   // leg-independent ranking key
```

Pass that down so every `SubQueryGroupCollector` elects the same doc id for a group while `group.score` stays the leg's own score. Later join adds legs for one doc id instead of splitting them.

src/test/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopGroupsCollectorTests.java
```diff
@@ -0,0 +1,1328 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+import static org.mockito.Mockito.mock;
+
+public class HybridCollapsingTopGroupsCollectorTests extends HybridCollectorTestCase {
```
martin-gaievski, line 49: please increase coverage, currently there is no test that can catch the missing functionality from my other code level comment, existing tests are only for per-leg representative election.
- shared helper `collectWithGroupDerivedScores` writes only `getSubQueryScores()[0]`, that's one leg, so there is nothing to disagree on
- `testCollapse_whenMultipleSubQueries_thenEachSubQueryHasResults` does build disagreeing legs ("high scores for even docs" / "high scores for odd docs"), so actual condition is present in its data. But its assertions are for existence-only, so it cannot fail on a wrong representative

Ideally would be great to have a two-leg IT asserting group order, not just group count

Ystk-hsn: Added in 5ff02f0b:

  - `testCollapse_whenLegsDisagreeWithinGroup_thenSameRepresentativeElectedAcrossLegs`, two legs ranking a group's documents differently, asserting the same representative in every leg with each leg's own score, and leg independent group order. This one fails on the previous code.
  - `testCollapse_whenGroupStrongInOneLegOnly_thenEvictionUsesSummedScore`, eviction is also leg independent.
  - `testCollapse_whenSortByFieldAndLegsMatchDifferentDocs_thenSameRepresentativeElected`, the same guarantee under field sort.
  - The two leg IT you suggested, `testCollapse_whenLegsDisagreeAndDistinctGroupsEnabled_thenGroupsOrderedByFusedScore`, legs scoring different fields and the full group order asserted.
  - `testCollapse_whenMultipleSubQueries_thenEachSubQueryHasResults` is updated to the new contract, a sub-query that matched none of the elected representatives emits an empty list while its total hits still count its own matches.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopGroupsCollector.java
Ystk-hsn: Fixed in 5ff02f0b. You were right, the per leg election split multi doc groups exactly as you described, and the new unit test fails on the previous code with leg 1 electing its own preference.
One simplification over the sketch. Instead of passing the proxy down to per sub-query collectors, the collector now runs a single election shared by all sub-queries. For score sort the comparators read `HybridSubQueryScorer#score()`, which is already the sum over sub-queries, so the proxy needed no extra plumbing, and field sort keys are document attributes so the same election covers that case too. Each sub-query then reports its own score for the one elected representative.

Two deliberate consequences worth calling out. A sub-query that did not match the elected representative leaves it out of its list, since a zero score entry would distort its min/max normalization stats. And group evictions no longer propagate per sub-query competitive score thresholds to `HybridSubQueryScorer#getMinScores()`, because a low score in one sub-query does not disqualify a document whose other sub-query scores make it the representative.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopGroupsCollector.java
```diff
@@ -0,0 +1,409 @@
...
+                return true;
+            } else if (compIDX == compIDXEnd) {
+                // Ties lose: docs are visited in doc id order
+                return false;
+            }
+        }
+    }
+
+    private void collectNewGroup(int doc, float[] subScoresByQuery) throws IOException {
+        if (groupMap.size() < numHits) {
```
martin-gaievski: If my understanding is correct, one `groupMap` with one budget of `numHits` shared by all sub-queries, and membership decided by the compound score the comparators. Because `HybridSubQueryScorer#score()` is the plain sum of un-normalized leg scores, leg with the larger score magnitude decides which groups survive, and `topDocs()` then emits nothing for a leg that didn't match any surviving representative. Example scenario: with leg 0 at 15/12/11 and leg 1 at 0.05/0.04 over disjoint groups and `numHits = 3`, leg 1 goes from `[gX, gY]` before this commit to `[]` after it.

What I was after is the outcome where the candidate set doesn't move: each sub-query still keeps its own
top-`numHits` groups ranked by its own score, and leg-independent key is used only to decide which
document represents a group. Concretely: keep a shared group → representative election, and layer the
per-sub-query top-`numHits` selection on top of it, ranking group `g` in leg `s` by leg `s` score for the
shared representative. 

If you'd rather keep the single shared election, then I think the setting's contract has to say so explicitly:
that enabling it makes the candidate set the top-`size` groups by *un-normalized score sum*, so a sub-query whose scores are smaller in magnitude may contribute no candidates at all. My concern with documenting rather than fixing is that it's true for nearly every multi-leg hybrid query, which would leave the setting recommending itself only for single-leg or same-scale queries.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopGroupsCollector.java
```diff
@@ -0,0 +1,409 @@
...
+            buildSortedSet();
+        }
+
+        int numComparators = comparators.length;
+        for (int subQuery = 0; subQuery < numSubQueries; subQuery++) {
+            List<FieldDoc> fieldDocs = new ArrayList<>();
+            List<Object> collapseValues = new ArrayList<>();
+            for (CollectedGroup<T> group : orderedGroups) {
+                float subQueryScore = group.scoresPerSubQuery[subQuery];
+                if (subQueryScore <= 0) {
```
martin-gaievski: sub-query's `scoreDocs` are no longer sorted descending by that sub-query's score. with this logic every sub-query now walks `orderedGroups` in the one shared (compound) order and stamps its own score. 
If you address previous comment and make the per-leg selection, then it fixes itself, each leg's ordered set would be keyed on its own score for the representative.
If the single shared election stays, then in `topDocs()` re-sort each leg's `fieldDocs` and `collapseValues` **together** by that leg's score descending, doc id ascending, and only when the request sorts by score

Please add two tests either way: unit test asserting each leg's `scoreDocs[i].score` is non-increasing (about 15 lines), and an IT with `technique: rrf` + collapse + the setting enabled on a single shard asserting group order, that combination has no coverage at all today.

martin-gaievski (commented):
Thanks for turning this around quickly, and one of my comment is genuinely fixed. Your new `testCollapse_whenLegsDisagreeWithinGroup_thenSameRepresentativeElectedAcrossLegs` really does fail on the pre-fix commit (`leg 1 must elect doc 0 for groupA expected:<0> but was:<1>`) and passes on this head, the whole surrounding unit suite is green, and the three new unit tests plus the two-leg IT are exactly the coverage I asked for. `HybridSubQueryScorer#score()` being the sum is the right key for a score sort, and you're right that field-sort keys are document attributes so one election covers that case too.

The part I need us to talk about is the simplification. My suggestion kept one group collector per sub-query and made only the choice of representative inside a group leg-independent. Collapsing that to a single shared election also collapses the candidate set: there is now one group budget of `numHits` for all sub-queries, and membership is decided by the **raw, un-normalized** sum of sub-query scores. Hybrid legs do not share a score scale — BM25 lands around 1–30, cosine/kNN around 0–1 — so the sum is dominated by the larger-scaled leg, and the elected set is effectively that leg's top-`size` groups.

Measured on the real collector, same input on both commits. Five disjoint groups, leg 0 scoring 15/12/11 on gA/gB/gC and leg 1 scoring 0.05/0.04 on gX/gY, `topNGroups = 3`:

 | | leg 0 emits | leg 1 emits |
 |---|---|---|
 | pre-fix (`1ba61673`) | `[gA, gB, gC]` | `[gX, gY]` |
 | this PR | `[gA, gB, gC]` | **`[]`** — 2 matched documents, 0 candidates |

`gX` was leg 1's rank-1 document; under `min_max` + `arithmetic_mean` it normalizes to `1.0` and would be a top-two result. On this head it is not in the response at all. That is the shape of essentially every real hybrid query, so I don't think we can treat it as an edge case — and it's the leg the setting is supposed to help, since enabling the setting is the only way to get the #1947 fix.

It also diverges from how the rest of hybrid search picks candidates: `HybridTopScoreDocCollector` keeps a per-sub-query `HitQueue(numOfHits)` and the default collapse collector keeps per-sub-query `FieldValueHitQueue`s. Same
 input through `HybridTopScoreDocCollector` as a control and leg 1 keeps `[3@0.05, 4@0.04]`.

There's a second consequence in breaking sorting invariant: because every leg now emits in one shared order, a sub-query's list is no longer sorted descending by that sub-query's own score, which `HybridQueryScoreDocsMerger` states as its precondition and which `rrf` depends on for its ranks.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopGroupsCollector.java
Ystk-hsn: Fixed in e86de6e6, went with the per-leg selection. Your scenario is a unit test now with the same numbers, failing on the previous commit with leg 1 collecting `[]` and passing here with leg 0 `[gA, gB, gC]`, leg 1 `[gX, gY]`.
Two existing tests pinned the old behavior, so their assertions flipped. `testCollapse_whenGroupStrongInOneLegOnly` now expects groupY to survive in leg 1, and the multi-sub-query test expects the odd-docs leg to keep its own five groups instead of an empty list. One rule is unchanged from the earlier round. A group appears in a sub-query's list only when the representative matched that sub-query, so the setting javadoc's "top-`size` distinct groups per sub-query" holds as written now.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopGroupsCollector.java
Ystk-hsn: With the per-leg selection this fixed itself. Both tests are in, the per-leg ordering unit test and the single-shard `rrf` + collapse IT. The IT fails on the previous commit with `[groupB, groupA, groupC, groupD]` and passes here with `[groupB, groupC, groupA, groupD]`. One side effect worth noting is that for a score sort the `FieldDoc` fields now carry the leg's own score instead of the sum, matching what the per-leg collector emitted before the opt-in rework.

Ystk-hsn:
@martin-gaievski Pushed e86de6e6 going with the per-leg selection, specifics on the two threads.

martin-gaievski (commented):
With the recent push the shared budget is gone, shared representative isn't. Example of failing scenario: when a group holds a lexical-only document and a vector-only document, election runs on the raw un-normalized sum. Assuming that BM25 wins, and vector leg — which never matched the winner — emits nothing for that group:

  - this PR      leg0 [gP, gQ] docs=[0,2] | leg1 []           totalHits=2
  - default      leg0 [gP, gQ] docs=[0,2] | leg1 [gP, gQ] docs=[1,3]
  - no collapse  leg0 [0@20.0, 2@18.0]    | leg1 [1@0.9, 3@0.8]

It is pinned by a test, but under an int field sort, where the election key is scale-free and the outcome is defensible; there's no score-sort equivalent. A second finding falls out of the same design: a leg's candidate set can end up permanently below numHits (measured: 1 group where numHits=2 and 3 documents matched).

Proposed fix, which removes both: make scoresPerSubQuery[s] running max over the group's documents in leg s rather than the representative's own score. Shared representative keeps topDoc so fusion still sees one document per group, but  every leg that matched anything keeps the group — and because a max never decreases, the under-fill and the re-keying wrinkle both vanish. code change should be limited to `collectExistingGroup`.

Ystk-hsn:
Thanks for all the reviews on this one. Made `scoresPerSubQuery` the running max over the group's documents.
Both findings are pinned. The lexical-only/vector-only group test fails on the previous commit with the vector leg collecting `[]`, and the candidate set one fails with 1 group where numHits=2. There's also an e2e version where the rating-only doc's evidence lifts its group under min_max.
Two notes. The running max applies under a field sort too, so the field-sort test you mentioned flipped. And it changed the `rrf` IT's per-leg ranks enough to make the old dataset an exact reciprocal-rank tie, so one rating value moved there.

martin-gaievski (commented):
Thanks for adressing my comments, I think you closed all the gaps I flagged there. I think with that request I made you add a new gap - running max is the wrong per-leg aggregate under `sort:[{"_score":"asc"}]`.
Under a reversed score sort the collector minimizes when electing a representative but maximizes when ranking and evicting groups, so it (a) drops the group that should rank first and (b) labels survivors with a score no returned document earned.

Example scenario:
One segment, one sub-query, keyword collapse, `Sort(SortField(null, SCORE, true))`, `numHits=1`, setting on.
- **Layout 1**: doc0 g=A 1.0, doc1 g=A 100.0, doc2 g=B 2.0: current main `[groupB doc=2 score=2.0]`; parent `e86de6e6` `[groupA doc=0 score=1.0]`; default collector `[groupA doc=0 score=1.0]`. Group owning the globally lowest-scoring document is evicted by `pollLast` on its max of 100.0 and then removed from `groupMap`.
- **Layout 2**: doc0 g=A 1.0, doc1 g=B 2.0, doc2 g=A 100.0: current main  `[groupA doc=0 score=100.0 fields0=100.0]`; parent and default `score=1.0`. doc0 scored 1.0; the reported score and the ascending re-sort key are 100.0.

this is a genuine correctness regression relative to the default path in a reachable, currently-working configuration.

One way to fix that: in `collectExistingGroup`, make the aggregate direction-aware: precompute the score component's index in the constructor and, fold with `Math.min` when `reversed[scoreIdx] == -1` and `Math.max` otherwise. A min needs its sentinel fixed too, seed to `Float.POSITIVE_INFINITY` and fold only over scores `> 0`, otherwise `min(x, 0.0)` would drop exactly the groups round-4a wanted kept; a monotone min preserves guarantees. 
Cheaper alternative if you would rather not carry two directions: keep the running max only when no score component is reversed and fall back to the representative's own score otherwise. 
Either way add one unit test built on `new SortField(null, SortField.Type.SCORE, true)`.

Ystk-hsn:
Made the fold direction-aware, min seeded with +Infinity over matched scores only when the score sort is reversed, max otherwise.
The unit test is built on `new SortField(null, SortField.Type.SCORE, true)` with your layout 1 numbers. It fails on the previous commit with `[groupB]` and passes here with groupA at 1.0.

martin-gaievski (approved):
Change looks good to me, thanks

Ystk-hsn:
@martin-gaievski Thanks a lot for the reviews. Resolved the changelog conflict, no code changes, should be good to go now.

martin-gaievski:
@vibrantvarun can you take a look please, IMO we now in a good shape with this change

martin-gaievski:
@Ystk-hsn please open an issue in https://github.com/opensearch-project/documentation-website/issues for documentation update related to this functionality and new setting

src/test/java/org/opensearch/neuralsearch/e2e/HybridCollapseIT.java
```diff
@@ -72,6 +83,345 @@ public void testCollapse_withMultipleShard_thenSuccessful() {
         testCollapse_whenE2E_withMinScore_thenSuccessful(0.5005f, 1, 2);
     }
 
+    @SneakyThrows
+    public void testCollapse_whenOneGroupOwnsMultipleTopDocsAndDistinctGroupsEnabled_thenDistinctGroupsReturned() {
```
vibrantvarun, line 87: Test-coverage gaps on the ON (distinct-groups) path — highest-value first:

1. Every distinct-groups IT uses NUMBER_OF_SHARDS_ONE, so the per-shard-top-size-groups → coordinator merge/dedup path has zero multi-shard coverage (the default path IS tested at multiple shards). Completeness is provable (a globally-top-size group is top-size on the shard where it peaks) and the emitted CollapseTopFieldDocs shape is identical to the default collector, so this is a coverage gap, not a suspected bug — but the downstream cross-shard dedup should be exercised with the distinct-groups shape, including the null-collapse-value bucket. Add a multi-shard ON IT.
2. Missing/null collapse-field value is not tested on the new collector: a doc lacking the field collapses into a single null group (groupMap.get(null)) and a null enters collapseValues. It's fails-open / no-NPE and matches the default collector, but it's untested on this path and puts a null into the downstream array — worth a case.
3. The hitsThresholdChecker early-termination path (CollectionTerminatedException + GREATER_THAN_OR_EQUAL_TO relation) is never triggered — the threshold (1001) exceeds every test dataset. And there's no single test that queries, PUTs the Dynamic setting, and re-queries the same index; both modes are proven only on separately-created indices.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopGroupsCollector.java
```diff
@@ -0,0 +1,470 @@
...
+            List<FieldDoc> fieldDocs = new ArrayList<>();
+            List<Object> collapseValues = new ArrayList<>();
+            for (CollectedGroup<T> group : orderedGroupsPerSubQuery.get(subQuery)) {
+                float subQueryScore = group.scoresPerSubQuery[subQuery];
+                Object[] fields = new Object[numComparators];
+                for (int k = 0; k < numComparators; k++) {
+                    // Score components carry this sub-query's own score; other components are attributes of the representative
+                    fields[k] = isScoreSortField[k] ? subQueryScore : comparators[k].value(group.comparatorSlot);
+                }
+                fieldDocs.add(new FieldDoc(group.topDoc, subQueryScore, fields));
```
vibrantvarun, line 180: Worth an explicit confirmation (not a bug): this emits a FieldDoc whose `doc` is the group's single elected representative (the summed-score winner) but whose `score` is this sub-query's running-max, which can come from a *different* document in the group (see the leg-exclusive-docs case). The default collector always emits a (doc, score) pair from the same document. The design intends this — one representative per group so downstream per-doc fusion aligns all legs on one docId — and testCollapse_whenLegExclusiveDocsAndDistinctGroupsEnabled pins the fused ordering, but please confirm the coordinator CollapseDataCollector / normalization keys on docId+collapseValue+score and never assumes `score` belongs to `doc`. That's the one place a hidden assumption could bite, and it isn't asserted at the coordinator boundary.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
```diff
@@ -82,6 +82,21 @@ public final class NeuralSearchSettings {
...
 
+    /**
+     * When enabled, a hybrid query with collapse collects the top {@code size} distinct groups per sub-query
+     * (each represented by its most competitive document), so the response contains {@code size} groups whenever
+     * that many exist. When disabled (default), the collector keeps the top {@code size} documents per sub-query,
+     * which preserves score parity with the same hybrid query without collapse but can return fewer groups than
+     * {@code size} when one group owns several top-scoring documents. The two behaviors are mutually exclusive;
+     * see https://github.com/opensearch-project/neural-search/issues/1947 for the design discussion.
+     */
+    public static final Setting<Boolean> HYBRID_COLLAPSE_DISTINCT_GROUPS_ENABLED = Setting.boolSetting(
```
vibrantvarun, line 98: Two behavioral caveats worth adding to this setting's javadoc (both LOW, neither crashes):
1. It's Dynamic, read per-request at collector creation — flipping it between page 1 and page 2 of the same paginated search yields inconsistent page semantics (deduplicated-docs vs distinct-groups). Inherent to a dynamic setting, but worth calling out.
2. Because it's IndexScope, a single _search spanning an index with it enabled and one with it disabled runs different collectors per shard and merges heterogeneous CollapseTopFieldDocs into one response — structurally compatible and non-crashing, but semantically mixed. Worth a documented note that the two modes shouldn't be mixed across indices in one request.

vibrantvarun (commented):
Reviewed across three lenses (logic/line-by-line, adversarial edge-cases, tests) and cross-checked the new collector against Lucene FirstPassGroupingCollector and the existing HybridCollapsingTopDocsCollector. No correctness bug, no crash, no wrong/dropped/duplicated groups under anything I could construct — the collector is a correct generalization of FirstPassGroupingCollector to N per-sub-query top-size sets sharing one representative election. Verified: slot capacity (numSubQueries*numHits+2) is exactly the transient peak so freeSlots.pop() can't underflow; the remove-before-mutate TreeSet invariant is correct (a group is removed from exactly the member sets whose key changes, and foldScore is a no-op precisely when it isn't removed); election reads the summed score while the emitted FieldDoc score is the per-leg running max via the isScoreSortField guard (no summed-score leak); evict-then-readmit is self-correcting for ranking/membership (competitive threshold is monotonic); and a globally-top-size group cannot be missed on every shard. No regression — the default path (HybridCollapsingTopDocsCollector, HybridLeafFieldComparator, its tests) is byte-identical to main and only reached when the setting is off. Both central claims are oracle-anchored (default score-parity vs an independent no-collapse run; both modes pinned to exact 4-hit/5-hit group sets).

Everything below is LOW/MED — non-blocking, several shared with the existing default collector. Inline: the multi-shard test gap, the doc/score downstream confirmation, and two doc caveats. Nits not inline: (a) the CHANGELOG entry is under 'Bug Fixes' but this is an opt-in enhancement (new setting) — please move it to the enhancements/features section; (b) the slot-capacity math is tight-but-correct and would be worth an explicit boundary stress test; (c) getMaxScore() returns the max single-sub-query score not the summed election score (harmless, matches the default collector); (d) the TreeSet tiebreak `o1.topDoc - o2.topDoc` is safe here (doc ids bounded) but Integer.compare(...) is the safer idiom.

src/test/java/org/opensearch/neuralsearch/e2e/HybridCollapseIT.java
Ystk-hsn: Added in the latest commit. A five-shard distinct-groups IT over the skewed dataset plus a document with no collapse field, asserting all six groups come back through the cross-shard dedup, the null bucket included. A same-index toggle test, default 4 hits, flip on 5, flip back 4. An early-termination unit test pinning the `GREATER_THAN_OR_EQUAL_TO` relation, and a collector-level null-group unit test.

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopGroupsCollector.java
Ystk-hsn: Went through the coordinator side. The per-document leg scores never leave the shard, a `CollapseTopFieldDocs` entry only carries doc, score and fields, so there is nothing there to re-derive a document's own score from. `CollapseDataCollector` keys on the collapse value and compares whole FieldDocs, min_max takes the (doc, score) pairs as they are, and rrf only looks at positions. If a wrong pairing ever slipped in shard-side, the leg-exclusive IT would flip its fused order, groupP sits above groupQ only because of the rating evidence.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
Ystk-hsn: Both caveats are in the setting's javadoc now, the per-request read under pagination and not mixing the two modes across indices in one request.

Ystk-hsn:
@vibrantvarun Thanks a lot for the detailed review. Everything is addressed in the latest commit, specifics on the threads. From the notes outside the threads, the changelog entry moved to Enhancements, would appreciate if you could change the label on this PR on your end to match. The tiebreak is `Integer.compare` now, and the slot capacity has an explicit boundary stress test, two legs at full occupancy pushing new groups through the transient slot.

@martin-gaievski opened opensearch-project/documentation-website#13130 for the docs update, planning to raise the docs PR once this one merges.

vibrantvarun (approved):
Re-reviewed the `Close the review's test-coverage gaps and document the setting's caveats` commit — every item I raised is addressed, verified against the diff.

- **Multi-shard coverage**: `testCollapse_whenDistinctGroupsEnabledOnMultipleShards_thenCoordinatorMergesAllGroups` runs the distinct-groups path over five shards on the skewed dataset with a no-collapse-field document, and asserts all six groups — the null bucket included — come back through the cross-shard dedup. The same-index toggle (`...ToggledOnSameIndex...`, 4 → 5 → 4) and the collector-level null-group test (`...whenDocumentHasNoCollapseField_thenItFormsTheNullGroup`) close the rest of that gap.
- **doc/score pairing**: thanks for tracing the coordinator boundary — `CollapseTopFieldDocs` carries only (doc, score, fields), `CollapseDataCollector` keys on the collapse value and compares whole `FieldDoc`s, min_max takes the pairs as they are and rrf looks only at positions, so the representative's per-leg `score` can't be misattributed downstream. That was exactly the confirmation I was after (not a bug).
- **Setting javadoc**: both caveats are in — the dynamic per-request read under pagination, and not mixing the two modes across indices in one request.
- Also picked up the two nits: the TreeSet tiebreak is now `Integer.compare(o1.topDoc, o2.topDoc)` (overflow-safe, behavior-identical), and there's an explicit slot-capacity stress test (`...whenAllComparatorSlotsOccupied...`) plus an early-termination test pinning the `GREATER_THAN_OR_EQUAL_TO` relation.

The only production change since my review is the overflow-safe comparator swap, so no regression risk. The collector's correctness (single shared representative election, per-sub-query candidacy, slot capacity, evict-then-readmit) I validated in the earlier pass and still holds. LGTM.

Ystk-hsn:
@vibrantvarun Thanks for the review!
@martin-gaievski Rebased to resolve the changelog conflict, no code changes. Would also appreciate if you could switch the PR label to enhancement on your end to match the changelog section!

martin-gaievski:
> @vibrantvarun Thanks for the review! @martin-gaievski Rebased to resolve the changelog conflict, no code changes. Would also appreciate if you could switch the PR label to enhancement on your end to match the changelog section!

sure, switched label and restarted CI, once main actions are green I'll merge

PR #1432, opened 2025-07-08
Handle a hybrid query extended with DLS rules by the security plugin

kt-eliatra:
### Description
Handle a hybrid query extended with DLS rules by the security plugin

### Related Issues
Resolves #1303

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/neural-search/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1957, opened 2026-08-20
Add hybrid DLS cross-plugin integration tests

sharp-pixel:
### Description

Adds model-free, cross-plugin integration coverage for document-level security
applied to Neural Search hybrid queries and completes filter traversal for
`NeuralQueryBuilder`.

`NeuralQueryBuilder#visit(QueryBuilderVisitor)` now exposes its embedded
`queryfilter` through `visitor.getChildVisitor(BooleanClause.Occur.FILTER)`.
The default `QueryBuilder#visit` implementation visits only the builder itself,
so the override is required for visitors to traverse the complete neural query
tree. This aligns neural queries with other composite query builders, including
`KNNQueryBuilder`.

The companion Security change can therefore use the public `filter()` and
`visit()` contract to discover and verify embedded filters without reflecting
on Neural Search or k-NN getter names. Unit tests cover traversal both with and
without an embedded filter.

The secure integration test verifies that:

- an administrator sees all matching documents while a DLS-restricted user sees
  only allowed documents;
- DLS protects hybrid hits and global aggregations;
- suggestion-bearing hybrid requests remain compatible with DLS and return
  DLS-filtered hits plus a valid suggestion response;
- an existing hybrid filter is combined with DLS;
- a single-clause hybrid query remains supported; and
- hybrid k-NN and neural clauses receive the DLS filter.

The suggestion assertion covers request compatibility, not filtering of term
dictionary candidates. Candidate filtering retains the Security plugin's
existing behavior and is not changed by either PR.

The developer guide documents how to test unpublished Security plugin changes by
publishing the Security plugin ZIP to Maven Local. Neural Search continues to
resolve the fixed Security artifact coordinate matching `opensearch_build`; it
does not accept an arbitrary plugin archive path.

> [!IMPORTANT]
> This draft depends on
> [opensearch-project/security#6428](https://github.com/opensearch-project/security/pull/6428).
> It should not be merged until that Security change is merged and available to
> the Neural Search integration-test build.

This builds on the adaptive hybrid DLS support merged in
[opensearch-project/security#6416](https://github.com/opensearch-project/security/pull/6416)
and complements the hybrid filter support introduced by
[neural-search#1432](https://github.com/opensearch-project/neural-search/pull/1432).

### Related Issues

Related to #1303.

Blocking dependency:
[opensearch-project/security#6428](https://github.com/opensearch-project/security/pull/6428).

### Testing

Published the Security plugin from the companion PR to Maven Local:

```bash
cd /path/to/opensearch-security
./gradlew publishPluginZipPublicationToMavenLocal
```

Then ran the focused test on a three-node Neural Search cluster:

```bash
cd /path/to/neural-search
./gradlew :integTest \
  --tests org.opensearch.neuralsearch.query.HybridQueryDlsIT \
  -x spotlessApply \
  -Dsecurity.enabled=true \
  -PnumNodes=3 \
  --refresh-dependencies
```

Five of the six cross-plugin tests pass. The dense-neural test is currently
blocked during ML model registration by
[opensearch-project/security#6433](https://github.com/opensearch-project/security/issues/6433),
before the neural query executes.

The visitor unit tests and formatting checks pass:

```bash
./gradlew test --tests "org.opensearch.neuralsearch.query.NeuralQueryBuilderTests"
./gradlew spotlessCheck
```

### Changelog

No changelog entry is proposed because the visitor override completes the
existing `QueryBuilder` traversal contract and the remaining changes add test
coverage and developer documentation. Please apply the `skip-changelog` label.

### Check List

- [x] API changes companion pull request — not applicable; this implements an
  existing public visitor contract without adding API surface.
- [x] Public documentation issue/PR — not applicable; this is developer-only
  test infrastructure and an existing query traversal contract.

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.


kkhatua:
@sharp-pixel 
Can we have this out of draft and ready for review? We'll need both PRs to go in simultaneously.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryDlsIT.java
```diff
@@ -0,0 +1,530 @@
...
+            .collect(Collectors.toMap(bucket -> (String) bucket.get("key"), bucket -> ((Number) bucket.get("doc_count")).intValue()));
+
+        assertEquals(
+            expectedBuckets.values().stream().mapToInt(Integer::intValue).sum(),
+            ((Number) allDocuments.get("doc_count")).intValue()
+        );
+        assertEquals(expectedBuckets, actualBuckets);
+    }
+
+    private void assertSuggestionContains(Map<String, Object> responseBody, String suggestionName, String expectedOption) {
```
martin-gaievski, line 466: please add `assertSuggestionDoesNotContain(...)` for blocked terms, or add an explicit comment explaining why absence is intentionally not asserted. current method does not assert that blocked-only terms ("confidential", "classified") are absent, if security plugin regresses and stops filtering term-dictionary candidates, this test will still pass.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryDlsIT.java
```diff
@@ -0,0 +1,530 @@
...
+    private static final Set<String> PUBLIC_DOCUMENT_IDS = Set.of("allowed-alpha", "blocked-alpha", "blocked-beta");
+    private static final Set<String> PUBLIC_ALLOWED_DOCUMENT_IDS = Set.of("allowed-alpha");
+
+    @BeforeClass
+    public static void requireSecurityPlugin() {
+        assumeTrue("requires the Security plugin", isSecurityPluginEnabled());
+    }
+
+    @Before
+    public void setUpDlsResources() throws IOException {
```
martin-gaievski, line 61: After creating DLS role and user, test immediately searches as the restricted user. This is potential race window on a 3-node cluster (`-PnumNodes=3`). security plugin propagates config changes synchronously via its internal cache, if config hasn't reached the node handling the request results are unreliable.
I suggest you call `DELETE /_plugins/_security/api/cache` after user creation, or add a retry loop around the first restricted search.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryDlsIT.java
```diff
@@ -0,0 +1,530 @@
...
+                "/" + INDEX_NAME
+            )
+        );
+    }
+
+    private static boolean isSecurityPluginEnabled() {
+        return Boolean.parseBoolean(System.getProperty("security.enabled", "false"));
+    }
+
+    public void testDlsFiltersHybridHitsAndAggregations() throws Exception {
```
martin-gaievski, line 100: that's six different cases asserted in one method. Consider splitting it into smaller individual methods, reasoning is - if first assertion fails we don't know if other five are good or not. Individual test methods will show that (fail or pass individually).

src/test/java/org/opensearch/neuralsearch/query/HybridQueryDlsIT.java
sharp-pixel: Thanks for flagging this. I traced the Security REST update path: role/user PUTs use `RefreshPolicy.IMMEDIATE`, then execute `ConfigUpdateAction` across all nodes and return success only after every node reload completes without failure.

Therefore, the subsequent restricted search should not have a propagation race. `DELETE /_plugins/_security/api/cache` would repeat the same cluster-wide update, while retrying could hide a real propagation regression. I propose leaving the immediate assertion unchanged. If you have observed a specific flaky case despite this synchronization, please share it and I will investigate further.

References: [update path](https://github.com/opensearch-project/security/blob/99866a357bf214edb6ed06cced94dfe8bd5dc7bf/src/main/java/org/opensearch/security/dlic/rest/api/AbstractApiAction.java#L511-L566), [per-node reload](https://github.com/opensearch-project/security/blob/99866a357bf214edb6ed06cced94dfe8bd5dc7bf/src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java#L128-L145).

src/test/java/org/opensearch/neuralsearch/query/HybridQueryDlsIT.java
sharp-pixel: Addressed in [`9cdb0d2`](https://github.com/sharp-pixel/neural-search/commit/9cdb0d2). I made the comment explicit that blocked-only suggestion candidates are intentionally not asserted because term-dictionary filtering is existing Security plugin behavior outside this regression test scope. This test continues to verify suggestion-bearing request compatibility, DLS-filtered hits, and a valid suggestion response.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryDlsIT.java
sharp-pixel: Agreed and addressed in [`9cdb0d2`](https://github.com/sharp-pixel/neural-search/commit/9cdb0d2). I split the combined method into four behavior-focused tests covering hits/aggregations, suggestions, an existing hybrid filter, and a single-clause hybrid query. Admin and restricted-user requests remain paired where the admin result serves as the control. All four tests pass independently on the three-node Security-enabled cluster.

src/main/java/org/opensearch/neuralsearch/query/NeuralQueryBuilder.java
```diff
@@ -525,6 +527,19 @@ public QueryBuilder filter(QueryBuilder filterToBeAdded) {
...
+    /**
+     * Exposes the embedded query filter as a {@link BooleanClause.Occur#FILTER} child. The default
+     * {@link QueryBuilder#visit(QueryBuilderVisitor)} implementation visits only this builder, so composite query
+     * builders must override it to make their complete query tree available to visitors.
+     */
+    @Override
+    public void visit(QueryBuilderVisitor visitor) {
+        visitor.accept(this);
+        if (queryfilter != null) {
+            queryfilter.visit(visitor.getChildVisitor(BooleanClause.Occur.FILTER));
```
vibrantvarun: Correct override — it recurses (`queryfilter.visit(getChildVisitor(FILTER))`), so it descends the full filter subtree, matching `KNNQueryBuilder` and unlike core's `boosting`/`script_score` which hand their child to `accept()` without recursing.

Worth flagging a beneficial side effect on the `feature/fusion-hybrid-query` branch: the fused-mode guards (`countFusedLegSearches` / `requireFusedQueryIsPartOfRequestQuery`) walk queries via `visit`, and before this a fused hybrid embedded in a neural query's `filter` was invisible to them — counted as zero legs, i.e. a potential `max_leg_searches` blind spot. Since those are the same `visit` traversal, this override closes that gap too. Might be worth a one-line test on that branch after it rebases, but nothing needed here.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryDlsIT.java
```diff
@@ -0,0 +1,537 @@
...
```
vibrantvarun, line 100: Following up on the propagation-timing thread: the `RefreshPolicy.IMMEDIATE` + synchronous `ConfigUpdateAction` rationale is sound for the happy path, so no change needed now. If this ever flakes on `-PnumNodes=3` (the restricted-user search running before the DLS role/user is visible on the node that serves it), the low-cost hardening is to wrap the first restricted-user assertion in `assertBusy(...)` so it retries rather than failing once. Just noting the fallback in case CI surfaces it.

vibrantvarun (commented):
The production change (NeuralQueryBuilder#visit) is correct and well-tested; the rest is a cross-plugin DLS IT + dev-guide docs, and martin's three comments are all addressed (method split, explicit suggestion-scope comment, and a traced rationale for the propagation timing). Posting COMMENT rather than APPROVE only because the PR itself gates merge on Security #6428, which isn't merged yet — the cross-plugin IT can't build/run against the required Security artifact until then. Two notes inline.

sharp-pixel:
> The production change (NeuralQueryBuilder#visit) is correct and well-tested; the rest is a cross-plugin DLS IT + dev-guide docs, and martin's three comments are all addressed (method split, explicit suggestion-scope comment, and a traced rationale for the propagation timing). Posting COMMENT rather than APPROVE only because the PR itself gates merge on Security #6428, which isn't merged yet — the cross-plugin IT can't build/run against the required Security artifact until then. Two notes inline.

The security plugin PRs have been merged

vibrantvarun (approved):
LGTM. Now that the Security dependency (#6428) is merged, the gate that had this at COMMENT is lifted. Code is clean: NeuralQueryBuilder#visit correctly recurses the embedded filter (KNN-aligned, unit-tested), and the DLS IT's assertions are DLS-meaningful — I cross-checked every case against the dataset (hits, DLS-filtered global aggregation, existing-filter combination, single-clause, knn, neural, mixed), all with total/count/set + dedup checks that fail loudly if DLS isn't applied to a vector clause, so no false-greens. Model/pipeline/index cleanup is handled by the IT plus BaseNeuralSearchIT#cleanUp. One non-blocking note stands (the restricted-search-immediately-after-role-PUT is safe per the author's traced ConfigUpdateAction synchronization; assertBusy only if CI ever flakes on -PnumNodes=3). Please make sure CI re-runs against the merged Security snapshot before merging so the cross-plugin DLS IT validates end-to-end.

vibrantvarun:
@martin-gaievski can you also check this once.

martin-gaievski (commented):
Change looks good to me. Let's wait for https://github.com/opensearch-project/security/pull/6428 to be merged, then I'll approve and merge this PR. I assume with #6428 CI in neural-search should pass

martin-gaievski (approved):

PR #1960, opened 2026-08-21
[z_score] Compute the per-subquery statistics in one pass instead of four

dbwiddis:
### Description

`ZScoreNormalizationTechnique.getZScoreResults` needs four statistics per sub query — mean, standard deviation, max and min. Each had its own `calculate*PerSubquery` helper, and **every one of those independently called `calculateStatsPerSubquery`**:

```java
float[] maxPerSubquery = calculateMaxScorePerSubquery(queryTopDocs, numOfSubqueries);
float[] minPerSubquery = calculateMinScorePerSubquery(queryTopDocs, numOfSubqueries);
float[] meanPerSubQuery = calculateMeanPerSubquery(queryTopDocs, numOfSubqueries);
float[] stdPerSubquery = calculateStandardDeviationPerSubquery(queryTopDocs, numOfSubqueries);
```

So a single normalization walked every hit four times and built four sets of N `DescriptiveStatistics`, each retaining a copy of every score in a resizable `double[]`. This builds the `DescriptiveStatistics[]` once and reads all four statistics off it.

`explain()` goes through `getZScoreResults` too, so it benefits identically.

### Why this is behaviour-preserving

`DescriptiveStatistics` accumulates in insertion order, and the walk order is unchanged, so the statistics — and therefore every normalized score — are bit-identical to before. Not "within a delta": bit-identical.

### Measurements

Allocation, via `ThreadMXBean.getThreadAllocatedBytes` over the modelled computation, 1000 iterations after warmup. The ratio converges on 4.00x as hit counts grow, which is what you'd expect from collapsing four identical passes into one:

| fixture | four-pass | one-pass | reduction |
|---|---|---|---|
| 1 shard × 2 subqueries × 10 hits | 7,280 B | 2,024 B | 3.60x |
| 3 shards × 2 subqueries × 100 hits | 71,264 B | 17,912 B | 3.98x |
| 5 shards × 3 subqueries × 500 hits | 795,552 B | 199,008 B | 4.00x |
| 10 shards × 5 subqueries × 1000 hits | 5,258,592 B | 1,314,792 B | 4.00x |

I'm quoting allocation rather than latency because it is the reproducible signal here; wall-clock on a fixture this small is dominated by noise.

### Testing

New `testNormalization_whenStatsComputedInOnePass_thenBitIdenticalToFourPassReference` builds randomized `CompoundTopDocs` across 1–3 shards and 2–4 sub queries, computes each statistic via four fully independent walks (the pre-change shape), applies the z-score formula, and asserts `Float.floatToIntBits` equality against `normalize()`'s output.

I mutation-checked it rather than assuming it bites — mutating `getMean`, `getStandardDeviation`, or `getMax` in the production code all fail the test.

One honest gap, called out in the test's javadoc: **min is not pinned by any end-to-end assertion, and cannot be.** `minScore` is only read on the `standardDeviation == 0` branch, but a zero standard deviation means every score equals the mean, so the preceding `mean == score` check returns `maxScore` first. That branch is unreachable for any realistic score. Pre-existing, unchanged here, and noted only so the "bit-identical" claim isn't read as broader than it is.

The existing `ZScoreNormalizationTechniqueTests` are untouched and pass, as does the wider `*Normaliz*` / `*Combination*` unit sweep (404 tests, 27 classes).

### Related Issues

Groundwork for #1941 — the shared `z_score` core that both the classic shard-side path and coordinator fused mode will call needs this cleaned up first. Mirrors #1942, which did the same for RRF's rank arithmetic.

### Check List
- [x] Functionality includes testing.
- [x] API changes companion pull request [created], if applicable.
- [x] Public documentation issue/PR [created], if applicable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,3 +25,4 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Refactoring
 * [SemanticHighlighter] Traverse the query tree with a QueryBuilderVisitor instead of a "manual" walk ([#1915](https://github.com/opensearch-project/neural-search/pull/1915))
 * [RRF] Extract the rank arithmetic into a shared RRFScoreNormalizer and hoist the workflow duplicated between NormalizationProcessor and RRFProcessor into AbstractScoreHybridizationProcessor ([#1944](https://github.com/opensearch-project/neural-search/pull/1944))
+* [z_score] Compute the per-subquery mean, standard deviation, max and min in a single DescriptiveStatistics pass instead of four, reducing normalization allocation by 4x ([#1960](https://github.com/opensearch-project/neural-search/pull/1960))
--- a/src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
+++ b/src/main/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizationTechnique.java
@@ -126,28 +126,6 @@ public class ZScoreNormalizationTechnique implements ScoreNormalizationTechnique
         return getDocIdAtQueryForNormalization(normalizedScores, this);
     }
 
-    private static float[] calculateMaxScorePerSubquery(final List<CompoundTopDocs> queryTopDocs, final int numOfSubqueries) {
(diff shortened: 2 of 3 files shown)


bzhangam (approved):

owaiskazi19 (approved):
Thanks for the improvement

PR #1961, opened 2026-08-21
Extract the shared z_score and l2 normalization cores

dbwiddis:
> [!IMPORTANT]
> Stacked on #1960. Until that merges this PR's diff also shows #1960's commit — review only `Extract the shared z_score and l2 normalization cores`, or use the [compare view against #1960's branch](https://github.com/dbwiddis/neural-search/compare/perf/zscore-single-pass-stats...dbwiddis:refactor/score-normalizer-shared-cores). I'll rebase once #1960 lands.

### Description

`min_max` and RRF each have a shared arithmetic core that both the classic shard-side path and coordinator-side fused mode call — `MinMaxScoreNormalizer` (#1933) and `RRFScoreNormalizer` (#1944). `z_score` and `l2` do not: both keep their formulas in `private` methods inside their classic technique classes.

Fused mode needs the same formulas. A second copy of each would be a standing source of classic-vs-fused score drift, so this extracts them first. After this, any classic-vs-fused difference can only be the hit set, never a second copy of the formula.

Adds `ZScoreNormalizer` and `L2ScoreNormalizer` next to the existing cores, and has the classic techniques delegate to them — exactly as `MinMaxScoreNormalizationTechnique.normalizeSingleScore` already delegates to `MinMaxScoreNormalizer`.

### Why each core has two halves

The two paths reach the arithmetic differently: classic streams scores as it walks `CompoundTopDocs`, while the coordinator already holds a complete per-leg score map. So each core exposes the aggregate step as well as the scalar formula:

- **`ZScoreNormalizer.StatsAccumulator`** wraps `DescriptiveStatistics` and exposes `mean`/`standardDeviation`/`max`/`min`. Sharing the accumulator rather than only the final formula means both paths share the *definition* of those statistics, including the `double`→`float` narrowing. Values go in one at a time so classic needn't materialize an intermediate collection.
- **`L2ScoreNormalizer.NormAccumulator`** accumulates the sum of squares, plus an `l2Norm(Collection<Float>)` convenience for the coordinator's single-pass case.

### The one subtlety worth reviewer attention

**The `l2` running sum stays a `float`.** That is what the classic path has always done (`float[] l2Norms`). Widening it to `double` would round differently and silently change every existing `l2` score, so it is deliberately preserved and there's a test pinning it.

Float addition is also not associative, so the norm depends on accumulation order. That's documented on `NormAccumulator`, because classic accumulates per shard while the coordinator will accumulate over the merged leg — on a single shard the two agree bit-for-bit, but across shards `l2` can differ in the last bit. Flagging it here so it's a known, documented property before fused `l2` is wired up rather than a surprise later.

### Testing

Move-only otherwise: no formula, constant, or edge case is altered.

- **The existing `ZScoreNormalizationTechniqueTests` and `L2ScoreNormalizationTechniqueTests` are untouched** and are the regression net. They exercise `normalize()` and `explain()` end to end and never touch the private arithmetic, so a green run is the proof the move changed nothing.
- New `ZScoreNormalizerTests` (7) and `L2ScoreNormalizerTests` (6) covering the cores directly, mirroring `MinMaxScoreNormalizerTests`. Two assert bit-level equality rather than a delta: that the incremental accumulator and the collection convenience agree exactly, and that the sum of squares is accumulated in `float`.
- Full unit suite: 2650 tests, green.

### Incidental removals

`ZScoreNormalizationTechnique.SINGLE_RESULT_SCORE` was already dead before this change and goes with it; the two `MIN_SCORE` constants move to the cores that now own them.

### Related Issues

Groundwork for #1941. Mirrors #1944, which did the same extraction for RRF.

### Check List
- [x] Functionality includes testing.
- [x] API changes companion pull request [created], if applicable.
- [x] Public documentation issue/PR [created], if applicable.


src/test/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizerTests.java
```diff
@@ -0,0 +1,58 @@
...
+        List<Float> scores = List.of(0.5f, 0.25f, 1.75f, 0.125f);
+        L2ScoreNormalizer.NormAccumulator accumulator = new L2ScoreNormalizer.NormAccumulator();
+        for (float score : scores) {
+            accumulator.add(score);
+        }
+
+        assertEquals(Float.floatToIntBits(L2ScoreNormalizer.l2Norm(scores)), Float.floatToIntBits(accumulator.norm()));
+    }
+
+    public void testNormAccumulator_accumulatesInFloat() {
```
martin-gaievski, line 46: seems this test cannot fail. First,`expectedFloatSum` is itself a float expression, so the expectation mirrors the implementation, any accumulator computing the same product-sum passes, in float or double.
And even a genuine float-vs-double comparison would pass on this fixture. For `{1.0f, 1e-4f}` both give bits well under `ulp(1.0)/2`, addend is swallowed either way.

src/test/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizerTests.java
```diff
@@ -0,0 +1,66 @@
...
+        assertEquals(ZScoreNormalizer.MIN_SCORE, ZScoreNormalizer.normalizeSingleScore(2.0f, 2.0f, 5.0f, 9.0f, 1.0f), DELTA);
+    }
+
+    public void testStatsAccumulator_exposesMeanStandardDeviationMaxAndMin() {
+        ZScoreNormalizer.StatsAccumulator accumulator = new ZScoreNormalizer.StatsAccumulator();
+        for (float score : new float[] { 2.0f, 7.0f, 8.0f }) {
+            accumulator.add(score);
+        }
+
+        // Population statistics as DescriptiveStatistics defines them: mean 5.667, sample standard deviation 3.215.
```
martin-gaievski: asserted value is sample (n−1) statistic. Since z_score's behaviour hinges on convention
`DescriptiveStatistics` uses, this is precisely the comment reader will trust. IMO drop "Population" and name sample (n−1) convention explicitly.

src/test/java/org/opensearch/neuralsearch/processor/normalization/L2ScoreNormalizerTests.java
dbwiddis: Right on both counts -- the expectation was a float expression mirroring the implementation, and `{1.0f, 1e-4f}` can't discriminate anyway, since `1e-8` sits under half an ulp of `1.0f`.

Kept the test rather than dropping it, since the float running sum is the behaviour this refactor has to preserve. The fixture is now `1.0f` followed by 100 addends of `2.0e-4f`, each squaring to `4.0e-8`, just under half an ulp: a float sum swallows all 100 and stays exactly `1.0f`, a double sum reaches `1.000004` and narrows to `1.000002f`. Expectation is the literal `1.0f`, and an `assertNotEquals` against a double-accumulated reference guards the fixture so it fails if that margin ever stops holding.

Confirmed by flipping `NormAccumulator.sumOfSquares` to `double`: `expected:<1065353216> but was:<1065353233>`. Fixed in a86ed5cf.

src/test/java/org/opensearch/neuralsearch/processor/normalization/ZScoreNormalizerTests.java
dbwiddis: Agreed -- and the old comment contradicted itself, calling them population statistics and then asserting the sample standard deviation.

It now names the convention explicitly: `DescriptiveStatistics#getStandardDeviation` is the sample (n-1, bias-corrected) value, `sqrt(20.667/2) = 3.215` for `{2, 7, 8}`, with the population `sqrt(20.667/3) = 2.625` given for contrast so the asserted number is unambiguous. Fixed in a86ed5cf.

martin-gaievski (approved):
Looks good, thank you

PR #1962, opened 2026-08-21
[Fusion] Wire z_score and l2 into fused mode

dbwiddis:
### Description

Widens the coordinator-side (fused) mode of the `hybrid` query from `min_max` only to the whole score-normalization family — `min_max`, `z_score`, `l2` — with `arithmetic_mean`. A user who already runs a `z_score` or `l2` `normalization-processor` pipeline can now turn on `fusion` without touching their config.

Both new normalizers are thin `ScalarNormalizer` wrappers over the shared cores extracted in #1961, which the classic shard-side path also calls — so there is one implementation of each formula, not two. Adding a technique stays exactly what `ScalarNormalizers`' javadoc promises: one class plus one registry entry, with no change to `CoordinatorScoreFusion` or the orchestrator.

> [!IMPORTANT]
> Stacked on #1943 (`feature/fusion-advanced-query-features`), which is not merged yet, plus #1960 and #1961. The diff below therefore includes those commits; **the reviewable change is the single top commit `e83a6cb4`**. Once #1943, #1960 and #1961 land, this branch gets rebased onto `feature/fusion-hybrid-query` and the diff drops to that one commit.

| commit | what | review? |
|:--|:--|:--|
| `8b2605b2` | from #1960 | no |
| `95bba47f` | from #1961 | no |
| `e83a6cb4` | **this PR** | yes |

### The gate

`requireSupportedTechniques` was an equality test against the two supported names. It becomes three checks:

1. **Fused combination scope** — a `Set` that currently holds only `arithmetic_mean`. This runs *first* on purpose: RRF is modelled as `combination=rrf` with `normalization=none`, so a registry-first order would tell an RRF user that `none` is an unsupported *normalization*, blaming a clause they never wrote.
2. **Normalizer availability** — `ScalarNormalizers.supportedTechniques()`, so the gate and the registry cannot drift apart, and the error names the alternatives.
3. **The classic compatibility matrix** — read through a new static `ScoreNormalizationFactory.supportedCombinationTechniques(String)` rather than copied, so fused mode cannot admit a pairing classic forbids. `z_score` with geometric or harmonic mean is the case that exists today. This check is not load-bearing yet (the scope check admits only `arithmetic_mean`, which every normalization accepts) but it starts biting the moment that scope widens — which is the next PR.

### Verification

Two levels, both verified to actually fail when broken:

- **`CoordinatorScoreFusionDifferentialTests`** now runs every case across all three techniques, and asserts **bit equality** instead of a `1e-6` delta. On a single shard both paths run the same arithmetic over the same values in the same order, so a tolerance could only hide a real divergence. Two new hit sets cover a wide score range and an all-zero leg (`l2`'s zero-norm guard, `min_max`'s `min == max` branch). Pinning the coordinator normalizer to `min_max` fails all 6 cases.
- **`HybridQueryFusedModeIT.testFusedMode_forEveryNormalizationTechnique_thenMatchesClassicPipeline`** is the end-to-end counterpart: for each technique, the same two legs fused on the coordinator vs. the equivalent classic `normalization-processor` pipeline — same documents, same order, same scores. Single-shard index (see below), scores from a numeric field rather than BM25 so nothing depends on per-shard term statistics, and the second leg is filtered to the top ranks so most documents match one leg only, exercising the "leg did not match" path. Pinning the fused technique to `min_max` fails it: `expected:<1.2270122> but was:<1.0>`.

Locally: full unit suite 2708 green; `integTest` for `*HybridQueryFusedMode*`, `*RRFProcessorIT*`, `*HybridQueryIT*` 43 green.

### Two findings worth flagging

**Single-shard parity is bit-exact; cross-shard `l2` is not.** `l2` accumulates its sum of squares in `float` (deliberately — widening to `double` would silently change every existing classic `l2` score) and float addition is not associative. Classic accumulates per shard, the coordinator over the already-merged leg, so on one shard the norms are bit-identical and across shards they can differ in the last bit. That is a difference in accumulation order, not in the formula, and it is why the parity IT uses a single shard. Documented on `L2ScalarNormalizer`.

**The `0.0` sentinel is fine for geometric/harmonic mean — no presence mask needed.** `ScalarNormalizers`' javadoc warned that wiring those combiners would likely need an explicit presence mask rather than `CoordinatorScoreFusion`'s `0.0` "leg did not match" sentinel, because `l2` can legitimately normalize a real score to `0.0`. That turns out not to be a fused-vs-classic problem: `ScoreCombiner#getNormalizedScoresPerDocument` allocates `new float[legCount]` and fills only the legs that matched, so **classic feeds the same `0.0`**, and geometric/harmonic skip `<= 0` identically on both paths. The residual ambiguity — a matched doc whose `l2` score really is `0.0`, reachable only when a leg's whole norm is 0 — exists identically in classic, so it is not a divergence. The note is replaced with that finding; the sentinel stays. This is what makes the next PR (geometric/harmonic in fused mode) a gate change rather than a fusion-core change.

### Coordination note for #1943

#1943's `CoordinatorScoreFusion` predates #1948 and has no `fuseRrf`; its gate allowed only `min_max` + `arithmetic_mean`. This PR keeps that behavior (RRF is still refused in fused mode, now with a clearer message) rather than restoring RRF routing, since that belongs to whichever of #1943/#1948 merges second. Flagging so it is not discovered at merge time.

### Related Issues
Part of #1941.

### Check List
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Enhancements
 * In-query fusion in hybrid search. Implement base classes and enable fusion (min_max and arithmetic mean) ([#1933](https://github.com/opensearch-project/neural-search/pull/1933))
-* In-query fusion in hybrid search. Support nested hybrid queries, search across multiple indices, aggregations, collapse with group expansion, and point in time; refuse fused mode while any node in the cluster is below 3.8.0; cap the leg sub-searches one request may fan out with the `plugins.neural_search.hybrid.fusion.max_leg_searches` cluster setting ([#1943](https://github.com/opensearch-project/neural-search/pull/1943))
+* In-query fusion in hybrid search. Support nested hybrid queries, search across multiple indices, aggregations, collapse with group expansion, and point in time; refuse fused mode while any node in the cluster is below 3.8.0; refuse `scroll` in fused mode with a validation error (use `point_in_time` instead); cap the leg sub-searches one request may fan out with the `plugins.neural_search.hybrid.fusion.max_leg_searches` cluster setting ([#1943](https://github.com/opensearch-project/neural-search/pull/1943))
 
 ### Bug Fixes
 * [SemanticHighlighter] Fix SemanticHighlighterExtBuilder.toXContent ([#1906](https://github.com/opensearch-project/neural-search/issues/1906)) (query-insights [#651](https://github.com/opensearch-project/query-insights/issues/651))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Enhancements
 * In-query fusion in hybrid search. Implement base classes and enable fusion (min_max and arithmetic mean) ([#1933](https://github.com/opensearch-project/neural-search/pull/1933))
 * In-query fusion in hybrid search. Support nested hybrid queries, search across multiple indices, aggregations, collapse with group expansion, and point in time; refuse fused mode while any node in the cluster is below 3.8.0; refuse `scroll` in fused mode with a validation error (use `point_in_time` instead); cap the leg sub-searches one request may fan out with the `plugins.neural_search.hybrid.fusion.max_leg_searches` cluster setting ([#1943](https://github.com/opensearch-project/neural-search/pull/1943))
+* In-query fusion in hybrid search. Support `z_score` and `l2` normalization in fused mode, so the whole score-normalization family is available with `arithmetic_mean`; both run the same shared normalization cores as the classic shard-side path ([#1962](https://github.com/opensearch-project/neural-search/pull/1962))
 
 ### Bug Fixes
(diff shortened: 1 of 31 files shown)


martin-gaievski:
@dbwiddis, your branch needs a rebase, base is `74f0a903` (#1944) but `feature/fusion-hybrid-query` tip is now `965379e9` with #1943 squash-merged. Once rebased the diff should collapse to just `e83a6cb4` as expected. Rest LGTM architecturally.

dbwiddis:
@martin-gaievski rebased. Note this PR has 3 commits, the first two represent currently open PRs that I'll need to rebase onto once they merge:
 - #1960 
 - #1961

martin-gaievski (approved):
Looks good, thank @dbwiddis

Issue #1558, opened 2025-09-10
[FEATURE]Avoid multiple inference call for same query text in neural query


heemin32:
### Is your feature request related to a problem?
In the following query, I’m passing the same query text, house, to the same model to search across two different fields in the same index. Internally, this results in two separate ml-common predict calls to generate embeddings for the same text. We should consider avoiding multiple predict calls when the query text and model ID are identical. Also, we would like to reduce the predict call happening in each shard of the index.
```
{
  "query": {
    "bool": {
      "should": [
        {
          "neural": {
            "title_embedding": {
              "model_id": "asdfasdf",
              "query_text": "house",
              "k": 100
            }
          }
        },
        {
          "neural": {
            "body_embedding": {
              "model_id": "asdfasdf",
              "query_text": "house",
              "k": 100
            }
          }
        }
      ]
    }
  }
}
```

### What solution would you like?
We aim to reduce redundant embedding generation for identical query text with the same model ID by leveraging caching. In addition, we’d like to explore whether generating embeddings at the coordinator level provides any advantages compared to the shard level even with cache is implemented.

There are several possible implementation approaches:

####Cache storage options####
1. In-memory cache on each data node (node-level cache)
2. System index for cluster-wide cache
3. External storage outside the cluster, such as Redis

####Embedding generation location####
1. Coordinator level
2. Shard level

### What alternatives have you considered?
Implementing the cache in ml-common plugin

### Do you have any additional context?
N/A

linjiaye929:
Hi maintainers, I'd like to work on this. I propose starting with request/coordinator-local deduplication keyed by model ID, input, and type, without a node cache, system index, or Redis. I'll add unit/integration and BWC coverage. Is this narrow first scope acceptable? @linjiaye929

bzhangam:
Thanks for picking this up, @linjiaye929. Sharing how I'm thinking about the problem — I'd split it into ingest and search, because they have different right answers.

**Ingest**

For ingest we already collect a list of texts before inferencing (batch ingest), so we have the full set in hand at one point. Deduplication here is straightforward and stateless: unique the inference requests by `(model_id, input, type)`, predict once per unique entry, and scatter the results back. Low risk, no shared state, no invalidation concerns. This is a clean first step.

**Search**

[...]

PR #1967, opened 2026-08-25
[Fusion] Report matched_queries for named sub-queries in fused mode

martin-gaievski:
### Description
In fused mode (`hybrid` with the `fusion` parameter), `matched_queries` silently disappears from the response
for a sub-query that carries `_name`. Today hybrid reports it in every configuration. This PR closes both
halves of that gap.

`matched_queries` is a fetch-phase field built from the names registered while the query is converted on the
shard (`QueryShardContext#addNamedQuery` → `ParsedQuery#namedFilters()`); `MatchedQueriesPhase` then builds its **own** `Weight` per registered name and evaluates it per hit. So a named sub-query has to be **registered**,
never executed, for its name to be reported.

Fused mode converted legs only via the non-scoring Tail, and the coordinator passed an empty leg list whenever
the Tail was not built (`needsTail` false). The result is an empty `namedFilters()`, so
`MatchedQueriesPhase.getProcessor` returns `null` and the field is absent — at HTTP 200, with no warning. The
boundary is a request-shape detail rather than anything the user said about names: `track_total_hits` one above
the fused window reports names, exactly at the window does not. Same for the other three Tail triggers
(aggregations, highlight, non-`_score` sort, collapse expansion) — turning any of them on brings the field back.

A second, independent loss existed in **every** configuration, Tail or not: a materializable kNN/neural leg is
replaced by an ids-address substitute, and the substitute was a fresh builder carrying no name.

Examples

1. Create index and ingest documents
```json
PUT /products
{
  "settings": { "index": { "knn": true, "number_of_shards": 1, "number_of_replicas": 0 } },
  "mappings": { "properties": {
    "title": { "type": "text" },
    "brand": { "type": "keyword" },
    "vec":   { "type": "knn_vector", "dimension": 2,
               "method": { "name": "hnsw", "space_type": "l2", "engine": "lucene" } }
  }}
}

PUT /products/_doc/1?refresh=true
{ "title": "wireless noise cancelling headphones", "brand": "acme",   "vec": [1.0, 1.0] }

PUT /products/_doc/2?refresh=true
{ "title": "wireless mechanical keyboard",         "brand": "globex", "vec": [1.2, 1.0] }
```

2. Search query 1
```json
POST /products/_search?size=2
{
  "track_total_hits": false,
  "query": {
    "hybrid": {
      "fusion": {
        "window_size": 10,
        "combination":   { "technique": "arithmetic_mean" }
      },
      "queries": [
        { "match": { "title": { "query": "wireless", "_name": "text_leg" } } },
        { "term":  { "brand": { "value": "acme",     "_name": "brand_leg" } } }
      ]
    }
  }
}
```

**Before fix no `matched_queries` anywhere:**

```json
"hits": [
  { "_id": "1", "_score": 0.5005 },
  { "_id": "2", "_score": 0.5 }
]
```

**After fix, and byte-for-byte what classic hybrid returns for the same legs:**

```json
"hits": [
  { "_id": "1", "_score": 0.5005, "matched_queries": ["text_leg", "brand_leg"] },
  { "_id": "2", "_score": 0.5,    "matched_queries": ["text_leg"] }
]
```

Flipping `track_total_hits` to `true` — or adding an aggregation, a highlight, a non-`_score` sort, or collapse
expansion — brings the field back on the old code, which is what makes this hard to notice: nothing in the
request mentions named queries, and the ranking is unaffected either way.

### Related Issues

### Check List


bzhangam (approved):

src/main/java/org/opensearch/neuralsearch/query/HybridFusionQueryBuilder.java
```diff
@@ -176,13 +200,22 @@ protected QueryBuilder doRewrite(QueryRewriteContext queryRewriteContext) throws
             rewrittenInnerHits.add(r);
             changed |= r != q;
         }
+        // The name-only legs must be rewritten too: a name is registered against the query its builder compiles to, and an
+        // un-rewritten builder either compiles to something else or refuses to compile at all.
+        List<QueryBuilder> rewrittenNamedOnly = new ArrayList<>(namedOnlyQueries.size());
+        for (QueryBuilder q : namedOnlyQueries) {
+            QueryBuilder r = q.rewrite(matchSetContext);
```
bzhangam, line 207: For neural sparse we need to trigger inference call again during the rewrite. Same thing can happen for the tail rewrite. Seems like there is no elegant way to avoid this unless we rewrite the leg in the coordinator and dispatch the rewritten leg query?

Or prioritize https://github.com/opensearch-project/neural-search/issues/1558 can help on duplicated inference call.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -351,11 +370,56 @@ private static List<QueryBuilder> legQueriesForTail(List<QueryBuilder> legs, Sea
...
+            }
+            // A materialized leg answers to its own _name: matched_queries is reported from the names registered while
+            // this query is converted, and the substitute is a fresh builder that would otherwise carry none — so a named
+            // kNN/neural leg would silently lose the field even with the Tail present. What it then reports is the
+            // documents the leg returned, which is the same bound materialization already accepts for the match set.
+            // Under include_named_queries_score the reported value is the substitute's score rather than the ANN
+            // similarity: the shard never sees the vector query, and re-running it for a reporting field is exactly the
+            // graph walk materialization exists to avoid. For the same reason only the leg's own name is inherited — a
+            // _name nested inside the leg (on a knn filter, say) has no clause left here to be registered against.
+            tail.add(materializedLeg(legHits[legIndex]).queryName(leg.queryName()));
```
vibrantvarun: For a materialized kNN/neural leg, the substitute registered under the leg's `_name` here is a filter-only `bool{ids, _index}`. Two behavioral gaps vs classic hybrid seem worth pinning:

- Under `include_named_queries_score`, a filter-only bool contributes 0 to score, so the named-query score is reported as `0.0` rather than the ANN similarity classic reports.
- The substitute only addresses the ids the leg *returned* (≤ `window_size`), so a doc matched by this leg but outside its own window won't get the name; classic re-evaluates the weight and would.

Both are inherent to not re-running the vector query and are documented in the code comment, but neither is covered by an IT against the classic oracle (the scoring IT uses `function_score`, i.e. non-materializable, legs). Could we add an IT pinning the materialized-leg score (expected `0.0`) and the recall bound, so the asymmetry is anchored to behavior rather than only a comment?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
 * In-query fusion in hybrid search. Support nested hybrid queries, search across multiple indices, aggregations, collapse with group expansion, and point in time; refuse fused mode while any node in the cluster is below 3.8.0; refuse `scroll` in fused mode with a validation error (use `point_in_time` instead); cap the leg sub-searches one request may fan out with the `plugins.neural_search.hybrid.fusion.max_leg_searches` cluster setting ([#1943](https://github.com/opensearch-project/neural-search/pull/1943))
 * In-query fusion in hybrid search. Support `z_score` and `l2` normalization in fused mode, so the whole score-normalization family is available with `arithmetic_mean`; both run the same shared normalization cores as the classic shard-side path ([#1962](https://github.com/opensearch-project/neural-search/pull/1962))
+* In-query fusion in hybrid search. Report `matched_queries` for a sub-query carrying `_name`, matching classic hybrid: named legs are registered on the shard even when the non-scoring Tail is not built, and a materialized kNN/neural leg keeps its `_name`
```
vibrantvarun: Two small things:
1. This reads "matching classic hybrid", but for materialized kNN/neural legs the `include_named_queries_score` value and recall differ (see the orchestrator comment) — a short caveat here would keep it from being read as full parity.
2. The sibling fusion entries (#1933/#1943/#1962) each end with a `([#1967](...))`-style PR link; this line is missing it.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -109,9 +116,21 @@ static QueryBuilder buildFusedQuery(
         if (ranked.ids().length == 0) {
             return new MatchNoneQueryBuilder();
         }
-        List<QueryBuilder> tail = needsTail(source, ranked.ids().length) ? legQueriesForTail(legs, legHits) : List.of();
+        boolean tailNeeded = needsTail(source, ranked.ids().length);
+        // A leg's _name only reaches matched_queries if its builder is converted on the shard, and the Tail is the only
+        // thing that converts legs. When the Tail is not built, carry the same leg forms for registration alone: the fetch
+        // phase re-evaluates every named query from its own weights, so nothing has to execute for one to be reported.
+        boolean namesOnly = tailNeeded == false && anyLegNamed(legs);
+        List<QueryBuilder> legsInTailForm = tailNeeded || namesOnly ? legQueriesForTail(legs, legHits) : List.of();
```
vibrantvarun: In the names-only path `legsInTailForm` carries *all* legs, and `registerNamedOnlyQueries` then calls `toQuery()` on each — including unnamed, non-materializable legs that a Top-only query would otherwise never compile on the shard. For a request with one named leg + one unnamed heavy leg (e.g. `script_score`), that's an extra shard-side compile per unnamed leg for no reporting benefit. Bounded by `max_leg_searches` so minor — but would filtering the name-only carry to legs that actually render a `_name` be worth it, or is avoiding a re-decision of named-ness at conversion time the deliberate tradeoff?

src/main/java/org/opensearch/neuralsearch/query/HybridFusionQueryBuilder.java
```diff
@@ -144,6 +164,9 @@ public HybridFusionQueryBuilder(StreamInput in) throws IOException {
         this.scores = in.readFloatArray();
         this.tailQueries = in.readNamedWriteableList(QueryBuilder.class);
         this.innerHitsQueries = in.readNamedWriteableList(QueryBuilder.class);
+        // No wire-version gate: this query is built only for a cluster whose every node supports fused mode (see
+        // HybridQueryBuilder#requireClusterSupportsFusedMode), and it has never shipped in a released version.
+        this.namedOnlyQueries = in.readNamedWriteableList(QueryBuilder.class);
```
vibrantvarun, line 169: `namedOnlyQueries` is (de)serialized unconditionally with no `TransportVersion` gate. This looks fine for the released contract (type is new-in-3.8.0, version-gated via `requireClusterSupportsFusedMode`, never shipped). The one residual case is a mixed cluster of two *different* 3.8.0 pre-release snapshots — both pass the min-version gate but disagree on wire format, so the old reader stops after `innerHitsQueries` and the trailing list corrupts the stream. Worth a one-line note in the PR description so the unconditional read is a recorded decision.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -351,11 +370,56 @@ private static List<QueryBuilder> legQueriesForTail(List<QueryBuilder> legs, Sea
...
+        }
+        return false;
+    }
+
+    private static boolean rendersQueryName(QueryBuilder leg) {
+        // toXContent opens and closes the query's own object, so the builder is used at its root position: wrapping it in
+        // another object makes every render fail, and this method's fail-open would then carry every leg unconditionally.
+        try (XContentBuilder builder = XContentFactory.jsonBuilder()) {
+            leg.toXContent(builder, ToXContent.EMPTY_PARAMS);
+            return builder.toString().contains(QUERY_NAME_KEY);
```
vibrantvarun, line 431: nit: `builder.toString()` works here (the json builder is byte-backed), but `Strings.toString(builder)` is the idiomatic way to render an `XContentBuilder`. Also — already acknowledged in the javadoc, just flagging — the `"_name":` substring check false-positives on a leg that queries a field literally named `_name` (e.g. `{"term":{"_name":...}}`); harmless (one no-op registration), since the shallow `queryName()` check covers a leg's own name.

vibrantvarun (commented):
Read the fused-mode `matched_queries` fix end-to-end (Top-only registration + materialized-leg name inheritance). The design mirrors the existing `innerHitsQueries` register-don't-execute pattern cleanly, and the oracle-anchored ITs are strong. A few things worth considering inline — mostly the materialized-leg deviation being under-tested/undocumented, plus a couple of nits. Nothing that looks like a correctness defect in the released contract.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionQueryBuilder.java
martin-gaievski: that's mostly pre-existing rather than new here. Tail is rewritten in this same method and is on by default, so a `neural_sparse` leg with a `model_id` already pays 2 inferences in fused mode vs classic's 1. Measured: classic 1, fused+Tail 2, Top-only+named 2. It's one extra coordinator call per model leg, never per shard.

This PR adds that cost in one shape only: Top-only *and* a named leg. `knn`/`neural`/`neural_knn` are unaffected, they arrive materialized to ids.

Pre-rewriting leg on the coordinator doesn't work cleanly: it registers an async action mid-drain, and a nested fused-hybrid leg can't share one rewritten form between the round-1 fan-out and Tail. #1558 is the right approqch, same coordinator node, one request, and it also fixes the pre-existing Tail case. I'll file the follow-up.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: ack, will add tests

CHANGELOG.md
martin-gaievski: ack, will correct this entry

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: It's not a deliberate tradeoff. you and @bzhangam landed on the same line from opposite ends. You're counting the wasted shard-side `toQuery` compile per unnamed leg; his concern was that a carried model-resolving leg re-runs its rewrite and pays a second ML inference on the coordinator. One filter fixes both, so it's going in this PR.

`anyLegNamed` already computes the per-leg predicate, so this is roughly ten lines: filter `legsInTailForm` on `queryName() != null || rendersQueryName(leg)` in the names-only branch. It has to be that predicate rather than the shallow `queryName()` check, or a `_name` nested under `nested`/`function_score`/a `knn` filter gets dropped. Semantics-preserving: `AbstractQueryBuilder.toQuery` registers only when a name is set, and `namedOnlyQueries` never reaches `buildSelfErasedQuery()`. It also removes the mutually-exclusive dual ternary the bot flagged, since the two branches stop sharing one list.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionQueryBuilder.java
martin-gaievski: makes sense, will do the change

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: there's no `Strings.toString(XContentBuilder)` overload in core OpenSearch, form that would compile, `Strings.toString(XContentType.JSON, leg)`, but imo it's worse here: it catches `IOException` internally and returns an error-JSON string, which contains no `"_name":`. I incline for keeping existing code as is

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
vibrantvarun: Confirmed on `dfee4926` — the two new ITs (`testFusedKnnTail_whenIncludeNamedQueriesScore_thenMaterializedLegReportsZeroAndRealLegReportsItsScore` and `testFusedKnnTail_whenAnnLegMatchedBeyondItsWindow_thenTheTruncatedDocLosesTheAnnName`) pin both the score-`0.0` and the recall bound against the real behavior. Thanks for adding them.

CHANGELOG.md
vibrantvarun: The updated entry reads well — dropping "matching classic hybrid" for the caveat about the returned-docs bound and the substitute score under `include_named_queries_score`, plus the `#1967` link, covers both points. Thanks.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
vibrantvarun: The `namedLegsForRegistration` + per-leg `carriesQueryName` refactor is exactly it: unnamed legs are no longer carried or compiled on the shard, and as a bonus it removes the extra `neural_sparse` coordinator inference in the unnamed Top-only case @bzhangam raised. Nice touch pinning the later-named-leg hit alignment (`testBuildFusedQuery_whenOnlyALaterLegIsNamed_thenTheSubstituteAddressesThatLegsOwnHits`) — that's the exact off-by-one the refactor could have introduced.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionQueryBuilder.java
vibrantvarun: The min-supported-version round-trip test (`testSerializationRoundTrip_whenStreamPinnedToMinimumSupportedVersion_thenAllThreeListsSurvive`) is a good way to record the wire contract. The one case it doesn't cover is old-reader/new-writer across two *different* pre-release 3.8.0 snapshots — but that's unreleased and low-risk, so fine to leave (or a one-line note in the description if you want it on record). No blocker either way.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
vibrantvarun: Agreed — withdrawing the suggestion. I checked `org.opensearch.core.common.Strings`: there's no `toString(XContentBuilder)` overload, and the `toString(MediaType, ToXContent)` form catches `IOException` and returns error-JSON with no `"_name":`, which would turn the current fail-open into fail-closed and silently drop names. Keeping the existing code is the correct call. Thanks for folding the `_name` false-positive note into the `carriesQueryName` javadoc.

vibrantvarun (approved):
Thanks for the quick turnaround — all five points are addressed. The materialized-leg score/recall deviation is now pinned by ITs, the CHANGELOG caveat + link are in, the name-only carry is filtered to named legs (which also removes the extra neural_sparse coordinator inference in the unnamed Top-only case), and the serialization round-trip is version-pinned. Confirmed my Strings.toString nit was wrong — that form would flip the fail-open detection to fail-closed. LGTM.

PR #1969, opened 2026-08-26
[Fusion] Block fused hybrid query outside main search query

martin-gaievski:
### Description
Blocking resolver query in it's set in one of these:
 - post_filter
 - aggregation filter
 - sort
 - internal rescore query
 - highlight query
 - wrapper
 
Reason is to limit resource consumption by cascading nested fan outs of sub-queries. This will create endless nested loop of fusion queries that will eventually lead to stack overflow exception and node crush.

Example of the query that we' are blocking
```json
  POST /my-index/_search
  {
    "query": { "match_all": {} },
    "post_filter": {
      "hybrid": {
        "fusion": {
          "window_size": 5,
        },
        "queries": [ { "match": { "text": "hello" } }, { "term": { "grp": "A" } } ]
      }
    }
  }
```

### Related Issues

### Check List


src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -665,6 +668,88 @@ private static void requireClusterSupportsFusedMode() {
...
+        }
+
+        @Override
+        public void accept(final QueryBuilder queryBuilder) {
+            found = found || queryBuilder == target;
+        }
+
+        @Override
+        public QueryBuilderVisitor getChildVisitor(final Occur occur) {
+            return this;
```
bzhangam, line 796: Should we early stop the visit when we found it?

bzhangam (approved):

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
vibrantvarun: The visitor API has no early-termination signal — `QueryBuilderVisitor` always descends via `getChildVisitor`, so a true short-circuit would mean throwing a control-flow exception to break out of `visit`. For the small query trees this runs on (a request body the user had to spell out, bounded by `max_leg_searches`), that trades readability for a walk that's already cheap, so I'd lean toward leaving it as a full walk. Happy to add a one-line note on the intentional non-short-circuit if useful.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
 * In-query fusion in hybrid search. Support nested hybrid queries, search across multiple indices, aggregations, collapse with group expansion, and point in time; refuse fused mode while any node in the cluster is below 3.8.0; refuse `scroll` in fused mode with a validation error (use `point_in_time` instead); cap the leg sub-searches one request may fan out with the `plugins.neural_search.hybrid.fusion.max_leg_searches` cluster setting ([#1943](https://github.com/opensearch-project/neural-search/pull/1943))
 * In-query fusion in hybrid search. Support `z_score` and `l2` normalization in fused mode, so the whole score-normalization family is available with `arithmetic_mean`; both run the same shared normalization cores as the classic shard-side path ([#1962](https://github.com/opensearch-project/neural-search/pull/1962))
+* In-query fusion in hybrid search. Fix ceiling's counting scope (leg budget counted a fan-out of zero) when Core rewrites `post_filter`, aggregation filters, `rescore` and `highlight_query` ([#1969](https://github.com/opensearch-project/neural-search/pull/1969))
```
vibrantvarun: Two things on this entry:
1. **Incomplete list** — the guard also refuses a fused hybrid in **`sort`** and inside a **`wrapper`** query (per `requireFusedQueryIsPartOfRequestQuery` and its error message), but only `post_filter`, aggregation filters, `rescore` and `highlight_query` are listed here.
2. **Emphasis** — the user-observable change is a new **validation error** for a fused hybrid outside the request `query`, whereas this reads as an internal counting fix. Might be clearer in the sibling #1943 style, e.g.: *"refuse a fused hybrid outside the request `query` (in `post_filter`, an aggregation filter, `sort`, `rescore`, `highlight`, or a `wrapper`) with a validation error — use a `bool` with the same clauses as `should` instead."*

vibrantvarun (commented):
Reviewed the position guard end-to-end. The mechanism is sound: `requireFusedQueryIsPartOfRequestQuery` runs first in `doRewriteFused` round 1 (before the budget count and config resolution), and reachability-by-identity from `source().query()` correctly keeps the legitimate shapes (top-level, nested in `bool`, and a fused leg of an enclosing fused hybrid) while refusing the unbounded ones. The identity (`==`) match depends on `getSearchRequest().source().query()` still referencing this builder at round-1 time — which holds because the guard is skipped once `fusedSupplier` is set, and the `…isAdmitted`/`…isServed` tests exercise the real `source.rewrite(...)` path, so it's regression-guarded. Test coverage is strong on both the served and abuse paths (inline-config-in-post_filter refused instead of recursing, wrapper refused, classic-hybrid-outside-query unaffected, over-budget still hits the ceiling). One CHANGELOG accuracy nit inline; otherwise this looks good to me.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
martin-gaievski: Checked the API before answering, and there is no cheap version of this.

`QueryBuilderVisitor` exposes only `accept` and `getChildVisitor` — no early-exit signal — and the two things that look like they would work do not:

- **Handing back `NO_OP_VISITOR` from `getChildVisitor` once found** stops the comparisons, not the descent. Every `visit` implementation recurses unconditionally, whatever the visitor returns — `BoolQueryBuilder`, `DisMaxQueryBuilder`, `ConstantScoreQueryBuilder`, `NestedQueryBuilder` and `FunctionScoreQueryBuilder` all call `child.visit(subVisitor)` — so the tree still gets walked end to end, just with a no-op `accept`.
- **Skipping the comparison** is already what happens: `found = found || queryBuilder == target` does not evaluate the right operand once `found` is true. (Which is also why the automated suggestion on this same hunk is a literal no-op — it rewrites the `||` into an `if` and calls it a short-circuit.)

So a real short-circuit means throwing out of `accept` for control flow. Against that: the walk runs **once per fused hybrid, at rewrite round 1 only** — round 2 returns at the `fusedSupplier` check above this guard, and the match-set path returns before that — over a body the user had to spell out, whose fused hybrids `max_leg_searches` caps at 25 legs by default. Not worth an exception for control flow, so I have left it a full walk and put the reasoning on `IdentityFinder` rather than leaving the next reader to ask the same question:

```java
 * <p>The walk is not cut short once the target is found, deliberately: {@link QueryBuilderVisitor} has no
 * early-termination signal, and every {@link QueryBuilder#visit} implementation recurses unconditionally, so even
 * handing back {@link QueryBuilderVisitor#NO_OP_VISITOR} would keep descending — only the comparison would stop, and
 * that already short-circuits. Stopping the traversal itself would mean throwing out of {@code accept} for control
 * flow, for a walk that runs once per fused hybrid at rewrite round 1 over a body the user had to spell out and that
 * {@link #validateFusedLegSearchBudget} caps.
```

Worth pinning down though — chasing it is what turned up that `boosting` and `script_score` do not recurse at all, which was a real gap in this guard's message. Details on the CHANGELOG thread.

CHANGELOG.md
martin-gaievski: Both fair, and verifying the first one turned up a third gap in the same list.

**`sort` — confirmed, and now measured.** `SearchSourceBuilder.rewrite` rewrites the sort list alongside the query, and `FieldSortBuilder` / `GeoDistanceSortBuilder` / `NestedSortBuilder` each rewrite their nested filter, so a fused hybrid in a sort's `nested.filter` does reach this guard and is refused. The message claimed it; nothing measured it. Now it is a case in `HybridQueryFusedFanOutTests`' positions map (which drives core's own `source.rewrite(...)`) and in `HybridQueryFusedQueryPositionIT` — the IT index gains one unused nested property so a sort has a path to point at.

**`wrapper` — confirmed**, already pinned by `testFusedHybridInsideAWrapperQuery_isRefused`.

**The third gap: `boosting` and `script_score`.** Both call `visitor.getChildVisitor(...).accept(query)` — `accept`, not `visit` — so they expose their immediate inner query and nothing under it. Measured on this branch:

| body | `countFusedLegSearches` |
|---|---|
| `boosting{positive: <fused>}` | 2 |
| `boosting{positive: bool{should: <fused>}}` | **0** |
| `script_score{query: <fused>}` | 2 |
| `script_score{query: bool{should: <fused>}}` | **0** |
| `nested{query: bool{should: <fused>}}` | 2 (recurses, so it is the builder's blindness and not the walk's) |

So a fused hybrid more than one level under either is refused as well — which the message did not name, and whose only advice ("use a `bool` query with the same clauses as `should` instead") is useless for a body that already **is** a `bool`. The refusal is right, and for a stronger reason than "conservative": `countFusedLegSearches` is the same `visit` walk, so those two bodies count as **zero** leg sub-searches — they are precisely the uncounted fan-out this PR exists to stop. What was wrong was the wording. Fixed:

- the message now states the requirement as *part of the request's `query` **and reachable within it*** and splits the remedy by cause — wrong position → use a `bool` there; hidden by `wrapper` / `boosting` / `script_score` → move the `hybrid` out of it;
- dropped the javadoc's *"No shape that fusion can answer is lost"*, which was simply false, and replaced it with the three ways a hybrid becomes invisible plus the shared-traversal argument for why refusing them is consistent;
- `testFusedHybridUnderAQueryThatHidesItsChildren_isRefusedAndWasNeverCounted` pins the table above, the refusal, and `nested` as the contrast.

**New CHANGELOG entry**, in the #1943 shape you suggested, naming every position and both invisibility causes:

> In-query fusion in hybrid search. Refuse a fused hybrid query that is not part of the request's own `query` with a validation error, so the `plugins.neural_search.hybrid.fusion.max_leg_searches` ceiling can no longer be evaded: core rewrites `post_filter`, aggregation filters, sorts, `rescore` and `highlight_query` alongside the `query` but the ceiling counts the `query` alone, so a fused hybrid in any of those positions used to fan out at a counted budget of zero (and one in `post_filter`, which every leg sub-search inherits, recursed without a fixed point) — use a `bool` query with the same clauses as `should` in those positions instead; a fused hybrid the query-tree walk cannot reach is refused for the same reason, whether it is inside a `wrapper` query or more than one level under a `boosting` or `script_score` query ([#1969](https://github.com/opensearch-project/neural-search/pull/1969))

martin-gaievski:
Review round addressed in f4fb9ad9. Thanks both — the two comments were narrow but chasing the second one surfaced a real gap.

**@bzhangam, early-stopping the reachability walk** — no cheap version exists: `QueryBuilderVisitor` has no early-termination signal, every `visit` implementation recurses unconditionally (so `NO_OP_VISITOR` would stop the comparisons and keep descending), and the comparison already short-circuits through the `||`. A genuine stop means throwing out of `accept` for control flow, for a walk that runs once per fused hybrid at rewrite round 1 over a body `max_leg_searches` caps at 25 legs. Left as a full walk with the reasoning recorded on `IdentityFinder`. [Detail](https://github.com/opensearch-project/neural-search/pull/1969#discussion_r3882732603)

**@vibrantvarun, the CHANGELOG entry** — rewritten in #1943's shape, naming every position and both reasons a hybrid can be unreachable. Verifying your "incomplete list" point turned up that the *error message and the javadoc* had the same gap, plus one neither of us had named:

- **`sort`** is genuinely refused (`SearchSourceBuilder.rewrite` rewrites the sort list; `FieldSortBuilder` / `GeoDistanceSortBuilder` / `NestedSortBuilder` each rewrite their nested filter) but was claimed by the message and measured by nothing. Now a case in the unit positions map and in the IT.
- **`boosting` and `script_score`** hand their inner query to `accept` without recursing into it, so a fused hybrid more than one level under either is invisible to the walk. Measured: `boosting{positive: bool{should: <fused>}}` and `script_score{query: bool{should: <fused>}}` both count **0** leg sub-searches, against 2 one level up and 2 through a `nested` query. Those bodies were being refused under a message that did not name them and whose only advice — use a `bool` — is useless for a body that already is one.

The refusal stands, and for a better reason than caution: `countFusedLegSearches` is the *same* `visit` walk, so anything invisible to the guard is counted as zero legs by the ceiling — which is exactly the uncounted fan-out this PR exists to stop. The wording was what was wrong. The message now states the requirement as *part of the request's `query` and reachable within it* and splits the remedy by cause; the javadoc's "No shape that fusion can answer is lost" — false — is replaced by the three ways a hybrid becomes invisible and the shared-traversal argument. [Detail](https://github.com/opensearch-project/neural-search/pull/1969#discussion_r3882733931)

**Verification on f4fb9ad9:** unit suite 2751 tests / 0 failures, `spotlessJavaCheck` clean, `HybridQueryFusedQueryPositionIT` 6/6 green including the new `sort` position. One new unit test (`testFusedHybridUnderAQueryThatHidesItsChildren_isRefusedAndWasNeverCounted`) and one new position case; no production behaviour changed beyond the refusal message text.

vibrantvarun (approved):
LGTM. Both review threads addressed in f4fb9ad9, verified against the diff and OpenSearch core: the CHANGELOG/message/javadoc now name every refused position (including `sort`) and both invisibility causes, and chasing the message gap correctly surfaced that `boosting`/`script_score` expose only their immediate child (`accept`, not `visit`) — so a fused hybrid more than one level under either is counted as zero legs by `countFusedLegSearches` and rightly refused, closing a real `max_leg_searches` evasion rather than just a wording issue. Guard runs once per fused hybrid at rewrite round 1; identity reachability is regression-guarded by the admitted/served tests through the real `source.rewrite(...)` path; new `testFusedHybridUnderAQueryThatHidesItsChildren_isRefusedAndWasNeverCounted` plus the `sort` position case cover the additions. Thanks for the thorough follow-up.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
martin-gaievski: Correcting myself here, because the census in my reply above is wrong.
 
"Every visit implementation recurses unconditionally, whatever the visitor returns" appears incorrect. Of core's 47 AbstractQueryBuilder subclasses, 15 override visit, and only the 5 I listed recurse. Ten hand their inner query to getChildVisitor(...).accept(child) and stop there … and the other 32 expose nothing at all.

Proper answer is: no implementation consults a stop signal, so NO_OP_VISITOR would still stop only the comparisons. What did not survive is the walk itself … Both walks now call visit themselves, keyed by an identity set so nothing is entered twice.

CHANGELOG.md
martin-gaievski: Correction to the middle of my reply above: table is right, but conclusion I drew from it was backwards.
  
I read boosting{positive: bool{should: <fused>}} counting 0 as proof the refusal was correct … That zero was the bug, not the safeguard. [then the per-nesting-level explanation and the 10-vs-30 / 20-vs-84 / 20-vs-340 table]. 
Double checking this turned wording problem into a measured ceiling bypass, so the code changed rather than just changelog entry. Force-pushed as b9cdbf9f, replacing f4fb9ad9

vibrantvarun (approved):
Re-approving on b9cdbf9fea — my earlier approval was on the rebased-out f4fb9ad9 and shouldn't carry, since that revision still had the bypass this commit closes.

The fix is the right one: both walks now descend themselves (`accept` → `visit`) instead of trusting builders to recurse, identity-deduped so each node is counted once and the accept↔visit re-entrancy terminates. That turns `max_leg_searches` from a per-nesting-level cap back into a per-request one — closing the geometric fan-out under `boosting`/`script_score` (340 vs a ceiling of 25) — while correctly admitting the deeply-nested shapes that were over-refused, and keeping `wrapper`/`template` fail-closed since they expose no child to count. Nice catch surfacing a real ceiling evasion out of a changelog nit. LGTM.

PR #1971, opened 2026-08-27
Fix inaccurate hits.total.value on hybrid queries with small size

idsoke:
### Description
`HybridTopScoreDocCollector` never called `hitsThresholdChecker.incrementHitCount()`, so `isThresholdReached()` was always false. Once a sub-query's top-k heap filled to `size` (page size), its eviction score was fed into `minCompetitiveScore`, letting the bulk scorer prune documents shard-wide before `track_total_hits`' threshold was reached, undercounting `hits.total.value`.

Fix: increment the hit counter on every collected doc, and only feed eviction scores into `minCompetitiveScore` once the threshold is reached.

Added a regression test with 5000 matches and a small `size`; confirmed it fails without the fix (`expected:<5000> but was:<4096>`, matching the reported bug) and passes with it. Full `search`/`query`/`collector` test suites pass with no regressions.

### Related Issues
Related to opensearch-project/OpenSearch#22823

### Check List


vibrantvarun (commented):
Traced the fix through the collector, HitsThresholdChecker, HybridCollectorManager and the two sibling collectors — it's correct. Root cause confirmed (incrementHitCount was never called, so isThresholdReached() was always false and eviction scores fed minCompetitiveScore as soon as the top-k heap filled at page size, pruning before track_total_hits' threshold). The fix aligns HybridTopScoreDocCollector with HybridTopFieldDocSortCollector / HybridCollapsingTopDocsCollector (both already increment), so it's not a double-count; top-k results are unchanged (the heap still evicts on overflow — only the feedback into pruning is now gated); and threshold = max(numDocs, trackTotalHitsUpTo) behaves correctly across track_total_hits settings. The shared non-atomic checker across concurrent slices is pre-existing (siblings share it) and benign — a lost increment only delays isThresholdReached, never causes premature pruning. One test-coverage note inline; approve-worthy once that path is covered.

Behavior note for the changelog/consumers: hits.total.relation now correctly reports `gte` for hybrid queries matching more than the threshold, where it previously always reported `eq` (with an undercounted value).

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopScoreDocCollector.java
```diff
@@ -161,7 +164,11 @@ public void collect(int doc) throws IOException {
                 // this way we're inserting into heap and do nothing else unless we reach the capacity
                 // after that we pull out the lowest score element on each insert
                 ScoreDoc evictedScoreDoc = pq.insertWithOverflow(currentDoc);
-                if (Objects.nonNull(evictedScoreDoc)) {
+                // Only start feeding the eviction score back into minCompetitiveScore (used by the
+                // bulk scorer to skip non-competitive docs) once we've reached the total hits threshold.
+                // Otherwise the heap fills up after just `numHits` (page size) docs and pruning kicks in
+                // long before track_total_hits' accuracy threshold is satisfied, undercounting totalHits.
+                if (Objects.nonNull(evictedScoreDoc) && hitsThresholdChecker.isThresholdReached()) {
```
vibrantvarun, line 171: The fix newly *activates* two paths that only fire once the threshold is reached: this gated min-competitive-score update, and the `totalHitsRelation → GREATER_THAN_OR_EQUAL_TO` flip a few lines up (line 157, dead until now because `isThresholdReached()` was always false).

The regression test only exercises the **threshold-not-reached** half: `numMatchingDocs=5000` with `trackTotalHitsUpTo=10000` gives `threshold = max(numDocs, 10000) = 10000`, which 5000 never reaches — so it asserts `value=5000, relation=EQUAL_TO` (the exact path) but never enters this branch or flips the relation.

Could we add a test with matches **>** threshold (e.g. 15000 matches, `trackTotalHitsUpTo=10000`) asserting `value >= 10000` and `relation == GREATER_THAN_OR_EQUAL_TO`, and ideally that early termination still prunes past the threshold (so the perf optimization isn't lost)? That locks down the newly-live behavior rather than just the fixed undercount.

idsoke: Added a regression test for this (`testReduce_whenMatchedDocsExceedThreshold_thenTotalHitsIsAtLeastThresholdButRelationStaysEqualTo`, 15k matches / `trackTotalHitsUpTo=10000`) and it confirms `totalHits` never falls below the threshold once pruning resumes - so the undercount fix holds here too.

It also surfaced a gap, though: `hits.total.relation` still comes back `EQUAL_TO` in this case, not `GREATER_THAN_OR_EQUAL_TO`. My "behavior note" above was wrong. The `totalHitsRelation` flip inside `collect()` (line 157) does fire correctly once the threshold is reached, but `HybridSearchCollectorResultUtil#getTotalHits()` never reads that field - it derives the relation purely from whether `track_total_hits` is disabled (`trackTotalHitsUpTo == SearchContext.TRACK_TOTAL_HITS_DISABLED`), defaulting to `EQUAL_TO` for every other value, including the default 10000. So the flip is effectively dead for the relation the client sees. This looks pre-existing and also affects `HybridTopFieldDocSortCollector` / `HybridCollapsingTopDocsCollector`, which track the same field the same way - i.e. not introduced by this PR, and outside its scope (fixing it means touching the shared result-reduction path, not the collector/pruning logic this PR targets).

I've left the new test asserting the current (EQUAL_TO) behavior with a comment explaining the gap, rather than expanding this PR to fix it. Let me know if you'd rather it be folded in here, otherwise I can file a follow-up issue for the relation bug separately.

src/test/java/org/opensearch/neuralsearch/search/query/HybridCollectorManagerTests.java
```diff
@@ -266,6 +266,206 @@ public void testReduce_whenMatchedDocs_thenSuccessful() {
...
+                numMatchingDocs,
+                totalHitsValue
+            ),
+            totalHitsValue < numMatchingDocs
+        );
+        // Documents the gap described above: relation stays EQUAL_TO even though totalHitsValue is only a
+        // lower bound at this point. If HybridSearchCollectorResultUtil#getTotalHits() is fixed to honor the
+        // collector's own relation tracking, this assertion should be updated to expect
+        // GREATER_THAN_OR_EQUAL_TO instead.
+        assertEquals(TotalHits.Relation.EQUAL_TO, topDocsAndMaxScore.topDocs.totalHits.relation());
```
martin-gaievski: 🟡 **Above the threshold hits.total is a size-dependent lower bound still reported as EQUAL_TO, and the new test pins it**

The gate this PR adds makes hitsThresholdChecker.isThresholdReached() live for the first time in this collector, so once a shard passes max(numDocs, track_total_hits_up_to) unique hits, eviction scores start pruning again and totalHits stops being exact. Probe run on the PR head (15000 matches, track_total_hits=10000, single segment): pagination_depth 10 -> totalHits 12288, 100 -> 12288, 1000 -> 12508, 3000 -> 12948, relation EQUAL_TO in every case. The collector does flip totalHitsRelation to GREATER_THAN_OR_EQUAL_TO at line 157, but HybridSearchCollectorResultUtil.getTotalHits() (lines 201-212) derives the relation solely from trackTotalHitsUpTo == TRACK_TOTAL_HITS_DISABLED and never reads it. The second new test asserts this EQUAL_TO outcome as expected behaviour.

```java
assertEquals(TotalHits.Relation.EQUAL_TO, topDocsAndMaxScore.topDocs.totalHits.relation());
```

**Why it matters:** The reported symptom of #22823 ('total.value differs by size') survives above the threshold, now with an exactness claim attached -> a client comparing hybrid against a plain query (which returns value 10000 / relation gte) sees a different, request-shape-dependent number labelled exact -> pagination and result-count UIs built on hits.total mislead users; the pinned assertion means the next engineer has to change a passing test to fix it.

**Suggestion:** Complete the threshold semantics in this PR - it is small and lives on the path the PR already changes: add `TotalHits.Relation getTotalHitsRelation()` to HybridSearchCollector (HybridTopFieldDocSortCollector and both collapse collectors already track the field), and in HybridSearchCollectorResultUtil.getTotalHits() use `trackTotalHitsUpTo == TRACK_TOTAL_HITS_DISABLED ? GREATER_THAN_OR_EQUAL_TO : hybridSearchCollector.getTotalHitsRelation()`. TopDocsMerger.getMergedTotalHits() already ORs relations, so the concurrent-search reduce needs no change. Then flip this assertion to GREATER_THAN_OR_EQUAL_TO. If you still prefer to defer, drop the EQUAL_TO assertion (keep the two value bounds) and link the filed follow-up issue in the test comment and PR body so the wrong relation is not locked in by a green test.

_[code-review-sage]_

src/main/java/org/opensearch/neuralsearch/search/collector/HybridTopScoreDocCollector.java
martin-gaievski: 🟡 **The same ungated eviction feedback remains in HybridCollapsingTopDocsCollector, so collapse requests keep undercounting**

Sibling inventory for 'eviction score fed into compoundQueryScorer.getMinScores() before the hits threshold': HybridTopScoreDocCollector - ungated before, gated after this PR; SimpleFieldCollector / PagingFieldCollector - no min-score feedback at all; HybridCollapsingTopDocsCollector - still ungated at lines 375-377 (`if (isSortByScore) { minScoreThresholds[subQuery] = ...; compoundQueryScorer.getMinScores()[subQuery] = Math.max(..., evictedScore); }`), and isSortByScore is true for the default collapse request because HybridCollectorFactory passes `new Sort(new SortField(null, SortField.Type.SCORE))` when no sort is given; HybridCollapsingTopGroupsCollector (main, #1956) deliberately never propagates min scores. The collapse collector also calls updateHitCount() before its sub-query loop, so it has exactly the pre-fix shape: heap fills at `size`, feedback prunes the next 4096-doc window, hits counted before CollectionTerminatedException are undercounted.

```java
if (Objects.nonNull(evictedScoreDoc) && hitsThresholdChecker.isThresholdReached()) {
```

**Why it matters:** Ungated feedback in the collapse collector -> HybridBulkScorer applies `score > minScore` and setMinCompetitiveScore to later windows -> a `collapse` hybrid query with small `size` and more matches than one window still reports a too-small hits.total.value, i.e. #22823 stays reproducible on the collapse path while the PR and CHANGELOG say the bug is fixed; the two collectors now disagree on the same user-visible field depending on whether `collapse` is present.

**Suggestion:** Either apply the same `&& hitsThresholdChecker.isThresholdReached()` guard to the collapse collector's minScore update (lines 375-377) with a test mirroring testReduce_whenManyMatchedDocsAndSmallSize_thenTotalHitsIsAccurate under collapse, or state explicitly in the PR body and CHANGELOG that the fix covers the non-collapse collector only and link a follow-up issue for HybridCollapsingTopDocsCollector. HybridCollapsingTopGroupsCollector's 'never propagate' comment is the reference for the conservative option.

_[code-review-sage]_

Issue #1802, opened 2026-03-20
[Design] Native Engine for Sparse ANN Search


chishui:
**Status:** current. This document reads as the design as it now stands; earlier revisions are listed under
[Revision history](#10-revision-history) with links to the comments carrying their rationale and benchmark data.

## 1. Overview

This document outlines the high-level design for creating an open-source C++ sparse vector index/query engine library and integrating it into the OpenSearch Neural Search plugin via JNI.

### Current State

* OpenSearch Neural Search plugin supports sparse vector search through Lucene's native capabilities
* SEISMIC algorithm has been integrated into OpenSearch for efficient sparse ANN search
* The k-NN plugin successfully demonstrates native library integration with Faiss through JNI

### Motivation

* Create a high-performance, unified, open-source C++ library that serves as a sparse vector index/query engine
* Enable pluggable architecture to integrate various sparse ANN algorithms (including SEISMIC)
* Achieve better performance through native C++ implementations for compute-intensive operations
* Keep index memory out of the JVM heap, and let the OS manage its residency (via mmap) instead of reimplementing caching and eviction in the plugin

### 3.1 Vision

We propose creating an **open-source C++ library** specifically designed as a sparse vector index and query engine. This library will:

* Serve as a standalone, reusable component for sparse vector operations
* Support pluggable algorithm backends, allowing integration of different sparse ANN algorithms
* Include SEISMIC as the first integrated algorithm (already proven in OpenSearch)
* Be designed for extensibility to accommodate future sparse ANN algorithms
* Follow similar patterns to established libraries like Faiss for dense vectors, in-memory engine 

The library is built here and almost ready: https://github.com/opensearch-project/neural-sparse-cpp

### 3.2 Algorithm Integration

The library will provide a unified interface for multiple sparse ANN algorithms:

|Algorithm	|Description	|Status	|
|---	|---	|---	|
|SEISMIC	|State-of-the-art sparse ANN algorithm	|First integration target	|
|Linscan(Inverted Index)	|Traditional sparse retrieval	|Baseline implementation	|
|Future algorithms	|Extensible plugin architecture	|Planned	|

### 3.3 Design Principles

* **Modularity**: Clean separation between index management, search execution, and algorithm implementations
* **Extensibility**: New algorithms can be added without modifying core interfaces
* **Performance**: Optimized for high-throughput, low-latency sparse vector operations
* **Portability**: Support for Linux and macOS platforms

### 4.1 High-Level Architecture

[...]

PR #1972, opened 2026-08-27
[Sparse ANN] [Native] Add the nsparse JNI layer for the native sparse engine

chishui:
### Description

First of two PRs splitting the native sparse ANN engine. This one adds **only the JNI layer**: the C++/JNI bridge to [neural-sparse-cpp](https://github.com/opensearch-project/neural-sparse-cpp) (nsparse) plus the CMake and Gradle wiring to build it. The plugin-side code that calls into it (codec, mapper, query) follows in a second PR, so nothing here is reachable from a user-facing path yet.

**What's in it**

- `jni/src` — the JNI entry points, an nsparse wrapper (index lifecycle, query, filtered query, serialization) and a streaming writer that pipes a serialized index through Lucene's `IndexOutput`.
- `jni/tests` — a googletest suite driving the real nsparse library against a fake `JNIEnv`, so the layer is testable without a JVM.
- Java side — `NativeLibrary` (native declarations and library loading), plus `SparseQueryResult` and `IndexOutputWrapper`, both of which native code resolves by name at load time.
- `build.gradle` — `cmakeJniLib` / `buildJniLib` / `jniTest` tasks, and `java.library.path` for the test and integTest JVMs and the test clusters.

The nsparse submodule is pinned to `439b992`.

**Index types supported:** `inverted`, `seismic`, `seismic_sq`, and `disk_seismic` (`DiskSeismicIndex`, which keeps cluster summaries in RAM and borrows per-document forward vectors from an mmap).

### Related Issues

Part of #1802 (design: Native Engine for Sparse ANN Search). Not closed by this PR — the plugin integration lands separately.

### Testing

| | Result |
|---|---|
| `./gradlew jniTest` (Linux) | 60 C++ tests pass |
| `./gradlew jniTest -PjniSanitizers=true` | 60 pass clean under ASan/LSan, zero leaks |
| `./gradlew check -x integTest` | green, 2588 Java tests |
| Windows Server 2025, MSVC 19.44 | 60 C++ tests pass; `gradlew.bat jniTest` and `gradlew.bat test` green |

Two new CI jobs: `JNI-native-tests` (ubuntu-latest and windows-latest) and `JNI-native-tests-sanitizers` (Linux only, since `-fsanitize` is GCC/Clang). All existing checkout steps now use `submodules: recursive`.

The sanitizer job is the substantive coverage for the ownership rules: this layer hands raw pointers across the language boundary and frees them natively, which is exactly what assertions cannot see.

### Notes for reviewers

Three things are deliberately **not** in this PR:

1. **Packaging.** The `.so`/`.dll` is not in the plugin zip; it is found via `-Djava.library.path=$rootDir/jni/build` pointing into the build tree. Following k-NN's model needs three pieces — build every SIMD variant, append `lib/` to the zip, and add `plugins/opensearch-neural-search/lib` to the `opensearch-build` startup scripts. The third is a separate repo. Tracked as follow-up.
2. **Runtime SIMD selection.** One variant is built, for whatever machine ran the build, and the Java side locates it by probing file names. That is fine for build-from-source but wrong for a distributed artifact, and it is the prerequisite for (1).
3. **Windows SIMD.** Windows is pinned to the generic nsparse build, matching k-NN. Worth knowing that this is *not* a toolchain limitation for us: k-NN builds Windows with MinGW, we use MSVC, and nsparse has MSVC `/arch:AVX2` / `/arch:AVX512` branches. I verified `nsparse_avx512` compiles clean with MSVC, passes the suite, and produces a DLL containing real `zmm`-register instructions. The blocker is variant selection, i.e. item (2). The rationale is recorded in `jni/cmake/init-nsparse.cmake` so it is not re-derived incorrectly.

Some deliberate design points worth a look:

- The JNI sources compile into an **object library** consumed by both the shared library and the test binary. Linking tests against the shared library only works on ELF, which exports everything by default; on Windows a DLL exports only the `JNIEXPORT`'d entry points, which left 20 unresolved externals.
- `ScopedArrayElements` / `ScopedLocalRef` / `ScopedStringChars` in `jni_util.h` exist because hand-written `Get.../Release...` pairs skip the release on any throw, stranding the buffer for the life of the JVM — and for a copy-back release, silently discarding values the native side wrote.
- `NsparseStreamWriter::pos()` reports the absolute **file** offset, seeded from `IndexOutputWrapper.getFilePointer()`. nsparse pads each serialized array to its element alignment from `pos()`, and the mmap reader recomputes that padding from the offset in the file; a writer counting only its own bytes agrees only while the payload starts at offset 0. Only `DiskSeismicIndex` is exposed to this, since the other types copy and a copy is aligned wherever it lands.
- `NativeLibraryContractTests` pins the reflective Java↔native bindings (the Lombok-generated `(IF)V` constructor, `writeBytes([BII)V`, `getFilePointer()J`, native-method ↔ exported-symbol agreement both ways, and the `SIMD_SUFFIXES` ↔ `init-nsparse.cmake` pairing). These have no compile-time guard otherwise.

### Check List


chishui:
Analyzer: googletest now pinned to immutable commit b514bdc (was a mutable tag). The submodule finding needs maintainer ack: origin is first-party opensearch-project/neural-sparse-cpp per #1802, pinned to 99bc33f.

codecov/patch: uncovered lines are NativeLibrary's static loader, untestable without the shared library present.

jni/src/nsparse_wrapper.cpp
```diff
@@ -0,0 +1,394 @@
...
+        }
+    }
+    std::vector<nsparse::term_t> termTokens(tokens->begin(), tokens->end());
+
+    omp_set_num_threads(threadCount);
+    index->add_with_ids(static_cast<nsparse::idx_t>(numIds),
+                        reinterpret_cast<const nsparse::idx_t*>(indptr->data()),
+                        termTokens.data(), values->data(),
+                        reinterpret_cast<const nsparse::idx_t*>(ids));
+
```
yuye-aws, line 292: Is there any check on the valid inputs?

jni/src/nsparse_wrapper.cpp
```diff
@@ -0,0 +1,394 @@
...
+        if (token < 0 || token > termLimit) {
+            throw std::invalid_argument(
+                "sparse token id " + std::to_string(token) +
+                " is outside the range [0, " + std::to_string(termLimit) +
+                "] supported by the native engine");
+        }
+    }
+    std::vector<nsparse::term_t> termTokens(tokens->begin(), tokens->end());
+
+    omp_set_num_threads(threadCount);
```
yuye-aws, line 287: Is there any where to restore the omp thread count?

jni/src/common.h
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * If an address is 0, a new vector is heap-allocated and the address is written back.
+ * Otherwise, data is appended to the existing vector.
+ */
+inline void transferVectors(
+    int64_t* memoryAddresses,
+    const int32_t* indices, int indicesLen,
+    const int32_t* tokens, int tokensLen,
+    const float* weights, int weightsLen
+) {
```
yuye-aws, line 33: There is an edge case:

  1. Call 1 with indicesLen == 0 → takes the allocate branch, produces a live but
  empty vector with a non-null data().
  2. Call 2 with real data → sees an allocated vector, takes the append branch,
  calls back() on empty.

The fix should be int32_t offset = indicesVec->empty() ? 0 : indicesVec->back();

jni/src/nsparse_wrapper.cpp
chishui: on API layer, we have input data verification

jni/src/nsparse_wrapper.cpp
chishui: Yes, it's from `plugins.neural_search.sparse.algo_param.index_thread_qty` setting

jni/src/common.h
chishui: Good catch — it's a `SIGSEGV`, not a misread.

But `empty() ? 0 : back()` still skips `indices[0]`, giving `{3, 7}` instead of `{0, 3, 7}`: silent CSR corruption. Keyed the branch off `indicesVec->empty()` instead — an empty destination copies the indptr verbatim.

Two tests added, 62 pass. 8448ef25

yuye-aws (approved):
Please fix the merge conflicts

PR #1973, opened 2026-08-28
[Fusion] Fixed score override when request contains rescore

martin-gaievski:
### Description
`rescore` on a fused query may override the hybrid's score scale and reorder its hits, but it must not promote a document the fusion never ranked. The second half was broken by two independent mechanisms, one of which fires at **entirely default configuration**.

`rescore` is deliberately never propagated to a leg sub-search (`CandidateScope`, `Disposition.NOT_PROPAGATED`):
it is core's own two-round mechanism and runs once per shard, after the coordinator has already fused and
self-erased the `hybrid` into

```
bool {
  should: constant_score(bool{ filter: ids(<fused window>), filter: term(_index, …) })^<fused score>,  // Top
  filter: bool{ should: [ <every leg query> ] }                                                        // Tail
}
```

Where Tail is a non-scoring `filter`. Round 2 therefore matches **every document any leg matched**, and every
document outside the fused window sits at exactly `0.0` , legal rescore candidate. Two request-independent
multipliers decide how many of them a rescore actually reaches:

1. **The rescore window can exceed the fused window.** `rescore.window_size` is the user's own number and has
   no relation to `fusion.window_size`. The moment it is larger, the surplus is filled from the Tail.
2. **Per-shard shortfall.** Core sizes the rescore window against the **shard's** own reader
   (`RescoreContext`/`QueryPhase`), while the fused window is coordinator-global. A shard holding 2 of 5 ranked
   documents rescores 3 unranked ones **even when `rescore.window_size` equals `fusion.window_size` exactly**.

To fix this  each `rescore` query is rewritten to `bool{must: <the user's query>, filter: <the fused window>}`: neither
multiplier can widen what the rescore query matches, and the user's query stays the only scoring clause, so its
contribution to the score is untouched. Because it is the **query** that is confined rather than the page, the
same change covers the `top_hits` site. Separately, a fused score of exactly `0.0` is floored to `Float.MIN_VALUE`, that keeps a ranked document above the non-scoring Tail while leaving every existing ordering intact, since every other positive float is already at or above it.

Below are several example where rescore can get incorrect hits and how that changes after this PR

#### Example 1 — specification

The hybrid ranks `[1, 2, 3, 4, 5]`; the rescore matches `[5, 6]`. Document 5 must be promoted to the top — the
rescore has overridden the hybrid's scale outright — and document 6 must not appear, because the hybrid never
ranked it.

```json
POST /rescore/_search
{
  "size": 5,
  "query": <fused>,
  "rescore": { "window_size": 20,
               "query": { "rescore_query": { "ids": { "values": ["5", "6"] } },
                          "query_weight": 1.0, "rescore_query_weight": 10.0, "score_mode": "total" } }
}
```

| | Page | `_score` of the first hit |
|---|---|---|
| Before | `[5, 6, 1, 2, 3]` — document 6 lifted, document 4 evicted | `10.001` |
| After | `[5, 1, 2, 3, 4]` | `10.001` |

The promotion and the scale are unchanged; only the unranked document is gone. With a rescore query that
matches *every* document from 5 up (`range`), the before-page is `[5, 6, 7, 8, 9]` — four of five slots.

#### Example 2 — stock case: default rescore window on a sharded index

`window_size` is **absent**, so core's default applies.

```json
POST /rescore-20-shards/_search
{
  "size": 5,
  "query": <fused>,
  "rescore": { "query": { "rescore_query": { "range": { "s": { "lte": 995 } } },
                          "query_weight": 1.0, "rescore_query_weight": 10.0, "score_mode": "total" } }
}
```

| | Page |
|---|---|
| Before | **`[5, 11, 13, 17, 19]`** |
| After | `[5, 1, 2, 3, 4]` |

20 shards over 30 documents leaves most shards holding no ranked document at all, so each shard's rescore
window is filled almost entirely from the Tail. This is the row to read: no oversized window, no adversarial
query, nothing the user could be told to avoid.

#### Example 3 — the second rescore site: inside a `top_hits` bucket

```json
POST /rescore/_search
{
  "size": 0,
  "query": <fused>,
  "aggs": { "all": { "filter": { "match_all": {} }, "aggs": { "top": { "top_hits": { "size": 5 } } } } },
  "rescore": { "window_size": 20,
               "query": { "rescore_query": { "range": { "s": { "lte": 995 } } },
                          "query_weight": 1.0, "rescore_query_weight": 10.0, "score_mode": "total" } }
}
```

| | Bucket's five visible hits |
|---|---|
| Before | `[5, 6, 7, 8, 9]` |
| After | `[5, 1, 2, 3, 4]`, document 5 at `10.001`, document 1 at `1.0` |

The bucket still **sees** every document the Tail matched — that is what an aggregation is for, and documents
6–10 are present below at `0.0`. What stops is the promotion.

#### Example 4 — score tie, with no rescore anywhere in the request

Three legs at weights `[1.0, 0.0, 0.0]`: the weighted leg matches only documents 1–4, `ids[26]` contributes
document 26 at weight `0.0` (so it fuses to exactly `0.0`), and a third `match_all` leg at weight `0.0` puts
documents 6–25 in the Tail.

```json
POST /rescore/_search   { "size": 6, "query": <fused, weights [1.0, 0.0, 0.0]> }
```

| | Page |
|---|---|
| Before | `[1, 2, 3, 4, 5, 6]` — document 6 was never ranked; document 26, which was, is gone |
| After | `[1, 2, 3, 4, 5, 26]`, and no ranked document is reported at `0.0` |

#### Example 5 — what still works

Row 2 is the reason this is *confinement* rather than a ban: when the fused window covers the corpus, every
document is a ranked document and a rescore may legitimately promote any of them. It is the one behaviour a
too-narrow fix would take away, and it is pinned by a test that only a window filter admitting all 30
documents can satisfy.

### Related Issues

### Check List


src/main/java/org/opensearch/neuralsearch/query/FusedRescoreScope.java
```diff
@@ -0,0 +1,300 @@
...
+ * placeholders are never rewritten, which {@link #requireReachedTheExecutedRequest} detects at round 2 and turns into a
+ * failed request. The one thing that must not happen — returning results with an unconfined rescore — is the one thing
+ * that cannot.
+ *
+ * <p><b>Known narrowing with more than one fused hybrid in a request.</b> Each fused hybrid installs its own placeholder
+ * over the shared rescore list, so with two of them under a {@code bool} the second wraps the first's replacement and the
+ * rescore ends up confined to the <i>intersection</i> of the two windows, where the union would be the faithful answer.
+ * That is deliberate: it is the conservative direction — the intersection can only under-apply the boost, never lift a
+ * document neither hybrid ranked — and reaching it needs no structural inspection of the user's own rescore query, which
+ * any "have I already confined this?" check would. A single fused hybrid, the case this is written for, is exact.
```
vibrantvarun: This multi-fused-hybrid **intersection narrowing** is a deliberate and safe (conservative) limitation, but it's the one behavior here not pinned by a test. Since it depends on install order (each hybrid wrapping the previous replacement) and the safe direction is load-bearing — the intersection must never become a union/lift of a doc neither hybrid ranked — a test with two fused hybrids under a `bool` asserting the rescore is confined to the intersection would lock the intent in and guard against a future refactor silently widening it. Everything else in the class is thoroughly covered.

src/main/java/org/opensearch/neuralsearch/query/FusedRescoreScope.java
```diff
@@ -0,0 +1,300 @@
...
+ * widen what the rescore query matches. The user's query keeps its scoring contribution untouched — it is the sole
+ * {@code must}, so the bool's score is exactly its score, and the window rides along as a non-scoring {@code filter}.
+ * Rewriting the query rather than the page is also what makes one fix cover core's <i>other</i> application site:
+ * {@code TopHitsAggregator} widens every {@code top_hits} bucket's collector by {@code rescore.window_size} and rescores
+ * that bucket with the request's own rescore contexts.
+ *
+ * <p><b>Why the window arrives late, and why that dictates this shape.</b> The window is not known until the leg
+ * {@code MultiSearch} comes back, which is an async action — so the obvious implementation (fuse, then edit
+ * {@code source.rescores()} from the callback) writes to the rescore list <i>after</i> core has already snapshotted it.
+ * {@code SearchSourceBuilder#rewrite} rewrites the query at {@code :1187} and the rescorers at {@code :1199} in the same
```
vibrantvarun: nit: the design relies on core rewriting the query *before* the rescorer list within one `SearchSourceBuilder#rewrite` pass — which is exactly right — but citing the concrete line numbers (`:1187`/`:1199`, also at line 133) will go stale as core evolves. Consider phrasing it as the ordering contract ("the query is rewritten before the rescorer list in the same pass") without the line refs, so the rationale survives a core reformat.

src/main/java/org/opensearch/neuralsearch/query/FusedRescoreScope.java
```diff
@@ -0,0 +1,300 @@
...
+     * defect this class exists to close, silently and only for that rescorer.
+     *
+     * <p>Refused at install time, so the refusal depends on the request alone — not on whether this particular fusion
+     * happened to rank a document.
+     */
+    private static QueryRescorerBuilder requireQueryRescorer(final RescorerBuilder<?> declared) {
+        if (declared instanceof QueryRescorerBuilder queryRescorer) {
+            return queryRescorer;
+        }
+        throw new IllegalArgumentException(
```
vibrantvarun, line 202: Just confirming this is the intended behavior change (it's in the CHANGELOG): a non-`query` rescorer (e.g. a custom/plugin rescorer that classic hybrid would accept) is now a hard 400 in fused mode. Makes sense given confinement only works by rewriting the rescore query, which only `QueryRescorerBuilder` exposes — flagging it as the user-visible restriction to be sure it's deliberate rather than incidental.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -353,11 +381,54 @@ private static RankedDocs toRankedDocs(Map<String, Float> scoresByKey, Map<Strin
...
+     * wrong response to them: it would quietly collapse every negative or {@code NaN} score onto one value and destroy
+     * whatever order they had, which is the opposite of what this method is for. {@code -0.0f} is not one of them — it is
+     * zero, and is floored like any other zero.
+     *
+     * <p>This floor is applied here, at fusion time, rather than when the Top clauses are built, so the builder's own
+     * state, its {@code toXContent}/profile form and the {@code _score} the user is shown all agree on one number.
+     */
+    private static float scoreAboveTail(final float fusedScore) {
+        if (Float.isFinite(fusedScore) == false || fusedScore < 0.0f) {
+            throw new IllegalStateException(
```
vibrantvarun, line 466: nit: `scoreAboveTail` throws `IllegalStateException` for a non-finite/negative fused score, while `HybridFusionQueryBuilder#requireUsableAsBoosts` throws `IllegalArgumentException` for the same invariant. Both are defensive on an unreachable condition, so not important — just two exception types for one contract; might be worth aligning them.

vibrantvarun (commented):
Reviewed the rescore-confinement end-to-end (FusedRescoreScope + the score floor + the boost validation). Really solid: the late-window placeholder installed inside round 1's query rewrite (before core snapshots the rescore list) and verified fail-closed at round 2 is subtle but holds up, the `1e-30f` floor choice is genuinely load-bearing and well justified against subnormal rounding, and the test coverage maps to every scenario incl. the per-shard shortfall and both zero-score paths. No correctness defect found — a few minor notes inline, the only real ask being a test for the documented multi-hybrid intersection behavior.

src/main/java/org/opensearch/neuralsearch/query/FusedRescoreScope.java
martin-gaievski: ack, will add test

src/main/java/org/opensearch/neuralsearch/query/FusedRescoreScope.java
martin-gaievski: ack, will update doc comment

src/main/java/org/opensearch/neuralsearch/query/FusedRescoreScope.java
martin-gaievski: yup, that's correct understanding

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: exceptions are different by design, and I would rather keep them: in core the two types map to different HTTP statuses, so collapsing them loses information

src/main/java/org/opensearch/neuralsearch/query/FusedRescoreScope.java
vibrantvarun: Confirmed the test landed in fd666e583a (HybridQueryFusedFanOutTests) — thanks.

src/main/java/org/opensearch/neuralsearch/query/FusedRescoreScope.java
vibrantvarun: Verified the :1187/:1199 citations are dropped across the three javadocs and replaced with the ordering contract — good.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
vibrantvarun: Fair — you're right, IllegalStateException→500 (unreachable internal invariant) vs IllegalArgumentException→400 (input/wire contract) carry different information, so keeping them distinct is correct. Withdrawing the nit.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -761,6 +809,87 @@ public QueryBuilderVisitor getChildVisitor(final Occur occur) {
...
+
+    /**
+     * Records whether it was ever shown a fused hybrid. Walks the whole tree the way {@link LegSearchCounter} does —
+     * one instance, handed back as its own child visitor — so a hybrid at any depth the builders expose is seen.
+     */
+    private static final class FusedHybridFinder implements QueryBuilderVisitor {
+        private boolean found;
+
+        @Override
+        public void accept(final QueryBuilder queryBuilder) {
```
vibrantvarun: `FusedHybridFinder` here relies on each builder's `visit()` to recurse (it records the match but doesn't call `queryBuilder.visit(this)` itself) — the same pattern #1969 replaced in `IdentityFinder`/`LegSearchCounter` precisely because `boosting`, `script_score` and the `span_*` builders hand their inner query to `accept()` **without recursing**. So `containsFusedHybrid` catches a fused hybrid one level under a `boosting` but misses one ≥2 levels down (`boosting{positive: bool{should: <fused>}}`).

Not a safety hole — such a request still fails closed at round 2 rather than running an unconfined rescore — but for that shape the intended clean *pre-fan-out* 400 doesn't fire (it degrades to a wasted fan-out + a less-specific refusal), and the javadoc's "a hybrid this walk cannot see — a `wrapper` query carries its inner query as bytes" enumerates only `wrapper`, omitting the `boosting`/`script_score` non-recursion. Suggest making this finder descend itself (identity-deduped `accept → visit`) like the two #1969 finders, which makes it exact at any depth and reduces the "cannot see" set to just the genuinely-opaque `wrapper`/`template`.

vibrantvarun (commented):
Re-reviewed fd666e583a — the four earlier threads are all addressed (or, for the exception types, rightly rebutted), and this commit also fixed a real defect (fused hybrid in a rescore query 500'ing after fan-out → clean 400) plus the atomic write-back and the parallel-arrays check. One new consistency finding inline on the fused-hybrid-in-rescore detector; it's LOW (fail-closed still holds), not blocking.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
martin-gaievski: Right on both counts, and thanks — confirmed locally that boosting{positive: bool{should: <fused>}} slips past this finder, exactly as you describe

Rather than teach it to descend, the whole guard is gone. #1969 merged into the base branch while this PR was in review, and its requireFusedQueryIsPartOfRequestQuery already refuses a fused hybrid the request's own query does not expose — rescore included — from the top of doRewriteFused, i.e. ahead of where this ran. Its IdentityFinder self-descends, which is the fix you suggested, so the refusal is now exact at any depth and the "cannot see" set really is just wrapper/template. It's also already covered twice: testFusedHybridOutsideTheRequestQuery_isRefusedBeforeAnyFanOut (whose position table includes rescore) and HybridQueryFusedQueryPositionIT's REST rescore body asserting a 400.

PR is rebased onto #1969

vibrantvarun (approved):
LGTM on 7ce0bd1299. All review threads resolved: the intersection test and doc-ref cleanup landed, the exception-type distinction is rightly kept, and my FusedHybridFinder finding is closed the better way — the guard is removed entirely now that the PR rebases onto #1969, whose requireFusedQueryIsPartOfRequestQuery already refuses a fused hybrid the request's query doesn't expose (rescore included), self-descends via IdentityFinder so it's exact at any depth, and runs ahead of where the removed guard did. The core rescore-confinement (late-window placeholder installed pre-snapshot, fail-closed verification, 1e-30 floor) is sound and well-tested. Nice work chasing the rescore-nested-hybrid 500 and the atomic write-back out of review.

src/main/java/org/opensearch/neuralsearch/query/FusedRescoreScope.java
```diff
@@ -0,0 +1,306 @@
...
+ * this pass — the same assumption every other fused-mode read of the request makes. If it ever breaks (a
+ * search-pipeline request processor handing back a different {@code SearchRequest} whose source is a deep copy), the
+ * placeholders are never rewritten, which {@link #requireReachedTheExecutedRequest} detects at round 2 and turns into a
+ * failed request. The one thing that must not happen — returning results with an unconfined rescore — is the one thing
+ * that cannot.
+ *
+ * <p><b>Known narrowing with more than one fused hybrid in a request.</b> Each fused hybrid installs its own placeholder
+ * over the shared rescore list, so with two of them under a {@code bool} the second wraps the first's replacement and the
+ * rescore ends up confined to the <i>intersection</i> of the two windows, where the union would be the faithful answer.
+ * That is deliberate: it is the conservative direction — the intersection can only under-apply the boost, never lift a
```
bzhangam: I think there also can be case the hybrid is in a must_not clause and in that case our current logic will fail. I'm thinking since the issue is introduced by the global leg combined with the final rank doc query why not simply dispatch them as two requests and merge them to produce the final response?

martin-gaievski: I've checked `must_not` both ways and it doesn't error, though you're pointing at something real. Measured on the fan-out harness:
 - bool{must_not:[<fused hybrid>]} is admitted by #1969's reachability guard (its IdentityFinder returns the same child visitor for every Occur, MUST_NOT included), fans out once, and settles to bool{must_not:[hybrid_fusion{…}]} — no throw.
- same with a rescore is also clean: rescore is still confined to bool{must:<userQ>, filter:hybrid_fused_window}, requireReachedTheExecutedRequest() passes.
 
Most importantly the `must_not` excludes window from the result set, so the window filter on the rescore matches none of the surviving docs → rescore no-ops. It can't add to the hybrid's hits because there are none left, so there's no rescore defect to fix here.

What you're really flagging is that, because fused mode splices the ranking into the query tree, any score-transforming context around it changes what fusion means — must_not negates it, a low boost rescales it, a bare should makes it optional. must_not is the sharpest case, but it's a property of the rewrite-and-splice approach and a policy question for #1969's position guard (which today deliberately admits any identity-reachable position, looser than existing top-level-only rule), not something this PR introduces. If we want to refuse score-inverting positions like must_not, I'm happy to file that as a follow-up on the #1969 guard, just it's not directly rescore functionality.
 
On two-requests-and-merge: agreed that's the cleaner long-term shape, and it's exactly where Retriever core work is going — resolve the ranked docs, then run/merge the final query coordinator-side, so the ranking never has to survive arbitrary boolean composition. This code rewrites-and-splices because it runs entirely on the coordinator via existing async-rewrite + msearch with no core changes, which is what let fused mode ship on top of hybrid; tradeoff is the composition fragility you're pointing at. I'd rather not swap architectures inside this rescore PR, but it's the right strategic call, let's keep track of it in https://github.com/opensearch-project/OpenSearch/issues/22826 thread.

bzhangam: ok if we want to limit the scope. But if we plan to simply use must clause to build the rescore window should we block other use case? Fail the request is better than silently return inaccurate response?

martin-gaievski: I measured possible effect across all score-transforming placements that #1969 allows by driving the whole source through core's rewrite . In every one rescore is still limited to bool{must: <userQ>, filter: hybrid_fused_window} and none throw or lets the rescore reach a non-window doc. Placements I checked: must_not, boosting(negative), function_score, optional should, constant_score

That's by construction: window filter sits on the rescore query itself, so whatever the enclosing query does, the rescore can only ever match the fused window — it can't promote a doc fusion didn't rank, in any placement.

What changes across those placements is the ranking of the fused subquery: must_not excludes the window, function_score multiplies the fused score, constant_score flattens it, boosting-negative demotes it. those are defined compositions of the hybrid's self-erased form, and several are things people legitimately want (function_score to post-scale fusion by a field, constant_score to use fusion as a pure filter). They aren't wrong answers to fail on — blocking them would refuse useful queries.

So I'd keep this PR as-is — the rescore boundaries are safe everywhere.

martin-gaievski: after deep dive I found that in general results we return can be wrong, so I pushed new comment where we are blocking queries with rescore if resolver is not the top clause

bzhangam (approved):

PR #1977, opened 2026-08-31
[Fusion] Detailed per-leg profiler in fused mode

martin-gaievski:
### Description
With this change hybrid with `fusion` block and `profile: true` reports one profile entry per leg, per shard. Today a profiled fused hybrid returns a single shard entry whose tree is the query the fused rewrite *substituted* — `bool{ constant_score(_id:X)^norm … }` plus the non-scoring Tail — so the kNN leg never appears as a kNN query anywhere, and the entry that is there is unlabelled.

Request:

```json
POST /test-fused-profile-research/_search
{
  "profile": true,
  "track_total_hits": true,
  "query": {
    "hybrid": {
      "fusion": {
        "window_size": 10,
        "normalization": { "technique": "min_max" },
        "combination": { "technique": "arithmetic_mean" }
      },
      "queries": [
        { "knn":  { "vec":  { "vector": [1.1, 1.0], "k": 10 } } },
        { "term": { "text": "hello" } }
      ]
    }
  }
}
```
Response from classic hybrid query (baseline, with normalization pipeline, no fusion in request)

```json
{
  "took": 39,
  "_shards": { "total": 1, "successful": 1, "skipped": 0, "failed": 0 },
  "hits": { "total": { "value": 6, "relation": "eq" }, "max_score": 1.0, "hits": [ { "…": "… the 6 docs …" } ] },
  "profile": {
    "shards": [
      {
        "id": "[8bJ4SbGxTBmEVao2Ur_Gmg][test-fused-profile-research][0]",
        "searches": [
          {
            "query": [
              {
                "type": "HybridQuery",
                "description": "(OSKnnFloatVectorQuery:vec[1.1,...][100] | text:hello)",
                "time_in_nanos": 9178878,
                "breakdown": { "create_weight": 3166883, "build_scorer": 5970327, "next_doc": 41668, "score": 0, "…": "…" },
                "children": [
                  {
                    "type": "LuceneEngineKnnVectorQuery",
                    "description": "OSKnnFloatVectorQuery:vec[1.1,...][100]",
                    "time_in_nanos": 1545871,
                    "breakdown": { "create_weight": 1500640, "…": "…" },
                    "children": [
                      { "type": "OSKnnFloatVectorQuery", "description": "OSKnnFloatVectorQuery:vec[1.1,...][100]", "time_in_nanos": 0, "breakdown": { "…": "… all zero …" } }
                    ]
                  },
[...]

PR #1979, opened 2026-09-01
[Fusion] Explain a fused hybrid score from its legs' explanations

martin-gaievski:
### Description

Adding `explain` for fused `hybrid` query. Structurally this is the `profile` mechanism of #1977 a second time, on the same approach and with the same hook: filter attaches a collector before the rewrite, fan-out fills it, filter's response listener reads from it. New DTO class FusedDocExplanations carry the report. 

Where we attach new explain classes, and where it deliberately do not

| Report | Attaches to | Why not more |
|---|---|---|
| `profile` (#1977) | every fused hybrid in the request | a profile section is a list, so N hybrids can each contribute |
| `explain` (this PR) | **only** the fused hybrid that *is* the request's own `query`, by reference identity | a hit's explanation is one tree describing that hit's score; nested, the fused score is an input to the enclosing query's scoring, so replacing the tree would describe the wrong thing |

It follows same main fusion/resolver paradigm - no pipeline or processor is required, specifically existing [explanation response processor](https://docs.opensearch.org/latest/search-plugins/search-pipelines/explanation-processor/). One more advantage over existing logic for explain - with this change we run actual query once and explain part only processes scores differently, so it's faster. Today explain in hybrid query runs query phase twice to get raw queries scores. 

---

#### Query examples

Index used by the ITs: 6 documents, ids `1`–`6`, each with the same `text` and a 2-dimensional `vec`, queried by a
`knn` leg and a `term` leg that matches all six.

#### 1. The flat case — every leg described under the fused score

```json
GET test-fused-explain/_search
{
  "explain": true,
  "track_total_hits": true,
  "size": 7,
  "query": {
    "hybrid": {
      "fusion": {
        "window_size": 10,
      },
      "queries": [
        { "knn":  { "vec": { "vector": [1.1, 1.0], "k": 10 } } },
        { "term": { "text": "hello" } }
      ]
    }
  }
}
```

Each ranked hit carries:

```json
"_explanation": {
  "value": 0.75,
  "description": "arithmetic_mean combination of:",
  "details": [
    {
      "value": 1.0,
      "description": "min_max normalization of:",
      "details": [
        { "value": 0.98, "description": "within top 10 docs", "details": [] }
      ]
    },
    {
      "value": 0.5,
      "description": "min_max normalization of:",
      "details": [
        { "value": 0.287, "description": "weight(text:hello in 0) [PerFieldSimilarity], result of:", "details": [ "..." ] }
      ]
    }
  ]
}
```

#### 2. Weighted combination — the weights are named, not folded in

```json
"combination": { "technique": "arithmetic_mean", "parameters": { "weights": [0.7, 0.3] } }
```

```json
"_explanation": {
  "value": 0.85,
  "description": "arithmetic_mean, weights [0.7, 0.3] combination of:",
  "details": [ "…leg 0, value 1.0…", "…leg 1, value 0.5…" ]
}
```

`0.7 × 1.0 + 0.3 × 0.5 = 0.85` — the weights apply to the values the children report, so the children stay readable
as the normalized leg scores they are.

#### 3. document matching one leg

```json
{ "knn": { "vec": { "vector": [1.1, 1.0], "k": 2 } } }
```

With `k: 2` the ANN leg contributes two candidates, so the other four documents reach fusion through the `term` leg
alone. Their trees carry **one** leg node, not two with a zero — the same choice classic makes.

#### 4. `rescore` — the fused score nests under the final score

```json
{
  "explain": true,
  "query": { "hybrid": { "fusion": { "...": "..." }, "queries": [ "..." ] } },
  "rescore": {
    "window_size": 10,
    "query": {
      "rescore_query": { "term": { "text": "hello" } },
      "query_weight": 1.0,
      "rescore_query_weight": 2.0
    }
  }
}
```

When the score round 2 returned is not the fused score, an extra node is inserted above the combination:

```json
"_explanation": {
  "value": 1.324,
  "description": "score of the fused hybrid query after post-fusion scoring, computed from:",
  "details": [
    { "value": 0.75, "description": "arithmetic_mean combination of:", "details": [ "..." ] }
  ]
}
```

`rescore` is the **only** way a fused hit's score can differ from its fused score: `boost` on a `hybrid` is rejected
at parse time (`[hybrid] query does not support [boost]`, HTTP 400).

### Related Issues

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
@@ -14,6 +14,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 
+import org.apache.lucene.search.Explanation;
 import org.opensearch.ExceptionsHelper;
 import org.opensearch.OpenSearchStatusException;
 import org.opensearch.action.search.MultiSearchRequest;
@@ -37,6 +38,8 @@ import org.opensearch.neuralsearch.processor.normalization.RRFScoreNormalizer;
 import org.opensearch.neuralsearch.processor.combination.ScoreCombinationFactory;
 import org.opensearch.neuralsearch.processor.combination.ScoreCombinationTechnique;
 import org.opensearch.neuralsearch.processor.combination.ScoreCombinationUtil;
+import org.opensearch.neuralsearch.processor.explain.ExplainableTechnique;
+import org.opensearch.neuralsearch.search.explain.FusedDocExplanations;
 import org.opensearch.neuralsearch.search.profile.FusedCoordinatorTimings;
 import org.opensearch.search.SearchHit;
 import org.opensearch.search.builder.SearchSourceBuilder;
@@ -66,8 +69,6 @@ import lombok.NoArgsConstructor;
 final class HybridFusionOrchestrator {
 
     private static final ScoreCombinationFactory SCORE_COMBINATION_FACTORY = new ScoreCombinationFactory();
-    /** Separator for the composite _index+_id fusion key. Never parsed back — see computeRankedDocs. */
-    private static final String KEY_SEPARATOR = "#";
     /** The {@code _name} key as it appears in a rendered query — see {@link #anyLegNamed}. */
     private static final String QUERY_NAME_KEY = String.format(Locale.ROOT, "\"%s\":", AbstractQueryBuilder.NAME_FIELD.getPreferredName());
 
@@ -161,14 +162,23 @@ final class HybridFusionOrchestrator {
         FusionSpec fusion,
         int windowSize
     ) {
-        return buildFusedQuery(source, multiSearchResponse, legs, fusion, windowSize, new FusedCoordinatorTimings());
+        return buildFusedQuery(
+            source,
+            multiSearchResponse,
+            legs,
+            fusion,
+            windowSize,
+            new FusedCoordinatorTimings(),
+            new FusedDocExplanations()
+        );
     }
 
     /**
-     * As above, recording each fusion phase's span into {@code timings} for the coordinator profile entry. The timings
-     * instance is always present — never null-checked here — so that instrumentation costs the same handful of
-     * {@code nanoTime} calls whether or not the request asked to be profiled, and an unprofiled request simply discards
-     * what was measured. Phase boundaries are the method boundaries below, so a span is never attributed to two phases.
+     * As above, recording each fusion phase's span into {@code timings} for the coordinator profile entry and each window
+     * document's per-leg breakdown into {@code explanations} for the request's {@code explain}. Both instances are always
+     * present — never null-checked here — so that instrumentation costs the same handful of {@code nanoTime} calls whether
+     * or not the request asked to be profiled, and a request that asked for neither simply discards what was collected.
+     * Phase boundaries are the method boundaries below, so a span is never attributed to two phases.
      */
     static QueryBuilder buildFusedQuery(
         SearchSourceBuilder source,
@@ -176,13 +186,14 @@ final class HybridFusionOrchestrator {
         List<QueryBuilder> legs,
         FusionSpec fusion,
         int windowSize,
-        FusedCoordinatorTimings timings
+        FusedCoordinatorTimings timings,
+        FusedDocExplanations explanations
     ) {
         MultiSearchResponse.Item[] items = multiSearchResponse.getResponses();
         long windowMergeStart = System.nanoTime();
         SearchHit[][] legHits = groupLegHits(items, legs.size());
         timings.windowMergeNanos(System.nanoTime() - windowMergeStart);
-        RankedDocs ranked = computeRankedDocs(legHits, fusion, windowSize, timings);
+        RankedDocs ranked = computeRankedDocs(legHits, fusion, windowSize, timings, explanations);
         timings.rankedDocs(ranked.ids().length);
         if (ranked.ids().length == 0) {
             return new MatchNoneQueryBuilder();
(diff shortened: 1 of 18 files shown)


vibrantvarun (commented):
Intense review — traced the explain path end-to-end (FusedDocExplanations, FusedExplanationMerger, fuseDetailed, recordExplanations, the identity attach, and the merge order). It's solid: fuse() now returns fuseDetailed().fused() so explained/unexplained scores can't diverge; legNormalizedScores is exactly what the combiner consumed; the top-level-only identity attach is correct; and profiles-then-explanations merge order is right. One LOW correctness inconsistency in the explain tree for an edge case (inline), plus one robustness nit on the cast. Approve-worthy after the first.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -333,14 +358,57 @@ private static RankedDocs computeRankedDocs(SearchHit[][] legHits, FusionSpec fu
...
+            List<FusedDocExplanations.LegContribution> contributions = new ArrayList<>(legNormalizedScores.size());
+            for (int leg = 0; leg < legNormalizedScores.size(); leg++) {
+                Float normalizedScore = legNormalizedScores.get(leg).get(key);
+                if (Objects.isNull(normalizedScore)) {
+                    // The leg did not match this document. Classic renders no node for it either.
+                    continue;
+                }
+                contributions.add(new FusedDocExplanations.LegContribution(leg, normalizedScore, legExplanations.get(leg).get(key)));
+            }
+            explanations.addDocument(key, ranked.scores()[doc], contributions);
```
vibrantvarun: This stores `ranked.scores()[doc]` as the combination-node value, but that value is post-`scoreAboveTail` (line 504), which floors a `0.0` fused score to a tiny positive epsilon (#1973). The node's leg children, however, are the normalized values, which arithmetic-mean to the **raw** `0.0`. Because the Top boost is also the floored score, `hitScore == fusedScoreByKey`, so `FusedDocExplanations.explain` returns the combination node directly (no `FINAL_SCORE` wrapper) — a tree whose parent value (~epsilon) its children don't produce (0.0).

Reachable exactly per #1973's own analysis: a `weights` entry of `0.0` for a doc matching only that leg, or `l2` over an all-zero leg (`MIN_SCORE == 0.0`).

Suggest passing the **raw** fused score instead:
```java
explanations.addDocument(key, fused.fused().get(key), contributions);
```
`fused` and `key` are already in scope and the value is non-null for any ranked doc. That makes the combination node consistent with its children **and** surfaces the floor naturally through the existing `FINAL_SCORE_DESCRIPTION` node (since `hitScore` (floored) would then differ from `fusedScore` (raw)) — which is what that "score after post-fusion scoring" node is for. Worth a test pinning a `weights:[0,1]` (or l2-zero-leg) doc's explanation.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -333,14 +358,57 @@ private static RankedDocs computeRankedDocs(SearchHit[][] legHits, FusionSpec fu
...
+        final List<Map<String, Explanation>> legExplanations,
+        final ScoreCombinationTechnique combination,
+        final ScalarNormalizer normalizer
+    ) {
+        if (legExplanations.stream().allMatch(Map::isEmpty)) {
+            return;
+        }
+        // Same format strings as classic: ScoreCombiner#explainByShard and ExplanationUtils#getDocIdAtQueryForNormalization.
+        explanations.combinationDescription(
+            String.format(Locale.ROOT, "%s combination of:", ((ExplainableTechnique) combination).describe())
```
vibrantvarun: nit/robustness: this cast to `ExplainableTechnique` is unguarded. All four current combination techniques implement it and classic hybrid does the identical cast, so it's safe today — but `ScoreCombinationTechnique` doesn't enforce `ExplainableTechnique`, so a future technique that omits it would throw a `ClassCastException` (500) only on explained requests. An `instanceof` guard (skip the fused breakdown, keep round-2's explanation) or folding `describe()` into the combination contract would make it future-proof.

martin-gaievski: ack, Your second option is the right fix and it's free, since `describe()` is already a default method. I'll make
`ScoreCombinationTechnique extend ExplainableTechnique` and all three casts delete themselves

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: ack, and taking the fix

martin-gaievski:
> Intense review — traced the explain path end-to-end (FusedDocExplanations, FusedExplanationMerger, fuseDetailed, recordExplanations, the identity attach, and the merge order). It's solid: fuse() now returns fuseDetailed().fused() so explained/unexplained scores can't diverge; legNormalizedScores is exactly what the combiner consumed; the top-level-only identity attach is correct; and profiles-then-explanations merge order is right. One LOW correctness inconsistency in the explain tree for an edge case (inline), plus one robustness nit on the cast. Approve-worthy after the first.

I take two action out of this comment:
1. javadoc in test is factually wrong. `HybridQuerySearchRequestFilterTests` claims test "pins that the one rebuild carries the very `SearchHit` instances over: a rebuild that copied them would drop the explanation written afterwards." **It pins no such thing** — under profiles-first the explanation is written *after*  the rebuild, so a copying rebuild would receive the write on the copies and the test would still pass. Reword to what it actually pins ("asking for both yields both") and point at `FusedLegProfileMergerTests.java:166` for the `assertSame` that does pin it.
2. **Coverage gap** `profile` + `explain` together is exercised only through mocked consumers in a unit
 test: `grep -c profile HybridQueryFusedModeExplainIT.java` = 0 and `grep -c explain HybridQueryFusedModeProfileIT` = 0. Adding `profile: true` to one existing explain IT case closes it.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
vibrantvarun: Confirmed fixed in f365c1e132 — now records `fused.fused().get(key)`, so a floored-0.0 doc's combination node reads 0.0 (consistent with its leg children) under the FINAL_SCORE node showing the actual round-2 score. Tests cover it too. Thanks.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
vibrantvarun: Just flagging the interface change didn't land in f365c1e132 (that commit was Finding 1 only) — `ScoreCombinationTechnique` still doesn't extend `ExplainableTechnique` and the cast at line 396 is still there. Non-blocking nit, so no objection to merging without it, but noting it in case the follow-up got dropped.

martin-gaievski: yes, my miss, pushed commit with the change

vibrantvarun (approved):
LGTM on 739d6c5cee (rebased clean off the earlier conflict). Both review findings resolved and re-verified at this head: the explain combination node now records the raw computed fused score (fused.fused().get(key)) so a floored-0.0 doc reads 0.0 under the FINAL_SCORE node — consistent with its leg children — and ScoreCombinationTechnique now extends ExplainableTechnique with every (ExplainableTechnique) cast removed, so an explained request can't 500 on a future technique. Explain path is sound overall: fuse() == fuseDetailed().fused() so scores can't diverge, legNormalizedScores is exactly what the combiner consumed, and explain attaches only to the top-level fused hybrid by identity. Thanks for the quick turnaround.

PR #1980, opened 2026-09-01
[Fusion] Report a soft-timeout-truncated leg on a fused hybrid response

martin-gaievski:
### Description
Follow-up to a behavior note raised in the [review comments](https://github.com/opensearch-project/neural-search/pull/1977#pullrequestreview-5072244086) from @vibrantvarun in #1977.

A fused leg that exceeds request's soft `timeout` returns the candidates it had rather than failing. Due to that fused
window may have missing hist comparing to complete run's.  `timed_out` stayed `false` on a request whose ranking was
computed from a truncated candidate set.

We already has access to timed out flag in `HybridQueryBuilder`. Based on that falg existing check in  `attachFusedReporting` is extended. 

The union is a logical OR that only ever sets:

- a leg that finished cannot undo another leg's truncation;
- one truncated hybrid cannot be talked back down by a completed sibling hybrid;
- the response's own flag is an **input** to the OR rather than something to overwrite, since round 2 can time out on
  its own, entirely apart from the legs.

Partial shard failures and `terminate_after` are deliberately **not** reported. N legs search the same shards while a
response carries one set of shard counters and one `terminated_early`, so neither has a defensible union — a fused
hybrid would have to either double-count or discard, and both misreport.

Cluster-level `search.default_search_timeout` is not read either. Doing so means holding cluster settings and
re-reading them per request on a path whose whole justification is that it costs nothing when unused. Such a request
still sees round 2's own `timed_out`; it simply gets no fused contribution (it's today's behavior).

## Example

```json
GET test-index/_search
{
  "timeout": "5ms",
  "query": {
    "hybrid": {
      "fusion": {
        "window_size": 100,
      },
      "queries": [
        { "knn":  { "vec": { "vector": [1.1, 1.0], "k": 100 } } },
        { "term": { "text": "hello" } }
      ]
    }
  }
}
```

With this change, a run in which either leg was cut short answers:

```json
{ "took": 7, "timed_out": true, "hits": { "...": "..." } }
```

Before it, the same run answered `"timed_out": false` with nothing in the response indicating that the ranking came
out of a partial candidate set. A request that carries no `timeout` is untouched: no gate opens, no listener wrap, no
rebuild.

### Related Issues
#1930 

### Check List


src/main/java/org/opensearch/neuralsearch/search/FusedLegTimeoutMerger.java
```diff
@@ -0,0 +1,59 @@
...
+ * fused {@code hybrid} the request fans out, published to by the leg fan-out callback, read once when the response comes
+ * back. Unlike that one it is attached whether or not the request asked to be profiled — an incomplete answer has to be
+ * reported as one either way, and the profile entry's per-leg {@code timed_out} under {@code debug} is a diagnostic for
+ * someone already looking, not a flag a client can act on.
+ *
+ * <p>Across legs and across hybrids the signal is a logical OR, and only ever sets the flag: it means "some part of this
+ * answer is incomplete", which is what core's own {@code timed_out} means, and a leg that finished cannot make an
+ * incomplete answer complete. The two neighbouring per-leg signals are deliberately <b>not</b> reported, because they
+ * cannot be composed the same way — N legs search the same shards while a response carries exactly one set of
+ * shard counters and one {@code terminated_early}, so there is no union of them that is true of the response as a whole.
```
vibrantvarun: The rationale for not reporting partial shard failures / `terminated_early` is sound (N legs, one set of shard counters, no defensible union). Just confirming the intended boundary: a fused window narrowed by a **leg shard failure** (rather than a soft timeout) still answers `timed_out:false` with nothing else signalling the ranking came from an incomplete candidate set — i.e. this closes the timeout case but leaves the shard-failure case as-is. That's no worse than today and I think it's the right scope, but it's the one user-visible gap that survives this PR, so worth having on the record as deliberate.

src/main/java/org/opensearch/neuralsearch/search/FusedResponseRebuilder.java
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * <p>Reconstructed from {@link SearchResponseSections#profile()} because the section itself is {@code protected} in
+     * core and unreachable from here. That round-trip is exact in every case this can be reached in: a rebuild that
+     * substitutes nothing is one driven by {@code timed_out} alone, which means the request was not profiled, which means
+     * core built no profile section at all. The one divergence it cannot express — a section that is present but holds no
+     * shard entries, which would render as an empty {@code profile} block and comes back as no block — needs a search
+     * pipeline to have switched profiling on after this plugin's {@code ActionFilter} already saw the request, and a
+     * profiled search that resolved to zero shards.
+     */
+    private static SearchProfileShardResults ownProfileResults(final SearchResponseSections sections) {
```
vibrantvarun, line 100: Minor/doc: this reasons that reaching `ownProfileResults` (profileResults == null) implies the request was unprofiled, hence core built no profile section. That holds for every reachable state — a profiled fused request that ran always collects leg profiles, so `mergedProfileResults` is non-null — but it's "unreachable in practice" rather than impossible by construction (the caller also passes null when a profiled merger is `isEmpty()`). Behavior is fine either way (the round-2 section is preserved, just un-relabelled); just flagging the invariant is a touch softer than the comment states.

vibrantvarun (commented):
Thanks for taking the #1977 follow-up. Traced the whole path (leg timed_out → collectLegTimings → FusedCoordinatorTimings.anyLegTimedOut → legTimeoutConsumer → FusedLegTimeoutMerger → filter OR → FusedResponseRebuilder). It's correct: OR-only and never clears, published after all legs are counted and before fusion, per-hybrid unioned into one request-scoped merger, gated by mayHitASoftTimeout so a no-timeout request pays nothing, and the rebuild no-ops unless something actually changed. Nice bonus that extracting FusedResponseRebuilder (widest constructors + canary test) also closes the response-rebuild fragility I raised on #1977 — one argument list to keep correct instead of one per merger. No correctness issues; two notes inline, both non-blocking.

src/main/java/org/opensearch/neuralsearch/search/FusedResponseRebuilder.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/FusedLegTimeoutMerger.java
martin-gaievski: agree, will push the change

martin-gaievski:
> Thanks for taking the #1977 follow-up. Traced the whole path (leg timed_out → collectLegTimings → FusedCoordinatorTimings.anyLegTimedOut → legTimeoutConsumer → FusedLegTimeoutMerger → filter OR → FusedResponseRebuilder). It's correct: OR-only and never clears, published after all legs are counted and before fusion, per-hybrid unioned into one request-scoped merger, gated by mayHitASoftTimeout so a no-timeout request pays nothing, and the rebuild no-ops unless something actually changed. Nice bonus that extracting FusedResponseRebuilder (widest constructors + canary test) also closes the response-rebuild fragility I raised on #1977 — one argument list to keep correct instead of one per merger. No correctness issues; two notes inline, both non-blocking.

Thanks — and both notes are worth more than "non-blocking" suggests, because between them they caught five doc claims in this PR's own new code that are wrong or overstated, three of which neither note actually names. Fixing all five.

vibrantvarun (approved):
LGTM on ef8cd2e24f. Both notes resolved via accurate doc corrections (behavior was already right). The shard-failure case in particular is better than my note assumed — it's not silent: HybridFusionOrchestrator emits a Warning header naming the degraded legs, so the surviving distinction is header-vs-body-field, and the _shards triple genuinely has no defensible OR; terminated_early is unreachable since terminate_after is rejected at rewrite. The ownProfileResults invariant now states the softer truth (a request the ActionFilter saw as unprofiled can be profiled by response time). Core change is sound: OR-only timed_out published after all legs are counted and before fusion, per-hybrid unioned into one request-scoped merger, gated by mayHitASoftTimeout, single FusedResponseRebuilder pass with widest constructors + canary test. Thanks for the thorough follow-up.

PR #1981, opened 2026-09-02
Deduplicate ingest inference inputs within sub-batches

linjiaye929:
### Description
Deduplicates equal text inference inputs within a single ingest sub-batch, while preserving the existing document and field order. Each unique input is inferred once and its result is scattered to every original position with independent mutable result objects.

This applies only to the dense `text_embedding` and sparse `sparse_encoding` ingest batching paths. Sparse WORD and TOKEN_ID requests remain separate inference boundaries. Existing `skip_existing` filtering and batch error propagation are unchanged.

This does not add a search cache, in-flight search coalescing, cross-request cache, node-local TTL cache, or cluster-wide cache.

### Related Issues
Refs #1558

### Test evidence
- `gradle.bat test -PskipSpotlessEclipse -x spotlessApply -x buildJniLib -x cmakeJniLib --tests "org.opensearch.neuralsearch.processor.InferenceProcessorTests" --tests "org.opensearch.neuralsearch.processor.TextEmbeddingProcessorTests" --tests "org.opensearch.neuralsearch.processor.SparseEncodingProcessorTests" --no-daemon --max-workers=1` — passed: 112 tests, 0 failures (6 shared, 69 dense, 37 sparse).
- Regression tests failed before the production change with dense input `[tiny, duplicate, duplicate, lengthy input]` instead of `[tiny, duplicate, lengthy input]`, and sparse format inputs `[value1, value1]` instead of `[value1]`.

The standard `spotlessJavaCheck` could not configure locally because the Eclipse JDT formatter P2 mirror timed out; no formatter configuration change is included.

### Check List
- [x] New functionality has been documented in the changelog.
- [x] API changes companion pull request is not applicable; there is no API change.
- [x] Public documentation issue/PR is not applicable; this changes no public API or configuration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Add `model_selection` (language_option/model_type) parameter to semantic field to resolve the model id from cluster settings ([#1918](https://github.com/opensearch-project/neural-search/issues/1918))
 
 ### Bug Fixes
+* [Ingest] Deduplicate equal inference inputs within each ingest sub-batch and scatter the result to every original field position ([#1558](https://github.com/opensearch-project/neural-search/issues/1558))
 * [Hybrid Query] Fix NoSuchElementException in hybrid query with sort/search_after when a shard returns no results ([#1939](https://github.com/opensearch-project/neural-search/pull/1939))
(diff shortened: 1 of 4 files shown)


heemin32:
@linjiaye929 Thanks for the contribution! However, this issue is about the query side rather than ingestion — specifically, users querying the same value against multiple fields. On the ingestion side, I don't think there's a strong needs where users would want to store the same value in multiple fields. And if they did, it could be handled by a [copy processor](https://docs.opensearch.org/latest/ingest-pipelines/processors/copy/) without generating the embedding multiple times.

PR #1985, opened 2026-09-02
[Fusion] Search the fusion config parameters

martin-gaievski:
### Description
Improved handling for search configs in fusion hybrid query:
- read and apply rank_constant for RRF technique
- block lower and upper bounds for min_max, currently they are silently ignored. Adding proper support for those techniques should be in a separate PR
- default is min_max+arithmetic_mean, same as in existing HQ: `fusion: {}` and `fusion: {"window_size": N}` are the defaults form (min_max + arithmetic_mean, equal weights), not a pipeline read

### Related Issues
#1930 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/query/FusionSpec.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/FusionSpec.java
@@ -8,6 +8,8 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
+import java.util.TreeSet;
 
 import lombok.AccessLevel;
 import lombok.Getter;
@@ -41,7 +43,8 @@ import org.opensearch.neuralsearch.processor.normalization.RRFScoreNormalizer;
  *       combination clause itself, NOT under {@code parameters} — that is where {@code RRFProcessorFactory} reads it) +
  *       optional {@code combination.parameters.weights}. RRF is rank-based, so this shape carries no normalization
  *       clause; it still resolves to {@code normalization = rrf}, because that is the normalization classic applies for
- *       it — see {@link #readNormalizationTechnique}.</li>
+ *       it — see {@link #readNormalizationTechnique}. An inline block may also spell {@code rank_constant} the other
+ *       shape's way, under {@code normalization.parameters} — see {@link #resolveScoreRankerRankConstant}.</li>
  * </ul>
  *
  * <p>Note that {@code rrf} appears in both, and means different things: under a {@code normalization-processor} it is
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/neuralsearch/query/FusionSpec.java
```diff
@@ -269,13 +374,27 @@ private static String readNormalizationTechnique(Map<String, Object> config, Str
...
     @SuppressWarnings("unchecked")
     private static float[] readWeights(Map<String, Object> combinationClause) {
         if ((combinationClause.get(PARAMETERS_KEY) instanceof Map) == false) {
             return new float[0];
         }
         Map<String, Object> parameters = (Map<String, Object>) combinationClause.get(PARAMETERS_KEY);
         if ((parameters.get(WEIGHTS_KEY) instanceof List) == false) {
+            if (parameters.containsKey(WEIGHTS_KEY)) {
+                throw new IllegalArgumentException(
+                    String.format(Locale.ROOT, "parameter [%s] must be a collection of numbers", WEIGHTS_KEY)
```
vibrantvarun: This branch gives the clean 400 for a `weights` value that isn't a list — but the element loop just below (`weights[i] = ((Number) raw.get(i)).floatValue()`) casts each element without validating it, so a list with a non-numeric or null element throws an unchecked exception instead:

- `weights: ["a", 0.5]` → `ClassCastException`
- `weights: [null]` → `NullPointerException`

Both surface as HTTP 500, which is exactly the silent-vs-classic mismatch this method's javadoc says it exists to prevent — `testFromInlineFusion_whenWeightsIsNotAList_thenRejected` only covers the whole value not being a list, so the bad-element case is uncovered. Classic answers it with this same `must be a collection of numbers` 400.

Suggest validating per element before the cast:
```java
Object w = raw.get(i);
if (!(w instanceof Number)) {
    throw new IllegalArgumentException(String.format(Locale.ROOT, "parameter [%s] must be a collection of numbers", WEIGHTS_KEY));
}
weights[i] = ((Number) w).floatValue();
```
plus a test with `weights: ["a"]` asserting the 400.

vibrantvarun (commented):
Intense pass over the fusion-config parsing. It's solid and I verified the larger picture: rank_constant is genuinely applied end-to-end (FusionSpec reads it from the correct clause per shape → HybridFusionOrchestrator.normalizerParams threads fusion.rankConstant() into ScalarNormalizers.forTechnique), so the normalization-processor+rrf pipeline that was stuck at the default 60 is actually fixed. Validation is consistent with the 'reject what classic rejects rather than diverge at HTTP 200' philosophy (min_max bounds → 400, rank_constant dual-spelling + conflict detection, object-clause guards double-defended at parse and rewrite), and the Shape-keyed compatibility gate correctly rejects the normalization-processor rrf+mean pairing while exempting score-ranker rrf+rrf. One correctness gap inline; otherwise LGTM.

src/main/java/org/opensearch/neuralsearch/query/FusionSpec.java
martin-gaievski: ack, adding validation and tests

vibrantvarun (approved):
LGTM on 638a4d64c2. The weights element-validation gap is fixed exactly as discussed — a non-Number (or null) element now returns the same 'must be a collection of numbers' 400 as the not-a-list case, covered by both a unit test and an IT. Rest of the intense review stands: rank_constant is applied end-to-end (FusionSpec parse fix → normalizerParams → ScalarNormalizers), validation consistently rejects what classic rejects (min_max bounds, rank_constant dual-spelling/conflict, object-clause guards), and the Shape-keyed compatibility gate handles the rrf pairings correctly. Thanks for the quick fix.

PR #1986, opened 2026-09-03
[Fusion] Count a kNN/neural leg's whole match set in the fused Tail

martin-gaievski:
### Description
Fixed total hits count for neural sparse and queries with internal count like k in knn query

Example of scenario when issue is visible:

```
PUT /docs
{"settings":{"index":{"knn":true,"number_of_shards":1,"number_of_replicas":0}},
 "mappings":{"properties":{
   "owner":{"type":"keyword"},
   "s":{"type":"integer"},
   "vec":{"type":"knn_vector","dimension":2,
          "method":{"name":"hnsw","space_type":"l2","engine":"lucene"}}}}}

POST /docs/_doc/1  {"owner":"owner-a","s":99,"vec":[1.1,1.0]}
POST /docs/_doc/2  {"owner":"owner-a","s":98,"vec":[1.2,1.0]}
POST /docs/_doc/3  {"owner":"owner-a","s":97,"vec":[1.3,1.0]}
```

```
GET /docs/_search?size=20
{
  "query": {
    "hybrid": {
      "fusion": {
        "window_size": 2,
        "normalization": {"technique": "min_max"},
        "combination":   {"technique": "arithmetic_mean"}
      },
      "queries": [
        {"knn":  {"vec": {"vector": [1.1, 1.0], "k": 3}}},
        {"term": {"s": 99}}
      ]
    }
  },
  "aggregations": {"by_owner": {"terms": {"field": "owner", "size": 10}}},
  "track_total_hits": true
}
```

Knn leg matches all three documents; `term` leg matches document 1 alone. So the match set is
three documents, and that is what classic hybrid counts.

| | before | after |
|---|---|---|
| `hits.total.value` | **2** | **3** |
| `by_owner` → `owner-a` `doc_count` | **2** | **3** |

Before, knn leg was materialized as an address of the two hits it returned (documents 1 and 2), so document 3 that matched, and never returned, was absent from the count and from the bucket. The same request with `window_size: 10` was already correct before this PR, because the leg then returned
its whole match set. 

Fixing that by splitting  two questions the name check was answering at once:

- Worth materializing stays a type question (`isMaterializableLeg`): only an ANN leg, because only  those re-walk an HNSW graph — and re-run inference — when the Tail re-executes them.
- Correct to materialize becomes a per-response question: a leg that came back with fewer hits than window was not truncated, so the address is exactly its match set. A leg that filled the window is kept as the real query and counted for real.

Runtime cost of such change is running Tail/Round 2 (it's not additional round 2): only a leg that (a) is an ANN leg, (b) filled the window, and (c) is in a request that needs the Tail at all pays a second ANN walk  (aggs, highlight, non-_score sort, collapse.inner_hits, or accurate index-wide total_hits). 
Default `k` (10) is well below default `window_size` (100), so the common case is unchanged. Conservative direction is a false negative, leg whose match set happens to be exactly `window_size` is re-walked although the address would have been exact, never a wrong count.

### Related Issues
#1930

### Check List


vibrantvarun (commented):
Intense pass, no defects found — this closes the materialized-leg undercount I flagged on #1967 (and the deferred note that was in isMaterializableLeg's own javadoc). Verified end-to-end on the example (knn k=3 / window=2 → leg fills the window → kept as the real query → Round-2 Tail counts all 3, matching classic), and the cost/correctness/names split is cleanly separated. Two confirmations inline, both non-blocking.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -593,19 +593,39 @@ static float scoreAboveTail(final float fusedScore) {
...
+     * hits it already returned.
+     *
+     * <p>The Tail is a {@code filter}, so it <i>is</i> the match set {@code total_hits} and every aggregation are computed
+     * from, and an address of the returned hits stands for the real match set only when the leg returned all of it. That is
+     * decidable from the response alone: {@code newLegRequest} caps every leg at {@code size = window_size}, so a leg that
+     * came back with fewer hits than the window was not truncated and the address is exact — while one that filled the
+     * window may have matched documents it never returned, and is therefore kept as the real query and counted properly.
+     */
+    private static QueryBuilder legInTailForm(QueryBuilder leg, SearchHit[] hits, int windowSize) {
+        if (hits.length >= windowSize) {
```
vibrantvarun, line 618: This is the right decision and correctly conservative: `hits.length >= windowSize` keeps the real query whenever the leg *might* be truncated. Two things worth noting for the record, neither blocking:

- A leg whose true match set is **exactly** `windowSize` also takes this branch and re-runs, even though its address would have been exact — the harmless false-positive direction (an extra ANN walk over correctness). Fine as-is.
- This is the first path that places a **real kNN query inside the Tail `filter`** (previously ANN legs were always materialized to ids). Good that `testFusedKnnTail_whenAnnLegMatchedBeyondItsWindow_thenTotalHitsAndAggsStillCountTheMatchSet` exercises it end-to-end with an oracle control run — that's what proves knn-in-filter actually produces the match-set count rather than just compiling.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -704,23 +730,19 @@ private static QueryBuilder materializedLeg(SearchHit[] hits) {
...
-     * hits is exact by construction, since it was not truncated. Raising {@code window_size} to at least {@code k} ×
-     * shards restores an exact count. The fix — using the original leg in the Tail when a materialized leg came back full —
-     * is deferred: it costs a second ANN walk in precisely the case a user chose a deep {@code k} for.
+     * Legs for which materializing is <i>worth</i> it: an ANN leg re-run in the Tail re-walks the HNSW graph (and a
+     * {@code neural} one re-runs inference) purely to recount what the fan-out already retrieved, while a term-defined leg
+     * costs almost nothing to re-run. This is a cost filter only — whether an address of the returned hits is <i>correct</i>
+     * is decided per response by the truncation test in {@link #legInTailForm}, not here.
+     *
+     * <p>Keeping the two separate is what makes the rule safe for a leg whose writeable name does not determine its match
+     * set. {@code neural} is exactly that: against a {@code rank_features} semantic embedding field it rewrites into a
```
vibrantvarun, line 739: This `neural` → `neural_sparse` case is the genuinely important catch, worth calling out: materializing on the writeable name alone silently undercounted a sparse leg's match set at HTTP 200, and the coordinator can't tell dense from sparse pre-rewrite. Making `isMaterializableLeg` a pure cost filter and letting the per-response truncation test decide correctness is the right factoring — a sparse leg fills the window, so it's kept as the real query and counted properly. `testBuildFusedQuery_whenNeuralLegFilledTheWindow_...` and the IT parity run pin it.

vibrantvarun (approved):
LGTM. Intense pass found no defects — this closes the materialized-leg undercount from #1967: an ANN leg that filled the window is now kept as the real query in the Tail and counted for real, while one that came back short is still materialized. The cost (isMaterializableLeg) / correctness (per-response truncation test) / names split is cleanly separated, windowSize is threaded consistently, and the neural→neural_sparse case is handled by deciding on truncation rather than the writeable name. Verified end-to-end on the example and the oracle-anchored IT (truncating vs window-wide control agree). No action items from my two inline comments — they're confirmations. Nice work.

PR #1988, opened 2026-09-03
Guard eclipse() to spotless tasks and keep P2 mirror on pinned version (neural-search)

chishui:
### Description
`eclipse()` resolves at Gradle **configuration** time, so a slow P2 update site kills every task in the repo — including BWC, integ and JNI jobs that format nothing. Neither `-x spotlessCheck` nor `--offline` avoids it:

```
> java.io.IOException: Failed to load eclipse jdt formatter: SocketTimeoutException: timeout
```

spotless 8.10.0 added embedded lockfiles for Eclipse JDT 4.9–4.40 (diffplug/spotless#3000), so a pinned version resolves `org.eclipse.jdt:org.eclipse.jdt.core` from Maven Central through the repositories already in `buildscript`, with Gradle's caching and retry. Pinning **4.29** keeps today's formatter version, so no source files change. `withP2Mirrors` no longer participates in resolution.

Verified with the P2 cache deleted: `spotlessCheck` passes, `~/.m2/repository/dev/equo` is never recreated, and `./gradlew help` succeeds with the network blackholed (fails on `main`).

### Check List

By submitting this pull request, I confirm my contribution is made under the Apache 2.0 license.

### DIFF
--- a/.github/workflows/backwards_compatibility_tests_workflow.yml
+++ b/.github/workflows/backwards_compatibility_tests_workflow.yml
@@ -22,6 +22,7 @@ jobs:
   Restart-Upgrade-BWCTests-NeuralSearch:
     needs: Get-CI-Image-Tag
     strategy:
+      fail-fast: false
       matrix:
         java: [21, 25]
         os: [ubuntu-latest]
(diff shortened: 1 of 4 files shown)


peterzhuamazon:
Even if pinning for 4.29 I would still suggest you to put on the mirror. We are testing the mirror and make changes since last week, so please avoid removing it or we cannot test all the component for stress testing. Thanks.

peterzhuamazon:
Offline discussed with Liyun and I will include another fix and put this as part of another batch PRs before merging. Thanks.

peterzhuamazon:
expected version pin

peterzhuamazon (approved):

PR #1992, opened 2026-09-03
Expose neural filters to query visitors

sharp-pixel:
## Description
Expose a neural query's embedded filter to QueryBuilderVisitor traversal as a FILTER child.

## Testing
- ./gradlew test --tests org.opensearch.neuralsearch.query.NeuralQueryBuilderTests --no-daemon


martin-gaievski (commented):
Looks good, just add the changelog entry

martin-gaievski (approved):

martin-gaievski:
@vibrantvarun can you take a look, that's the cod change part from https://github.com/opensearch-project/neural-search/pull/1957, we have split it into two to unblock security plugin change

vibrantvarun (approved):
LGTM. This is the code-only split of #1957's NeuralQueryBuilder#visit override, which I already vetted there — byte-for-byte the same and correct: accept(this) then queryfilter.visit(getChildVisitor(FILTER)), so it recurses the full filter subtree (KNN-aligned, not the boosting/script_score accept-without-recurse pattern), null-guarded. Both unit tests pin it (filter present -> [neural, filter] with Occur.FILTER; absent -> [neural] only). Clean extraction with no cross-plugin dependency. Thanks for splitting it out.

PR #1993, opened 2026-09-03
[Fusion] Gate fused mode behind opt-in cluster setting

martin-gaievski:
### Description
Fused (in-query) mode ships as an experimental, opt-in feature. This adds
```
plugins.neural_search.hybrid.fusion.enabled   (boolean, default false, NodeScope + Dynamic)
```
by default this refuses a `hybrid` query carrying a `fusion` block:

```json
POST /my-index/_search
{"query":{"hybrid":{"fusion":{"normalization":{"technique":"min_max"}},
                    "queries":[{"match":{"text":"hello"}},{"term":{"text":{"value":"place"}}}]}}}
```
response
```
HTTP 400
[hybrid] query [fusion] is a feature that is turned off on this cluster.
Set [plugins.neural_search.hybrid.fusion.enabled] to true to enable it, or drop the [fusion] block and use classic hybrid
```

There are two routes to turn it on, and both properties on the setting are load-bearing rather than belt-and-braces.

**`Dynamic` — `PUT /_cluster/settings`.** Takes effect on the next request, no restart, and it is the only route on a
cluster whose node configuration cannot be edited in place:

```json
PUT /_cluster/settings
{"persistent":{"plugins.neural_search.hybrid.fusion.enabled": true}}
```

Use `persistent`, not `transient`: a transient value silently overrides the persistent one while it exists and is then
lost on a full-cluster restart, which would switch an experimental feature off underneath a workload already using it.

**`NodeScope` — `opensearch.yml`.** Declarative, at home in whatever configures the nodes, carried by a replacement node with no API call, and the only way to bring a cluster up already on:

```yaml
plugins.neural_search.hybrid.fusion.enabled: true
```

The two are not equal when both are used. A cluster-state value wins over the node's own `opensearch.yml`. Either way the transition is logged once at WARN by `NeuralSearchSettingsAccessor`. 

### Related Issues
#1930

### Check List


src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettingsAccessor.java
```diff
@@ -54,5 +57,18 @@ private void registerSettingsCallbacks(ClusterService clusterService, Settings s
                 int maxThreadQty = OpenSearchExecutors.allocatedProcessors(settings);
                 ClusterTrainingExecutor.updateThreadPoolSize(maxThreadQty, setting);
             });
+        clusterService.getClusterSettings().addSettingsUpdateConsumer(HYBRID_FUSION_ENABLED, value -> {
```
vibrantvarun, line 60: Minor / doc-vs-behavior: `addSettingsUpdateConsumer` fires only on a *transition*, so a cluster brought up already on via `opensearch.yml` (the NodeScope route) never triggers this and logs nothing at startup — whereas the PR description says "Either way the transition is logged once at WARN." Accurate for the dynamic `PUT /_cluster/settings` route; the yml-at-boot case is silent (arguably fine — there's no transition to log — but the "either way" wording overstates it). Non-blocking; either tweak the wording or, if a startup signal is wanted for the yml route, log the resolved value once at init.

src/main/java/org/opensearch/neuralsearch/settings/NeuralSearchSettings.java
```diff
@@ -143,4 +143,33 @@ public final class NeuralSearchSettings {
...
+     *
+     * <p>The two routes are not equal when both are used. A cluster-state value wins over the node's own
+     * {@code opensearch.yml} — {@code AbstractScopedSettings} resolves the applied cluster settings first and falls back
+     * to the node's settings — so once the key has been set through the API the yml line no longer decides anything.
+     *
+     * <p>Set it under {@code persistent}, not {@code transient}: a transient value silently overrides the persistent one
+     * while it exists and is then lost on a full-cluster restart, which would turn fused mode off under a workload already
+     * relying on it.
+     */
+    public static final Setting<Boolean> HYBRID_FUSION_ENABLED = Setting.boolSetting(
```
vibrantvarun, line 169: Minor / test coverage: every fused-mode IT enables the feature via the dynamic route (`enableFusedMode()` → `updateClusterSettings`), and the read semantics for the NodeScope route are unit-covered through the mock `ClusterSettings(nodeSettings)` construction — so the yml route is effectively proven at unit level. There's no IT that actually brings a node up with `plugins.neural_search.hybrid.fusion.enabled: true` in `opensearch.yml` and asserts a fusion query is served. Non-blocking, but a single such IT would close the NodeScope route end-to-end (it's the one you can't set dynamically-only).

vibrantvarun (approved):
LGTM. Clean experimental opt-in: default-off and fail-closed (no cluster service → getDefault() → refuse), the gate is only entered for a fusion-carrying query, on the coordinator in round 1, and deliberately ahead of requireClusterSupportsFusedMode so a lagging+off cluster is told to enable rather than upgrade. Refuse-not-downgrade is the right call (a body-carried fusion config has no classic equivalent without a pipeline). I verified the concern that mattered — that clusterSettings.get() reflects an opensearch.yml (NodeScope) value and not only dynamic updates — via NeuralSearchClusterTestUtils constructing ClusterSettings from node settings and the gate reading true; so the yml route is honored and the cluster-state-wins-over-yml precedence is standard. Two minor, non-blocking notes inline. Nice work.

bzhangam (approved):

martin-gaievski:
had to add a fix for BWC and not check for new fusion specific feature flag/setting in older versions

PR #1996, opened 2026-09-07
[Sparse ANN] Report sparse vector field adoption as neural info stats

chishui:
### Description

Adds four info stats under `index.sparse`: `sparse_vector_indices`, `sparse_vector_fields`, `sparse_native_engine_indices`, `sparse_native_engine_fields`.

Derived from cluster-state mappings on each stats call, not event counters. No hook fires once per user-created index — a mapping is parsed on the cluster manager, on every node applying the state, on each update, and again on restart — so an event counter over-counts and drifts. Lucene counts are total minus native; an omitted `engine` counts as lucene, matching `SparseMethodContext#parse`. The walk recurses into `properties`, so nested fields count. Full rationale in the commit message.

### Testing

Unit tests cover per-index counting, nested fields, engine defaults, and mappings without properties. Verified on a local cluster: counts follow index create/delete.

### Related Issues

Related to #1802

### Check List
- [x] Commits are signed per the DCO.

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
+++ b/src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
@@ -280,18 +280,17 @@ public class InfoStatsManager {
         }
 
         for (Map<String, Object> mapping : indexMappings) {
-            SparseFieldCounts counts = new SparseFieldCounts();
-            countSparseVectorFields(asMap(mapping.get(PROPERTIES_KEY)), counts);
+            SparseFieldCounts counts = countSparseVectorFields(asMap(mapping.get(PROPERTIES_KEY)));
 
-            if (counts.total == 0) {
+            if (counts.total() == 0) {
                 continue;
             }
             increment(stats, InfoStatName.SPARSE_VECTOR_INDICES);
-            incrementBy(stats, InfoStatName.SPARSE_VECTOR_FIELDS, counts.total);
+            incrementBy(stats, InfoStatName.SPARSE_VECTOR_FIELDS, counts.total());
 
-            if (counts.nativeEngine > 0) {
+            if (counts.nativeEngine() > 0) {
                 increment(stats, InfoStatName.SPARSE_NATIVE_ENGINE_INDICES);
-                incrementBy(stats, InfoStatName.SPARSE_NATIVE_ENGINE_FIELDS, counts.nativeEngine);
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
```diff
@@ -256,6 +265,84 @@ private void addIngestProcessorStats(Map<InfoStatName, CountableInfoStatSnapshot
...
+            if (fieldConfig == null) {
+                continue;
+            }
+            if (SparseVectorFieldMapper.CONTENT_TYPE.equals(asString(fieldConfig.get(TYPE_KEY)))) {
+                counts.total++;
+                if (isNativeEngine(fieldConfig)) {
+                    counts.nativeEngine++;
+                }
+            }
+            countSparseVectorFields(asMap(fieldConfig.get(PROPERTIES_KEY)), counts);
```
yuye-aws: Is there a depth on recursion?

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
```diff
@@ -256,6 +265,84 @@ private void addIngestProcessorStats(Map<InfoStatName, CountableInfoStatSnapshot
...
+        return SparseEngine.NATIVE == SparseEngine.fromName(engine);
+    }
+
+    /**
+     * The sparse vector fields found in one index's mapping
+     */
+    private static class SparseFieldCounts {
+        private long total;
+        private long nativeEngine;
+    }
```
yuye-aws, line 350: Can these 2 fiels be final?

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
chishui: Bounded by the mapper, not by anything here: `index.mapping.depth.limit` (20 by default) is enforced before a mapping reaches cluster state, so the walk is at most ~20 frames deep. Noted that on the method in ad8aa992.

src/main/java/org/opensearch/neuralsearch/stats/info/InfoStatsManager.java
chishui: They were accumulators mutated during the walk. Fixed in ad8aa992: `SparseFieldCounts` is now a record and the walk returns counts per level instead of mutating a holder, so both are final. Behaviour unchanged.

yuye-aws (approved):

PR #1999, opened 2026-09-09
[Fusion] Limit rescore pool to documents ranked by fusion

martin-gaievski:
### Description
Fixed scenario when `rescore` on a fused  hybrid could return documents the hybrid never ranked, and drop documents it did.

By design fused mode's round 2 is a self-erased query:

```
bool {
  should: [ constant_score(ids(window doc i))^fusedScore_i, ... ]   // "Top", the ranked window
  filter: bool{ should: [ ...the original legs... ] }               // "Tail" non-scoring part
}
```

`minimum_should_match` is left null, and core's `Queries.applyMinimumShouldMatch` is a no-op for null, so with a
`FILTER` clause present Lucene compiles this to `ReqOptSumScorer(req = Tail, opt = Top)`: Tail alone decides
matching and Top only adds score. Every document any leg matched therefore matches round 2, at score exactly
`0.0f`, while only the window was ranked (score should be >=`HybridFusionOrchestrator.MIN_RANKED_SCORE`).

Any weighting that lands a ranked document on exactly `0.0f` makes it indistinguishable from a Tail-only doc. And worse, the Tail-only document with a lower doc id pushed higher in the final list of hits.

Proposed fix is to attach custom rescorer that filters out the unwanted hits/docs. Coordinator installs the request's own rescorer chain inside a plugin-owned rescorer. At shard time it records which documents arrived unranked, applies the user's rescorers in order, and then puts the unranked ones in a score band below every ranked one.

```
FusedRescoreScope.install (coordinator, rewrite round 1)
    source.rescores()  =  [ FusedWindowGuardRescorerBuilder{ the user's window-confined chain, in order } ]

FusedWindowGuardRescorer.rescore (shard)
    1. unranked := { doc : score <= 0.0f }        // read from the PRISTINE pool, before any delegate runs
    2. for each delegate: topDocs = delegate.rescore(topDocs, ...)
    3. unranked -> UNRANKED_SCORE ; ranked -> max(score, RANKED_FLOOR)
    4. re-sort (score desc, doc id asc)
```

Rescore may reorder the hybrid's hits, but it may not change which documents are eligible to be returned. That is exactly the ordering the same request has with no rescore — ranked documents at or above the floor, Tail-only documents at `0.0f`. 

Alternative approach I considered is with use of query phase searcher as injection point, but I want to avoid usage of query phase classes of hybrid query for new functionality.

##### Example A — `query_weight: 0` at default windows. 

Index: **300 documents**, one integer field `s` with `s = _id`, **20 shards**. Two identical legs scoring by `s`, so the default `fusion.window_size` of 100 makes the ranked window documents 201–300, highest doc ids, while never-ranked documents 1–200 are lowest.

```json
POST /idx/_search
{
  "size": 6,
  "query": {
    "hybrid": {
      "fusion": {
        "combination": { "technique": "arithmetic_mean", "parameters": { "weights": [0.5, 0.5] } }
      },
      "queries": [
        { "function_score": { "query": { "match_all": {} },
                              "field_value_factor": { "field": "s", "modifier": "none", "missing": 1 } } },
        { "function_score": { "query": { "match_all": {} },
                              "field_value_factor": { "field": "s", "modifier": "none", "missing": 1 } } }
      ]
    }
  },
  "rescore": {
    "query": {
      "rescore_query": { "ids": { "values": ["300"] } },
      "query_weight": 0.0,
      "rescore_query_weight": 10.0,
      "score_mode": "total"
    }
  }
}
```
Before this change

```
ids    = [300,  5,   11,  202, 206, 227]
scores = [10.0, 0.0, 0.0, 0.0, 0.0, 0.0]
```

Documents `5` and `11` are outside the fused window: two of six hits are documents the fusion never ranked, and two
it did rank were pushed off the page to make room.

With/after this PR

```
ids    = [300,  202, 206, 227, 231, 240]
scores = [10.0, 0.0, 0.0, 0.0, 0.0, 0.0]
```

Every hit is inside the window. The scores are unchanged, only the *membership* is fixed.

##### Example B — `query_weight: -1.0`, whole page was unranked.

Same index and query as A, with `"query_weight": -1.0` and a `rescore_query` matching nothing
(`{"ids":{"values":["9999"]}}`). Core applies no sign check.

Before this change

```
ids    = [5,    11,   35,   73,   93,   110 ]
scores = [-0.0, -0.0, -0.0, -0.0, -0.0, -0.0]
```

Every hit is a document the fusion never ranked. 

With/after this PR

```
ids    = [201,    202,       203,       204,        205,       206      ]
scores = [-0.001, -0.010101, -0.020202, -0.0303030, -0.040404, -0.050505]
```

All ranked. The ordering is inverted, which is literally what a negative `query_weight` asks for, but nothing unranked
appears.

##### Example C — `score_mode: multiply` with `rescore_query_weight: 0`, on a normally-scored ranked document

This one shows the defect is not an artefact of the ranked floor. Index: 30 documents, 1 shard. Three legs at weights
`[0.5, 0.5, 0.0]` — a range leg matching documents 1–4, an `ids` leg matching 26 and 27, and a `match_all` leg at weight
zero so the Tail covers everything. `fusion.window_size: 6`.

Baseline, no rescore, document 26 is ranked at an ordinary fused score of `0.5`, with a high doc id:

```
ids    = [1,   26,  2,          3,          4,      27    ]
scores = [0.5, 0.5, 0.33333334, 0.16666667, 5.0E-4, 5.0E-4]
```

Add:

```json
"rescore": { "window_size": 20,
  "query": { "rescore_query": { "ids": { "values": ["26"] } },
             "query_weight": 1.0, "rescore_query_weight": 0.0, "score_mode": "multiply" } }
```

Before this change `[1, 2, 3, 4, 27, 5]` — ranked document 26 is gone, never-ranked document 5 took its slot. `score_mode: min` returned exactly the same page.

After this change `[1, 2, 3, 4, 27, 26]` — document 26 keeps its slot, document 5 is absent.

##### Example D, full `rescore_query_weight` × `score_mode` matrix, fused vs plain

`rescore_query_weight ∈ {1.0, 0.0, −5.0}` plus a both-negative row, × `score_mode ∈ {total, avg, max, min, multiply}`, on the 300-document / 20-shard index. With this fix, none of the 20 fused combinations returned a document outside the window. Plain-query arm is included to show the behaviour is consistent with the rest of the engine:

| weighting | fused (after) | plain `function_score` | reading |
|---|---|---|---|
| `rqw=-5.0, qw=1.0` | `[300,299,…]` | `[300,299,…]` | identical |
| `rqw=0.0, qw=1.0` | `[300,299,…]` | `[300,299,…]` | identical — under `multiply` the matched document collapses and drops off the page; nothing unranked appears |
| `rqw=-5.0, qw=-1.0` | `[201…206]` — all ranked, inverted | `[25,52,58,63,67,68]` | both invert; fused's inversion is confined to the ranked set |
| `multiply, rqw=-5.0, qw=-1.0` | `[250,201,202,…]` | `[250,25,52,…]` | same shape — a double negative makes the matched document win |

I've added some of these scenario to test suite.

### Related Issues
#1930 

### Check List

### DIFF
--- a/src/main/java/org/opensearch/neuralsearch/query/FusedWindowGuardRescorerBuilder.java
+++ b/src/main/java/org/opensearch/neuralsearch/query/FusedWindowGuardRescorerBuilder.java
@@ -1,0 +1,181 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.opensearch.neuralsearch.query;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+import org.opensearch.core.common.io.stream.StreamInput;
+import org.opensearch.core.common.io.stream.StreamOutput;
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.index.query.QueryRewriteContext;
+import org.opensearch.index.query.QueryShardContext;
+import org.opensearch.search.rescore.QueryRescorerBuilder;
+import org.opensearch.search.rescore.RescoreContext;
+import org.opensearch.search.rescore.RescorerBuilder;
+
+/**
+ * Carries the request's own rescorer chain to the shard wrapped in {@link FusedWindowGuardRescorer}, so that the
+ * guarantee "a rescore may reorder the hybrid's hits but may not change which documents are eligible" survives any
+ * weights the request used. See {@link FusedWindowGuardRescorer} for why the guard is needed and why it demotes rather
+ * than removes.
+ *
+ * <p><b>Not user-typeable, deliberately.</b> This is registered as a named WRITEABLE only — via
+ * {@code NeuralSearch#getNamedWriteables()} under the {@link RescorerBuilder} category — and NOT through
+ * {@code SearchPlugin#getRescorers()}. That matters: {@code SearchModule#registerRescorer} adds a
+ * {@code NamedXContentRegistry} entry in the same call as the writeable one, and the XContent entry is what makes a
+ * rescorer name parseable from a request body. Registering the writeable alone gives the shard what it needs to
+ * deserialize this builder while adding no new request syntax for a user to discover, mistype, or come to depend on.
+ * {@code fromXContent} therefore does not exist; {@link #doXContent} is render-only, for the slow log and
+ * {@code _tasks?detailed}.
+ *
+ * <p><b>One guard wraps the whole chain.</b> The delegates are held in the order the user declared them and are applied
+ * in that order, so per-element {@code window_size} semantics are preserved: each delegate keeps its own
+ * {@link RescoreContext} built from its own builder. This builder's own {@code window_size} is the maximum over the
+ * chain, because core sizes the shard's first-pass pool from the largest window across all rescore contexts and this
+ * builder is the only context core can see.
+ *
+ * <p><b>Why this needs no version gate of its own.</b> It is a new {@code NamedWriteable} on the shard wire, so the
+ * obvious question is what a node that cannot resolve the name does — it fails while deserializing the shard request,
+ * and with {@code allow_partial_search_results} at its default the client still gets a 200 with those shards' documents
+ * silently missing. That is already the exact hazard {@code HybridQueryBuilder#requireClusterSupportsFusedMode} exists
+ * for: it refuses fused mode cluster-wide, at rewrite and before any fan-out, unless every node can resolve
+ * {@code hybrid_fusion}. This builder is only ever installed on a request that has passed that gate, and it ships in the
+ * same release as fused mode itself — which has never shipped, so there is no released build that has fused mode and
+ * lacks this. The gate therefore covers both names with one check.
+ *
+ * <p><b>The dependency that makes that true, and it is not yet satisfied:</b>
+ * {@code MinClusterVersionUtil.MINIMAL_SUPPORTED_VERSION_FUSED_MODE_IN_HYBRID_QUERY} must name the release that actually
+ * contains fused mode. It currently reads {@code V_3_8_0}, which is already released and contains none of it — so the
+ * constant has to be corrected before either name reaches a mixed-version cluster. That correction is tracked
+ * separately; do not add a second gate here to work around it, because a per-name gate would answer a fused rescore
+ * UNGUARDED on a cluster that failed it, which is worse than refusing.
+ *
+ * <p><b>Installed at coordinator rewrite round 1</b> by {@code FusedRescoreScope#install}, which replaces the request's
+ * rescore list with a single element holding the already-window-confined delegates. A replacement made at round 1
+ * reaches the shards: {@code SearchSourceBuilder#rewrite} rewrites the query before it reads the rescorer list, so the
+ * list core snapshots is the one this builder is in.
+ */
+public class FusedWindowGuardRescorerBuilder extends RescorerBuilder<FusedWindowGuardRescorerBuilder> {
+
+    public static final String NAME = "hybrid_fused_window_guard";
+
+    /**
+     * The score unranked documents are demoted to. Finite on purpose: a non-finite score is quoted as a JSON string by
+     * the response serializer, so {@code -Infinity} would turn {@code _score} into a string for every client.
+     *
+     * <p>Public because the coordinator has to recognise it again — {@code FusedRescoreScoreNormalizer} maps it back to
+     * {@code 0.0} before the response leaves, so the sentinel is never user-visible.
+     */
+    public static final float UNRANKED_SCORE = -Float.MAX_VALUE;
+
+    /**
+     * The lowest score a ranked document may report, one float above {@link #UNRANKED_SCORE}. Nothing is representable
+     * between the two, so the ranked and unranked bands cannot meet however extreme the request's weights are.
+     */
+    public static final float RANKED_FLOOR = Math.nextUp(-Float.MAX_VALUE);
+
+    private final List<QueryRescorerBuilder> delegates;
+
+    FusedWindowGuardRescorerBuilder(final List<QueryRescorerBuilder> delegates) {
+        if (Objects.isNull(delegates) || delegates.isEmpty()) {
+            throw new IllegalArgumentException("[" + NAME + "] requires at least one rescorer to guard");
+        }
+        this.delegates = List.copyOf(delegates);
+        // Core sizes the first-pass pool from the largest window across the request's rescore contexts, and after the
+        // wrap this builder is the only context it sees — so it has to stand in for the whole chain's depth.
+        Integer widest = null;
+        for (QueryRescorerBuilder delegate : delegates) {
+            if (Objects.nonNull(delegate.windowSize()) && (Objects.isNull(widest) || delegate.windowSize() > widest)) {
+                widest = delegate.windowSize();
+            }
+        }
+        if (Objects.nonNull(widest)) {
+            windowSize(widest);
+        }
+    }
+
+    public FusedWindowGuardRescorerBuilder(final StreamInput in) throws IOException {
+        super(in);
+        this.delegates = List.copyOf(in.readList(QueryRescorerBuilder::new));
+    }
+
+    @Override
+    public String getWriteableName() {
+        return NAME;
+    }
+
+    @Override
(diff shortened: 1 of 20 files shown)


src/main/java/org/opensearch/neuralsearch/search/FusedRescoreScoreNormalizer.java
```diff
@@ -0,0 +1,69 @@
...
+    static SearchResponse normalize(final SearchResponse response, final float sentinel) {
+        if (Objects.isNull(response) || Objects.isNull(response.getInternalResponse())) {
+            return response;
+        }
+        SearchHits hits = response.getInternalResponse().hits();
+        if (Objects.isNull(hits) || Objects.isNull(hits.getHits())) {
+            return response;
+        }
+        for (SearchHit hit : hits.getHits()) {
+            if (Float.compare(hit.getScore(), sentinel) == 0) {
```
vibrantvarun: Edge artifact (LOW, but worth deciding): this normalizes only the exact `UNRANKED_SCORE` (-Float.MAX_VALUE) back to 0.0, but the guard also produces `RANKED_FLOOR` (= `nextUp(-Float.MAX_VALUE)`) whenever a *ranked* doc's post-rescore score saturated to ≤ -MAX or NaN — which the guard deliberately clamps (see `rankedScore`). That value is never normalized, so such a ranked doc is returned at `_score ≈ -3.4028235E38`, positioned *above* the unranked docs that were normalized to `0.0`. Result: hits are ordered correctly (ranked above unranked) but their `_score` values are non-monotonic (a higher hit shows a lower score), and an internal sentinel-adjacent value leaks to the client. Reachable only with pathological negative weights, but the guard clearly considers that path real (it's why the NaN/saturation clamp exists). Should `RANKED_FLOOR` also be mapped here (e.g. to `MIN_RANKED_SCORE`, keeping ranked strictly above the normalized-unranked 0.0), and is the saturated-ranked case covered by a test? Same question for the `maxScore` rebuild below — a top hit at `RANKED_FLOOR` isn't `== sentinel`, so it isn't corrected either.

src/main/java/org/opensearch/neuralsearch/query/FusedWindowGuardRescorerBuilder.java
```diff
@@ -0,0 +1,181 @@
...
+ * silently missing. That is already the exact hazard {@code HybridQueryBuilder#requireClusterSupportsFusedMode} exists
+ * for: it refuses fused mode cluster-wide, at rewrite and before any fan-out, unless every node can resolve
+ * {@code hybrid_fusion}. This builder is only ever installed on a request that has passed that gate, and it ships in the
+ * same release as fused mode itself — which has never shipped, so there is no released build that has fused mode and
+ * lacks this. The gate therefore covers both names with one check.
+ *
+ * <p><b>The dependency that makes that true, and it is not yet satisfied:</b>
+ * {@code MinClusterVersionUtil.MINIMAL_SUPPORTED_VERSION_FUSED_MODE_IN_HYBRID_QUERY} must name the release that actually
+ * contains fused mode. It currently reads {@code V_3_8_0}, which is already released and contains none of it — so the
+ * constant has to be corrected before either name reaches a mixed-version cluster. That correction is tracked
```
vibrantvarun: Flagging this as a release-gate, not a change request — the javadoc already calls it out and I agree with not adding a second gate here. But to make sure it isn't lost: this builder's wire safety (and `hybrid_fusion`'s) rests entirely on `requireClusterSupportsFusedMode`, which keys off `MINIMAL_SUPPORTED_VERSION_FUSED_MODE_IN_HYBRID_QUERY` — and that constant still reads `V_3_8_0`, an already-released version that contains none of fused mode. Until it's corrected to the release that actually ships fused mode, the version gate is effectively a no-op: a mixed-version cluster could route a `hybrid_fusion`/`hybrid_fused_window_guard` to a node that can't resolve the NamedWriteable, which under default `allow_partial_search_results` is a silent short answer (200 with shards missing). Can you link the tracking item for the constant correction and confirm it lands before fused mode is enabled in any release? It gates the whole feature's mixed-cluster safety, not just this PR.

vibrantvarun (commented):
Intense review — traced the guard end-to-end (membership read from the pristine pool before any delegate; whole-chain wrap so a query_weight:0 element can't reclassify ranked docs; two-band clamp with an unbridgeable gap at ±nextUp(-MAX); demote-not-remove to avoid the RescoreProcessor scoreDocs[0] AIOOBE; per-delegate windows preserved, builder window = max over chain; NamedWriteable-only so it's not user-typeable). Also verified the coordinator side: the sentinel-normalize path fires exactly when the fused hybrid is the request's own query (finder early-returns on empty, and doRewriteFused refuses a rescore unless the hybrid is top-level), so there's no empty-list access and no nested-hybrid leak. This is a solid, well-reasoned fix and the test coverage is strong. Two things inline: one edge correctness artifact, and one release-gating dependency the PR itself flags.

src/main/java/org/opensearch/neuralsearch/query/FusedWindowGuardRescorerBuilder.java
martin-gaievski: ack

src/main/java/org/opensearch/neuralsearch/search/FusedRescoreScoreNormalizer.java
martin-gaievski: ack on both halves, and it's a bit worse than that.

RANKED_FLOOR is never mapped, so saturated ranked doc comes back at -3.4028233E38 above hits reading 0.0, and max_score lands strictly below the minimum _score on the page. The rebuild also hardcodes 0.0f, which is the wrong corrected value for a  saturated ranked top hit.
On coverage: not just untested — pinned. 

And your maxScore question found a channel I'd missed: the raw sentinel leaks into top_hits buckets with ordinary positive weights. Core rescores each bucket's TopDocs and writes the score onto the bucket hit; normalizer only walks top-level hits. My bucket IT can't see it because its size equals the fused window.

Fixing: map the ranked band member above 0.0 (new constant next to UNRANKED_SCORE/RANKED_FLOOR — MIN_RANKED_SCORE is package-private in the wrong package), widen the maxScore branch to "in the bottom band" and recompute from hits, normalize bucket hits, and rewrite that test. The MIN_RANKED_SCORE collision you flag is real but nothing branches on the value, so I don't think it blocks the mapping.

src/main/java/org/opensearch/neuralsearch/query/FusedWindowGuardRescorerBuilder.java
martin-gaievski: opened https://github.com/opensearch-project/neural-search/issues/2002 to track version gate

vibrantvarun (approved):
Re-reviewed 2a8ee10 objectively — verified the fixes are correct, not just present. Both earlier comments are resolved: decode() maps BOTH band floats by one rule (unranked→0.0, RANKED_FLOOR→RANKED_FLOOR_NORMALIZED=1e-30=MIN_RANKED_SCORE), maxScore goes through the same decode (not a hardcoded 0.0), and top_hits buckets are walked/decoded — so ranked-saturated docs report 1e-30 above the unranked 0.0, scores are monotonic, and no sentinel-adjacent value leaks; and the version gate is tracked by #2002 + guarded by testFusedModeMinimumVersion_isNotBehindTheVersionUnderDevelopment (correctly broadened to note classic hybrid also desyncs the wire against a released 3.8.0 via doWriteTo).\n\nAlso checked the fixes you added beyond my comments: ChainRescoreContext forwarding getQueries()/getParsedQueries() is a real catch (without it the wrap would drop _name from matched_queries on rescore queries and give the rescore query local-IDF stats under dfs_query_then_fetch); needsTheRankedFloor clamps NaN/saturated ranked docs even in an all-ranked pool; the global-agg vacuous-inference handling is sound; and the effective-window fix uses core's own DEFAULT_WINDOW_SIZE=10 (verified against RescorerBuilder#buildContext), so a chain with an unset window sizes the pool correctly.\n\nResiduals are documented and acceptable: the normalizeSentinel submit-vs-rewrite coupling (no shipped request processor adds a rescore into a fused hybrid; disclosed with a re-check note), and the top_hits bucket max_score that can't be rebuilt from a plugin. #2002 (the version constant) is the one cross-branch release blocker and is tracked + test-guarded. LGTM for the feature branch.

PR #2000, opened 2026-09-10
[SemanticHighlighter] Support lists and scalars with per-element fragments

### LINKED ISSUE #1813: [FEATURE] Better handle the behavior diff for semantic highlight sing vs batch
We have behavior diff between the semantic highlight using standard Highlight framework vs System generated search processor approach:

Standard Highlighter Framework:

- Cannot support non-string fields
- Fails early when encountering lists or objects - returns null and silently skips highlighting (Need to double confirm)

Batch Inference Processor Approach:

- Can support non-string fields
- Can handle lists and objects successfully (Need to double confirm)

The User Experience Gap:

Currently, when a user tries to highlight a field that's a list:

The standard approach silently returns null and skips highlighting (no feedback to user)
The processor approach can actually highlight the list successfully

Rather than silently skipping, we should inform users when they attempt to highlight a list or object field using the standard approach. Should we try to throw an error(may not be easy to do this since we may fail in core before we get into the neural plugin) to let user know if they want to highlight non string field they can try the processor approach? Or we simply update the documentation to make this clear and let user figure it out by themselves?

_A clear and concise description of any alternative solutions or features you've considered._

ezhang3333:
### Description
Extends semantic highlighting to non-string field types (lists and scalars) with per-element fragments, matching the built-in OpenSearch highlighters.

Previously the batch request builder joined list values into a single string for inference, but the result applier only accepted `String` source values and silently dropped the highlight. The single-inference path likewise returned null for any non-string field.

Behavior after this change:
- Lists yield one `Text` fragment per matching array element (e.g. `["alpha", "beta"]` with a hit on `beta` returns `<em>beta</em>`), like the core highlighters.
- Scalars (`Number`, `Boolean`) are coerced via `toString` on both paths.
- Objects / maps are skipped on both paths.
- Batch keeps a single inference per hit (joined text) and splits model spans back per element by offset; single inference highlights per element.

Files: new `highlight/utils/HighlightValueUtils.java` (shared conversion + per-element split); updated `HighlightContextBuilder`, `HighlightResultApplier`, `HighlightExtractorUtils` (new `getFieldTexts()`), `SemanticHighlighter`, plus a `HighlightTagApplier.parseAndValidateSpans` extraction reused by the new utility; unit tests and CHANGELOG entry included.

### Related Issues
Resolves #1813

### Testing
- Follow-up coverage tests cover batch size checks, no-match snippets, HTML encoding, and element mapping. The changed Java files compile and `git diff --check` passes. A local Gradle rerun was blocked before tests by a Zscaler 403 on an OpenSearch snapshot JAR; CI will provide the test result.
- `./gradlew spotlessJavaCheck`, `./gradlew javadoc`: pass.
- `./gradlew test`: 2802 tests, 0 failures.
- `SemanticHighlightingLocalModelIT`: fails identically on base and branch at model upload in `setUp` (`BaseNeuralSearchIT.uploadModel` returns null - test env cannot provision ML models locally), before any changed code runs. Relying on CI for the integration signal.


src/main/java/org/opensearch/neuralsearch/highlight/utils/HighlightValueUtils.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Joins elements the same way the batch request builder does.
+     *
+     * @param elements non-null text elements
+     * @return space-joined string, empty when no elements
+     */
+    public static String joinElements(final List<String> elements) {
+        if (elements == null || elements.isEmpty()) return "";
+        return String.join(" ", elements);
```
RexyZhang2001: Here, batch inference joins array values with a regular space, which can merge two values into one sentence. For example, `["Apple is red", "My car is outside"]` becomes `Apple is red My car is outside`. The sentence-highlighting model may return one span covering both values, and the split logic then highlights both elements even if only the first one is relevant, like `<em>Apple is red My car is outside</em>`. Can we preserve a hard boundary between values or create one batch request per element?

src/main/java/org/opensearch/neuralsearch/highlight/batch/utils/HighlightResultApplier.java
```diff
@@ -103,45 +103,53 @@ private void applyRow(
...
+        String joinedText = HighlightValueUtils.joinElements(elements);
+        if (joinedText.isEmpty()) return;
 
-        String highlighted = HighlightTagApplier.applyTags(text, highlights, preTag, postTag);
-        if (highlighted == null) {
+        List<String> fragments = HighlightValueUtils.applyHighlightsPerElement(elements, highlights, preTag, postTag);
+        if (fragments.isEmpty()) {
             if (noMatchSize > 0) {
-                String snippet = text.length() <= noMatchSize ? text : text.substring(0, noMatchSize);
+                String snippet = joinedText.length() <= noMatchSize ? joinedText : joinedText.substring(0, noMatchSize);
```
RexyZhang2001: For a list value, `no_match_size` is currently applied to the joined text. For example, `["cat", "dog"]` with `no_match_size: 5` returns `cat d`, even though that text does not exist in any source element. Can we apply the no-match fallback to the first element instead?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
```diff
@@ -98,21 +99,32 @@ public HighlightField highlight(FieldHighlightContext fieldContext) {
...
-
-        if (highlightedResponse == null || highlightedResponse.isEmpty()) {
+        if (fragments.isEmpty()) {
             // no_match_size: emit first N chars when model returns no spans.
             int noMatchSize = extractNoMatchSize(options);
-            if (noMatchSize > 0 && !fieldText.isEmpty()) {
-                String snippet = fieldText.length() <= noMatchSize ? fieldText : fieldText.substring(0, noMatchSize);
-                if (SemanticHighlightingConstants.ENCODER_HTML.equalsIgnoreCase(extractEncoder(options))) {
+            if (noMatchSize > 0 && !joinedText.isEmpty()) {
+                String snippet = joinedText.length() <= noMatchSize ? joinedText : joinedText.substring(0, noMatchSize);
```
RexyZhang2001: The single-inference path has the same `no_match_size` issue, should be updated together.

src/main/java/org/opensearch/neuralsearch/highlight/utils/HighlightValueUtils.java
ezhang3333: Done

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
ezhang3333: Done

src/main/java/org/opensearch/neuralsearch/highlight/batch/utils/HighlightResultApplier.java
ezhang3333: Done

src/main/java/org/opensearch/neuralsearch/highlight/batch/config/HighlightContextBuilder.java
```diff
@@ -140,7 +141,9 @@ private static int resolveMaxBatchSize(HighlightConfig config) {
     /**
      * Reads a string value from a hit's {@code _source} under {@code fieldName}.
      * Falls back to the leaf name when the dotted path is absent, because inner
-     * hit sources key by leaf, not the fully qualified name.
+     * hit sources key by leaf, not the fully qualified name. Lists are joined
+     * with spaces and scalars are coerced via toString, matching the per-element
```
heemin32:
```suggestion
     * with HighlightValueUtils.ELEMENT_SEPARATOR and scalars are coerced via toString, matching the per-element
```

src/test/java/org/opensearch/neuralsearch/highlight/batch/config/HighlightContextBuilderTests.java
```diff
@@ -402,7 +402,7 @@ public void testListSourceValueWithNullsAreSkipped() {
 
         HighlightContext ctx = builder.build(config, response, 0L);
         assertEquals(1, ctx.size());
-        assertEquals("alpha gamma", ctx.getRequests().get(0).getContext());
+        assertEquals("alpha\n\ngamma", ctx.getRequests().get(0).getContext());
```
heemin32: Could you also add a test case where the original value contains `\n\n` in it?

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
```diff
@@ -98,21 +99,32 @@ public HighlightField highlight(FieldHighlightContext fieldContext) {
...
-        );
-
-        if (highlightedResponse == null || highlightedResponse.isEmpty()) {
+        if (fragments.isEmpty()) {
             // no_match_size: emit first N chars when model returns no spans.
             int noMatchSize = extractNoMatchSize(options);
-            if (noMatchSize > 0 && !fieldText.isEmpty()) {
-                String snippet = fieldText.length() <= noMatchSize ? fieldText : fieldText.substring(0, noMatchSize);
-                if (SemanticHighlightingConstants.ENCODER_HTML.equalsIgnoreCase(extractEncoder(options))) {
+            if (noMatchSize > 0 && !firstElement.isEmpty()) {
```
heemin32, line 125: Wouldn't it be better to find the first non empty element similar to unified highlighter?

src/test/java/org/opensearch/neuralsearch/highlight/batch/utils/HighlightResultApplierTests.java
```diff
@@ -284,6 +284,39 @@ public void testApplyWithIndicesSliceSizeMismatchThrows() {
         assertTrue(e.getMessage(), e.getMessage().contains("Batch results size"));
     }
 
+    public void testAppliesListFieldAsPerElementFragments() {
```
heemin32, line 287: Can we add the case where the value contains `\n\n` in it?

src/main/java/org/opensearch/neuralsearch/highlight/utils/HighlightValueUtils.java
RexyZhang2001: Changing the separator to `\n\n` may not fully resolve the boundary issue because `DefaultSentenceSegmenter` does not treat a blank line by itself as a hard sentence boundary. Without terminal punctuation, separate array values can still be interpreted as one sentence.

For example, the values `Apple is red` and `My car is outside` are sent to the model as `Apple is red\n\nMy car is outside`. The first element occupies offsets `[0, 12)`, while the second occupies `[14, 31)`. If the model treats the joined text as one sentence, it can return a span of `[0, 31)`.

When `applyHighlightsPerElement()` processes the first element, it skips this span because it extends beyond offset `12`. When it processes the second element, it skips the same span because it starts before offset `14`. Therefore, neither element receives the model's highlight and batch mode returns no highlighted fragment. In single mode, `Apple is red` is processed independently, so a span of `[0, 12)` can be highlighted correctly.

Could we send each array element as a separate batch input and preserve the element mapping when applying the results? It would also be helpful to add a regression test using adjacent array values that do not end with punctuation.

src/main/java/org/opensearch/neuralsearch/highlight/SemanticHighlighter.java
ezhang3333: Dont think so because the `fieldTexts` we get from `HighlightExtractorUtils` already has had empty strings removed because it ends up calling `HighlightValueUtils.toTextElements()`

src/main/java/org/opensearch/neuralsearch/highlight/batch/config/HighlightContextBuilder.java
ezhang3333: Done

src/test/java/org/opensearch/neuralsearch/highlight/batch/utils/HighlightResultApplierTests.java
ezhang3333: Done

RexyZhang2001:
> Changing the separator to `\n\n` may not fully resolve the boundary issue because `DefaultSentenceSegmenter` does not treat a blank line by itself as a hard sentence boundary. Without terminal punctuation, separate array values can still be interpreted as one sentence.
> 
> For example, the values `Apple is red` and `My car is outside` are sent to the model as `Apple is red\n\nMy car is outside`. The first element occupies offsets `[0, 12)`, while the second occupies `[14, 31)`. If the model treats the joined text as one sentence, it can return a span of `[0, 31)`.
> 
> When `applyHighlightsPerElement()` processes the first element, it skips this span because it extends beyond offset `12`. When it processes the second element, it skips the same span because it starts before offset `14`. Therefore, neither element receives the model's highlight and batch mode returns no highlighted fragment. In single mode, `Apple is red` is processed independently, so a span of `[0, 12)` can be highlighted correctly.
> 
> Could we send each array element as a separate batch input and preserve the element mapping when applying the results? It would also be helpful to add a regression test using adjacent array values that do not end with punctuation.

@ezhang3333 Please address this comment as well and sign off in the next commit. Thanks.

RexyZhang2001:
LGTM! Please sign off to pass the DCO check: https://github.com/opensearch-project/neural-search/pull/2000/checks?check_run_id=104600902254

RexyZhang2001 (approved):

heemin32:
@ezhang3333 please fix the DCO and conflict.

RexyZhang2001:
@ezhang3333 Please improve the test coverage: 
https://app.codecov.io/gh/opensearch-project/neural-search/pull/2000?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=opensearch-project

heemin32 (approved):

PR #2001, opened 2026-09-10
[Sparse ANN] [Native] Add support for neural-sparse-cpp's wider 64-bit nnz-offset

zirui-song-18:
### Description
neural-sparse-cpp [#54](https://github.com/opensearch-project/neural-sparse-cpp/pull/54) widened the CSR forward-index offset from 32-bit to a new `offset_t = int64_t` so a single segment can hold more than ~2.1B non-zeros. The 32-bit cap was hit by large corpora — e.g. MS MARCO V2 (138M docs, ~28.7B nnz) overflows a signed-int32 `indptr` and cannot be built as a single force-merged segment on the native engine.

This PR bumps the submodule to the merged nsparse commit and adapts the plugin's JNI + Java side so the wider offset flows through end to end. Doc-ids, labels, term-ids, and the per-block on-disk structures deliberately **stay 32-bit** (nsparse `idx_t`) — only the CSR nnz-offset widens.

#### Changes
- **Submodule bump** — `jni/external/neural-sparse-cpp` `354c756` → `da0d43c` ("Widen the CSR offset to 64 bit", opensearch-project/neural-sparse-cpp#54, now merged to nsparse `main`).
- **Streaming path** (`jni/src/common.h`) — the off-heap CSR `indptr` accumulator widens from `std::vector<int32_t>` to `std::vector<int64_t>`, so a segment's cumulative nnz can exceed `INT32_MAX`. Each per-flush indptr is still a **relative** int32 array (reset to 0 per flush, bounded by the streaming flush limit) and is widened before the cumulative offset is added — so the `transferVectors`/`insertToIndex` JNI signatures are unchanged.
- **Insert / query** (`jni/src/nsparse_wrapper.cpp`) — hands `add_with_ids` and `search` the `indptr` as `nsparse::offset_t` (was `idx_t`) on both the insert and the single-query paths.
- **Disk `.csr` writer** (`CsrSparseVectorsFile.java`) — emits int64 `indptr` entries (8 bytes/row, matching nsparse `read_csr`); the id file's `external_ids` stay int32 (`idx_t`). Removes the now-obsolete `INT_MAX` nnz cap. `CsrSparseVectorsFileTests` updated for the 8-byte indptr (reads + byte-offset math); all 11 tests pass.

#### Cost (measured on MS MARCO V1, 8.8M docs, native `seismic_sq`, 1 segment)                                              
Widening the offset is effectively free — only the `indptr` array grows 4→8 bytes/row:                                                                                                                         
  | metric | int32 | int64 | Δ |                                                                                             
  |---|---|---|---|                                            
  | disk size (1 segment) | 41.361 GB | 41.406 GB | **+0.11%** (~45 MB) |                          
  | peak JVM RSS | 105,617 MB | 105,616 MB | ~0 |
  | build time | 2227 s | 2223 s | ~0 |                        

### Related Issues

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * [Hybrid Query] Read the current document and its sub-query matches from the positioned disjunction iterator, fixing an ArrayIndexOutOfBoundsException and silently misattributed scores when a sub-query has a two-phase iterator ([#1946](https://github.com/opensearch-project/neural-search/issues/1946))
 * [RRF] Reject a combination technique other than rrf when creating a score-ranker-processor, instead of accepting the pipeline and throwing NullPointerException on every query ([#1949](https://github.com/opensearch-project/neural-search/pull/1949))
 * [Sparse ANN] Skip cache cleanup for a closed index's shards, which have no MapperService, so reopening a sparse index no longer leaks the shard lock and leaves the shard unassigned ([#1982](https://github.com/opensearch-project/neural-search/issues/1982))
+* [Sparse ANN] Widen the CSR indptr offset to 64-bit across the JNI/Java boundary (matching neural-sparse-cpp `offset_t=int64`) so a single segment can hold more than 2.15B cumulative non-zeros without int32 overflow ([#2001](https://github.com/opensearch-project/neural-search/pull/2001))
 
 ### Infrastructure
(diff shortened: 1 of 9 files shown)


chishui (commented):
you should update DefaultNativeIndexWriter as well, although it's not used

zirui-song-18:
> you should update DefaultNativeIndexWriter as well, although it's not used

Ack. The streaming path's indptr lives in `OffHeapSparseVectorsBuffer` (`DefaultNativeIndexWriter` only orchestrates it), so the guard goes there.

chishui (approved):

PR #2004, opened 2026-09-12
Fused adaptive tail totals

martin-gaievski:
### Description
Latest resolver related changes (mainly #1999) in hybrid query introduced extra latency, and made resolver query ~1.8-2x of classic hybrid query

```
  ┌────────┬─────────┬──────────────────┬───────────────────┐
  │ metric │ classic │ resolver (fused) │ Δ (fused−classic) │
  ├────────┼─────────┼──────────────────┼───────────────────┤
  │ p50    │     111 │              216 │           +105 ms │
  ├────────┼─────────┼──────────────────┼───────────────────┤
  │ p90    │     132 │              255 │           +123 ms │
  ├────────┼─────────┼──────────────────┼───────────────────┤
  │ p99    │     616 │             1037 │           +421 ms │
  ├────────┼─────────┼──────────────────┼───────────────────┤
  │ mean   │     131 │              252 │           +122 ms │
  └────────┴─────────┴──────────────────┴───────────────────┘
```

Numbers taken on 3 nodes cluster, 12 shards, Quora dataset from https://github.com/beir-cellar/beir

For a plain top-k fused hybrid, round 2 carried the Tail (a non-scoring filter that re-runs every leg on every shard, including a second HNSW walk) purely to count hits.total — which core caps at the track_total_hits threshold anyway, so the  answer is {10000, gte} whenever the union exceeds it.

This PR chnages this logic - it has the legs count up to that threshold themselves in round 1, and when any leg's own count is capped (gte) — a leg being a subset of the union, so the union is past the threshold too — round 2 runs Top-only and the response gets the identical {threshold, gte} via the ActionFilter seam; when no leg proves it, or any feature needs the Tail's documents (aggregations, highlight, sort, collapse, search_after, min_score, a page past the window, exact totals, named ANN legs, response processors), the Tail is kept exactly as before.

Fix this fix resolver came back to be on par with classic.hybrid query
```
  ┌────────────────────────────┬─────────┬───────────┬───────────┬───────────┬────────────┬──────────┬──────────────┐
  │           build            │ cls p50 │ fused p50 │ fused p90 │ fused p99 │ fused mean │ paired Δ │ fused faster │
  ├────────────────────────────┼─────────┼───────────┼───────────┼───────────┼────────────┼──────────┼──────────────┤
  │ POC baseline               │     108 │       114 │       130 │       230 │        117 │    +6 ms │           2% │
  ├────────────────────────────┼─────────┼───────────┼───────────┼───────────┼────────────┼──────────┼──────────────┤
  │ #1999 (regression)         │     111 │       216 │       255 │      1037 │        252 │  +105 ms │           0% │
  ├────────────────────────────┼─────────┼───────────┼───────────┼───────────┼────────────┼──────────┼──────────────┤
  │ adaptive-tail-totals (fix) │     106 │       106 │       122 │       374 │        115 │    +0 ms │          44% │
  └────────────────────────────┴─────────┴───────────┴───────────┴───────────┴────────────┴──────────┴──────────────┘
```

### Related Issues

### Check List


vibrantvarun (commented):
Reviewed across three lenses (logic, tests, adversarial). Bottom line: the core hits.total arithmetic is sound and I found no HIGH-severity correctness bug or crash. Verified the linchpin against core — SearchPhaseController.TopDocsStats.getTotalHits() caps each sub-search's own reduced count at track_total_hits and forces gte, and it runs on each leg's MultiSearch with the SAME threshold the response stitches, so "a leg reports {T,gte}" ⟺ leg's true count > T ⟺ (leg ⊆ union) union > T ⟺ the Tail path would emit the identical {T,gte}. This defeats the multi-shard attack (the cap is global per sub-search, not per-shard), the ==threshold boundary (a leg at T reports eq, not gte, so it doesn't trigger), and exact/disabled totals (never enter the Top-only path). The Tail-keep predicate is complete — aggs, highlight, non-_score/asc sort, collapse, search_after, min_score, page-past-window, named-ANN-filled-window, response processors are all excluded, and post_filter is safe because it's propagated to the legs so the subset argument survives it. FusedTotalHitsMerger stitches only the total, only on the Top-only path, and composes with the other seam rebuilders.

Actionable items inline (2 medium, then coverage/scope). None block correctness; #1 (rescore) is the one I'd most want closed before merge.

Minor/nits not inline: explain + derived-total is an untested two-merger chained-rebuild path (likely fine — explanations are coordinator-assembled and don't read the Tail); the arming gate in HybridQueryBuilder re-checks threshold/page/response-processors but not requestShapeAllowsDerivedTotalHits, so a request processor that adds aggs between filter-time and rewrite-time arms the legs needlessly (perf-only — the authoritative gate re-checks and keeps the Tail); and requestShapeAllowsDerivedTotalHits's unit test omits the highlighter branch (covered at the filter level).

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -860,6 +904,129 @@ private static boolean wantsTotalsBeyondWindow(SearchSourceBuilder source, int n
...
+     * with them absent the page changes. The page bound itself needs the ranked count and is checked separately.
+     */
+    static boolean requestShapeAllowsDerivedTotalHits(SearchSourceBuilder source) {
+        if (Objects.isNull(source)) {
+            return false;
+        }
+        if (Objects.nonNull(source.aggregations()) || Objects.nonNull(source.highlighter())) {
+            return false;
+        }
+        if (Objects.nonNull(source.collapse()) || Objects.nonNull(source.searchAfter()) || Objects.nonNull(source.minScore())) {
```
vibrantvarun, line 945: MEDIUM — `rescore` is not excluded here and is untested on the derived path. This shape check covers aggregations/highlighter/collapse/searchAfter/minScore (and sorts below), but not `source.rescores()`. So a fused request with `rescore` + default totals + a leg over the threshold drops the Tail and runs BOTH the #1999 rescore-sentinel normalizer and the totals merger on the same response. I traced it as *probably* correct (the page fits within numRankedDocs, rescore doesn't affect total_hits, and the window guard bands unranked docs), but the argument is subtle once `query_weight <= 0`, `score_mode: multiply`, or `rescore_window_size > numRankedDocs` interact with band separation — and with Top-only there are no Tail-only docs in the pool for the guard to band, which differs from the Tail-kept path. There's zero coverage: the only rescore test turns totals off. Please either add `source.rescores()` to this predicate, or add an IT (fused + rescore + default totals + a capped leg) asserting the page AND total are identical to the exact-totals Tail-kept path.

src/main/java/org/opensearch/neuralsearch/query/CandidateScope.java
```diff
@@ -546,11 +581,14 @@ void enableLegExplain() {
...
-        SearchSourceBuilder legSource = new SearchSourceBuilder().query(leg)
-            .size(windowSize)
-            .from(0)
-            .fetchSource(false)
-            .trackTotalHits(false);
+        SearchSourceBuilder legSource = new SearchSourceBuilder().query(leg).size(windowSize).from(0).fetchSource(false);
+        // A leg counts to the threshold when the rewrite asked for it (see legTotalHitsThreshold); otherwise totals stay
+        // off, as before.
+        if (Objects.nonNull(legTotalHitsThreshold)) {
+            legSource.trackTotalHitsUpTo(legTotalHitsThreshold);
```
vibrantvarun, line 588: MEDIUM (perf) — arming every leg to count to the threshold is a strict slowdown for the population that gets no benefit. For a default fused search over an index/leg below the threshold, each leg now visits up to `track_total_hits` competitive docs per shard before it can stop, and when no leg reaches the threshold the Tail is kept anyway — so the enumeration is duplicated (the `enableLegTotalHits` javadoc admits this). This PR is sold as a latency win but regresses exactly the below-threshold case. Worth a benchmark on a below-threshold corpus before merge, and consider gating the arming behind an estimated-benefit signal rather than default-on for every fused request.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFusedModeTotalHitsIT.java
```diff
@@ -0,0 +1,216 @@
...
+        return (List<Map<String, Object>>) ((Map<String, Object>) response.get("hits")).get("hits");
+    }
+
+    /**
+     * The lexical leg has DOCS matches, above THRESHOLD, so its count comes back capped and round 2 runs Top-only — and the
+     * response's total is exactly what the Tail path reports: {@code {THRESHOLD, gte}}. The Tail path is produced here by
+     * asking for exact totals, which always keeps it, then compared at the same threshold semantics.
+     */
+    @SneakyThrows
+    public void testTotalHits_whenALegExceedsTheThreshold_thenTheTailIsDroppedAndTheTotalIsUnchanged() {
```
vibrantvarun, line 208: Test-coverage gap on the number that matters. The mechanism is oracle-anchored (the profiled test proves the Tail was physically dropped), but the derived `{THRESHOLD, gte}` value+relation here is compared only to hand-coded constants and to a *different-valued* exact-totals path ({DOCS,eq}) — the ranked page is asserted identical, not `hits.total`. Nothing runs the SAME request at the SAME `track_total_hits` through a Tail-*kept* path (e.g. add an aggregation, which forces the Tail while still tracking to 8) and asserts it also yields `{8, gte}` — that symmetric parity oracle is what would catch an off-by-one or eq-vs-gte at the cap. Also missing: a comparison against classic (non-fused) hybrid, and an explicit per-shard-below/union-above assertion (the 2-shard setup exercises it but never pins that each shard's count < threshold). And please confirm the IT actually enables fused mode (the #1993 opt-in) — the four non-profiled tests could pass trivially in classic mode if BaseNeuralSearchIT doesn't default it on.

src/main/java/org/opensearch/neuralsearch/search/FusedTotalHitsMerger.java
```diff
@@ -0,0 +1,73 @@
...
+ * count at the request's {@code track_total_hits} threshold: whenever the union has more matches than the threshold the
+ * response says {@code {threshold, gte}} and nothing more — see {@code SearchPhaseController.TopDocsStats#getTotalHits}.
+ * A leg's own total is a lower bound on the union's and is capped the same way, so a leg that reports {@code gte} at the
+ * threshold already proves what the Tail would have reported. The rewrite then drops the Tail and hands that
+ * {@code {threshold, gte}} here; round 2 reports its own (window-sized) total, and {@link #getMergedResponse} puts the
+ * derived one in its place on the way out.
+ *
+ * <p>The derived total describes round 1's view of the index, a moment before round 2 ran; the two say the same thing
+ * whenever both rounds reach the same shards and see the same reader state — the assumption the two-round design already
+ * makes about the ranked window. Under {@code allow_partial_search_results} a shard lost in round 2 alone is not reflected
```
vibrantvarun, line 28: Good that this documents the partial-results / `now` / terms-lookup cases where the derived bound may not equal what round 2 would count. One more member of that family isn't named or tested: a leg that hits a soft `timeout` after crossing the threshold derives `{T,gte}` and drops the Tail, whereas a timed-out Tail rerun would report a smaller partial count — a difference in both value and relation. Worth adding to this note so "identical to the Tail" is scoped to complete, stable-index results, and (optionally) excluding the derivation when a round-1 leg reported `timed_out`.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: Ack, will add test. Adding `source.rescores()` is not practical, risk is actually inverted. Top-only round 2 has no `filter` clause, so match set == ranked set and there are no unranked docs for any weighting to promote. Excluding `rescore` would force these requests back onto the riskier Tail path and neglet the win.

src/main/java/org/opensearch/neuralsearch/query/CandidateScope.java
martin-gaievski: ack, mechanism is right: `TopScoreDocCollector` cannot call `setMinCompetitiveScore` until the hits threshold is reached, so arming disables dynamic pruning for the first `T` hits. Already measured at ~11 ms

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFusedModeTotalHitsIT.java
martin-gaievski: Detailed response for each of the asks:
- no symmetric parity oracle: same request, same `track_total_hits`, Tail-kept path - ACK
- per-shard-below / union-above never pinned - ACK
- compare against classic hybrid - I'd push back, that's out of scope for this PR. Classic runs a different shard-side collector with known totals divergences (`min_score` + aggs). this PR's claim is "derived total equals what Tail path reports", not "equals classic"
- confirm the IT actually enables fused mode - Confirmed/not actionable: `BaseNeuralSearchIT` has [`@Before optInToFusedMode()` → `tryEnableFusedMode()`](https://github.com/opensearch-project/neural-search/blob/feature/fusion-hybrid-query/src/testFixtures/java/org/opensearch/neuralsearch/BaseNeuralSearchIT.java#L189-L192), so extending `BaseNeuralSearchIT` is sufficient

src/main/java/org/opensearch/neuralsearch/search/FusedTotalHitsMerger.java
martin-gaievski: ack on updating soft-`timeout` case in `FusedTotalHitsMerger`'s note. I would not optionally exclude  derivation for a round-1 leg reported `timed_out` scenario. It would make the answer less accurate: a leg that counted ≥ T before timing out proves ≥ T real matches, so `{T,gte}` is true of the union, while a timed-out Tail's smaller partial count is the weaker statement. besides, `timed_out:true` still reaches response via `FusedLegTimeoutMerger`.

src/main/java/org/opensearch/neuralsearch/query/CandidateScope.java
martin-gaievski: Details on the requested measurements - From test cluster before/after run of this change (adaptive-Tail section, same collection, drift-controlled):

| ratio | before | after |
|---|---|---|
| `std_mm ÷ classic` | 1.783x | 1.293x |
| `std_mm ÷ fast_mm` (Tail-free, legs not counting) | 1.491x | **1.172x** |

Both arms of the second row are Top-only round 2; only difference is the legs counting to 10 000. That
residual 11 ms of a 75 ms p50 is the leg-counting cost, and it is what a below-threshold request pays for nothing. Against it, the change removed ~25–27 ms of Tail cost on the above-threshold path.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
vibrantvarun: Agreed, and I concede the exclude suggestion — you're right the risk is inverted: Top-only round 2 has no filter clause, so match set == ranked set and there are no unranked docs for the #1999 guard to demote; excluding rescore would only force these onto the riskier Tail path. testRewrite_whenTheRequestCarriesARescore… + the IT page-parity test pin it. Resolved.

src/main/java/org/opensearch/neuralsearch/query/CandidateScope.java
vibrantvarun: Thanks for measuring — ~11ms and the std_mm÷fast_mm ratio improving 1.491→1.172 answers the 'benchmark it' ask. Residual (below-threshold indices still pay the counting cost for no derivation benefit) is acceptable at that magnitude. Resolved.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFusedModeTotalHitsIT.java
vibrantvarun: The same-threshold Tail-kept parity oracle and the per-shard-below/union-above (_shards:N) test are exactly what was needed — resolved. Conceding the classic-hybrid comparison: fair, the claim is derived==Tail, not ==classic, so it's out of scope here.

src/main/java/org/opensearch/neuralsearch/search/FusedTotalHitsMerger.java
vibrantvarun: Agreed — {T,gte} from a leg that counted past T before timing out is a true (and stronger) lower bound on the union than a timed-out Tail's partial count, and timed_out still surfaces via FusedLegTimeoutMerger. The doc note is the right resolution; withdrawing the optional-exclude suggestion. Resolved.

vibrantvarun (approved):
LGTM on 095f60b397. Re-reviewed against my earlier three-lens review: all four inline findings are resolved (rescore correctly not-excluded + now test-pinned; perf measured; same-threshold parity oracle + per-shard-below/union-above ITs added; soft-timeout javadoc), and the three secondary suggestions I raised were soundly rebutted (I've conceded each in-thread). The one production change is exactly the arming-symmetry fix — the arm gate re-checks requestShapeAllowsDerivedTotalHits — which is surgical and correct; the rest is tests + a doc note, core derivation logic unchanged. The central {T,gte} inference remains verified against core's TopDocsStats cap, the Tail-keep predicate is complete, and the merger composes correctly with the other seam rebuilders. Nice work.

bzhangam:
This optimization only helps on certain cases and we still have cases suffering the extra latency. Probably later in documentation we should provide some guide on how to configure the track_total_hits when use hybrid + fusion. By default we should recommend to disable the track_total_hits when it's not needed. If we do need it to support features like pagination we should recommend to use the smallest threshold meeting their UI requirement to best leverage this optimization.

bzhangam (approved):

martin-gaievski:
> This optimization only helps on certain cases and we still have cases suffering the extra latency. Probably later in documentation we should provide some guide on how to configure the track_total_hits when use hybrid + fusion. By default we should recommend to disable the track_total_hits when it's not needed. If we do need it to support features like pagination we should recommend to use the smallest threshold meeting their UI requirement to best leverage this optimization.

Agreed on all points. This belongs in the fused-mode documentation, and I'll fold it into the public docs issue with a small decision table. Two mechanics worth writing down because they sharpen the  "smallest threshold" advice:

1. `track_total_hits: false` is the fastest configuration and stays so in the follow-ups: no leg counting, Top-only round 2, and in the coordinator fast path (next PR on this branch) round 2 is skipped entirely for the common shape.

2. There's a free tier for tracked totals: an integer threshold at or below window_size never arms leg counting at all (legTotalHitsThreshold returns null for T ≤ window). ranked window itself proves {T, gte} whenever it fills, so the Tail is dropped at zero added cost; when fewer than T documents match, the Tail returns and reports the exact count, which is what the caller wants there anyway. Since pagination independently needs from + size inside the window (a page past it keeps the Tail regardless), "smallest threshold meeting the UI requirement" concretely means: size window_size to the deepest page you'll serve and keep the threshold at or under it — e.g. 10 pages × 10 hits → window 100, threshold 100, no counting armed. Above the window, smaller is still better twice: cheaper per-shard counting and a higher chance a leg proves the count.

Population that pays without benefit is a corpus whose union never reaches the threshold. Legs count (~11 ms at c=8 in my measurements, and load-dependent: near zero with worker headroom) and the Tail is kept anyway. That's inherent to proving counts from the legs, so the guide should say it plainly: if your corpus is small relative to the threshold, disable totals or drop the threshold under the window. Shapes that keep the Tail for their documents (aggregations, highlight, exact totals) are a separate follow-up PR.

PR #2008, opened 2026-09-15
[Fusion] Skip round 2 of fused query and use legs information to get final search hits results

martin-gaievski:
### Description
Next optimization round aiming to lower latency by cutting round 2 for a typical hybrid query case. For top-k by fused score, document's own fields, nothing computed over the full match set, round 2 is mainly a waste.

With this fix legs themselves now carry the page: when the fast path is eligible they fetch requested details (`_source` with its includes/excludes, stored_fields, docvalue_fields, fields, version, seq_no_primary_term). Coordinator assembles the page from the window it already holds, then rewrite self-erases into match_none, and the assembled page is placed on the response through the ActionFilter (already part of resolver). 

As per benchmarks this took typical fused request from 1.1-1.29x classic hybrid to 0.9-0.95x depending on dataset and resource capacity.

Everything for taking/skipping round 2 decided before legs run: aggregations, highlight, any sort, collapse, search_after, min_score, rescore, profile, exact totals, script_fields, a named leg or a leg with inner_hits, a search pipeline with response processors, and a nested fused hybrid. 

This PR has three main functional parts:

1. fast path itself: leg fetch (CandidateScope.enableLegFetch, propagateFetchFields), post-leg decision and page assembly (HybridFusionOrchestrator.buildFusedResult/assemblePage), response swap (FusedHitsMerger, wired first in the filter's rebuild order so explanations and totals land on the swapped page)

2. drop of leg sort. 
in current code each leg returns results sorted by [_score, _doc], then doc ids to the coordinator so it could copy round 2's tie order exactly. this costs ~10 ms per leg, roughly the same 10-13 ms the fast path saves by skipping round 2. 
with this PR legs are unsorted, which is cheap. Case of scores tie on the assembled page: first by shard (same as round 2), then within a shard by index-name + _id. That rule is deterministic, same answer on every replica, before and after segment merges. Doc-id order, by contrast, changes across replicas and merges.
One client-visible consequence: for same-shard bit-identical fused scores the order — and at a page edge, which of the tied documents the page shows — can differ from round 2. This effect depends on exact normalization technique - duplicates are rare for min_max/l2/z_score (scores are rarely bit identical), but more often for rrf due to rank based computation.

3. fetch-volume gate. 
Current code over-fetches documents,  in round 2 fetches exactly documents on the page, size of them (say 10). The fast path skips round 2 by having each leg fetch its whole candidate window up front — legs × window documents (say 2 × 100 = 200) — because at that moment nobody knows yet which 10 will make the page. it over-fetches ~190 documents to save one network round. It's optimal if documents are small, costs microseconds. But for larger documents like vectors (easy can reach 1K dimensions) fetch itself became a bottleneck. 
I made this logic smarter, we do estimates extra bytes that will be fetched:
```(documents over-fetched: legs × window − size) × (embedding bytes per document, read from the index mapping: vector dimension × bytes per number, ~4 KB flat for rank_features)```
If the estimate exceeds 1 MB, the fast path is refused and the request runs the normal two rounds.

Following table shows latency percentiles before (A) and after (B) this change. Cluster configuration: 3 data r5.xlarge nodes, Quora dataset, 12 shards, 0 replicas.

  ```
┌──────┬───────────┬────────────┬───────────┬────────────┬──────────────┐
  │ stat │ A classic │ A resolver │ B classic │ B resolver │ resolver B−A │
  ├──────┼───────────┼────────────┼───────────┼────────────┼──────────────┤
  │ p50  │     114.0 │      117.0 │     114.0 │      108.0 │         −9.0 │
  ├──────┼───────────┼────────────┼───────────┼────────────┼──────────────┤
  │ p90  │     131.0 │      133.0 │     132.0 │      124.0 │         −9.0 │
  ├──────┼───────────┼────────────┼───────────┼────────────┼──────────────┤
  │ p99  │     141.0 │      140.0 │     147.0 │      133.0 │         −7.0 │
  ├──────┼───────────┼────────────┼───────────┼────────────┼──────────────┤
  │ mean │     115.2 │      118.4 │     116.0 │      110.0 │         −8.4 │
  └──────┴───────────┴────────────┴───────────┴────────────┴──────────────┘
```

### Related Issues
#1930 

### Check List


src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -1028,7 +1056,236 @@ private static TotalHits[] legTotalHits(MultiSearchResponse.Item[] items) {
...
+            return FusedResult.twoRound(new MatchNoneQueryBuilder());
+        }
+        // The page has to come from the ranked window alone: past it, the two-round path returns Tail-only documents.
+        boolean pageFits = requestedPageEnd(source) <= ranked.ids().length;
+        // Totals: disabled → none; otherwise what round 2 would have reported must already be known (see below).
+        Integer trackTotalHitsUpTo = source.trackTotalHitsUpTo();
+        boolean totalsDisabled = Objects.nonNull(trackTotalHitsUpTo) && trackTotalHitsUpTo == SearchContext.TRACK_TOTAL_HITS_DISABLED;
+        TotalHits totalHits = totalsDisabled ? null : totalHitsForFastPath(source, ranked.ids().length, legTotalHits(items), windowSize);
+        boolean totalsSettled = totalsDisabled || Objects.nonNull(totalHits);
+        if (pageFits == false || totalsSettled == false) {
```
vibrantvarun, line 1293: Perf (worth fixing before merge) — fusion is computed twice on the two-round fallback. buildFusedResult has already run groupLegHits + computeRankedDocs (and recorded FusedCoordinatorTimings + FusedDocExplanations) by the time it reaches this fallback; buildFusedQuery then runs groupLegHits + computeRankedDocs again, discarding the RankedDocs just computed. Two consequences: (1) every request that arms the fast path but then needs round 2 (page past the window, or the count not provable from the legs) pays coordinator-side fusion/normalization/explanation-recording twice; (2) it double-writes FusedCoordinatorTimings, so a profiled fallback reports only the second pass. Not a correctness issue (explanation re-record is an idempotent map-put), but a real regression on the fallback path in a latency-focused PR — thread the already-computed RankedDocs/legHits into an overload of buildFusedQuery instead of recomputing.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFusedModeFastPathIT.java
```diff
@@ -0,0 +1,397 @@
...
+
+    /**
+     * The one client-visible difference, pinned down. One shard, twelve identical documents: every fused score ties, so
+     * round 2 would order the page by Lucene doc id while the fast path orders it by {@code _id} — "1", "10", "11", "12",
+     * "2", ... — deterministically, on every replica, across merges. Ids, scores, sources, total and max_score are the
+     * two-round path's. The {@code _id} order is itself the proof the fast path ran: round 2 cannot produce it here.
+     */
+    @SneakyThrows
+    @SuppressWarnings("unchecked")
+    public void testFastPath_whenFusedScoresTie_thenTiesAreOrderedByIdAndEverythingElseMatchesTwoRounds() {
```
vibrantvarun, line 271: Tie-ordering coverage has two gaps, both around the exact behavior this PR changes:

1. (High) No rrf test. This tie test — and the unit one — use min_max/arithmetic_mean, where equal raw scores normalize to 1.0 only in a contrived corpus. rrf produces bit-identical scores *by construction* (same rank position across legs → identical rank-constant contribution), so it's precisely where the new shard→_index+_id tie-break replaces round 2's _doc order. Add a unit + IT tie case with technique: rrf.

2. (Medium) The one sanctioned client-visible difference — a tied-score run straddling from/from+size showing a different doc at the page edge than round 2 — has no positive test. This case uses size==window (no straddle) and canonicalTieOrder re-sorts within equal-score runs on both sides, so it structurally cannot observe the boundary difference the roundTwoOrder javadoc documents. Add a test with an equal-score run crossing the page edge (from>0) asserting the fast path keeps a valid tied occupant.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFusedModeFastPathIT.java
```diff
@@ -0,0 +1,397 @@
...
+ * {@code profile: true}, which the fast path refuses (round 2's own tree is part of what profile reports), with the
+ * profile section stripped before comparing. Which path ran is proved intrinsically where the paths are
+ * distinguishable at all: on the one-shard all-ties index the fast path's {@code _id} tie order differs from round 2's
+ * doc-id order, so {@link #testFastPath_whenFusedScoresTie_thenTiesAreOrderedByIdAndEverythingElseMatchesTwoRounds}
+ * passing proves the fast path really runs; path selection for every other shape is pinned at unit level
+ * ({@code requestShapeAllowsFastPath}, {@code legsAllowFastPath}, the fetch budget, and the fan-out rewrite tests).
+ */
+public class HybridQueryFusedModeFastPathIT extends BaseNeuralSearchIT {
+
+    private static final String INDEX = "test-hybrid-fused-fast-path";
```
vibrantvarun, line 44: Every fast-path IT is single-index, so multi-index page assembly is unproven end-to-end — the composite _index+_id keying, per-hit _index/_source, and cross-index _index-then-_id tie order are only unit-touched. Worth a multi-index identity IT (same request, fast path vs profile:true control). Related edge worth a note: under custom routing the same _id can live on two shards of one index as two distinct docs; the fast path assembles one hit (deduped by key) while round 2's _id-term matches both → two hits, so the paths disagree on hit count. It's rooted in the documented _index+_id keying limitation (HybridFusionOrchestrator ~line 509) rather than new here, but the fast path does change the observable count vs round 2 — worth documenting, or normalizing so the two paths at least agree.

vibrantvarun (commented):
Reviewed across three lenses (logic, adversarial edge-cases, tests) and cross-verified against core. No correctness bug — the round-2-equivalence contract holds and is genuinely oracle-anchored (the IT runs the same request through the fast path and a forced round-2 control via profile:true and asserts hits/_source/fields/version/seqNo/score/order identical). Verified the eligibility gate is complete (post_filter + PIT propagated to legs; suggest is query-independent and survives the match_none self-erase; terminate_after/indices_boost rejected upstream; _explain supported via enableLegExplain + fused-breakdown re-apply; the 14 disqualifiers all gate correctly), page assembly reproduces round 2 field-for-field, the tie-break is deterministic + replica-stable and cross-shard ties match core's shardIndex order, and ReturnedEmbeddingFields is a pure latency predictor (fails closed, cannot drop or mis-select a field).

One thing I checked and am NOT flagging: the totalHitsForFastPath `numRankedDocs == threshold → {threshold, eq}` branch is correct, not a gte/eq divergence — it only fires when threshold is not below numRankedDocs, so round 2 wouldn't build the Tail either and would also report {threshold, eq}. It matches round 2; the eq-vs-gte-'truer' question is the pre-existing needs-Tail boundary, unchanged here.

Actionable items inline: one perf regression on the fallback path (worth fixing before merge) and two test-coverage gaps. Minor notes not inline: under explain:true a hit's _shard/_node comes from the leg's shard target and may name a different replica than round 2 (cosmetic; replica selection is non-deterministic anyway); a small fully-counted result set (numRankedDocs < windowSize, exact total already in hand) falls back to two rounds unnecessarily — missed optimization, not a bug; and stored_fields / derived-source / which-leg-instance-wins-on-duplicate aren't asserted end-to-end.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: ack, working on the fix

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFusedModeFastPathIT.java
martin-gaievski: ack, adding both requested tests: rrf and equal score run that is over the page boundary

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFusedModeFastPathIT.java
martin-gaievski: ack, will add one for two indexes

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: Fixed in 3d1cf60c. Extracted the substitute-assembly tail into `buildSubstitute(source, items, legHits, ranked, ...)`; the armed fallback now calls it with the `legHits`/`ranked` already computed in `buildFusedResult`, so `groupLegHits` + `computeRankedDocs` (and their `FusedCoordinatorTimings`/`FusedDocExplanations` side effects) run exactly once. The un-armed path keeps the recomputing `buildFusedQuery` overload since it never ran fusion before reaching it. No behaviour change — empty-ranked and totals-consumer semantics preserved on both paths; a profiled fallback now records one pass.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFusedModeFastPathIT.java
martin-gaievski: Both added in 3d1cf60c.

1. rrf tie order — unit `testBuildFusedQuery_rrf_equalScoreTiesOrderByIndexThenId`: two docs holding rank 0 in their respective legs get bit-identical rrf scores and `toRankedDocs` orders that run by `(_index, _id)`, not insertion. This is where the tie can be constructed exactly. Plus IT `testFastPath_whenRrf_thenPageMatchesTwoRounds` asserting a whole rrf request equals its two-round control (ties included) on a real index. (Note: the all-ties corpus does not make rrf ties — rrf ties on equal *rank across legs*, so the exact-tie proof is the unit test; the IT proves the rrf page assembles identically end to end.)

2. page-edge straddle — IT `testFastPath_whenATiedRunStraddlesThePageEdge_thenTheFastPathKeepsValidTiedOccupants`: `from:5,size:3` over the all-ties index slices an equal-score run mid-run; asserts the fast path returns the `_id`-ordered occupants of the straddled slice (valid tied occupants), the documented boundary allowance.

src/test/java/org/opensearch/neuralsearch/query/HybridQueryFusedModeFastPathIT.java
martin-gaievski: Added IT `testFastPath_whenSpanningTwoIndices_thenPageMatchesTwoRounds` in 3d1cf60c: one fused request over two indices that hold the **same** `_ids` (1..6), 2 shards each, asserting the assembled page equals the `profile:true` two-round control field for field — exercising composite `_index+_id` keying, per-hit `_index`/`_source`, and cross-index tie order end to end, and proving same-`_id` docs across indices are not merged.

On the custom-routing duplicate-`_id` edge: documented rather than normalized (normalizing would be a behaviour change beyond this PR). Added a note on `computeRankedDocs` that `_index+_id` keying makes such a document appear once on the fast path vs twice under round 2 s `ids` clause — a property of the keying the fast path makes visible in `hits`, rooted in the existing limitation.

martin-gaievski:
Thanks @vibrantvarun — addressed the three actionable items:

- **Perf (fallback double-fusion):** `buildSubstitute` reuses the already-computed `legHits`/`ranked`; fusion + timings/explanations run once on the armed fallback.
- **rrf tie coverage:** unit test pinning the `(_index, _id)` tie order on bit-identical rrf scores, plus an rrf identity IT.
- **page-edge straddle:** IT with `from>0` across an equal-score run.
- **multi-index identity IT** (same `_ids` across two indices).

The info-only notes I left as-is with a reply: the `{threshold, eq}` branch (verified correct), the explain `_shard`/`_node` replica cosmetic, and the small-fully-counted-result missed optimization (tracking as a follow-up, not this PR). The custom-routing duplicate-`_id` count divergence is now documented on `computeRankedDocs`.

vibrantvarun (approved):
Approving 3d1cf60c after a fresh tester-lens pass on top of the earlier three-lens review (logic / adversarial edge-cases / test-coverage). Nothing concrete left to tear down.

Adversarially re-checked the new commit — the buildSubstitute extraction is the only new production code and it is behavior-preserving: buildFusedQuery (the un-armed path) now computes ranked and delegates the identical tail to buildSubstitute, and the fast-path fallback calls buildSubstitute with the legHits/ranked already computed in buildFusedResult, so fusion runs exactly once and its timings/explanations are written once. Same inputs from the same MultiSearchResponse → identical substitute; fusion is correctly dropped from buildSubstitute since it was only needed for computeRankedDocs. No regression.

All three of my earlier findings are resolved: (1) the double-fusion-on-fallback perf issue is fixed as above; (2) tie coverage now includes rrf (unit testBuildFusedQuery_rrf_equalScoreTiesOrderByIndexThenId + an end-to-end rrf IT) and the page-edge straddle case; (3) multi-index page assembly has a same-_id two-index IT asserting field-for-field parity against the profile:true control, and the custom-routing same-_id-two-shards hit-count difference is now documented on computeRankedDocs.

Earlier verification still stands: the fast-path == round-2 invariant is oracle-anchored via the profile:true differential control, the eligibility gate is complete (post_filter/PIT propagated to legs; suggest survives the match_none self-erase; terminate_after/indices_boost rejected upstream; _explain supported), page assembly reproduces round 2 field-for-field, the tie-break is deterministic + replica-stable, and ReturnedEmbeddingFields is a pure latency predictor that fails closed. The residual notes I left are informational (explain _shard/_node replica cosmetic; small fully-counted sets fall back — missed optimization, not a bug). LGTM.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
```diff
@@ -1028,7 +1085,227 @@ private static TotalHits[] legTotalHits(MultiSearchResponse.Item[] items) {
...
+            return false;
+        }
+        if (Objects.nonNull(source.scriptFields()) && source.scriptFields().isEmpty() == false) {
+            return false;
+        }
+        if (source.profile()) {
+            return false;
+        }
+        Integer trackTotalHitsUpTo = source.trackTotalHitsUpTo();
+        return Objects.isNull(trackTotalHitsUpTo) || trackTotalHitsUpTo != SearchContext.TRACK_TOTAL_HITS_ACCURATE;
```
bzhangam, line 1152: **Latency regression on the default-totals path.**

Default `track_total_hits` is 10,000, and a default request arms the fast path (only `TRACK_TOTAL_HITS_ACCURATE` is rejected), so every leg fetches full `_source`/fields for its whole window in round 1.

But for the default threshold, settlement rests entirely on a single leg reporting `gte(10000)`. `numRankedDocs >= threshold` can't help — `numRankedDocs` is `ranked.ids().length`, capped at `window_size` (~100), so it's never ≥ 10,000. So unless a leg individually matches ≥ 10,000 docs, `totalsSettled` fails *after* the legs have already fetched, and we fall back to `buildSubstitute`, whose round-2 Top re-fetches the page.

That's a high-recall condition, and often unmet: for the common case where legs match a few hundred–few thousand docs, the default request pays the heavy round-1 over-fetch **and** a full round 2 — worse than classic's id-only round 1.

So the regression isn't an edge case — it's the default shape unless a leg individually matches over 10,000. Do we have data on how often that holds? If it's the minority, the fast path over-fetches then does round 2 anyway on most default requests. So not sure if this optimization really useful or not. Options: don't arm the heavy fetch until the count looks leg-derivable, or reuse the round-1 source on fallback instead of re-fetching.

bzhangam:
Before I approve, I'd like to align on the value case, because I think this optimization targets a narrower slice than fused mode's motivation suggests.

Per the RFC (#1930), the headline motivation for fused mode is **composability** — the classic hybrid's central limitation is that it "must be the top-level query," and the RFC's marquee example is a hybrid nested inside a `bool`. But this fast path structurally requires the hybrid to be the literal top-level query (`mayTakeFastPath` checks `source.query() instanceof HybridQueryBuilder`, and the hits consumer only attaches when `finder.found.get(0) == source.query()`). So the fast path can't apply to any of the nested/repeated-hybrid cases that motivated fused mode in the first place — it only helps the top-level top-K shape, which is the shape classic hybrid already handled.

That leaves the justification resting on the latency argument (RFC motivation #2) for a specific cohort: existing top-level hybrid users who adopt fused mode purely for the ~10–13ms/req win. A couple of questions:

1. **Coverage** — what fraction of fused-hybrid traffic actually clears the full arming gate (top-level + no aggs/highlight/sort/collapse/search_after/min_score/rescore/script_fields/profile/exact-totals + page-fits + count-provable + under the 1MB fetch budget)? A 10ms win on a small slice is hard to weigh against the maintenance cost.

2. **Divergence containment** — this adds a second page-assembly path that must reproduce round 2 field-for-field, with correctness resting on the `profile:true` differential IT rather than a structural guarantee, plus one permanent client-visible tie-order difference. That's a durable tax: every future change to fused-mode behavior now has to be kept bit-identical across both paths. Is owning that justified by the coverage in (1)?

To be clear, I think the implementation is careful and well-tested — this is about scope/ROI, not correctness. If there's data showing the migrating-top-level cohort is large, I'm comfortable. Otherwise I'd want to weigh narrowing scope or deferring.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: Your reading of the mechanism is right: with the default threshold 10,000 and `window_size` 100, `numRankedDocs` can never reach the threshold, so settlement rests on a single leg reporting `gte(10000)`, knowable only after the legs have already fetched. I measured that case on the same cluster I ran benchmarks (3 r5.xlarge nodes, Quora, 12 shards)

**Coverage of the settlement condition, n=400:**  default-operator lexical leg proves `gte(10000)` in 100% of queries; with `operator: and`, 0%. So the failure case is not a per-query lottery — it is a leg shape, and I have it as its own arm.

failing case: `operator: and` legs, 0% settlement, every request over-fetches then runs round 2:

| c | armed | disarmed | classic | armed − disarmed (paired p50) | armed slower share |
|---|---|---|---|---|---|
| 1 | 106.0 / 105.9 ms | 106.1 / 106.0 | 85.2 / 85.2 | **−0.1 / −0.1 ms** | 0.48 / 0.47 |
| 8 | 203.7 / 207.8 ms | 208.2 / 209.0 | 131.5 / 135.9 | **−5.5 / +0.5 ms** | 0.46 / 0.51 |

Shard-op counters confirm the mechanism independently: `armed` and `disarmed` do identical work here, 36.0 query ops and 19.2 fetch ops per request, both, so fallback really does run the same two rounds as before. The only difference is the leg-fetch payload, and it is within noise.

Why it is practically free: over-fetch is `2 × 100 − 10 = 190` docs × ~3.2 KB ≈ 0.6 MB, under the 1 MB budget, what gate bounds. For the same reason I checked the other proposed approach, reusing the round-1 page on fallback instead of re-fetching: round-2 **fetch phase** on the two-round path measures **0.0 ms** (paired, `_source` on vs off, c=1 and c=8); fallback's cost is the round-2 query phase plus the extra coordinator↔shard round, which reuse would not touch. Not worth a third response-composition mode.

Where you are right, and what I'm changing in this PR. gate estimates embedding payload only (`knn_vector` dimension × width, ~4 KB per `rank_features`). A large text document in `_source` contributes 0 to the estimate, so on a 50 KB/doc corpus with no vector field the legs fetch 190 × 50 KB ≈ 9.5 MB against round 2's 0.5 MB page, and at the ~3–10 ms/MB I measured on a tiered topology that is a real 30–95 ms loss, on the success path too, not only the fallback. This is where fast path can be slower than two rounds.

Rather than tighten the budget statically, I'm extending the gate to a measured `_source` size. With that fast path arms only when the actual extra payload of vectors or text is under budget. On Quora corpus (3.2 KB/doc) nothing changes; at ~5.5 KB/doc and above (window 100, size 10) it now correctly takes two rounds. I'll add the liveness IT using per-request `_stats/search` fetch-op deltas (deterministic on a 1-shard index: 3 ops two-round vs 2 fast path), which also gives the suite the intrinsic path oracle it lost when the counter was dropped.

For completeness, the benefit side on the same cluster and queries: armed − disarmed = **−8.2 / −7.9 ms at c=1**, **−23.3 / −22.9 ms at c=8** (armed slower in 1% / 13% of pairs); armed fused within ±5 ms of classic

martin-gaievski:
> Before I approve, I'd like to align on the value case, because I think this optimization targets a narrower slice than fused mode's motivation suggests.
> 
> Per the RFC (#1930), the headline motivation for fused mode is **composability** — the classic hybrid's central limitation is that it "must be the top-level query," and the RFC's marquee example is a hybrid nested inside a `bool`. But this fast path structurally requires the hybrid to be the literal top-level query (`mayTakeFastPath` checks `source.query() instanceof HybridQueryBuilder`, and the hits consumer only attaches when `finder.found.get(0) == source.query()`). So the fast path can't apply to any of the nested/repeated-hybrid cases that motivated fused mode in the first place — it only helps the top-level top-K shape, which is the shape classic hybrid already handled.
> 
> That leaves the justification resting on the latency argument (RFC motivation #2) for a specific cohort: existing top-level hybrid users who adopt fused mode purely for the ~10–13ms/req win. A couple of questions:
> 
> 1. **Coverage** — what fraction of fused-hybrid traffic actually clears the full arming gate (top-level + no aggs/highlight/sort/collapse/search_after/min_score/rescore/script_fields/profile/exact-totals + page-fits + count-provable + under the 1MB fetch budget)? A 10ms win on a small slice is hard to weigh against the maintenance cost.
> 2. **Divergence containment** — this adds a second page-assembly path that must reproduce round 2 field-for-field, with correctness resting on the `profile:true` differential IT rather than a structural guarantee, plus one permanent client-visible tie-order difference. That's a durable tax: every future change to fused-mode behavior now has to be kept bit-identical across both paths. Is owning that justified by the coverage in (1)?
> 
> To be clear, I think the implementation is careful and well-tested — this is about scope/ROI, not correctness. If there's data showing the migrating-top-level cohort is large, I'm comfortable. Otherwise I'd want to weigh narrowing scope or deferring.

Your point on composability is accurate: fast path requires hybrid to be the request's own query, so none of the nested/repeated-hybrid shapes from RFC can take it. It targets RFC motivation #2 (latency), not #1, and it is strictly additive, nested and repeated hybrids keep going through the two-round path unchanged. The target group that it serves is the existing top-level hybrid. For such users fused mode without this change is a latency regression if they migrate (1.29× classic after #2004; 0.87–0.95× with it).

On the two questions:

**1. Coverage.** Settlement has three routes; only the third can fail after legs have run:

| Route | Condition | Settles |
|---|---|---|
| 1 | `track_total_hits: false` | always, before the legs run |
| 2 | `track_total_hits: T ≤ window_size` | always (`numRankedDocs ≥ T`) |
| 3 | `T > window_size`, including the **default 10,000** | only if a leg reports `gte(T)` |

Routes 1 and 2 are a free tier. Route 3 is the default shape and there coverage is bimodal rather than "a small slice": for Quora queries the default-operator lexical leg cleared 10,000 in **100%** of cases; an `operator: and` leg in **0%**. And in the 0% case the measured cost of having armed is ~0 (table in the inline reply). The payoff is asymmetric — **−8 ms (c=1) / −23 ms (c=8) on a hit, ~0 on a miss**. This makes arming the default shape defensible. I'm careful not to generalize 100%: one corpus, English, large index. The narrower claim I will stand behind is that route-3 settlement is the norm for OR-style matching over a large index and the exception for conjunctive legs, and that both outcomes are safe.

**2. Divergence containment.** This is the fair cost. What bounds it: page assembly reads the same `RankedDocs` that two-round Top is built from, so changes to fusion or ranking move both paths together by construction. Differential IT covers the fetch/response surface, which is where the paths genuinely differ; the one permanent client-visible difference is confined to same-shard bit-identical fused scores and is now pinned by tests at both levels (`rrf` tie order unit test, page-edge straddle IT, multi-index identity IT); and the gate fails closed on anything it does not recognize.

I'm preping one more commit for this PR raised in your other commit. It surfaced one regime where the fast path is slower: large text documents in `_source`, which fetch gate cannot see because it estimates embedding payload only. Instead of shrinking the fast path to shapes where settlement is guaranteed pre-legs I'll do following: the gate learns the actual returned `_source` size per `(index, _source filter shape)` from responses the filter already handles, uses it in the budget, and fails closed until observed. That keeps the whole measured win, removes the only known way for the fast path to be a regression, and adds no new response path.

src/main/java/org/opensearch/neuralsearch/query/HybridFusionOrchestrator.java
martin-gaievski: Collected timing data with the chages I pushed in latest commit:

**1. regime you pointed, large text the mapping cannot see.** I reindexed Quora with a ~19.5 KB body added to every document (`_source` per hit 19.6 KB with the vector excluded → 3.7 MB over-fetch at size 10 / window 100 / 2 legs; the mapping-only gate estimated 0 and armed). Measured on both builds, on the full 522,931-document corpus (a 100K-document cut under-stated the fast path — the extra round of round 2 gets dearer as the workers load up, the 3.7 MB fetch does not):

| 523K docs × 19.6 KB | c | fetch ops/req armed / disarmed | armed p50 (p90) ms | disarmed p50 (p90) ms | classic p50 (p90) ms | armed − disarmed, paired p50 | as % of disarmed | armed slower share |
|---|---|---|---|---|---|---|---|---|
| before, `ff50e64e` (mapping-only gate → fast path fires) | 1 | **24.0** / 31.0 | 89.3 (94) / 89.1 (94) | 93.4 (99) / 93.0 (97) | 87.7 (95) / 87.7 (94) | **−3.8 / −3.5** | −4.1 % / −3.8 % | 0.05 / 0.09 |
| before | 8 | | 223.2 (262) / 213.4 (242) | 238.5 (285) / 227.9 (262) | 206.0 (240) / 197.8 (227) | **−15.0 / −14.4** | −6.3 % / −6.3 % | 0.30 / 0.29 |
| after, `1a91ae7f` (observed 19.6 KB → refused) | 1 | **31.0** / 31.0 | 93.5 (97) / 93.6 (98) | 94.0 (98) / 94.3 (98) | 88.8 (95) / 89.1 (96) | −0.6 / −0.7 | −0.6 % / −0.7 % | 0.32 / 0.31 |
| after | 8 | | 225.7 (250) / 223.9 (246) | 228.5 (250) / 224.4 (243) | 197.0 (220) / 193.5 (214) | −2.5 / −0.9 | −1.1 % / −0.4 % | 0.46 / 0.49 |

Read the "after" rows as the control's own bias. Net of it, the un-gated fast path on 20 KB documents was **about 3 ms (3–4 %) ahead at c=1 and 13 ms (6 %) ahead at c=8** on this cluster — so two corrections to my earlier reply. First, on co-located hardware the fast path on large text was **not** slower than two rounds; the 30–95 ms figure I quoted came from the per-MB fetch cost on a tiered coordinator/worker topology, where the inversion is real (measured at ~1.5 MB, +5 ms at size 10). The margin shrinks with document size everywhere and inverts only where the fetch is dear relative to the round. Which means the learned gate at a fixed 1 MB budget **forgoes ~3 ms (c=1) / ~13 ms (c=8) on large-text corpora on co-located clusters** while protecting the tiered ones — I would keep the default at 1 MB and make the budget a dynamic cluster setting so operators on co-located clusters can raise it (4 MB arms this case); happy to add that here or as a follow-up, your call. Second, the liveness IT is on two-shard indices with the two reference counts calibrated in-test (a named leg as the two-round reference, `_source: false` as the fast-path reference), not the "1-shard, 3 vs 2" I wrote — a single-shard index runs query and fetch in one shard round and counts a fetch even for `match_none`.

**2. Nothing given back on the shape that benefits.** Quora as-is (3.2 KB/doc, 0.6 MB over-fetch, observed → under budget → armed):

| Quora 3.2 KB, `_source` minus vector | c | fetch ops/req armed / disarmed | armed p50 (p90) ms | disarmed p50 (p90) ms | classic p50 (p90) ms | armed − disarmed, paired p50 | as % of disarmed | armed slower share |
|---|---|---|---|---|---|---|---|---|
| before, `ff50e64e` | 1 | 24.0 / 31.0 | 88.3 (91) / 89.0 (93) | 96.8 (100) / 97.2 (101) | 89.8 (95) / 91.0 (96) | **−8.2 / −7.9** | −8.5 % / −8.1 % | 0.01 / 0.01 |
| before | 8 | | 170.8 (189) / 170.9 (192) | 192.8 (213) / 193.8 (215) | 164.8 (188) / 168.1 (192) | **−23.3 / −22.9** | −12.1 % / −11.8 % | 0.14 / 0.13 |
| after, `1a91ae7f` | 1 | 24.0 / 31.0 | 81.3 (84) / 80.3 (84) | 91.2 (95) / 88.8 (93) | 83.3 (89) / 82.3 (87) | **−9.5 / −8.4** | −10.4 % / −9.5 % | 0.01 / 0.01 |
| after | 8 | | 182.2 (228) / 171.4 (200) | 209.7 (259) / 191.4 (221) | 176.3 (217) / 166.7 (192) | **−28.4 / −20.7** | −13.5 % / −10.8 % | 0.22 / 0.23 |

**3. fallback is still free.** `operator: and` legs (0 % settle → over-fetch, then round 2 anyway):

| Quora 3.2 KB, AND legs | c | fetch ops/req armed / disarmed | armed p50 (p90) ms | disarmed p50 (p90) ms | classic p50 (p90) ms | armed − disarmed, paired p50 | as % of disarmed | armed slower share |
|---|---|---|---|---|---|---|---|---|
| before, `ff50e64e` | 1 | 19.2 / 19.2 | 106.0 (108) / 105.9 (108) | 106.1 (109) / 106.0 (109) | 85.2 (87) / 85.2 (87) | **−0.1 / −0.1** | −0.1 % / −0.1 % | 0.48 / 0.47 |
| before | 8 | | 203.7 (248) / 207.8 (252) | 208.2 (247) / 209.0 (251) | 131.5 (166) / 135.9 (166) | −5.5 / +0.5 | −2.6 % / +0.2 % | 0.46 / 0.51 |
| after, `1a91ae7f` | 1 | 19.2 / 19.2 | 105.1 (108) / 105.3 (108) | 105.3 (108) / 105.2 (108) | 84.6 (86) / 84.5 (86) | **−0.2 / −0.1** | −0.2 % / −0.1 % | 0.46 / 0.48 |
| after | 8 | | 199.8 (241) / 201.6 (250) | 201.2 (245) / 201.5 (245) | 127.8 (161) / 130.8 (164) | −3.6 / +4.1 | −1.8 % / +2.0 % | 0.48 / 0.53 |

bzhangam (approved):

bzhangam:
Approved. But still have a feel that we make this more complicated and user needs to learn more details to understand why sometimes latency is better and sometimes is worse compared to the classic hybrid. Not sure what we can do to reduce this learning curve/confusion other than document it(more details in the documentation may also confuse users).

martin-gaievski:
> Approved. But still have a feel that we make this more complicated and user needs to learn more details to understand why sometimes latency is better and sometimes is worse compared to the classic hybrid. Not sure what we can do to reduce this learning curve/confusion other than document it(more details in the documentation may also confuse users).

I agree, latency is changing depending on the multiple factors, I look at this from different angle - without last 2 PRs resolver performed worse than classic HQ, not it can do better in certain cases, which is by itself a good thing.
What we can do as follow up change is make this information part of the profiler response for resolver, this will make the decision regarding the fast/not fast mode more like a self served.

Issue #866, opened 2024-08-14
[RFC] Sorting in Hybrid Search


vibrantvarun:
# Introduction
This document proposes a solution for enabling the sorting feature in [Hybrid Search](https://opensearch.org/docs/latest/search-plugins/hybrid-search/). The proposal aims to use existing extensions from OpenSearch as much as possible.

# Problem Statement
The problem statement is that Hybrid search results should be in sorted form as per the sort criteria provided in the search request. Since, Hybrid search went GA in 2.11 there are multiple asks from the customers via Github issues to enable different type of features like [Aggregations](https://github.com/opensearch-project/neural-search/issues/604), [Post Filter](https://github.com/opensearch-project/neural-search/issues/508). One such ask is to enable the Sorting in Hybrid Search. Therefore, we are proposing this feature to be released in the upcoming release 2.15.

# Scope
The scope of this document is to enable the support all the [scenarios](https://opensearch.org/docs/latest/search-plugins/searching-data/sort/) through which sorting is supported in OpenSearch. However, there are certain limitations wrt to Hybrid Search which we will discuss in the later sections.

# Current State
### Overview
In the Query Phase of traditional search query like match, term etc  individual shard results are retrieved in a sorted form as per the sort criteria given in the search request. The results are then sent to the Fetch Phase.

### Low level Functioning
Since sorting is performed at the shard level, the collector adds the query result (docId) to the priority queue, which operates based on the sort field comparator.

### Note
1. When sort is applied then Query Phase results will not contain scores as scoring is not done in the Query Phase.  The results will be purely based of sort criteria whether it be any field, docId, _id etc. The scores are calculated in the Fetch Phase only if track_scores=true is sent in the search request. 
2. If track_scores=true then the field relevance score is calculated which says that how relevant that field is to the search result.
3. If track_scores=false then scoring will not be done at all in the entire search process. Consequently, the hits will contain null scores.
4. As per the OpenSearch architecture, Fetch phase will only set scores in the search hits in the following cases
    1. sort criteria is by _score
    2. sort criteria is by any field (not _score) and track_scores=true
    3. If no sorting is applied ( in hybrid query scenario)

## Internal Functioning of Hybrid Search
[...]

PR #2010, opened 2026-09-16
#1984: Allow sort: [_score, <field>] for hybrid queries with collapse

### LINKED ISSUE #1984: [FEATURE] Allow sort: [_score, <field>] for hybrid queries with collapse (deterministic policy-driven group head)

With a hybrid query + collapse, the group head is chosen post-fusion by fused score (CollapseDataCollector). A collapse group often holds several candidate docs that differ by a business dimension - e.g. the same item targeted at different users, distinguished by a field and the application needs a specific candidate to represent the group.

Today there's no way to express that:

- On an exact fused-score tie the head falls back to Lucene doc id - unstable across segment merges, reindexing, and shards, so the representative flips unpredictably with no way to pin it.
- Any attempt to add the deciding field is rejected up front by HybridCollectorManager.validateSortCriteria: `_score sort criteria cannot be applied with any other criteria. Please select one sort criteria out of them.`

Fused-score ties are not rare: RRF is a sum over a small discrete set, so docs sharing a rank profile tie exactly; min-max normalization bunches values similarly. So, the wrong or an unpredictable doc frequently becomes the group head.

Allow sort: [_score desc, \<field> desc] when collapse is present: _score stays primary (relevance order unchanged), and the field decides the representative only on exact fused-score equality.

Note the semantics deliberately chosen here: this is score-primary with a field tiebreaker. A higher-scoring but less-specific doc still wins the head, the field decides ties only. The alternative "most-specific doc always wins regardless of score" is field-primary selection, which is a different mechanism and out of scope here.

#### Why this is a parity gap

For non-hybrid queries, sort: [_score desc, id asc] already gives a deterministic collapse head. Hybrid is the only query type that forbids mixing _score with a field. This closes that gap for the collapse case (score stays populated; matching and primary order unchanged).

#### Scope

Deliberately narrow to the collapse path - it bypasses the score/field collector fork and uses its own `HybridCollapsingTopDocsCollector`, so no changes to `HybridTopScoreDocCollector` /  `HybridTopFieldDocSortCollector`.

The collapse collector already retains sort field values (its FieldValueHitQueue is built from the full sort, and topDocs() extracts all comparator values). The gap is per-sub-query score handling for a multi-key sort: its score path assumes a single-key `_score` (wraps one comparator in `HybridLeafFieldComparator`). A [_score, field] sort takes the multi-comparator branch (plain `MultiLeafFieldComparator`) while isSortByScore is true, so updateExistingEntry / addNewEntry hit:
  
```
assert comparator instanceof HybridLeafFieldComparator;   // AssertionError when assertions enabled
((HybridLeafFieldComparator) comparator)...     // ClassCastException in production (assertions off)
```

So the substantive work is: make the multi-key branch hybrid-aware by wrapping the `_score` sub-comparator (leaving field sub-comparators plain) so it still receives per-sub-query scores. This works because `MultiLeafFieldComparator` propagates setScorer to every child comparator, so a hybrid-wrapped score child transparently feeds the per-sub-query score while field children read doc-values as usual.
  
### Proposed changes

1. `validateSortCriteria` - allow _score-primary + field when collapse is present (only when _score is the first sort key; keep rejecting field-primary + _score, and preserve the track_scores / min_score invariants).
2. `HybridCollapsingTopDocsCollector` - hybrid-wrap the score component in the multi-key path (ensureLeafComparatorsInitialized, updateExistingEntry, addNewEntry).
3. `ScoreCombiner.getDocIdSortFieldsMap` - on a multi-key sort-by-score, replace only the _score slot with the normalized score and preserve the trailing tiebreaker value (previously the whole sort-fields array was collapsed to [score], silently dropping the tiebreaker before it reached the coordinator's collapse comparator).
4. Head tiebreak is then inherited - `CollapseDataCollector.collapseComparator` is already sort-driven and multi-key aware; it becomes correct once Point 3 preserves the field layout.

**Out of scope for this issue:** general non-collapse hybrid [_score, field]

Boolean query instead of Hybrid queries. 

NA

### Ask
Does this direction seem acceptable? If so, I can raise a PR soon.

ercsonusharma:
### Description
[FEATURE] Allow sort: [_score, <field>] for hybrid queries with collapse (deterministic policy-driven group head)

### Related Issues
Resolves #1984 

### Check List


vibrantvarun:
@ercsonusharma score as a sort criteria cannot be applied by any other criteria. That is why it is blocked in collapse as well

check https://github.com/opensearch-project/neural-search/issues/866 
https://opensearch.org/blog/internal-functioning-of-sorting-in-hybrid-query/

#### TLDR:

Hybrid query runs subqueries. 

sorting by _score cannot be combined with other fields like age because subquery results must be combined and normalized based on either a specific sort order or the relevance score, but not both simultaneously.

#### Why the Limitation Exists

##### Conflicting Processing Phases: 
Hybrid queries calculate scores during an initial query phase, normalize them across different scales (like lexical BM25 and neural vector k-NN scores), and combine them. When you apply a custom sort like age, the engine switches to sorting purely by field values during the query phase, bypassing relevance scoring entirely unless track_scores=true is enabled. 

##### Recalculation Mismatch: 
Enabling track_scores=true forces the engine to recalculate scores during the fetch phase. This breaks hybrid search because the newly calculated scores do not reflect the global normalization processor steps that happened earlier in the pipeline, leading to inaccurate results. 

##### Architectural Ambiguity: 
The engine cannot deterministically tie-break or interleave multi-scale normalized hybrid scores with arbitrary field criteria (like numerical or date fields) within the distributed execution framework without breaking the sorted order constraints.

ercsonusharma:
Thanks @vibrantvarun - good points, and you are right about both of the approaches you describe. This proposal takes a third path that avoids both, so I want to clarify the scope. `_score` stays the primary key. The field is a tiebreaker consulted only on an exact fused-score tie, applied post-fusion. It's not a field-mode sort and does not use track_scores.

Scope is deliberately narrow: relaxation is gated on `collapse() != null` and `_score` being the first sort key. `field-primary + _score` stays rejected, non-collapse hybrid unchanged. This just brings hybrid to parity with non-hybrid, where sort: [_score, <field>] already gives a deterministic collapse head which is also the established, expected behavior for field collapsing in other mainstream search engines. Patch + tests are up if you'd like to take a look.

vibrantvarun (commented):
Reviewed across three lenses (logic/line-by-line, adversarial edge-cases, tests) and cross-checked the collapse path against the vendored Lucene MultiLeafFieldComparator and the existing default collector. No correctness bug, no regression. The core fix is right: the pre-PR ClassCastException on multi-key `[_score, <field>]`+collapse (a MultiLeafFieldComparator blind-cast to HybridLeafFieldComparator) is fixed by wrapping only the index-0 SCORE child and tracking it via `scoreComparators[]`. My main regression worry is refuted — the collapse value lives in the separate `collapseValues[]` array, not `fieldDoc.fields`, so single-key `[_score]`+collapse has `fields.length==1` and takes the unchanged branch (byte-identical to main); the clone/swap branch fires only for the newly-allowed shape and is actually required there. Verified: reverseMul applied exactly once, guard scoping airtight (`[<field>, _score]` and no-collapse both stay rejected; score-at-non-primary can't slip through), multi-shard head elected from a single global map.

Actionable items — none blocking, but the first two are worth a look before merge:

1. (MED, correctness) `_score asc` + field + collapse is now accepted, but the collector's min-competitive-score pruning assumes score-descending. `validateSortCriteria` gates on `SortField.Type.SCORE.equals(sortFields[0].getType())` only — it doesn't check the sort order. So `[{_score:{order:asc}}, <field>]`+collapse passes the guard, and the `minScoreThresholds`/`getMinScores()` -> HybridBulkScorer path can then prune docs that are actually competitive under ascending order -> potentially missing/wrong group heads. This is latent for single-key `[_score asc]` today, but this PR widens the accepted surface to include it alongside a tiebreak field. Suggest rejecting `_score` unless descending in the guard, or confirming the pruning path is correct/disabled for ascending.

2. (MED, defense-in-depth) `ScoreCombiner.isSortOrderByScore` returns true when score is *anywhere* in the sort, but `getDocIdSortFieldsMap` hard-swaps index 0 (`sortFields[0] = normalizedScore`). That's correct only because the guard enforces score-primary; if the guard ever loosens (e.g. to allow `[<field>, _score]`+collapse), ScoreCombiner would normalize the wrong slot. Aligning `isSortOrderByScore` (or the swap site) to require `sortFields[0]` be SCORE would keep the two in lockstep. Same 'index-0 assumption held only by the guard' theme as #1.

3. (Tests) Coverage is happy-path-narrow. Highest-value gaps: (a) no `validateSortCriteria` unit test pinning the *rejection* arms — `[<field>, _score]`+collapse rejected, `[_score, <field>]` without collapse rejected — yet that guard is the whole safety boundary this PR relaxes; (b) the multi-shard test sets assertDeterministicHead=false, so the cross-shard elected head isn't oracle-anchored; (c) no `ScoreCombiner` unit test for the new clone/swap branch; (d) the collector 'deterministic head' unit test puts each doc in its own group, so it never exercises intra-group election; (e) missing: keyword/string tiebreak field, null tiebreak value, multi-segment multi-key, and `_score asc`.

4. (LOW, docs) When both `_score` and the tiebreak field tie exactly, the head falls back to doc-id -> not replica-stable (inherent, same as single-key today; this PR strictly improves determinism by adding the tiebreak). And `min_score` + `[_score, <field>]` + collapse stays rejected (a feature gap vs single-key `[_score]`, not a regression). Both worth a one-line note.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -201,9 +201,17 @@ private static void validateSortCriteria(SearchContext searchContext, boolean tr
...
         if (hasScoreSort && hasFieldSort) {
-            throw new IllegalArgumentException(
-                "_score sort criteria cannot be applied with any other criteria. Please select one sort criteria out of them."
-            );
+            // Collapse-gated relaxation: a hybrid query with collapse may mix _score with a field so the
+            // collapse group head is deterministic (field breaks exact fused-score ties). Allowed only when
+            // _score is the primary (first) sort key; field-primary + _score stays rejected.
+            boolean isScorePrimaryWithCollapse = searchContext.collapse() != null
+                && sortFields.length > 0
+                && SortField.Type.SCORE.equals(sortFields[0].getType());
```
vibrantvarun: This guard checks the sort *type* is SCORE but not its *order*, so `[{_score:{order:asc}}, <field>]`+collapse is accepted. The collapsing collector's min-competitive-score pruning (`minScoreThresholds` / `compoundQueryScorer.getMinScores()` -> HybridBulkScorer) assumes score-descending; under `_score asc` it can prune docs that are actually competitive -> potentially missing or wrong group heads. Latent for single-key `[_score asc]` already, but this change widens the accepted surface to include it with a tiebreak field. Suggest rejecting `_score` unless descending here, or confirming the pruning path is correct/disabled for ascending order.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
```diff
@@ -247,13 +247,23 @@ private Map<Integer, Object[]> getDocIdSortFieldsMap(
...
-                        sortFields = new Object[] { combinedNormalizedScoresByDocId.get(fieldDoc.doc) };
+                        Float normalizedScore = combinedNormalizedScoresByDocId.get(fieldDoc.doc);
+                        if (fieldDoc.fields == null || fieldDoc.fields.length <= 1) {
+                            // Single-key [_score] sort (or no field values) — original behavior
+                            sortFields = new Object[] { normalizedScore };
+                        } else {
+                            // Multi-key [_score, field] sort — _score is always the primary key (index 0),
+                            // so preserve the trailing tiebreaker(s) and swap only the score slot
+                            sortFields = fieldDoc.fields.clone();
+                            sortFields[0] = normalizedScore;
```
vibrantvarun, line 260: Defense-in-depth (not a bug today): `isSortOrderByScore` (line 165) returns true when SCORE appears *anywhere* in the sort, but this swap hard-codes index 0. It's correct only because HybridCollectorManager.validateSortCriteria enforces score-primary before this ever runs. If that guard is ever loosened (e.g. to allow `[<field>, _score]`+collapse), this would normalize the wrong slot. Worth making `isSortOrderByScore` — or this swap site — require `sortFields[0]` to be SCORE, so the two stay in lockstep with the guard.

vibrantvarun:
Thanks @ercsonusharma, the third-path framing holds up — I traced it through and agree it escapes the #866 limitation rather than reopening it. That limitation is fundamentally about *field-primary* sorting: scoring is skipped in the query phase, so recovering scores forces `track_scores=true` and a fetch-phase recompute that clobbers normalization (#866 functional limitation). With `_score` primary, hybrid scores in the query phase exactly as it always does, normalization runs untouched, and the field is a stable per-doc value used only as a post-fusion tiebreak. `ScoreCombiner.getDocIdSortFieldsMap` bears this out — it swaps only index 0 to the normalized score and preserves the doc's real field value, so `[fused-score desc, field]` is a coherent ordering and the field value never moves under normalization. Parity with non-hybrid collapse is the right bar.

One residual I'd still like closed (left inline at `HybridCollectorManager` line 209): the guard gates on `sortFields[0]` being type SCORE but not on its order, so `sort: [{"_score":{"order":"asc"}}, <field>]` + collapse passes `isScorePrimaryWithCollapse`. The collector's min-competitive-score pruning (`minScoreThresholds` / `getMinScores()` → `HybridBulkScorer`) assumes score-descending, so ascending could prune genuinely-competitive docs and yield a wrong/missing group head. The third-path argument holds precisely because `_score` is descending; `_score asc` is the one shape that breaks the invariant. Suggest rejecting `_score` unless descending in the guard (or confirming the pruning path is order-safe). Everything else LGTM.

src/main/java/org/opensearch/neuralsearch/processor/combination/ScoreCombiner.java
ercsonusharma: done.

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
ercsonusharma: done.

ercsonusharma:
Thanks for the review @vibrantvarun I have incorporated the suggested changes. I will take up the documentation update in a separate PR.

vibrantvarun (approved):
Re-reviewed the `review comments impl` commit — both of my earlier comments are addressed correctly, verified against the code and Lucene's sort semantics.

- **`_score asc` guard**: `validateSortCriteria` now gates on `HybridSearchCollapseUtil.isScorePrimaryDescendingSort` (`isScorePrimarySort && !sortFields[0].getReverse()`), which correctly rejects `[{_score:{order:asc}}, <field>]` + collapse — `SortField.Type.SCORE` defaults to `reverse=false` (descending), and `_score asc` maps to `reverse=true`. The collector's min-competitive-score pruning stays on its safe score-descending assumption, and `testCreateCollectorManager_whenScoreAscendingWithFieldAndCollapse_thenFail` pins it.
- **`ScoreCombiner` defense-in-depth**: `isSortOrderByScore` now delegates to `isScorePrimarySort` (slot-0 must be SCORE), staying in lockstep with the guard. It's equivalent to the old "score anywhere" check for every reachable hybrid sort (field-primary + `_score` is rejected upstream), so no behavior change on the existing paths.
- Nice call extracting `HybridSearchCollapseUtil` so the guard, `ScoreCombiner`, and the collapse collector share one definition of "score-primary", with unit tests covering score-primary / descending / ascending / field-primary / absent / null.

The design is sound (I confirmed earlier in the thread that the score-primary + post-fusion field-tiebreak carve-out escapes the #866 limitation rather than reopening it), no regression to the single-key `[_score]` + collapse path, and the guard-rejection boundary is now tested. Deferring the doc caveats (tie → doc-id determinism, `min_score` + `[_score, <field>]` + collapse still rejected) to a follow-up PR is fine. LGTM.

vibrantvarun:
@martin-gaievski could you review this PR

ercsonusharma:
@vibrantvarun @martin-gaievski just following up when you get a chance: are we good to merge this? Thanks!

src/main/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollector.java
```diff
@@ -340,19 +355,18 @@ private void updateExistingEntry(HybridSubQueryScorer compoundQueryScorer, int s
...
+                    scoreComparators[subQuery].setCurrentSubQueryScore(score);
                 }
 
                 boolean accepted = false;
                 try {
                     accepted = reverseMuls[subQuery] * comparator.compareBottom(doc) > 0;
                 } finally {
                     if (!accepted && isSortByScore) {
-                        ((HybridLeafFieldComparator) comparator).setCurrentSubQueryScore(scoreOfLastTopEntry);
+                        scoreComparators[subQuery].setCurrentSubQueryScore(scoreOfLastTopEntry);
```
martin-gaievski: 🔴 **Min-score feedback prunes score-tied docs that the new field tiebreaker should keep**

Rule-pack R1/R2 (made-live predicate / threshold propagation). The eviction path propagates the evicted entry's raw score into HybridSubQueryScorer.getMinScores(), and HybridBulkScorer.scoreWindowIntoBitSetWithSubqueryScorers (line 146) drops every later-window doc with `score > minScore` false, i.e. it discards docs whose score EQUALS the threshold. That was a valid optimization for the single-key [_score] sort because the queue also rejects ties (`compareBottom(doc) > 0` is strict), so 'score == evicted' was never competitive. This PR's new [_score desc, <field>] shape makes ties competitive: a doc that ties the bottom's score but wins on the field is accepted by MultiLeafFieldComparator, yet the same doc is filtered out before it reaches the collector if it lives in a later 4096-doc window. Trace (numHits=2, sort [_score desc, price asc], one segment > 4096 docs): window 1 collects d10 (1.0, price 30) and d11 (1.0, 20) -> queue full, bottom d10; d12 (1.0, 10) beats the bottom on price -> accepted, d10 evicted, evictedScore 1.0 -> getMinScores()[0] = 1.0, bottom is now d11 (20). Window 2: d5000 (1.0, price 5) is filtered in HybridBulkScorer (1.0 > 1.0 is false) and never reaches the collector. Correct top-2 under the declared sort is {d12(10), d5000(5)}; actual is {d12(10), d11(20)}. The same doc placed in window 1 would have been kept, so the outcome depends on doc-id layout. Direction table for the touched score-sorted path: FIELD_SCORE single-key -> ties non-competitive, prune is a subset of queue rejection (OK, unchanged); [_score desc, field] -> ties competitive via field, prune drops them (WRONG, new); [_score asc, *] -> rejected by isScorePrimaryDescendingSort (not reachable here; single-key asc is the separate PR #2026 defect). Sibling inventory: HybridTopScoreDocCollector has the same strict feedback but only ever sorts by score (ties non-competitive, OK); SimpleFieldCollector/PagingFieldCollector send no feedback; HybridCollapsingTopGroupsCollector (on main since #1956) deliberately sends no min-score feedback and is the template for the multi-key case. Lucene's own setMinCompetitiveScore keeps ties (skips only score < min), so the drop comes solely from the plugin's strict filter. The new unit tests feed the leaf collector directly and stay under 4096 docs, so this path is never exercised. EMPIRICAL REPRODUCTION (this review, PR head ba336f6c): a scratch test drove HybridCollapsingTopDocsCollector through the real HybridBulkScorer over a single 6000-doc segment where every doc scores 1.0 and integerField = 6000 - docId (best tiebreak values in the second window), sort [_score desc, integerField asc], numHits=2. Result on head: survivors integerField=[1905, 1906] and per-sub-query totalHits=4096 - every second-window doc was pruned, the returned pair is the best of window 1 only, and the count is 6000 -> 4096. With the one-line fix below (propagate only when sort.getSort().length == 1) the same test returns [1, 2] with totalHits=6000, and all 21 existing HybridCollapsingTopDocsCollectorTests still pass.

```java
// Update minScore from the evicted entry's score
if (isSortByScore) {
    minScoreThresholds[subQuery] = Math.max(minScoreThresholds[subQuery], evictedScore);
    compoundQueryScorer.getMinScores()[subQuery] = Math.max(compoundQueryScorer.getMinScores()[subQuery], evictedScore);
}
```

**Why it matters:** Tiebreaker sort makes score-ties competitive -> the evicted score is still propagated as a hard `score > minScore` filter in HybridBulkScorer -> on any segment with more than 4096 matching docs, tied candidates in later windows (including the one with the best tiebreaker value) never enter the per-sub-query top-N -> the coordinator's CollapseDataCollector elects the head from an incomplete, doc-id-order-dependent candidate set, and a group whose only surviving members were pruned loses that sub-query's contribution entirely -> the 'deterministic policy-driven group head' the feature promises is wrong and still non-deterministic across segment merges, exactly the instability issue #1984 set out to remove, and it appears precisely in the tied-score scenario the feature targets (variants with identical text produce identical BM25 scores). The pruning also removes those docs before collect(), so the shard's hits.total is undercounted for the new shape (6000 -> 4096 in the reproduction); the single-key variant of that undercount is the subject of open PR #2026.

**Suggestion:** Only propagate the min-competitive score when the sort is single-key score (ties non-competitive): change the eviction block to `if (isSortByScore && sort.getSort().length == 1) { ... }` (verified in this review: fixes the reproduction and keeps the existing 21 collector tests green) (this mirrors HybridCollapsingTopGroupsCollector, which sends no feedback). Alternatively keep WAND for the multi-key case by propagating `Math.nextDown(evictedScore)` so `score > minScore` and setMinCompetitiveScore both keep exact ties. Update the class Javadoc ('evicted entries propagate their scores as minimum thresholds') to state the multi-key exception. Add a regression test that goes through HybridBulkScorer with a single-segment index of > 4096 tied-score docs (e.g. 5000 docs with identical text and a monotone price field, numHits small) and asserts the survivors are the lowest prices regardless of doc id; a leaf-collector-only fixture cannot catch this.

_[code-review-sage]_

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -201,9 +202,17 @@ private static void validateSortCriteria(SearchContext searchContext, boolean tr
...
         }
         if (hasScoreSort && hasFieldSort) {
-            throw new IllegalArgumentException(
-                "_score sort criteria cannot be applied with any other criteria. Please select one sort criteria out of them."
-            );
+            // Collapse-gated relaxation: a hybrid query with collapse may mix _score with a field so the collapse
+            // group head is deterministic (field breaks exact fused-score ties). Allowed only when _score is the
+            // primary (first) sort key AND descending; field-primary + _score, and ascending _score, stay rejected
+            // (see HybridSearchCollapseUtil#isScorePrimaryDescendingSort for why descending is required).
+            boolean isScorePrimaryWithCollapse = searchContext.collapse() != null
```
martin-gaievski: 🟡 **Relaxed guard now also feeds the distinct-groups collector merged after this PR's base**

Design CONCERNS + cross-change conflict. The PR is based on f15d73ac (2026-09-16). Since then PR #1956 merged into main (ab55fe53, 2026-09-21) and HybridCollectorFactory now routes collapse to HybridCollapsingTopGroupsCollector when index.neural_search.hybrid_collapse_distinct_groups_enabled is set. This guard relaxation therefore admits [_score desc, <field>] into two collectors, but the PR only adapted HybridCollapsingTopDocsCollector and only tests that one. Reading main's TopGroups collector, it is position-generic (isScoreSortField[i], reversed[i], value(k) per slot) and sends no min-score feedback, so it is probably correct for the new shape, but nothing in this PR proves it, and the PR body/#1984 discussion predate its existence. Separately, open PR #2026 (head 8f964a99) rewrites the same constructor hunk (isSortByScore/isSortByScoreDescending) and the collect() skip predicate in HybridCollapsingTopDocsCollector; `git merge-tree` between the two heads reports one textual conflict there, and both PRs' test files touch HybridCollapsingTopDocsCollectorTests. Once #2026 lands, the Javadoc on isScorePrimaryDescendingSort ('pruning ... does not consult the reverse flag') and the identical comment in HybridCollectorManagerTests become stale, and the descending-only restriction is no longer motivated by a bug. A git merge of this PR into current main is clean (0 conflicts), so the rebase itself is cheap.

```java
boolean isScorePrimaryWithCollapse = searchContext.collapse() != null
    && HybridSearchCollapseUtil.isScorePrimaryDescendingSort(sortFields);
```

**Why it matters:** Guard widened for a collector family that grew after the PR was written -> the new sort shape reaches an untested code path selected by an index setting -> a regression there would surface only for users of the opt-in setting, with no test to catch it; and two PRs rewriting the same predicate land in unknown order -> whichever merges second inherits a conflict and a stale justification for the descending-only rule.

**Suggestion:** Rebase onto current main and add one IT (or collector unit test) that runs [_score desc, <field>] + collapse with index.neural_search.hybrid_collapse_distinct_groups_enabled=true, asserting the same tie-broken head. Coordinate with #2026: if it merges first, adopt its isSortByScoreDescending flag instead of re-deriving the direction here and reword the isScorePrimaryDescendingSort Javadoc so the asc restriction is stated as a deliberate scope limit rather than a workaround for the pruning bug; if this PR merges first, leave a note on #2026 about the constructor conflict.

_[code-review-sage]_

src/test/java/org/opensearch/neuralsearch/search/collector/HybridCollapsingTopDocsCollectorTests.java
```diff
@@ -475,6 +475,247 @@ public void testCollapse_whenSortByField_thenCorrectOrdering() throws IOExceptio
...
+    }
+
+    /**
+     * on an exact score tie, the trailing field must break the tie deterministically.
+     *
+     * <p>Every doc shares the same per-sub-query score, so the primary key is fully tied. The sort
+     * is {@code [_score DESC, integerField ASC]}, so the collapse collector must fall back to the
+     * integer field (ascending) instead of an unstable doc-id order — giving a deterministic group head.
+     *
+     * <p><b>TDD status: RED.</b> Fails until multi-key score handling is fixed in the collector.
```
martin-gaievski: 🟡 **Tie tests never cross the 4096-doc window and two new tests pass unchanged on main**

Rule-pack R4. Discriminating (fail on merge-base by construction): the three new collector tests (ClassCastException/AssertionError in addNewEntry on base), the ScoreCombiner tiebreak test (fields.length 1 vs 2 on base), the util tests (methods absent on base) and the IT (HTTP 400 on base). Non-discriminating (pass on base): testCollectCollapseData_whenScoreTieWithinGroup_thenFieldElectsDeterministicHead (CollapseDataCollector is unchanged) and testCreateCollectorManager_whenScoreAscendingWithFieldAndCollapse_thenFail (base already throws the same message for any score+field mix) - these document inherited behaviour and must not be counted as coverage of the change. Gaps: (1) no test drives the collector through HybridBulkScorer with more than 4096 matching docs, which is the only way to observe the min-score feedback (see the red finding); every fixture here is 10-50 docs fed straight to the leaf collector. (2) No positive unit test that validateSortCriteria ACCEPTS [_score desc, <field>] with collapse - the relaxation itself is only covered end-to-end. (3) No negative test that [_score desc, <field>] WITHOUT collapse is still rejected; the existing HybridQuerySortIT case is field-primary ([stock, _score]), so the collapse gate has no direct guard. (4) The multi-shard IT variant only asserts acceptance and _score-desc ordering, so cross-shard tiebreak is unverified. (5) Three Javadocs still read 'TDD status: RED ... fails until ... is fixed', describing a pre-fix state that is no longer true. Executed in this review: on head ba336f6c the four touched unit-test classes pass (54 tests: ScoreCombinationTechniqueTests 11, CollapseDataCollectorTests 10, HybridCollapsingTopDocsCollectorTests 21, HybridSearchCollapseUtilTests 12). Copied onto the merge-base f15d73ac, exactly the four new collector/ScoreCombiner tests FAIL and the new CollapseDataCollector test PASSES (42 run, 4 failed) - confirming the discrimination split above. A 6000-doc HybridBulkScorer-driven variant of the tie test fails on head (see the red finding), demonstrating the gap. CI workflows on head are all 'action_required' and have not run, so there is no observable green gate for the IT, spotless or the changelog verifier (spotlessJavaCheck run locally: all 10 PR files clean).

```java
 * <p><b>TDD status: RED.</b> Fails until multi-key score handling is fixed in the collector.
 */
public void testCollapse_whenScoreTie_thenFieldBreaksTieDeterministically() throws IOException {
```

**Why it matters:** The suite is green for the shipped shape while the one mechanism that can silently drop competitive docs is untested -> the red defect ships undetected -> a later revision has to re-open the collector; the stale RED comments mislead the next reader into thinking the tests are expected to fail.

**Suggestion:** Add a HybridBulkScorer-driven test (real HybridQuery over a single-segment index with >4096 tied-score docs) asserting the top-N by tiebreaker; add a HybridCollectorManagerTests case that [_score desc, field]+collapse creates a collector without throwing and one that the same sort without collapse still throws; strengthen the multi-shard IT to assert the tie-broken head (use a single-node, deterministic-IDF fixture or a keyword tiebreak on both tied docs); delete the 'TDD status: RED' paragraphs; describe the two base-passing tests as invariant guards, not regression tests for the fix.

_[code-review-sage]_

src/main/java/org/opensearch/neuralsearch/search/query/HybridCollectorManager.java
```diff
@@ -201,9 +202,17 @@ private static void validateSortCriteria(SearchContext searchContext, boolean tr
...
+                && HybridSearchCollapseUtil.isScorePrimaryDescendingSort(sortFields);
+            if (!isScorePrimaryWithCollapse) {
+                throw new IllegalArgumentException(
```
martin-gaievski: 🟡 **Rejection message still says _score can never mix with a field, which is no longer true**

The same pre-relaxation message is reused for three now-distinct rejections: [_score, field] without collapse, [field, _score] with collapse, and [_score asc, field] with collapse. For the third case in particular the user has done everything the feature asks except the sort order, and the error tells them the combination is impossible rather than that ascending _score is unsupported. The IT and unit test both assert this exact string, so changing it requires touching those assertions.

```java
if (!isScorePrimaryWithCollapse) {
        "_score sort criteria cannot be applied with any other criteria. Please select one sort criteria out of them."
    );
}
```

**Why it matters:** Misleading error -> a user who read the docs/issue for the new capability cannot tell which rule they violated (order, key position, or missing collapse) -> support round-trips and the feature appears broken.

**Suggestion:** Branch the message: when collapse is present and slot 0 is SCORE but reversed -> '_score must be sorted descending when combined with a field in a hybrid query with collapse'; when collapse is present but _score is not the first key -> '_score must be the primary sort key when combined with a field in a hybrid query with collapse'; otherwise keep the existing text but append 'unless the query uses collapse with _score as the primary, descending key'. Update the two test assertions accordingly.

_[code-review-sage]_

martin-gaievski:
One thing I forgot to mention - please add changelog entry for this PR

PR #2020, opened 2026-09-17
Fused fast path profile

martin-gaievski:
### Description
Follow up to #2008, addressing the review comment that fused mode's latency varies. 

Two changes I'm adding here:
- profiler now reports the fast-path verdict. `debug.fast_path` in the `[coordinator][fused:hybrid_N]` entry says whether the same request without `profile` would take the fast path on this coordinator. If it's not, profiler output has first reason in check order
- fetch budget is a dynamic cluster setting.`plugins.neural_search.hybrid.fusion.fast_path_fetch_budget`, default `1mb` (previous constant), range is `0b`–`1gb`.

### Examples of new profiler responses

Cold: the first `_source`-returning request on a fresh index, nothing observed yet on this coordinator.
```json
"fast_path": {
  "would_take": false,
  "refused_by": "source_size_unobserved",
  "detail": "no response with this _source filter has been observed for the index on this coordinator yet; the two-round page is the first observation",
  "fetch_budget_bytes": 1048576
}
```

same request again (profiled response taught the size), `track_total_hits: false`:
```json
"fast_path": {
  "would_take": true,
  "fetch_estimate_bytes": 27645,
  "fetch_budget_bytes": 1048576,
  "count_settled": true
}
```

Default totals on a small corpus — legs decide, after they answer:
```json
"fast_path": {
  "would_take": false,
  "refused_by": "count_not_settled",
  "detail": "the request wants a count beyond the window and no leg reported enough matches to prove it",
  "fetch_estimate_bytes": 27645,
  "fetch_budget_bytes": 1048576,
  "count_settled": false
}
```

`track_total_hits: true`:
```json
"fast_path": { "would_take": false, "refused_by": "exact_totals", "detail": "track_total_hits: true needs the Tail's exact count" }
```

leg carrying `_name`:
```json
"fast_path": { "would_take": false, "refused_by": "leg_name", "detail": "leg 0 carries _name; round 2 registers it against every returned document" }
```

request with aggregation:
```json
"fast_path": { "would_take": false, "refused_by": "request_shape", "detail": "aggregations need round 2 over the fused ranking" }
```

fused hybrid nested inside a `bool`:
```json
"fast_path": { "would_take": false, "refused_by": "nested_hybrid", "detail": "only the request's own top-level hybrid can have its page assembled" }
```

`size: 20` with 8 documents ranked:
```json
"fast_path": {
  "would_take": false,
  "refused_by": "page_beyond_window",
  "detail": "the page ends at 20 but only 8 documents were ranked",
  "fetch_estimate_bytes": 22800,
  "fetch_budget_bytes": 1048576,
  "count_settled": true
}
```

### Related Issues
#1930

### Check List


vibrantvarun (commented):
Reviewed across three lenses (logic/regression-equivalence, adversarial profiled-twin, tests), grounding each against the head-SHA source. No correctness bug, no regression — the PR is a clean observability + settings change.

The refactor that folds the fast-path checks into one `FastPathDecision` is behavior-equivalent to #2008 on every input: the before-legs gate and the after-legs fallback map 1:1 to the old inline `requestShapeAllowsFastPath` / `legsAllowFastPath` / pipeline / budget checks and the old `pageFits || totalsSettled` fallback (the OR/AND semantics are preserved; check order only picks the reported reason). The one intentional divergence — re-reading request shape in the arming path rather than trusting consumer presence — strictly *narrows* arming to shapes the fast path cannot serve (e.g. a search-request-processor that injects an aggregation after the filter), so it can only cost latency, never arm where the old code didn't. The central profiled-twin claim also holds in every shape I could construct: `would_take` equals what the same request without `profile` would actually do (same identity test for the root hybrid, same `requestShapeFastPathRefusal` with only `profile` masked, same `items`/`ranked` read after the legs). Default budget (1 MB) is byte-identical to the retained constant.

Actionable items, all non-blocking (two MED test gaps + nits inline):

LOW (doc wording) — for a profiled top-level fused hybrid that a *shape* feature refuses (e.g. `profile` + `sort`), the report says `refused_by: request_shape`, whereas the unprofiled twin's own rewrite would internally tag `nested_hybrid` (its filter gate refuses on shape, so `rootHybrid` is never set and `decideFastPathBeforeLegs` short-circuits to `nested_hybrid` before it checks the shape). `would_take` is `false` in both, and `request_shape` is arguably the *truer* reason — but it contradicts `FastPathDecision`'s javadoc claim that "every condition ... is evaluated exactly as the unprofiled request evaluates it." Worth a one-line footnote, not a code change.

LOW (nit) — `legsAllowFastPath` is now orphaned in production. Arming inlines the leg checks (`carriesQueryName` loop + `innerHitsLegs(...).isEmpty()`) inside `decideFastPathBeforeLegs`, so the helper is reachable only from a `{@link}` javadoc and its own unit test `testLegsAllowFastPath` — which now exercises dead code duplicating what `testDecideFastPathBeforeLegs` covers. Either re-point `decideFastPathBeforeLegs` at the helper (one source of truth for the leg checks) or drop the helper and its test.

LOW (nit) — `requestShapeFastPathRefusal` returns a bare `String[]{reason, detail}` that callers/tests index positionally (`[0]`/`[1]`); a small record/pair would make it self-documenting. And `FetchVolume.extraBytes()` (`extraDocuments * perDocument.bytes()`) is a `long` product that is theoretically overflowable but not reachable under the enforced `window`/`legs` bounds (~1e11 vs Long.MAX ~9.2e18) — no fix needed, just noting it's untested.

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
```diff
@@ -719,10 +738,17 @@ private QueryBuilder doRewriteFused(QueryRewriteContext queryRewriteContext) thr
...
+        //
+        // The checks are one verdict (FastPathDecision) so that what arms the fast path and what the profile reports
+        // about it are the same evaluation. The request's shape is re-read here rather than trusted from the consumer's
+        // presence, for the reason given for totals above: a search request processor that adds, say, an aggregation
+        // between the filter and this rewrite must keep round 2, whose aggregations the assembled page cannot carry.
+        // The consumer's presence stays the arming condition — it is what says this hybrid is the request's own query
+        // and that the filter's wrapper is there to swap the page in. A profiled request has no consumer (profile keeps
+        // two rounds) but may carry fastPathReportRoot, so its verdict is still evaluated, for the report alone.
+        boolean rootHybrid = Objects.nonNull(fusedHitsConsumer) || fastPathReportRoot;
+        FastPathDecision fastPath = HybridFusionOrchestrator.decideFastPathBeforeLegs(searchRequest, legs, window, rootHybrid);
```
vibrantvarun, line 750: MED test gap. This is the one genuinely *new* arming behavior in the PR: arming now re-reads request shape via `decideFastPathBeforeLegs` (`requestShapeFastPathRefusal(source, true)`) instead of trusting the consumer's presence, so a search-request-processor that mutates the source after the filter attached the consumer — e.g. injects an aggregation or a sort — correctly flips the fast path off. The equivalence tests cover the old refusal set, but no test drives this specific case (consumer present + a shape that becomes disallowed only *after* the filter). Worth a test: it's the single new safety guarantee here, and it's currently unverified.

src/main/java/org/opensearch/neuralsearch/query/ReturnedEmbeddingFields.java
```diff
@@ -74,11 +76,58 @@ final class ReturnedEmbeddingFields {
...
+                return Long.MAX_VALUE;
+            }
+            // extraDocuments is at most legs × window (tens of thousands) and bytes per document is bounded by the
+            // observed size of a document, so the product fits comfortably.
+            return extraDocuments * perDocument.bytes();
+        }
+
+        /** Refuse when unknown (fail closed) or over budget. */
+        boolean exceedsBudget() {
+            return perDocument.isUnknown() || extraBytes() > budgetBytes;
```
vibrantvarun, line 134: MED test gap. `exceedsBudget()` uses strict `>`, so `extraBytes() == budgetBytes` must ARM. The tests only straddle the boundary (1.33 MB over a 1 MB budget refuses; 630 KB under passes) — none construct the exact-equal case, so a `>`→`>=` change would pass every existing test while silently refusing a request that fits exactly. Add a case where `extraDocuments * perDocument.bytes()` equals `budgetBytes` and assert the page still assembles.

martin-gaievski: I think that's not entirely accurate. strict `>` reading is right, and it is right that no nonzero exact-equal case exists. But the mutation argument does not hold: boundary is already pinned at the degenerate point, and a `>`=>`>=` change fails an existing test.

`HybridQueryFusedFanOutTests:1661-1680` sets `plugins.neural_search.hybrid.fusion.fast_path_fetch_budget` to `0b` and then sends a `_source: false` request. `_source: false` yields `PerDocumentEstimate.NOTHING` (`bytes = 0`, `unknown = NONE`), so it is *not* short-circuited by `isUnknown()` and reaches the comparison with `extraBytes() == 0` and `budgetBytes == 0`:

```java
assertNotNull("but a request that fetches nothing is never refused on volume", assembled[0]);  // :1680
```

Under `>=`, `0 >= 0` refuses, nothing is assembled, and that assertion fails. `ReturnedEmbeddingFieldsTests` additionally pins that `extraBytes()` performs the multiplication rather than returning early (`assertEquals(0L, nothing.extraBytes())` with `extraDocuments == 190`).

Another note: this gate is documented as a latency predictor, not a correctness condition
(`ReturnedEmbeddingFields`, both paths return the same page), and its per-document term is an
EWMA of observed page means (`ObservedSourceSizes`). A one-byte difference at the boundary can only
pick the slower of two paths that return an identical response; there is no "silently refusing" outcome a
user can observe beyond latency.

Happy to add a nonzero exact-equal assertion for its documentary value — two lines, in
`testFastPathFetchVolume_thenTheGateAndTheProfileReadOneCalculation` where the rest of the boundary
arithmetic lives:

```java
// exactly at the budget fits: the gate refuses only what is over it (guards a `>` → `>=` drift at a nonzero
// point; the degenerate 0 == 0 point is pinned by HybridQueryFusedFanOutTests' zero-budget `_source: false` arm)
ObservedSourceSizes.clear();
observe(VECTOR_INDEX, source(), 8_192);
ReturnedEmbeddingFields.FetchVolume exactlyAtBudget = ReturnedEmbeddingFields.fastPathFetchVolume(request(source().size(72)), 2, 100);
assertEquals(128, exactlyAtBudget.extraDocuments());                             // 2 × 100 − 72
assertEquals(exactlyAtBudget.budgetBytes(), exactlyAtBudget.extraBytes());       // 128 × 8192 == 1 << 20
assertFalse("a volume exactly at the budget is not over it", exactlyAtBudget.exceedsBudget());
```

src/main/java/org/opensearch/neuralsearch/query/HybridQueryBuilder.java
martin-gaievski: ack, and by incident I've added such test to next PR in fused pipeline https://github.com/opensearch-project/neural-search/blob/5fddf6bb50b21bf7f6df04f4a718d00426b1e4f5/src/test/java/org/opensearch/neuralsearch/query/HybridQueryFusedFanOutTests.java#L1593-L1614.

martin-gaievski:
> Reviewed across three lenses (logic/regression-equivalence, adversarial profiled-twin, tests), grounding each against the head-SHA source. No correctness bug, no regression — the PR is a clean observability + settings change.
> 
> The refactor that folds the fast-path checks into one `FastPathDecision` is behavior-equivalent to #2008 on every input: the before-legs gate and the after-legs fallback map 1:1 to the old inline `requestShapeAllowsFastPath` / `legsAllowFastPath` / pipeline / budget checks and the old `pageFits || totalsSettled` fallback (the OR/AND semantics are preserved; check order only picks the reported reason). The one intentional divergence — re-reading request shape in the arming path rather than trusting consumer presence — strictly _narrows_ arming to shapes the fast path cannot serve (e.g. a search-request-processor that injects an aggregation after the filter), so it can only cost latency, never arm where the old code didn't. The central profiled-twin claim also holds in every shape I could construct: `would_take` equals what the same request without `profile` would actually do (same identity test for the root hybrid, same `requestShapeFastPathRefusal` with only `profile` masked, same `items`/`ranked` read after the legs). Default budget (1 MB) is byte-identical to the retained constant.
> 
> Actionable items, all non-blocking (two MED test gaps + nits inline):
> 
> LOW (doc wording) — for a profiled top-level fused hybrid that a _shape_ feature refuses (e.g. `profile` + `sort`), the report says `refused_by: request_shape`, whereas the unprofiled twin's own rewrite would internally tag `nested_hybrid` (its filter gate refuses on shape, so `rootHybrid` is never set and `decideFastPathBeforeLegs` short-circuits to `nested_hybrid` before it checks the shape). `would_take` is `false` in both, and `request_shape` is arguably the _truer_ reason — but it contradicts `FastPathDecision`'s javadoc claim that "every condition ... is evaluated exactly as the unprofiled request evaluates it." Worth a one-line footnote, not a code change.
> 
> LOW (nit) — `legsAllowFastPath` is now orphaned in production. Arming inlines the leg checks (`carriesQueryName` loop + `innerHitsLegs(...).isEmpty()`) inside `decideFastPathBeforeLegs`, so the helper is reachable only from a `{@link}` javadoc and its own unit test `testLegsAllowFastPath` — which now exercises dead code duplicating what `testDecideFastPathBeforeLegs` covers. Either re-point `decideFastPathBeforeLegs` at the helper (one source of truth for the leg checks) or drop the helper and its test.
> 
> LOW (nit) — `requestShapeFastPathRefusal` returns a bare `String[]{reason, detail}` that callers/tests index positionally (`[0]`/`[1]`); a small record/pair would make it self-documenting. And `FetchVolume.extraBytes()` (`extraDocuments * perDocument.bytes()`) is a `long` product that is theoretically overflowable but not reachable under the enforced `window`/`legs` bounds (~1e11 vs Long.MAX ~9.2e18) — no fix needed, just noting it's untested.

ack on all three findings, will push commit with requested changes

bzhangam (approved):

vibrantvarun (approved):
Re-reviewed the `Address review nits` commit — everything I raised is handled, verified against the diff.

- **`requestShapeFastPathRefusal` now returns a `FastPathDecision.Refusal` record** instead of the positional `String[]{reason, detail}`, with every call site and `refuse(Refusal)` wired through — the self-documenting shape I was after.
- **The orphaned `legsAllowFastPath` is gone**, and I confirmed there are no dangling references (the `{@link}` was removed too), so no dead code and no compile risk; arming inlines the leg checks.
- **The `refused_by`/`would_take` divergence is now documented** in `FastPathDecision` — not just the label footnote I suggested, but also the deeper case where a request processor *removes* the feature the filter refused on and the report can read `would_take: true` for a request that runs two rounds (core's `script` processor). Good, honest documentation of the one seam.
- **`extraBytes()` overflow** is now bounded explicitly in the comment (MAX_FUSION_LEG_SEARCHES cap + the window ceiling), so the product is provably orders of magnitude below `Long.MAX_VALUE`.

On the fetch-budget boundary: @martin-gaievski is right and I was wrong — a `>` → `>=` drift is already caught, at the degenerate `0 == 0` point, by the zero-budget `_source: false` arm in `HybridQueryFusedFanOutTests` (`0 >= 0` would refuse and fail its `assertNotNull(assembled)`), and the gate is a latency predictor where both paths return the same page, so there's no user-observable "silent refusal" beyond latency. The nonzero exact-equal assertion added to `testFastPathFetchVolume...` (128 × 8192 == 1 MB → `exceedsBudget()==false`) closes it at a nonzero point too. The one new arming behavior (shape re-read after the filter) is exercised by the disarm test in the follow-up #2021, which I've reviewed.

No new issues and no regression — the `Refusal` record is a behavior-identical type-safe refactor, and the rest is dead-code removal plus docs and tests. LGTM.

PR #2021, opened 2026-09-18
[Fusion] Fixed incorrect filtering and runtime error when pipeline with processors is attached

martin-gaievski:
### Description
Fixed two defects related to scenario with attached pipeline request processor:
1. `filter_query` was silently bypassed on the fast path
2. Runtime exception/HTTP 500 response `topDocs already consumed` with a hybridization pipeline attached

Fixes are following:
1.  re-derive position and shape at the rewrite, on the request core is executing — `isRequestQuery = searchRequest.source().query() == this` (the identity test the filter applied to the submitted tree) and `requestShapeAllowsFastPath` re-read — and gate the fast path, the leg explain and the totals-consumer hand-off on them. A processor-nested hybrid then behaves exactly like a user-nested one: consumers attached but never fed, which every merger already treats as "nothing to merge". The identity test make sense here because `doRewrite` dispatches straight to `doRewriteFused` on the original instance and later passes exit through `fusedSupplier` before this code runs.
2. skip null and consumed results in `isHybridQuery`, this is classic-processor fix and belongs on `main` as well.

### Before / after request and response, same requests

Fixture: index `tenants`, 12 shards, 12 documents `"quarterly report number N for the sales team"`, `tenant` = `acme` for even N, `globex` for odd N. Pipelines:

```json
PUT _search/pipeline/tenant-acme
{"request_processors":[{"filter_query":{"query":{"term":{"tenant":"acme"}}}}]}

PUT _search/pipeline/legacy-norm
{"phase_results_processors":[{"normalization-processor":{"normalization":{"technique":"min_max"},"combination":{"technique":"arithmetic_mean"}}}]}
```

**A. tenant-scoped caller, fast-path shape**

```json
POST tenants/_search?search_pipeline=tenant-acme
{"size":10,"track_total_hits":false,"_source":false,
 "query":{"hybrid":{"fusion":{"window_size":50,"normalization":{"technique":"min_max"},"combination":{"technique":"arithmetic_mean"}},
                    "queries":[{"match":{"title":"quarterly report"}},{"match":{"title":"sales team"}}]}}}
```

| | response |
|---|---|
| before | `200` — **10 hits: `7, 5, 2, 6, 9, 1, 11, 8, 10, 12`** — five of them `globex` documents the filter must exclude; no `Warning` header |
| after | `200` — **6 hits: `2, 6, 8, 4, 10, 12`** — the `acme` documents only, identical to the two-round path |

**A2. A filter matching no document** (`filter_query: {"term":{"tenant":"nobody"}}`, same request): before **10 hits**, the same page as with no filter; after **0 hits**.

**B. Same filter, two-round request with default totals** (a named leg keeps round 2): before and after both `200`, 6 hits, `total {6, eq}` — on this 12-document index the count settles inside the window. On a corpus above the default threshold the before-state returned **0 hits with `total {10000, gte}`** (the legs' unfiltered count); after, the total is the filtered count (`HybridQueryFusedModeFastPathIT` asserts `{0, eq}`).

**C. The user's normalization pipeline still attached, fast-path shape** (`max_concurrent_shard_requests=1` makes the null shard result deterministic; without it, 11–12 of 12 attempts failed on the live 12-shard index)

```json
POST tenants/_search?search_pipeline=legacy-norm&max_concurrent_shard_requests=1
{"size":5,"track_total_hits":false,"_source":false,
 "query":{"hybrid":{"fusion":{"window_size":50,"normalization":{"technique":"min_max"},"combination":{"technique":"arithmetic_mean"}},
                    "queries":[{"match":{"title":"quarterly report"}},{"match":{"title":"sales team"}}]}}}
```

| | response |
|---|---|
| before | `500 search_phase_execution_exception` → `illegal_state_exception: topDocs already consumed` |
| after | `200` — 5 hits `7, 5, 2, 6, 9`, the same page on every attempt |

### Related Issues
#1930 

### Check List


src/main/java/org/opensearch/neuralsearch/processor/AbstractScoreHybridizationProcessor.java
```diff
@@ -152,9 +152,17 @@ <Result extends SearchPhaseResult> boolean shouldSkipProcessor(SearchPhaseResult
      */
     @VisibleForTesting
     boolean isHybridQuery(final SearchPhaseResult searchPhaseResult) {
+        // A shard whose request was built after the first shard answered may reply with QuerySearchResult.nullInstance()
+        // when its query cannot match (core's can-match shortcut for a match_none rewrite — the fused fast path's round 2,
+        // or any plain query rewritten to match_none), and a partial reduce may already have consumed a shard's topDocs.
+        // Neither carries anything to read: QuerySearchResult#topDocs() throws on both, so they are simply not hybrid.
+        if (Objects.isNull(searchPhaseResult.queryResult())
+            || searchPhaseResult.queryResult().isNull()
+            || searchPhaseResult.queryResult().hasConsumedTopDocs()) {
```
vibrantvarun, line 168: This guard is correct for the `shouldSkipProcessor` scan, but the comment above ("a partial reduce may already have consumed a shard's topDocs … so they are simply not hybrid") overclaims: for the consumed-but-not-`isNull()` case it only prevents the throw *here*. When such a result coexists with a readable hybrid, `shouldSkipProcessor`'s `noneMatch` returns false and the processor runs — and `NormalizationProcessorWorkflow.filterValidResults` (~line 417) filters only `Objects.nonNull(r) && !r.isNull()`, NOT `hasConsumedTopDocs()`, so the consumed survivor reaches `getQueryTopDocs` (~line 224) / `unprocessedDocIds` (~line 400) / `getTopDocsForCompoundQuery` (~line 349), all of which call `topDocs()` → the same `IllegalStateException("topDocs already consumed")`, just moved downstream. Grounded in core `QuerySearchResult`: `isNull` is a final flag set only by `nullInstance()`; `consumeTopDocs()` nulls `topDocsAndMaxScore` but leaves `isNull()==false`, so a partial-reduce-consumed result has `hasConsumedTopDocs()==true` and `topDocs()` throws. The null-instance/`match_none` sub-case the IT exercises is fully fixed (`isNull()==true` → filtered by `filterValidResults`); the >512-shard consumed sub-case that this comment (and `filterValidResults`'s own comment) claims to handle is not. Suggest mirroring this exact 3-part guard in `filterValidResults` (and not assuming every survivor has readable topDocs), or narrowing the claim in both comments. Separately: if *every* hybrid shard result is consumed at processing time, `noneMatch` becomes true and the processor is skipped, returning raw un-normalized scores silently — pre-PR that state threw the 500; worth confirming it isn't reachable, since silent wrong scores are worse than an error.

vibrantvarun (commented):
Reviewed across three lenses (logic/correctness, adversarial edge, tests), grounded against the head-SHA source plus the OpenSearch core `QuerySearchResult`/`QueryPhaseResultConsumer` these build on.

**Both headline fixes are correct.** Fix 1 (re-derive `isRequestQuery = source().query() == this` + re-read `requestShapeAllowsFastPath`): the identity test is valid at the round-1 rewrite (`this` is the original submitted instance; round 2 returns early via `fusedSupplier`), it agrees with the filter's own `found.get(0) == source.query()`, and when a request processor wraps (`filter_query` → `bool{must:[hybrid],filter}`) or reshapes (min_score) the request, all three consumers disarm and the hybrid keeps its two-round Tail so the enclosing filter intersects the true match set — no page assembled, no leg-derived total, legs unexplained. I also checked the deliberately-safe cases: a processor that sets `post_filter` while keeping `source().query()==this` is fine because legs inherit `post_filter` via `CandidateScope`, so the assembled page is already filtered; a processor that replaces `query()` with a different hybrid instance leaves that instance without consumers → clean two-round, no NPE. Fix 2's guard short-circuits correctly (null → isNull() → hasConsumedTopDocs() before `topDocs()`), and because `isHybridQuery` runs inside `shouldSkipProcessor`'s `noneMatch` (not as a shard filter) a null/consumed result beside a readable hybrid does not hide it — the match_none/null-instance 500 the IT exercises is genuinely fixed.

**One thing I'd flag before claiming the partial-reduce case is closed (MED, see inline):** the new `isHybridQuery` comment states a partial reduce that consumed a shard's topDocs is "simply not hybrid" and thus handled — but that's only true for the `shouldSkipProcessor` scan. If a consumed-but-not-`isNull()` result sits *beside* a readable hybrid, `shouldSkipProcessor` correctly returns false and the processor RUNS, and `NormalizationProcessorWorkflow.filterValidResults` (line ~417) filters only `!isNull()`, NOT `hasConsumedTopDocs()`, so that survivor reaches `getQueryTopDocs` (line ~224) / `unprocessedDocIds` (line ~400), which call `topDocs()` → the same `IllegalStateException("topDocs already consumed")`, just relocated downstream. Grounded in core: `isNull` is a final flag set only by `nullInstance()`; `consumeTopDocs()` nulls `topDocsAndMaxScore` but not `isNull`, so a consumed result has `isNull()==false, hasConsumedTopDocs()==true, topDocs() throws`. Both the pre-existing `filterValidResults` comment and this PR's new comment claim to cover the >512-shard partial-reduce case; only the null-instance sub-case actually is. Fix: mirror the same 3-part guard in `filterValidResults` (and don't assume every survivor has readable topDocs), or narrow the comments' claim.

**PLAUSIBLE (MED if reachable, and a NEW behavior): all-consumed → silent wrong scores.** If *every* hybrid shard result is consumed at processing time (shard count an exact multiple of the effective batch size → empty tail buffer), `noneMatch(isHybridQuery)` is true → the processor SKIPS → the response carries raw, un-normalized/un-combined scores with no error. Pre-PR this same state threw the 500; post-PR it may silently return wrong scores, which is worse. I couldn't confirm the empty-tail-at-processor-time reachability without a large-shard cluster, but the mechanism is confirmed — worth verifying.

**Test gaps (MED):**
1. The strongest proof of Fix 1's totals-nulling under a real `filter_query` is missing: the IT asserts the derived total only for the *empty* filter (`{0, eq}`), which is weak (many bugs also yield 0). A non-empty case — the single-doc filter under the counting shape asserting `total {1, eq}` vs the legs' ~12 union — would actually discriminate that the filtered count is used and the leg-derived total was disabled.
2. `testFastPath_whenAHybridizationPipelineIsAttached` has only a status-200 + page-stability oracle (no content-correctness check against a control), and unlike the `filter_query` IT it does not `primeThroughPipeline`, so if the fast path never arms (cold gate) round 2 is never `match_none`, no null-instance is produced, and the test passes green without exercising the skip path. Add priming + a correctness oracle (or an assertion that a null-instance shard was actually produced).
3. The Fix-2 guard lives on the shared `AbstractScoreHybridizationProcessor`, but only `RRFProcessorTests` covers `shouldSkipProcessor`; `NormalizationProcessorTests` has none. A mirrored unit test would make the "belongs on main too" claim explicit and guard a future override.

**LOW/nits:** no unit test for a processor that replaces `query()` with a *different* fused hybrid (same `!= this` branch, minor); the added `import static org.mockito.Mockito.mock;` is out of alphabetical order (after `never`/`verify`) and may trip a spotless/import-order check; `ids(...)` uses an inline fully-qualified `java.util.stream.Collectors.toList()` instead of an import.

src/main/java/org/opensearch/neuralsearch/processor/AbstractScoreHybridizationProcessor.java
martin-gaievski: Agree, comment overclaims for classic results, and I'll narrow it: the guard makes null/consumed results *non-hybrid for the skip decision*; it does not make a partially-reduced classic result processable. Two things change the picture. 

First, fused results are never hybrid-formatted, so skipping is the complete answer for them. And guard also fixes the pre-existing partial-reduce 500 for any non-hybrid query behind a hybridization pipeline (measured with`batched_reduce_size=2` on 12 shards: fused two-round, fused fast path and plain `match` all went 500×8 → 200×8; classic was 200×8 on both builds). 

Second, a classic result consumed before the processor runs is not reachable in a supported configuration: `HybridQuerySearchRequestFilter` forces `batched_reduce_size=MAX_VALUE` whenever the submitted top-level query is a classic hybrid, for exactly this reason. 

The remaining corner is a classic hybrid a custom request processor creates after the filter (core's `script` processor refuses to set `query`). There I'd rather not mirror the guard in`filterValidResults`: a consumed classic result's documents are already in the partial-reduce TopDocs, un-normalized and sentinel-formatted, so dropping it from the workflow would combine normalized and raw docs into a silently wrong page.

I would made that corner fail fast with a clear message instead

martin-gaievski:
> Reviewed across three lenses (logic/correctness, adversarial edge, tests), grounded against the head-SHA source plus the OpenSearch core `QuerySearchResult`/`QueryPhaseResultConsumer` these build on.
> 
> **Both headline fixes are correct.** Fix 1 (re-derive `isRequestQuery = source().query() == this` + re-read `requestShapeAllowsFastPath`): the identity test is valid at the round-1 rewrite (`this` is the original submitted instance; round 2 returns early via `fusedSupplier`), it agrees with the filter's own `found.get(0) == source.query()`, and when a request processor wraps (`filter_query` → `bool{must:[hybrid],filter}`) or reshapes (min_score) the request, all three consumers disarm and the hybrid keeps its two-round Tail so the enclosing filter intersects the true match set — no page assembled, no leg-derived total, legs unexplained. I also checked the deliberately-safe cases: a processor that sets `post_filter` while keeping `source().query()==this` is fine because legs inherit `post_filter` via `CandidateScope`, so the assembled page is already filtered; a processor that replaces `query()` with a different hybrid instance leaves that instance without consumers → clean two-round, no NPE. Fix 2's guard short-circuits correctly (null → isNull() → hasConsumedTopDocs() before `topDocs()`), and because `isHybridQuery` runs inside `shouldSkipProcessor`'s `noneMatch` (not as a shard filter) a null/consumed result beside a readable hybrid does not hide it — the match_none/null-instance 500 the IT exercises is genuinely fixed.
> 
> **One thing I'd flag before claiming the partial-reduce case is closed (MED, see inline):** the new `isHybridQuery` comment states a partial reduce that consumed a shard's topDocs is "simply not hybrid" and thus handled — but that's only true for the `shouldSkipProcessor` scan. If a consumed-but-not-`isNull()` result sits _beside_ a readable hybrid, `shouldSkipProcessor` correctly returns false and the processor RUNS, and `NormalizationProcessorWorkflow.filterValidResults` (line ~417) filters only `!isNull()`, NOT `hasConsumedTopDocs()`, so that survivor reaches `getQueryTopDocs` (line ~224) / `unprocessedDocIds` (line ~400), which call `topDocs()` → the same `IllegalStateException("topDocs already consumed")`, just relocated downstream. Grounded in core: `isNull` is a final flag set only by `nullInstance()`; `consumeTopDocs()` nulls `topDocsAndMaxScore` but not `isNull`, so a consumed result has `isNull()==false, hasConsumedTopDocs()==true, topDocs() throws`. Both the pre-existing `filterValidResults` comment and this PR's new comment claim to cover the >512-shard partial-reduce case; only the null-instance sub-case actually is. Fix: mirror the same 3-part guard in `filterValidResults` (and don't assume every survivor has readable topDocs), or narrow the comments' claim.
> 
> **PLAUSIBLE (MED if reachable, and a NEW behavior): all-consumed → silent wrong scores.** If _every_ hybrid shard result is consumed at processing time (shard count an exact multiple of the effective batch size → empty tail buffer), `noneMatch(isHybridQuery)` is true → the processor SKIPS → the response carries raw, un-normalized/un-combined scores with no error. Pre-PR this same state threw the 500; post-PR it may silently return wrong scores, which is worse. I couldn't confirm the empty-tail-at-processor-time reachability without a large-shard cluster, but the mechanism is confirmed — worth verifying.
> 
> **Test gaps (MED):**
> 
> 1. The strongest proof of Fix 1's totals-nulling under a real `filter_query` is missing: the IT asserts the derived total only for the _empty_ filter (`{0, eq}`), which is weak (many bugs also yield 0). A non-empty case — the single-doc filter under the counting shape asserting `total {1, eq}` vs the legs' ~12 union — would actually discriminate that the filtered count is used and the leg-derived total was disabled.
> 2. `testFastPath_whenAHybridizationPipelineIsAttached` has only a status-200 + page-stability oracle (no content-correctness check against a control), and unlike the `filter_query` IT it does not `primeThroughPipeline`, so if the fast path never arms (cold gate) round 2 is never `match_none`, no null-instance is produced, and the test passes green without exercising the skip path. Add priming + a correctness oracle (or an assertion that a null-instance shard was actually produced).
> 3. The Fix-2 guard lives on the shared `AbstractScoreHybridizationProcessor`, but only `RRFProcessorTests` covers `shouldSkipProcessor`; `NormalizationProcessorTests` has none. A mirrored unit test would make the "belongs on main too" claim explicit and guard a future override.
> 
> **LOW/nits:** no unit test for a processor that replaces `query()` with a _different_ fused hybrid (same `!= this` branch, minor); the added `import static org.mockito.Mockito.mock;` is out of alphabetical order (after `never`/`verify`) and may trip a spotless/import-order check; `ids(...)` uses an inline fully-qualified `java.util.stream.Collectors.toList()` instead of an import.

Regarding the main concern - in core this is not reachable. In `QueryPhaseResultConsumer.PendingReduces.consumeResult` a partial reduce is triggered by an arriving non-null result, which is then added to the just-cleared buffer, so after any non-null result the buffer holds ≥ 1 unconsumed result until the final `reduce()` — and that final reduce runs in `FetchSearchPhase`, after `onPhaseDone` has run the phase-results processors. An exact multiple of the batch size still leaves the last result unconsumed. With zero non-null results there is nothing hybrid, and skipping is right. (Classic hybrid also never partially reduces, per the filter.)

Ack on test gaps and nits.

bzhangam:
On `post_filter` handling (not a #2021 blocker — a `CandidateScope` design question that predates it):

`post_filter` is classified `PROPAGATED`, so it's pushed to each leg. That's what makes it work on the fast path — the legs filter before returning, so the page is filtered and the window stays full. But the legs apply it *before* the coordinator normalizes, so per-leg min/max is computed over the filtered subset and `post_filter` ends up affecting the fused scores/ranking. That breaks parity with classic hybrid, where `post_filter` runs post-normalization on the coordinator and is score-neutral. Is that divergence intended?

If we instead matched classic and applied `post_filter` post-normalization, how would the fast path handle it? It assembles the page from already-fetched `SearchHit`s (`assemblePage`), and `post_filter` is a Lucene query the coordinator can't evaluate against hits without re-running a shard round (defeating the fast path). So it looks like it'd have to disarm the fast path whenever `post_filter` is present, and the two-round path would need to accept a possibly-short page.

So it's a real tradeoff:
- propagate to legs (today): full page, fast-path-compatible, but diverges from classic and isn't score-neutral;
- post-normalization: matches classic, but the fast path disarms on `post_filter` and pages can come back short.

Could we confirm which we intend and document it? The `CandidateScope` comment justifies propagation only via round-2 score-0 backfill and doesn't mention this parity/score-neutrality tradeoff.

bzhangam (approved):

vibrantvarun (approved):
Re-reviewed the `Address review: refuse partially reduced hybrid results loudly` commit — the finding I raised is resolved, and well.

- **The relocated `topDocs already consumed` 500 is now handled at the source**: `NormalizationProcessorWorkflow.filterValidResults` refuses a consumed-but-not-`isNull()` result with a clear `IllegalStateException` that names the remedy, instead of letting it surface as a cryptic error deep in `getQueryTopDocs`. Throwing (rather than silently dropping) is the right call — dropping would hand the final reduce a mix of normalized and raw delimiter-format documents at HTTP 200. And it's correctly scoped: a legit top-level hybrid never hits it (the filter disables batched reduction for it), and a non-hybrid query behind the pipeline is skipped before the workflow runs, so the throw fires only for the genuinely unsupported case of a request processor creating/replacing the hybrid after the filter.
- **The silent un-normalized-scores risk is closed**: the `isHybridQuery` comment now documents (and you measured, batched_reduce_size=2 on 12 shards) that core buffers the partial-reduce-triggering result and reduces the rest only after the phase-results processors, so at least one readable result always remains — the all-consumed skip window isn't reachable, and any consumed result that does appear is refused loudly rather than skipped.
- **Tests close the gaps I noted**: the workflow refusal is pinned (`testExecute_whenAResultWasConsumedByAPartialReduce_thenRefusesWithAClearMessage`, asserting the message and the named remedy); `NormalizationProcessor` now has its own `shouldSkipProcessor` coverage of the shared guard (not just RRF); the `filter_query` IT asserts the discriminating non-empty derived total (`{value:1, relation:eq}` on the single-doc filter, where the legs' union would otherwise prove `{THRESHOLD, gte}`); and the hybridization-pipeline IT now carries a real correctness oracle (`assertClientVisibleIdentical` vs a two-round control through the same pipeline) and proves the match_none round 2 actually ran by reading the shard fetch-phase counter.

The two headline fixes remain correct (I verified the `isRequestQuery` identity re-derivation and the guard short-circuit in the prior pass), the guard logic is unchanged, and the new throw is confined to the unsupported path — no regression. bzhangam's `post_filter`/normalization-parity note is a real but pre-existing `CandidateScope` design question, out of scope here as they framed it. LGTM.

martin-gaievski:
> On `post_filter` handling (not a #2021 blocker — a `CandidateScope` design question that predates it):
> 
> `post_filter` is classified `PROPAGATED`, so it's pushed to each leg. That's what makes it work on the fast path — the legs filter before returning, so the page is filtered and the window stays full. But the legs apply it _before_ the coordinator normalizes, so per-leg min/max is computed over the filtered subset and `post_filter` ends up affecting the fused scores/ranking. That breaks parity with classic hybrid, where `post_filter` runs post-normalization on the coordinator and is score-neutral. Is that divergence intended?
> 
> If we instead matched classic and applied `post_filter` post-normalization, how would the fast path handle it? It assembles the page from already-fetched `SearchHit`s (`assemblePage`), and `post_filter` is a Lucene query the coordinator can't evaluate against hits without re-running a shard round (defeating the fast path). So it looks like it'd have to disarm the fast path whenever `post_filter` is present, and the two-round path would need to accept a possibly-short page.
> 
> So it's a real tradeoff:
> 
> * propagate to legs (today): full page, fast-path-compatible, but diverges from classic and isn't score-neutral;
> * post-normalization: matches classic, but the fast path disarms on `post_filter` and pages can come back short.
> 
> Could we confirm which we intend and document it? The `CandidateScope` comment justifies propagation only via round-2 score-0 backfill and doesn't mention this parity/score-neutrality tradeoff.

Thanks for raising this, because the conclusion depends on where classic applies post_filter, and I think the premise doesn't hold: classic hybrid also applies post_filter before normalization, on the shard, not post-normalization on the coordinator.

In core, QueryPhase adds createFilteredCollectorContext(searcher, parsedPostFilter().query()) right after the top-docs collector context whenever post_filter is set; that context wraps the inner collector in a FilteredCollector(in,   filterWeight). For a classic hybrid the inner collector is ours — HybridCollectorManager.isHybridFilteredCollector exists precisely to unwrap that FilteredCollector and reach the HybridSearchCollector inside. So each shard's per-sub-query top docs are already post-filtered when they leave the shard, and the normalization processor then computes per-sub-query min/max over those filtered results. There is no post-normalization post_filter step on the coordinator — core never evaluates post_filter there — and it is not score-neutral in classic either: filtered-out documents don't participate in the range.

Fused does the same thing one level up: a leg is an ordinary search, so its shards run exactly that FilteredCollector path, and the coordinator normalizes over the leg's post-filtered window. One thing that genuinely differs is the scope of the range — classic takes it over the union of each shard's post-filtered top-size per sub-query, fused over the leg's post-filtered top-window_size across the index — but that is the general fused/classic normalization difference and holds
with or without post_filter.

So propagating post_filter to the legs is the parity-preserving choice rather than a divergence, and the fast path doesn't need to disarm on it: the page it assembles is filtered the same way classic's shard results are. Alternative that you describe (filter after normalization) would actually be the new behaviour, and it would also have the round-2 problem the CandidateScope comment describes: unfiltered window decimated in round 2 and back-filled with score-0 Tail documents.

Agreed CandidateScope comment justifies propagation only via that round-2 argument; I'll add a sentence noting it also matches classic (whose shard collectors are post-filtered before normalization), will push commit

